PR #13, opened 2020-11-20
Enable external ownership of ConcreteBuffer

yungyuc:
This PR suggests to allow the internal data buffer of `ConcreteBuffer` to be owned outside the shared-pointer-controlled object.  Before the change, creation of `ConcreteBuffer` mandates dynamic memory allocation.

A custom `std::unique_ptr` `Deleter` is added to support memory management of the internal buffer of `ConcreteBuffer`.  The customer `Deleter` uses a polymorphic "remover" to control the lifecycle of the internal buffer.  By default, the remover is null, and the `Deleter` behaves the same as `std::default_delete<T>`.

The `remover` makes it possible to bind the lifecycle of the internal data buffer outside.  For example, `struct ConcreteBufferNdarrayRemover` is derived from `ConcreteBuffer::remover_type` and bind the lifecycle of the buffer to a `PyArrayObject` (i.e., `pybind11::array`).

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -19,7 +19,8 @@ Checks: >
     -llvm-include-order,
     -performance-noexcept-move-constructor,
     -readability-named-parameter,
-    -readability-magic-numbers
+    -readability-magic-numbers,
+    -misc-non-private-member-variables-in-classes
 WarningsAsErrors: ''
 HeaderFilterRegex: 'modmesh.*'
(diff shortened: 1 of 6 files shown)


yungyuc:
@tigercosmos This may be an example for the 'system-level' work that should be done for the mesh code base.  For your information.  Feel free to comment.

PR #31, opened 2021-12-26
Add basic code to support multi-dimensional (2/3D) unstructured mesh

yungyuc:
Add multi-dimensional (2/3D) unstructured mesh object without any verification

The mesh is named StaticMesh* because it does not change when being used.

1. The code was copied from SOLVCON (https://github.com/solvcon/solvcon).
2. Add basic shape tests for the mesh object.
3. Add CellType but it is not used nor tested at all.


yungyuc:
This should be merged to the project branch `project/ustmesh` (not `master`).

PR #61, opened 2022-04-12
Fix viewer CMakeLists.txt

yanagiragi:
This PR aim to solve the raised error [^1] when trying to build viewer

```bash
k45vs@K45VS:$ CMAKE_PREFIX_PATH=/home/k45vs/Tools/Qt-6.2.4/6.2.4/gcc_64/lib/cmake make
k45vs@K45VS:$ cd build/dev38
k45vs@K45VS:$ make

In file included from /home/k45vs/Desktop/modmesh/src/viewer/modmesh/viewer/RMainWindow.cpp:32:
/home/k45vs/Desktop/modmesh/src/viewer/modmesh/viewer/PythonInterpreter.hpp:38:7: error: ‘modmesh::PythonInterpreter’ declared with greater visibility than the type of its field ‘modmesh::PythonInterpreter::m_interpreter’ [-Werror=attributes]
   38 | class PythonInterpreter
```

To solve this I copy the CXX_VISIBILITY flag settings in other CMakeLists.txt. [^2]

Also, I've set default value of  `HIDE_SYMBOL` to `ON` since for now the only use case it is set to off is to combine it when `BUILD_QT=OFF`. However in current Makefile it is default set to `BUILD_QT=ON`, so I assume by default the Makefile is targeting to build both modmesh and the viewer without providing additional arguments.

[^1]: https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes
[^2]: https://github.com/solvcon/modmesh/blob/master/src/pymod/CMakeLists.txt#L16


yungyuc:
Thanks.  It works pretty well.  It would be more convenient for me to check your code if the branch name differed from `master`.

yanagiragi:
Good suggestion. I forget to create another branch this time.

Issue #57, opened 2022-04-05
Fix unwanted overlay in the viewer


yungyuc:
When the mesh is shown in the viewer in the first time, everything looks OK:

![image](https://user-images.githubusercontent.com/399122/161776685-b2c8a2c6-c865-4754-a26d-7f40b0a3bb3e.png)

But showing a (slightly) different mesh using `modmesh.view.show()` does not remove the first mesh and have the first and second overlaid in the widget:

![image](https://user-images.githubusercontent.com/399122/161776883-f001a71c-2d0a-4f54-8c16-0d47e69eab79.png)

There should be a way to avoid the unwanted overlay.

yungyuc:
After merging #63 the camera controller stops working.

yanagiragi:
@yungyuc You're right. It may takes sometimes to figure out the reason, perhaps it is due to the camera controller is worked as an component of the mesh rather than individual nodes.

Perhaps we should first revert to the commit before merging #63 ?

yungyuc:
I am guessing the same thing and investigating.  I think #63 is OK.  Deleting child entities should not break camera controller like that so I should find the root cause and fix it.

yungyuc:
I hope #64 is the last piece of this work.  If we do not find other bugs, this issue can be closed later.

PR #62, opened 2022-05-01
fix: Clear scene before assign new mesh

yanagiragi:
This PR aims to solve https://github.com/solvcon/modmesh/issues/57 by clear the scene nodes of the viewer each time when the `Run` button is clicked.


yanagiragi:
I discover a typo error in branch name, will fix and re-open another PR.

PR #63, opened 2022-05-01
fix: Clear scene before assign new mesh

yanagiragi:
This PR aims to solve https://github.com/solvcon/modmesh/issues/57 by clear the scene nodes of the viewer each time when the Run button is clicked.

### DIFF
--- a/src/viewer/modmesh/viewer/PythonInterpreter.cpp
+++ b/src/viewer/modmesh/viewer/PythonInterpreter.cpp
@@ -133,6 +133,11 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
             [](std::shared_ptr<StaticMesh2d> const & mesh)
             {
                 RApplication * app = RApplication::instance();
+                RScene * scene = app->main()->viewer()->scene();
+                for (RStaticMesh<2> * child : scene->findChildren<RStaticMesh<2> *>())
+                {
+                    delete child;
+                }
                 new RStaticMesh<2>(mesh, app->main()->viewer()->scene());
             });
 }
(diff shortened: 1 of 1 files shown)


src/viewer/modmesh/viewer/PythonInterpreter.cpp
```diff
@@ -133,7 +133,12 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
             [](std::shared_ptr<StaticMesh2d> const & mesh)
             {
                 RApplication * app = RApplication::instance();
-                new RStaticMesh<2>(mesh, app->main()->viewer()->scene());
+                RScene * scene = app->main()->viewer()->scene();
+                for (RStaticMesh<2> * child : scene->findChildren<RStaticMesh<2> *>())
```
yungyuc, line 137: I read about `qDeleteAll()` for children removal.  I keep the loop because I do not know which way is better.

src/viewer/modmesh/viewer/PythonInterpreter.cpp
```diff
@@ -133,7 +133,12 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
...
+                {
+                    delete child;
+                }
+                new RStaticMesh<2>(mesh, scene);
```
yungyuc, line 141: The `scene` variable is already there for the pointer so I reuse it.

yungyuc (approved):

src/viewer/modmesh/viewer/PythonInterpreter.cpp
yanagiragi: Sure. I agree it is better to reuse it.

PR #64, opened 2022-05-03
Correct the wrong deletion of the controller of scene

yungyuc:
The camera controller of scene is also a child of it.  For some reason, `QObject::findChildren<RStaticMesh<ND> *>()` does not only return `RStaticMesh<ND> *`, but also the controller.  Consequently the controller gets deleted when "show" is called to replace the `RStaticMesh`.

The fix uses a loop through `QNode::childNodes()` and RTTI in the loop to correctly delete only `RStaticMesh<ND>`.

This is to fix the issue introduced in #63 , and eventually addresses #57 .

### DIFF
--- a/src/viewer/modmesh/viewer/PythonInterpreter.cpp
+++ b/src/viewer/modmesh/viewer/PythonInterpreter.cpp
@@ -134,9 +134,12 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
             {
                 RApplication * app = RApplication::instance();
                 RScene * scene = app->main()->viewer()->scene();
-                for (RStaticMesh<2> * child : scene->findChildren<RStaticMesh<2> *>())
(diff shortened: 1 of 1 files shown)


src/viewer/modmesh/viewer/PythonInterpreter.cpp
```diff
@@ -134,9 +134,12 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
             {
                 RApplication * app = RApplication::instance();
                 RScene * scene = app->main()->viewer()->scene();
-                for (RStaticMesh<2> * child : scene->findChildren<RStaticMesh<2> *>())
```
yungyuc, line 137: The `child` includes the camera controller which gets deleted although it shouldn't be.

yungyuc:
@yanagiragi FYI.

Issue #60, opened 2022-04-06
Add a 3D example mesh


yungyuc:
The unit test has a 2D example mesh: https://github.com/solvcon/modmesh/blob/dc0a9f6808e85264995054323a6aea75d724913f/tests/test_mesh.py#L82 but it does not have a 3D mesh.  A 3D example mesh should be created in the unit test.

The new prototype viewer should help creating and inspecting the mesh.

PR #65, opened 2022-05-04
Add a single tetrahedron mesh for testing 3D mesh code

yungyuc:
Add a single tetrahedron to show that the 3D mesh code basically works, although may have some issues that are not known.

This is to address issue #60.


yungyuc:
@yanagiragi Do you want to take a look before I merge?

yanagiragi:
LGTM.

Issue #75, opened 2022-05-17
Add viewer build binary in CI


tai271828:
Per discussion on 220517 we all agree to do this. See https://hackmd.io/ZEN4vslFSs-40rGKaIgA2g?view

tai271828:
Per this discussion https://github.com/solvcon/modmesh/discussions/53#discussioncomment-2781471 and the support of Ubuntu Jammy on Github https://github.com/actions/virtual-environments . I think we are ready to move forward because I can build the viewer binary with `devevn` (which helps automation) on my x86 machine running Ubuntu Jammy.

tai271828:
I will try to move forward a bit this weekend. Fun weekend project :star_struck:

PR #76, opened 2022-05-23
ci(viewer): init CI run for viewer binary

tai271828:
Beginning to build the `viewer` binary with Ubuntu Jammy (22.04). The CI job will pass with Ubuntu, and will fail with macos. I leave the fix for macos to anyone with mac. 

Ref issue #75

### DIFF
--- a/.github/workflows/modmesh_viewer.yml
+++ b/.github/workflows/modmesh_viewer.yml
@@ -1,0 +1,95 @@
+name: modmesh_viewer
+
+on:
+  push:
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+        matrix:
+          os: [ubuntu-22.04, macos-11]
+          cmake_build_type: [Release, Debug]
+
+        fail-fast: false
+
+    steps:
+
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+
+    - name: dependency (ubuntu)
+      if: matrix.os != 'macos-11'
+      run: |
+        sudo apt-get -qqy update
+        sudo apt-get -qy install \
+            sudo curl git build-essential make cmake libc6-dev gcc-10 g++-10 \
+            python3 python3-dev python3-venv \
+            qt6-3d-dev
+        # TODO: not figure out why this is needed. Without this package, we do
+        # not make cmake navigate to the right path to get the qt lib
+        sudo apt-get -qy install \
+            libgl1-mesa-dev
+
+        sudo pip3 install numpy pytest flake8
+
+    - name: dependency (macos)
+      if: matrix.os == 'macos-11'
+      run: |
+        brew install python3 llvm
+        ln -s "$(brew --prefix llvm)/bin/clang-format" "/usr/local/bin/clang-format"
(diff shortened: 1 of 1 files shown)


.github/workflows/modmesh_viewer.yml
```diff
@@ -0,0 +1,95 @@
...
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
+    - name: viewer
+      run: |
+        make viewer \
+          VERBOSE=1 BUILD_QT=ON \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
+    - name: flake8
```
yungyuc: We do not need flake8 for this action.

.github/workflows/modmesh_viewer.yml
```diff
@@ -0,0 +1,95 @@
...
+  push:
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+        matrix:
+          os: [ubuntu-22.04, macos-11]
```
yungyuc: `macos-11` does not work: https://github.com/tai271828/modmesh/runs/6560081058?check_suite_focus=true#step:7:70 .

.github/workflows/modmesh_viewer.yml
```diff
@@ -0,0 +1,95 @@
...
+            python3 python3-dev python3-venv \
+            qt6-3d-dev
+        # TODO: not figure out why this is needed. Without this package, we do
+        # not make cmake navigate to the right path to get the qt lib
+        sudo apt-get -qy install \
+            libgl1-mesa-dev
+
+        sudo pip3 install numpy pytest flake8
+
+    - name: dependency (macos)
```
yungyuc: This is not needed because macos does not build with Qt.

.github/workflows/modmesh_viewer.yml
```diff
@@ -0,0 +1,95 @@
...
+        echo "flake8 version: $(flake8 --version)"
+
+    - name: buildext
+      run: |
+        make buildext \
+          VERBOSE=1 BUILD_QT=ON \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
+    - name: pytest
```
yungyuc, line 66: We do not need pytest.  I think we should make a new unit test runner different from this in the future.

yungyuc (changes requested):
Great addition!  Could you fix the failed run and do some minor tweaks?

yungyuc:
@ianlienfa FYI

.github/workflows/modmesh_viewer.yml
yungyuc: I am keeping pytest because this uses a new OS version that is not included in other actions.

yungyuc (approved):

tai271828:
Thanks for the follow-up fixup. I went to a holiday for several days without accessing internet! I hope everyone does not mind the late response.

yungyuc:
No problem at all.

Issue #69, opened 2022-05-06
Add axis marker in the view window


yungyuc:
The view widget should use an axis marker.

yungyuc:
Now we have a simple axis marker.  It will need a lot of improvements, e.g., add options for using arrow heads, making it to stay in the scene, etc.  Create new issues for the improvements.

PR #77, opened 2022-05-25
Add a simple axis marker in the view widget

yungyuc:
This is to address #69 .  This is a really simple axis marker consisting of 3 line segments of unit length (1).  The mark is an entity in the view widget and uses fixed coordinate.

### DIFF
--- a/cpp/binary/viewer/viewer.cpp
+++ b/cpp/binary/viewer/viewer.cpp
@@ -31,16 +31,6 @@
 
 #include <modmesh/view/view.hpp>
 
-int main(int argc, char ** argv)
-{
-    using namespace modmesh;
-
-    RApplication app(argc, argv);
-    app.main()->resize(1000, 600);
-
-    return app.exec();
-}
-
 namespace modmesh
 {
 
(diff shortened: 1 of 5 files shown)


cpp/binary/viewer/viewer.cpp
```diff
@@ -50,6 +40,19 @@ namespace python
 namespace detail
 {
 
+static void show_mark()
+{
+    RScene * scene = RApplication::instance()->main()->viewer()->scene();
+    for (Qt3DCore::QNode * child : scene->childNodes())
```
yungyuc, line 46: For now, I only need one marker in a view.

cpp/modmesh/view/RAxisMark.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <Qt3DCore/QGeometry>
+#include <Qt3DCore/QAttribute>
+#include <Qt3DCore/QTransform>
+
+#include <Qt3DExtras/QDiffuseSpecularMaterial>
+
+namespace modmesh
+{
+
+class RLine
```
yungyuc, line 53: A simple line segment.

cpp/modmesh/view/RAxisMark.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+private:
+
+    Qt3DCore::QGeometry * m_geometry = nullptr;
+    Qt3DRender::QGeometryRenderer * m_renderer = nullptr;
+    Qt3DExtras::QDiffuseSpecularMaterial * m_material = nullptr;
+
+}; /* end class RLine */
+
+class RAxisMark
```
yungyuc, line 69: The axis marker.  I prefer to keep it shorter so it's named as "mark" not "marker".

Issue #78, opened 2022-05-26
Add a feature to screenshot


yungyuc:
The screenshot feature should allow:

1. Shot the full or part of the window and/or 3D view.
2. Copy the shot to clipboard.  This helps paste online.
3. Save the shot to a (compressed and uncompressed) bitmap file.

PR #87, opened 2022-06-04
Take the pixmap from the 3D window

yungyuc:
Add two Python helper functions for the pixmap:

1. clipImage(): send the pixmap to clipboard.
2. saveImage(): save the pixmap to a file.


yungyuc:
Fixes #78

Issue #59, opened 2022-04-05
[viewer] Add menubar


yungyuc:
Add basic menubar to the viewer.

yanagiragi:
Shouldn't be hard. Please let me have a try.

yungyuc:
mac may have some special behaviors.  See https://doc.qt.io/qt-6/qmenubar.html#qmenubar-as-a-global-menu-bar and #90 for more information.

yungyuc:
I forgot to close this issue which is considered as done.

PR #89, opened 2022-06-10
feat: Support menubar

yanagiragi:
Solves https://github.com/solvcon/modmesh/issues/59 and add basic example:

```
File
├─ New
│ └─ New File
└─ Exit
```


cpp/modmesh/view/RApplication.cpp
```diff
@@ -43,6 +43,27 @@ RApplication::RApplication(int & argc, char ** argv)
...
+    RMenu * newMenu = new RMenu(QString("New"));
+    RAction * newFileAction = new RAction(
+        QString("New file"),
+        QString("Create new file"),
+        []()
+        {
+            qDebug() << "Create new file!";
+        });
+    RAction * exitAction = new RAction(
+        QString("Exit"),
```
yungyuc, line 57: Interestingly, on mac, when this menu item string is `Exit`, the item does not show.  It shows after the string changes to something else.  I will study why it happens and fix it later.

yungyuc (approved):

cpp/modmesh/view/RApplication.cpp
yanagiragi: To be clear, the case we discovered should be the menu item string **starts with `Exit`**, not sure the contains `Exit` will also fail or not.

PR #90, opened 2022-06-12
Do not add "exit" menu item when it is macos

yungyuc:
See https://doc.qt.io/qt-6/qmenubar.html#qmenubar-as-a-global-menu-bar . When it is macos, Qt search for "quit" or "exit" menu item and merge them with the corresponding item in the default menu. It does not need to be added by the app.

Also fix a typo for `fileMenu`.

### DIFF
--- a/cpp/modmesh/view/RApplication.cpp
+++ b/cpp/modmesh/view/RApplication.cpp
@@ -44,24 +44,33 @@ RApplication::RApplication(int & argc, char ** argv)
     pybind11::exec("modmesh.view = _modmesh_view");
 
     RMenuBar * menuBar = new RMenuBar();
-    RMenu * fileMmenu = new RMenu(QString("File"));
+    RMenu * fileMenu = new RMenu(QString("File"));
     RMenu * newMenu = new RMenu(QString("New"));
+
     RAction * newFileAction = new RAction(
         QString("New file"),
         QString("Create new file"),
         []()
         {
+            // FIXME: This is only a demo.
             qDebug() << "Create new file!";
         });
+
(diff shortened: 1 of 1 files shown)


cpp/modmesh/view/RApplication.cpp
```diff
@@ -44,24 +44,33 @@ RApplication::RApplication(int & argc, char ** argv)
     pybind11::exec("modmesh.view = _modmesh_view");
 
     RMenuBar * menuBar = new RMenuBar();
-    RMenu * fileMmenu = new RMenu(QString("File"));
+    RMenu * fileMenu = new RMenu(QString("File"));
```
yungyuc, line 47: @yanagiragi I think this is a typo?

cpp/modmesh/view/RApplication.cpp
```diff
@@ -44,24 +44,33 @@ RApplication::RApplication(int & argc, char ** argv)
...
         []()
         {
+            // FIXME: This is only a demo.
             qDebug() << "Create new file!";
         });
+
+    newMenu->addAction(newFileAction);
+    fileMenu->addMenu(newMenu);
+
+#ifndef Q_OS_MACOS
```
yungyuc, line 62: Only build this when this is not in mac.

yungyuc (commented):
@yanagiragi Let me know if this looks good to you.

cpp/modmesh/view/RApplication.cpp
yanagiragi: Yes. I believe this is a typo.

yungyuc:
ref #59

PR #95, opened 2022-06-26
Refactor: Embed the base Python extension module to the viewer executable

yungyuc:
The "base" _modmesh Python extension module was built in a separate shared object and loaded into the viewer binary.  Some object files need to be linked in the two images.  It is unnecessarily complex.  Embedding the "base" Python extension module in the same way that the "_modmesh_view" module is embedded to the viewer executable simplifies the linkage.

The pure Python module (package) `modmesh` is changed accordingly.  When it is used without the Qt viewer, it still tries to load the `_modmesh` extension module relatively.


cpp/binary/viewer/viewer.cpp
```diff
@@ -26,14 +26,20 @@
...
-#include <modmesh/python/wrapper/view/view.hpp> // Must be the first include.
+#include <modmesh/python/python.hpp> // Must be the first include.
+#include <modmesh/python/wrapper/modmesh/modmesh.hpp>
+#include <modmesh/python/wrapper/view/view.hpp>
 #include <modmesh/modmesh.hpp>
-
 #include <modmesh/view/view.hpp>
 
-PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod)
+PYBIND11_EMBEDDED_MODULE(_modmesh, mod) // NOLINT
```
yungyuc, line 35: Embed `_modmesh` to the viewer executable.

cpp/modmesh/python/common.hpp
```diff
@@ -385,6 +387,92 @@ WrapBase
...
+
+public:
+
+    OneTimeInitializer(OneTimeInitializer const &) = delete;
+    OneTimeInitializer(OneTimeInitializer &&) = delete;
+    OneTimeInitializer & operator=(OneTimeInitializer const &) = delete;
+    OneTimeInitializer & operator=(OneTimeInitializer &&) = delete;
+    ~OneTimeInitializer() = default;
+
+#ifdef IAMWWRONG
```
yungyuc: Dead code that I forgot to delete.  Will delete it in the next commit.

Issue #52, opened 2022-03-27
Set data from ndarray memory buffer to SimpleArray


yungyuc:
Provide better support for the Python wrapper of the class template `SimpleArray` to get the data [ndarray](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html) through the memory buffer.

```pycon
>>> import modmesh as mm
>>> import numpy as np
>>> sarr = mm.SimpleArrayFloat64((2, 3, 4))
>>> ndarr = np.arange(2*3*4, dtype='float64').reshape((2, 3, 4))
```

Using Python [ellipsis](https://docs.python.org/dev/library/constants.html#Ellipsis) should [copy the data in the memory buffer](https://numpy.org/doc/stable/user/basics.indexing.html?highlight=ellipsis#slicing-and-striding) from `ndarr` to `sarr`, but it does not work yet:

```pycon
>>> sarr[...] = ndarr[...]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __setitem__(): incompatible function arguments. The following argument types are supported:
    1. (self: modmesh._modmesh.SimpleArrayFloat64, arg0: int, arg1: float) -> None
    2. (self: modmesh._modmesh.SimpleArrayFloat64, arg0: List[int], arg1: float) -> None

Invoked with: <modmesh._modmesh.SimpleArrayFloat64 object at 0x104196bb0>, Ellipsis, array([[[ 0.,  1.,  2.,  3.],
        [ 4.,  5.,  6.,  7.],
        [ 8.,  9., 10., 11.]],

       [[12., 13., 14., 15.],
        [16., 17., 18., 19.],
        [20., 21., 22., 23.]]])
```

The range indexing (slice) should work in a similar way:

```pycon
>>> sarr[:,:,:] = ndarr[:,:,:]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __setitem__(): incompatible function arguments. The following argument types are supported:
    1. (self: modmesh._modmesh.SimpleArrayFloat64, arg0: int, arg1: float) -> None
    2. (self: modmesh._modmesh.SimpleArrayFloat64, arg0: List[int], arg1: float) -> None

Invoked with: <modmesh._modmesh.SimpleArrayFloat64 object at 0x104196bb0>, (slice(None, None, None), slice(None, None, None), slice(None, None, None)), array([[[ 0.,  1.,  2.,  3.],
[...]

PR #96, opened 2022-06-28
SimpleArray: support copy data from ndarray using ellipsis

tigercosmos:
solve the first part of #52

### DIFF
--- a/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
@@ -149,7 +149,9 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def(
                 "__setitem__",
                 [](wrapped_type & self, const py::ellipsis &, pybind11::array & arr_in)
-                { copy_array_using_ellipsis(self, arr_in); })
+                {
+                    broadcast_array_using_ellipsis(self, arr_in);
+                })
             .def(
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
@@ -161,15 +163,59 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             ;
     }
 
-    static void copy_array_using_ellipsis(wrapped_type & self, pybind11::array & arr_in)
+    template <typename D /* for destination type */>
+    struct TypeBroadCast
     {
-        const auto left_shape = self.shape();
+        static void broadcast(wrapped_type & arr_out, pybind11::array & arr_in)
+        {
+            using shape_type = small_vector<size_t>;
+            using out_type = typename std::remove_reference<decltype(arr_out[0])>::type;
+
+            auto * arr_new = reinterpret_cast<pybind11::array_t<D> *>(&arr_in);
+
+            shape_type sidx_init(arr_out.ndim());
+
+            for (size_t i = 0; i < arr_out.ndim(); ++i)
+            {
+                sidx_init[i] = 0;
+            }
+
+            std::function<void(shape_type, int)> copy_idx;
+            copy_idx = [&](shape_type sidx, int dim)
+            {
+                if (dim < 0)
+                    return;
+
+                for (size_t i = 0; i < arr_out.shape(dim); ++i)
+                {
+                    sidx[dim] = i;
+
+                    size_t offset_in = 0;
+                    for (pybind11::ssize_t it = 0; it < arr_in.ndim(); ++it)
+                    {
+                        offset_in += arr_in.strides(it) / arr_in.itemsize() * sidx[it];
+                    }
+
+                    const D * ptr_in = arr_new->data();
+                    arr_out.at(sidx) = static_cast<out_type>(*(ptr_in + offset_in));
+                    copy_idx(sidx, dim - 1);
+                }
+            };
+
+            copy_idx(sidx_init, arr_out.ndim() - 1);
+        }
+    };
+
+    static void broadcast_array_using_ellipsis(
+        wrapped_type & arr_out, pybind11::array & arr_in)
+    {
+        const auto left_shape = arr_out.shape();
         const auto * right_shape = arr_in.shape();
 
-        auto throw_error = [=]()
+        auto throw_shape_error = [&]()
         {
             std::ostringstream msg;
-            msg << "Input array from shape(";
+            msg << "Broadcast input array from shape(";
             for (pybind11::ssize_t i = 0; i < arr_in.ndim(); ++i)
             {
                 msg << right_shape[i];
@@ -179,10 +225,10 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 }
             }
             msg << ") into shape(";
-            for (size_t i = 0; i < self.ndim(); ++i)
+            for (size_t i = 0; i < arr_out.ndim(); ++i)
             {
                 msg << left_shape[i];
-                if (i != self.ndim() - 1)
+                if (i != arr_out.ndim() - 1)
                 {
                     msg << ", ";
                 }
@@ -192,20 +238,67 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             throw std::runtime_error(msg.str());
         };
 
-        if (self.ndim() != static_cast<size_t>(arr_in.ndim()))
+        if (arr_out.ndim() != static_cast<size_t>(arr_in.ndim()))
         {
-            throw_error();
+            throw_shape_error();
         }
 
         for (size_t i = 0; i < left_shape.size(); ++i)
         {
             if (left_shape[i] != static_cast<size_t>(right_shape[i]))
             {
-                throw_error();
+                throw_shape_error();
             }
         }
 
-        std::memcpy(self.data(), arr_in.data(), arr_in.nbytes());
+        if (dtype_is_type<bool>(arr_in))
+        {
+            TypeBroadCast<bool>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<int8_t>(arr_in))
+        {
+            TypeBroadCast<int8_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<int16_t>(arr_in))
+        {
+            TypeBroadCast<int16_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<int32_t>(arr_in))
+        {
+            TypeBroadCast<int32_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<int64_t>(arr_in))
+        {
+            TypeBroadCast<int64_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<uint32_t>(arr_in))
+        {
+            TypeBroadCast<uint32_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<uint16_t>(arr_in))
+        {
+            TypeBroadCast<uint16_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<uint32_t>(arr_in))
+        {
+            TypeBroadCast<uint32_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<uint64_t>(arr_in))
+        {
+            TypeBroadCast<uint64_t>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<float>(arr_in))
+        {
+            TypeBroadCast<float>::broadcast(arr_out, arr_in);
+        }
+        else if (dtype_is_type<double>(arr_in))
+        {
+            TypeBroadCast<double>::broadcast(arr_out, arr_in);
+        }
+        else
+        {
+            throw std::runtime_error("input array data type not support!");
+        }
     }
 
     static shape_type make_shape(pybind11::object const & shape_in)
--- a/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
@@ -147,6 +147,10 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 [](wrapped_type & self, std::vector<ssize_t> const & key, T val)
                 { self.at(key) = val; })
             .def(
+                "__setitem__",
+                [](wrapped_type & self, const py::ellipsis &, pybind11::array & arr_in)
+                { copy_array_using_ellipsis(self, arr_in); })
+            .def(
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
                 { return self.reshape(make_shape(shape)); })
@@ -155,6 +159,53 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def_property_readonly("nbody", &wrapped_type::nbody)
             //
             ;
+    }
+
+    static void copy_array_using_ellipsis(wrapped_type & self, pybind11::array & arr_in)
+    {
+        const auto left_shape = self.shape();
+        const auto * right_shape = arr_in.shape();
+
+        auto throw_error = [=]()
+        {
+            std::ostringstream msg;
+            msg << "Input array from shape(";
+            for (pybind11::ssize_t i = 0; i < arr_in.ndim(); ++i)
+            {
+                msg << right_shape[i];
+                if (i != arr_in.ndim() - 1)
+                {
+                    msg << ", ";
+                }
+            }
+            msg << ") into shape(";
+            for (size_t i = 0; i < self.ndim(); ++i)
+            {
+                msg << left_shape[i];
+                if (i != self.ndim() - 1)
+                {
+                    msg << ", ";
+                }
+            }
+            msg << ")";
+
+            throw std::runtime_error(msg.str());
+        };
+
+        if (self.ndim() != static_cast<size_t>(arr_in.ndim()))
+        {
+            throw_error();
+        }
+
+        for (size_t i = 0; i < left_shape.size(); ++i)
+        {
+            if (left_shape[i] != static_cast<size_t>(right_shape[i]))
+            {
+                throw_error();
+            }
+        }
+
+        std::memcpy(self.data(), arr_in.data(), arr_in.nbytes());
     }
 
     static shape_type make_shape(pybind11::object const & shape_in)
--- a/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
@@ -165,9 +165,9 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
 
     struct TypeBroadCast
     {
-        static void check_shape(wrapped_type & arr_out, pybind11::array & arr_in)
+        static void check_shape(wrapped_type const & arr_out, pybind11::array const & arr_in)
         {
-            const auto left_shape = arr_out.shape();
+            const auto & left_shape = arr_out.shape();
             const auto * right_shape = arr_in.shape();
 
             if (arr_out.ndim() != static_cast<size_t>(arr_in.ndim()))
@@ -184,7 +184,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             }
         }
 
-        static void broadcast(wrapped_type & arr_out, pybind11::array & arr_in)
+        static void broadcast(wrapped_type & arr_out, pybind11::array const & arr_in)
         {
             if (dtype_is_type<bool>(arr_in))
             {
@@ -236,7 +236,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             }
         }
 
-        static void throw_shape_error(wrapped_type & arr_out, pybind11::array & arr_in)
+        static void throw_shape_error(wrapped_type const & arr_out, pybind11::array const & arr_in)
         {
             const auto & left_shape = arr_out.shape();
             const auto * right_shape = arr_in.shape();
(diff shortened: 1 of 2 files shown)


cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -146,6 +146,17 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "__setitem__",
                 [](wrapped_type & self, std::vector<ssize_t> const & key, T val)
                 { self.at(key) = val; })
+            .def(
+                "__setitem__",
+                [](wrapped_type & self, py::ellipsis, pybind11::array & arr_in)
+                {
+                    auto arr_in_size = static_cast<size_t>(arr_in.nbytes());
+                    if (self.nbytes() != arr_in_size)
```
yungyuc: Please add a python test for this exception.

yungyuc (changes requested):

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: done

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -146,6 +146,17 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
                 [](wrapped_type & self, std::vector<ssize_t> const & key, T val)
                 { self.at(key) = val; })
+            .def(
+                "__setitem__",
+                [](wrapped_type & self, const py::ellipsis &, pybind11::array & arr_in)
+                {
+                    auto arr_in_size = static_cast<size_t>(arr_in.nbytes());
+                    if (self.nbytes() != arr_in_size)
+                    {
+                        throw std::out_of_range("Array sizes mismatch");
```
yungyuc: The error message should include the mismatch shapes, including the input and self.

yungyuc (changes requested):

yungyuc:
If the added patch is clear enough it is not necessary to squash.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: done

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +161,53 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+
+            throw std::runtime_error(msg.str());
+        };
+
+        if (self.ndim() != static_cast<size_t>(arr_in.ndim()))
+        {
+            throw_error();
+        }
+
+        for (size_t i = 0; i < left_shape.size(); ++i)
```
yungyuc: `SimpleArrayFloat64((2, 3, 4))[...] = np.ones(24, dtype='float64')[:]` should work, but this shape test seems to disallow it.  Could you confirm and fix?

I think we should not check for the shape but should check for total element size and item size.

Please also test for `nghost != 0` (see https://github.com/solvcon/modmesh/blob/3b769a47941f5fb2b726de8bb8fca1e6b3f09ea8/tests/test_buffer.py#L254).  The ellipsis syntax should populate everything including the ghost elements.

The full slice (`[:]`) should also cover everything including ghost.  It should use a separate PR, but when the slice works, we may use `[0:]` to populate only the body (excluding ghost).

Note that the ghost index only works in the leading index and the negativity works differently from numpy negative index.

yungyuc (changes requested):

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: The following numpy example gets `ValueError: could not broadcast input array from shape (24,) into shape (2,4,3)`. If we want to follow the behavior of numpy, we should only check the shape.
```
ndarr2 = np.arange(2*3*4, dtype='float64').reshape((2, 4, 3))
ndarr4 = np.ones(24, dtype='float64')[:]
ndarr2[...] = ndarr4[...]
```

tigercosmos: Could you explain what is a ghost? is there any reference?

yungyuc: It's good to follow numpy broadcasting rule.  Let's continue with the shape check.

The test case I posted contains the working example of the ghost.  Here I repost for clarity:

https://github.com/solvcon/modmesh/blob/3b769a47941f5fb2b726de8bb8fca1e6b3f09ea8/tests/test_buffer.py#L254

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +161,53 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        {
+            if (left_shape[i] != static_cast<size_t>(right_shape[i]))
+            {
+                throw_error();
+            }
+        }
+
+        std::memcpy(self.data(), arr_in.data(), arr_in.nbytes());
```
tigercosmos: I found that we also need to support broadcasting array with different types, which is implemented in Numpy:

```py
ndarr = np.arange(2*3*4, dtype='float32').reshape((2, 3, 4))
ndarr2 = np.arange(2*3*4, dtype='int32').reshape((2, 3, 4))
ndarr[...] = ndarr2[...]
```

I try to implement this:
```cpp
if  (dtype_is_type<T>(arr_in))
{
  std::memcpy(self.data(), arr_in.data(), arr_in.nbytes());
}
else
{
  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
  const auto * ptr = reinterpret_cast<const int*>(arr_in.data()); // should decide the type from arr_in

  for (size_t size = 0; size < self.size(); ++size)
  {
      T * it = self.begin() + size;
      *it = *(ptr + size);
  }
}
```
Now I just use `reinterpret_cast<const int*>` for demo, but I found that I cannot get the data type directly from `arr_in`.
I can get the type name by `pybind11::array::dtype()`, so it's possible to write the code like the following which is ugly:
```cpp
if(arr_in.dtype() == "int32") {
  const auto * ptr = reinterpret_cast<const int*>(arr_in.data()); 

  for (size_t size = 0; size < self.size(); ++size)
  {
      T * it = self.begin() + size;
      *it = *(ptr + size);
  }  
} else if (arr_in.dtype() == "float32") {
  const auto * ptr = reinterpret_cast<const float*>(arr_in.data());
...
```
Is there a better way to achieve it?

yungyuc: I would be surprised if `reinterpret_cast<float *>(pointer_to_int)` work for the type casting.  It should need a `static_cast`.  I am not sure how numpy implements the broadcasting.  But perhaps you can reference its approach?

I would guess broadcasting support is significantly more complex than ellipsis.  It might be a good idea to use a separate PR for it.

yungyuc: Sorry I misinterpret your intention of using `reinterpret_cast`.  It's for interpreting the input ndarray and it's fine.

yungyuc: In addition to learning from numpy, perhaps you can also try to build a mapping table for the type broadcasting.  A class or class template might probably help implementation:

```cpp
template <typename D /* for destination type */> struct TypeBroadCast
{

// Populating the map should be straight-forward.
std::map<std::string /* dtype string */, std::type_info> m_type_map;

D * broadcast(pybind11::array const & arr)
{

// I don't know how to implement the conversion from dtype to static type D.

}

};
```

tigercosmos: The problem is when the `__setitem__` is called, the parameter is `pybind11::array` ([ref](https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L680)) which doesn't return any type information so that I cannot use decltype like `reinterpret_cast<decltype(arr_in.sth)>`,
So now maybe the ugly implementation (as the example above) could be a workaround and we can improve it in the future?

yungyuc: We can improve it in the future, but do not need a workaround at the moment.  The type broadcasting is not part of the original request.  We can (and maybe should) still do it, but it may be done using a separate PR.

pybind11 should also contain code for the same purpose.  Perhaps you can take a look at [pybind11::array_t](https://github.com/pybind/pybind11/blob/479e9a50f38cfa8e525a697d7f5d850f851223b5/include/pybind11/numpy.h#L1016).

tigercosmos: `pybind11:array_t<T>` do have the type information, however `__setitem__` only provide `pybind11:array` at the moment. `array_t` is inherited from `array`

it's possible to do this but have the same issue (I mean, ugly, and even harder to maintain code)
```
            .def(
                "__setitem__",
                [](wrapped_type & self, const py::ellipsis &, pybind11::array_t<int> & arr_in)
                ...
            .def(
                "__setitem__",
                [](wrapped_type & self, const py::ellipsis &, pybind11::array_t<float> & arr_in)
                ...
  ```           
  
  I would try to use the aforementioned mapping class. I will only implement the basic things and leave the remaining for the next PR.

Comment (yungyuc) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: That's not what I meant.  To convert a `numpy.ndarray` (which stores the element type information in `dtype` object) to `pybind11::array_t` and fill in the type in the static type argument `T`, pybind11 detects the type information from the `dtype` object.  It is what you want to do, isn't it?

File: cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp (hunk 3 above)
Comment (tigercosmos) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: > The ellipsis syntax should populate everything including the ghost elements.

then why do we need to test `nghost != 0`? `memcpy` will have also copy ghost elements.

current PR behavior
```
sarr = modmesh.SimpleArrayFloat64(12)
ndarr = np.arange(12, dtype='int32')
sarr.nghost = 2

sarr[...] = ndarr[...]

print(sarr[-2]) # 0.0
print(sarr[-1]) # 1.0
print(sarr.ndarray) # [ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10. 11.]
print(ndarr) # [ 0  1  2  3  4  5  6  7  8  9 10 11]
```

Comment (yungyuc) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: That is the behavior I am expecting and an expected behavior should be added in unit tests.

File: cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp (hunk 4 above)
Comment (yungyuc) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: What is the reason to use `std::memcpy` but not `std::copy_n`?

Comment (tigercosmos) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: The main reason is that `pybind11:array` doesn't provide iterators, resulting in not supporting `copy_n`.
In this case, since the buffer type and size will be the same, I think `memcpy` should work well.
Even if `std::copy_n` is available, I wonder about the performance difference between calling `memcpy` and `copy_n`?

Comment (yungyuc) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: `copy_n` can be used for the buffer just like `memset`.  When using `memset`, I guess you are relying more on how OS chooses to implement it because compiler needs a good reason not to use the OS implementation.  With `copy_n` it's more on the stdlib.

`memset` and `copy_n` have different definition and perhaps the former has a better chance to be optimized away: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=8537 .

I prefer `copy_n`, because it induces less change of code when we move from pointer to an alternate iterator.  Using `memset` at this point requires one to recode it to `copy_n` later.

Comment (yungyuc) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: Please note that it is dangerous to directly use the memory buffer of the input `ndarray`, because the input may use non-simple strides to skip elements, e.g., `SimpleArrayFloat64[...] = ndarray[::2,:,:]`.  It would be interesting to add a unit test for the strided input.

If the test fails, I am still fine to use the memory buffer for this feature, as a draft implementation.  But the failed test should be added as a known failure and a new issue should be created to track the fix.

**Comment by tigercosmos** [comment]
Now only `test_SimpleArray_broadcast_ellipsis_ghost` need to resolve

File: cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp (hunk 4 above)
Comment (tigercosmos) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: resolve type broadcast

File: cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp (hunk 3 above)
Comment (tigercosmos) [reply cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp]: resolved

File: cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp [hunk 5]
```diff
@@ -161,15 +163,59 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
     }
 
-    static void copy_array_using_ellipsis(wrapped_type & self, pybind11::array & arr_in)
+    template <typename D /* for destination type */>
+    struct TypeBroadCast
     {
-        const auto left_shape = self.shape();
+        static void broadcast(wrapped_type & arr_out, pybind11::array & arr_in)
+        {
+            using shape_type = small_vector<size_t>;
```
yungyuc: You do not need this alias for it was done in line 45:

```cpp
using shape_type = typename wrapped_type::shape_type;
```

And even if we create an alias, the existing `wrapped_type` should be used.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -149,7 +149,9 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def(
                 "__setitem__",
                 [](wrapped_type & self, const py::ellipsis &, pybind11::array & arr_in)
```
yungyuc: (Missed in previous review.)  Other places in the file use `T const &` but here you write `const T &`.  It's OK, but being consistent with the rest in the file is better.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,162 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+    template <typename D /* for destination type */>
+    struct TypeBroadCast
+    {
+        static void broadcast(wrapped_type & arr_out, pybind11::array & arr_in)
+        {
+            using shape_type = small_vector<size_t>;
+            using out_type = typename std::remove_reference<decltype(arr_out[0])>::type;
+
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+            auto * arr_new = reinterpret_cast<pybind11::array_t<D> *>(&arr_in);
```
yungyuc: This looks hacky.  I cannot think of a better way to do it, though.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,162 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    sidx[dim] = i;
+
+                    size_t offset_in = 0;
+                    for (pybind11::ssize_t it = 0; it < arr_in.ndim(); ++it)
+                    {
+                        offset_in += arr_in.strides(it) / arr_in.itemsize() * sidx[it];
+                    }
+
+                    const D * ptr_in = arr_new->data();
+                    // NOLINTNEXTLINE
```
yungyuc: Please make it explicit what tidy checks you skip.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,162 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    copy_idx(sidx, dim - 1);
+                }
+            };
+
+            copy_idx(sidx_init, arr_out.ndim() - 1);
+        }
+    };
+
+    // NOLINTNEXTLINE(readability-function-cognitive-complexity)
+    static void broadcast_array_using_ellipsis(
```
yungyuc: The signature is short.  Please put it in one line.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,162 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    arr_out.at(sidx) = static_cast<out_type>(*(ptr_in + offset_in));
+                    copy_idx(sidx, dim - 1);
+                }
+            };
+
+            copy_idx(sidx_init, arr_out.ndim() - 1);
+        }
+    };
+
+    // NOLINTNEXTLINE(readability-function-cognitive-complexity)
```
yungyuc: I agree with clang-tidy.  This function is longer than 100 lines but it is not a complex tight loop.  Please follow tidy and break it.

Perhaps you can create a helper class like `TypeBroadCast`?

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,162 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+    static void broadcast_array_using_ellipsis(
+        wrapped_type & arr_out, pybind11::array & arr_in)
+    {
+        const auto left_shape = arr_out.shape();
+        const auto * right_shape = arr_in.shape();
+
+        auto throw_shape_error = [&]()
```
yungyuc: It seems that the lambda can be moved to `TypeBroadCast` as a static member function.

A lambda here does not make code easy to read.  The reference capture does not make it easy to maintain.  The only benefit of using this lambda looks like a quick hack.  Let me know if I miss something.

tests/test_buffer.py
```diff
@@ -363,4 +363,104 @@ def test_SimpleArray_from_ndarray_content(self):
         sarr.ndarray.fill(100)
         self.assertTrue((ndarr == 100).all())
 
+    def test_SimpleArray_broadcast_ellipsis_shape(self):
```
yungyuc, line 366: I like the additional tests.

yungyuc (changes requested):
It's great to see the ellipsis support is taking shape.  Thanks a lot and there are some things to be fixed.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: I think it's a smart way :P

yungyuc: I can't say it's not :)

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,177 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                }
+            }
+            msg << ")";
+
+            throw std::runtime_error(msg.str());
+        };
+    };
+
+    template <typename D /* for destination type */>
+    struct TypeBroadCastImpl
```
tigercosmos, line 270: In the beginning, I want to just implement a template function, `broadcast_impl<D>`, in the origin `TypeBroadCast` struct (not a template class now), but I found I couldn't do that (error: "use of function template name with no prior declaration in function call with explicit template"). So I created another struct to handle the template. Actually, I still don't know why I cannot just put a template function in a non-template class?

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,177 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             ;
     }
 
+    struct TypeBroadCast
+    {
+        static void check_shape(wrapped_type & arr_out, pybind11::array & arr_in)
```
yungyuc: Argument should use const reference because you do not write to them.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
yungyuc: If you have code like that, post to gist and I can help analyze.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,177 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            msg << ")";
+
+            throw std::runtime_error(msg.str());
+        };
+    };
+
+    template <typename D /* for destination type */>
+    struct TypeBroadCastImpl
+    {
+        static void broadcast(wrapped_type & arr_out, pybind11::array & arr_in)
```
yungyuc: `arr_in` should be a const reference

yungyuc (changes requested):

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -157,6 +163,177 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    size_t offset_in = 0;
+                    for (pybind11::ssize_t it = 0; it < arr_in.ndim(); ++it)
+                    {
+                        offset_in += arr_in.strides(it) / arr_in.itemsize() * sidx[it];
+                    }
+
+                    const D * ptr_in = arr_new->data() + offset_in;
+                    // NOLINTNEXTLINE(bugprone-signed-char-misuse, cert-str34-c)
+                    arr_out.at(sidx) = static_cast<out_type>(*ptr_in);
+                    copy_idx(sidx, dim - 1);
```
yungyuc, line 308: It's good to use a comment to signify the recursion.

tigercosmos: done

yungyuc (approved):

yungyuc:
Thanks a lot @tigercosmos

PR #107, opened 2022-07-21
SimpleArray: setitem parser & broadcast array by slices

tigercosmos:
#52

### DIFF
--- a/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
@@ -158,9 +158,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         if (args.size() == 2)
         {
             // sarr[K] = V
-            if (py::isinstance<py::int_>(args[0]) &&
-                (py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) ||
-                 py::isinstance<py::float_>(args[1])))
+            if (py::isinstance<py::int_>(args[0]) && !py::isinstance<py::array>(args[1]))
             {
                 const auto key = args[0].cast<ssize_t>();
 
@@ -168,9 +166,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 return;
             }
             // sarr[K1, K2, K3] = V
-            if (py::isinstance<py::tuple>(args[0]) &&
-                (py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) ||
-                 py::isinstance<py::float_>(args[1])))
+            if (py::isinstance<py::tuple>(args[0]) && !py::isinstance<py::array>(args[1]))
             {
                 const auto key = args[0].cast<std::vector<ssize_t>>();
 
--- a/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
@@ -138,21 +138,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "__getitem__",
                 [](wrapped_type const & self, std::vector<ssize_t> const & key)
                 { return self.at(key); })
-            .def(
-                "__setitem__",
-                [](wrapped_type & self, ssize_t key, T val)
-                { self.at(key) = val; })
-            .def(
-                "__setitem__",
-                [](wrapped_type & self, std::vector<ssize_t> const & key, T val)
-                { self.at(key) = val; })
-            .def(
-                "__setitem__",
-                [](wrapped_type & self, py::ellipsis const &, pybind11::array & arr_in)
-                {
-                    broadcast_array_using_ellipsis(self, arr_in);
-                })
-            .def("__setitem__", &copy_array)
+            .def("__setitem__", &setitem_parser)
             .def(
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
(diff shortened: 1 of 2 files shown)


tigercosmos:
@yungyuc  please review

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             ;
     }
 
+    static void setitem_parser(wrapped_type & arr_out, pybind11::args & args)
```
yungyuc: Make `args` const &

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+    {
+        namespace py = pybind11;
+
+        if (args.size() == 2)
+        {
+            // sarr[K] = V
+            if (py::isinstance<py::int_>(args[0]) &&
+                (py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) ||
```
yungyuc: It seems this is just `!isinstance<py::array>` and does not need to use a list of types.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                 py::isinstance<py::float_>(args[1])))
+            {
+                const auto key = args[0].cast<std::vector<ssize_t>>();
+
+                arr_out.at(key) = args[1].cast<T>();
+                return;
+            }
+            // multi-dimension with slice and ellipsis
+            // sarr[slice, slice, ellipsis] = ndarr
```
yungyuc, line 177: The comments are useful.  Thanks.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            if (py::isinstance<py::tuple>(args[0]) && py::isinstance<py::array>(args[1]))
+            {
+                const auto tuple = args[0].cast<py::tuple>();
```
yungyuc: Maybe `const py::tuple args_in = args[0];` would just work.  It looks a little bit strange to cast a Python type to another Python type.

yungyuc: And naming a variable using its type name is hard to read.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        return slices;
+    }
+
+    static void process_slice(pybind11::tuple const & tuple,
+                              std::vector<small_vector<int>> & slices,
+                              size_t ndim)
+    {
+        namespace py = pybind11;
+
+        // copy slices from the front untill an ellipsis
```
yungyuc: *until

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        }
+    }
+
+    static void broadcast_array_using_slice(wrapped_type & arr_out,
+                                            std::vector<small_vector<int>> const & slices,
+                                            pybind11::array const & arr_in)
+    {
+        TypeBroadCast::check_shape(arr_out, slices, arr_in);
+
+        size_t nghost = 0;
```
yungyuc: It'll be more readable:

```cpp
size_t const nghost = arr_out.nghost();
if (0 != nghost) { arr_out.set_nghost(0); }
TypeBroadCast ...;
if (0 != nghost) { arr_out.set_nghost(nghost); }
```

yungyuc: There is a theory that constants or literals should be put as the first argument for comparison operation so that compiler will surely make a noise with a typo happens: `0 = nghost`.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-            const auto & left_shape = arr_out.shape();
-            const auto * right_shape = arr_in.shape();
+            shape_type right_shape(arr_in.ndim());
+            for (pybind11::ssize_t i = 0; i < arr_in.ndim(); i++)
+            {
+                right_shape[i] = arr_in.shape(i);
+            }
+
+            shape_type left_shape(arr_out.ndim());
+            // todo: range check
```
yungyuc: Thanks for the reminder.  I used to use all-capital for TODO in this code base.

tests/test_buffer.py
```diff
@@ -462,5 +462,133 @@ def test_SimpleArray_broadcast_ellipsis_dtype(self):
...
+            1*2*3*4*5, dtype='float64').reshape((1, 2, 3, 4, 5))
+        ndarr_input += 999  # just a magic number
+
+        def init(sarr):
+            v = 0
+            for x1 in range(sarr.shape[0]):
+                for x2 in range(sarr.shape[1]):
+                    for x3 in range(sarr.shape[2]):
+                        for x4 in range(sarr.shape[3]):
+                            for x5 in range(sarr.shape[4]):
```
yungyuc, line 476: This is a good place for some humourous comments :)

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            for (size_t i = 0; i < arr_out.ndim(); i++)
+            {
+                const auto & slice = slices[i];
```
yungyuc: Please spelling out the type for `slice`.  `auto` makes it harder to read here.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
     {
-        static void check_shape(wrapped_type const & arr_out, pybind11::array const & arr_in)
+        static void check_shape(wrapped_type const & arr_out,
+                                std::vector<small_vector<int>> const & slices,
+                                pybind11::array const & arr_in)
         {
-            const auto & left_shape = arr_out.shape();
-            const auto * right_shape = arr_in.shape();
+            shape_type right_shape(arr_in.ndim());
+            for (pybind11::ssize_t i = 0; i < arr_in.ndim(); i++)
```
yungyuc, line 343: The slice handling code here is a good staring point for adding view support for `SimpleArray` in the future.

tests/test_buffer.py
```diff
@@ -462,5 +462,133 @@ def test_SimpleArray_broadcast_ellipsis_dtype(self):
...
+
+        sarr[::STEP] = ndarr[...]
+        ndarr2[::STEP] = ndarr[...]
+
+        print(ndarr2.shape)
+
+        for i in range(0, N, STEP):
+            self.assertEqual(ndarr2[i], sarr[i - G])
+
+    def test_SimpleArray_broadcast_slice_ghost_md(self):
```
yungyuc, line 574: Love it.

yungyuc (changes requested):

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: I don't think so, since you may also do something like `sarr[K] = (1, 2, 3)`

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
```diff
@@ -163,99 +150,308 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                (py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) ||
+                 py::isinstance<py::float_>(args[1])))
+            {
+                const auto key = args[0].cast<ssize_t>();
+
+                arr_out.at(key) = args[1].cast<T>();
+                return;
+            }
+            // sarr[K1, K2, K3] = V
+            if (py::isinstance<py::tuple>(args[0]) &&
```
tigercosmos: actually, I think the `py::tuple` here should be better to check if the items' types are number.

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: I will need to discuss this with you later.

tests/test_buffer.py
tigercosmos: could you provide some examples(?

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
yungyuc: I presume an exception is thrown with `sarr[K] = (1, 2, 3)` when the check is only `!isinstance<py::array>`.  Am I right?

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
yungyuc: We can use dc or https://github.com/solvcon/modmesh/discussions to discuss.

tests/test_buffer.py
yungyuc: That won't be humourous anymore.  Let's try it next time :)

cpp/modmesh/python/wrapper/modmesh/wrap_SimpleArray.cpp
tigercosmos: you are right!
    sarr[0, 0, 0, 0, 0] = (1, 2, 4)
RuntimeError: Unable to cast Python instance of type <class 'tuple'> to C++ type 'double'

yungyuc: That exception seems to come from pybind11.  Using that exception is more maintainable than explicit Python type inspection.

yungyuc (approved):

yungyuc:
Thanks a lot.  Merging.

PR #108, opened 2022-07-22
Refactor spacetime and view

yungyuc:
Merge the code of spacetime to only a couple of files, and from the `spacetime` namespace to `modmesh::spacetime` namespace.

Move the Python source files of view to be with the rest view source files, because the whole directory of `view/` depends on Python.


yungyuc:
Again, the diff is large and I think it does not make sense for anyone else to review it.

PR #109, opened 2022-07-23
Create a new option to animate the linear solver plot

yungyuc:
In the `linear_solver` app, animate the plot.  Also, make the animation the default in the app code.


yungyuc:
@EN-Chou You can take a look at this simple PR.

Issue #70, opened 2022-05-07
Improve camera control


yungyuc:
The current camera control uses vanilla `Qt3DExtras::QOrbitCameraController` and should be improved.  The camera control is a critical component in the viewer.  It at least needs the following functions and more:

* Resetting.
* WSAD.
* Change of origin.

----

Amended on 27th July 2024

- [x] "Up" and "w" should move toward the top of the window.  "Down" and "s" should move toward the bottom of the window.  "Q" and "e" should change to move along the z axis (perpendicular to the window).
  - I am not sure about the binding to page up/down.  We can use discussions.
  - Implemented with PR #408.
- [x] Camera reset/original position to be configurable and exposed through Python API.  More information: https://github.com/solvcon/modmesh/pull/408#discussion_r1718535001
  - Done with PR #418

yanagiragi:
Please let me have a try.

yungyuc:
@tychuang1211 and @YenPeiChen07 are working on it

yungyuc:
@Tucchhaa Thanks for the PR #385.  After testing I found w and q should swap and s and e should swap.  For a simulation / cad code, it is unintuitive for "up" and "down" to move along the z axis (perpendicular to the screen/window).  "Up" should go toward the top of window and "down" should go toward the bottom of the window.

We should also add the movements as items in the "View" menu.  Could you please help it?

Tucchhaa:
> We should also add the movements as items in the "View" menu. Could you please help it?

@yungyuc do you mean assigning custom keys to the camera's movement?

yungyuc:
> > We should also add the movements as items in the "View" menu. Could you please help it?
> 
> @yungyuc do you mean assigning custom keys to the camera's movement?

No, but that is also an important feature.  It is substantial work and will need a separate issue to track.  The ultimate form may be like how vscode and clion (JetBrains IntelliJ-series IDE) work.

----

The "View" menu is like below, and it should contain the movements with hotkey labeled:

<img width="425" alt="image" src="https://github.com/user-attachments/assets/69517b15-81e4-4039-b201-14dbed1ec261">

Tucchhaa:
@yungyuc should the movement hotkey items be clickable or disabled?

yungyuc:
> @yungyuc should the movement hotkey items be clickable or disabled?

I was thinking they should be clickable (enabled).  I didn't think of a use case for the movement menu items to be disabled.  But perhaps I missed something?

[...]

Issue #84, opened 2022-06-03
Add global feature toggles


yungyuc:
We are adding more and more features to the project and there is not yet a system to manage the settings of the features.  A [feature-toggling system](https://martinfowler.com/articles/feature-toggles.html) should be implemented to house and control the features that were and will be added.  Related discussions are in #165 .

The feature toggles should support two types of parameters:

1. Set once and not change during the whole execution.  This is for flags used in the tight loops.
2. Being changed multiple or many times during the execution.  This is for interactive features.

Tasks:

- [x] Develop the `FixedToggle` to house compile-time-named, runtime-initialized values.  Addresses of the values are determined by the compiler/linker and it does not take table lookup.  It does not use hierarchy.
- [x] Develop the `DynamicToggleTable` to house the key-value pairs determined only during runtime.  Access requires table lookup.  The keys are dynamic strings.  It support hierarchical access.
- [x] Develop serialization by using json/yaml as the file format.

After this issue is done, update modmesh engine and apps to use the toggles (create new issues to track the follow-up work).

yungyuc:
I am starting to add some simple code for it.

yungyuc:
See also discussions in #165

PR #112, opened 2022-07-30
feat: Use FPSCamera to improve control

yanagiragi:
https://github.com/solvcon/modmesh/issues/70

I replace to original orbital camera to first person camera to improve camera controls, In my opinion the controls are more straight forward. 

Currently the new camera contain some cons:

1. It does not support to reverse the vertical direction to rotation the camera
2. The controls should be more straight forward if it supports WASDQE instead of Arrow/PageUpDown

However, auto focus to the whole scene is the killer feature. I highly recommend to give it a try.

P.S. The supported controls is documented in [here](https://doc.qt.io/qt-6/qt3dextras-qfirstpersoncameracontroller.html#details).


yungyuc (changes requested):
Both `QOrbitCameraControl` and `QFirstPersonCameraControl` look useful.  Can we keep both and switch using a menu entry?

yanagiragi:
Sure. Which one should be the default camera?

yungyuc:
> Sure. Which one should be the default camera?

Either is OK.  It will be configurable once we figure out a good way to implement #84 .

yanagiragi:
@yungyuc I've upload the changes.

cpp/modmesh/view/RApplication.cpp
```diff
@@ -96,6 +101,39 @@ RApplication::RApplication(int & argc, char ** argv)
...
+        QString("Use First Person Camera Controller"),
+        [&]()
+        {
+            qDebug() << "Use First Person Camera Controller";
+            auto * viewer = this->m_main->viewer();
+            viewer->scene()->setFirstPersonCameraController();
+            viewer->scene()->controller()->setCamera(viewer->camera());
+        });
+
+    auto * cameraGroup = new QActionGroup(this);
```
yungyuc, line 126: We have been mixing camelCase and snake_name for local variables.  You didn't start it.  But we should start to think about how to end it.  It will probably take more time for us to accumulate code before coming up with a good guideline.

cpp/modmesh/view/RApplication.cpp
```diff
@@ -96,6 +101,39 @@ RApplication::RApplication(int & argc, char ** argv)
...
+    cameraGroup->addAction(use_orbit_camera);
+    cameraGroup->addAction(use_fps_camera);
+    use_orbit_camera->setCheckable(true);
+    use_fps_camera->setCheckable(true);
+    use_orbit_camera->setChecked(true);
+
+    cameraMenu->addAction(use_orbit_camera);
+    cameraMenu->addAction(use_fps_camera);
+    menuBar->addMenu(cameraMenu);
```
yungyuc, line 135: Cool.  I will try it later.

yungyuc (approved):

PR #116, opened 2022-08-18
Add modmesh macos viewer CI

minchuncho:
Add modmesh viewer CI for Mac OS: `macos-latest`.

### DIFF
--- a/.github/workflows/modmesh_viewer.yml
+++ b/.github/workflows/modmesh_viewer.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-22.04, macos-latest]
+          os: [ubuntu-22.04, macos-12]
           cmake_build_type: [Release, Debug]
 
         fail-fast: false
@@ -42,11 +42,19 @@ jobs:
         sudo pip3 install numpy pytest flake8
         
     - name: dependency (macos)
-      if: ${{ startsWith(matrix.os, 'macos-latest') }}
+      if: ${{ startsWith(matrix.os, 'macos-12') }}
       run: |
-        brew update
-        brew install wget make cmake libccd gcc python3 qt6
-        pip3 install numpy pytest flake8
+        # brew install python3 llvm
+        # brew install qt6
+        # brew install wget make cmake libccd gcc python3 qt6
+        # pip3 install -U numpy pytest flake8
+
+        brew install python3 llvm qt6
+        ln -s "$(brew --prefix llvm)/bin/clang-format" "/usr/local/bin/clang-format"
+        ln -s "$(brew --prefix llvm)/bin/clang-tidy" "/usr/local/bin/clang-tidy"
+        echo "which pip3: $(which pip3)"
+        pip3 install -U setuptools
+        pip3 install -U numpy pytest flake8
 
     - name: dependency (manual)
       run: sudo ${GITHUB_WORKSPACE}/contrib/dependency/install.sh pybind11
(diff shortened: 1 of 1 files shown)


.github/workflows/modmesh_viewer.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-22.04]
+          os: [ubuntu-22.04, macos-latest]
```
yungyuc: Please use the explicit version `macos-12` rather than the alias `macos-latest`, so that it is consistent to other workflow, e.g., https://github.com/solvcon/modmesh/blob/9eef4aaf004cbf86478368714e13493d34e0d59a/.github/workflows/modmesh_pytest.yml#L20 

modmesh is supposed to be developed using a newer system.  That is why Qt 6 is chosen rather than Qt 5.  So it is preferred to use macos 12 instead of 11.

The alias `macos-latest` changes as GitHub Action upgrades its system.  At this point it is set to `macos-11` (see [the github doc](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)).  Even if we would like to test for macos 11, we should use the explicit version.

.github/workflows/modmesh_viewer.yml
```diff
@@ -40,6 +40,13 @@ jobs:
             libgl1-mesa-dev
 
         sudo pip3 install numpy pytest flake8
+        
+    - name: dependency (macos)
+      if: ${{ startsWith(matrix.os, 'macos-latest') }}
+      run: |
+        brew update
```
yungyuc: Could you please align the commands used in https://github.com/solvcon/modmesh/blob/9eef4aaf004cbf86478368714e13493d34e0d59a/.github/workflows/modmesh_pytest.yml#L48 

If you have a better way to install using brew, feel free to update the other places.

yungyuc (changes requested):
Thanks a lot for the addition, @minchuncho .  I think we can use a couple of changes in the PR.

.github/workflows/modmesh_viewer.yml
minchuncho: Sounds reasonable. I will switch it.

.github/workflows/modmesh_viewer.yml
minchuncho: Okay, let me check this.

.github/workflows/modmesh_viewer.yml
```diff
@@ -40,6 +40,21 @@ jobs:
             libgl1-mesa-dev
 
         sudo pip3 install numpy pytest flake8
+        
+    - name: dependency (macos)
+      if: ${{ startsWith(matrix.os, 'macos-12') }}
+      run: |
+        # brew install python3 llvm
```
yungyuc, line 47: Why do we use the comments here?

yungyuc (changes requested):
Could you please also use a different branch for the PR?  You are requesting to merge from your `master` to this `master`.  A merge commit like that will look confusing.

I suggest to include one or two keywords of the purpose of the branch in the name.

But I am not totally sure whether or not you can change the branch name of a PR.  If that is not possible, it is OK to close this one and file a new one.

.github/workflows/modmesh_viewer.yml
minchuncho: Do you mean merging from, for example, `add-viewer-macos-ci` to `master`?

Btw about the comments, I forgot to remove them before pushing. I will fix it afterwards.

yungyuc: Yes that's what I meant, and I saw your new PR #117

Issue #111, opened 2022-07-24
Add/evaluate macOS (12) build in CI


yungyuc:
Currently the `modmesh-viewer` has only Ubuntu build test.  It will be convenient to have macOS (12) viewer build as well.

minchuncho:
I'm on this.

yungyuc:
Thanks @minchuncho

PR #117, opened 2022-08-19
Add macos build test to modmesh_viewer.yml

minchuncho:
1. Specify the version of mac os: 12
2. Align the commands with the ones in https://github.com/solvcon/modmesh/blob/9eef4aaf004cbf86478368714e13493d34e0d59a/.github/workflows/modmesh_pytest.yml#L48

### DIFF
--- a/.github/workflows/modmesh_viewer.yml
+++ b/.github/workflows/modmesh_viewer.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-22.04]
+          os: [ubuntu-22.04, macos-latest]
           cmake_build_type: [Release, Debug]
 
         fail-fast: false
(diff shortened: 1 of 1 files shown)


minchuncho:
My old version of CI is as follows:
```
brew install wget make cmake libccd gcc python3 qt6
pip3 install -U numpy pytest flake8
```
Curious, I didn't write the following line but the old version still worked.
https://github.com/solvcon/modmesh/blob/9eef4aaf004cbf86478368714e13493d34e0d59a/.github/workflows/modmesh_pytest.yml#L52

I found that `setuptools` is a Python package development process library, i.e., enhanced `disutils`, while `disutils` is a Python standard library.
May I reach to this conclusion that if `setuptools` is not installed, then `disutils` is chosen by default so the old version still ran well?

.github/workflows/modmesh_viewer.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-22.04]
+          os: [ubuntu-22.04, macos-latest]
```
yungyuc: You have changed it to `macos-12` in #116 but here it goes back to `macos-latest`.  Perhaps an overlook?

yungyuc (changes requested):
LGTM except the regression to os version alias.

yungyuc:
> Curious, I didn't write the following line but the old version still worked.
> https://github.com/solvcon/modmesh/blob/9eef4aaf004cbf86478368714e13493d34e0d59a/.github/workflows/modmesh_pytest.yml#L52
> 
> I found that `setuptools` is a Python package development process library, i.e., enhanced `disutils`, while `disutils` is a Python standard library. May I reach to this conclusion that if `setuptools` is not installed, then `disutils` is chosen by default so the old version still ran well?

That is a reasonable guess, I think.  I do not know the answer either.  It probably does not matter ATM, but could be interesting to learn.

.github/workflows/modmesh_viewer.yml
minchuncho: Yes, it is an overlook and I have already fixed it.

yungyuc: LGTM.  I will merge once the CI passes.

yungyuc (approved):

yungyuc:
Thanks @minchuncho .  This addressed #111

Issue #118, opened 2022-08-19
Crash with the RHI rendering backend


minchuncho:
On a Mac of an older version, supposedly below `12`, the viewer app is likely to crash with the RHI rendering backend.

In my case, the following error messages first appeared while running `sample_mesh` and the viewer app failed to render the edges and axes as expected.
```
2022-08-05 15:04:47.683571+0800 viewer[9581:98839] Failed to create render pipeline state: Vertex attribute vertexNormal(1) is missing from the vertex descriptor
2022-08-05 15:04:47.683640+0800 viewer[9581:98839] [Qt3D.Renderer.RHI.Backend] Failed to build graphics pipeline: Creation Failed
2022-08-05 15:04:47.683704+0800 viewer[9581:98839] Failed to create render pipeline state: Vertex attribute vertexNormal(1) is missing from the vertex descriptor
2022-08-05 15:04:47.683727+0800 viewer[9581:98839] [Qt3D.Renderer.RHI.Backend] Failed to build graphics pipeline: Creation Failed
2022-08-05 15:04:47.683986+0800 viewer[9581:98839] Failed to create render pipeline state: Vertex attribute vertexNormal(1) is missing from the vertex descriptor
2022-08-05 15:04:47.684011+0800 viewer[9581:98839] [Qt3D.Renderer.RHI.Backend] Failed to build graphics pipeline: Creation Failed
2022-08-05 15:04:47.684224+0800 viewer[9581:98839] Failed to create render pipeline state: Vertex attribute vertexNormal(1) is missing from the vertex descriptor
2022-08-05 15:04:47.684256+0800 viewer[9581:98839] [Qt3D.Renderer.RHI.Backend] Failed to build graphics pipeline: Creation Failed
```
Even, a while after the running, the viewer app straight crashed with a segmentation fault.

As specified in the document (https://doc.qt.io/qtforpython/overviews/qt3drender-porting-to-rhi.html), the workaround is to switch the Qt6 rendering backend from RHI to OpenGL by setting the environment variable `QT3D_RENDERER=opengl` before running the viewer app.
Moreover, the doc also lists several limitations of RHI rendering backend which may be the culprit of the crash.

_Sidenote: The issue originated with a macOS of version `11.6.8` and Qt of version `6.3.1`._

yungyuc:
I wonder is there a way to set the renderer other than using the environment variable?  If there is, we may try to do it in the code with a config file.

minchuncho:
[...]

PR #119, opened 2022-08-24
Add environment variable from the setup to solve issue 118

minchuncho:
@yungyuc

`QT3D_RENENDER := opengl` is first added in `setup.mk`. 
Then, by this line (https://github.com/minchuncho/modmesh/blob/54fc5160f1da289c25f41eca7d351ae77ac81136/Makefile#L140) `QT3D_RENENDER` is treated as one of the CMake argument and hopefully by this line (https://github.com/minchuncho/modmesh/blob/54fc5160f1da289c25f41eca7d351ae77ac81136/CMakeLists.txt#L100), `QT3D_RENENDER=opengl` can be set as an environment variable.

Unfortunately, it failed to do as expected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,8 @@ endif()
 add_subdirectory(cpp/binary/pymod_modmesh)
 if(BUILD_QT)
 add_subdirectory(cpp/binary/viewer)
+set(ENV{QT3D_RENDERER} ${QT3D_RENDERER})
+message(STATUS "QT3D_RENDERER: " $ENV{QT3D_RENDERER})
 endif()
 
 add_custom_target(flake8)
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ CMAKE_PREFIX_PATH ?=
 CMAKE_ARGS ?=
 VERBOSE ?=
 FORCE_CLANG_FORMAT ?=
+QT3D_RENDERER ?=
 
 pyextsuffix := $(shell python3-config --extension-suffix)
 pyvminor := $(shell python3 -c 'import sys; print("%d%d" % sys.version_info[0:2])')
@@ -136,6 +137,7 @@ CMAKE_CMD = cmake $(MODMESH_ROOT) \
 	-DUSE_CLANG_TIDY=$(USE_CLANG_TIDY) \
 	-DLINT_AS_ERRORS=ON \
 	-DMODMESH_PROFILE=$(MODMESH_PROFILE) \
+	-DQT3D_RENDERER=$(QT3D_RENDERER) \
 	$(CMAKE_ARGS)
 
 $(BUILD_PATH)/Makefile: CMakeLists.txt Makefile


CMakeLists.txt
```diff
@@ -97,6 +97,8 @@ endif()
 add_subdirectory(cpp/binary/pymod_modmesh)
 if(BUILD_QT)
 add_subdirectory(cpp/binary/viewer)
+set(ENV{QT3D_RENDERER} ${QT3D_RENDERER})
```
yungyuc, line 100: IIUC, you would like to set the name `QT3D_RENDERER` to have the value set in `ENV{QT3D_RENDERER}`.

You seem to use a wrong syntax of `set`.  Please see https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable .  Quote here:

```cmake
set(<variable> <value>... [PARENT_SCOPE])
```

The variable to be set should be the first argument to `set`, while the value is the second.  So one way to fix it is to rewrite this line to

```cmake
set(QT3D_RENDERER $ENV{QT3D_RENDERER})
```

Please note that you need the `$` for reading from the environment variable: https://cmake.org/cmake/help/latest/variable/ENV.html .

However this might not be ideal.  Perhaps it should be considered to use `option`, which can be passed in with the `-D` argument to the cmake command line.

yungyuc (changes requested):
Could you provide some reference to how `QT3D_RENDERER` works?

CMakeLists.txt
minchuncho: @yungyuc 
Hmm not exactly, let me rephrase the problem.

After the building process is done, we need to export the environment variable `QT3D_RENDERER=opengl` in the shell before running the viewer, which allows the viewer to successfully run with the opengl backend instead of the rhi backend.
That being said, I wasn't trying to load the environment variable from the shell and create a cmake cache entry called `QT3D_RENDERER`; instead, my intension was to export `QT3D_RENDERER=opengl` in the shell thru cmake.

As you initially suggested in https://github.com/solvcon/modmesh/issues/118, I tried to set the environment variable thru `setup.mk` and passed it in with the `-D` argument to the cmake command line, as you described above.
Then, I thought the following line
https://github.com/solvcon/modmesh/blob/54fc5160f1da289c25f41eca7d351ae77ac81136/CMakeLists.txt#L100
would do until I came across [the cmake doc](https://cmake.org/cmake/help/latest/command/set.html#set-environment-variable) which clearly states:
> This command affects only the current CMake process, not the process from which CMake was called, nor the system environment at large, nor the environment of subsequent build or test processes.

That is why `QT3D_RENDERER=opengl` wasn't exported as expected and is exactly the problem I wanted to address here.

minchuncho: FIY, [this doc](https://doc.qt.io/qtforpython/overviews/qt3drender-porting-to-rhi.html) specifies how `QT3D_RENDERER` works.

yungyuc: I see.  Since it is an environment variable effective during runtime, it probably does not make sense to set it during compile time.

To fix #118, we probably need to figure out how to configure to use opengl using Qt6 API.

yungyuc: While I have guessed it could be like it, I am still surprised a little bit that it is really it.  Please take a look how [Qt itself handles it](https://github.com/qt/qt3d/blob/fcc2f7d398a31c96434a350f358e3349b582c6f5/src/extras/defaults/qt3dwindow.cpp#L230
).  We should do something similar:

```cpp
    // Default to using RHI backend is not specified We want to set the
    // variable to ensure any 3rd party relying on it to detect which rendering
    // backend is in use will get a valid value.
    bool useRhi = false;
    if (qEnvironmentVariableIsEmpty("QT3D_RENDERER")) {
#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0) && QT_CONFIG(qt3d_rhi_renderer)
        qputenv("QT3D_RENDERER", "rhi");
#else
        qputenv("QT3D_RENDERER", "opengl");
#endif
    }
#if QT_CONFIG(qt3d_rhi_renderer)
    useRhi = qEnvironmentVariable("QT3D_RENDERER") == QStringLiteral("rhi");
#endif
```

But I think modmesh should not use the Qt API `qputenv` for manipulating environment variable.  If C++ doesn't provide a standard way doing it, a simple system should be implemented in modmesh.

yungyuc:
The PR has been updated with #195 .  Thank you @minchuncho for the initial work!

PR #121, opened 2022-09-11
setup docker environment

dboyliao:
- Setup docker environment for development
- Add README for instructions of building docker environment

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,0 +1,32 @@
+FROM ubuntu:22.04
+
+WORKDIR /workspace
+ENV PYTHONDONTWRITEBYTECODE=1
+
+RUN apt-get update && \
+  apt-get install -y curl \
+  git \
+  build-essential \
+  cmake \
+  libc6-dev \
+  g++ \
+  clang-tidy-14 \
+  qt6-base-dev
+RUN apt-get install -y \
+  python3 \
+  python3-dev \
+  python3-venv \
+  python3-setuptools \
+  python3-pip \
+  python3-numpy \
+  python3.10-dev
+RUN python3 -m pip install \
+  jupyter \
+  pytest \
+  flake8
+RUN apt-get purge && \
+  apt-get autoremove && \
+  apt-get autoclean && \
+  rm -rf /var/lib/apt/lists/*
+
+CMD [ "/bin/bash" ]
--- a/contrib/docker/README.md
+++ b/contrib/docker/README.md
@@ -1,0 +1,11 @@
+# Docker
+
+## Build
+```bash
+$ docker build -t solvcon/modmesh:latest contrib/docker
+```
+
+## Run
+```bash
+$ docker run -it --rm -v $(pwd):/workspace solvcon/modmesh:latest
+```


dboyliao:
FYI, I'm able to build `modmesh` and import it in the docker container.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "thirdparty/pybind11"]
```
yungyuc: Dependency should be installed outside the project, except some very special code that will be considered an integrated part in the source form, e.g., [metal-cpp](https://github.com/solvcon/modmesh/blob/master/thirdparty/metal-cpp.sh).

One way to build the dependency is to use https://github.com/solvcon/devenv .

Makefile
```diff
@@ -112,8 +112,15 @@ cmake: $(BUILD_PATH)/Makefile
 .PHONY: xcode
 xcode: $(BUILD_PATH)_xcode/Makefile
 
+.PHONY: pybind11
```
yungyuc: pybind11 is a header only library and should not use a target.  Please revert the change in the make file.

README.md
```diff
@@ -0,0 +1,11 @@
+# Docker
```
yungyuc: We would like to avoid use of container because it is hard to maintain.  Container is good for deployment but it is cumbersome for source-level debugging.  For example, it's hard to compare the debug and optimized builds using a container.  It also adds complexity and overheads in instrumenting and profiling.

It's probably OK to add optional scripts for container, but we should not bless the use.

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:22.04
```
yungyuc: If adding docker script, it should be put under `contrib/`, not the root, because container is not a blessed way to build it.

yungyuc (changes requested):

Makefile
dboyliao: reverted.

.gitmodules
dboyliao: Moving docker files into `/contrib`

docker/Dockerfile
dboyliao: Done

dboyliao:
Screen Shot 2022-09-11 at 3 04 55 PM

I run the CI locally and the error is due to missing`pip3` command.
Looks like `pip` is not installed correctly in the CI image

yungyuc:
> Screen Shot 2022-09-11 at 3 04 55 PM
> 
> I run the CI locally and the error is due to missing`pip3` command. Looks like `pip` is not installed correctly in the CI image

Could you please provide a link to the error at your CI run?  Image is not helpful in pin-pointing the error.

The nightly CI is healthy: https://github.com/solvcon/modmesh/actions/runs/3028860237 .  I would guess the error comes from your change.

.gitmodules
yungyuc: Perhaps I didn't make it clear.  What I meant was that there should not be sub-module in the repository and devenv is one way we did to avoid it.

README.md
yungyuc: Since container is not a blessed way to build the code, it should not be mentioned in README especially when a blessed approach is not included.

That is, at the moment, the project is not ready for a README.  This file should not exist (yet).

.gitmodules
dboyliao: I do remove the submodule
Please check the `Files changed` tab for detail.
I'm just telling you I also move all the docker files into `/contrib`.

README.md
dboyliao: Ok. I'll also move the README into `/contrib`

dboyliao:
> > Screen Shot 2022-09-11 at 3 04 55 PM
> > I run the CI locally and the error is due to missing`pip3` command. Looks like `pip` is not installed correctly in the CI image
> 
> Could you please provide a link to the error at your CI run? Image is not helpful in pin-pointing the error.
> 
> The nightly CI is healthy: https://github.com/solvcon/modmesh/actions/runs/3028860237 . I would guess the error comes from your change.

I'm using `act` to run the CI locally.
https://github.com/nektos/act

After inspecting the workflow, that's because I use `pyenv` on my machine and it's fighting with it.
My PR should not introduce any CI-breaking change after all.

.gitmodules
yungyuc: But the file is still in your changeset: https://github.com/dboyliao/modmesh/commit/730bef36c6e76806bd7c9da9e6385c42bea28f89

README.md
yungyuc: In that way, how about adding the information as the comments in the docker file you add?

.gitmodules
dboyliao: Oh, I forgot to remove the file.
My bad.
Sorry.

yungyuc (approved):

yungyuc:
LGTM.  Thanks for the PR.

tigercosmos:
wow love it!!

Issue #106, opened 2022-07-20
Port to windows and add it to modmesh_viewer CI workflow


yungyuc:
This issue is opened to track the ongoing work of porting to windows.

yungyuc:
@tychuang1211 is working on it.

yanagiragi:
Will we have a CI specific for Windows build? or just compatible to Windows but CI stills runs on Linux workflow?

yungyuc:
GitHub action has windows runner.  We do not need a separate CI system.

It would be ideal that both Linux and windows use the same workflow.  If that is too hard, using two workflows is acceptable.

tychuang1211:
I'm trying to port modmesh on Windows. it works smoothly in Release mode but crashes when running linear_wave app in Debug mode. 
The problem is, PySide from PyPI has it's own copy of Qt DLLs, and will try to load them when being called.
It's not a problem in Release mode because if a DLL with the same module name is already loaded in memory, the system uses the loaded DLL.
![image](https://user-images.githubusercontent.com/33109379/185093292-4b099297-1b8e-4281-aec9-b6d20eb0b2e4.png)
In Debug mode however, as shown in the picture, it loads QtCored.dll instead (which is QtCore.dll with debugging symbols). And then QtCore.dll is loaded from PySide directory, causing our problem
What I am trying to do is to build PySide6 from source so it use existing Qt DLLs,
![image](https://user-images.githubusercontent.com/33109379/185094153-1792df96-6f8a-49a3-bd12-2e64cf4d4127.png)
though according to [README.md](https://github.com/pyside/pyside-setup#building-additional-options) of pyside-setup it seems that there will be copies anyway, since there is no proper rpath support on Windows.

PR #122, opened 2022-09-16
Port to windows and Add windows build test to modmesh_viewer CI workflow

tychuang1211:
#106 

1. Support building with MSVC
2. Add OS `windows-2022` to CI workflow


.github/workflows/modmesh_viewer.yml
```diff
@@ -51,10 +56,27 @@ jobs:
         pip3 install -U setuptools
         pip3 install -U numpy pytest flake8
 
+    - name: Install Qt (windows)
+      if: runner.os == 'Windows'
```
yungyuc: Is there a reason not to use ``${{ startsWith(matrix.os, 'windows') }}``?

cpp/modmesh/base.hpp
```diff
@@ -47,6 +47,10 @@
 
 namespace modmesh
 {
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
```
yungyuc: Is it a right way to include in a namespace?

cpp/modmesh/spacetime/core.hpp
```diff
@@ -7,6 +7,7 @@
 
 #include <memory>
 #include <vector>
+#include <functional>
```
yungyuc, line 10: Thanks for catching this!

.github/workflows/modmesh_viewer.yml
```diff
@@ -15,8 +15,9 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-22.04, macos-12]
+          os: [ubuntu-22.04, macos-12, windows-2022]
           cmake_build_type: [Release, Debug]
+          arch: [amd64]
```
yungyuc: Do we need it?  Isn't the cloud runner always uses x86 (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)?

yungyuc (changes requested):
This is exciting!  wakuwaku

.github/workflows/modmesh_viewer.yml
tychuang1211: Actually, no. Well, except for I think it's easier to read.

yungyuc: In that case, could you use the original way to test it?  `startsWith` reads more accurate semantically.

.github/workflows/modmesh_viewer.yml
tychuang1211: Hmm, I don't really know about this. It's the input to msvc-dev-cmd and the default value seems sufficient, I think we could just delete the line.

yungyuc: Agree.  If it's not needed, let's delete it.

cpp/modmesh/base.hpp
tychuang1211: Come to think of it, we should put this outside the namespace. XD

.github/workflows/modmesh_viewer.yml
```diff
@@ -26,6 +27,10 @@ jobs:
       with:
         fetch-depth: 1
 
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: ${{ matrix.arch }}
```
tychuang1211: We'll also need to delete these two lines.

.github/workflows/modmesh_viewer.yml
tychuang1211: No problem.

cpp/modmesh/base.hpp
yungyuc: The whole fenced code section can be moved out.

yungyuc (approved):

yungyuc:
This is wonderful.  Thank you so much, @tychuang1211

Issue #103, opened 2022-07-16
Make "viewer" act like a Python interpreter


yungyuc:
The `viewer` binary invokes `pybind11::scoped_interpreter()` in the beginning of the process.  From https://github.com/solvcon/modmesh/blob/e2de7ce856b906f306d1db48f8b5c0018e480cd9/cpp/binary/viewer/viewer.cpp#L52 to https://github.com/solvcon/modmesh/blob/e2de7ce856b906f306d1db48f8b5c0018e480cd9/cpp/modmesh/python/common.cpp#L65

It might not be too far away for us to make the `viewer` binary work like a Python interpreter, i.e., equip with interactive REPL in text console.  Then the binary can work as both an entry point of the GUI and the interpreter.

yungyuc:
It's probably a good way to do it by using IPython: https://ipython.org/ipython-doc/stable/interactive/reference.html#embedding-ipython

yungyuc:
Now `viewer` can work just like a normal Python interpreter.  The next step may be to give it a better name.

PR #136, opened 2022-10-28
Make the viewer binary a multi-purpose binary

yungyuc:
Address #103 

Change the start-up code so that the viewer binary can run in the following 4 modes:
1. viewer: open a windows
2. command: take a string as Python code
3. terminal: start text-based iPython prompt
4. pytest: run the Python test cases

In this way, the viewer binary can be used to run the Python test cases.  A new target run_viewer_pytest is added to the make file.  CI (Ubuntu, macos, and Windows) uses the new target to run unit tests.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -99,6 +99,13 @@ jobs:
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
 
+    - name: make run_viewer_pytest
+      run: |
+        make run_viewer_pytest \
+          ${JOB_MAKE_ARGS} \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
   build_macos:
 
     if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
@@ -187,6 +194,13 @@ jobs:
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
             CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
 
+      - name: make run_viewer_pytest
+        run: |
+          make run_viewer_pytest \
+            ${JOB_MAKE_ARGS} \
+            CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
   build_windows:
 
     if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
@@ -249,4 +263,12 @@ jobs:
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
-          cmake --build ${{ github.workspace }}/build --config ${{ matrix.cmake_build_type }} --target ALL_BUILD
+          cmake --build ${{ github.workspace }}/build `
+            --config ${{ matrix.cmake_build_type }} `
+            --target ALL_BUILD
+
+      - name: cmake run_viewer_pytest
+        run: |
+          cmake --build ${{ github.workspace }}/build `
+            --config ${{ matrix.cmake_build_type }} `
+            --target run_viewer_pytest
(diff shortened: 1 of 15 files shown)


.github/workflows/devbuild.yml
```diff
@@ -249,4 +263,12 @@ jobs:
...
\ No newline at end of file
+          cmake --build ${{ github.workspace }}/build `
+            --config ${{ matrix.cmake_build_type }} `
+            --target ALL_BUILD
+
+      - name: cmake run_viewer_pytest
+        run: |
+          cmake --build ${{ github.workspace }}/build `
+            --config ${{ matrix.cmake_build_type }} `
+            --target run_viewer_pytest
```
yungyuc, line 274: Unit testing on Windows is now running, but the return code is not checked yet.

There is a minor error that can be fixed in the future when we change to respect the return code.

Makefile
```diff
@@ -78,6 +78,10 @@ pytest: $(MODMESH_ROOT)/modmesh/_modmesh$(pyextsuffix)
 viewer: $(MODMESH_ROOT)/modmesh/_modmesh$(pyextsuffix)
 	make -C $(BUILD_PATH) VERBOSE=$(VERBOSE) viewer $(MAKE_PARALLEL)
 
+.PHONY: run_viewer_pytest
+run_viewer_pytest: viewer
+	cmake --build $(BUILD_PATH) --target run_viewer_pytest VERBOSE=1
```
yungyuc, line 83: Use cmake target.  This should be more robust.  Other make file targets should change to be like this in the future.

cpp/binary/viewer/viewer.cpp
```diff
@@ -53,16 +53,18 @@ PYBIND11_EMBEDDED_MODULE(_modmesh_view, mod) // NOLINT
...
     modmesh::device::MetalManager::instance();
 #endif // MODMESH_METAL
 
-    modmesh::RApplication app(argc, argv);
-    app.main()->resize(1000, 600);
-    return app.exec();
+    // Initialize the Python interpreter.
+    modmesh::python::Interpreter::instance()
+        .initialize()
+        .setup_modmesh_path();
```
yungyuc, line 63: Move the setup code to distinct functions.

cpp/modmesh/python/common.cpp
```diff
@@ -100,6 +99,21 @@ void Interpreter::setup_path()
         sys.path.insert(0, path)
 _set_modmesh_path())"""";
     pybind11::exec(cmd);
+    return *this;
+}
+
+Interpreter & Interpreter::setup_process(int argc, char ** argv_in)
+{
+    std::vector<std::string> argv;
```
yungyuc, line 107: This is a prototype implementation.  Better organization should be done in the future.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -225,11 +232,16 @@ void wrap_view(pybind11::module & mod)
                 viewer->grabPixmap().save(filename.c_str());
             },
             py::arg("filename"))
+        .def(
+            "app",
```
yungyuc, line 236: Change it to be a function so that it always get the singleton from C++.

modmesh/system.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class ModmeshArgumentParser(argparse.ArgumentParser):
+    def __init__(self, *args, **kw):
+        super().__init__(*args, **kw)
+        self.exited = False
+        self.exited_status = None
+        self.exited_message = None
+
+    def exit(self, status=0, message=None):
```
yungyuc, line 54: Override to avoid calling sys.exit.

modmesh/system.py
```diff
@@ -0,0 +1,130 @@
...
+def _run_pytest():
+    # Import pytest locally to avoid making it a dependency to the whole
+    # modmesh.
+    import pytest
+    mmpath = os.path.join(os.path.dirname(modmesh.__file__), '..', 'tests')
+    mmpath = os.path.abspath(mmpath)
+    return pytest.main(['-v', '-x', mmpath])
+
+
+def setup_process(argv):
```
yungyuc, line 111: `setup_process` is probably not a good name because the role is the main function.  But let's use it for now.

tigercosmos:
there is a bug, even if I clean up the code in the interpreter and run it, the graph result is still there.
Also, I think it's better to have a console window with the interpreter.

tigercosmos:
Actually, I think improving the interpreter could be a good project. probably can open another issue. like and the syntax highlight or something.

yungyuc:
> there is a bug, even if I clean up the code in the interpreter and run it, the graph result is still there.
> Also, I think it's better to have a console window with the interpreter.

Look forward to a PR.  This PR is merged and it should be easier for you to work on it.

PR #137, opened 2022-10-29
ignore vs environment files

tigercosmos:


.gitignore
```diff
@@ -12,3 +13,9 @@ setup.mk
 *.pyo
 *.so
 *.dylib
+*.pyd
+*.vcx*
+*.sln
+*log
```
yungyuc, line 19: `*log` looks like a little bit wild.  Could it be more specific?

tigercosmos: VS creates a `modmesh.tlog` folder, please refer to this: https://stackoverflow.com/questions/41417471/what-are-these-tlog-files-should-these-be-part-of-my-source-control
I just think something like `debug.log`, `tmp.log.1` should also be ignored.

yungyuc: `*log` will also ignore a directory named like `syslog`.  I do not want that.

What about making it like `*.tlog`?  Additional ignoring entries can be added when we see them.

tigercosmos: sure

tigercosmos:
I am closing this one, another PR will include this

yungyuc:
No problem.  Just for your information.  I am fine to have small PRs as long as they make sense.  This one makes sense.

PR #143, opened 2022-11-03
Fix Python version to 3.10 on mac

yungyuc:
Github action macos runner upgrades to Python 3.11 and the installation conflicts with the way modmesh ci scripts installs dependencies.  Fix the Python version to 3.10 for now to work around the conflict.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -119,6 +119,8 @@ jobs:
 
     strategy:
       matrix:
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
         os: [macos-11, macos-12]
         cmake_build_type: [Release]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -137,13 +139,18 @@ jobs:
...
+          brew link --force --overwrite python@3.10
+          brew install qt6
 
       - name: dependency by pip
         run: |
-          echo "which pip3: $(which pip3)"
-          pip3 install -U setuptools
-          pip3 install -U numpy pytest flake8
+          echo "which python3: $(which python3)"
+          ls -al $(which python3)
```
yungyuc, line 150: Add additional debug logging.

.github/workflows/devbuild.yml
```diff
@@ -137,13 +139,18 @@ jobs:
...
+          python3 -m pip -v install --upgrade setuptools
```
yungyuc, line 151: Change to use the Python interpreter to launch pip to make sure it's the pip installed in the interpreter.

PR #144, opened 2022-11-04
WIP: add python console

tigercosmos:


cpp/modmesh/view/RMainWindow.cpp
```diff
@@ -36,9 +36,16 @@ namespace modmesh
 
 void RMainWindow::setUp()
 {
-    m_pytext = new RPythonText(QString("Python"), this);
+    auto console = std::make_shared<RPythonConsole>(QString("Console"), this);
+
+    m_pyconsole = console;
+    m_pyconsole->setAllowedAreas(Qt::LeftDockWidgetArea | Qt::RightDockWidgetArea);
+
+    m_pytext = new RPythonText(QString("Python"), this, Qt::WindowFlags(), console);
```
tigercosmos: share the console to `RPythonText`, since we need to update the output

cpp/modmesh/view/RPythonConsole.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void RPythonConsole::setUp()
+{
+    m_text->setReadOnly(true);
+    m_text->setPlainText(QString(""));
+}
+
+void RPythonConsole::setConsole(std::string const & stdout_str, std::string const & stderr_str)
+{
+    std::stringstream ss;
```
tigercosmos, line 61: could be a temporary format
a more complete version can be in future work, such as red color for error.

modmesh/apputil.py
```diff
@@ -86,14 +95,19 @@ def get_appenv(name=None):
 get_appenv(name='master')
 
 
-def run_code(code):
+def run_code(code, redirectStdOutFile=None, redirectStdErrFile=None):
```
tigercosmos, line 100: a possible improvement could be recording the history of the outputs.

cpp/modmesh/view/RMainWindow.hpp
```diff
@@ -52,13 +54,15 @@ class RMainWindow
...
     RPythonText * pytext() { return m_pytext; }
+    RPythonConsole * pyconsole() { return m_pyconsole.get(); }
     R3DWidget * viewer() { return m_viewer; }
 
 private:
 
     void setUp();
 
     RPythonText * m_pytext = nullptr;
+    std::shared_ptr<RPythonConsole> m_pyconsole = nullptr;
```
yungyuc: A very good reason is needed to justify the use of shared pointer with Qt objects, which has its own ownership system.  Shared pointer does not work well with Qt pointers.  I do not see why you need the shared pointer here.

cpp/binary/viewer/CMakeLists.txt
```diff
@@ -46,6 +47,15 @@ qt_add_executable(
     ${MODMESH_PYMOD_SPACETIME_SOURCES}
 )
 
+if(WIN32)
+    set_target_properties(
+        viewer
+        PROPERTIES
+        LINK_FLAGS_DEBUG "/SUBSYSTEM:CONSOLE" # open the console for debug mode
```
yungyuc, line 54: If it opens an additional debug console in release build, it needs to be changed.

cpp/modmesh/view/RPythonConsole.cpp
yungyuc: Use `modmesh::Formatter` for string formatting.

cpp/modmesh/view/RPythonText.cpp
```diff
@@ -65,16 +69,50 @@ void RPythonText::runCode()
 {
     namespace py = pybind11;
 
+    const std::string redirect_stdout_file_path = "stdout.txt";
```
yungyuc: We may not redirect buffered I/O to files unconditionally like this.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -160,6 +160,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRApplication
                 {
                     return self.main()->pytext();
                 })
+            .def_property_readonly(
```
yungyuc, line 235: Things get exposed from C++ to Python need corresponding unit tests.  At the moment I do not see the need to access it from Python.

modmesh/apputil.py
yungyuc: This is not a right way to redirect buffered I/O in either Python or C++.

yungyuc (changes requested):
Please take a look at #145.

cpp/modmesh/view/RMainWindow.hpp
tigercosmos: I use shared_ptr because the pointer is shared. I think a raw pointer is also fine.

cpp/binary/viewer/CMakeLists.txt
tigercosmos: in release mode, there is no cmd prompt.

cpp/modmesh/view/RPythonText.cpp
tigercosmos: what is a proper way?

yungyuc: If by default it's redirected, there must be a runtime-configurable way to change it back.  If it's undirected by default, a runtime-configurable way needs to be provided to turn on the redirection.

There may not be hard-coded file names to be written.

cpp/modmesh/view/RMainWindow.hpp
yungyuc: You should use signal/slot for accessing other Qt objects to make the event loop and ownership manager happy.

cpp/binary/viewer/CMakeLists.txt
yungyuc: To make it apparent, I recommend to only add the switch when it's not in release mode.

cpp/modmesh/view/RMainWindow.hpp
tigercosmos: could you give some references?

cpp/modmesh/view/RPythonText.cpp
tigercosmos: I see, in this case, it should be redirected, so there must be a configuration file?

cpp/modmesh/view/RMainWindow.hpp
yungyuc: Googling the keywords in the sentence should be sufficient.

cpp/modmesh/view/RPythonText.cpp
yungyuc: Yes, it should be redirected, but there does not need to be a configuration file.  Runtime configurability does not depend on configuration file.

Issue #104, opened 2022-07-17
Turn the "python" widget into a REPL interactive shell


yungyuc:
Currently the viewer is using a primitive widget to take Python code to run and it is not easy to use:

<img width="1112" alt="image" src="https://user-images.githubusercontent.com/399122/179379311-7bda9788-ff55-478f-86fc-5461caf8a45d.png">

There should be a REPL widget working like the Python REPL or iPython.

- [x] Add an interactive Python command console #145
- [x] Remove the Python text box at the right-hand side #146
- [x] Redirect the buffered outputs to the console display, see discussions in https://github.com/solvcon/modmesh/pull/144#discussion_r1014557732 @tigercosmos
  - #148 
- ~~Support <shift-enter> to type  more than one line.~~
- ~~Add a `run` button for the command line, it will be more straightforward.~~
  - It might not be the best way for the console UI.  Need more thoughts.

tai271828:
REPL is a best-practice commonly seen in this use case. +1 to the idea.

yanagiragi:
LGTM. How about also provide an another feature to load *.py and execute it in the viewer?

yungyuc:
> LGTM. How about also provide an another feature to load *.py and execute it in the viewer?

Agree.  It could be done by extending the new "app" sub system (#100).

yungyuc:
@tigercosmos As we discussed in https://github.com/solvcon/modmesh/pull/144#discussion_r1014557732 , we should redirect the buffered outputs to the console display.

tigercosmos:
@yungyuc yes, I will work on it.
I think we can also add the following as TODOs
- add a `run` button for the command line, it will be more straightforward.
- support <shift-enter> to go to the next line

yungyuc:
> @yungyuc yes, I will work on it.
> 
> I think we can also add the following as TODOs
> 
> - add a `run` button for the command line, it will be more straightforward.
> 
> - support <shift-enter> to go to the next line

Thanks.  Agree.  Items added.

yungyuc:
@tigercosmos There is one remaining part unfinished:

* Support to type more than one line.

Do you plan to finish it?

tigercosmos:
@yungyuc nit for now, I will unassign myself

yungyuc:
> @yungyuc nit for now, I will unassign myself

Thanks for letting me know.  I am closing it.

PR #145, opened 2022-11-05
Add an interactive Python command console

yungyuc:
For #104 

Add a dock widget at the bottom of the main window that a horizontally split text edit areas.  The upper and larger one is for displaying the history of the input Python commands.  The lower and narrow one is to input a Python command for execution.

The command line supports history navigation through up and down arrow keys on the keyboard, and uses enter for sending the command for execution.  It does not yet support multi-line Python code.

Improvements that are needed in the future:
* The output from Python goes to the process stdout and stderr that are not redirected to the Qt widget.  It should be in the future.
* The Python code dock widget on the right-hand side of the main window will be kept for a while before all the apps are migrated to use the new Python command console.

### DIFF
--- a/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
+++ b/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2022, Yung-Yu Chen <yyc@solvcon.net>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/view/RPythonConsoleDockWidget.hpp>
+#include <QVBoxLayout>
+#include <QKeyEvent>
+
+namespace modmesh
+{
+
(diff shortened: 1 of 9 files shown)


cpp/modmesh/view/RMainWindow.cpp
```diff
@@ -40,6 +40,10 @@ void RMainWindow::setUp()
     m_pytext->setAllowedAreas(Qt::LeftDockWidgetArea | Qt::RightDockWidgetArea);
     addDockWidget(Qt::RightDockWidgetArea, m_pytext);
 
+    m_pycon = new RPythonConsoleDockWidget(QString("Console"), this);
+    m_pycon->setAllowedAreas(Qt::BottomDockWidgetArea | Qt::TopDockWidgetArea);
```
yungyuc, line 44: Allow the console to be at the bottom or the top.  Having it in the left or right could make it too narrow.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -0,0 +1,155 @@
...
+    m_history_edit->insertPlainText(m_command_edit->toPlainText());
+    m_history_edit->insertPlainText("\n");
+    m_command_edit->setPlainText("");
+    m_command_string = "";
+    m_current_command_index = static_cast<int>(m_past_command_strings.size());
+    auto & interp = modmesh::python::Interpreter::instance();
+    interp.exec_code(code);
+}
+
+void RPythonConsoleDockWidget::navigateCommand(int offset)
```
yungyuc, line 103: Navigate through the history.  Slight complexity.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+    int const cmdsize = static_cast<int>(m_past_command_strings.size()); // make msc happy.
+    if (cmdsize == m_current_command_index)
+    {
+        m_command_string = m_command_edit->toPlainText().toStdString();
+    }
+
+    int new_index = m_current_command_index + offset;
```
yungyuc, line 111: Limit the range of `new_index`.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -0,0 +1,155 @@
...
+    if (new_index > cmdsize)
+    {
+        new_index = cmdsize;
+    }
+    else if (new_index < 0)
+    {
+        new_index = 0;
+    }
+
+    if ((0 == m_past_command_strings.size()) || (new_index == m_current_command_index))
```
yungyuc, line 121: If there's nothing in the history or the new index does not change, do nothing.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -0,0 +1,155 @@
...
+            m_current_command_index = new_index;
+        }
+        else // new_index < m_current_command_index
+        {
+            m_command_edit->setPlainText(QString::fromStdString(m_past_command_strings[new_index]));
+            m_current_command_index = new_index;
+        }
+        // Move cursor to the end of line.
+        {
+            QTextCursor cursor = m_command_edit->textCursor();
```
yungyuc, line 146: `textCursor()` returns a copy.

modmesh/apputil.py
```diff
@@ -52,16 +52,18 @@ class AppEnvironment:
...
 
+    :ivar globals:
+        The global namespace of the application.
     :ivar locals:
         The local namespace of the application.
     """
     def __init__(self, name):
-        self.globals = {}
-        self.locals = {
+        self.globals = {
```
yungyuc, line 61: Change to have the symbols to appear in the "module global".

yungyuc (commented):
@tigercosmos It happens that I am working on a feature similar to your #144.  But in #144 there is not history navigation.  I think we can take both PRs after reviewing both.

yungyuc:
@tigercosmos I am checking it in for discussions in the sprint later this afternoon.  We can still discuss in this PR.

modmesh/system.py
```diff
@@ -111,4 +114,14 @@ def enter_main(argv):
         sys.stderr.write('mode "{}" is not supported'.format(args.mode))
     return ret
 
+
+def exec_code(code):
+    try:
+        apputil.run_code(code)
+    except Exception as e:
```
tigercosmos, line 121: this doesn't work well I think.
for example, I enter `print(a)` (with undefined `a`) in the console, it will just crash.

tigercosmos:
@yyc so based on this change, I will modify #144 as the followings:
- `RPythonText`'s stdout and stderr will go to this new `RPythonConsoleDockWidget`.
-  `RPythonConsoleDockWidget`'s stdout and stderr will just show right after the command.

yungyuc:
You tagged another one but not me.  Hope he/she does not mind :)

> * `RPythonText`'s stdout and stderr will go to this new `RPythonConsoleDockWidget`.

Yes.  But in general, the buffered output of the Python interpreter should be redirected.  We do not need to limit it to the `RPythonText`, which will be removed later.

> * `RPythonConsoleDockWidget`'s stdout and stderr will just show right after the command.

And we should find an apparent way to distinguish the command from the output.

PR #146, opened 2022-11-06
Remove the Python text box at the right-hand side

yungyuc:
Change all applications to set the Python command line instead.

Ref #104


cpp/modmesh/view/CMakeLists.txt
```diff
@@ -9,7 +9,6 @@ set(MODMESH_VIEW_HEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/RAxisMark.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/RMainWindow.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/RPythonConsoleDockWidget.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/RPythonText.hpp
```
yungyuc, line 12: Bye bye, `RPythonText`.

cpp/modmesh/view/RMainWindow.cpp
```diff
@@ -28,20 +28,15 @@
...
 
 void RMainWindow::setUp()
 {
-    m_pytext = new RPythonText(QString("Python"), this);
-    m_pytext->setAllowedAreas(Qt::LeftDockWidgetArea | Qt::RightDockWidgetArea);
-    addDockWidget(Qt::RightDockWidgetArea, m_pytext);
-
     m_pycon = new RPythonConsoleDockWidget(QString("Console"), this);
-    m_pycon->setAllowedAreas(Qt::BottomDockWidgetArea | Qt::TopDockWidgetArea);
+    m_pycon->setAllowedAreas(Qt::AllDockWidgetAreas);
```
yungyuc, line 39: Allow narrow command line dock on left and right.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -109,25 +109,34 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRLine
...
 
         namespace py = pybind11;
 
         (*this)
-            .def_property("code", &wrapped_type::code, &wrapped_type::setCode)
+            .def_property(
+                "command",
+                [](wrapped_type const & self)
+                {
+                    return self.command().toStdString();
```
yungyuc, line 129: Later when we have sufficiently many such wrappers, we should add a caster for `QString`.

modmesh/app/bad_euler1d.py
```diff
@@ -37,10 +37,8 @@
 
 
 def load_app():
-    view.app().pytext.code = """
-# Need to hold the win object to keep PySide alive.
-win, svr = mm.app.bad_euler1d.run(animate=True, interval=10)
-""".lstrip()
+    cmd = "win, svr = mm.app.bad_euler1d.run(animate=True, interval=10)"
```
yungyuc, line 40: `pycon` takes a one liner.  It does not yet have a check in the engine so that we can fine-tune the implementation.

yungyuc:
@tigercosmos I am merging and would like you have your comments.

cpp/modmesh/view/wrap_view.cpp
tigercosmos: maybe open an issue, otherwise it's easy to forget lol

yungyuc: Sounds like a good idea.  After filing one could you tag it here?

PR #148, opened 2022-11-09
Console stdout stderr

tigercosmos:
TODO checklist:
- [x] runtime configuration
- [x] print(module)
- [x] mouse double click and select
- [x] scroll to the end
- [x] prompt with `>>>`
- [ ]  ~~save logs of execution~~ (not in this PR)


yungyuc:
Is it correct that the PR is still in draft mode and not ready for review?

yungyuc:
I tested the code a little bit.  Some issues:

* `print(module)` does not show anything.
* Mouse double-clicking in the history box selects either a single word (good; that's what we want) or the whole box (not helpful).  I would like double-clicking selects either a single words or a whole line but excluding the prompt.  Selecting only an output section may be useful too.
* The history text box does not scroll to the bottom when new command is added beyond the text area.
* Conventionally the Python interactive command prompt is `>>> `, not `$ `, which looks strange to a Python user.

image

cpp/modmesh/python/common.hpp
```diff
@@ -490,7 +490,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
     Interpreter & setup_process();
 
     int enter_main();
-    void exec_code(std::string const & code);
+    void exec_code(std::string const & app_env_name, std::string const & code);
```
tigercosmos: the user can reuse the certain environment to run the code

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -45,9 +45,21 @@ void RPythonConsoleDockWidget::appendPastCommand(const std::string & code)
...
+    int endPos = static_cast<int>(text.size());
+    cursor.setPosition(startPos);
+    cursor.setPosition(endPos, QTextCursor::KeepAnchor);
+    setTextCursor(cursor);
+}
+
 void RPythonCommandTextEdit::keyPressEvent(QKeyEvent * event)
 {
-    if (Qt::Key_Return == event->key())
+    if (Qt::Key_Return == event->key() || Qt::Key_Enter == event->key())
```
tigercosmos, line 62: the normal enter + the number pad enter

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +124,90 @@ void RPythonConsoleDockWidget::executeCommand()
...
-    m_history_edit->insertPlainText(m_command_edit->toPlainText());
-    m_history_edit->insertPlainText("\n");
+    printCommandHistory();
     m_command_edit->setPlainText("");
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    interp.exec_code(code);
+
+    interp.exec_code(m_config["appEnvName"], code);
```
tigercosmos: run the code with the provided app environment according to the config

consider that in the future, viewer could support switching app environments, which will be helpful

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
         self.globals = {
             # Give the application an alias of the top package.
             'mm': importlib.import_module('modmesh'),
             'appenv': self,
         }
         self.locals = {}
-        self.name = name
         # Each run of the application appends a new environment.
-        environ[name] = self
+        environ[config['appEnvName']] = self
```
tigercosmos: IMO, this line should live in `new_appenv`

tigercosmos: IMO, this line should live in `new_appenv`

modmesh/system.py
```diff
@@ -119,13 +131,8 @@ def enter_main(argv):
     return ret
 
 
-def exec_code(code):
-    try:
-        apputil.run_code(code)
-    except Exception as e:
```
tigercosmos: should not put exception here, otherwise we cannot catch stderr

cpp/modmesh/view/RPythonConsoleDockWidget.hpp
```diff
@@ -79,14 +96,25 @@ public slots:
...
     std::deque<std::string> m_past_command_strings;
     int m_current_command_index = 0;
     size_t m_past_limit = 1024;
 
+    QString m_commandHtml = "<font color=\"Black\"> <b>&#62;&#62;&#62;</b> ";
+    QString m_stderrHtml = "<font color=\"Red\">";
+    QString m_stdoutHtml = "<font color=\"Blue\">";
+    QString m_endHtml = "</font><br>";
+
+    std::unordered_map<std::string, std::string> m_config;
```
tigercosmos: the same config shared with `AppEnvironment`

modmesh/system.py
```diff
@@ -83,6 +82,19 @@ def _run_viewer(argv):
...
     wm.resize(w=1000, h=600)
     wm.show()
+
+    config = {
+        'appEnvName': 'main',
+        'redirectStdOutFile': 'stdout_viewer.txt',
+        'redirectStdErrFile': 'stderr_viewer.txt',
+    }
+
+    # setup new app environment
```
tigercosmos: Python Console shares the same config with `AppEnvironment`

currently, only one `AppEnvironment` is created at the beginning.

in the future, if there is new `AppEnvironment` created, Python Console should also update its config

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
-def get_appenv(name=None):
-    if None is name:
+    def run_code(self, code):
+        with open(self._config['redirectStdOutFile'], 'w') as f1:
+            with contextlib.redirect_stdout(f1):
+                with open(self._config['redirectStdErrFile'], 'w') as f2:
+                    with contextlib.redirect_stderr(f2):
+                        try:
+                            exec(code, self.globals, self.locals)
+                        except Exception as e:
```
tigercosmos: put exception to here

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
yungyuc: viewer can already load different app in the same process.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +124,90 @@ void RPythonConsoleDockWidget::executeCommand()
...
+    printCommandOutput();
+}
+
+void RPythonConsoleDockWidget::printCommandOutput()
```
yungyuc: GUI/Qt functions use the `camelCase` style.  Non-GUI/Qt should use `snake_case` style.

cpp/modmesh/view/RPythonConsoleDockWidget.hpp
yungyuc: We cannot do configuration in this way, please fix with:

1. Use individual data members for the information specific to this widget.
2. Exclude information (configuration) that is not part of this widget.

Note: for an HPC application tree and hash are usually no-go, and it's especially so for configuration.  Oftentimes we need a singleton in memory for speed.

This PR is associated with #104 for GUI.  The configuration added should be ad hoc to the GUI widget.  General configuration is outside the scope of the associated issue, and should be separately considered in #84.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -122,12 +122,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRPythonConsoleDockWidget
         namespace py = pybind11;
 
         (*this)
+            .def("setConfig", [](wrapped_type & self, std::string const & key, std::string const & value)
```
yungyuc: Exposing a hash table to Python is pointless.  The Python `dict` and attribute access already works like that.  Please use individual data members for widget information and expose the accessors accordingly.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -122,12 +122,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRPythonConsoleDockWidget
...
+                 { self.setConfig(key, value); })
             .def_property(
-                "command",
-                [](wrapped_type const & self)
-                {
-                    return self.command().toStdString();
-                },
+                "command", [](wrapped_type const & self)
+                { return self.command().toStdString(); },
```
yungyuc: Do not change the style.

modmesh/apputil.py
yungyuc: Do not mingle the non-GUI construct `environ` with GUi widget.  Avoid taking the name from `config`.  The `config` object should be removed too.

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+        # replace the config value; otherwise keep the default value
+        for k, v in config.items():
+            if v is not None:
+                self._config[k] = v
 
-
-def get_appenv(name=None):
-    if None is name:
+    def run_code(self, code):
+        with open(self._config['redirectStdOutFile'], 'w') as f1:
```
yungyuc: Please use a single level of nesting.  In this revision, you descend by 4 levels before the try/except block, and it is too deep for readability and maintainability.

yungyuc: It does not look right to open a file in Python for consumption in C++.  By doing this, when accessing the file from C++, it could wind back in Python for I/O.  It is hard to maintain and slow.

Please open a file in C++ and then wrap it to Python instead.

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+            with contextlib.redirect_stdout(f1):
+                with open(self._config['redirectStdErrFile'], 'w') as f2:
+                    with contextlib.redirect_stderr(f2):
+                        try:
+                            exec(code, self.globals, self.locals)
+                        except Exception as e:
+                            print(("{}: {}".format(
```
yungyuc: One convention I didn't mention clearly: `print` is only used for debugging in modmesh.  Please use `sys.stdout` for writing to buffered I/O.  That was how it worked: https://github.com/solvcon/modmesh/blob/e40e0d9fd7c5d16963425518a31d24e7df86e151/modmesh/system.py#L126

We will need a general logging facility working in both C++ and Python.  That means we should not directly use Python logging.

At the time being we do not have home brewed logging, and can temporarily accept the use of Python `logging`.

modmesh/system.py
```diff
@@ -83,6 +82,19 @@ def _run_viewer(argv):
     wm.windowTitle = "Modmesh Viewer"
     wm.resize(w=1000, h=600)
     wm.show()
+
+    config = {
```
yungyuc: Do not use a Python dictionary or C++ hash or tree to keep configurations.  They are slow for HPC.

There are cases that we may tolerate the slowness of a hash/tree, but I do not want to make it enter modmesh so soon.  It delivers a wrong impression to devs and users.

modmesh/system.py
```diff
@@ -83,6 +82,19 @@ def _run_viewer(argv):
...
+        'appEnvName': 'main',
```
yungyuc: We generally do not use `camelCase`.  In modmesh, only GUI related code uses `camelCase` because that is Qt convention.  Everywhere else we should use `snake_case`.

yungyuc (changes requested):
I think most of the change in apputil.py and system.py should be reverted.  You added too much Python code and it does not look right.  Most of the I/O redirection should be done in C++, not Python.  The change in Python should simply changing the underneath file descriptors.

modmesh/system.py
yungyuc: And please keep the name `master`.  Do not use `main`.  I do not like the pronunciation of `main`, which easily gets confused with that of "M".

modmesh/system.py
```diff
@@ -83,6 +82,19 @@ def _run_viewer(argv):
...
+        'redirectStdOutFile': 'stdout_viewer.txt',
```
yungyuc: Why do we have file names?  stdout and stderr from Python should be redirected to the Qt widget but not any files on the disks.  We do not want to pay for the additional cost of the storage.

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+                self._config[k] = v
 
-
-def get_appenv(name=None):
-    if None is name:
+    def run_code(self, code):
+        with open(self._config['redirectStdOutFile'], 'w') as f1:
+            with contextlib.redirect_stdout(f1):
+                with open(self._config['redirectStdErrFile'], 'w') as f2:
+                    with contextlib.redirect_stderr(f2):
```
ezio-melotti: Here you can use (parenthesized context managers)[https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers) or [ExitStack](https://docs.python.org/3.10/library/contextlib.html#supporting-a-variable-number-of-context-managers) to avoid the multiple levels of indentation.

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+                        try:
+                            exec(code, self.globals, self.locals)
+                        except Exception as e:
+                            print(("{}: {}".format(
+                                type(e).__name__, str(e))), file=sys.stderr)
```
ezio-melotti:
```suggestion
                            print(f"{type(e).__name__}: {e}", file=sys.stderr)
```

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+                            print("traceback:", file=sys.stderr)
+                            traceback.print_stack()
+                        sys.stdout.flush()
+                        sys.stderr.flush()
```
ezio-melotti: You should be able to do `print(..., flush=True)` directly if you only have one `print`.  If you have multiple, doing it once at the end might be better.

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
+
+
+def new_appenv(config=None):
+    name = config.get('appEnvName', None)
+    if None is name or name == '':
```
ezio-melotti:
```suggestion
    if name is None or name == '':
```

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
         i = 0
         name = 'anonymous%d' % i
```
ezio-melotti:
```suggestion
        name = f'anonymous{i}'
```

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
         while name in environ:
             i += 1
             name = 'anonymous%d' % i
```
ezio-melotti: Is there an upper limit to this?  Maybe it would be better to rewrite it as 
```suggestion
        for i in range(10):
            name = f'anonymous{i}'
            if name not in environ:
                break
        else:
            raise ValueError("can't find app name")
```

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
     app = environ.get(name, None)
+
     if None is app:
```
ezio-melotti:
```suggestion
    if app is None:
```

modmesh/apputil.py
```diff
@@ -57,46 +58,68 @@ class AppEnvironment:
...
-def run_code(code):
-    has_key = False
-    for k in reversed(environ):
-        has_key = True
-        break
-    if not has_key:
-        raise KeyError("No AppEnviron is available")
-    aenv = environ[k]
-    aenv.run_code(code)
+    if None is app:
```
ezio-melotti:
```suggestion
    if app is None:
```

modmesh/apputil.py
yungyuc: For making the two conditions in the line use consistent style, changing the order is good for reading.

If not in this scenario, (like a single condition), I prefer putting constant in the first place in the conditional checking.  It is the same as a convention commonplace in C code, which is to reduce the chance of typo to treat equality `==` as assignment `=`, because

```cpp
if (nullptr = pointer)
```

is an error compiler must complain about.

With the aid of modern compilers or clang-tidy, the trick is not absolutely necessary, but good to have.

My preference may change if a convincing argument is provided.

modmesh/apputil.py
yungyuc: There was not an upper limit in mind.  It's harmless to have one, but there need to be a warning or error when the upper limit is hit.

modmesh/apputil.py
yungyuc: This is where constant should be put in the first position of the condition.

modmesh/apputil.py
ezio-melotti: Three reasons:
1. With a single `=` in the condition Python already complains, so there is no risk of accidental assignment:
```py
>>> name = 'foo'
>>> if name = None: pass
  File "<stdin>", line 1
    if name = None: pass
       ^^^^^^^^^^^
SyntaxError: invalid syntax. Maybe you meant '==' or ':=' instead of '='?
```
2. In this case we are using `is` instead of `==`, so there is no risk of typos and accidentally missing an `=`.
3. `variable == constant` reads better and is the convention followed in Python (and many other languages), and in this specific situation it is also symmetric with the second part of the condition

modmesh/apputil.py
ezio-melotti: I updated the suggestion above to raise an error in case the name can't be found after checking `anonymous0-9` (you might want to tweak the error message/exception type).

yungyuc: Side track: it's hard for me to decide whether or not to put a period at the end of an error message.  Sometimes I do sometimes I don't.  Any idea for a guideline?

ezio-melotti: I tried a few expressions that raised errors and they don't seem to include the period.  Generally they are in the form of `ExceptionName: message`.

yungyuc: Good to know that.  Thanks.  The for/else looks great too.

yungyuc: I am separating this enhancement to a distinct PR: https://github.com/solvcon/modmesh/pull/153

modmesh/apputil.py
```diff
@@ -68,8 +70,32 @@ def __init__(self, name):
...
+
+        oldStdoutFd = sys.stdout
+        oldStderrFd = sys.stderr
+        try:
+            if self._stdoutFd <= 0:
+                raise Exception("wrong stdout fd:", self._stdoutFd)
+            if self._stderrFd <= 0:
+                raise Exception("wrong stderr fd:", self._stderrFd)
+
+            sys.stdout = msvcrt.get_osfhandle(self._stdoutFd)
```
tigercosmos: failed here, although `self._stdoutFd` could be `3` from C++ `fopen`, and `msvcrt.get_osfhandle` translates it to Python's handle, but it seems python cannot take this FD, probably cannot share the same FD for different processes.
```
OSError: [Errno 9] Bad file descriptortraceback:  File "C:\Users\tiger\modmesh\modmesh\system.py", line 105, in enter_main
    ret = _run_viewer(argv)
  File "C:\Users\tiger\modmesh\modmesh\system.py", line 80, in _run_viewer
    return view.launch()
  File "C:\Users\tiger\modmesh\modmesh\view.py", line 95, in launch
    return app.exec()
  File "C:\Users\tiger\modmesh\modmesh\system.py", line 117, in exec_code
    apputil.run_code(code, redirectStdOutFd, redirectStdErrFd)
  File "C:\Users\tiger\modmesh\modmesh\apputil.py", line 129, in run_code
    aenv.run_code(code)
  File "C:\Users\tiger\modmesh\modmesh\apputil.py", line 99, in run_code
    traceback.print_stack()
```

modmesh/apputil.py
```diff
@@ -70,21 +70,32 @@ def __init__(self, name):
...
-                            traceback.print_stack()
+        import msvcrt
+
+        oldStdoutFd = sys.stdout
+        oldStderrFd = sys.stderr
+        try:
+            if self._stdoutFd <= 0:
+                raise Exception("wrong stdout fd:", self._stdoutFd)
+            if self._stderrFd <= 0:
+                raise Exception("wrong stderr fd:", self._stderrFd)
```
ezio-melotti: `Exception` is for catching, not for raising.  A `ValueError` would be more appropriate here:
```suggestion
            if self._stdoutFd <= 0:
                raise ValueError("wrong stdout fd:", self._stdoutFd)
            if self._stderrFd <= 0:
                raise ValueError("wrong stderr fd:", self._stderrFd)
```

modmesh/apputil.py
```diff
@@ -70,21 +70,32 @@ def __init__(self, name):
...
+
+            sys.stdout = msvcrt.get_osfhandle(self._stdoutFd)
+            sys.stderr = msvcrt.get_osfhandle(self._stderrFd)
+            exec(code, self.globals, self.locals)
+            
+            sys.stdout = oldStdoutFd
+            sys.stderr = oldStderrFd
+        except Exception as e:
```
ezio-melotti: Here you could either use `ValueError` or leave `Exception` if you want to catch other types of exception too.

modmesh/apputil.py
```diff
@@ -70,21 +70,32 @@ def __init__(self, name):
...
+            sys.stdout = oldStdoutFd
+            sys.stderr = oldStderrFd
+            sys.stderr.write(("{}: {}".format(type(e).__name__, str(e))))
```
ezio-melotti:
```suggestion
            sys.stderr.write(f"{type(e).__name__}: {e}")
```
There might be something in the `traceback` module that does this for you already.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +124,112 @@ void RPythonConsoleDockWidget::executeCommand()
...
     m_command_edit->setPlainText("");
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    interp.exec_code(code);
+
+#ifdef _MSC_VER
+    FILE *stdoutFile, *stderrFile;
+
+    errno_t err = fopen_s(&stdoutFile, "stdout.txt", "wb");
```
yungyuc: Writing to disks incurs unwanted runtime overhead.  I do not understand why do you use the file names.  Could you elaborate?

modmesh/apputil.py
```diff
@@ -68,8 +70,32 @@ def __init__(self, name):
...
         environ[name] = self
 
+        self._stdoutFd = -1
+        self._stderrFd = -1
+
     def run_code(self, code):
-        exec(code, self.globals, self.locals)
+        import msvcrt
+
+        oldStdoutFd = sys.stdout
```
yungyuc: The names baffled me.  I think you intended to mean `old_stdout_fobj`?

modmesh/apputil.py
yungyuc: I concur.  I do not find a lot of use cases of `raise Exception`.  We should always use a specific derived exception type.

If invalid fd is error state, the checking section should be moved outside the try block.  But I guess you intend to allow invalid fd and simply skip the fd processing.  In this case, the checking section should still be moved outside the try block.  You should use a different branch to handle the flow.

modmesh/apputil.py
yungyuc: Is there a reason that you have been trying to deal with the fd in Python, but not directly in C/C++?  Wouldn't it be easier to simply handle it in CPython API?

modmesh/apputil.py
```diff
@@ -88,14 +114,17 @@ def get_appenv(name=None):
 get_appenv(name='master')
 
 
-def run_code(code):
+def run_code(code, redirectStdOutFd=-1, redirectStdErrFd=-1):
```
yungyuc: In my imagination, the code handling fd would be more straight-forward if they are in C/C++.  In Python, the fd should be wrapped by the Python file object, which is used almost everywhere in Python.

yungyuc (changes requested):

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
tigercosmos: I just want to get file descriptor. indeed we can use `pipe()`, but more work to support both Windows and Unix.

modmesh/apputil.py
tigercosmos: I haven't tried if I can get a python file object in C++, I could try.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
yungyuc: I think it is OK to focus on one system and then the other.

modmesh/system.py
```diff
@@ -114,13 +115,26 @@ def enter_main(argv):
...
+def exec_code(code, redirectStdOutFd=-1, redirectStdErrFd=-1):
+    if redirectStdOutFd <= 0:
+        raise ValueError("wrong fd:", redirectStdOutFd)
+    if redirectStdErrFd <= 0:
+        raise ValueError("wrong fd:", redirectStdErrFd)
+
+    oldStdout = sys.stdout
+    oldStderr = sys.stderr
     try:
+        sys.stdout = msvcrt.get_osfhandle(redirectStdOutFd)
```
tigercosmos: still `OSError: [Errno 9] Bad file descriptor`

@yungyuc  I feel that we cannot open fds in C++, because Python cannot access those fds.
I also tried opening fds in Python and share to C++, but also not work.
C++ and Python should be the same process, so fd should be able to share?

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +125,101 @@ void RPythonConsoleDockWidget::executeCommand()
...
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    interp.exec_code(code);
+
+    int pipe_stdout_fd[2];
+    int pipe_stderr_fd[2];
+
+#ifdef _MSC_VER
+    /* Open a set of pipes */
+    if (_pipe(pipe_stdout_fd, 256, _O_BINARY) == -1)
```
tigercosmos: maybe it's not a good idea to call `_pipe()` here.
how about call `os.pipe()` in Python and pass to here?

tigercosmos:
I also tried to add the following in `system.py`

```py
r_stdout, w_stdout = os.pipe()
r_stderr, w_stderr = os.pipe()

sys.stdout = os.fdopen(r_stdout, "w+")
sys.stderr = os.fdopen(r_stderr, "w+")

def get_stdout_redirect_fd():
    return w_stdout

def get_stderr_redirect_fd():
    return w_stderr
```

and in the C++ code, call `interp.get_stdout_redirect_fd()` to get the fd, and then call `_read(fd)` ; however, this way doesn't work as well.

modmesh/system.py
tigercosmos: > Is there a reason that you have been trying to deal with the fd in Python, but not directly in C/C++? Wouldn't it be easier to simply handle it in CPython API?

maybe I should try this

tychuang1211:
I was testing the code and it seemed that not only did `print(module)` not show anything, but everything enclosed in angle brackets also didn't show. And I also checked the stream before and after redirection, both were correct. 

After this I found that the inserted Html in `RPythonConsoleDockWidget::printCommandStdout` and `RPythonConsoleDockWidget::printCommandStderr` were not escaped. We could simply change `QString::fromStdString(message)` to `QString::fromStdString(message).toHtmlEscaped()` in both function to fix the problem.

image

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +125,101 @@ void RPythonConsoleDockWidget::executeCommand()
...
+    cursor.movePosition(QTextCursor::End);
+    m_history_edit->setTextCursor(cursor);
+}
+
+void RPythonConsoleDockWidget::printCommandStdout(const std::string & stdout_message)
+{
+    QTextCursor cursor = m_history_edit->textCursor();
+    cursor.movePosition(QTextCursor::End);
+    m_history_edit->setTextCursor(cursor);
+    m_history_edit->insertHtml(m_stdoutHtml + QString::fromStdString(stdout_message) + m_endHtml);
```
tychuang1211:
```suggestion
    m_history_edit->insertHtml(m_stdoutHtml + QString::fromStdString(stdout_message).toHtmlEscaped() + m_endHtml);
```

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +125,101 @@ void RPythonConsoleDockWidget::executeCommand()
...
+    cursor.movePosition(QTextCursor::End);
+    m_history_edit->setTextCursor(cursor);
+}
+
+void RPythonConsoleDockWidget::printCommandStderr(const std::string & stderr_message)
+{
+    QTextCursor cursor = m_history_edit->textCursor();
+    cursor.movePosition(QTextCursor::End);
+    m_history_edit->setTextCursor(cursor);
+    m_history_edit->insertHtml(m_stderrHtml + QString::fromStdString(stderr_message) + m_endHtml);
```
tychuang1211:
```suggestion
    m_history_edit->insertHtml(m_stderrHtml + QString::fromStdString(stderr_message).toHtmlEscaped() + m_endHtml);
```

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
yungyuc: I think this is good!  But perhaps for simple coloring, we may also consider to directly use Qt API for changing color?  I am curious which way works better for the terminal.

tigercosmos: thanks @tychuang1211, I didn't think about `toHtmlEscaped`. btw, which code did you test, I think it's not the latest commit?
@yungyuc, it's inconvenient to set different colors for text if we use QT way. let's just use HTML way.

yungyuc: @tigercosmos I am OK to use HTML there for now, but convenience should not be the major factor for deciding which way to go.  The more indirect the longer runtime it takes.  To some point we will need to evaluate and pick a "better" way, of which the metrics are not defined at the moment.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -33,6 +33,43 @@
...
+    pybind11::object _stdout;
+    pybind11::object _stderr;
+    pybind11::object _stdout_buffer;
+    pybind11::object _stderr_buffer;
+
+public:
+    PyStdErrOutStreamRedirect()
+    {
+        auto sysm = pybind11::module::import("sys");
+        _stdout = sysm.attr("stdout");
```
yungyuc: Please avoid using underscore `_` in the beginning of a variable, because the standard reserves them: https://softwareengineering.stackexchange.com/a/191833

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -33,6 +33,43 @@
 namespace modmesh
 {
 
+class PyStdErrOutStreamRedirect
+{
+    pybind11::object _stdout;
```
yungyuc: Member data names should start with `m_`.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -33,6 +33,43 @@
...
+        auto sysm = pybind11::module::import("sys");
+        _stdout = sysm.attr("stdout");
+        _stderr = sysm.attr("stderr");
+        auto stringio = pybind11::module::import("io").attr("StringIO");
+        _stdout_buffer = stringio(); // Other filelike object can be used here as well, such as objects created by pybind11
+        _stderr_buffer = stringio();
+        sysm.attr("stdout") = _stdout_buffer;
+        sysm.attr("stderr") = _stderr_buffer;
+    }
+    std::string stdoutString()
```
yungyuc: It's not a Qt function, and the name should follow the `snake_case` convention.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -112,13 +161,45 @@ void RPythonConsoleDockWidget::executeCommand()
...
     std::string const code = m_command_edit->toPlainText().toStdString();
     appendPastCommand(code);
-    m_history_edit->insertPlainText(m_command_edit->toPlainText());
-    m_history_edit->insertPlainText("\n");
+    printCommandHistory();
     m_command_edit->setPlainText("");
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
+    PyStdErrOutStreamRedirect pyOutputRedirect{};
```
yungyuc: `snake_case` for local variables.

yungyuc: This is elegant.  Not even need a line of comment.

yungyuc (changes requested):
With some cosmetic improvements we are good to check in!

tigercosmos:
@yungyuc  ready for merge

tigercosmos:
```
/home/runner/work/modmesh/modmesh/cpp/modmesh/python/common.cpp:208:28: error: an exception may be thrown in function '~PyStdErrOutStreamRedirect' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
```
any idea?

yungyuc:
> ```
> /home/runner/work/modmesh/modmesh/cpp/modmesh/python/common.cpp:208:28: error: an exception may be thrown in function '~PyStdErrOutStreamRedirect' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> ```
> any idea?

https://releases.llvm.org/13.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-exception-escape.html

tychuang1211:
> ```
> /home/runner/work/modmesh/modmesh/cpp/modmesh/python/common.cpp:208:28: error: an exception may be thrown in function '~PyStdErrOutStreamRedirect' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
> ```
> 
> any idea?

the `pybind11::module::import("sys")` in `~PyStdErrOutStreamRedirect` may throw an exception (`pybind11::error_already_set`)

cpp/modmesh/python/common.cpp
```diff
@@ -171,7 +171,52 @@ void Interpreter::exec_code(std::string const & code)
...
+    return pybind11::str(m_stdout_buffer.attr("read")());
+}
+
+std::string PyStdErrOutStreamRedirect::stderr_string()
+{
+    m_stderr_buffer.attr("seek")(0);
+    return pybind11::str(m_stderr_buffer.attr("read")());
+}
+
+PyStdErrOutStreamRedirect::~PyStdErrOutStreamRedirect() noexcept(true)
```
yungyuc: Because you still may throw an exception, it should be `noexcept(false)`.  Since no specifying noexcept means `noexcept(false)`, it's not really needed.

yungyuc (changes requested):

cpp/modmesh/python/common.cpp
tigercosmos: because I only catch `pybind11::error_already_set` here?

also, if I use `try  catch(...)` for the whole deconstruction function, will it throw?

yungyuc: I do not fully understand your questions.  Could you be more specific?

tigercosmos: > Because you still may throw an exception, it should be `noexcept(false)`

is it because I only catch `pybind11::error_already_set` in the code?

if I write:
```
    try
    {
     // ..
    }
    catch (...) // capture all exception
    {
        // ...
    }
```

will it still throw exception?

yungyuc: I do not know.  `noexcept` means the function does not throw exception.  I am not sure it means no exception handling is allowed in the function, or no except is thrown outside the function.  I guess it's the latter, but if it's the former, a `try catch` block in `noexcept` would be problematic.

I would be interested in the behavior too.

yungyuc (approved):

cpp/modmesh/python/common.cpp
tigercosmos: it seems no difference for the behavior when `noexcept` is `true` or `false` if we catch all errors

```
struct Foo {
    Foo(){}
    
    ~Foo() noexcept(true){
        try {
            throw "123";
        }
        catch(...) {
            //
        }
    }
};

int main()
{
  Foo foo;
}
```

the program can run without panic with both conditions.

tigercosmos: however, this panic
```
struct Foo {
    Foo(){}

    ~Foo() noexcept(true){
        try {
            throw "123";
        }
        catch(int a) {
                std::cerr << "XX" ;
        }
    }
};

int main()
{
  Foo foo;
}
```

tychuang1211: image
https://en.cppreference.com/w/cpp/language/noexcept_spec

tigercosmos: so I think the answer is, the current code might need to use `noexcept(false)` because it only capture `error_already_set`.
if it is `catch(...)` then we can set it `noexcept(true)`

PR #155, opened 2022-11-24
add DISABLE_REDIRECT

tigercosmos:
DISABLE_REDIRECT: Disable redirecting stdout and stderr to python console

### DIFF
--- a/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
+++ b/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
@@ -129,10 +129,19 @@ void RPythonConsoleDockWidget::executeCommand()
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    modmesh::python::PyStdErrOutStreamRedirect pyOutputRedirect{};
+
+#ifdef DISABLE_REDIRECT
+    bool disable_redirect = true;
+#else
+    bool disable_redirect = false;
+#endif
+    modmesh::python::PyStdErrOutStreamRedirect pyOutputRedirect(disable_redirect);
     interp.exec_code(code);
-    printCommandStdout(pyOutputRedirect.stdout_string());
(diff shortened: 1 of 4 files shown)


cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -129,10 +129,19 @@ void RPythonConsoleDockWidget::executeCommand()
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    modmesh::python::PyStdErrOutStreamRedirect pyOutputRedirect{};
+
+#ifdef DISABLE_REDIRECT
```
yungyuc, line 133: I think we should not use a compile-time macro for this simple switch.  The switch controls something outside a loop (not to say a tight loop), so we do not worry about its runtime.  A runtime switch would do.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -129,10 +129,19 @@ void RPythonConsoleDockWidget::executeCommand()
...
+    bool disable_redirect = true;
+#else
+    bool disable_redirect = false;
+#endif
+    modmesh::python::PyStdErrOutStreamRedirect pyOutputRedirect(disable_redirect);
```
yungyuc, line 138: We should take this opportunity to fix the code style. It's not a qt object so it should be `snake_case`.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
yungyuc: The runtime switch can be added on `RPythonConsoleDockWidget` and exposed to Python, so that we can control it from the console itself.

yungyuc (changes requested):
The code needs to be changed.  It might be a quick fix and I'll see if I have time to look into it today.

yungyuc:
@tigercosmos I am working on this: https://github.com/yungyuc/modmesh/tree/feature/switch-redirect .  It should be ready later tonight.

PR #156, opened 2022-11-24
add error message if cannot find QT6

### LINKED ISSUE #149: Cmake setting for QT
for Linux, I always need to add the line manually to make QT work. Maybe have a variable to specify the QT path and also try to guess where is QT?

```
set(CMAKE_PREFIX_PATH "/home/user/Qt/6.3.2/gcc_64/lib/cmake")
```

tigercosmos:
```
ion "2.10.0")
-- pybind11_INCLUDE_DIRS: C:/Users/tiger/AppData/Local/Programs/Python/Python310/Lib/site-packages/pybind11/include;C:/Users/tiger/AppData/Local/Programs/Python/Python310/Include
-- NUMPY_INCLUDE_DIR: C:/Users/tiger/AppData/Local/Programs/Python/Python310/Lib/site-packages/numpy/core/include
CMake Error at cpp/modmesh/CMakeLists.txt:13 (message):
  Qt6 Not Found.  Try `cmake -DCMAKE_PREFIX_PATH=/path/to/Qt/6.3.2/gcc_64/lib/cmake`?


-- Configuring incomplete, errors occurred!
See also "C:/Users/tiger/modmesh/build/CMakeFiles/CMakeOutput.log".
```

fix #149


yungyuc (changes requested):
Could you also change the following? 
https://github.com/solvcon/modmesh/blob/e4c997aad6101753bc535969b8ef4ef2435ebaac/cpp/binary/viewer/CMakeLists.txt#L20

If you do not like the duplicated code, making a `.cmake` file is another way to do it, just much more involved.

cpp/modmesh/CMakeLists.txt
```diff
@@ -8,13 +8,17 @@ if (BUILD_QT)
...
-    find_package(Qt6 REQUIRED COMPONENTS Core)
-    find_package(Qt6 REQUIRED COMPONENTS Widgets)
-    find_package(Qt6 REQUIRED COMPONENTS Gui)
-    find_package(Qt6 REQUIRED COMPONENTS 3DCore)
-    find_package(Qt6 REQUIRED COMPONENTS 3DRender)
-    find_package(Qt6 REQUIRED COMPONENTS 3DInput)
-    find_package(Qt6 REQUIRED COMPONENTS 3DExtras)
+    find_package(Qt6 QUIET)
+    if(NOT Qt6_FOUND)
+        message(FATAL_ERROR  "Qt6 Not Found. Try `cmake -DCMAKE_PREFIX_PATH=/path/to/Qt/6.3.2/gcc_64/lib/cmake`?")
```
yungyuc: In an error message it's probably more elegant to not specifying minor versions, like:

```
/path/to/Qt/6.x.y/<platform>
```

And I think `lib/cmake` is redundant because `CMAKE_PREFIX_PATH=/path/to/qt/6.2.3/macos` works for me.  If it is not the case in Linux or Windows, could you add a couple of more words to make the message complete?

tigercosmos: in Windows and Linux, the path is different. I can just list three paths for all platforms.

yungyuc: Listing 3 paths sounds mouthy.  How about just indicate that "mac does not need lib/cmake"?

tigercosmos: another way is to use if condition to print messages according to the platform, I think it would be clearer for user.

yungyuc: That sounds about right.

cpp/modmesh/CMakeLists.txt
```diff
@@ -8,6 +8,19 @@ if (BUILD_QT)
...
+    find_package(Qt6 QUIET)
+    if(NOT Qt6_FOUND)
+        if(MSVC)
+            message(FATAL_ERROR  "Qt6 Not Found.\nTry to add `C:\\Qt\\6.x.y\\msvcYYYY_64\\lib\\cmake\\Qt` into \"Path\" of \"Enviroment Variables\"?")
+        endif()
+        if(APPLE)
+            message(FATAL_ERROR  "Qt6 Not Found.\nTry `CMAKE_PREFIX_PATH=/path/to/qt/6.x.y/macos`?")
+        endif()
+        if(UNIX AND NOT APPLE)
+            message(FATAL_ERROR  "Qt6 Not Found.\nTry `cmake -DCMAKE_PREFIX_PATH=/path/to/Qt/6.y.z/gcc_64/lib/cmake`?")
```
yungyuc: I think it should follow the `6.x.y` convention as above.  Or the above two may change to `6.y.z`.  I am fine either way, but they should be consistent.

yungyuc (changes requested):

cpp/modmesh/CMakeLists.txt
tigercosmos: sure

yungyuc (approved):

PR #157, opened 2022-11-25
Use a runtime switch to turn off Python console redirect

yungyuc:
The console widget redirect is turned on by default, and it can be turned off in runtime from Python:

```
# Default is true.
assert modmesh.view.app.pycon.python_redirect == True
# Turn it off by using the setter.
modmesh.view.app.pycon.python_redirect = False
```

Before we have a configuration file, which requires some careful thinking, we should use the interpreter for configuration in this way.

This PR is recreated from #155.  @tigercosmos , I think `a` is not a very distinguishable branch name.

### DIFF
--- a/cpp/modmesh/python/common.cpp
+++ b/cpp/modmesh/python/common.cpp
@@ -200,8 +200,8 @@ PythonStreamRedirect & PythonStreamRedirect::activate()
 
         // Create string-IO objects. Other file-like object can be used here as
         // well, such as objects created by pybind11.
-        auto string_io = pybind11::module::import("io").attr("StringIO");
         {
+            auto string_io = pybind11::module::import("io").attr("StringIO");
             m_stdout_buffer = string_io();
             sys_module.attr("stdout") = m_stdout_buffer;
             m_stderr_buffer = string_io();
@@ -228,13 +228,13 @@ PythonStreamRedirect & PythonStreamRedirect::deactivate()
     return *this;
 }
 
-std::string PythonStreamRedirect::stdout_string()
+std::string PythonStreamRedirect::stdout_string() const
 {
     m_stdout_buffer.attr("seek")(0);
     return pybind11::str(m_stdout_buffer.attr("read")());
 }
 
-std::string PythonStreamRedirect::stderr_string()
+std::string PythonStreamRedirect::stderr_string() const
 {
     m_stderr_buffer.attr("seek")(0);
     return pybind11::str(m_stderr_buffer.attr("read")());
--- a/cpp/modmesh/python/common.cpp
+++ b/cpp/modmesh/python/common.cpp
@@ -31,6 +31,8 @@
 
 #include <modmesh/toggle/toggle.hpp>
 
+#include <utility>
+
 #include <pybind11/numpy.h>
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
 #include <numpy/arrayobject.h>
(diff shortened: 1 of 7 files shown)


cpp/modmesh/python/common.cpp
```diff
@@ -181,44 +183,63 @@ void Interpreter::exec_code(std::string const & code)
...
 }
 
-PyStdErrOutStreamRedirect::PyStdErrOutStreamRedirect()
+PythonStreamRedirect & PythonStreamRedirect::activate()
+{
+    if (m_enabled)
+    {
+        auto sys_module = pybind11::module::import("sys");
+        // Back up old streams.
+        if (!bool(m_stdout_backup))
```
yungyuc, line 192: Do not backup if the backup variables are not empty.

cpp/modmesh/python/common.hpp
```diff
@@ -505,21 +505,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
 // Suppress the warning "greater visibility than the type of its field"
 #pragma GCC diagnostic ignored "-Wattributes"
 #endif
-class PyStdErrOutStreamRedirect
+class PythonStreamRedirect
```
yungyuc, line 508: Make the class name shorter.

cpp/modmesh/python/common.hpp
```diff
@@ -505,21 +505,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
...
 {
 
 public:
-    PyStdErrOutStreamRedirect();
-    std::string stdout_string();
-    std::string stderr_string();
-    ~PyStdErrOutStreamRedirect() noexcept(false);
+
+    PythonStreamRedirect(bool enabled)
```
yungyuc, line 513: Only constructor is needed, because of the introduction of `activate()` and `deactivate()`.

cpp/modmesh/python/common.hpp
```diff
@@ -505,21 +505,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
...
+        : m_enabled(enabled)
+    {
+    }
+
+    std::string stdout_string() const;
```
yungyuc, line 518: Fix constness.

cpp/modmesh/python/common.hpp
```diff
@@ -505,21 +505,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
...
+
+    bool is_activated() const { return bool(m_stdout_backup) || bool(m_stderr_backup); }
+    bool is_deactivated() const { return !is_activated(); }
 
 private:
-    pybind11::object m_stdout;
-    pybind11::object m_stderr;
+
+    bool m_enabled;
+    pybind11::object m_stdout_backup;
```
yungyuc, line 539: Make the member data name more accurate.

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -129,10 +129,15 @@ void RPythonConsoleDockWidget::executeCommand()
     m_command_string = "";
     m_current_command_index = static_cast<int>(m_past_command_strings.size());
     auto & interp = modmesh::python::Interpreter::instance();
-    modmesh::python::PyStdErrOutStreamRedirect pyOutputRedirect{};
+
+    m_python_redirect.activate();
     interp.exec_code(code);
-    printCommandStdout(pyOutputRedirect.stdout_string());
-    printCommandStderr(pyOutputRedirect.stderr_string());
+    if (m_python_redirect.is_activated())
```
yungyuc, line 135: Only print strings if the redirect object is activated.

tests/test_view.py
```diff
@@ -0,0 +1,50 @@
...
+    view = None
+
+
+@unittest.skipUnless(modmesh.HAS_VIEW, "Qt view is not built")
+class ViewTC(unittest.TestCase):
+
+    def test_import(self):
+        self.assertTrue(hasattr(modmesh.view, "app"))
+
+    @unittest.skip("headless testing is not ready")
```
yungyuc, line 43: This test, even if expectedFailure, crashes with macos 11.  Before we figure out a way for headless testing of GUI code, this should be skipped.

cpp/modmesh/python/common.hpp
```diff
@@ -505,21 +505,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY Interpreter
...
+    std::string stdout_string() const;
+    std::string stderr_string() const;
+
+    PythonStreamRedirect & set_enabled(bool enabled)
+    {
+        m_enabled = enabled;
+        return *this;
+    }
+
+    bool is_enabled() const { return m_enabled; }
```
tigercosmos, line 527: it looks redundant to have is_enabled and is_disable?

yungyuc: It is intentional.  Having the symmetric helpers encapsulates better.  When the system grows we will forget how it is implemented inside and wonder it's `is_enabled()` or `is_disabled()`.  Implementing both resolves the ambiguity.

An IDE with auto-completion helps, but we cannot assume everyone uses an IDE.

PR #160, opened 2022-11-28
fix backgrond color and command edit text color

j8xixo12:
Fix command windows background color to white also fix command edit text color to black.


yungyuc (approved):
LGTM

PR #164, opened 2022-12-07
Fix a bug of wrong dtype of the numpy.ndarray created from ConcreteBuffer on aarch64

yungyuc:
On aarch64 (arm64 on Linux), `char` is unsigned but it is signed on other platform.  To avoid the ambiguity the `format_descriptor` to construct the buffer protocol object is changed to use `int8_t` (signed) type.

### DIFF
--- a/cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp
@@ -81,7 +81,7 @@ WrapConcreteBuffer::WrapConcreteBuffer(pybind11::module & mod, char const * pyna
                 return py::buffer_info(
                     self.data(), /* Pointer to buffer */
                     sizeof(int8_t), /* Size of one scalar */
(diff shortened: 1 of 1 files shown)


tai271828 (commented):
How do I verify whether this pull request works or not?  If compilation passes on aarch64, does it mean the fix works?

yungyuc:
> How do I verify whether this pull request works or not? If compilation passes on aarch64, does it mean the fix works?

In this case, I think it is good as long as it looks good to you.  I have tested it on my aarch64 VM (where I found this issue).

tai271828 (approved):
The change looks good to me from the portability point of view since:
1. it is suggested by arm development guide. For example, see [this example](https://developer.arm.com/documentation/den0013/d/Porting/Miscellaneous-C-porting-issues/unsigned-char-and-signed-char) `In fact, on an ARM core, it is usually better to use ints rather than chars, even for small values, for performance reasons.` from the guide.
2. `int8_t` is a general type, and should be available to any compiler and standard.

PR #167, opened 2022-12-08
Fix qt geometry render issue by modified vertex size

j8xixo12:
Fix render can not show `Line` by modified vertex size from **3** to **5**
but still no idea why vertex size have to `>= 5`

ref: [Qt official docs](https://doc.qt.io/qt-6/qt3dcore-qattribute.html#vertexSize-prop)


yungyuc (approved):
Sweet!

Issue #133, opened 2022-10-20
Report the stack trace when Python exception is thrown from viewer


yungyuc:
Some Python exceptions are not properly handled when they are thrown from the viewer, e.g., the stack trace shown in #132:

```
 	KernelBase.dll!00007ff9cb82428c()	Unknown
 	vcruntime140d.dll!00007ff901cab650()	Unknown
>	viewer.exe!pybind11::module_::import(const char * name) Line 1199	C++
 	viewer.exe!modmesh::RApplication::{ctor}::__l2::<lambda_3>::operator()() Line 96	C++
 	[External Code]	
```

I guess it's not the only place to have the issue.

tychuang1211:
Just leaving reference.
[handling-python-exceptions-cpp](https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-python-exceptions-cpp)

tychuang1211:
I think this issue is solved in #168

yungyuc:
> I think this issue is solved in #168

Thanks.  Let's wait for the feedback from other crashing issues including #115, #132, #169.

yungyuc:
> > I think this issue is solved in #168
> 
> Thanks. Let's wait for the feedback from other crashing issues including #115, #132, #169.

The three issues seem to be more involved than what we solve here.  Close this issue.

PR #168, opened 2022-12-09
Output stack trace to std::cerr when Python exception is thrown from viewer

tychuang1211:
#133 

For example:
image

### DIFF
--- a/cpp/modmesh/python/common.cpp
+++ b/cpp/modmesh/python/common.cpp
@@ -163,7 +163,7 @@ void Interpreter::preload_module(std::string const & name)
     }
     catch (const py::error_already_set & e)
     {
-        if (std::string::npos == std::string(e.what()).find("ModuleNotFoundError"))
+        if (e.matches(PyExc_ModuleNotFoundError))
         {
             throw;
         }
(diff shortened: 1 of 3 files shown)


cpp/modmesh/python/common.cpp
```diff
@@ -105,26 +105,49 @@ Interpreter & Interpreter::setup_modmesh_path()
...
+    }
     return *this;
 }
 
 // NOLINTNEXTLINE(readability-convert-member-functions-to-static)
 int Interpreter::enter_main()
 {
-    // NOLINTNEXTLINE(misc-const-correctness)
-    pybind11::object mod_sys = pybind11::module_::import("modmesh.system");
+    int ret = -1;
```
yungyuc, line 138: Thanks for using a default return value.  It improves readability.

cpp/modmesh/python/common.cpp
```diff
@@ -140,7 +163,7 @@ void Interpreter::preload_module(std::string const & name)
     }
     catch (const py::error_already_set & e)
     {
-        if (std::string::npos == std::string(e.what()).find("ModuleNotFoundError"))
+        if (e.matches(PyExc_ModuleNotFoundError))
```
yungyuc, line 166: Sweet!

yungyuc (approved):
LGTM

PR #177, opened 2022-12-10
build(Makefile): make python3-config protable

tai271828:
When using python virtual environment, python-config or python3-config are maybe not included in PATH, and thus the command may return incorrect information of python configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,11 @@ CMAKE_ARGS ?=
 VERBOSE ?=
 FORCE_CLANG_FORMAT ?=
 
-pyextsuffix := $(shell python3-config --extension-suffix)
+WHICH_PYTHON := $(shell which python)
+REALPATH_PYTHON := $(shell realpath $(WHICH_PYTHON))
+DIRNAME_PYTHON := $(shell dirname $(REALPATH_PYTHON))
+
+pyextsuffix := $(shell $(DIRNAME_PYTHON)/python3-config --extension-suffix)
 pyvminor := $(shell python3 -c 'import sys; print("%d%d" % sys.version_info[0:2])')
 
 ifeq ($(CMAKE_BUILD_TYPE), Debug)
--- a/Makefile
+++ b/Makefile
@@ -38,8 +38,8 @@ VERBOSE ?=
 FORCE_CLANG_FORMAT ?=
 
 WHICH_PYTHON := $(shell which python3)
-REALPATH_PYTHON := $(shell realpath $(WHICH_PYTHON))
-DIRNAME_PYTHON := $(shell dirname $(REALPATH_PYTHON))
+REALPATH_PYTHON := $(realpath $(WHICH_PYTHON))
+DIRNAME_PYTHON := $(dir dirname $(REALPATH_PYTHON))
 
 pyextsuffix := $(shell $(DIRNAME_PYTHON)/python3-config --extension-suffix)
 pyvminor := $(shell python3 -c 'import sys; print("%d%d" % sys.version_info[0:2])')
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ CMAKE_ARGS ?=
 VERBOSE ?=
 FORCE_CLANG_FORMAT ?=
 
-WHICH_PYTHON := $(shell which python)
+WHICH_PYTHON := $(shell which python3)
 REALPATH_PYTHON := $(shell realpath $(WHICH_PYTHON))
 DIRNAME_PYTHON := $(shell dirname $(REALPATH_PYTHON))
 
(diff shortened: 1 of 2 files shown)


cpp/binary/pymod_modmesh/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ if (CLANG_TIDY_EXE AND USE_CLANG_TIDY)
 endif ()
 
 execute_process(
-    COMMAND python3-config --extension-suffix
+    COMMAND $(dirname $(realpath $(which python)))/python3-config --extension-suffix
```
yungyuc: I'd be curious whether or not windows accepts it.

Makefile
```diff
@@ -37,7 +37,11 @@ CMAKE_ARGS ?=
 VERBOSE ?=
 FORCE_CLANG_FORMAT ?=
 
-pyextsuffix := $(shell python3-config --extension-suffix)
+WHICH_PYTHON := $(shell which python)
```
yungyuc: It mush be `python3`.  We should not allow any possibility of running python 2.

yungyuc (changes requested):

Makefile
```diff
@@ -37,7 +37,11 @@ CMAKE_ARGS ?=
...
+REALPATH_PYTHON := $(shell realpath $(WHICH_PYTHON))
```
yungyuc: Please use make function while we can: https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html .  It runs faster than shell.

yungyuc (changes requested):

Makefile
```diff
@@ -37,7 +37,11 @@ CMAKE_ARGS ?=
...
+DIRNAME_PYTHON := $(shell dirname $(REALPATH_PYTHON))
+
+pyextsuffix := $(shell $(DIRNAME_PYTHON)/python3-config --extension-suffix)
```
tai271828, line 52: FWIW, this will bring the build flow the support of using python virtual environment.

For example, in my case it is:
- `which python`: /home/tai271828/.cache/pypoetry/virtualenvs/modmesh-venv-py3-9-13-LgJ7FNfF-py3.9/bin/python
- `realpath`: /home/tai271828/.pyenv/versions/3.9.13/bin/python3.9
- `dirname`: /home/tai271828/.pyenv/versions/3.9.13/bin/

and additionally:
```
$ ls -alh /home/tai271828/.cache/pypoetry/virtualenvs/modmesh-venv-py3-9-13-LgJ7FNfF-py3.9/bin/python
lrwxrwxrwx 1 tai271828 tai271828 49 Dec  9 21:46 /home/tai271828/.cache/pypoetry/virtualenvs/modmesh-venv-py3-9-13-LgJ7FNfF-py3.9/bin/python -> /home/tai271828/.pyenv/versions/3.9.13/bin/python
```

Makefile
tai271828: This is fair. I will update it.

Makefile
yungyuc: Please summarize in a comment line that the treatment is for virtual env.  It should also be mentioned that a virtual env is strongly discouraged in developing modmesh.  We do not actively resolve bugs related to any virtual env including venv or conda.

cpp/binary/pymod_modmesh/CMakeLists.txt
tai271828: I am wondering too. I am thinking if you know someone interested in testing this?

yungyuc:
I would encourage you to make sure CI passes in your fork before filing a PR.

cpp/binary/pymod_modmesh/CMakeLists.txt
yungyuc: It actually passes CI ... well, interesting.

But it does not hurt to use cmake native helpers: https://cmake.org/cmake/help/v3.0/command/get_filename_component.html

tai271828:
> I would encourage you to make sure CI passes in your fork before filing a PR.

This is a good idea. I will do that next time.

Makefile
tai271828: Will do!

Makefile
tai271828: Let me check. Thanks for the tip!

cpp/binary/pymod_modmesh/CMakeLists.txt
tai271828: Thanks for the tip. Let me check.

tai271828:
Updated the pull request according to the above comments and suggestions. Ready to review again. Thank you!

yungyuc (approved):
LGTM.  Thanks a lot.

PR #180, opened 2022-12-12
Fix execvp permission denied error

yungyuc:
There is a typo in the Makefile causing the following error:

```
$ make invalid
make: execvp: ./: Permission denied
make: *** No rule to make target `invalid'.  Stop.
```


tai271828 (approved):
Thanks! I also noticed that but I was not aware that it was my bad!

PR #181, opened 2022-12-13
Add a sample Makefile to extract buffer and make a standalone build

yungyuc:
The new contrib/standalone_buffer/Makefile extracts the buffer subsystem and build it standalone.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -8,6 +8,69 @@ on:
 
 jobs:
 
+  standalone_buffer:
+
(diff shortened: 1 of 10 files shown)


.github/workflows/devbuild.yml
```diff
@@ -8,6 +8,69 @@ on:
 
 jobs:
 
+  standalone_buffer:
+
+    if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
+
+    name: standalone_buffer_${{ matrix.os }}
```
yungyuc, line 15: Make sure it is always possible to build the buffer as a header-only sub-system standalone.

Issue #179, opened 2022-12-11
CMake Error with the latest codebase which introduces gtest


tai271828:
I have no luck when building the latest codebase (commit 0a1f53841c0fc3c7b13066a19b2f05be9d5aac7a ) with my branch ( see PR https://github.com/solvcon/modmesh/pull/177 ) . I guess it may be my configuration since the CI of Ubuntu passed. I will check in more details later to see what's going on.

Environment:
- Ubuntu mate 20.04
- Python 3.9 venv

Error message:

```
(modmesh-venv-py3-9-13-py3.9) [O_o]─[~/work-my-projects/solvcon/modmesh] tai271828@syakaro: 221211-23:51:25 15 file 104Kb
$ ls; make clean; rm -rf build/; rm -f _modmesh.cpython-3*-x86_64-linux-gnu.so; make buildext VERBOSE=1 BUILD_QT=ON CMAKE_BUILD_TYPE=Debug CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)" | tee /tmp/build.log ; ls
build  cmake  CMakeLists.txt  contrib  COPYING  cpp  gtests  Makefile  modmesh  README.md  resources  setup.py  tai.py  tests  thirdparty
make: execvp: ./: Permission denied
rm -f /home/tai271828/work-my-projects/solvcon/modmesh/modmesh/_modmesh
make -C build/dev39 clean
make[1]: *** build/dev39: No such file or directory.  Stop.
make: *** [Makefile:156: clean] Error 2
make: execvp: ./: Permission denied
mkdir -p build/dbg39 ; \
cd build/dbg39 ; \
[...]

Issue #182, opened 2022-12-20
CI for windows began to fail


tai271828:
I noticed CI began to fail since yesterday.

- https://github.com/solvcon/modmesh/actions/runs/3725891142
- https://github.com/solvcon/modmesh/actions/runs/3733775280

tai271828:
btw, the CI for my personal fork failed since 3 days ago https://github.com/tai271828/modmesh/actions/runs/3720939984

yungyuc:
It's being discussed at https://discord.com/channels/730297880140578906/730393342818648204/1054553931776278618

PR #183, opened 2022-12-20
Correct cmake support version of gtests

tai271828:
In #179 , we noticed the google tests introduced into modmesh recently
may need newer CMake. According to CMake document,
DOWNLOAD_EXTRACT_TIMESTAMP is supported since CMake version 3.24. See
 https://cmake.org/cmake/help/latest/module/ExternalProject.html

 Issue: #179


tai271828:
Note the Ci fails for this known issue https://github.com/solvcon/modmesh/issues/182 but not from my PR.

yungyuc:
LGTM.

PR #187, opened 2022-12-22
WIP: implement configuration

tigercosmos:
The configuration is accessible for both C++ (`Config`) and Python (`modmesh.config`) . (#165)
The data of configuration is decided to use `py::dict` to leverage Python `json` module and to avoid third-party libraries in C++
Now we don't use `modmesh.view.app.pycon.python_redirect = False`, instead, we will just set `modmesh.config["python_redirect"] = True`

---

The config file will look like:
```json
{
    "A": 1,
    "B": "B",
    "C": {
        "D": 1.2
    }
}
```

C++ access:
```c++
    std::cout << "A: " << pybind11::cast<int>(config.get("A")) << std::endl; // A: 1
    config.set("A", 5);

    std::cout << "A: " << pybind11::cast<int>(config.get("A")) << std::endl; // A: 5
    std::cout << "B: " << pybind11::cast<std::string>(config.get("B")) << std::endl; // B: B
 
    // TODO: support a list of keys
    std::cout << "D: " << pybind11::cast<std::string>(config.get({"C", "D"}, "B")) << std::endl; // D: 1.2
```

Python access, for both interpreter mode or viewer mode:
```python
import modmesh.config as config
a = config["A"]
config["A"] = 5
```

### DIFF
--- a/cpp/binary/viewer/CMakeLists.txt
+++ b/cpp/binary/viewer/CMakeLists.txt
@@ -97,6 +97,8 @@ endif()
 
 add_custom_target(run_viewer_pytest $<TARGET_FILE:viewer> --mode=pytest)
 
+file(COPY ${PROJECT_ROOT_DIR}/config.json DESTINATION ${viewer_BINARY_DIR})
+
 install(TARGETS viewer
     RUNTIME DESTINATION "${INSTALL_VIEWERDIR}"
     BUNDLE DESTINATION "${INSTALL_VIEWERDIR}"
--- a/cpp/modmesh/python/common.hpp
+++ b/cpp/modmesh/python/common.hpp
@@ -545,6 +545,49 @@ private:
 #ifdef __GNUC__
 #pragma GCC diagnostic pop
 #endif
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+// Suppress the warning "greater visibility than the type of its field"
+#pragma GCC diagnostic ignored "-Wattributes"
+#endif
+class Config
+{
+public:
+
+    static Config & instance();
+
+    Config(Config const &) = delete;
+    Config(Config &&) = delete;
+    Config & operator=(Config const &) = delete;
+    Config & operator=(Config &&) = delete;
+    ~Config() { delete m_data; };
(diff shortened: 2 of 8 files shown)


cpp/modmesh/python/module.cpp
```diff
@@ -52,6 +53,7 @@ void initialize(pybind11::module_ mod)
 #else // QT_CORE_LIB
     mod.attr("HAS_VIEW") = false;
 #endif // QT_CORE_LIB
+    wrap_Config(mod);
```
tigercosmos, line 56: I suppose this line will make `modmesh.config` work, but I don't know why it doesn't work.

cpp/binary/viewer/CMakeLists.txt
```diff
@@ -97,6 +97,8 @@ endif()
 
 add_custom_target(run_viewer_pytest $<TARGET_FILE:viewer> --mode=pytest)
 
+file(COPY ${PROJECT_ROOT_DIR}/config.json DESTINATION ${viewer_BINARY_DIR})
```
tigercosmos, line 100: We will need to provide a default setting `config.json` later

cpp/modmesh/python/common.hpp
```diff
@@ -545,6 +545,49 @@ class PythonStreamRedirect
...
+    // In Python, "number" is analogous to the float type.
+    void set(const char * key, const float value);
+    // void set(const std::vector<const char *> keys, const float value); // TODO
+
+    void set(const char * key, const char * value);
+    // void set(const std::vector<const char *> keys, const char * value); // TODO
+
+private:
+    Config() = default;
+    // need to be initialized after `Interpreter` is initialized
```
tigercosmos, line 584: another way to avoid the Interpreter initialization issue is to use `std::unordered_map` directly, the drawback is that a nested dict is not allowed.

yungyuc:
@tigercosmos , is this PR ready for review?  CI is not clean yet.  Please change the PR to draft if it is not ready.  If you are reqeusting review, please make it clear what are your points.

Issue #135, opened 2022-10-23
Add profiling system for the shock tube problem


yungyuc:
The shock tube problem (`modmesh.onedim.euler1d.ShockTube`) is a classical compressible flow problem.  It uses a numerical solver along with an analytical solution.

The numerical solver is supposed to be high-performance but it is not yet optimized.  Before optimization, profiling should be done.  We do not yet have a useable profiling system in modmesh, but there is a prototype: https://github.com/solvcon/modmesh/blob/master/cpp/modmesh/profile.hpp .

This issue is to track the augmentation of the prototype to a useable profiling system.  It does not need to be comprehensive, but should help understand the performance of the shock tube problem.

yungyuc:
To have a basic understandings of the shock tube problem, the notes in https://github.com/solvcon/mmnote/ may be helpful.

tai271828:
I will help profiling on Ubuntu Linux.

tai271828:
Since pull request #190 is landed, I think my next step will try to make the profiling data able to be exported to a file, which could be consumed by a GUI-based reporter, dashboard, or something.

easy_profiler mentioned in [this thread](https://github.com/solvcon/modmesh/pull/188#discussion_r1056893194) has a nice GUI application to represent its profiling data. The application is written in Qt5, so I believe we can be enlightened a lot by its code when writing a Qt6-based dashboard for our own profiling data.

tai271828:
> Since pull request #190 is landed, I think my next step will try to make the profiling data able to be exported to a file, which could be consumed by a GUI-based reporter, dashboard, or something.
> 
> easy_profiler mentioned in [this thread](https://github.com/solvcon/modmesh/pull/188#discussion_r1056893194) has a nice GUI application to represent its profiling data. The application is written in Qt5, so I believe we can be enlightened a lot by its code when writing a Qt6-based dashboard for our own profiling data.

Besides, if the data could be consumed by the other application, that means it will be time for us to put profiling benchmark in our CI.

yungyuc:
Yes, I think it's a great idea to steal profling visualization code from easy profiler, and probably also the profile data file format.

Profiling in CI is tricky because the CI servers usually are noisy.  We should bring it to discord and use more discussions for it.

yungyuc:
Clear assignees for lack of activities.

tigercosmos:
[...]

PR #188, opened 2022-12-24
profiling: make a draft for discussion

tai271828:
An easy script to show what I know after reading the source of modmesh about how to use the following items:
- manipulate the euler1d solver of modmesh and create a 1D shock tube problem based on the solver.
- profiling the problem by using profile.hpp. That is to say, a stop watch in this code change.

I am looking forward to more discussion or suggestions that help us make #135 move forward and keep us on the same page. For example, is this what you expect about the approach of profiling?

### DIFF
--- a/contrib/profiling.py
+++ b/contrib/profiling.py
@@ -1,0 +1,29 @@
+from modmesh.onedim import euler1d
+from _modmesh import stop_watch
+
+
+shock_tube = euler1d.ShockTube()
+shock_tube.build_constant(
+    gamma=1.4,
+    pressure1=1.0,
+    density1=1.0,
+    pressure5=0.1,
+    density5=0.125,
+)
+
+shock_tube.build_numerical(xmin=-1, xmax=1, ncoord=21, time_increment=0.05)
+
+# reset lap
+stop_watch.lap()
+shock_tube.calc_velocity2(0, 10)
+print(f"calc_velocity2: \t{stop_watch.lap()}")
+
+# reset lap
+stop_watch.lap()
+shock_tube.build_field(t=10)
+print(f"build_field t=10:\t{stop_watch.lap()}")
+
+# reset lap
+stop_watch.lap()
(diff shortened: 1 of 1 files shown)


contrib/profiling.py
```diff
@@ -0,0 +1,29 @@
...
+    gamma=1.4,
+    pressure1=1.0,
+    density1=1.0,
+    pressure5=0.1,
+    density5=0.125,
+)
+
+shock_tube.build_numerical(xmin=-1, xmax=1, ncoord=21, time_increment=0.05)
+
+# reset lap
```
yungyuc, line 16: Is there a report or something for the test case?

tai271828: No. We can (and should) implement that for sure. 

Do we want to add profiling a part of test case? If yes, then we may want to have a benchmark that is agreed on mutually.

yungyuc: Before making automatic tests for runtime, perhaps we can first work on the (automatic generation of the) profiling reports.  The reports are what we want from a profiler.  One example of the reporting is https://github.com/yse/easy_profiler .

However here I think we should first develop a [scoped-based profiler](https://yyc.solvcon.net/en/latest/nsd/11arraydesign/arraydesign.html?highlight=scope#scoped-based-timer), which is usually ad hoc and does not have a lot of simple examples online.  We may use the examples of the sample-based profiling, but start with a scope-based one.

tai271828: Thanks for the comments. The roadmap sounds good to me. Let me have a look of  https://github.com/yse/easy_profiler

tai271828:
Close this PR since we now have this driving script from gist [euler1d_profiling-py](https://gist.github.com/tai271828/d5f50e4a06274b1203ed79f6739039dd#file-euler1d_profiling-py).

This driving script runs through a real shock tube problem and the result is easier to ready.

PR #189, opened 2023-01-06
perf(profiler): integrate easy_profiler

tai271828:
Begin to integrate easy_profiler via CMake for march_half_so0

Issue: #135

### DIFF
--- a/cpp/binary/pymod_modmesh/CMakeLists.txt
+++ b/cpp/binary/pymod_modmesh/CMakeLists.txt
@@ -9,10 +9,15 @@ pybind11_add_module(
     module.cpp
 )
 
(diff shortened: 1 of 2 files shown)


yungyuc:
I am not sure why do you try to integrate easy profiler.  I do not plan to use it.

tai271828:
> I am not sure why do you try to integrate easy profiler. I do not plan to use it.

Got it. To be clear: is your plan making profiler on our own?

yungyuc:
> > I am not sure why do you try to integrate easy profiler. I do not plan to use it.
> 
> 
> 
> Got it. To be clear: is your plan making profiler on our own?

Yes, I would like to write code as much as we can.

PR #197, opened 2023-02-17
Shocktube add legend

j8xixo12:
I added legend and xy axis labels to improve readability of shock tube plot.

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -79,9 +79,11 @@ class Plot:
         self.ax.set_ylim(-0.2, 1.2)
         self.ax.grid()
 
-        self.density = QuantityLine(name="density", unit="kg/m^3")
-        self.velocity = QuantityLine(name="velocity", unit="m/s")
-        self.pressure = QuantityLine(name="pressure", unit="pa")
+        self.density = QuantityLine(name="density",
+                                    unit=r"$\mathrm{kg}/\mathrm{m}^3$")
+        self.velocity = QuantityLine(name="velocity",
+                                     unit=r"$\mathrm{m}/\mathrm{s}$")
+        self.pressure = QuantityLine(name="pressure", unit=r"$\mathrm{Pa}$")
 
     def build_lines(self, x):
         self.ax.set_xlim(x[0], x[-1])
@@ -91,6 +93,8 @@ class Plot:
                 (self.velocity, 'g'),
                 (self.pressure, 'b'),
         )):
+            # Plot multiple data line with same X axis, it need to plot a
+            # data line on main axis first
             if i == 0:
                 data.ana, = self.ax.plot(x.copy(), np.zeros_like(x),
                                          f'{color}-',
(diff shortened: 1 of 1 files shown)


modmesh/app/euler1d.py
```diff
@@ -78,19 +79,43 @@ def __init__(self, figsize=(15, 10)):
         self.ax.set_ylim(-0.2, 1.2)
         self.ax.grid()
 
-        self.density = QuantityLine(name="density")
-        self.velocity = QuantityLine(name="velocity")
-        self.pressure = QuantityLine(name="pressure")
+        self.density = QuantityLine(name="density", unit="kg/m^3")
```
yungyuc: Please use latex math: `$\mathrm{kg}/\mathrm{m}^3$`

modmesh/app/euler1d.py
```diff
@@ -78,19 +79,43 @@ def __init__(self, figsize=(15, 10)):
...
+        self.velocity = QuantityLine(name="velocity", unit="m/s")
```
yungyuc: Ditto.  $\mathrm{m}/\mathrm{s}$

modmesh/app/euler1d.py
```diff
@@ -78,19 +79,43 @@ def __init__(self, figsize=(15, 10)):
...
+        self.pressure = QuantityLine(name="pressure", unit="pa")
```
yungyuc: Ditto.  $\mathrm{Pa}$

modmesh/app/euler1d.py
```diff
@@ -78,19 +79,43 @@ def __init__(self, figsize=(15, 10)):
...
+        lines = []
+        for i, (data, color) in enumerate((
                 (self.density, 'r'),
                 (self.velocity, 'g'),
                 (self.pressure, 'b'),
-        ):
-            data.ana, = self.ax.plot(x.copy(), np.zeros_like(x), f'{color}-')
-            data.num, = self.ax.plot(x.copy(), np.zeros_like(x), f'{color}x')
+        )):
+            if i == 0:
```
yungyuc, line 98: It's more readable to check for `isinstance(data, self.density)`.  In that way you can freely change the order of the data lines in the future.

yungyuc (changes requested):

j8xixo12:
@yungyuc I replaced unit raw string with `latex`.

modmesh/app/euler1d.py
j8xixo12: Checking `isinstance(data, self.density)` it will force `density` plot on main axis, therefore I can not freely change the order of the data lines.
Due to plot multiple data line on same x axis, it need plot a data line on main axis, therefore I check for loop index `I == 0` to plot data line on main axis, others plot on auxiliary axis.

yungyuc: So you want the first data but rather than density to be plot as the main line.  That makes sense.

yungyuc (approved):

yungyuc:
Let's fix the mac CI failure later.

PR #198, opened 2023-02-28
improve CMake

tigercosmos:
- let`USE_GOOGLETEST` be `option`
- let `modmesh_primary` be a shared library


tigercosmos:
I think I need more investments. close this one

PR #204, opened 2023-04-02
Avoid Qt API qputenv which is hard for refactoring to non-Qt code

yungyuc:
Use POSIX setenv instead.  The POSIX setenv is not thread-safe.  Add a TODO for thread safety in the future.

### DIFF
--- a/cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
+++ b/cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
@@ -151,6 +151,12 @@ WrapProcessInfo::WrapProcessInfo(pybind11::module & mod, char const * pyname, ch
             [](py::object const &) -> auto & {
                 return wrapped_type::instance();
             })
+        .def(
+            "set_environment_variables",
+            [](wrapped_type & self) -> auto & {
+                return self.set_environment_variables();
+            },
+            py::return_value_policy::reference_internal)
         .def_property_readonly(
             "command_line",
             [](wrapped_type & self) -> auto & {
--- a/cpp/modmesh/toggle/toggle.cpp
+++ b/cpp/modmesh/toggle/toggle.cpp
@@ -34,6 +34,26 @@
 namespace modmesh
 {
 
+int setenv(const char * name, const char * value, int overwrite)
+{
+#ifdef _WIN32
+    int errcode = 0;
+    if (!overwrite)
+    {
+        size_t envsize = 0;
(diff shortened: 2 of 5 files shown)


cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -151,6 +151,12 @@ WrapProcessInfo::WrapProcessInfo(pybind11::module & mod, char const * pyname, ch
             [](py::object const &) -> auto & {
                 return wrapped_type::instance();
             })
+        .def(
```
yungyuc, line 154: I did not add a unit test because once set it cannot be redone in the same process.  Testing it requires a clever way that I haven't thought of.

cpp/modmesh/toggle/toggle.cpp
```diff
@@ -34,6 +34,26 @@
 namespace modmesh
 {
 
+int setenv(const char * name, const char * value, int overwrite)
+{
+#ifdef _WIN32
```
yungyuc, line 39: Windows does not have `::setenv()`.

cpp/modmesh/toggle/toggle.cpp
```diff
@@ -34,6 +34,26 @@
...
+        size_t envsize = 0;
+        errcode = getenv_s(&envsize, NULL, 0, name);
+        if (errcode || envsize)
+        {
+            return errcode;
+        }
+    }
+    return _putenv_s(name, value);
+#else // _WIN32
+    // NOLINTNEXTLINE(concurrency-mt-unsafe) TODO: make the wrapper thread safe.
```
yungyuc, line 52: Another benefit of the custom wrapper to `setenv()`: we can make it thread safe in the future.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -497,14 +497,6 @@ OneTimeInitializer<view_pymod_tag> & OneTimeInitializer<view_pymod_tag>::me()
 
 void initialize_view(pybind11::module & mod)
 {
-// TODO: There is some bug in QT RHI, this is a workaround to force QT use opengl directly
-// see: https://doc.qt.io/qtforpython/overviews/qt3drender-porting-to-rhi.html
-// for more detail
-#if defined(QT3D_USE_RHI)
-    qputenv("QT3D_RENDERER", "rhi");
```
yungyuc, line 504: Avoid `qputenv` that is only available in Qt

PR #206, opened 2023-04-04
Add dynamic feature toggle

yungyuc:
Ref issue #84 

Add a class `modmesh::DynamicToggleTable` to store key-value pairs whose number can change during runtime. It accepts limited types: Boolean, (signed) int8,
16, 32, 64, (double) real, and string.

The Pythonic interface `__getattr__` and `__setattr__` are provided.  In Python, addition of a key-value pair should be done using the `set_TYPE()` function.  The
Pythonic `__getattr__` should get any type.

### DIFF
--- a/cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
+++ b/cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
@@ -56,6 +56,79 @@ protected:
 
     static std::string report(wrapped_type const & self);
 
+    static pybind11::object getattr(wrapped_type const & self, std::string const & key)
+    {
+        namespace py = pybind11;
+
+        DynamicToggleIndex const index = self.get_dynamic_index(key);
+        switch (index.type)
+        {
+        case DynamicToggleIndex::TYPE_NONE:
+            throw py::attribute_error(Formatter() << "Cannt get non-existing key \"" << key << "\"");
+            break;
+        case DynamicToggleIndex::TYPE_BOOL:
+            return py::cast(self.get_bool(key));
+            break;
+        case DynamicToggleIndex::TYPE_INT8:
+            return py::cast(self.get_int8(key));
+            break;
+        case DynamicToggleIndex::TYPE_INT16:
+            return py::cast(self.get_int16(key));
+            break;
+        case DynamicToggleIndex::TYPE_INT32:
+            return py::cast(self.get_int32(key));
+            break;
+        case DynamicToggleIndex::TYPE_INT64:
+            return py::cast(self.get_int64(key));
+            break;
+        case DynamicToggleIndex::TYPE_REAL:
+            return py::cast(self.get_real(key));
+            break;
+        case DynamicToggleIndex::TYPE_STRING:
+            return py::cast(self.get_string(key));
+            break;
+        default:
+            return py::none();
+            break;
+        }
+    }
+
+    static void setattr(wrapped_type & self, std::string const & key, pybind11::object & value)
+    {
+        namespace py = pybind11;
+
+        DynamicToggleIndex const index = self.get_dynamic_index(key);
+        switch (index.type)
+        {
+        case DynamicToggleIndex::TYPE_NONE:
+            throw pybind11::attribute_error(Formatter() << "Cannot set non-existing key \"" << key << "\"; use set_TYPE() instead");
+            break;
+        case DynamicToggleIndex::TYPE_BOOL:
+            self.set_bool(key, py::cast<bool>(value));
+            break;
+        case DynamicToggleIndex::TYPE_INT8:
+            self.set_int8(key, py::cast<int8_t>(value));
+            break;
+        case DynamicToggleIndex::TYPE_INT16:
+            self.set_int16(key, py::cast<int16_t>(value));
+            break;
+        case DynamicToggleIndex::TYPE_INT32:
+            self.set_int32(key, py::cast<int32_t>(value));
+            break;
+        case DynamicToggleIndex::TYPE_INT64:
+            self.set_int64(key, py::cast<int64_t>(value));
+            break;
+        case DynamicToggleIndex::TYPE_REAL:
+            self.set_real(key, py::cast<double>(value));
+            break;
+        case DynamicToggleIndex::TYPE_STRING:
+            self.set_string(key, py::cast<std::string>(value));
+            break;
+        default:
+            break;
+        }
+    }
+
 }; /* end class WrapToggle */
 
 WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const * pydoc)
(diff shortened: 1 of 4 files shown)


cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -56,6 +56,85 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapToggle
 
     static std::string report(wrapped_type const & self);
 
+    static pybind11::object getattr(wrapped_type const & self, std::string const & key)
```
yungyuc, line 59: This is the Pythonic getter.

cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -56,6 +56,85 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapToggle
...
+        case DynamicToggleIndex::TYPE_STRING:
+            return py::cast(self.get_string(key));
+            break;
+        default:
+            return py::none();
+            break;
+        }
+    }
+
+    static void setattr(wrapped_type & self, std::string const & key, pybind11::object & value)
```
yungyuc, line 96: The Pythonic setter.

cpp/modmesh/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -68,6 +147,30 @@ WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const *
...
         ;
 
+    // Dynamic properties.  Number of the properties can be freely changed
+    // during runtime.
+    (*this)
+        .def("__getattr__", getattr)
+        .def("__setattr__", setattr)
+        .def("dynamic_keys", &wrapped_type::dynamic_keys)
+        .def("dynamic_clear", &wrapped_type::dynamic_clear)
+        .def("get_bool", &wrapped_type::get_bool, py::arg("key"))
```
yungyuc, line 157: The repetition is tedious, but tolerable at the moment.  When more types are added to the support we may consider to make a helper to create them.

cpp/modmesh/toggle/toggle.cpp
```diff
@@ -60,6 +63,90 @@ Toggle & Toggle::instance()
     return o;
 }
 
+// NOLINTNEXTLINE(fuchsia-statically-constructed-objects,readability-redundant-string-init,cert-err58-cpp)
+std::string const DynamicToggleTable::sentinel_string = "";
+
+/* The macro gives debuggers a hard time. Manually expand it if you need to
```
yungyuc, line 69: I didn't not find a way to use template to reduce the duplicated code.  The macros are the best idea I came up with.  The downside is the difficulty in debug-stepping.  Since we can manually  expand it when debugging, I trade the capability of stepping for the conciseness.

cpp/modmesh/toggle/toggle.cpp
```diff
@@ -60,6 +63,90 @@ Toggle & Toggle::instance()
...
+        if (it != m_key2index.end())                                     \
+        {                                                                \
+            if (it->second.is_##MTYPE())                                 \
+            {                                                            \
+                return m_vector_##MTYPE.at(it->second.index);            \
+            }                                                            \
+        }                                                                \
+        return SENTINEL;                                                 \
+    }
+MM_DECL_DYNGET(bool, bool, 0)
```
yungyuc: The sentinel should have been `false`.  I overlooked this and will fix it in a new commit.

cpp/modmesh/toggle/toggle.hpp
```diff
@@ -34,26 +34,102 @@
...
+        TYPE_BOOL,
+        TYPE_INT8,
+        TYPE_INT16,
+        TYPE_INT32,
+        TYPE_INT64,
+        TYPE_REAL,
+        TYPE_STRING
+    };
+
+    operator bool() const { return type != TYPE_NONE; }
```
yungyuc, line 62: If no type is set, this index is invalid and evaluated as false.

cpp/modmesh/toggle/toggle.hpp
```diff
@@ -34,26 +34,102 @@
...
+    bool is_int16() const { return type == TYPE_INT16; }
+    bool is_int32() const { return type == TYPE_INT32; }
+    bool is_int64() const { return type == TYPE_INT64; }
+    bool is_real() const { return type == TYPE_REAL; }
+    bool is_string() const { return type == TYPE_STRING; }
+
+    // Index upper bound 2**32 is more than sufficient.  2**16 (65536) may be
+    // too little.
+    uint32_t index = 0;
+    Type type = TYPE_NONE;
```
yungyuc, line 74: Use a single byte after the 32-bit index.

cpp/modmesh/toggle/toggle.hpp
```diff
@@ -34,26 +34,102 @@
...
 
-    static constexpr bool USE_PYSIDE =
 #ifdef MODMESH_USE_PYSIDE
-        true
-#else // MODMESH_USE_PYSIDE
-        false
-#endif // MODMESH_USE_PYSIDE
-        ;
-
+    MM_TOGGLE_CONSTEXPR_BOOL(USE_PYSIDE, true);
```
yungyuc, line 129: Refactoring change.  I cannot get rid of the `#ifdef` here because the conditional macro is not supported within `#define`.

tests/test_toggle.py
```diff
@@ -45,6 +46,154 @@ def test_instance(self):
         self.assertTrue(hasattr(modmesh.Toggle.instance, "show_axis"))
 
 
+class ToggleDynamicTC(unittest.TestCase):
+
+    def test_all_types(self):
```
yungyuc, line 51: Explicit tests for all supported type using the `get_TYPE` and `set_TYPE` functions.

tests/test_toggle.py
```diff
@@ -45,6 +46,154 @@ def test_instance(self):
...
+        tg = modmesh.Toggle.instance
+        tg.dynamic_clear()
```
yungyuc, line 53: Clear the container before testing.

tests/test_toggle.py
```diff
@@ -45,6 +46,154 @@ def test_instance(self):
...
+        # Raise exception when the requested key is not available (no need to
+        # test for all types).
+        with self.assertRaisesRegex(
+                AttributeError,
+                r'Cannt get non-existing key "dunder_nonexist"'
+        ):
+            tg.dunder_nonexist
+
+        # Need to use set_TYPE() to create the dynamic key-value pair.
+        # (Make sure all supported types are tested.)
```
yungyuc, line 167: Only `set_TYPE` can create new key-value pairs.  Test all types for the new keys can be got by `__getattr__`.

yungyuc (commented):
So far I am only trying to make the (global) dynamic toggle container to support `__getattr__` and `__setattr__`.  I am not sure whether or not `__getitem__` and `__setitem__` are useful.

Issue #213, opened 2023-04-18
CI fail to find Numpy with pip install


j8xixo12:
CI fail to find `Numpy`, due to latest `pip` (23.1.0),
https://github.com/solvcon/modmesh/actions/runs/4724133935/jobs/8380959174#step:7:123

j8xixo12:
It is hard to debug inside `pip`, the temporary solution of this issue is to use older pip (23.0.1) to install modmesh in CI test case.

yungyuc:
It seems to fail again: https://github.com/solvcon/modmesh/actions/runs/4735343779/jobs/8405439558

Maybe we need to fix the `setup.py`.

j8xixo12:
In order to fix #220, it will encounter symbol resolution issue in GitHub Actions environment and let CI fail.
I has fixed this in **ubuntu** and **macOS** environments but **Windows** environment still need some time to fix it.

yungyuc:
Thanks @j8xixo12 .  What about creating a PR to discuss possible ways to move forward?

j8xixo12:
> Thanks @j8xixo12 . What about creating a PR to discuss possible ways to move forward?

@yungyuc I jsut file a PR #221.

yungyuc:
This issue is fixed with #222 .  We track the new windows CI failure with #220.

PR #214, opened 2023-04-18
fix ci nouse_install fail

j8xixo12:
Temporary cancel upgrade pip version to avoid ci fail, because nouse_install will fail if we are using latest pip (23.1.0), more detail can be found at #213


yungyuc (approved):
Thanks a lot for finding this workaround!

PR #215, opened 2023-04-20
change pyside api visibility to default

j8xixo12:
Fix link error by changing pyside api visibility from hidden to default. If your using clang then this issue will be by pass, due to modmesh symbol only be hiidden when using gcc.


yungyuc (approved):
LGTM.  Thanks for fixing it.

PR #216, opened 2023-05-06
add Qt main window module level handler in view module

j8xixo12:
If a Qt object is created and managed by C++, when python needs to access it via `PySide6`, it needs to create a variable first to handle it first.

During python create a `PySide6` variable to handle c++ Qt object instance, it also register this Qt object to `Shiboken6` resource manager, once python need to access this Qt object, `Shiboken6` will check this c++ object valid or not, if user not register this Qt object to `Shiboken6` first it will raise runtime error.

`mainWindow` is parent node of `modmesh viewer`, therefore modems apps or widgets usually set `mainWindow` as parent, therefore it is high probability that `mainWindow` will be accessed by python.

Here is the runtime error message:
```
  File "path/to/usr/lib/python3.10/site-packages/matplotlib/backends/backend_qt.py", line 239, in showEvent
    window.screenChanged.connect(self._update_screen)
RuntimeError: Internal C++ object (PySide6.QtGui.QWindow) already deleted.
```

### DIFF
--- a/modmesh/view.py
+++ b/modmesh/view.py
@@ -52,6 +52,13 @@ enable = False
 try:
     from _modmesh import view as _vimpl  # noqa: F401
     enable = True
+    # Before using _modmesh mianWindow api, the QMainWindow signature need
+    # to be imported or will get type error:
+    # TypeError: Unable to convert function return value to a Python type!
+    # Creating variable to handle Qt MainWindow which is created by c++ and
+    # registered it to Shiboken6 to prevent runtime error occured:
+    # RuntimeError:
+    # Internal C++ object (PySide6.QtGui.QWindow) already deleted.
     from PySide6.QtWidgets import QMainWindow
     _main_window = _vimpl.RManager.instance.mainWindow
 except ImportError:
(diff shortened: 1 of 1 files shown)


modmesh/view.py
```diff
@@ -52,6 +52,8 @@
 try:
     from _modmesh import view as _vimpl  # noqa: F401
     enable = True
+    from PySide6.QtWidgets import QMainWindow
```
yungyuc: Please summarize why you need to add this import and variable in comments above this line.

I think it's better to move this action to the pybind11 wrapper of `RWindow`.  You can leave it to me if not sure how and why to do it.

yungyuc (changes requested):
Thanks a lot.  This makes sense.  But the code will need some explaining comments inline.

yungyuc:
@j8xixo12 Could you please also add (quote) how the runtime error looks like in the PR description?

modmesh/view.py
j8xixo12: I can try to move this action to pybind11 wrapper, but should I finish it in this PR?

j8xixo12:
> @j8xixo12 Could you please also add (quote) how the runtime error looks like in the PR description?

I have attached runtime error message in PR description.

modmesh/view.py
yungyuc: Let's use another PR for moving it to pybind11.

yungyuc:
Please use a new PR to move the Python code to pybind11.  Let me know if you are not sure how to do it.

PR #217, opened 2023-05-10
move pyside6 module init flow to pybind11

j8xixo12:
Based on PR #216 moved pyside6 module init flow
to pybnid11 wrapper.

### DIFF
--- a/cpp/modmesh/view/wrap_view.cpp
+++ b/cpp/modmesh/view/wrap_view.cpp
@@ -483,6 +483,21 @@ void wrap_view(pybind11::module & mod)
     WrapRManagerProxy::commit(mod, "RManagerProxy", "RManagerProxy");
 
     mod.attr("mgr") = RManagerProxy();
+
+    try
+    {
+        // Creating module level variable to handle Qt MainWindow which is
+        // created by c++ and registered it to Shiboken6 to prevent runtime
+        // error occured.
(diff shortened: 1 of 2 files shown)


cpp/modmesh/view/wrap_view.cpp
```diff
@@ -483,6 +483,21 @@ void wrap_view(pybind11::module & mod)
...
+        // created by c++ and registered it to Shiboken6 to prevent runtime
+        // error occured.
+        // RuntimeError:
+        // Internal C++ object (PySide6.QtGui.QWindow) already deleted.
+        // py::module::import("PySide6.QtWidgets");
+        py::globals()["_mainWindow"] = RManager::instance().mainWindow();
+    }
+    catch (const pybind11::error_already_set & e)
+    {
+        std::cerr << e.what() << std::endl;
```
yungyuc, line 499: Is it intentional that you do not rethrow?  May I understand the rationale?

j8xixo12: Yes it is.
I have referenced some similar code in modmesh, these codes just print the error, I was thought these error used to notify developer there are some error.

It seems I should do something to handle this if error is occurred.

yungyuc:
Sorry, this slipped through my fingers.  Merging.

Issue #220, opened 2023-05-19
PySide6 should default enable in modmesh


j8xixo12:
Since https://github.com/solvcon/modmesh/commit/214b4743bae1995cf7c451855880a397eae559fc has mentioned that `PySide6` is mandatory, `MODMESH_USE_PYSDIE` option should be removed.

j8xixo12:
I just file a PR #221 that remove `MODMESH_USE_PYSIDE` in modmesh project.

yungyuc:
@j8xixo12 @terrychan999 Windows CI is still failing: https://github.com/solvcon/modmesh/actions/runs/5036470016 .  Could you please fix it?

j8xixo12:
@yungyuc I am working on it!
I found pyside6 and shiboken6 abi library name should be `pyside6.*.lib`, `pyside6.*.dll`, `shiboken6.*.lib` and `shiboken6.*.dll` instead of `libpyside6.*.so` and `libshiboken6.*.so`.
So I fixed this by modified `CMakeLists.txt`, but I got a new error:  https://github.com/j8xixo12/modmesh/actions/runs/5036515452/jobs/9032713027#step:12:68

j8xixo12:
The return code `-1073741515` can be convert to hex `0xc0000135` which is `STATUS_DLL_NOT_FOUND`.
I also find a command can dump a execute file depends on what dll called `dumpbin`, it just like Linux's `ldd`.
But it didn't show dll can be found or not....
https://github.com/j8xixo12/modmesh/actions/runs/5059791675/jobs/9081829684#step:12:72

yungyuc:
`dumpbin` sounds like a great tool that we must include in the CI.  Maybe we should add `ldd` in the Ubuntu run as well (later if we have time).  Any way, it looks like we are heading toward a right direction.

PR #210, opened 2023-04-07
add plot manager and several physics parameters calculation

j8xixo12:
In this PR added following physics parameters calculation:
1. Temperature
2. Entopy
3. Internal energy

Also added plot manager which can save current plot via file explorer window and it can switch plot layout between
grid layout or single plot by click button. 

Due to plot manager inherit from `PySide6.QWidgets.QWidget`, `PySide6` is mandatory now. 

Connect to #202

PR #221, opened 2023-05-20
Remove modmesh use pyside

j8xixo12:
#210 has mentioned `PySide6` is mandatory, I removed `MODMESH_USE_PYSIDE` in modmesh project.

This patch encounter `PySide6` symbol resolution issue in GitHub Actions `windows` environment.

link to #220

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -123,7 +123,7 @@ jobs:
       run: |
         sudo pip3 install setuptools
         sudo pip3 install numpy pytest flake8
-        sudo pip3 install pyside6 shiboken6
+        sudo pip3 install pyside6==$(qmake6 -query QT_VERSION)
 
     - name: dependency by manual script
       run: sudo ${GITHUB_WORKSPACE}/contrib/dependency/install.sh pybind11
@@ -193,7 +193,8 @@ jobs:
 
     - name: make run_viewer_pytest
       run: |
-        make run_viewer_pytest VERBOSE=1
+        export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
+        make run_viewer_pytest VERBOSE=0
 
   build_macos:
 
@@ -239,7 +240,7 @@ jobs:
           python3 -m pip -v install --upgrade setuptools
           python3 -m pip -v install --upgrade pip
           python3 -m pip -v install --upgrade numpy pytest flake8
-          python3 -m pip -v install --upgrade pyside6 shiboken6
+          python3 -m pip -v install --upgrade pyside6==$(qmake -query QT_VERSION)
 
       - name: dependency by manual script
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/devbuild.yml
```diff
@@ -123,7 +123,7 @@ jobs:
       run: |
         sudo pip3 install setuptools
         sudo pip3 install numpy pytest flake8
-        sudo pip3 install pyside6 shiboken6
+        sudo pip3 install pyside6==$(qmake6 -query QT_VERSION)
```
j8xixo12, line 126: OS package manager Qt version may not match with default version of pip installed pyside6, therefore need to specific version of pyside6 when using pip install.

ubuntu 22.04 need to use `qmake6` in stead of `qmake`.
ref: https://askubuntu.com/questions/1460242/ubuntu-22-04-with-qt6-qmake-could-not-find-a-qt-installation-of

.github/workflows/lint.yml
```diff
@@ -122,6 +122,7 @@ jobs:
 
     - name: make run_viewer_pytest
       run: |
+        export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
```
j8xixo12, line 125: Directly specific `shiboken6` path to `LD_LIBRARY_PATH` that allow OS can find `shiboken6` during runtime.

cpp/modmesh/toggle/toggle.cpp
```diff
@@ -64,13 +64,7 @@ Toggle & Toggle::instance()
...
 
 SolidToggle::SolidToggle()
-    : m_use_pyside(
-#ifdef MODMESH_USE_PYSIDE
-          true
-#else
-          false
-#endif
-      )
+    : m_use_pyside(true)
```
j8xixo12, line 67: The reason of I leaves `m_use_pyside` here is modmesh may can select different UI backend in the future, we can use toggle system to select different backend.

.github/workflows/devbuild.yml
```diff
@@ -297,14 +298,18 @@ jobs:
             VERBOSE=1 USE_CLANG_TIDY=OFF \
             BUILD_QT=ON \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DMODMESH_USE_PYSIDE:BOOL=ON"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
           make buildext VERBOSE=1
 
       - name: make pytest BUILD_QT=ON
         run: |
+          # PySide6 installed by pip will bundle with a prebuilt Qt,
```
yungyuc, line 306: The comments clearly describe why we need to align the Qt library files and are important for future maintenance.  Thanks a lot.

cpp/modmesh/toggle/toggle.cpp
yungyuc: You read my mind.

yungyuc (approved):
LGTM.  Let me merge.

PR #223, opened 2023-05-21
Add a simple, demonstrative json schema in the unit tests for the toggle system

yungyuc:
The demonstrative json data are:

```json
[
  {
    "fixed":{
      "show_axis":false
    }
  },
  {
    "dynamic":{
      "apps":{
        "euler1d":{
          "use_sub":false
        }
      }
    }
  }
]
```

### DIFF
--- a/tests/test_toggle.py
+++ b/tests/test_toggle.py
@@ -275,6 +275,28 @@ class ToggleHierarchicalTC(unittest.TestCase):
                           'level1p.level2p.test_int32',
                           'level1p.test_bool', 'test_int8'])
 
+    def test_get_value(self):
+        tg = modmesh.Toggle.instance.clone()
+        tg.dynamic_clear()
+
+        tg.add_subkey("level1")
+        tg.set_int8("level1.test_int8", 21)
+        self.assertEqual(tg.level1.test_int8, 21)
+        self.assertEqual(tg.get_value('level1.test_int8'), 21)
+        self.assertEqual(tg.get_value(key='level1.test_int8'), 21)
+        self.assertEqual(tg.get_value('level1.test_int8', 22), 21)
+        self.assertEqual(tg.get_value('level1.test_int8', default=22), 21)
+        self.assertEqual(tg.get_value(key='level1.test_int8', default=22), 21)
+
+        self.assertEqual(tg.get_value('level1.non_exist', 22), 22)
+        self.assertEqual(tg.get_value('level1.non_exist', default=22), 22)
+        self.assertEqual(tg.get_value(key='level1.non_exist', default=22), 22)
+        with self.assertRaisesRegex(
+                AttributeError,
+                r'Cannot get non-existing key "level1.non_exist"'
+        ):
+            self.assertEqual(tg.level1.non_exist, 21)
+
 
 class ToggleSerializationTC(unittest.TestCase):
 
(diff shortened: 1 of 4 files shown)


tests/test_toggle.py
```diff
@@ -275,6 +275,28 @@ def test_multi_level(self):
...
+    def test_get_value(self):
+        tg = modmesh.Toggle.instance.clone()
+        tg.dynamic_clear()
+
+        tg.add_subkey("level1")
+        tg.set_int8("level1.test_int8", 21)
+        self.assertEqual(tg.level1.test_int8, 21)
+        self.assertEqual(tg.get_value('level1.test_int8'), 21)
+        self.assertEqual(tg.get_value(key='level1.test_int8'), 21)
+        self.assertEqual(tg.get_value('level1.test_int8', 22), 21)
```
yungyuc, line 287: Add tests for the default value.  If the value is there, the default value is not used.

tests/test_toggle.py
```diff
@@ -275,6 +275,28 @@ def test_multi_level(self):
...
+        self.assertEqual(tg.get_value('level1.test_int8', default=22), 21)
+        self.assertEqual(tg.get_value(key='level1.test_int8', default=22), 21)
+
+        self.assertEqual(tg.get_value('level1.non_exist', 22), 22)
```
yungyuc, line 291: If the key is not there, the default value is used.

tests/test_toggle.py
```diff
@@ -336,6 +358,38 @@ def test_dynamic_to_json(self):
...
+        self.assertEqual(tg.apps.euler1d.use_sub, False)
+
+        fixture = '''[{"fixed": {"show_axis": false}},
+{"dynamic": {"apps": {"euler1d": {"use_sub": true}}}}]'''
+        tg = modmesh.toggle.load(
+            fixture,
+            toggle_instance=modmesh.Toggle.instance.clone())
+        self.assertEqual(tg.apps.euler1d.use_sub, True)
+
+    @unittest.skip("the lifecycle issue may cause segfault")
```
yungyuc, line 378: I did something wrong ... This test function is to remind us that I need to fix the mistake I made.

tests/test_toggle.py
```diff
@@ -336,6 +358,38 @@ def test_dynamic_to_json(self):
...
+{"dynamic": {"apps": {"euler1d": {"use_sub": false}}}}]'''
+        # TODO: Need to fix this later. It may be wrong lifecycle handling with
+        # WrapHierarchicalToggleAccess.
+        with self.assertRaisesRegex(
+                AttributeError,
+                r'Cannot get non-existing key "apps.euler1d"'
+        ):
+            modmesh.toggle.load(
+                fixture,
+                toggle_instance=modmesh.Toggle.instance.clone()).apps.euler1d
```
yungyuc, line 390: `(temporary).apps` works.  But `(temporary).apps.euler1d` crashes or fails.

I am guessing `HierarchicalToggleAccess` does not hold the lifecycle of the parent `Toggle` object.  I tried to use return value policy but it does not help.  I need more time to study how to deal with it.

tests/test_toggle.py
```diff
@@ -336,6 +358,38 @@ def test_dynamic_to_json(self):
         self.assertEqual(json.loads(json.dumps(data)), golden)
 
 
+class ToggleLoadTC(unittest.TestCase):
+
+    def test_load(self):
+        fixture = '''[{"fixed": {"show_axis": false}},
```
yungyuc, line 364: @buganini, this is for your reference.

It's far from complete, but a starting point for how a json configuration file may look like.

PR #225, opened 2023-05-27
Use toggle for python_redirect

yungyuc:
The `PythonStreamRedirect` object now uses toggle to check for its enablement.

### DIFF
--- a/cpp/modmesh/python/common.hpp
+++ b/cpp/modmesh/python/common.hpp
@@ -511,8 +511,8 @@ class PythonStreamRedirect
 public:
 
     PythonStreamRedirect(bool enabled)
-        : m_enabled(enabled)
     {
+        set_enabled(enabled);
     }
 
     std::string stdout_string() const;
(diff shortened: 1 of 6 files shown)


cpp/modmesh/python/common.hpp
```diff
@@ -511,21 +511,21 @@ class PythonStreamRedirect
...
+        set_enabled(enabled);
     }
 
     std::string stdout_string() const;
     std::string stderr_string() const;
 
     PythonStreamRedirect & set_enabled(bool enabled)
     {
-        m_enabled = enabled;
+        Toggle::instance().fixed().set_python_redirect(enabled);
```
yungyuc, line 523: Use the `Toggle` for enablement.

cpp/modmesh/view/RPythonConsoleDockWidget.hpp
```diff
@@ -116,7 +116,7 @@ public slots:
     int m_current_command_index = 0;
     size_t m_past_limit = 1024;
 
-    python::PythonStreamRedirect m_python_redirect{/* enabled */ true};
+    python::PythonStreamRedirect m_python_redirect;
```
yungyuc, line 119: Do not use compile-time default value here.

PR #226, opened 2023-05-27
Label the executed commands with serial number

yungyuc:
Also switch back to use plain text for showing the text, instead of HTML.

After the change, the Python console looks like:

image

The HTML style looks like:

image

The HTML style has unnecessary blank lines between the Python command and its output.  I am not sure how to get rid of it.  It is also getting in my way of inserting the command serial number.  The best way seems to be switching back to use plain text for showing the text.

If coloring is needed, we may study how to use Qt native way to do it.

### DIFF
--- a/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
+++ b/cpp/modmesh/view/RPythonConsoleDockWidget.cpp
@@ -155,8 +155,17 @@ void RPythonConsoleDockWidget::printCommandHistory()
     QTextCursor cursor = m_history_edit->textCursor();
     cursor.movePosition(QTextCursor::End);
     m_history_edit->setTextCursor(cursor);
+    std::string lead;
+    if (m_past_command_strings.size() > 1)
+    {
+        lead = Formatter() << "\n[";
+    }
+    else
+    {
+        lead = Formatter() << "[";
+    }
     m_history_edit->insertPlainText(QString::fromStdString(
-        Formatter() << "[" << m_last_command_serial << "] "
+        Formatter() << lead << m_last_command_serial << "] "
                     << m_command_edit->toPlainText().toStdString() << "\n"));
     cursor.movePosition(QTextCursor::End);
     m_history_edit->setTextCursor(cursor);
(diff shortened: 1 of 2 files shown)


cpp/modmesh/view/RPythonConsoleDockWidget.cpp
```diff
@@ -154,7 +155,9 @@ void RPythonConsoleDockWidget::printCommandHistory()
     QTextCursor cursor = m_history_edit->textCursor();
     cursor.movePosition(QTextCursor::End);
     m_history_edit->setTextCursor(cursor);
-    m_history_edit->insertHtml(m_commandHtml + m_command_edit->toPlainText() + m_endHtml);
+    m_history_edit->insertPlainText(QString::fromStdString(
+        Formatter() << "[" << m_last_command_serial << "] "
```
yungyuc: The prompt will look like:

```
[1] print(True)
True
```

cpp/modmesh/view/RPythonConsoleDockWidget.hpp
```diff
@@ -115,14 +115,9 @@ public slots:
     std::deque<std::string> m_past_command_strings;
     int m_current_command_index = 0;
     size_t m_past_limit = 1024;
+    int m_last_command_serial = 0;
```
yungyuc, line 118: `m_last_command_serial` is for showing in the console text area.  `m_current_command_index` is the index to `m_past_command_strings`.  It is possible that `m_last_command_serial > m_past_command_strings.size()`.

tigercosmos:
LGTM. I just think using HTML is easier to achieve coloring text, though.
In the future, it is likely that we will require the development of a versatile rendering class to handle colored text. This class may also encompass functionalities for specifying padding, borders, background color, and other properties for individual commands and their corresponding output.

tigercosmos (approved):

cpp/modmesh/view/RPythonConsoleDockWidget.cpp
tigercosmos: suggestion, maybe one more blank line between two commands.
```
[1] print(True)
True

[2] print(True)
True
```

yungyuc: Good idea.  Patch is coming.

image

yungyuc:
Agree

PR #227, opened 2023-05-28
Fix windows ci

j8xixo12:
Due to #220, `PySide6` and `Shiboken6` dependencies needs to be considered in windows ci environment.

`PySide6` and `Shiboken6` has different library name in windows, therefore `CMakeLists.txt` needs to be modified to address this problem.

`PySide6` and `Shiboken6` library path also needs to be added in `PATH`, that allow exe file can find dll during runtime.


.github/workflows/devbuild.yml
```diff
@@ -406,6 +406,9 @@ jobs:
 
       - name: cmake run_viewer_pytest
         run: |
+          $pyside6_path = $(python3 -c "import sys, os, PySide6; sys.stdout.write(os.path.dirname(PySide6.__file__))")
```
yungyuc, line 409: I am guess that you add the environment variable only in this step because you do not want to interfere the non-view step run_gtest?  If so, I think we can use a comment to spell out the subtle intention.  Otherwise maintainers will wonder why not simply pushing the environment PATH to the global when pyside6 is installed.

CMakeLists.txt
```diff
@@ -123,8 +123,13 @@ if(NOT "${SHIBOKEN6_PYTHON_PACKAGE_PATH}" STREQUAL "")
     link_directories("${SHIBOKEN6_PYTHON_PACKAGE_PATH}")
 endif()
 
-file(GLOB PYSIDE6_LIBFILE LIST_DIRECTORIES false "${PYSIDE6_PYTHON_PACKAGE_PATH}/libpyside6.*")
-file(GLOB SHIBOKEN6_LIBFILE LIST_DIRECTORIES false "${SHIBOKEN6_PYTHON_PACKAGE_PATH}/libshiboken6.*")
+if(MSVC)
```
yungyuc, line 126: We can use a comment for why we need these variable overrides on windows.  If there are some references (URLs) it's good to add them in the comments.  It will greatly help future maintainers.

yungyuc (approved):
It's wonderful!  A simple patch that comes with countless hours of troubleshoot.  I admire the hard work.

I have no problem for the code to be merged.  But addition of some comments will make it even better.

yungyuc:
Simple patch, countless hours of troubleshoot...

yungyuc:
Let me push the button to merge.  The comments can use another PR.

.github/workflows/devbuild.yml
j8xixo12: No, I have tried to add the environment in a new step, but it seems didn’t work, so I put these code in this step.
I will find out why and use other PR to add comment to explain about this!

CMakeLists.txt
j8xixo12: Oh I found this different by browse pyside6 and shiboken6 folder.
I will add a comment here!

.github/workflows/devbuild.yml
yungyuc: No problem.

PR #228, opened 2023-05-28
Remove ubuntu-20.04 and macos-11 from nouse_install.yml

yungyuc:
ubuntu-20.04 and macos-11 are older arches that modmesh devs are not using.  They can be removed from the CI to save waiting time.


yungyuc:
@j8xixo12 I'd like to learn what do you think?

j8xixo12:
`nouse_install` really takes lots of time waiting, remove old archs is good!
LGTM.

PR #229, opened 2023-05-30
enhance github workflow and add explain comment

j8xixo12:
Added PySide6 and Shiboken6 into system path in windows CI by modified GITHUB_PATH, which is the only way to modify system path in github actions container.

Also add some explanatory comment in CMakeLists.txt to explain why there are 2 ways to add library path.


yungyuc (approved):
LGTM

PR #232, opened 2023-07-10
Fix 1d cell type nface return value

j8xixo12:
`CellType` `nface()` only considers 2, 3 dimension, however `CellType` also contains 1 dimension cell type,
`face()` should considering 1 dimension case.

In this PR I also updated a test case to test 1 dimension cell type behavior.

### DIFF
--- a/cpp/modmesh/mesh/StaticMesh.hpp
+++ b/cpp/modmesh/mesh/StaticMesh.hpp
@@ -92,7 +92,15 @@ struct CellType : NumberBase<int32_t, double>
     uint8_t nedge() const { return m_nedge; }
     uint8_t nsurface() const { return m_nsurface; }
 
-    uint8_t nface() const { return 2 == m_ndim ? nedge() : nsurface(); }
+    uint8_t nface() const
+    {
+        switch (m_ndim)
+        {
+        case 1: return nnode(); break;
+        case 2: return nedge(); break;
+        default: return nsurface(); break;
+        }
+    }
 
     const char * name() const
     {
--- a/cpp/modmesh/mesh/StaticMesh.hpp
+++ b/cpp/modmesh/mesh/StaticMesh.hpp
@@ -75,10 +75,10 @@ struct CellType : NumberBase<int32_t, double>
     CellType(uint8_t id_in, uint8_t ndim_in, uint8_t nnode_in, uint8_t nedge_in, uint8_t nsurface_in)
         : m_id(id_in)
         , m_ndim(ndim_in)
-        , m_nnode(nnode_in)
-        , m_nedge(nedge_in)
-        , m_nsurface(nsurface_in)
     {
+        m_attrs[NNODE_IDX] = nnode_in;
+        m_attrs[NEDGE_IDX] = nedge_in;
+        m_attrs[NSURFACE_IDX] = nsurface_in;
     }
 
     CellType()
@@ -88,19 +88,10 @@ struct CellType : NumberBase<int32_t, double>
 
     uint8_t id() const { return m_id; }
     uint8_t ndim() const { return m_ndim; }
-    uint8_t nnode() const { return m_nnode; }
-    uint8_t nedge() const { return m_nedge; }
-    uint8_t nsurface() const { return m_nsurface; }
-
-    uint8_t nface() const
-    {
-        switch (m_ndim)
-        {
-        case 1: return nnode(); break;
-        case 2: return nedge(); break;
-        default: return nsurface(); break;
-        }
-    }
+    uint8_t nnode() const { return m_attrs[NNODE_IDX]; }
+    uint8_t nedge() const { return m_attrs[NEDGE_IDX]; }
+    uint8_t nsurface() const { return m_attrs[NSURFACE_IDX]; }
+    uint8_t nface() const { return m_attrs[m_ndim % 3]; } // Using mod can prevent idx out-of-boundary occurs.
 
     const char * name() const
     {
@@ -120,12 +111,13 @@ struct CellType : NumberBase<int32_t, double>
     }
 
 private:
+    static constexpr const uint8_t NNODE_IDX = 1;
+    static constexpr const uint8_t NEDGE_IDX = 2;
+    static constexpr const uint8_t NSURFACE_IDX = 0;
 
     uint8_t m_id : 6;
     uint8_t m_ndim : 2;
-    uint8_t m_nnode = 0;
-    uint8_t m_nedge = 0;
-    uint8_t m_nsurface = 0;
+    uint8_t m_attrs[3] = {0};
 
 }; /* end struct CellType */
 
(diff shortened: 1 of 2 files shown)


tests/test_mesh.py
```diff
@@ -201,4 +201,28 @@ def test_3d_single_tetrahedron(self):
...
+                          nedge=0)
+        self._check_metric_trivial(mh)
+
+        mh.build_interior()  # _do_metric=True, _build_edge=True
+        self._check_shape(mh, ndim=1, nnode=2, nface=2, ncell=1,
+                          nbound=0, ngstnode=0, ngstface=0, ngstcell=0,
+                          nedge=2)
+
+        # _do_metric do nothing due to dim == 1
+        self._check_metric_trivial(mh)
```
j8xixo12, line 227: I haven't completed the code study in `build_boundary` and `build_ghost`, so I decided not to call them in test case until I complete the code study.

cpp/modmesh/mesh/StaticMesh.hpp
```diff
@@ -92,7 +92,15 @@ struct CellType : NumberBase<int32_t, double>
     uint8_t nedge() const { return m_nedge; }
     uint8_t nsurface() const { return m_nsurface; }
 
-    uint8_t nface() const { return 2 == m_ndim ? nedge() : nsurface(); }
+    uint8_t nface() const
```
yungyuc: `nface()` should be a fast function.  It would be great if it has not any branch (if/else), but the original ternary expression already broke it.  When fixing the behavior bug, it would be great to also fix the performance issue.  Could you please take a look?  One way to do it is to compare the generated assembly.

tests/test_mesh.py
yungyuc: It's OK.  Maybe leave a comment for future work.

yungyuc (changes requested):
This is good start.  I left a comment for the performance concern.  Let's discuss how to deal with it.

cpp/modmesh/mesh/StaticMesh.hpp
j8xixo12: I can take a look on this issue!

The technique that first comes to my mind is table lookup, but it needs to maintain a table to store function's informations.

I'm trying to search more informations about branchless code technique.

j8xixo12:
I use a array to store `m_nnode`, `m_nedge` and `m_nsurface`, then `nface` can get these information without any branch, just using index to access it.

I also compare switch case version and branchless version assembly code in **debug build**, branchless version has shorter assembly code (27 lines to 24 lines), in release build I can not find `struct CellType` symbol, it should be striped.

But I think this commit may reduced the readability of `struct CellType`.

cpp/modmesh/mesh/StaticMesh.hpp
```diff
@@ -75,10 +75,10 @@ struct CellType : NumberBase<int32_t, double>
     CellType(uint8_t id_in, uint8_t ndim_in, uint8_t nnode_in, uint8_t nedge_in, uint8_t nsurface_in)
         : m_id(id_in)
         , m_ndim(ndim_in)
-        , m_nnode(nnode_in)
-        , m_nedge(nedge_in)
-        , m_nsurface(nsurface_in)
     {
+        m_attrs[NNODE_IDX] = nnode_in;
```
yungyuc: Why not using initialization list like:

```cpp
, m_attrs{0, 0, 0}
```

cpp/modmesh/mesh/StaticMesh.hpp
```diff
@@ -112,12 +111,13 @@ struct CellType : NumberBase<int32_t, double>
...
+    static constexpr const uint8_t NNODE_IDX = 1;
+    static constexpr const uint8_t NEDGE_IDX = 2;
+    static constexpr const uint8_t NSURFACE_IDX = 0;
 
     uint8_t m_id : 6;
     uint8_t m_ndim : 2;
-    uint8_t m_nnode = 0;
-    uint8_t m_nedge = 0;
-    uint8_t m_nsurface = 0;
+    uint8_t m_attrs[3] = {0};
```
yungyuc: Why `= {0}` rather than `= {0, 0, 0}`?  Invalid syntax?

If it's valid, the latter reads more explicit and less error-prone.

tests/test_mesh.py
yungyuc: Is there a reason not to leave a comment?

yungyuc (changes requested):

cpp/modmesh/mesh/StaticMesh.hpp
```diff
@@ -88,11 +88,10 @@ struct CellType : NumberBase<int32_t, double>
...
     uint8_t ndim() const { return m_ndim; }
-    uint8_t nnode() const { return m_nnode; }
-    uint8_t nedge() const { return m_nedge; }
-    uint8_t nsurface() const { return m_nsurface; }
-
-    uint8_t nface() const { return 2 == m_ndim ? nedge() : nsurface(); }
+    uint8_t nnode() const { return m_attrs[NNODE_IDX]; }
+    uint8_t nedge() const { return m_attrs[NEDGE_IDX]; }
+    uint8_t nsurface() const { return m_attrs[NSURFACE_IDX]; }
+    uint8_t nface() const { return m_attrs[m_ndim % 3]; } // Using mod can prevent idx out-of-boundary occurs.
```
yungyuc: Please just use `m_ndim` to avoid the unnecessary mod.  `m_ndim` should never be larger than 3.

yungyuc (changes requested):

tests/test_mesh.py
```diff
@@ -201,4 +201,30 @@ def test_3d_single_tetrahedron(self):
...
+        # TODO: Need to add build_boundary and build_ghost to make sure
```
j8xixo12, line 228: I left a comment to mention that there are some test need to be added in the future.

cpp/modmesh/mesh/StaticMesh.hpp
j8xixo12: `= {0}` and `= {0,0,0}` are valid syntax, but for more readability and less error-prone `= {0,0,0}` would be better.
I will fix it.

cpp/modmesh/mesh/StaticMesh.hpp
j8xixo12: I fixed this in the new commit.

cpp/modmesh/mesh/StaticMesh.hpp
j8xixo12: I use `m_ndim - 1` instead of **mod**.

tests/test_mesh.py
yungyuc: I see.  Thanks.

yungyuc (approved):

PR #233, opened 2023-07-22
add Gmsh mesh file parser

j8xixo12:
In this commit added mesh file io for modmesh, it lets modmesh has capabilty to read third-party mesh generator generated file and convert the content to a data structure that allow modmesh can use it to do some calculation.

Currently mesh io only support Gmsh format.


modmesh/app/sample_mesh.py
```diff
@@ -83,6 +83,25 @@ def help_other(set_command=False):
         view.mgr.pycon.command = cmd.strip()
 
 
+def help_mesh_viewer(path, set_command=False):
```
j8xixo12, line 86: I also added a sample command in sample_mesh app for, once mesh viewer app is completed, then this sample command can be removed.

yungyuc:
This PR contains almost 1000 lines of diff and will not be easy to review.  Would it be possible to break it into smaller pieces?  For review efficiency a batch of review should be around 300 lines of diff.

modmesh/core.py
```diff
@@ -64,6 +64,7 @@
     'METAL_BUILT',
     'metal_running',
     'HAS_VIEW',
+    'Gmsh',
```
yungyuc: This is a class and should be put before constants, ordered in the same convention as other classes or functions.

cpp/modmesh/io/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+
```
yungyuc: `io` the name conflict with Python `io` module.  It is unnecessary ambiguity.  Please rename this directory as `inout`.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+#include <unordered_map>
+#include <functional>
+
+#include <modmesh/base.hpp>
+#include <modmesh/mesh/mesh.hpp>
+#include <modmesh/buffer/buffer.hpp>
+
+namespace modmesh
+{
+namespace IO
```
yungyuc: namespace should not use uppercase, and it should not have the unnecessary ambiguity to the Python module named `io`. 

In addition, `IO` is a pretty short name and may conflict with symbols in foreign code base.

Please rename it to `inout`.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+{
+namespace Gmsh
```
yungyuc: Please do not use uppercase in namespace. Beside, I don't think it is necessary to have 3 levels of namespace. Gmsh is just one of the format. Please remove the namespace.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+{
+struct ElementDef
```
yungyuc: What about naming it like `GmshElementDef`?

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+    small_vector<uint8_t> mmcl() const { return m_mmcl; }
+
+private:
+    uint8_t m_ndim = 0; /* Number of dimension */
+    uint16_t m_nnds = 0; /* Number of nodes     */
+    uint8_t m_mmtpn = 0; /* modmesh cell type   */
+    small_vector<uint8_t> m_mmcl; /* modmesh cell order  */
+}; /* end struct ElementDef */
+
+inline ElementDef ElementDef::by_id(uint16_t id)
```
yungyuc: This is not a template.  Please move it to a cpp file.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+{
+#define VEC(...) __VA_ARGS__
+#define MM_DECL_SWITCH_ELEMENT_TYPE(ID, NDIM, NNDS, MMTPN, MMCL) \
```
yungyuc: The use of macro makes the code easy to read.  Thanks.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+    using number_base = NumberBase<int32_t, double>;
+    using int_type = typename number_base::int_type;
+    using uint_type = typename number_base::uint_type;
+    using real_type = typename number_base::real_type;
+
+public:
+    explicit Gmsh(const std::string & file_path)
+    {
+        bool meta_enter = false, node_enter = false, element_enter = false;
+        // clang-format off
```
yungyuc: You need to turn clang-format back on!!

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+    {
+        BEGIN,
+        META_END,
+        PYHSICAL_NAME_END,
+        NODE_END,
+        ELEMENT_END
+    };
+
+    // Check the finite state machine state transition is valid or not to check msh file format is correct
+    bool _isValidTransition(const std::string s)
```
yungyuc: Do not prepend `_` to any C++ symbols.  You already made them private and mixing the C++ access with Python private convention is a bad idea.

cpp/modmesh/io/gmsh.hpp
```diff
@@ -0,0 +1,397 @@
...
+class Gmsh
+    : public NumberBase<int32_t, double>
+{
+    using number_base = NumberBase<int32_t, double>;
+    using int_type = typename number_base::int_type;
+    using uint_type = typename number_base::uint_type;
+    using real_type = typename number_base::real_type;
+
+public:
+    explicit Gmsh(const std::string & file_path)
```
yungyuc: Could you make it also take a string as data, in addition to a string as path?  Both are useful.

gtests/test_nopython_io.cpp
```diff
@@ -0,0 +1,350 @@
+#include <modmesh/io/io.hpp>
```
yungyuc: Too long to review.  I trust you have made comprehensive testing.  If not, please speak up.

yungyuc (changes requested):
Thanks for the great work.  Let's polish the code.

gtests/test_nopython_io.cpp
j8xixo12: I have tested each cell type conversion between `gmsh` and `modmesh`. Additionally, I also tested if a user inputs a cell type that is not listed in `GmshElementDef`. The expected behavior is that `GmshElement::by_id` should return an empty `GmshElementDef` object.

cpp/modmesh/io/CMakeLists.txt
j8xixo12: I renamed the module name to `inout` in https://github.com/solvcon/modmesh/pull/233/commits/7432534167ee1da998cfb1a7024ed5e8f52f869f

cpp/modmesh/io/gmsh.hpp
j8xixo12: I renamed the namespace to `inout` in https://github.com/solvcon/modmesh/pull/233/commits/7432534167ee1da998cfb1a7024ed5e8f52f869f too.

cpp/modmesh/io/gmsh.hpp
j8xixo12: I have removed namespace `Gmsh` in  https://github.com/solvcon/modmesh/pull/233/commits/7432534167ee1da998cfb1a7024ed5e8f52f869f

cpp/modmesh/io/gmsh.hpp
j8xixo12: I renamed `ElementDef` to `GmshElementDef` in https://github.com/solvcon/modmesh/pull/233/commits/7432534167ee1da998cfb1a7024ed5e8f52f869f

cpp/modmesh/io/gmsh.hpp
j8xixo12: I change it to use string as data and python can take care of file stream to `bytes`, even if user is using `Gmsh` in cpp, cpp also can take care of file stream to string.

cpp/modmesh/io/gmsh.hpp
j8xixo12: `GmshElementDef::by_id` is a inline function should I move it to cpp file?
This function will be accessed by gtest test case, if move it to cpp file, gtest will encounter undefined reference.

cpp/modmesh/io/gmsh.hpp
j8xixo12: I have removed `_` from these private member function.

cpp/modmesh/io/CMakeLists.txt
yungyuc: Thanks

cpp/modmesh/inout/gmsh.hpp
```diff
@@ -0,0 +1,320 @@
...
+    {
+        BEGIN,
+        META_END,
+        PYHSICAL_NAME_END,
+        NODE_END,
+        ELEMENT_END
+    };
+
+    // Check the finite state machine state transition is valid or not to check msh file format is correct
+    bool isValidTransition(const std::string s)
```
yungyuc: Ah, I forgot to remind that C++ function names should follow snake_case.  You used snake_case in other functions.  This looks like an overlook.  Could you change it and review all in your PR?

yungyuc (changes requested):
It's quite close to completion.  Only some minor formatting issues remain.

cpp/modmesh/inout/gmsh.hpp
j8xixo12: I fixed this in https://github.com/solvcon/modmesh/pull/233/commits/f7de14ff4eb817cc2c858ee14dfcaa4ec52f2a49
I changed `isValidTransition` and `toblock` to snake case in `Gmsh` class

cpp/modmesh/inout/pymod/wrap_Gmsh.cpp
```diff
@@ -0,0 +1,46 @@
...
+    {
+        namespace py = pybind11; // NOLINT(misc-unused-alias-decls)
+
+        (*this)
+            .def(
+                py::init(
+                    [](const py::bytes & data)
+                    { return std::make_shared<inout::Gmsh>(data); }),
+                py::arg("data"))
+            .def("toblock", &wrapped_type::to_block);
```
yungyuc: I guess it's still an overlook here.  It makes sense to align the function names between Python and C++.

yungyuc (changes requested):
Yet a minor mismatch.  Should be a simple fix 🙂

yungyuc (approved):

yungyuc:
Thanks a lot!

PR #234, opened 2023-08-27
Fix mesh

j8xixo12:
There are 2 patch in this PR:

1. add point type face handler for Face builder:
The buffer used by mesh is pre-allocated when the `gmsh` parser parses the mesh. The buffer for faces is initialized to -1. If the Face builder does not handle point-type faces, then `make_ndfcs()` will get an index overflow because the initial value of -1 is used to calculate the pointer shift.

2. The index used by `m_ndcrd` in `StaticMesh::fill_ghost()` should be `node` index not `cell` index, this will cause index overflow.

### DIFF
--- a/cpp/modmesh/mesh/StaticMesh_boundary.cpp
+++ b/cpp/modmesh/mesh/StaticMesh_boundary.cpp
@@ -261,7 +261,7 @@ inline void StaticMesh::fill_ghost()
                 }
                 for (size_t idm = 0; idm < m_ndim; ++idm)
                 {
-                    m_ndcrd(igcl, idm) = m_ndcrd(icl, idm) + 2 * dist * m_fcnml(ibfc, idm);
+                    m_ndcrd(ignd, idm) = m_ndcrd(ind, idm) + 2 * dist * m_fcnml(ibfc, idm);
                 }
                 // decrement ghost node counter.
                 ignd -= 1;
--- a/cpp/modmesh/mesh/StaticMesh_interior.cpp
+++ b/cpp/modmesh/mesh/StaticMesh_interior.cpp
@@ -46,6 +46,21 @@ struct FaceBuilder
     using uint_type = typename number_base::uint_type;
     using real_type = typename number_base::real_type;
 
+    size_t add_point(int_type icl, int_type ifc)
+    {
+        constexpr const size_t NFACE = 1;
+        clfcs(icl, 0) = NFACE;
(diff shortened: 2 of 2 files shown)


cpp/modmesh/mesh/StaticMesh_boundary.cpp
```diff
@@ -261,7 +261,7 @@ inline void StaticMesh::fill_ghost()
                 }
                 for (size_t idm = 0; idm < m_ndim; ++idm)
                 {
-                    m_ndcrd(igcl, idm) = m_ndcrd(icl, idm) + 2 * dist * m_fcnml(ibfc, idm);
+                    m_ndcrd(ignd, idm) = m_ndcrd(ind, idm) + 2 * dist * m_fcnml(ibfc, idm);
```
yungyuc, line 264: What a bug I added!  Thanks for catching this.

cpp/modmesh/mesh/StaticMesh_interior.cpp
```diff
@@ -46,6 +46,21 @@ struct FaceBuilder
     using uint_type = typename number_base::uint_type;
     using real_type = typename number_base::real_type;
 
+    size_t add_point(int_type icl, int_type ifc)
```
yungyuc, line 49: I do not understand why do we need this type.  It is not useful in a mesh.  Could you please elaborate more?

yungyuc (commented):
Thanks a lot, @j8xixo12 .  I have provided my comments (my apology for delaying it so long).  Could you please take a look and answer the question?

cpp/modmesh/mesh/StaticMesh_interior.cpp
j8xixo12: In gmsh parser I using the number of elements to pre-allocate buffer, this number is provide by msh file, but this number contains point type cell which is we want to ignore. 

`FaceBuilder` constructor will use the length of input cell type buffer to pre-allocate buffer and give a initial value with -1, then `FaceBuilder::populate()` will going to fill face information, but if I ignore point type cell, then the `fcnds` still has initial value -1, that will cause the index overflow in `FaceBuilder::make_ndfcs()`.

That’s why I need to add point type cell handler in `FaceBuilder`.

yungyuc: Thanks.  It's clear now.  When I dealt with the older version of gmsh, I didn't remember it has node type.  Perhaps it's a new data format.

Issue #27, opened 2021-12-19
Single Python wrapper for SimpleArray


yungyuc:
At the time being, the (C++) `SimpleArray` class template is wrapped to Python by using distinct wrapper classes (template instantiations) for each of the fundamental `value_type` of interest:

```cpp
WrapSimpleArray<bool>::commit(mod, "SimpleArrayBool", "SimpleArrayBool");
WrapSimpleArray<int8_t>::commit(mod, "SimpleArrayInt8", "SimpleArrayInt8");
WrapSimpleArray<int16_t>::commit(mod, "SimpleArrayInt16", "SimpleArrayInt16");
WrapSimpleArray<int32_t>::commit(mod, "SimpleArrayInt32", "SimpleArrayInt32");
WrapSimpleArray<int64_t>::commit(mod, "SimpleArrayInt64", "SimpleArrayInt64");
WrapSimpleArray<uint8_t>::commit(mod, "SimpleArrayUint8", "SimpleArrayUint8");
WrapSimpleArray<uint16_t>::commit(mod, "SimpleArrayUint16", "SimpleArrayUint16");
WrapSimpleArray<uint32_t>::commit(mod, "SimpleArrayUint32", "SimpleArrayUint32");
WrapSimpleArray<uint64_t>::commit(mod, "SimpleArrayUint64", "SimpleArrayUint64");
WrapSimpleArray<float>::commit(mod, "SimpleArrayFloat32", "SimpleArrayFloat32");
WrapSimpleArray<double>::commit(mod, "SimpleArrayFloat64", "SimpleArrayFloat64");
```

And in Python, the construction of `SimpleArray` needs to use the specific type names:

```python
sarr = modmesh.SimpleArrayFloat64((2, 3, 4))
self.assertEqual(2*3*4*8, sarr.nbytes)
self.assertEqual((2, 3, 4), sarr.shape)
self.assertEqual((12, 4, 1), sarr.stride)
```

It will be desirable to avoid the fundamental type name (`Float64`) in the array type name, so that the Python code can be written as:

```python
sarr = modmesh.SimpleArray((2, 3, 4), type='float64')  # or dtype='float64'
self.assertEqual(sarr.ndarray.dtype, np.float64)
```

We may choose to mimic the numpy API by using `dtype` or invent our own way.

Struct arrays are a different topic.  This issue focuses on the arrays of the fundamental types.

# References

* https://github.com/pybind/pybind11/issues/1176#issuecomment-343312352

[...]

PR #237, opened 2023-10-06
add createSimpleArray to wrap WrapSimpleArray

terrychan999:
This is a more detailed version of my example: https://github.com/solvcon/modmesh/issues/27#issuecomment-1732625041
Which can handle (numpy array) or (shape, type) or (shape, value, type) now.

### DIFF
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -407,10 +407,136 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         return shape;
     }
 
+public:
+    static SimpleArray<T> init_array(pybind11::array arr_in)
+    {
+        if (!dtype_is_type<T>(arr_in))
+        {
+            throw std::runtime_error("dtype mismatch");
+        }
+        shape_type shape;
+        for (ssize_t i = 0; i < arr_in.ndim(); ++i)
+        {
+            shape.push_back(arr_in.shape(i));
+        }
+        std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
+            arr_in.nbytes(),
+            arr_in.mutable_data(),
+            std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
+        return wrapped_type(shape, buffer);
+    }
+    static SimpleArray<T> init_shape(pybind11::tuple const & shape)
+    {
+        return wrapped_type(make_shape(shape));
+    }
+    static SimpleArray<T> init_value(pybind11::tuple const & shape, T const & value)
+    {
+        return wrapped_type(make_shape(shape), value);
+    }
 }; /* end class WrapSimpleArray */
+
+pybind11::object createSimpleArray(pybind11::args args, const pybind11::kwargs & kwargs)
+{
+    namespace py = pybind11;
+
+    int argc = args.size();
+    int arg_len = args.size() + kwargs.size();
+    if (arg_len < 1 || arg_len > 3)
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -407,10 +407,136 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+    namespace py = pybind11;
+
+    int argc = args.size();
+    int arg_len = args.size() + kwargs.size();
+    if (arg_len < 1 || arg_len > 3)
+        throw std::runtime_error("Invalid number of arguments.");
+
+    // get type
+    std::string type;
+    if (kwargs.contains("type"))
```
yungyuc, line 449: Numpy uses `dtype` instead of `type`.  We should be consistent to numpy.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -407,10 +407,136 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            return py::cast(WrapSimpleArray<float>::init_value(args[0].cast<py::tuple>(), args[1].cast<float>()));
+        if (type == "float64")
+            return py::cast(WrapSimpleArray<double>::init_value(args[0].cast<py::tuple>(), args[1].cast<double>()));
+    }
+    throw std::runtime_error("unsupported type.");
+}
+
 void wrap_SimpleArray(pybind11::module & mod)
 {
+    mod.def("SimpleArray", &createSimpleArray);
```
yungyuc, line 539: This is not a class (type), but a helper function using the class naming convention.  The returned type depends on the input argument, e.g.,

```python
type(SimpleArray((2, 3), dtype='int32'))) != type(SimpleArray((2, 3), dtype='float32')))
```

But we want the are the same type.  Thus, in the pybind11 wrapper, there should be a Python type created as a facade.

yungyuc (changes requested):
Thanks, @terrychan999 .  The PR is a good start for us to discuss what we really want.  Please take a look at the inline comments.

And I'd like this work to be compatible to https://data-apis.org/array-api/latest/ (additional read: https://labs.quansight.org/blog/scipy-array-api).

terrychan999:
To expose single type to Python, I'm planning to introduce a base class, `SimpleArrayBase` for `WrapSimpleArray<T>`, inspired by tigercosmos's example https://github.com/solvcon/modmesh/issues/27#issuecomment-1203078356.

yungyuc:
> To expose single type to Python, I'm planning to introduce a base class, `SimpleArrayBase` for `WrapSimpleArray<T>`, inspired by tigercosmos's example [#27 (comment)](https://github.com/solvcon/modmesh/issues/27#issuecomment-1203078356).

It's worth trying.  When do you think it may be ready for review?

terrychan999:
I think I need one week to work on it.
That is on 10/29 to be ready for review.

tigercosmos:
@yungyuc  let's close this PR since we have `SimpleArrayPlex` now.

yungyuc:
Sounds good, @tigercosmos .  Any comments, @terrychan999 ?

yungyuc:
@terrychan999 any comments?

terrychan999:
> @yungyuc let's close this PR since we have `SimpleArrayPlex` now.
> Sounds good, @tigercosmos . Any comments, @terrychan999 ?

Certainly, @yungyuc and @tigercosmos. I'm grateful for the guidance!

yungyuc:
Thanks, @tigercosmos @terrychan999

PR #158, opened 2022-11-26
Add MDI support to viewer

yungyuc:
The multiple document interface in action:

image

Some code gets refactored because there are more than one "view" windows in the central widget now.

PR #240, opened 2023-10-16
Fix camera menu

tychuang1211:
Since PR #158, the camera setting became dummy. This PR add camera setting code. The new camera controller will be applied to all sub-windows when switched through the menu.

### DIFF
--- a/cpp/modmesh/view/RManager.cpp
+++ b/cpp/modmesh/view/RManager.cpp
@@ -153,17 +153,43 @@ void RManager::setUpMenu()
         auto * use_orbit_camera = new RAction(
             QString("Use Orbit Camera Controller"),
             QString("Use Oribt Camera Controller"),
-            []()
+            [this]()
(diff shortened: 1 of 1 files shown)


cpp/modmesh/view/RManager.cpp
```diff
@@ -153,17 +153,43 @@ void RManager::setUpMenu()
         auto * use_orbit_camera = new RAction(
             QString("Use Orbit Camera Controller"),
             QString("Use Oribt Camera Controller"),
-            []()
+            [this]()
             {
                 qDebug() << "Use Orbit Camera Controller (menu demo)";
+                for (auto subwin : m_mdiArea->subWindowList())
```
yungyuc, line 159: This duplicates the code at line 180.  In the future we should factor it out to be a common helper.

yungyuc (approved):
LGTM

Issue #212, opened 2023-04-09
euler1d app GUI does not look right


yungyuc:
The new visualizer added in #210 (issue #202) does not look right:

<img width="1541" alt="image" src="https://user-images.githubusercontent.com/399122/230759391-0dd90af1-e670-4516-aa0d-486069113af2.png">

@j8xixo12 Could you take a look?

# Update on Oct 20

See https://github.com/solvcon/modmesh/issues/212#issuecomment-1772420049

- [x] Visualization should look good in both sub-window and (standalone) popped-out window.
- [x] The main window size should be set to accommodate the sub-window.

j8xixo12:
The `Runtimeerror: Internal C++ object (pySide6.QtGui.QtGui)` was caused by #202, Euler1d need to access c++ side Qt object and this Qt object need to by managed by `shiboken6` or it will raise runtime error.

This issue has fixed by https://github.com/solvcon/modmesh/commit/93b751574f43288e67b70f1d0655ea4982630773

yungyuc:
The plots still looks as I shown above: very compressed and inconsistent.  If it looks better on your side, could you please share how it looks and how do you make it look better, @j8xixo12 ?

j8xixo12:
No it doesn't looks better on my side.
<img width="1112" alt="截圖 2023-05-22 上午8 25 03" src="https://github.com/solvcon/modmesh/assets/21517083/19013d21-dc2f-4aa3-8bbe-b15947813a48">

I try to enlarge the sub window size to 1000 x 1000, it looks better, but some part be hidden below the console window.
<img width="1112" alt="截圖 2023-05-22 上午8 26 31" src="https://github.com/solvcon/modmesh/assets/21517083/e99a3711-e4d9-4195-a7e0-dd54e79cc57e">

We can scroll down to see the hidden part, but I'm not sure is that a good way to adjust UI.

yungyuc:
Thanks for confirming, @j8xixo12 .  We need to do a better job for the plotting, before we can do anything serious for mesh.  Let's discuss more.

yungyuc:
I think two things need to be improved:
1. Visualization should look good in both sub-window and (standalone) popped-out window.
2. The main window size should be set to accommodate the sub-window.

j8xixo12:
I will do the following items to improve the euler1d app UI layout:
- [x] Using UI to configure solver
- [x] Add simulation control into UI
[...]

PR #242, opened 2023-10-20
euler1d app gui enhancement

j8xixo12:
There are 2 major changes in this commit:
1. Set minimum windows size to prevent the plot from looking distorted in a small window.
2. Introduced new layout and UI in Single plot mode, user can use checkboxes to select which parameter they want to plot, there is a limitation which allows 3 paramters can be plotted simultaneously to prevent the plot from looking too crowded.

The grid layout snapshot:
截圖 2023-10-20 上午8 14 09

The single layout snapsh:
截圖 2023-10-20 上午8 13 57
ot:

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -36,10 +36,11 @@ import numpy as np
 import matplotlib
 import matplotlib.pyplot
 from matplotlib.backends.backend_qtagg import FigureCanvas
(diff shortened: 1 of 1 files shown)


yungyuc (approved):
@j8xixo12 you rock.  This is a good start.  I will file new issues to follow up with subsequent enhancements.

yungyuc:
This improves #212

PR #244, opened 2023-10-20
Add custom camera controller

tychuang1211:
As mentioned in issue #70, the current camera control needs improvement. This pull request aims to create our own camera controller class derived from `QAbstractCameraController` to better suit our requirements. At this point, the code introduces a new camera controller to the menu, which functions in the same way as the `RFirstPersonCameraController`. Further discussion is needed to determinte which functions should be added to the controller.


cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -60,6 +60,45 @@ class ROrbitCameraController
     using Qt3DExtras::QOrbitCameraController::QOrbitCameraController;
 }; /* end class ROrbitCameraController */
 
+class RCustomCameraController
```
yungyuc, line 63: Even though this is prototype code, we may use a descriptive name for the class.  What name would be informative?

yungyuc (commented):
It looks nice and it is time for us to make tests for the UI code.  Do you have an idea for how to write Python unit tests for it?

tychuang1211:
> It looks nice and it is time for us to make tests for the UI code. Do you have an idea for how to write Python unit tests for it?

I've never wrote unit tests for UI, maybe try [pytest-qt](https://pypi.org/project/pytest-qt/)?

cpp/modmesh/view/R3DWidget.hpp
tychuang1211: hmm, I don't know, I think it depends on the behavior of the camera? like orbital or FPS

yungyuc: How would you like the camera to behave (move)?

yungyuc:
> I've never wrote unit tests for UI, maybe try [pytest-qt](https://pypi.org/project/pytest-qt/)?

If it does not take a lot of time (like more than 4 hours), it is a good idea to give it a try.

tigercosmos:
Close due to lack activity.

PR #248, opened 2023-11-05
[hotfix] reopen brew update

terrychan999:
macOS CI image version update, now ships with Python 3.12.
`brew link` and `brew unlink` does not work properly with old verion of homebrew.
The wheel of PySide6 6.5.2 from PyPI does not support Python 3.12 yet, which causes the error.

Passing job
https://github.com/solvcon/modmesh/actions/runs/6736224421/job/18311095744#step:5:13
Image Version: 20230921.1

Failing job
https://github.com/solvcon/modmesh/actions/runs/6748500321/job/18361376864#step:5:13
Image Version: 20231029.1

By reopening `brew update`, `homebrew/core` gets updated, fixes the problem.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -227,7 +227,9 @@ jobs:
 
       - name: dependency by homebrew
         run: |
-          # brew update  # No update because it is slow
+          # Reopen for runner image v20231029.1
+          # old homebrew/core cause brew link and unlink not working
+          brew update  # No update because it is slow
           # Force using python 3.10 from homebrew
           brew unlink python
           brew link --force --overwrite python@3.10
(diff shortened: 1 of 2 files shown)


yungyuc (commented):
Thanks a lot, @terrychan999 .  But there is another failure: https://github.com/solvcon/modmesh/actions/runs/6758853544/job/18372035694?pr=248 .  Could you please also take a look?

terrychan999:
> Thanks a lot, @terrychan999 . But there is another failure: https://github.com/solvcon/modmesh/actions/runs/6758853544/job/18372035694?pr=248 . Could you please also take a look?

Yes, currently working on it.

terrychan999:
It seems that the latest clang-tidy in LLVM 17 behaves differently from LLVM 16.
So I freeze the version of LLVM to 16.

Also, I've noticed that the Python version still isn't changing when using `brew link`.
However, the updated PySide6 wheel is compatible with Python 3.12, so it's working.

yungyuc:
Beautiful.  Merging.

PR #250, opened 2023-11-11
add euler1d app control button

j8xixo12:
I add euler1d app control button, in this commit I also refined solver initialization flow too.

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -88,7 +88,7 @@ class PlotManager(QWidget):
         self.setWindowTitle("Euler 1D")
         # Set minimum window size to prevent the plot from looking
         # distorted due to the window being too small
-        self.setMinimumSize(1150, 750)
+        self.setMinimumSize(1150, 700)
         self.setAcceptDrops(True)
 
         self.scroll_area = QScrollArea()
@@ -510,11 +510,35 @@ class Controller:
             self._subwin.resize(1150, 750)
 
         self.plt_mgr = PlotManager(self.shocktube)
+        self.plt_mgr.build_lines_grid_layout()
         # Ideally one would use self.addToolBar here, but it is slightly
         # incompatible between PyQt6 and other bindings, so we just add the
         # toolbar as a plain widget instead.
         layout = QVBoxLayout(self._main)
+        ctrl_button_layout = QHBoxLayout()
+
+        start_button = QPushButton()
+        start_button.setText('Start')
+        start_button.clicked.connect(self.start)
+        ctrl_button_layout.addWidget(start_button)
+
+        stop_button = QPushButton()
+        stop_button.setText('Stop')
+        stop_button.clicked.connect(self.stop)
+        ctrl_button_layout.addWidget(stop_button)
+
+        step_button = QPushButton()
+        step_button.setText('Step')
+        step_button.clicked.connect(self.step_button_cb)
+        ctrl_button_layout.addWidget(step_button)
+
+        reset_button = QPushButton()
+        reset_button.setText('Reset')
+        reset_button.clicked.connect(self.reset)
+        ctrl_button_layout.addWidget(reset_button)
+
         layout.addWidget(self.plt_mgr)
+        layout.addLayout(ctrl_button_layout)
 
         self.profiling = profiling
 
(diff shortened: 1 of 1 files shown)


modmesh/app/euler1d.py
```diff
@@ -545,14 +589,11 @@ def march_alpha2(self, steps=1):
             self.log(mm.time_registry.report())
         self.plt_mgr.update_lines()
 
-    def setup_solver(self, interval):
+    def setup_timer(self, interval):
```
j8xixo12, line 592: Since solver related flow has been removed from this function, only remaining timer related flow, therefore I changed the function name.

modmesh/app/euler1d.py
```diff
@@ -545,14 +589,11 @@ def march_alpha2(self, steps=1):
...
         """
         :param interval: milliseconds
         :return: nothing
         """
-        self.shocktube.build_field(t=0)
```
j8xixo12, line 553: These solver initial flow is redundant, they are initialized in [line 624](https://github.com/solvcon/modmesh/pull/250/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R624)

modmesh/app/euler1d.py
```diff
@@ -529,11 +553,31 @@ def start(self):
     def stop(self):
         self.timer.stop()
 
+    def step_button_cb(self):
```
j8xixo12, line 556: According to https://wiki.qt.io/How_to_Use_QPushButton
`QPushButton` clicked callback function has a boolean argument `checked`, if I bind `self.step()` directly, then `step()` get a 0 step marching, that is not expected behavior of the step button.

yungyuc: Then why this function does not need a Boolean argument?

I am fine having the function to take no arguments, as long as it works.  Just curious about the reason.

modmesh/app/euler1d.py
yungyuc: Sounds good to me.

yungyuc (approved):
I didn't test it myself.  I trust it works OK in your computer.

j8xixo12:
Reply https://github.com/solvcon/modmesh/pull/250#discussion_r1390231996

The boolean argument of clicked callback is to get the button state called `checked`, if user call `setCheckable()`.
If the button is a checkable button, then button will have 2 states : checked or not checked and the state would be rotated by  click button.

In euler1d app step button don't care the button's checked state, therefore I ignore the boolean argument!

Ref: https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QAbstractButton.html#PySide6.QtWidgets.PySide6.QtWidgets.QAbstractButton.setCheckable

yungyuc:
> Reply [#250 (comment)](https://github.com/solvcon/modmesh/pull/250#discussion_r1390231996)
> 

It's easier to keep track of the discussion if we comment in that location.

> The boolean argument of clicked callback is to get the button state called `checked`, if user call `setCheckable()`. If the button is a checkable button, then button will have 2 states : checked or not checked and the state would be rotated by click button.
> 
> In euler1d app step button don't care the button's checked state, therefore I ignore the boolean argument!
> 
> Ref: https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QAbstractButton.html#PySide6.QtWidgets.PySide6.QtWidgets.QAbstractButton.setCheckable

Your code does not have the argument while Qt asks you to have one, and it is confusing.  It seems that your code depends on undefined behavior of Qt/PySide.  A better way is to add the Boolean argument to your callback function and leave a comment to say it is explicitly ignored.  Then a maintainer will not need to consult the Qt manual because the call is self explanatory.

I suggest you to use a new PR to fix it or piggyback it in your next PR.

j8xixo12:
> > Reply [#250 (comment)](https://github.com/solvcon/modmesh/pull/250#discussion_r1390231996)
> 
> It's easier to keep track of the discussion if we comment in that location.
> 
> > The boolean argument of clicked callback is to get the button state called `checked`, if user call `setCheckable()`. If the button is a checkable button, then button will have 2 states : checked or not checked and the state would be rotated by click button.
> > In euler1d app step button don't care the button's checked state, therefore I ignore the boolean argument!
> > Ref: https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QAbstractButton.html#PySide6.QtWidgets.PySide6.QtWidgets.QAbstractButton.setCheckable
> 
> Your code does not have the argument while Qt asks you to have one, and it is confusing. It seems that your code depends on undefined behavior of Qt/PySide. A better way is to add the Boolean argument to your callback function and leave a comment to say it is explicitly ignored. Then a maintainer will not need to consult the Qt manual because the call is self explanatory.
> 
> I suggest you to use a new PR to fix it or piggyback it in your next PR.

According to [PySide6 Signal and Slots](https://doc.qt.io/qtforpython-6/tutorials/basictutorial/signals_and_slots.html), `Signal` can be connected to a free function, meanwhile I also checked [Qt Signal and Slots](https://doc.qt.io/qt-6/signalsandslots.html), I have a conclusion that `Signal` can be connected to `Slot` which has less argument signature then `Signal`, even in C++, and it's working.

So I think we don't need to worry about undefined behavior issue, but for good maintainability and readability I will file a pr to add the Boolean argument.

yungyuc:
Thanks, @j8xixo12 .  Then I agree it's not about undefined behavior but maintainability.  Your approach looks great.

PR #251, opened 2023-11-11
move Parameters QAction into file menu

buganini:
move Parameters QAction into file menu: macOS doesn't allow QAction to be directly under QMenuBar


yungyuc (approved):
Thanks!

PR #252, opened 2023-11-13
Refactor the experimental RadixTree code

yungyuc:
* Fix the function and variable naming style to use snake_case from the wrong camelCase.
* Use single-name accessor naming style when possible.
* Add proper type aliases and static assertion.
* Remove dead code.
* Move RadixTreeNode::info() to testing code because the function is only used in testing.
* Use namespace (alias mm = modmesh) in C++ testing code, which is considered API reference.


cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
-private:
-    using ChildList = std::list<std::unique_ptr<RadixTreeNode>>;
-
 public:
-    RadixTreeNode()
-        : prev(nullptr){};
-    RadixTreeNode(std::string name, int key)
-        : m_name(std::move(name))
+
+    using child_list_type = std::list<std::unique_ptr<RadixTreeNode<T>>>;
```
yungyuc, line 73: Use snake_case for internal type names.  Also add back the explicit `<T>` in `RadixTreeNode<T>`.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
+    using key_type = int32_t;
```
yungyuc, line 74: Add `key_type` so that in the future we can widen the index key type.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
+
+    static_assert(std::is_integral_v<key_type> && std::is_signed_v<key_type>, "signed integral required");
```
yungyuc, line 76: The key type needs to be a signed integer.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
+        , m_prev(nullptr)
     {
     }
 
-    const int & getKey() const { return m_key; }
-    const std::string & getName() const { return m_name; }
-    T & getData() { return m_data; }
-    const T & getData() const { return m_data; }
-    const ChildList & getChildren() const { return m_children; }
+    RadixTreeNode() = default;
```
yungyuc, line 85: Supplement all constructors.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
-    {
-        std::ostringstream oss;
-        if (nullptr != prev)
-        {
-            oss << getName() << " - ";
-            oss << getData();
-        }
-        return oss.str();
+        return m_prev;
     }
```
yungyuc: Move `Info()` to the test file because it is only used for testing.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
 
 private:
-    int m_key = -1;
+    key_type m_key = -1;
     std::string m_name;
     T m_data;
-    ChildList m_children;
-    RadixTreeNode<T> * prev;
+    child_list_type m_children;
+    RadixTreeNode<T> * m_prev = nullptr;
```
yungyuc, line 120: Use default value for `m_prev`.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
-    const T & getData() const { return m_data; }
-    const ChildList & getChildren() const { return m_children; }
+    RadixTreeNode() = default;
+    RadixTreeNode(RadixTreeNode const &) = default;
+    RadixTreeNode(RadixTreeNode &&) = default;
+    RadixTreeNode & operator=(RadixTreeNode const &) = default;
+    RadixTreeNode & operator=(RadixTreeNode &&) = default;
+    ~RadixTreeNode() = default;
+
+    key_type key() const { return m_key; }
```
yungyuc, line 92: Use single-name accessor style.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
+    const std::string & name() const { return m_name; }
+    T & data() { return m_data; }
+    const T & data() const { return m_data; }
+    const child_list_type & children() const { return m_children; }
 
-    // Add a child node
-    RadixTreeNode<T> * addChild(std::string childName, int childKey)
+    RadixTreeNode<T> * add_child(std::string const & name, key_type key)
```
yungyuc, line 98: Member functions use snake_case naming style, as other code in modmesh does.

gtests/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ add_executable(
     test_nopython_buffer.cpp
     test_nopython_modmesh.cpp
     test_nopython_inout.cpp
-    test_nonpython_radixtree.cpp
+    test_nopython_radixtree.cpp
```
yungyuc, line 25: Fix the file name.

gtests/test_nopython_radixtree.cpp
```diff
@@ -0,0 +1,78 @@
+#include <modmesh/toggle/RadixTree.hpp>
+#include <gtest/gtest.h>
+
+#ifdef Py_PYTHON_H
+#error "Python.h should not be included."
+#endif
+
+template <typename T>
+std::string get_info(modmesh::RadixTreeNode<T> & r)
```
yungyuc, line 9: Helper for printing the contents.

gtests/test_nopython_radixtree.cpp
```diff
@@ -0,0 +1,78 @@
...
+    if (nullptr != r.get_prev())
+    {
+        oss << r.name() << " - " << r.data();
+    }
+    return oss.str();
+}
+
+TEST(RadixTree, construction)
+{
+    namespace mm = modmesh;
```
yungyuc, line 21: The tests also work as examples for C++ API.  We do not want users to use the library by using the namespace.  Thus, use `mm = modmesh` alias here.

yungyuc (commented):
@q40603 , please take a look at the refactor.  Let me know what you think by adding global and inline comments.

cpp/modmesh/toggle/RadixTree.hpp
q40603: Thanks for helping adding back explicit T, which I fully forgot.
But will there be any negative scenarios for aliasing without passing template parameter in this case?

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -68,66 +68,59 @@ class TimedEntry
...
     {
         auto it = std::find_if(m_children.begin(), m_children.end(), [&](const auto & child)
-                               { return child->getKey() == key; });
+                               { return child->key() == key; });
         return (it != m_children.end()) ? it->get() : nullptr;
     }
 
-    // Get prev node
-    RadixTreeNode<T> * getPrev() const
+    RadixTreeNode<T> * get_prev() const
```
q40603: Is single-name accessor style also applicable to this?

gtests/test_nopython_radixtree.cpp
q40603: But if the tests should work like example, how can the user know what class members to be printed.
Also, when developing, should debug printing function also be added into unit tests?

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: It reads strange.  Semantically it's the same with or without the template argument, IIRC.  But one needs to consult the standard to be sure which is right.  It hurts productivity.  When in doubt, choose the least surprising way to write it.

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: I haven't decided, so leave it as a getter.  `get_prev` and `get_child` look like a group so that I keep them in the same style.  And I am not yet sure about whether a pointer or a reference should be returned.  It makes more sense to return a reference with the single-name style.

Let's keep it as is for a while.

gtests/test_nopython_radixtree.cpp
yungyuc: Keep in mind that the RadixTree is still a prototype.  It does not work for the profiling and stack printing yet.  It does not have a Python wrapper yet.  The usability cannot be defined yet.  It is too early to evaluate how users will use the information.  Because we do not have the information or experience for this feature, we do not add fragile code.

Adding fragile code in the tests is less intrusive, so we are doing it.

PR #253, opened 2023-11-16
Button cb enhance

j8xixo12:
I have modified push button clicked callback function to fit the function signature.

I also add `Slot` decorator to each button callback function, because of without `Slot` decorator may cause worse performance and induce bugs.

Here is the snippets of PySide6 document:
> We recommend marking all methods used by signal connections with a @QtCore.Slot() decorator. Not doing causes run-time overhead due to the method being added to the QMetaObject when creating the connection. This is particularly important for QObject classes registered with QML, where missing decorators can introduce bugs.

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -147,6 +147,13 @@ class PlotManager(QWidget):
 
     @Slot(bool)
     def save_all(self, checked=False):
+        """
+        This callback function don't care button's checked state,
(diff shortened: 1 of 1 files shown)


yungyuc:
Thanks, @j8xixo12 .  Could you please attach some screenshots to assist review?

yungyuc (approved):
Code looks good but it will be more informative to have screenshots.

PR #254, opened 2023-11-26
[CI] Force mac use Python 3.11 from homebrew

yungyuc:
The macos CI fails at linking Python 3.10, which is not available in the home brew installation anymore (https://github.com/solvcon/modmesh/actions/runs/6990491463/job/19020046164#step:4:127):

```
You have 27 outdated formulae and 1 outdated cask installed.
You can upgrade them with brew upgrade
or list them with brew outdated.
Unlinking /usr/local/Cellar/python@3.11/3.11.6_1... 13 symlinks removed.
Error: No such keg: /usr/local/Cellar/python@3.10
Error: Process completed with exit code 1.
```


yungyuc:
@terrychan999 FYI

PR #259, opened 2023-12-02
Implement CallProfiler 

tigercosmos:
This work is based on the previous #245, and the discussion on #260

This work introduces a new profiler, called `CallProfiler`, which can print the profiling result of a hierarchical caller stack.

Using the macro for profiler insertion.
```c++
void Foo()
{
    modmesh::USE_CALLPROFILER_PROFILE_THIS_FUNCTION();
    // ...
}
```

```log
Profiling Result
  foo1 - Total Time: 61 ms, Call Count: 1
    foo2 - Total Time: 54 ms, Call Count: 1
      foo3 - Total Time: 19 ms, Call Count: 1
  foo2 - Total Time: 54 ms, Call Count: 1
    foo3 - Total Time: 19 ms, Call Count: 1
  foo3 - Total Time: 38 ms, Call Count: 2
```


tigercosmos:
@q40603  would you like to help review this proposal and give some feedback?

q40603:
### What the proposal wants to solve
The new CallProfiler tends to complete the profiling utility by

- constructing the radixTree with each node holding CallerProfile that has function name, execution name, hit count. 
- using an STL stack to handle the call stack execution time calculation
- Using __Function__ macro together with the profiling probe insertion

### Comment:
The design of the radix tree is to replace the [std::map](https://github.com/solvcon/modmesh/blob/e1cea03f1e89c62a4e7d7179c1caa39555200def/cpp/modmesh/toggle/profile.hpp#L206) in the existing profiling util. A user can leverage the [TimeRegistry](https://github.com/solvcon/modmesh/blob/e1cea03f1e89c62a4e7d7179c1caa39555200def/cpp/modmesh/toggle/profile.hpp#L133) and the [entry API](https://github.com/solvcon/modmesh/blob/e1cea03f1e89c62a4e7d7179c1caa39555200def/cpp/modmesh/toggle/profile.hpp#L179) could be replaced by the [RadixTree entry API](https://github.com/solvcon/modmesh/blob/e1cea03f1e89c62a4e7d7179c1caa39555200def/cpp/modmesh/toggle/RadixTree.hpp#L139) to construct the profiling result. 

In this case, I think the STL stack could be avoided as the profiling probe is already on the stack frame.
From the performance perspective, I think memory alloc and dealloc could also be reduced during profiling.

@tigercosmos thanks for the proposal and using the radix tree.

tigercosmos:
> In this case, I think the STL stack could be avoided as the profiling probe is already on the stack frame.

I thought the current APIs not be able to get a hierarchical caller stack. Could you demonstrate how to get the same output as the example above with current APIs?

> From the performance perspective, I think memory alloc and dealloc could also be reduced during profiling.

It's a good point. I think the overhead is ignorable at least when the function call takes a few milliseconds.

q40603:
@tigercosmos I just created another [PR](https://github.com/solvcon/modmesh/pull/260) for the demo.
You may refer to [profileRadixTree.hpp](https://github.com/solvcon/modmesh/pull/260/files#diff-d07559b067b62b574afc233d090c44efed169a7170915b9b6513479aa961a0c6) for more details

yungyuc:
> > From the performance perspective, I think memory alloc and dealloc could also be reduced during profiling.
> 
> It's a good point. I think the overhead is ignorable at least when the function call takes a few milliseconds.

No it's not.  It's very significant when a function is called a million times in a tight loop.  A scoped profiler should not be called in a tight loop, but when it is misused in that way, we still want to get reasonable results.

For low-level helper like the profiler, we should avoid STL containers.

tigercosmos:
@q40603  I refer to your approach and improve the code. I would like to have your review first.

cpp/modmesh/toggle/callprofiler.hpp
yungyuc, line 1: This file is not long and it seems to be OK to put the contents in RadixTree.hpp directly.

cpp/modmesh/toggle/callprofiler.hpp
```diff
@@ -0,0 +1,212 @@
...
+        call_count++;
+    }
+
+    std::chrono::high_resolution_clock::time_point start_time;
+    std::function<void()> cancel_callback;
+    std::string caller_name;
+    std::chrono::nanoseconds total_time; /// use nanoseconds to have higher precision
+    int call_count = 0;
+    bool is_running = false;
+};
```
yungyuc: Use end-marking comment `/* end struct CallProfiler */`.

cpp/modmesh/toggle/callprofiler.hpp
```diff
@@ -0,0 +1,212 @@
...
+            }
+            m_radix_tree.move_current_to_parent();
+        }
+        m_radix_tree.reset();
+    }
+
+private:
+    CallProfiler() = default;
+
+    void _print_profiling_result(const RadixTreeNode<CallerProfile> & node, const int depth, std::ostream & outstream) const
```
yungyuc: Do not prefix underscore to symbols unless you are a compiler.

cpp/modmesh/toggle/callprofiler.hpp
```diff
@@ -0,0 +1,212 @@
...
+    bool is_running = false;
+};
+
+class CallProfilerTest; // for gtest
+
+/// The profiler that profiles the hierarchical caller stack.
+class CallProfiler
+{
+public:
+    /// A singleton.
```
yungyuc: Explicitly delete constructors.  (You should delete all of them.)

cpp/modmesh/toggle/callprofiler.hpp
yungyuc: And this function is long.  Please put it outside class declaration.

cpp/modmesh/toggle/callprofiler.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+    void cancel()
+    {
+        m_cancel = true;
+    }
+
+private:
+    const char * m_caller_name;
+    bool m_cancel = false;
+    CallProfiler & m_profiler;
```
yungyuc: Because of this reference member `m_profiler` and the const member `m_caller_name`, you should delete implicitly created constructors.

gtests/test_nopython_callprofiler.cpp
```diff
@@ -0,0 +1,191 @@
...
+TEST_F(CallProfilerTest, simple_case_2)
+{
+    pProfiler->reset();
+
+    foo1();
+    foo2();
+    foo3();
+    foo3();
+
+    // Example:
```
yungyuc, line 140: I like the examples in the comments.

yungyuc (changes requested):
- [x] Use end-marking comment.
- [x] Do not prefix underscore in symbol names.
- [x] Delete constructors.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -179,5 +200,151 @@ class RadixTree
     key_type m_unique_id = 0;
 }; /* end class RadixTree */
 
+// The profiling result of the caller
+struct CallerProfile
+{
+    CallerProfile()
+        : total_time(0)
```
yungyuc: Why don't you give it a default value and avoid this custom default constructor?  By the rule of 5, if you customize one of the 5, you should explicitly spell out all of the 5.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -179,5 +200,151 @@ class RadixTree
...
+
+    std::chrono::high_resolution_clock::time_point start_time;
+    std::function<void()> cancel_callback;
+    std::string caller_name;
+    std::chrono::nanoseconds total_time; /// use nanoseconds to have higher precision
+    int call_count = 0;
+    bool is_running = false;
+}; /* end struct CallerProfile */
+
+class CallProfilerTest; // for gtest
```
yungyuc, line 231: Why don't you put it in gtest file?

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -179,5 +200,151 @@ class RadixTree
...
+{
+public:
+    /// A singleton.
+    static CallProfiler & instance()
+    {
+        static CallProfiler instance;
+        return instance;
+    }
+
+    CallProfiler(CallProfiler const &) = delete;
```
yungyuc, line 248: It's hard to read because the default constructor is too far away from here.  It's more readable to move the private default constructor here, e.g.,

```cpp
private:
    CallProfiler() = default;

public:
    CallProfiler(CallProfiler const &) = delete;
    //...
```

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -179,5 +200,151 @@ class RadixTree
...
+        CallerProfile & callProfile = m_radix_tree.get_current_node()->data();
+        callProfile.caller_name = caller_name;
+        callProfile.start_stopwatch();
+    }
+
+    // Called when a function ends
+    void end_caller(const std::string & caller_name)
+    {
+        CallerProfile & callProfile = m_radix_tree.get_current_node()->data();
+
```
yungyuc: My personal preference is to avoid the unnecessary blank lines or even make them one-liner, e.g.,

```cpp
callProfile.stop_stopwatch(); // Update profiling information
m_radix_tree.move_current_to_parent(); // Pop the caller from the call stack
```

I feel it is more readable than

```cpp

// Update profiling information
callProfile.stop_stopwatch();

// Pop the caller from the call stack
m_radix_tree.move_current_to_parent();
```

But the latter is fine.  I don't have a problem with it.

yungyuc (changes requested):
Thanks a lot.  The code is getting cleaner.  More issues:

- [x] `CallerProfile` constructor
- [x] `CallProfilerTest` moving to test file

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: unfortunately, I found that `std::chrono::nanoseconds` needs to explicitly assign `0` to have the correct behavior.
sure, I will add the other 4.

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: we cannot do that, otherwise it cannot build. If I put this line to gtests file, then we get:
```
  RadixTree.cpp
C:\Users\tiger\modmesh\cpp\modmesh/toggle/RadixTree.hpp(297,28): error C2433: 'modmesh::CallProfilerTest': 'friend' not permitted on data decla
rations [C:\Users\tiger\modmesh\build\cpp\modmesh\modmesh_primary.vcxproj]
C:\Users\tiger\modmesh\cpp\modmesh/toggle/RadixTree.hpp(297,28): error C4430: missing type specifier - int assumed. Note: C++ does not support
default-int [C:\Users\tiger\modmesh\build\cpp\modmesh\modmesh_primary.vcxproj]
```

cpp/modmesh/toggle/callprofiler.hpp
tigercosmos: done

cpp/modmesh/toggle/callprofiler.hpp
tigercosmos: done

cpp/modmesh/toggle/callprofiler.hpp
tigercosmos: done

cpp/modmesh/toggle/callprofiler.hpp
tigercosmos: done

cpp/modmesh/toggle/callprofiler.hpp
tigercosmos: done

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: I still set them as default, not sure if it is what you want

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: fixed.

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: sure, let me change.

yungyuc: I also prefer local variables to use `snake_case`.  But in the same way, I have not a problem with the style in the local scope.

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: Ah, for accessing private members of `CallProfiler`?  Ok, although I don't like it, it's not intolerable.

But I think this `CallProfilerTest` should be at least put in `detail` namespace to highlight that it is not meant as part of the API.

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: I meant to use the default value of a member datum:

```cpp
std::chrono::nanoseconds total_time = 0; /// use nanoseconds to have higher precision
```

You already had the following.  I do not see why `total_time` does not use a default value.

```cpp
int call_count = 0;
bool is_running = false;
```

If you use the default value of the member datum, you do not need to spell out any of the constructors.

yungyuc (changes requested):
- [x] Clarify whether to use default value of member datum or spell out constructors of `CallerProfile`
- [x] Move `CallProfilerTest` to `detail` namespace

And please do make sure all CI including linters are OK every time.

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: ops, I missed this one.

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: `std::chrono::nanoseconds total_time = 0` is invalid, so I use `std::chrono::nanoseconds total_time = std::chrono::nanoseconds(0);`

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: done

tigercosmos:
there are many unnecessary commits, please squash the commits when the PR is merged.

yungyuc:
@tigercosmos I saw you click the request for review.  Next time when you want to request for review, please leave a global comment (not inline) with a summary of what is expected to be reviewed.  The diff is not self-explaining.

tigercosmos:
@yungyuc  Sure, the PR is ready for review, please take a look.

The `CallProfiler` implementation is done and the test for it is also finished.

In addition, the aforementioned issues are all addressed.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -179,5 +201,146 @@ class RadixTree
...
+    std::string caller_name;
+    std::chrono::nanoseconds total_time = std::chrono::nanoseconds(0); /// use nanoseconds to have higher precision
+    int call_count = 0;
+    bool is_running = false;
+}; /* end struct CallerProfile */
+
+namespace detail
+{
+class CallProfilerTest; // for gtest
+}
```
yungyuc: You forgot the end marking for the namespace

yungyuc (changes requested):
- [x] End marking

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: @yungyuc  sorry, fixed.

tigercosmos:
@yungyuc  all issues are clear. Ready for review.

yungyuc (approved):
Beautiful

Issue #257, opened 2023-11-29
AddressSanitizer: heap-buffer-overflow at `modmesh::spacetime::ElementBase<modmesh::spacetime::Selm>::xneg() const`


tigercosmos:
```log
tests/test_spacetime_bad_euler.py::EulerSolverTC::test_array_getter =================================================================
==3108952==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60f000005498 at pc 0x7f922c664854 bp 0x7ffeb6bfb4b0 sp 0x7ffeb6bfb4a8
READ of size 8 at 0x60f000005498 thread T0
    #0 0x7f922c664853 in modmesh::spacetime::ElementBase<modmesh::spacetime::Selm>::xneg() const /home/mujin/modmesh/cpp/modmesh/spacetime/core.hpp:53
    #1 0x7f922c699299 in modmesh::spacetime::ElementBase<modmesh::spacetime::Selm>::dx() const /home/mujin/modmesh/cpp/modmesh/spacetime/core.hpp:52
    #2 0x7f922ca010c4 in modmesh::spacetime::BadEuler1DSolver::march_half_so0(bool) /home/mujin/modmesh/cpp/modmesh/spacetime/kernel/BadEuler1DSolver.cpp:93
    #3 0x7f922c6e0293 in void modmesh::spacetime::BadEuler1DSolver::march_half2_alpha<2ul>() /home/mujin/modmesh/cpp/modmesh/spacetime/kernel/BadEuler1DSolver.hpp:287
    #4 0x7f922c6e05a9 in void modmesh::spacetime::BadEuler1DSolver::march_alpha<2ul>(unsigned long) /home/mujin/modmesh/cpp/modmesh/spacetime/kernel/BadEuler1DSolver.hpp:298
    #5 0x7f922c693259 in modmesh::python::WrapBadEuler1DSolver::def_group_so1<2ul>()::{lambda(modmesh::spacetime::BadEuler1DSolver&, unsigned long)#4}::operator()(modmesh::spacetime::BadEuler1DSolver&, unsigned long) const /home/mujin/modmesh/cpp/modmesh/spacetime/pymod/wrap_spacetime.cpp:586
[...]

PR #261, opened 2023-12-03
By pass address sanitizer for BadEuler1DSolver member function

j8xixo12:
@tigercosmos, we can continue the discussion of #257 here.

I have added `ElementBase::neg()` and `ElementBase::xpos()` to address sanitize whitelist.

After bypass address sanity check, modmesh still got tons of memory detection.

Here is the final result snippet : `SUMMARY: AddressSanitizer: 166560 byte(s) leaked in 2070 allocation(s).`

I also attached the sanitizer log : [asan_log.txt](https://github.com/solvcon/modmesh/files/13539427/asan_log.txt)

reference : [Google sanitizer - How to turn off instrumentation](https://github.com/google/sanitizers/wiki/AddressSanitizer#turning-off-instrumentation)


cpp/modmesh/modmesh.hpp
```diff
@@ -40,4 +40,10 @@
 #include <modmesh/mesh/mesh.hpp>
 #include <modmesh/toggle/toggle.hpp>
 
+#if defined(USE_SANITIZER) && (defined(__clang__) || defined(__GNUC__))
```
j8xixo12, line 44: Considering modmesh support Windows, we need to add MSVC case here, because MSVC compiler has different way to add function or variable to whitelist.

Reference: [MSVC - Turn off instrumentation](https://learn.microsoft.com/en-us/cpp/sanitizers/asan-building?view=msvc-170#__declspecno_sanitize_address)

tigercosmos: Thanks. It's a good start!

cpp/modmesh/modmesh.hpp
```diff
@@ -40,4 +40,10 @@
...
+#define ATTRIBUTE_NO_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
+#else
+#define ATTRIBUTE_NO_SANITIZE_ADDRESS
```
tigercosmos: we can leave a TODO here for Windows.

tigercosmos:
I also noticed that there are tons of issues from asan...
I think we need to have a list first, and start to address them one by one.

j8xixo12:
> I also noticed that there are tons of issues from asan... I think we need to have a list first, and start to address them one by one.

Agree. I'll help to organize these asan issues.

If this PR is ready to merge, I think we can open an issue to track these asan issue, then this PR can be close.

@tigercosmos, @yungyuc  what do you think ?

cpp/modmesh/modmesh.hpp
j8xixo12: Thanks for the advice!
I have left a TODO comment in https://github.com/solvcon/modmesh/pull/261/commits/0132dce1315ed0f27ee31410e6a15491652ea18e

tigercosmos:
thanks. LGTM

cpp/modmesh/modmesh.hpp
```diff
@@ -40,4 +40,11 @@
 #include <modmesh/mesh/mesh.hpp>
 #include <modmesh/toggle/toggle.hpp>
 
+// TODO Add MSVC case once sanitizer can be default turned on for CI testing
+#if defined(USE_SANITIZER) && (defined(__clang__) || defined(__GNUC__))
+#define ATTRIBUTE_NO_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
```
yungyuc: What about we name this kind of macros as `ASAN_...`?

j8xixo12: Thanks for the recommendations, I have changed the marco in https://github.com/solvcon/modmesh/pull/261/commits/58ed2df0460ce2ed2230d09a6cc411a0461d373a

yungyuc:
LGTM.  Merging.

PR #266, opened 2023-12-26
add SimpleArray caster

tigercosmos:
1. Introduced new `SimpleArrayPlex` and its wrapper `WrapSimpleArrayPlex`. The interface of `SimpleArray` and `WrapSimpleArrayPlex` should be the same as `SimpleArray` and `WrapSimpleArray`, but this should be a to-do for the next PR.
2. Implemented the customized caster for `SimpleArray` to allow `SimpleArrayPlex` implicit casting.
3. add a `testhelper` module for testing

the first step for https://github.com/solvcon/modmesh/issues/27

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -445,12 +445,26 @@ jobs:
           Get-Command clang-tidy
           Get-Command flake8
 
-      - name: cmake ALL_BUILD
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=OFF DEBUG_BUILD_OPTIMIZATION_ONE=ON
         run: |
           cmake `
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
+            -DDEBUG_BUILD_OPTIMIZATION_ONE=ON `
+            -S${{ github.workspace }} `
+            -B${{ github.workspace }}/build
+          cmake --build ${{ github.workspace }}/build `
+            --config ${{ matrix.cmake_build_type }} `
+            --target ALL_BUILD
+
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=ON DEBUG_BUILD_OPTIMIZATION_ONE=ON
+        run: |
+          Remove-Item -Path ${{ github.workspace }}/build -Recurse -Force
+          cmake `
+            -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
+            -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
             -DUSE_PYTEST_HELPER_BINDING=ON `
+            -DDEBUG_BUILD_OPTIMIZATION_ONE=ON `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
           cmake --build ${{ github.workspace }}/build `
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -159,7 +159,7 @@ jobs:
           VERBOSE=1 USE_CLANG_TIDY=OFF \
           BUILD_QT=OFF \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
         make buildext VERBOSE=1
 
     - name: make pytest BUILD_QT=OFF
@@ -167,7 +167,7 @@ jobs:
         python3 -c "import modmesh; assert modmesh.HAS_VIEW == False"
         make pytest VERBOSE=1
 
-    - name: make buildext BUILD_QT=ON
+    - name: make buildext BUILD_QT=ON USE_PYTEST_HELPER_BINDING=OFF
       run: |
         rm -f build/*/Makefile
         make cmake \
@@ -177,7 +177,18 @@ jobs:
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
         make buildext VERBOSE=1
 
-    - name: make pytest BUILD_QT=ON
+    # build with pytest helper binding for testing
+    - name: make buildext BUILD_QT=ON USE_PYTEST_HELPER_BINDING=ON
+      run: |
+        rm -f build/*/Makefile
+        make cmake \
+          VERBOSE=1 USE_CLANG_TIDY=OFF \
+          BUILD_QT=ON \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
+        make buildext VERBOSE=1
+
+    - name: make pytest BUILD_QT=ON USE_PYTEST_HELPER_BINDING=ON
       run: |
         python3 -c "import modmesh; assert modmesh.HAS_VIEW == True"
         make pytest VERBOSE=1
@@ -189,7 +200,7 @@ jobs:
           VERBOSE=1 USE_CLANG_TIDY=OFF \
           BUILD_QT=ON \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
     - name: make run_viewer_pytest
       run: |
@@ -205,7 +216,7 @@ jobs:
           VERBOSE=1 USE_CLANG_TIDY=OFF \
           BUILD_QT=OFF \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_SANITIZER=OFF"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_SANITIZER=OFF -DUSE_PYTEST_HELPER_BINDING=ON"
         make buildext VERBOSE=1
         make pytest VERBOSE=1
 
@@ -292,7 +303,7 @@ jobs:
             VERBOSE=1 USE_CLANG_TIDY=OFF \
             BUILD_QT=OFF \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
           make buildext VERBOSE=1
 
       - name: make pytest BUILD_QT=OFF
@@ -303,7 +314,7 @@ jobs:
           fi
           make pytest ${JOB_MAKE_ARGS}
 
-      - name: make buildext BUILD_QT=ON
+      - name: make buildext BUILD_QT=ON USE_PYTEST_HELPER_BINDING=OFF
         run: |
           rm -f build/*/Makefile
           make cmake \
@@ -313,7 +324,18 @@ jobs:
             CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
           make buildext VERBOSE=1
 
-      - name: make pytest BUILD_QT=ON
+      # build with pytest helper binding for testing
+      - name: make buildext BUILD_QT=ON USE_PYTEST_HELPER_BINDING=ON
+        run: |
+          rm -f build/*/Makefile
+          make cmake \
+            VERBOSE=1 USE_CLANG_TIDY=OFF \
+            BUILD_QT=ON \
+            CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
+          make buildext VERBOSE=1
+
+      - name: make pytest BUILD_QT=ON USE_PYTEST_HELPER_BINDING=ON
         run: |
           # PySide6 installed by pip will bundle with a prebuilt Qt,
           # this will cause duplicated symbol.
@@ -332,7 +354,7 @@ jobs:
             VERBOSE=1 USE_CLANG_TIDY=OFF \
             BUILD_QT=ON \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
       - name: make run_viewer_pytest
         run: |
@@ -347,7 +369,7 @@ jobs:
             VERBOSE=1 USE_CLANG_TIDY=OFF \
             BUILD_QT=OFF \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_SANITIZER=OFF"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_SANITIZER=OFF -DUSE_PYTEST_HELPER_BINDING=ON"
           make buildext VERBOSE=1
           make pytest VERBOSE=1
 
@@ -428,6 +450,7 @@ jobs:
           cmake `
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
+            -DUSE_PYTEST_HELPER_BINDING=ON `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
           cmake --build ${{ github.workspace }}/build `
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -118,7 +118,7 @@ jobs:
         make viewer \
           ${JOB_MAKE_ARGS} \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
     - name: make run_viewer_pytest
       run: |
@@ -126,7 +126,7 @@ jobs:
         make run_viewer_pytest \
           ${JOB_MAKE_ARGS} \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
     - name: make flake8
       run: |
@@ -208,7 +208,7 @@ jobs:
           make viewer \
             ${JOB_MAKE_ARGS} \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
       - name: make run_viewer_pytest
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -148,7 +148,7 @@ jobs:
     strategy:
       matrix:
         # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        os: [macos-13]
+        os: [macos-13, macos-14]
         cmake_build_type: [Debug]
 
       fail-fast: false
@@ -203,12 +203,20 @@ jobs:
       - name: make cinclude (check_include)
         run: make cinclude
 
-      - name: make viewer
+      - name: make viewer USE_PYTEST_HELPER_BINDING=OFF DEBUG_BUILD_OPTIMIZATION_ONE=ON
         run: |
           make viewer \
             ${JOB_MAKE_ARGS} \
             CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON -DDEBUG_BUILD_OPTIMIZATION_ONE=ON"
+
+      - name: make viewer USE_PYTEST_HELPER_BINDING=ON DEBUG_BUILD_OPTIMIZATION_ONE=ON
+        run: |
+          rm -f build/*/Makefile
+          make viewer \
+            ${JOB_MAKE_ARGS} \
+            CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+            CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON -DDEBUG_BUILD_OPTIMIZATION_ONE=ON"
 
       - name: make run_viewer_pytest
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,10 @@ option(LINT_AS_ERRORS "clang-tidy warnings as errors" OFF)
 # - Address Sanitization: Protecting against address-related vulnerabilities. See https://clang.llvm.org/docs/AddressSanitizer.html
 option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 
+# FIXME: A workaround flag, check more on https://github.com/solvcon/modmesh/issues/283
+# For macOS and Windows development, please turn on this flag to avoid the issue.
+option(DEBUG_BUILD_OPTIMIZATION_ONE "use -O1 for debug build. This is a workaround for issue#283." OFF)
+message(STATUS "DEBUG_BUILD_OPTIMIZATION_ONE: ${DEBUG_BUILD_OPTIMIZATION_ONE}")
 
 option(USE_PYTEST_HELPER_BINDING "use helper bindings to run pytest. should be ON if the build is for pytest" OFF)
 if(USE_PYTEST_HELPER_BINDING)
@@ -89,10 +93,20 @@ message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
 if(DEBUG_SYMBOL)
     if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DEBUG")
-        add_compile_options(/O2 /Ob2)
-        string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}") # /RTC1 incompatible with /O2
+        if(DEBUG_BUILD_OPTIMIZATION_ONE)
+            # MSVC optimization flags are different from GCC/Clang, so there is no -O1 in MSVC.
+            # /O1 means minimize size and /O2 means maximun speed.
+            # We found that only /O2 can have the correct behavior, so we set /O2 here.
+            # /Ob2 is the default value of /O2, but we found that it need to explicitly set /Ob2 to have the correct behavior.
+            # See more: https://learn.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed
+            add_compile_options(/O2 /Ob2)
+            string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}") # /RTC1 incompatible with /O1
+        endif()
     else()
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O2")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
+        if(DEBUG_BUILD_OPTIMIZATION_ONE)
+            add_compile_options(-O1)
+        endif()
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,13 @@ option(LINT_AS_ERRORS "clang-tidy warnings as errors" OFF)
 option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 
 
+option(USE_PYTEST_HELPER_BINDING "use helper bindings to run pytest. should be ON if the build is for pytest" OFF)
+if(USE_PYTEST_HELPER_BINDING)
+    add_definitions(-D USE_PYTEST_HELPER_BINDING)
+endif()
+message(STATUS "USE_PYTEST_HELPER_BINDING: ${USE_PYTEST_HELPER_BINDING}")
+
+
 if(BUILD_QT)
     option(QT3D_USE_RHI "Qt use RHI" OFF)
     message(STATUS "QT3D_USE_RHI: ${QT3D_USE_RHI}")
@@ -82,8 +89,10 @@ message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
 if(DEBUG_SYMBOL)
     if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DEBUG")
+        add_compile_options(/O2 /Ob2)
+        string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}") # /RTC1 incompatible with /O2
     else()
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O2")
     endif()
 endif()
 
--- a/cpp/modmesh/CMakeLists.txt
+++ b/cpp/modmesh/CMakeLists.txt
@@ -40,6 +40,10 @@ add_subdirectory(spacetime)
 add_subdirectory(view)
 add_subdirectory(inout)
 
+if(USE_PYTEST_HELPER_BINDING)
+    add_subdirectory(testhelper)
+endif() # USE_PYTEST_HELPER_BINDING
+
 set(MODMESH_ROOT_HEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/modmesh.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/base.hpp
@@ -84,6 +88,7 @@ set(MODMESH_TERMINAL_FILES
     ${MODMESH_SPACETIME_FILES}
     ${MODMESH_PYTHON_FILES}
     ${MODMESH_INOUT_FILES}
+    ${MODMESH_TESTHELPER_FILES}
     CACHE FILEPATH "" FORCE)
 
 set(MODMESH_GRAPHIC_FILES
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -1,0 +1,542 @@
+/*
+ * Copyright (c) 2024, An-Chi Liu <phy.tiger@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/buffer/SimpleArray.hpp>
+
+namespace modmesh
+{
+
+DataType get_data_type_from_string(const std::string & data_type_string)
+{
+    if (data_type_string == "bool")
+    {
+        return DataType::Bool;
+    }
+    if (data_type_string == "int8")
+    {
+        return DataType::Int8;
+    }
+    if (data_type_string == "int16")
+    {
+        return DataType::Int16;
+    }
+    if (data_type_string == "int32")
+    {
+        return DataType::Int32;
+    }
+    if (data_type_string == "int64")
+    {
+        return DataType::Uint64;
+    }
+    if (data_type_string == "uint8")
+    {
+        return DataType::Uint8;
+    }
+    if (data_type_string == "uint16")
+    {
+        return DataType::Uint16;
+    }
+    if (data_type_string == "uint32")
+    {
+        return DataType::Uint32;
+    }
+    if (data_type_string == "uint64")
+    {
+        return DataType::Uint64;
+    }
+    if (data_type_string == "float32")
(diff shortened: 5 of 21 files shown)


cpp/modmesh/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ add_subdirectory(python)
 add_subdirectory(spacetime)
 add_subdirectory(view)
 add_subdirectory(inout)
+add_subdirectory(testhelper)
```
tigercosmos: some tests require python binding functions or classes. We can put those test helpers in this `testhelper` module in the future.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -82,13 +84,17 @@ inline size_t buffer_offset(small_vector<size_t> const & stride, small_vector<si
 namespace detail
 {
 
+using shape_type = small_vector<size_t>;
```
tigercosmos, line 85: the `shape_type` is not related to `T`, I think it's better to put it outside.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -82,13 +84,17 @@ inline size_t buffer_offset(small_vector<size_t> const & stride, small_vector<si
...
+using sshape_type = small_vector<ssize_t>;
+using buffer_type = ConcreteBuffer;
+
 template <typename T>
 struct SimpleArrayInternalTypes
 {
+    using shape_type = detail::shape_type;
```
tigercosmos, line 92: backward compatibility

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +659,86 @@ using is_simple_array = std::is_same<
 template <typename S>
 inline constexpr bool is_simple_array_v = is_simple_array<S>::value;
 
+using SimpleArrayBool = SimpleArray<bool>;
```
tigercosmos, line 659: for later macro usage in `SimpleArrayCaster.hpp`

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +659,86 @@ using is_simple_array = std::is_same<
...
+    {
+        return m_data_type;
+    }
+
+    const void * instance_ptr() const
+    {
+        return m_instance_ptr;
+    }
+
+    /// TODO: add all SimpleArray public methods
```
tigercosmos, line 731: to let the PR be simple, no other methods for now.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+// Define the Pybind11 caster for SimpleArray<T>
+namespace pybind11
+{
+namespace detail
+{
+
+#define ARRAYPLEX_TYPE_CASTER(DATATYPE)                                                                                                                       \
+    template <> /* NOLINTNEXTLINE(bugprone-macro-parentheses) */                                                                                              \
+    struct type_caster<modmesh::SimpleArray##DATATYPE> : public type_caster_base<modmesh::SimpleArray##DATATYPE>                                              \
```
tigercosmos, line 48: the most important part for casting.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -0,0 +1,84 @@
...
+        : root_base_type(mod, pyname, pydoc, pybind11::buffer_protocol())
+    {
+        (*this)
+            .def_timed(
+                pybind11::init(
+                    [](pybind11::object const & shape, std::string const & datatype)
+                    { return wrapped_type(make_shape(shape), datatype); }),
+                pybind11::arg("shape"),
+                pybind11::arg("dtype"))
+            /// TODO: should have the same interface as WrapSimpleArray
```
tigercosmos, line 57: the python module should have more methods, but keep it simple for now.

cpp/modmesh/python/module.cpp
```diff
@@ -49,6 +50,10 @@ void initialize(pybind11::module_ mod)
     initialize_spacetime(spacetime_mod);
     pybind11::module_ onedim_mod = mod.def_submodule("onedim", "onedim");
     initialize_onedim(onedim_mod);
+
+    pybind11::module_ testhelper_mod = mod.def_submodule("testhelper", "testhelper");
```
tigercosmos, line 58: the helper module for testing python binding.

.clang-tidy
```diff
@@ -16,6 +16,7 @@ Checks: >
     -cppcoreguidelines-non-private-member-variables-in-classes,
     -cppcoreguidelines-macro-usage,
     -fuchsia-overloaded-operator,-fuchsia-default-arguments*,
+    -cppcoreguidelines-pro-type-reinterpret-cast,
```
tigercosmos: there are too many `reinterpret_cast` operations due to our low-level implementation manner, so disable the warning directly.

tigercosmos:
@yungyuc  the PR should be ready for review. There are CI errors for the debug build on macos/windows; however the release works for both platforms. I am still figuring out the root cause.

.clang-tidy
yungyuc: While it is not the devastating, I would like to ask you to try suppressing this in only the file having the new `reinterpret_cast`.  I am a little bit worried by the relaxation on the free casting.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: LGTM.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -82,13 +81,17 @@ inline size_t buffer_offset(small_vector<size_t> const & stride, small_vector<si
 namespace detail
 {
 
+using shape_type = small_vector<size_t>;
+using sshape_type = small_vector<ssize_t>;
+using buffer_type = ConcreteBuffer;
```
yungyuc: Please keep the `buffer_type` alias in `SimpleArrayInternalTypes` mixin.  It's not impossible to use other buffer types in the (distant) future.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Please do not reorder them.  The diff looks ugly.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: This is not beautiful, but I can live with it.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +656,86 @@ using is_simple_array = std::is_same<
...
+    Int16,
+    Int32,
+    Int64,
+    Uint8,
+    Uint16,
+    Uint32,
+    Uint64,
+    Float32,
+    Float64,
+};
```
yungyuc: End marking.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +656,86 @@ using is_simple_array = std::is_same<
...
+    Int8,
+    Int16,
+    Int32,
+    Int64,
+    Uint8,
+    Uint16,
+    Uint32,
+    Uint64,
+    Float32,
+    Float64,
```
yungyuc, line 684: Futuristic discussion: will this enum be useful when we want to support struct arrays?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: LGTM

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +656,86 @@ using is_simple_array = std::is_same<
...
+};
+
+DataType get_data_type_from_string(const std::string & data_type);
```
yungyuc: Rename argument name `data_type` to `type_string` or something including `string`.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/buffer/buffer.hpp>
+#include <pybind11/pybind11.h>
```
yungyuc: `Python.h` always need to be included in the first place.

It's probably more intuitive to name this file as something like `common.hpp` and make other files under `pymod` to include this one in the first place.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
yungyuc: It should use a comment to highlight the importance.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
```diff
@@ -0,0 +1,96 @@
...
+SIMPLE_ARRAY_CASTER(Int16);
+SIMPLE_ARRAY_CASTER(Int32);
+SIMPLE_ARRAY_CASTER(Int64);
+SIMPLE_ARRAY_CASTER(Uint8);
+SIMPLE_ARRAY_CASTER(Uint16);
+SIMPLE_ARRAY_CASTER(Uint32);
+SIMPLE_ARRAY_CASTER(Uint64);
+SIMPLE_ARRAY_CASTER(Float32);
+SIMPLE_ARRAY_CASTER(Float64);
+} // namespace detail
```
yungyuc: End marking should starts with "end".

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -0,0 +1,498 @@
+/*
+ * Copyright (c) 2023, Yung-Yu Chen <yyc@solvcon.net>
```
yungyuc: @tigercosmos , your name for new files that include contents that you write.

cpp/modmesh/testhelper/pymod/testbuffer_pymod.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/testhelper/pymod/testbuffer_pymod.hpp> // Must be the first include.
+
+#include <modmesh/modmesh.hpp>
+#include <modmesh/python/common.hpp>
+#include <pybind11/stl.h>
+namespace modmesh
```
yungyuc, line 35: Properly use blank lines.

cpp/modmesh/testhelper/pymod/testbuffer_pymod.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <pybind11/stl.h>
+
+#include <modmesh/python/common.hpp>
+
+namespace modmesh
+{
+
+namespace python
+{
+void initialize_testbuffer(pybind11::module & mod);
```
yungyuc, line 41: Blank lines.

cpp/modmesh/testhelper/pymod/wrap_TestSimpleArrayHelper.cpp
```diff
@@ -0,0 +1,81 @@
...
+protected:
+
+    WrapTestSimpleArrayHelper(pybind11::module & mod, char const * pyname, char const * pydoc)
+        : root_base_type(mod, pyname, pydoc)
+    {
+        (*this)
+            .def("test_cast_to_arrayint32", []() -> modmesh::SimpleArrayInt32
+                 {  
+                    SimpleArrayInt32 arr(100);
+                    return arr; })
```
yungyuc: I doubt this passes clang-format.  Let me know I am correct or not.

modmesh/core.py
```diff
@@ -73,6 +74,7 @@
     'Bezier3dFp64',
     'WorldFp32',
     'WorldFp64',
+    'testhelper',
```
yungyuc, line 79: I do not like to have testing code shipped with the production code.  Please come up with a plan to make it disappear from production build before we can merge it.  Once we conclude the plan it should be put in an issue.

tests/test_buffer.py
```diff
@@ -598,6 +598,52 @@ def test_SimpleArray_broadcast_slice_ghost_md(self):
                 for k in range(4):
                     self.assertEqual(ndarr2[i, j, k], sarr[i - G, j, k])
 
+    def test_SimpleArray_casting(self):
+        # first check the caster works if the argument is exact the same type
+        array_float64 = modmesh.SimpleArrayFloat64((2, 3, 4))
+        self.assertEqual(modmesh.testhelper.TestSimpleArrayHelper.
```
yungyuc: Multi-lines like this look ugly.  Please beautify.

tests/test_buffer.py
```diff
@@ -598,6 +598,52 @@ def test_SimpleArray_broadcast_slice_ghost_md(self):
...
+                         test_load_arrayfloat64_from_arrayplex(array_float64),
+                         True)
+
+        # init arrayplex
+        arrayplex_int32 = modmesh.SimpleArray((2, 3, 4), dtype="int32")
+        arrayplex_uint64 = modmesh.SimpleArray((2, 3, 4), dtype="uint64")
+        arrayplex_float64 = modmesh.SimpleArray((2, 3, 4), dtype="float64")
+
+        # check the type is the same with different data types
```
yungyuc, line 613: Please also assert the type name string.  It will make the test code more descriptive.

tests/test_buffer.py
```diff
@@ -598,6 +598,52 @@ def test_SimpleArray_broadcast_slice_ghost_md(self):
...
+            modmesh.testhelper.TestSimpleArrayHelper \
+                .test_load_arrayfloat64_from_arrayplex(arrayplex_int32)
+
+        # explicitly check the `cast` function of the customized caster works
+        array_int32 = modmesh.SimpleArrayInt32((2, 3, 4))
+        array_int32_2 = modmesh.testhelper \
+            .TestSimpleArrayHelper \
+            .test_cast_to_arrayint32()  # SimpleArray32 from casting
+        self.assertTrue(type(array_int32) is type(array_int32_2))
+
```
yungyuc, line 665: The following does not seem to work (I am ok with it now):

```python
# Obtain a plex from typed array
array_int32 = modmesh.SimpleArrayInt32((2, 3, 4))
arrayplex_int32 = array_int32.plex

# Obtain the typed array from a plex
arrayplex_int32 = modmesh.SimpleArray((2, 3, 4), dtype='int32')
array_int32 = arrayplex_int32.typed
```

How can we make the above work?

yungyuc (changes requested):
Thanks for the great patch.  Please leave a global comment when you are ready for the next review.  Please make sure all CI runs are clean for review.

- [x] Try to relax `reinterpret_cast` on the file-basis.
- [x] Keep the `buffer_type` alias in `SimpleArrayInternalTypes` mixin.
- [x] Review and add back all end marking.
- [x] Rename argument name `data_type` to `type_string` or something including `string`.
- [x] `Python.h` always need to be included in the first place.
- [x] Use your name for the new files of your major contribution.
- [x] Make a plan to remove testing code from production build.
- [x] Also assert the type name string.

cpp/modmesh/testhelper/pymod/wrap_TestSimpleArrayHelper.cpp
tigercosmos: I don't like it as well, this is formatted by clang-format...

tigercosmos: I think I can add `// clang-format off` for this line to make it pretty.

yungyuc: Oh.  If clang-format does it, then let it be so.  It's just a small part and perhaps clang-format will correct it in the future.

tests/test_buffer.py
tigercosmos: @yungyuc  current 80 length limit of flake8 makes it difficult if any variable has a long name. 80 sounds too short for a mordern monitor, maybe consider to 120?

tests/test_buffer.py
tigercosmos: done

tests/test_buffer.py
yungyuc: It brings more pain to violate PEP-8.  We should stick with it.

tests/test_buffer.py
tigercosmos: yes, we can do it. already added in the PR. will leave explanation comments later.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
tigercosmos: added comment at the beginning of this header file.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: I think yes.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
tigercosmos: > Python.h always need to be included in the first place

copied.

>  to name this file as something like common.hpp

only the files including pybind11 functions that use `SimpleArray` as parameters and return values need to include this header file. 

In addition, since no one is using the caster in this file (actually only test code uses it now), I think we can consider renaming it in the future.

>  make other files under pymod to include this one in the first place.

not sure if you are saying putting this header file into `buffer_pymod.hpp`

if yes, since `wrap_SimpleArray.hpp`  already includes  `buffer_pymod.hpp`, and when `wrap_SimpleArray.hpp` tries to include `SimpleArrayCaster.hpp`, there will be some casting issues on `SimpleArray`.

The easiest way to avoid the issues is not putting `SimpleArrayCaster.hpp` in  `buffer_pymod.hpp`. If you still think we should still do it, maybe we can leave an issue and resolve this later.

just for reference, error log if `wrap_SimpleArray.hpp` tries to include `SimpleArrayCaster.hpp` :
```
/home/mujin/modmesh/cpp/modmesh/python/common.hpp:273:29:   required from ‘static modmesh::python::WrapBase<Wrapper, Wrapped, Holder, WrappedBase>::wrapper_type& modmesh::python::WrapBase<Wrapper, Wrapped, Holder, WrappedBase>::commit(pybind11::module&, const char*, const char*) [with Wrapper = modmesh::python::WrapSimpleArray<float>; Wrapped = modmesh::SimpleArray<float>; Holder = std::unique_ptr<modmesh::SimpleArray<float>, std::default_delete<modmesh::SimpleArray<float> > >; WrappedBase = modmesh::SimpleArray<float>; modmesh::python::WrapBase<Wrapper, Wrapped, Holder, WrappedBase>::wrapper_type = modmesh::python::WrapSimpleArray<float>; pybind11::module = pybind11::module_]’
/home/mujin/modmesh/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp:430:29:   required from here
/usr/include/pybind11/pybind11.h:249:33: error: cannot bind rvalue reference of type ‘modmesh::SimpleArrayFloat32&&’ {aka ‘modmesh::SimpleArray<float>&&’} to lvalue of type ‘modmesh::SimpleArray<float>’
  249 |                 = cast_out::cast(std::move(args_converter).template call<Return, Guard>(cap->f),
      |                   ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  250 |                                  policy,
      |                                  ~~~~~~~
  251 |                                  call.parent);
      |                                  ~~~~~~~~~~~~
```

modmesh/core.py
tigercosmos: I added a `USE_PYTEST_HELPER_BINDING` flag for this. will have explanation comments on the code.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Let's explore more for how it will be done.  Maybe in discord.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
yungyuc: We can finish this PR with the correct order of `Python.h` inclusion before making a decision for where to put this file.  We may probably get a clearer picture for how the files should be organized.

File organization is one action of refactoring and may be done separately.

tigercosmos:
- [X] Try to relax reinterpret_cast on the file-basis.
 - [X] Keep the buffer_type alias in SimpleArrayInternalTypes mixin.
 - [X] Review and add back all end marking.
- [X]  Rename argument name data_type to type_string or something including string.
- [X]  Python.h always need to be included in the first place.
- [X]  Use your name for the new files of your major contribution.
- [X]  Make a plan to remove testing code from production build.
- [X]  Also assert the type name string.

.github/workflows/devbuild.yml
```diff
@@ -159,7 +159,7 @@ jobs:
           VERBOSE=1 USE_CLANG_TIDY=OFF \
           BUILD_QT=OFF \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
```
tigercosmos, line 162: `DUSE_PYTEST_HELPER_BINDING` flag is used to build the pytest helper.

.github/workflows/devbuild.yml
```diff
@@ -177,6 +177,17 @@ jobs:
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
         make buildext VERBOSE=1
 
+    # build with pytest helper binding for testing
```
tigercosmos, line 180: we first build the version without the helper, then build again with the helper, since later on we will run pytest.

.github/workflows/devbuild.yml
tigercosmos: most of the time we need this flag, since almost all builds are used for testing.

CMakeLists.txt
```diff
@@ -82,8 +89,10 @@ message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
 if(DEBUG_SYMBOL)
     if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DEBUG")
+        add_compile_options(/O2 /Ob2)
```
tigercosmos: unfortunately, pybind11 has bug for -O0 on mac/windows, customized caster will not work.
explicitly set -O2 for debug build.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: the order should be back.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: sure, let's put this topic in other issues.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
```diff
@@ -0,0 +1,104 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <pybind11/pybind11.h>
+
+#include <modmesh/buffer/buffer.hpp>
+
+/*
+ * The purpose of including this header is to facilitate implicit casting of
```
tigercosmos, line 34: add the comment to point out the purpose of the header.

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 /*
- * Copyright (c) 2019, Yung-Yu Chen <yyc@solvcon.net>
+ * Copyright (c) 2024, An-Chi Liu <phy.tiger@gmail.com>
```
tigercosmos, line 4: the author of the file is also me, so just changed the name at this time as well.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -27,8 +27,9 @@
  */
 
 #include <modmesh/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
-#include <modmesh/buffer/pymod/TypeBroadcast.hpp>
+
```
tigercosmos, line 30: a space here can prevent clang-format to reorder the includes.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -156,6 +157,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def_property_readonly("has_ghost", &wrapped_type::has_ghost)
             .def_property("nghost", &wrapped_type::nghost, &wrapped_type::set_nghost)
             .def_property_readonly("nbody", &wrapped_type::nbody)
+            .def_property_readonly("plex", &get_arrayplex)
```
tigercosmos: this is for
```py
array_plex = array_typed.plex
```

cpp/modmesh/testhelper/pymod/wrap_TestSimpleArrayHelper.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+protected:
+
+    WrapTestSimpleArrayHelper(pybind11::module & mod, char const * pyname, char const * pydoc)
+        : root_base_type(mod, pyname, pydoc)
+    {
+        (*this)
+            .def("test_cast_to_arrayint32", []() -> modmesh::SimpleArrayInt32
+                 {
+                     // clang-format off
```
tigercosmos, line 60: clang-format is stupid here, added this to make the code pretty

modmesh/core.py
```diff
@@ -73,6 +74,9 @@
     'Bezier3dFp64',
     'WorldFp32',
     'WorldFp64',
+
+    # TODO: have a way to enable this according to `USE_PYTEST_HELPER_BINDING`
```
tigercosmos, line 78: didn't find out a good way to enable `testhelper` in python when `USE_PYTEST_HELPER_BINDING` is set on C++.
there are some complicated ways, such as doing file editing in CMake, but I think it isn't worth it.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -0,0 +1,152 @@
...
+        : root_base_type(mod, pyname, pydoc, pybind11::buffer_protocol())
+    {
+        (*this)
+            .def_timed(
+                pybind11::init(
+                    [](pybind11::object const & shape, std::string const & datatype)
+                    { return wrapped_type(make_shape(shape), datatype); }),
+                pybind11::arg("shape"),
+                pybind11::arg("dtype"))
+            .def_property_readonly("typed", &get_typed_array)
```
tigercosmos, line 56: this is for `array_typed = array_plex.typed`

tigercosmos:
@yungyuc  the PR is ready for review. all aforementioned issues are addressed. There are also comments for more details.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: fixed.

yungyuc:
> @yungyuc the PR is ready for review. all aforementioned issues are addressed. There are also comments for more details.

Thanks, @tigercosmos .  You added a lot of changesets and it is a little bit confusing to me.  If you have time, could you please reorganize them into fewer commits and rebase?

tigercosmos:
@yungyuc  the old commits are like spaghetti, not easy to rebase them into a few commits, so I just rebase them into a single commit. hope this will help you.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
yungyuc: Please follow the doxygen format `/** * */`

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+SimpleArrayPlex::SimpleArrayPlex(SimpleArrayPlex const & other)
+    : m_data_type(other.m_data_type)
+{
+    if (!other.m_instance_ptr)
+    {
+        return; // other does not have instance
+    }
+
+    m_has_instance_ownership = true;
```
yungyuc, line 236: Use initialization list

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -40,7 +40,6 @@ typedef SSIZE_T ssize_t;
 
 namespace modmesh
 {
-
```
yungyuc: If change of blank lines is not by clang-format, please revert the unnecessary change.

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
yungyuc: Thanks.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Good to know

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Please use a lambda for `plex`.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: This is great

tests/test_buffer.py
yungyuc: Thanks for checking the type strings

.github/workflows/devbuild.yml
yungyuc, line 1: Is there a build action without `USE_PYTEST_HELPER_BINDING=ON`?

CMakeLists.txt
yungyuc: What is the pybind11 bug with `O0` and debug build?

Setting `O2` may be a workaround but it may not be merged without describing the root cause clearly.  A clear description must be given when the issue is found, or it will be overly difficult for reproduction later.

yungyuc (changes requested):
The updated patch is very good!  We need to pin-point the pybind11 issue discovered.

- [ ] **IMPORTANT** Clarify why the custom pybind11 caster does not work with `O0` and debug build with mac/windows
- [ ] Use doxygen format for file header comments.
- [ ] Use initialization list for `SimpleArrayPlex::m_has_instance_ownership`.
- [ ] Please use a lambda to wrap `plex`.
- [ ] Make sure there is a build action without `USE_PYTEST_HELPER_BINDING=ON`

tigercosmos:
here is a short report for early studying of the behavoiur of different `-O` levels

I inserted some stub by using asm like `asm("brk #0x99")`, to check if `pybind11::detail::type_caster<modmesh::SimpleArrayInt32>::load()`(in `SimpleArrayCaster.hpp`) is used, and use `objdump -D -S -g -m  _modmesh.cpython-311-darwin.so` to dump the assembly code.

for example, we can see `asm("brk #0x99")` in `-O0` build assembly code. 

```asm
__ZN8pybind116detail11type_casterIN7modmesh11SimpleArrayIiEEvE4loadENS_6handleEb:
    345bbc:	ff 43 02 d1	sub	sp, sp, #144
    345bc0:	fd 7b 08 a9	stp	x29, x30, [sp, #128]
    345bc4:	fd 03 02 91	add	x29, sp, #128
    345bc8:	a1 03 1f f8	stur	x1, [x29, #-16]
    345bcc:	a0 83 1e f8	stur	x0, [x29, #-24]
    345bd0:	28 00 80 52	mov	w8, #1
    345bd4:	48 00 08 0a	and	w8, w2, w8
    345bd8:	a8 73 1e 38	sturb	w8, [x29, #-25]
    345bdc:	a0 83 5e f8	ldur	x0, [x29, #-24]
    345be0:	e0 0f 00 f9	str	x0, [sp, #24]
    345be4:	20 13 20 d4	brk	#0x99
```

## -O0 build

`asm("brk #0x99")` is found in the following symbol (called S1 later), but it is never used in `-g0` build code.

```assembly
__ZN8pybind116detail11type_casterIN7modmesh11SimpleArrayIiEEvE4loadENS_6handleEb
```
```cpp
pybind11::detail::type_caster<modmesh::SimpleArray<int>, void>::load(pybind11::handle, bool))
```

also, this S1 symbol doesn't exis in `-O1`, `-O2`, `-O3` build

## -O1, -O2, -O3 build

on the other hand, for `-O1`, `-O2`, `-O3` build, `asm("brk #0x99")` are all found in the following symbol (called S2 later).

```assembly
__ZZN8pybind1112cpp_function10initializeIRPFbRN7modmesh11SimpleArrayIiEEEbJS5_EJNS_4nameENS_9is_methodENS_7siblingEEEEvOT_PFT0_DpT1_EDpRKT2_ENUlRNS_6detail13function_callEE_8__invokeESP_
```
```cpp
void pybind11::cpp_function::initialize<bool (*&)(modmesh::SimpleArray<int>&), bool, modmesh::SimpleArray<int>&, pybind11::name, pybind11::is_method, pybind11::sibling>(bool (*&)(modmesh::SimpleArray<int>&), bool (*)(modmesh::SimpleArray<int>&), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&)::'lambda'(pybind11::detail::function_call&)::__invoke(pybind11::detail::function_call&)
````

## Summary

The S2 symbol is the entrance of casting `SimpleArray`, and for `-O1`, `-O2`, `-O3` build, they all have `asm("brk #0x99")` in S2.

For `-O0` build, the S2 symbol also exists, but it never calls S1 (where the `asm("brk #0x99")` exists), and that's why the casting failed.

Somehow the compiler just ignores the customized caster template in `-O0` build, but the reason is still unknown.

@yungyuc  I would like to have your advice for further investigation.

**Comment by yungyuc** [comment]
I am guessing `-O1 -O2 -O3` inlines `load` while `-O0` does not.  Perhaps you can try [forcing clang](https://stackoverflow.com/a/25623448) to inline it and see what may happen.

**Comment by tigercosmos** [comment]
@yungyuc  very interesting, if I force inline for `pybind11::detail::type_caster<modmesh::SimpleArray<int>, void>::load(pybind11::handle, bool))`, the function is not complied. I cannot find `#0x99` in the assembly in `-O0` build.

**Comment by tigercosmos** [comment]
could be a related issue: https://github.com/pybind/pybind11/issues/1153

**Comment by tigercosmos** [comment]
also a possible related issue: https://github.com/pybind/pybind11/issues/2763

**Comment by yungyuc** [comment]
> @yungyuc very interesting, if I force inline for `pybind11::detail::type_caster<modmesh::SimpleArray<int>, void>::load(pybind11::handle, bool))`, the function is not complied. I cannot find `#0x99` in the assembly in `-O0` build.

By "not compiled", do you mean it is totally optimized away?  Please try tricks like https://programfan.github.io/blog/2015/04/27/prevent-gcc-optimize-away-code/ .  We need the asm showing up to know what's going on.

**Comment by tigercosmos** [comment]
@yungyuc  with using asm volatile trick:
- `inline __attribute__((always_inline)) bool load(pybind11::handle src, bool convert)`, I cannot find asm code in the compiled file
- `inline bool load(pybind11::handle src, bool convert)`, the asm is compiled, but the result is the same as the above report, and the function is never used.

the asm code is like this:
```
        inline bool load(pybind11::handle src, bool convert)                                                                                                         \
        {                                      
            asm volatile("mov w8, #0");
            asm volatile("" : : : "memory");
            asm("brk #0x99");       
 ```

tigercosmos:
aslo tried to use latest LLVM 17 toolchain to build ... the result is the same

yungyuc:
Perhaps [compiler explorer](https://godbolt.org) may help you find a way to insert the asm code.

tigercosmos:
@yungyuc please let me clarify again. By using asm volatile trick, I confirmed the following result in both modmesh, and a [simple code on compiler explorer](https://godbolt.org/z/dj68Tj668).

- `inline __attribute__((always_inline))  foo()`: by using this syntax for inline under `-O0`, the code needs to be called somewhere, otherwise the compiler will do nothing for this `foo`. Therefore, if I set `inline __attribute__((always_inline)) ` for the customized caster `load`, we cannot find the inserted asm code due to the function is never used.
- `inline foo()`: by using this syntax for inline  `-O0`, seems the compiler will just ignore the `inline` keyword and still compile `foo` as a function. In modmesh case, the customized caster `inline load` is compiled as a function but the function is never called by others.

To sum up, the compiler decides not to use the pybind11 customized caster under `-O0`  (it decides to take the generic caster for unknown reasons), and it doesn't matter whether the `load` function is inline, force inline or not inline.

yungyuc:
Can we say that, when we use LLVM, the custom caster is called only when it is inlined?  If that statement is (or may be) true, it may be used to create a smaller test case for studying.

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: @yungyuc  in this case is `m_has_instance_ownership` already in default member initializer as `false`, here 236 line changes to `true` if the previous if statement is true.  Or, do you mean I should use  initialization list to declare `m_has_instance_ownership` as false and not to use default member initializer?

yungyuc: Oh!  I overlooked it.  In this case we should not move this to initialization list.  Please keep it here.  Thanks for pointing it out.

cpp/modmesh/buffer/pymod/SimpleArrayCaster.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: fixed.

.github/workflows/devbuild.yml
tigercosmos: yes, please check `USE_PYTEST_HELPER_BINDING=OFF` case in CI

tigercosmos: generally, I made a build without the helper first, and then had a build with the helper for pytest.

tigercosmos:
- [x] IMPORTANT Clarify why the custom pybind11 caster does not work with O0 and debug build with mac/windows
 - [x] Use doxygen format for file header comments.
  - [x] ~~Use initialization list for SimpleArrayPlex::m_has_instance_ownership.~~ (no need)
  - [x] Please use a lambda to wrap plex.
  - [x] Make sure there is a build action without USE_PYTEST_HELPER_BINDING=ON
  - [X] Add a cmake switch to control optimization level for both debug and release build (`DEBUG_BUILD_OPTIMIZATION_ONE`)
 - [X] In Github Actions turn O1 for the problematic compiler setup
 - [X] Draft a Github issue to track the compiler issue and file it on GitHub (https://github.com/solvcon/modmesh/issues/283)

.github/workflows/devbuild.yml
```diff
@@ -423,11 +445,26 @@ jobs:
           Get-Command clang-tidy
           Get-Command flake8
 
-      - name: cmake ALL_BUILD
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=OFF DEBUG_BUILD_OPTIMIZATION_ONE=ON
```
tigercosmos, line 448: Here's an instance where `USE_PYTEST_HELPER_BINDING` is set to `OFF`. While there are additional scenarios, I'm highlighting one for reference.

`DEBUG_BUILD_OPTIMIZATION_ONE` is `ON` for both the debug build and the release build, but it only works for the debug build. I don't think we need to have a special condition branch to separate the two builds.

.github/workflows/lint.yml
```diff
@@ -148,7 +148,7 @@ jobs:
     strategy:
       matrix:
         # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        os: [macos-13]
+        os: [macos-13, macos-14]
```
tigercosmos, line 151: add missing `macos-14`

CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ option(LINT_AS_ERRORS "clang-tidy warnings as errors" OFF)
 # - Address Sanitization: Protecting against address-related vulnerabilities. See https://clang.llvm.org/docs/AddressSanitizer.html
 option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 
+# FIXME: A workaround flag, check more on https://github.com/solvcon/modmesh/issues/283
```
tigercosmos, line 34: the issue https://github.com/solvcon/modmesh/issues/283 is pointed out here.

CMakeLists.txt
```diff
@@ -82,8 +93,20 @@ message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
 if(DEBUG_SYMBOL)
     if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DEBUG")
+        if(DEBUG_BUILD_OPTIMIZATION_ONE)
+            # MSVC optimization flags are different from GCC/Clang, so there is no -O1 in MSVC.
+            # /O1 means minimize size and /O2 means maximun speed.
+            # We found that only /O2 can have the correct behavior, so we set /O2 here.
+            # /Ob2 is the default value of /O2, but we found that it need to explicitly set /Ob2 to have the correct behavior.
+            # See more: https://learn.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed
+            add_compile_options(/O2 /Ob2)
```
tigercosmos, line 102: a little complicated here, but please check the comment in the code.

tigercosmos:
> * [x]  IMPORTANT Clarify why the custom pybind11 caster does not work with O0 and debug build with mac/windows
> * [x]  Use doxygen format for file header comments.
> * [x]  ~Use initialization list for SimpleArrayPlex::m_has_instance_ownership.~ (no need)
> * [x]  Please use a lambda to wrap plex.
> * [x]  Make sure there is a build action without USE_PYTEST_HELPER_BINDING=ON
> * [x]  Add a cmake switch to control optimization level for both debug and release build (`DEBUG_BUILD_OPTIMIZATION_ONE`)
> * [x]  In Github Actions turn O1 for the problematic compiler setup
> * [x]  Draft a Github issue to track the compiler issue and file it on GitHub ([SimpleArray caster issue #283](https://github.com/solvcon/modmesh/issues/283))

@yungyuc All comments have been addressed. Please have a look at your convenience.

CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ option(LINT_AS_ERRORS "clang-tidy warnings as errors" OFF)
 # - Address Sanitization: Protecting against address-related vulnerabilities. See https://clang.llvm.org/docs/AddressSanitizer.html
 option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 
+# FIXME: A workaround flag, check more on https://github.com/solvcon/modmesh/issues/283
+# For macOS and Windows development, please turn on this flag to avoid the issue.
+option(DEBUG_BUILD_OPTIMIZATION_ONE "use -O1 for debug build. This is a workaround for issue#283." OFF)
```
yungyuc, line 36: This is great.

yungyuc (approved):
LGTM.  I'd like to merge it.  I guess there are some rough edges to round, but we can do it against `master`.

PR #269, opened 2024-01-27
Add example of triangular and quadrilateral mixed mesh

yungyuc:
In sample_mesh.py add `make_2dmix()` to show a 2D mesh of mixed element of triangles and quadrilateral.  A small and a larger mesh are included.

Also fix a bug: FaceBuilder::add_quadrilateral missed two lines of `++ifc` and resulted into wrong entries in fcnds.

The 2D mesh looks like:

image


yungyuc:
@j8xixo12 @Yu-Rou-Weng you may be interested in it.

PR #270, opened 2024-01-28
update minimum cmake version to 3.27 due to CMP0148

j8xixo12:
According to make official document, the CMP0148 policy was introduced in CMake version 3.27, therefore we need to set the minimum version to 3.27

Ref: https://cmake.org/cmake/help/latest/policy/CMP0148.html


yungyuc:
Thanks.  One of the ubuntu CI run fails but it seems to be an runner issue.  Merge anyway.

PR #271, opened 2024-01-28
The new euler1d app GUI layout, created by PUI

j8xixo12:
I refine euler1d application GUI layout by [PUI](https://github.com/buganini/PUI).

The latest layout contains dockable solver configure window and a plotting area which is a sub-window.

This PR still has a remaining feature that has not been implemented yet, which is the data selection in single figure layout, I will implement it in next PR.

This is the screenshot of latest GUI:
截圖 2024-01-28 晚上10 09 24


modmesh/app/new_euler1d.py
yungyuc, line 1: Why don't you directly update the existing `euler1d.py`?

yungyuc (commented):
We need some discussions for why you choose to use a new app file `new_euler1d.py` instead of updating the existing `euler1d.py`.

j8xixo12:
> We need some discussions for why you choose to use a new app file `new_euler1d.py` instead of updating the existing `euler1d.py`.

Because this UI still has a remaining feature that has not been implemented yet, which is the data selection in single figure layout.

I plan to implement it and using `new_euler1d.py` to instead of `euler1d.py`.
But I just realized how bad this plan is from the reviewer's perspective because replacing one file with another will create significant differences that are hard to review.

I think the new plan is to update the contents of `euler1d.py` with `new_euler1d.py` directly in this PR, and leave the remaining feature for the next PR.

yungyuc:
You are right it is impossible to review the `new_euler1d.py` without the diff.  When you are ready to request for the next review, please make an explicit global comment saying that.  You tend to use many checkins so that I will ignore the noisy notification from them.

Before you request for review, please properly annotate the code (diff/patch).

One more question: is there anything can be done in the old `euler1d.py` that cannot be done in your (current) `new_euler1d.py`?

j8xixo12:
> One more question: is there anything can be done in the old `euler1d.py` that cannot be done in your (current) `new_euler1d.py`?

Old `euler1d.py` allows user to select which data line is plotted in plotting area under single layout mode.
I plan to integrate this into plot area configuration window that allows user to configure data's color, maker...etc and select which data line to be plotted.
截圖 2024-01-31 晚上11 16 12

Old `euler1d.py` also has a feature that allow user to drag plot to change the layout under grid layout mode, but I think this feature can be replaced with plot configuration window, user can arrange layout on it.

yungyuc:
Thanks.  The data line switches are a good feature but not a must-have for analysis.  It can be added back after the new app layout is merged.  And I also agree that drag-n-drop reordering can be achieved by the plot configuration window.

Thus I think replacing the old `euler1d.py` with the `new_euler1d.py` is a right move.  And after it is done, we may further integrate the standalone application into the viewer system.  It will be a great move forward.  I feel excited.

modmesh/app/euler1d.py
```diff
@@ -29,101 +29,260 @@
...
+    """
     ana: matplotlib.lines.Line2D = None
     num: matplotlib.lines.Line2D = None
     axis: matplotlib.pyplot.axis = None
     name: str = ""
     unit: str = ""
 
-    def bind_axis(self, axis):
-        self.axis = axis
-
```
j8xixo12, line 75: Removing bind_axis method, assign axis directly in Line [381](https://github.com/solvcon/modmesh/pull/271/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R381) and [442](https://github.com/solvcon/modmesh/pull/271/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R442)
Just like assign `ana` and `num`.

modmesh/app/euler1d.py
```diff
@@ -29,101 +29,260 @@
...
+        :param adata: Analytical data.
+        :type adata: ndarray
+
+        :param ndata: Numerical data.
+        :type ndata: ndarray
+
+        :return: None
+        """
+        self.ana.set_ydata(adata)
+        self.num.set_ydata(ndata)
```
j8xixo12, line 97: Data lines are used for visualization, there is no need to keep an array copy, therefore using pass by reference instead of passing a copy of the array.

modmesh/app/euler1d.py
```diff
@@ -29,101 +29,260 @@
...
 
 
-class PlotManager(QWidget):
-    def __init__(self, shocktube):
-        super().__init__()
-        self.setWindowTitle("Euler 1D")
-        # Set minimum window size to prevent the plot from looking
-        # distorted due to the window being too small
-        self.setMinimumSize(1150, 700)
-        self.setAcceptDrops(True)
```
j8xixo12, line 92: Removing the native `PySide6` api and using the PUI to instead of it.

modmesh/app/euler1d.py
```diff
@@ -142,215 +301,72 @@ def __init__(self, shocktube):
...
-            if self.use_grid_layout:
-                self.build_lines_grid_layout()
-            else:
-                self.build_lines_single_plot()
-
-        self.main_layout.destroyed.connect(del_cb)
-
-        self.main_layout.update()
-        self.main_container.update()
-
```
j8xixo12, line 206: Removing the GUI layout update callback. Once the GUI needs to be refreshed, this part will be handled in the PUI DOM update.

modmesh/app/euler1d.py
```diff
@@ -142,215 +301,72 @@ def __init__(self, shocktube):
...
-            # Check if the widget is a layout
-            if isinstance(widget, QLayout):
-                self.delete_layout(widget)
-            else:
-                widget.setParent(None)
-                widget.deleteLater()
-
-        layout.update()
-        layout.setParent(None)
-        layout.deleteLater()
```
j8xixo12, line 232: Ditto.

modmesh/app/euler1d.py
```diff
@@ -142,215 +301,72 @@ def __init__(self, shocktube):
...
-            p1, p2 = (self.figure_layout.getItemPosition(i),
-                      self.figure_layout.getItemPosition(j))
-
-            self.figure_layout.addItem(self.figure_layout.takeAt(i), *p2)
-            self.figure_layout.addItem(self.figure_layout.takeAt(j), *p1)
-
-    def getWindowIndex(self, pos):
-        for i in range(self.figure_layout.count()):
-            if self.figure_layout.itemAt(i).geometry().contains(pos):
-                return i
```
j8xixo12, line 290: Removing drag and drop reordering feature, it would be reimplemented within the configuration window of the plotting area, but the implementation way may not using drag and drop.

modmesh/app/euler1d.py
```diff
@@ -442,288 +453,266 @@ def _update_single_figure(self):
...
-        return lines
 
-    def _build_checkbox(self, layout):
-        layout.setSpacing(50)
-        for name, data_line in self.data_lines.items():
-            check_box = QCheckBox(name)
-            if data_line[1]:
-                check_box.toggle()
-            check_box.clicked.connect(self._checkbox_cb)
-            layout.addWidget(check_box)
```
j8xixo12, line 475: Data line visualization selection will be moved to the configuration window of the plotting area later.
Therefore the checkbox is removed.

modmesh/app/euler1d.py
```diff
@@ -442,288 +453,266 @@ def _update_single_figure(self):
...
-        if self.use_sub:
-            # FIXME: sub window causes missing QWindow with the following
-            # error:
-            # RuntimeError:
-            # Internal C++ object (PySide6.QtGui.QWindow) already deleted.
-            # It is probably because RMainWindow is not recognized by PySide6
-            # and matplotlib.  We may consider to use composite for QMainWindow
-            # instead of inheritance.
-            self._subwin = view.mgr.addSubWindow(self._main)
-            self._subwin.resize(1150, 750)
```
j8xixo12, line 545: Since the Euler1d app is constructed with dockable widgets and sub-windows, the option of using sub-windows is meaningless.

modmesh/app/euler1d.py
```diff
@@ -442,288 +453,266 @@ def _update_single_figure(self):
...
-
-        layout.addWidget(self.plt_mgr)
-        layout.addLayout(ctrl_button_layout)
-
-        self.profiling = profiling
-
-    def show(self):
-        self._main.show()
-        if self.use_sub:
-            self._subwin.show()
```
j8xixo12, line 583: Using the `PUI` to instead of these lousy imperative code.
The GUI layout is implemented in [688](https://github.com/solvcon/modmesh/pull/271/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R688) and [645](https://github.com/solvcon/modmesh/pull/271/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R645)

modmesh/app/euler1d.py
```diff
@@ -442,288 +453,266 @@ def _update_single_figure(self):
...
-                self.checkbox_select_num -= 1
-                self.data_lines[checkbox.text()][1] = False
-
-        self._update_single_figure()
-
+        if self.max_steps and self.current_step > self.max_steps:
+            self.stop()
+            return
 
-class Controller:
```
j8xixo12, line 515: The `Controller` class was removed because it can be replaced by a configuration windows, that allows user modified the solver configuration via GUI.

yungyuc:
I would like to try the code but get

```
ModuleNotFoundError: No module named 'PUI'
```

I don't think we should use PUI as an external dependency because it is not stable yet.  @j8xixo12 could you please incorporate it into modmesh?

yungyuc (changes requested):
I would like to try the code but get

```
ModuleNotFoundError: No module named 'PUI'
```

I don't think we should use PUI as an external dependency because it is not stable yet.  @j8xixo12 could you please incorporate it into modmesh?

j8xixo12:
> I would like to try the code but get
> 
> ```
> ModuleNotFoundError: No module named 'PUI'
> ```
> 
> I don't think we should use PUI as an external dependency because it is not stable yet. @j8xixo12 could you please incorporate it into modmesh?

Does incorporating PUI into modmesh mean putting the entire PUI into modmesh, or can we add PUI as a submodule?
I'm thinking, using a submodule allows us to pin the stable commit of PUI for modmesh usage.

j8xixo12:
> > I would like to try the code but get
> > ```
> > ModuleNotFoundError: No module named 'PUI'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't think we should use PUI as an external dependency because it is not stable yet. @j8xixo12 could you please incorporate it into modmesh?
> 
> Does incorporating PUI into modmesh mean putting the entire PUI into modmesh, or can we add PUI as a submodule? I'm thinking, using a submodule allows us to pin the stable commit of PUI for modmesh usage.

@buganini Do you have any idea about how to incorporate PUI into modmesh ?

Since PUI is still under development, users may encounter problems with both modmesh and PUI simultaneously. This could result in a poor user experience and hard to debug for developer.

My suggestion is to add PUI as a submodule into modmesh. This way, modmesh can pin the stable commit of PUI for its usage. When modmesh needs additional features from PUI, we can update the submodule to the next stable commit.
With a submodule, users won't need to install PUI manually, and errors like `ModuleNotFoundError: No module named 'PUI'`should not occur again.

@yungyuc could you give me some suggestions, if I missed something important.

yungyuc:
@j8xixo12 I think we can start with git submodule pinning to a specific changeset.  We may need PUI to come up with a release scheme before we can properly use it as a dependency and manage using devenv.  Could you please help bring it to discord to discuss?

j8xixo12:
> @j8xixo12 I think we can start with git submodule pinning to a specific changeset. We may need PUI to come up with a release scheme before we can properly use it as a dependency and manage using devenv. Could you please help bring it to discord to discuss?

Ok, I will bring it to discord and have a discussion with bug.
But I guess PUI needs some time to come out the release scheme, I will pin the latest commit for now.
@yungyuc what do you think ?

yungyuc:
> Ok, I will bring it to discord and have a discussion with bug. But I guess PUI needs some time to come out the release scheme, I will pin the latest commit for now. @yungyuc what do you think ?

Let's pin to a changeset/commit at the moment and work with PUI for a release scheme.  Those are two separate things.

j8xixo12:
@yungyuc I have added PUI as a submodule into modmesh, because PUI is a thirdparty library, therefore I put it into `thirdparty` directory.
It will be preload during modmesh initialization.

modmesh/view.py
```diff
@@ -55,6 +57,29 @@
...
+            path = None
+            break
+        else:
+            path = os.path.dirname(path)
+    if path is None:
+        raise ImportError
+except ImportError:
+    sys.stderr.write('Can not find PUI in your environment.\n')
+    sys.stderr.write('Please run git submodule update --init\n')
+    sys.exit(0)
```
j8xixo12: If user forget to initialize the submodule it will raise a `ImportError` and terminate modmesh, also the error message will remind user to initialize the submodule to install the PUI.

yungyuc (changes requested):
- [x] Add an automation script under `thirdparty/` to document how to pull PUI.
- [x] Add a unit test to exercise importing PUI and make sure it basically works.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "thirdparty/PUI"]
```
yungyuc, line 1: It's good to put it under `thirdparty/` which is an existing directory.  I suggest to also add an automation script under it to document how to pull PUI.

tests/test_view.py
```diff
@@ -32,13 +32,17 @@
...
-
+import PUI.PySide6
 
 @unittest.skipUnless(modmesh.HAS_VIEW, "Qt view is not built")
 class ViewTC(unittest.TestCase):
 
     def test_import(self):
         self.assertTrue(hasattr(modmesh.view, "mgr"))
+        self.assertTrue(hasattr(PUI.PySide6, "PUINode"))
+        self.assertTrue(hasattr(PUI.PySide6, "PUIView"))
```
j8xixo12, line 51: Because `PUINode` and `PUIView` are key components in `PUI`, without these two attributes, `PUI` will not work. Therefore, I chose to check if these two attributes are imported properly.

tests/test_view.py
```diff
@@ -32,13 +32,18 @@
...
+        self.assertEqual(PUI.PySide6.PUI_BACKEND, "PySide6",
+                         "PUI backebd mismatch")
```
j8xixo12, line 53: Make sure PUI use the right backend which is `PySide6`.

modmesh/view.py
```diff
@@ -60,6 +62,28 @@ def _load():
...
+                    path = os.path.dirname(path)
+            if path is None:
+                raise ImportError
+        except ImportError:
+            sys.stderr.write('Can not find PUI in your environment.\n')
+            sys.stderr.write('Please run git submodule update --init\n')
+            sys.exit(0)
+
+        path = os.path.join(path, filename)
+        sys.path.append(path)
```
j8xixo12: If `view` is imported from `modmesh` library, that means modmesh is built with Qt. This indicates modmesh has GUI and the PUI needs to be imported otherwise PUI does not need to be imported.

CMakeLists.txt
```diff
@@ -35,6 +35,10 @@ option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 if(BUILD_QT)
     option(QT3D_USE_RHI "Qt use RHI" OFF)
     message(STATUS "QT3D_USE_RHI: ${QT3D_USE_RHI}")
+    execute_process(
+        COMMAND python3 PUI_install.py
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/thirdparty 
+    )
```
j8xixo12, line 41: If modmesh is built with Qt, then PUI also needs to be installed, therefore I let PUI to be installed automatically.

.gitmodules
j8xixo12: I add the `PUI_install.py` to pull and install PUI automatically.

modmesh/view.py
```diff
@@ -60,6 +62,30 @@ def _load():
...
+        try:
+            while True:
+                if os.path.exists(os.path.join(path, filename)):
+                    break
+                if path == os.path.dirname(path):
+                    path = None
+                    break
+                else:
+                    path = os.path.dirname(path)
+            if path is None or not os.path.exists(os.path.join(path,
```
j8xixo12, line 78: Check if the `PUI` folder in `thirdparty` contains `PUI` folder or not, if the user does not initialize the submodule, then the `PUI` in `thirdparty` will be an empty folder.

thirdparty/PUI_install.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+
+script_root = os.path.dirname(os.path.realpath(__file__))
+pui_folder = os.path.join(script_root, "PUI", "PUI")
```
j8xixo12, line 7: Using the same way in `modmesh/view.py` to check if the submodule is initialized or not.

j8xixo12:
- [x] Add an automation script under thirdparty/ to document how to pull PUI.
- [x] Add a unit test to exercise importing PUI and make sure it basically works.

@yungyuc  this PR is ready for review.

thirdparty/PUI_install.py
yungyuc, line 1: Please use a (bash) shell script.  It baffled me why you choose to use a Python script when the two other install scripts [googletest.sh](https://github.com/solvcon/modmesh/blob/master/thirdparty/googletest.sh) and [metal-cpp.sh](https://github.com/solvcon/modmesh/blob/master/thirdparty/metal-cpp.sh) in the directory exclusively uses bash.

Using bash allows the maintainers to manage the thirdparty without needing to install Python (3).  It's the same rationale of using bash for [devenv](https://github.com/solvcon/devenv).  On many server farms (or supercomputers) there may not have up-to-date Python or other dependencies installed.

tests/test_view.py
yungyuc: This is sweet.

tests/test_view.py
yungyuc: It's clear documentation.  Thanks.

yungyuc (changes requested):
Thanks, @j8xixo12 .  Please fix the following:

- [x] Use bash for the install script of PUI.

thirdparty/PUI_install.py
j8xixo12: The reason for using a Python script is that bash can't run on Windows. However, modmesh claims to support Windows. Also, our CI has a Windows version, and Python as a dependency of modmesh, user cannot run modmesh without Python. Therefore, I choose to use Python script to implement the install script.

yungyuc: I didn't notice that you made it in Python for Windows, which is reasonable.  It's good to use Python then.  Thank you for taking care of it.

modmesh/view.py
```diff
@@ -60,6 +62,30 @@ def _load():
     if enable:
         for name in _from_impl:
             globals()[name] = getattr(_vimpl, name)
+        # Try to find the PUI in thirdparty, if failed to find PUI
+        # modmesh will raise ImportError and terminate itself.
+        filename = os.path.join('thirdparty', 'PUI')
```
yungyuc, line 67: We should make the consumption of PUI more systematic in the future, so that we do not need to hard-code import path like here.  While we can live with it now, @j8xixo12 could you please create an issue to track the improvement for this?

yungyuc (approved):
Thanks @j8xixo12 for the great improvement.

yungyuc:
@j8xixo12 , I got it working once pulling and rebuilding out of the box.  Fantastic!

image

yungyuc:
This PR fixes #212

PR #275, opened 2024-02-14
[CI] enable macOS-14 runner

terrychan999:
In this PR, there are two changes:
1. Enable the macOS-14 runner.
2. In thirdparty/metal-cpp.sh, download the metal-cpp version based on the OS version.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -223,7 +223,7 @@ jobs:
     strategy:
       matrix:
         # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        os: [macos-13]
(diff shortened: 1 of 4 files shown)


yungyuc:
`tidy_flake8_macos` with macos-14 Debug is slow.  To save CI cycles (although it is free), perhaps we can turn on macos-14 with only `devbuild`.  What do you think?

terrychan999:
> `tidy_flake8_macos` with macos-14 Debug is slow. To save CI cycles (although it is free), perhaps we can turn on macos-14 with only `devbuild`. What do you think?

Sounds good, I'll make the commit.

yungyuc:
LGTM.  Thanks.

Issue #277, opened 2024-02-27
PySide6 can not be imported at macos-13 image


j8xixo12:
When the CI uses the macos-13 image, PySide6 cannot be imported successfully.

The GitHub actions log: https://github.com/solvcon/modmesh/actions/runs/7952874439/job/21708086025?pr=271#step:12:103

j8xixo12:
@terrychan999 can you help to file a PR to fix this issue ?

terrychan999:
I've just made a PR, thanks for the reminder.

j8xixo12:
Since macos-13 issue has fixed by @terrychan999. Let's close this issue!

PR #278, opened 2024-02-28
[CI] Fixing PySide6 Import Error

terrychan999:
PySide6's Qt library was removed to prevent duplicate symbols, but this resulted in PySide6 being imported incorrectly, as seen in https://github.com/solvcon/modmesh/issues/277.

We can use `install_name_tool`, an XCode utility, to add the `rpath` of Homebrew's Qt library to the existing shared library.


yungyuc:
LGTM.  Thanks a lot, @terrychan999

PR #291, opened 2024-03-12
implement constructors for SimpleArrayPlex

tigercosmos:
Support all constructors in `WrapSimpleArray` for `WrapSimpleArrayPlex` 

Now in Python, we can use the same way to construct `modmesh.SimpleArray` as `modmesh.SimpleArrayInt`

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -146,7 +146,10 @@ DataType get_data_type_from_type<double>()
     return DataType::Float64;
 }
 
-#define CREATE_SIMPLE_ARRAY(DataType, ArrayType, ...)                          \
+// According to the `DataType`, create the corresponding `SimpleArray<T>` instance
+// and assign it to `m_instance_ptr`. The `m_instance_ptr` is a void pointer, so
+// we need to use `reinterpret_cast` to convert the pointer of the array instance.
+#define MM_DECL_CREATE_SIMPLE_ARRAY(DataType, ArrayType, ...)                  \
     case DataType:                                                             \
         /* NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast) */      \
         m_instance_ptr = reinterpret_cast<void *>(new ArrayType(__VA_ARGS__)); \
@@ -158,17 +161,17 @@ SimpleArrayPlex::SimpleArrayPlex(const shape_type & shape, const DataType data_t
 {
     switch (data_type)
     {
-        CREATE_SIMPLE_ARRAY(DataType::Bool, SimpleArrayBool, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Int8, SimpleArrayInt8, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Int16, SimpleArrayInt16, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Int32, SimpleArrayInt32, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Int64, SimpleArrayInt64, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Uint8, SimpleArrayUint8, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Uint16, SimpleArrayUint16, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Uint32, SimpleArrayUint32, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Uint64, SimpleArrayUint64, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Float32, SimpleArrayFloat32, shape)
-        CREATE_SIMPLE_ARRAY(DataType::Float64, SimpleArrayFloat64, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Bool, SimpleArrayBool, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Int8, SimpleArrayInt8, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Int16, SimpleArrayInt16, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Int32, SimpleArrayInt32, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Int64, SimpleArrayInt64, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Uint8, SimpleArrayUint8, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Uint16, SimpleArrayUint16, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Uint32, SimpleArrayUint32, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Uint64, SimpleArrayUint64, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Float32, SimpleArrayFloat32, shape)
+        MM_DECL_CREATE_SIMPLE_ARRAY(DataType::Float64, SimpleArrayFloat64, shape)
     default:
         throw std::runtime_error("Unsupported datatype");
     }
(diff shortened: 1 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -146,83 +146,76 @@ DataType get_data_type_from_type<double>()
     return DataType::Float64;
 }
 
-SimpleArrayPlex::SimpleArrayPlex(const shape_type & shape, DataType data_type)
+#define CREATE_SIMPLE_ARRAY(DataType, ArrayType, ...)                          \
```
tigercosmos: reduce a lot of code, and allow us to call different constructor of `SimpleArray`

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -53,6 +53,32 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
                     { return wrapped_type(make_shape(shape), datatype); }),
                 pybind11::arg("shape"),
                 pybind11::arg("dtype"))
+            .def_timed(
+                // Note we use double instead of T to avoid the template (,which is not supported by pybind11::init)
+                // Value will later be converted to the correct datatype in the constructor
+                // This may cause the loss of precision, but should be fine to most cases
+                pybind11::init(
+                    [](pybind11::object const & shape, double const & value, std::string const & datatype)
```
tigercosmos: I decided to use `double` for `value`, please read the comment of explanation.

tests/test_buffer.py
```diff
@@ -700,4 +700,29 @@ def test_abs(self):
         sarr = sarr.abs()
         self.assertEqual(sarr.sum(), True)
 
+
+class SimpleArrayPlexTC(unittest.TestCase):
+
+    def test_SimpleArrayPlex_constructor(self):
```
tigercosmos, line 707: test cases for constructors

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -701,7 +701,19 @@ class SimpleArrayPlex
...
+    }
+
+    explicit SimpleArrayPlex(const shape_type & shape, const std::shared_ptr<ConcreteBuffer> & buffer, const std::string & data_type)
+        : SimpleArrayPlex(shape, buffer, get_data_type_from_string(data_type))
+    {
+    }
+
+    explicit SimpleArrayPlex(const shape_type & shape, const DataType data_type);
+    explicit SimpleArrayPlex(const shape_type & shape, const double & value, const DataType data_type);
+    explicit SimpleArrayPlex(const shape_type & shape, const std::shared_ptr<ConcreteBuffer> & buffer, const DataType data_type);
```
tigercosmos, line 710: these three constructors follows the python binding implementation of `SimpleArray`

tigercosmos:
@yungyuc  PR is ready for your review.

cpp/modmesh/buffer/SimpleArray.cpp
yungyuc: This is more concise, and provide you an opportunity to add comments to explain why using `reinterpret_cast`.  It will be helpful information for maintainer.

yungyuc: Please `undef` the macro once it is not used.  It will be better to prefix it with `MM_DECL`.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: Don't take `const &` for input argument of fundamental types.  It's implying the compiler to generating inefficient code.

yungyuc: But a better way is to use `py::object` for `value`, and then determine the type in the lambda to dispatch to the right constructor.

tests/test_buffer.py
yungyuc: It's good to have it.

yungyuc (changes requested):
- [x] Add comments to explain why use `reinterpret_cast`
- [x] `undef` the macro once it is not used and prefix it for clearer identification
- [x] Use `py::object` for `value` argument, or avoid taking `const &` for input argument of fundamental types

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: @yungyuc  I am not sure if using `py::object` for `value` is good or not.
here is my concern:
1. if we extract the data type at wrapper, we need to have template `T` for  `SimpleArrayPlex(const shape_type & shape, const T & value, const std::string & data_type)`, then we need to either (1) expose the implementation in the header (2) have many implementation for each data type `T`
2. if we don't extract the `py::object`, and pass it to `SimpleArrayPlex`, `SimpleArrayPlex` can pass the `py::object` to `SimpleArray`, and since `SimlpeArray<T>` already has the data type, then we can extract the `py::object` to the correct data type. However, the drawback is that we need to include pybind11 in `SimpleArray.hpp`

What do you think?

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: added comments and added `undef`

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: The new function `explicit SimpleArrayPlex(const shape_type & shape, double value, const std::string & data_type)` does not seem to be necessary.

You may use [`explicit SimpleArrayPlex(const shape_type & shape, DataType data_type)`](https://github.com/solvcon/modmesh/blob/48ace7cec150185529022673a1e438fba9bb5095/cpp/modmesh/buffer/SimpleArray.hpp#L704) followed by [`SimpleArray::fill(value_type const & value)`](https://github.com/solvcon/modmesh/blob/48ace7cec150185529022673a1e438fba9bb5095/cpp/modmesh/buffer/SimpleArray.hpp#L109) to do the same thing.

Conversion from `py::object` should happen at pybind11 wrapper level because `SimpleArrayPlex` should not know about `PyObject`.  Because `SimpleArray::fill()` does not know `py::object`, it probably saves some coding by providing a multiplexer from `py::object` to `SimpleArray::value_type` at the pybind11 layer.

tigercosmos: fixed.

tigercosmos:
@yungyuc  all addressed. Ready for review.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -53,13 +53,161 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
+                pybind11::arg("array"))
             .def_property_readonly("typed", &get_typed_array)
             /// TODO: should have the same interface as WrapSimpleArray
             ;
     }
 
-    /// return the typed function from the arrayplex
-    static pybind11::object get_typed_array(wrapped_type const & array_plex)
+    /// Initialize the arrayplex with the given value
+    /// NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
```
yungyuc: This (`NOLINT`) is not a doxygen comment.  Please use `//` instead of `///`.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -53,13 +53,161 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
+    {
+        const shape_type shape = make_shape(shape_in);
+        wrapped_type array_plex(shape, datatype);
+
+        // cast value to correct type based on the datatype
+        switch (array_plex.data_type())
+        {
+        case DataType::Bool:
+        {
+            if (!pybind11::isinstance<pybind11::bool_>(value))
```
yungyuc, line 94: The repeating error-checking code probably should use a macro to simplify.  Could you please take a look?  If it is an obviously good idea, you can try it in a later PR.

yungyuc (changes requested):
Everything looks good except a minor formatting issue.

- [x] For non-doxygen comment, use // instead of ///.  Please check everywhere in your patch again.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: For now, maybe maintaining the current state is suitable. I think that employing macros doesn't enhance readability unless we encounter excessive repetition of the same code.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: fixed.

tigercosmos:
@yungyuc  fixed the last issue. should be ready to merge.

yungyuc (approved):
LGTM.  Thanks.

Issue #284, opened 2024-03-05
Upgrade QT from 6.3 to 6.5


tigercosmos:
Our project is currently utilizing QT version 6.3, a non-LTS release and it is no longer available on the QT download page. To ensure stability and support for our application, we should migrate to the latest QT LTS version, which is [version 6.5](https://wiki.qt.io/Qt_6.5_Release).

yungyuc:
This issue is fixed by #292

Issue #289, opened 2024-03-11
Action macos-13 failed due to brew install


tigercosmos:
macos-13 failed due to brew install, see nightly build on master: https://github.com/solvcon/modmesh/actions/runs/8223688840/job/22490789414#step:4:189

```
==> Pouring double-conversion--3.3.0.ventura.bottle.tar.gz
🍺  /usr/local/Cellar/double-conversion/3.3.0: 26 files, 245.4KB
==> Installing qt dependency: python@3.12
==> Downloading https://ghcr.io/v2/homebrew/core/python/3.12/manifests/3.12.2_1
Already downloaded: /Users/runner/Library/Caches/Homebrew/downloads/db6a7027cff95a641d6d0c07e00f7b8fa26b0733b6d215c3752b[74](https://github.com/solvcon/modmesh/actions/runs/8223688840/job/22490789414#step:4:75)5a381ebcb8--python@3.12-3.12.2_1.bottle_manifest.json
==> Pouring python@3.12--3.12.2_1.ventura.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
Target /usr/local/bin/2to3
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.12

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.12
  ```

PR #292, opened 2024-03-13
[CI] add qt action to all platform

terrychan999:
In this PR, we manage Qt installation for all platforms using `jurplel/install-qt-action@v3`.
All the Qt versions are fixed to 6.5.3 LTS.
Also, `jurplel/install-qt-action@v3` has built-in `actions/setup-python@v4`, so we can avoid `brew unlink && brew link` issues like those in https://github.com/solvcon/modmesh/issues/289.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -113,19 +113,22 @@ jobs:
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ \
             python3 python3-dev python3-venv
-        # Install qt6 only with ubuntu-22.04
-        # This page explains why we need libgl1-mesa-dev
-        # https://doc-snapshots.qt.io/qt6-dev/linux.html
-        #
-        # In short, OpenGL libraries and headers are required. Without
-        # installing this package, cmake won't find out the correct lib path.
+        # # Install qt6 only with ubuntu-22.04
+        # # This page explains why we need libgl1-mesa-dev
+        # # https://doc-snapshots.qt.io/qt6-dev/linux.html
+        # #
+        # # In short, OpenGL libraries and headers are required. Without
+        # # installing this package, cmake won't find out the correct lib path.
+
         # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
         #   sudo apt-get -qy install \
         #       qt6-3d-dev xvfb \
         #       libgl1-mesa-dev
         # fi
 
-    - name: install qt
+        # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
+
+    - name: Install Qt
       uses: jurplel/install-qt-action@v3
       with:
         aqtversion: '==3.1.*'
@@ -280,6 +283,7 @@ jobs:
           export HOMEBREW_NO_AUTO_UPDATE=1
           export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
           # brew install qt6
+          # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
 
       - name: Install Qt
         uses: jurplel/install-qt-action@v3
@@ -458,7 +462,7 @@ jobs:
         run: |
           echo "github.event_name: ${{ github.event_name }}"
 
-      - name: install qt
+      - name: Install Qt
         uses: jurplel/install-qt-action@v3
         with:
           aqtversion: '==3.1.*'
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -113,20 +113,19 @@ jobs:
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ \
             python3 python3-dev python3-venv
-        # # Install qt6 only with ubuntu-22.04
-        # # This page explains why we need libgl1-mesa-dev
-        # # https://doc-snapshots.qt.io/qt6-dev/linux.html
-        # #
-        # # In short, OpenGL libraries and headers are required. Without
-        # # installing this package, cmake won't find out the correct lib path.
+        # Install qt6 only with ubuntu-22.04
+        # This page explains why we need libgl1-mesa-dev
+        # https://doc-snapshots.qt.io/qt6-dev/linux.html
+        #
+        # In short, OpenGL libraries and headers are required. Without
+        # installing this package, cmake won't find the correct lib path.
+        # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
 
         # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
         #   sudo apt-get -qy install \
         #       qt6-3d-dev xvfb \
         #       libgl1-mesa-dev
         # fi
-
-        # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
 
     - name: Install Qt
       uses: jurplel/install-qt-action@v3
@@ -282,8 +281,8 @@ jobs:
         run: |
           export HOMEBREW_NO_AUTO_UPDATE=1
           export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+          # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
           # brew install qt6
-          # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
 
       - name: Install Qt
         uses: jurplel/install-qt-action@v3
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -119,7 +119,8 @@ jobs:
         #
         # In short, OpenGL libraries and headers are required. Without
         # installing this package, cmake won't find the correct lib path.
-        # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
+        # This has been replaced by the 'install qt' section below to manage
+        # qt6 versioning independently from the OS.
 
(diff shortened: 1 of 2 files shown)


terrychan999:
I just notice that https://github.com/solvcon/modmesh/issues/289 still exist in `lint / tidy_flake8_macos` because llvm@16 depends on python 3.12 which install from homebrew...

yungyuc:
Is this PR intended for discussion or review?  When filing a PR, please write explicit words for the intention.

terrychan999:
@yungyuc this PR is now ready for review, thanks.

.github/workflows/devbuild.yml
```diff
@@ -113,11 +113,22 @@ jobs:
         #
         # In short, OpenGL libraries and headers are required. Without
         # installing this package, cmake won't find out the correct lib path.
-        if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
-          sudo apt-get -qy install \
-              qt6-3d-dev xvfb \
-              libgl1-mesa-dev
-        fi
+        # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
```
yungyuc, line 125: Please add comments to explain why the block of commands is commented out.

.github/workflows/devbuild.yml
```diff
@@ -258,7 +269,18 @@ jobs:
             # GitHub Action macOS 13 has a problem with python3
             brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done;
           fi
-          brew install qt6
+          # brew install qt6
```
yungyuc, line 288: Please add comments to explain why the command is commented out.

.github/workflows/lint.yml
```diff
@@ -78,11 +78,22 @@ jobs:
         # In short, OpenGL libraries and headers are required. Without
         # installing this package, cmake won't find out the correct lib path.
 
-        if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
-          sudo apt-get -qy install \
-              qt6-3d-dev xvfb \
-              libgl1-mesa-dev
-        fi
+        # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
```
yungyuc, line 83: Please add comments to explain why the block of commands is commented out.

yungyuc (changes requested):
Thanks, @terrychan999 .  Could you please address the following point?

- [x] Inline comment to explain the reason to disable code.

.github/workflows/devbuild.yml
```diff
@@ -258,7 +269,18 @@ jobs:
...
+
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: '==3.1.*'
```
tigercosmos, line 293: `==3.1` should be fine?

.github/workflows/devbuild.yml
```diff
@@ -107,17 +107,31 @@ jobs:
...
+        # # In short, OpenGL libraries and headers are required. Without
+        # # installing this package, cmake won't find out the correct lib path.
+
+        # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
+        #   sudo apt-get -qy install \
+        #       qt6-3d-dev xvfb \
+        #       libgl1-mesa-dev
+        # fi
+
+        # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
```
terrychan999: Added explanation for disabling here.

.github/workflows/devbuild.yml
```diff
@@ -258,7 +272,20 @@ jobs:
             # GitHub Action macOS 13 has a problem with python3
             brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done;
           fi
-          brew install qt6
+
+          # brew install qt6
+          # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
```
terrychan999: Added explanation for disabling here.

.github/workflows/lint.yml
```diff
@@ -71,18 +71,32 @@ jobs:
...
+        # # In short, OpenGL libraries and headers are required. Without
+        # # installing this package, cmake won't find out the correct lib path.
+
+        # if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
+        #   sudo apt-get -qy install \
+        #       qt6-3d-dev xvfb \
+        #       libgl1-mesa-dev
+        # fi
+
+        # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
```
terrychan999: Added explanation for disabling here.

.github/workflows/lint.yml
```diff
@@ -171,10 +185,25 @@ jobs:
             # GitHub Action macOS 13 has a problem with python3
             brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done;
           fi
-          brew install llvm@16 qt6
+
+          # brew install llvm@16 qt6
+          # # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
```
terrychan999: Added explanation for disabling here.

.github/workflows/devbuild.yml
terrychan999: this is a version specifier using in `pip install aqtinstall==<version specifier>`
https://github.com/jurplel/install-qt-action/blob/master/action/src/main.ts#L354

the version is just fellowing the current default.
https://github.com/jurplel/install-qt-action/tree/master?tab=readme-ov-file#aqtversion

tigercosmos: True, just realized it follows pip's rule.
Thanks for the explanation.

.github/workflows/devbuild.yml
yungyuc: I don't like the style of comments:

```
# # ...
```

It unnecessarily makes the lines longer.

And please write "... cmake won't find the correct ...".  It seems "find out" isn't what you meant.

.github/workflows/devbuild.yml
yungyuc: Please move the explanation above the code commented out.  Please avoid doubling the commenting mark like `# # ...`

.github/workflows/devbuild.yml
yungyuc: Explanation should go before the commented-out code.

yungyuc (changes requested):
The comments are not placed in right places.  Please review and update them all.

yungyuc:
@terrychan999 any comments?

terrychan999:
I've moved the comments to the position before the code and also removed the usage of double commenting mark.

I also added two changes:
1. Adding `PIP_BREAK_SYSTEM_PACKAGES: 1` in the env section of macOS-CI to avoid an error like the one in https://github.com/solvcon/modmesh/pull/301 with `jurplel/install-qt-action@v3`.
I used this environment variable because I couldn't directly pass the argument `--break-system-packages` into the action.

2. Changing the Qt version to `6.6.3`.
The reason for updating the Qt version to `6.6.3` is that `6.5.3 LTS` does not support Python 3.12, here's the reference:
https://wiki.qt.io/Qt_for_Python#Python_compatibility_matrix

.github/workflows/devbuild.yml
```diff
@@ -112,12 +112,25 @@ jobs:
...
         #
         # In short, OpenGL libraries and headers are required. Without
-        # installing this package, cmake won't find out the correct lib path.
-        if [ "${{ matrix.os }}" == "ubuntu-22.04" ] ; then \
-          sudo apt-get -qy install \
-              qt6-3d-dev xvfb \
-              libgl1-mesa-dev
-        fi
+        # installing this package, cmake won't find the correct lib path.
+        # This has been replaced by the 'Install Qt' section below to manage qt6 versioning independently from the OS.
```
yungyuc: You didn't pay attention to the line width (79-char).  The overly long lines make it harder to maintain.

.github/workflows/devbuild.yml
```diff
@@ -422,17 +448,17 @@ jobs:
         run: |
           echo "github.event_name: ${{ github.event_name }}"
 
-      - name: install qt
+      - name: Install Qt
```
yungyuc: It's not right to change from lower case to capitalization.  Look around, all names are using lower cases.

Please fix the style.

.github/workflows/devbuild.yml
```diff
@@ -230,6 +243,7 @@ jobs:
 
     env:
       QT_DEBUG_PLUGINS: 1
+      PIP_BREAK_SYSTEM_PACKAGES: 1 # disabling PEP668
```
yungyuc, line 259: I like the use of an environment variable to control installation behavior and the comment.

yungyuc (changes requested):
- [x] Fix line width
- [x] Fix action name styles (use lower case)

yungyuc:
@terrychan999 If you are ready for the next review, please leave a global comment.

terrychan999:
> @terrychan999 If you are ready for the next review, please leave a global comment.

Thanks for the reminder. Ready for review.

yungyuc (approved):
LGTM.  Thanks a lot.  Since the CI mac runs are unstable now, I will merge after it comes back to be stable.

yungyuc (changes requested):
@terrychan999 The CI fix #309 invalidated this PR.  Could you please help rebase to the latest master and resolve the conflict?

yungyuc:
@terrychan999 Can you take a look?

terrychan999:
> @terrychan999 The CI fix #309 invalidated this PR. Could you please help rebase to the latest master and resolve the conflict?

Rebased. Ready for review, thanks.

yungyuc (approved):

yungyuc:
Thanks a lot, @terrychan999

yungyuc:
This PR fixed #284

PR #293, opened 2024-03-14
fix macos-13 CI

tigercosmos:
dark magic, kinda ugly, but at least works

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -254,16 +254,28 @@ jobs:
         run: |
           export HOMEBREW_NO_AUTO_UPDATE=1
           export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+          if [ "${{ matrix.os }}" == "macos-13" ] ; then
+            # GitHub Action macOS 13 has a problem with python3
+            brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done;
+          fi
           brew install qt6
 
       - name: dependency by pip
         run: |
           echo "which python3: $(which python3)"
           ls -al $(which python3)
-          python3 -m pip -v install --upgrade setuptools
-          python3 -m pip -v install --upgrade pip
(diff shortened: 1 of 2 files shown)


tigercosmos:
@yungyuc  please check

.github/workflows/devbuild.yml
```diff
@@ -254,16 +254,28 @@ jobs:
         run: |
           export HOMEBREW_NO_AUTO_UPDATE=1
           export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+          if [ "${{ matrix.os }}" == "macos-13" ] ; then
+            # GitHub Action macOS 13 has a problem with python3
+            brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done;
```
tigercosmos, line 259: found the discussion here: https://github.com/orgs/Homebrew/discussions/5199

.github/workflows/devbuild.yml
```diff
@@ -254,16 +254,28 @@ jobs:
...
         run: |
           echo "which python3: $(which python3)"
           ls -al $(which python3)
-          python3 -m pip -v install --upgrade setuptools
-          python3 -m pip -v install --upgrade pip
-          python3 -m pip -v install --upgrade numpy pytest flake8
-          python3 -m pip -v install --upgrade pyside6==$(qmake -query QT_VERSION)
+          if [ "${{ matrix.os }}" == "macos-13" ] ; then
+            # suppress the warning of pip because of python link issue of brew
+            python3 -m pip -v install --upgrade setuptools  --break-system-packages
```
tigercosmos, line 269: unfortunately we need this `--break-system-packages` due to brew python link issue

yungyuc: This is not something we should treat as normal, but not a horrible no for a workaround for an issue specific to one version on one platform.

yungyuc (approved):
Many thanks for the workaround!

yungyuc:
This PR fixed #289

Issue #296, opened 2024-03-19
Improve dtype mapping


yungyuc:
In https://github.com/solvcon/modmesh/blob/44d4ccf8471678a1add2e830387bd02de348160a/cpp/modmesh/buffer/SimpleArray.cpp#L36 the list of if does not look like a perfect solution to the string to type code conversion:

```cpp
DataType get_data_type_from_string(const std::string & data_type_string)
{
    if (data_type_string == "bool")
    {
        return DataType::Bool;
    }
    if (data_type_string == "int8")
    {
        return DataType::Int8;
    }
    if (data_type_string == "int16")
    {
        return DataType::Int16;
    }
    if (data_type_string == "int32")
    {
        return DataType::Int32;
    }
    if (data_type_string == "int64")
    {
        return DataType::Uint64;
    }
    if (data_type_string == "uint8")
    {
        return DataType::Uint8;
    }
    if (data_type_string == "uint16")
    {
        return DataType::Uint16;
    }
    if (data_type_string == "uint32")
    {
        return DataType::Uint32;
    }
    if (data_type_string == "uint64")
    {
        return DataType::Uint64;
    }
    if (data_type_string == "float32")
    {
        return DataType::Float32;
    }
    if (data_type_string == "float64")
    {
        return DataType::Float64;
    }
    throw std::runtime_error("Unsupported datatype");
}
```

@visitorckw proposed to use a hash table for the mapping, like:

```cpp
DataType get_data_type_from_string(const std::string & data_type_string)
{
    const DataType table[4][4] = {
        {DataType::Bool, DataType::Bool, DataType::Bool, DataType::Bool},
        {DataType::Int32, DataType::Int64, DataType::Int16, DataType::Int8},
        {DataType::Uint32, DataType::Uint64, DataType::Uint16, DataType::Uint8},
        {DataType::Float32, DataType::Float64, DataType::Float32 DataType::Float64}
    };

    const char ch1 = data_type_string.front();
    const char ch2 = data_type_string.back();
    const int x = (ch1 - 'b');
    const int y = (ch2 - '0');
    const int i = (x & 3) ^ ((x & 4) >> 1);
    const int j = (y >> 1) - 1;

    return table[i & 3][j & 3];
}
```

The hash table should run faster than the branches.  The code is much more compact and should be more maintainable "in some ways".  The downside is obvious: the hash-table implementation is much more convoluted.  But it may be mitigated by proper comments and encapsulation.

A possible API could look like:

```cpp
[...]

PR #297, opened 2024-03-20
improve data type

tigercosmos:
#296

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -45,7 +45,7 @@ struct DataTypeHasher
     }
 };
 
-// NOLINTNEXTLINE(cert-err58-cpp, cppcoreguidelines-avoid-non-const-global-variables, fuchsia-statically-constructed-objects)
+// NOLINTNEXTLINE(misc-use-anonymous-namespace, cert-err58-cpp, cppcoreguidelines-avoid-non-const-global-variables, fuchsia-statically-constructed-objects)
 static std::unordered_map<std::string, DataType, DataTypeHasher> string_data_type_map = {
     {"bool", DataType::Bool},
     {"int8", DataType::Int8},
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -46,7 +46,7 @@ struct DataTypeHasher
     //      "float64" -> 'f' << 8 | '4' = 26164
     std::size_t operator()(const std::string & data_type_string) const
     {
-        assert(data_type_string.size() > 4); // the length of "bool" and "int8" is 4, and all other data types are longer than 4
+        assert(data_type_string.size() >= 4); // the length of "bool" and "int8" is 4, and all other data types are longer than 4
         const char ch1 = data_type_string.front();
         const char ch2 = data_type_string.back();
         return static_cast<std::size_t>(ch1) << 8 | static_cast<std::size_t>(ch2);
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -80,67 +80,67 @@ DataType::DataType(const std::string & data_type_string)
 }
 
 template <>
-DataType get_data_type_from_type<bool>()
+DataType DataType::from<bool>()
 {
     return DataType::Bool;
 }
 
 template <>
-DataType get_data_type_from_type<int8_t>()
+DataType DataType::from<int8_t>()
 {
     return DataType::Int8;
 }
 
 template <>
-DataType get_data_type_from_type<int16_t>()
+DataType DataType::from<int16_t>()
 {
     return DataType::Int16;
 }
 
 template <>
-DataType get_data_type_from_type<int32_t>()
+DataType DataType::from<int32_t>()
 {
     return DataType::Int32;
 }
 
 template <>
-DataType get_data_type_from_type<int64_t>()
+DataType DataType::from<int64_t>()
 {
     return DataType::Int64;
 }
 
 template <>
-DataType get_data_type_from_type<uint8_t>()
+DataType DataType::from<uint8_t>()
 {
     return DataType::Uint8;
 }
 
 template <>
-DataType get_data_type_from_type<uint16_t>()
+DataType DataType::from<uint16_t>()
 {
     return DataType::Uint16;
 }
 
 template <>
-DataType get_data_type_from_type<uint32_t>()
+DataType DataType::from<uint32_t>()
 {
     return DataType::Uint32;
 }
 
 template <>
-DataType get_data_type_from_type<uint64_t>()
+DataType DataType::from<uint64_t>()
 {
     return DataType::Uint64;
 }
 
 template <>
-DataType get_data_type_from_type<float>()
+DataType DataType::from<float>()
 {
     return DataType::Float32;
 }
 
 template <>
-DataType get_data_type_from_type<double>()
+DataType DataType::from<double>()
 {
     return DataType::Float64;
 }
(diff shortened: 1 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -668,23 +668,42 @@ using SimpleArrayUint64 = SimpleArray<uint64_t>;
...
+        Uint16,
+        Uint32,
+        Uint64,
+        Float32,
+        Float64,
+    }; /* end enum Type */
+
+    DataType() = default;
+
+    constexpr DataType(const DataTypeEnum datatype)
```
tigercosmos: This technique enables implicit casting between `DataType` and `DataType::DataTypeEnum`, providing the advantages of both a regular class and an enum class, with some similarities to the latter.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -31,53 +31,41 @@
...
-        return DataType::Uint64;
-    }
-    if (data_type_string == "float32")
-    {
-        return DataType::Float32;
-    }
-    if (data_type_string == "float64")
+struct DataTypeHasher
+{
+    std::size_t operator()(const std::string & data_type_string) const
```
tigercosmos, line 47: the customized hasher.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -28,56 +28,46 @@
...
+    {
+        const char ch1 = data_type_string.front();
+        const char ch2 = data_type_string.back();
+        return static_cast<std::size_t>(ch1) << 8 | static_cast<std::size_t>(ch2);
+    }
+};
```
yungyuc: End marking.  Please don't forget about it.  It's tedious to pay attention to basic styles like this.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -28,56 +28,46 @@
...
+    {"int16", DataType::Int16},
+    {"int32", DataType::Int32},
+    {"int64", DataType::Int64},
+    {"uint8", DataType::Uint8},
+    {"uint16", DataType::Uint16},
+    {"uint32", DataType::Uint32},
+    {"uint64", DataType::Uint64},
+    {"float32", DataType::Float32},
+    {"float64", DataType::Float64}};
+} // end of namespace detail
```
yungyuc: Use consistent style of namespace end marking.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -28,56 +28,46 @@
...
+
+DataType::DataType(const std::string & data_type_string)
```
yungyuc, line 72: Could you check if `inline` makes more efficient assembly?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -668,23 +668,41 @@ using SimpleArrayUint64 = SimpleArray<uint64_t>;
...
+        Int16,
+        Int32,
+        Int64,
+        Uint8,
+        Uint16,
+        Uint32,
+        Uint64,
+        Float32,
+        Float64,
+    }; /* end enum Type */
```
yungyuc: Please pay attention to end marking.  You are not consistent to the existing style nor your own code.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -668,23 +668,41 @@ using SimpleArrayUint64 = SimpleArray<uint64_t>;
...
-    Uint16,
-    Uint32,
-    Uint64,
-    Float32,
-    Float64,
-}; /* end enum class DataType */
-
-DataType get_data_type_from_string(const std::string & data_type_string);
+public:
+    enum DataTypeEnum
```
yungyuc: Perhaps `enum_type` may be used as the internal type name.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: This looks good.

cpp/modmesh/buffer/SimpleArray.cpp
yungyuc: Please add comments to explain why this hasher is applicable to the type string.

It may be illustrative to give some string:value pair as examples.

yungyuc (changes requested):
- [x] End marking.
- [x] Add unit tests.  If you don't think unit tests should be added at this point, please explain why not.

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: I tried, but the compiler didn't allow me to put "inline" here. There is a linking issue if I add `inline`. The signature should be the same(?)

if I added `inline`, error:
```
[ 82%] Linking CXX executable test_nopython
/usr/bin/ld: CMakeFiles/test_nopython.dir/test_nopython_buffer.cpp.o: in function `SimpleArray_DataType_from_string_Test::TestBody()':
/mnt/c/Users/tiger/modmesh/gtests/test_nopython_buffer.cpp:78: undefined reference to `modmesh::DataType::DataType(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
/usr/bin/ld: /mnt/c/Users/tiger/modmesh/gtests/test_nopython_buffer.cpp:81: undefined reference to `modmesh::DataType::DataType(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
/usr/bin/ld: /mnt/c/Users/tiger/modmesh/gtests/test_nopython_buffer.cpp:84: undefined reference to `modmesh::DataType::DataType(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
/usr/bin/ld: /mnt/c/Users/tiger/modmesh/gtests/test_nopython_buffer.cpp:85: undefined reference to `modmesh::DataType::DataType(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
```

tigercosmos:
@yungyuc  all issues addressed. (1) end mark, (2) unit test, (3) comments.

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: so I guess we need to put the implementation in hpp to avoid this issue; however, I think this implementation is better to keep in cpp, because it also use other static object in `modmesh::detail` in cpp.

yungyuc: You are right.  The whole things need to be in header files.  Let's try it later.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -28,120 +28,119 @@
...
+
+} /* end namespace detail */
+
+DataType::DataType(const std::string & data_type_string)
+{
+    auto it = detail::string_data_type_map.find(data_type_string);
+    if (it == detail::string_data_type_map.end())
     {
-        return DataType::Float64;
+        throw std::runtime_error("Unsupported datatype");
```
yungyuc: Be more rigorous on exception type.  Check https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html and https://en.cppreference.com/w/cpp/error/invalid_argument for more information.

yungyuc (changes requested):
- [x] Throw more suitable exception type

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: I see. thanks for the explanation. fixed.

tigercosmos:
@yungyuc all clear. please review.

yungyuc (approved):
LGTM.  Merging.

PR #298, opened 2024-03-22
refactor RadixTree

tigercosmos:
The `TimedEntry` struct located in `RadixTree.hpp` is intended for testing purposes. Let's relocate the struct to the test file.

### DIFF
--- a/gtests/test_nopython_radixtree.cpp
+++ b/gtests/test_nopython_radixtree.cpp
@@ -1,9 +1,36 @@
-#include <modmesh/toggle/RadixTree.hpp>
 #include <gtest/gtest.h>
+#include <modmesh/toggle/RadixTree.hpp>
 
 #ifdef Py_PYTHON_H
 #error "Python.h should not be included."
 #endif
+
+/**
+ * Simple Timed Entry for Testing
+ */
+class TimedEntry
+{
+public:
+    size_t count() const { return m_count; }
+    double time() const { return m_time; }
+
+    TimedEntry & add_time(double time)
+    {
+        ++m_count;
+        m_time += time;
+        return *this;
+    }
+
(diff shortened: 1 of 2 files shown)


tigercosmos:
cc @q40603  @yungyuc

gtests/test_nopython_radixtree.cpp
```diff
@@ -1,10 +1,37 @@
...
+#include <modmesh/toggle/RadixTree.hpp>
 
 #ifdef Py_PYTHON_H
 #error "Python.h should not be included."
 #endif
 
+/**
+ * Simple Timed Entry for Testing
+ */
+class TimedEntry
```
yungyuc: Moving the test code here makes sense to me.  What do you think, @q40603 ?

We have another `TimedEntry` at https://github.com/solvcon/modmesh/blob/d6ed9ac95f449ff77eb6e4d379c83d25f60eac23/cpp/modmesh/toggle/profile.hpp#L102 

For the test code here, what about renaming it to be different from `modmesh::TimedEntry` there?

yungyuc (changes requested):
- [x] Rename the class for the test code to be like test code.

gtests/test_nopython_radixtree.cpp
tigercosmos: renamed.

yungyuc (approved):
LGTM.  Wait for @q40603 comments

q40603:
@tigercosmos 
Sorry for the late response.
Separating the TimedEntry class from Radix Tree is more elegant, and it also makes more sense as Radixtree should accept other objects as well.

yungyuc:
Great!  Thank you @tigercosmos and @q40603

PR #301, opened 2024-03-30
fix macos-14 ci

terrychan999:
This is a temporary fix of macOS-14 CI, similar to the previous PR https://github.com/solvcon/modmesh/pull/293.

The error happens during pip install:
https://github.com/solvcon/modmesh/actions/runs/8483820632/job/23245585597

The root cause is Homebrew's enforcement of PEP668 policy, which prohibits the direct installation of python packages not managed by Homebrew.
https://docs.brew.sh/Homebrew-and-Python#pep-668-python312-and-virtual-environments

@yungyuc @tigercosmos ready for review, thanks.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -264,8 +264,8 @@ jobs:
         run: |
           echo "which python3: $(which python3)"
           ls -al $(which python3)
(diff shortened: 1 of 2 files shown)


tigercosmos:
LGTM

yungyuc:
It's great.  Thanks a lot, @terrychan999 !

PR #303, opened 2024-04-02
arrayplex setitem & refactoring

tigercosmos:
- moved shared code  to `array_common.hpp`
- supported arrayplex to set items.


cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -118,28 +99,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-                [](wrapped_type const & self)
-                {
-                    py::tuple ret(self.stride().size());
-                    for (size_t i = 0; i < self.stride().size(); ++i)
-                    {
-                        ret[i] = self.stride()[i];
-                    }
-                    return ret;
-                })
+            .def_property_readonly("shape", &get_array_shape<T>)
```
tigercosmos: use the function from `array_common.hpp`, same for other functions.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -251,20 +237,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
-                            }
-                            return pybind11::buffer_info(
-                                array.data(), /* Pointer to buffer */
-                                sizeof(data_type), /* Size of one scalar */
-                                pybind11::format_descriptor<data_type>::format(), /* Python struct-style format descriptor */
-                                array.ndim(), /* Number of dimensions */
-                                std::vector<size_t>(array.shape().begin(), array.shape().end()), /* Buffer dimensions */
-                                stride /* Strides (in bytes) for each index */
-                            ); });
+                            return get_buffer_info<data_type>(array); });
```
tigercosmos: use the function from `array_common.hpp`

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -285,43 +258,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
             .def("__getitem__", &get_typed_array_value<const std::vector<ssize_t> &>)
-            .def(
-                "reshape",
-                [](wrapped_type const & self, pybind11::object const & py_shape)
-                { return execute_callback_with_typed_array(
-                      self,
-                      // NOLINTNEXTLINE(fuchsia-trailing-return)
-                      [&](const auto & array) -> wrapped_type // need the return type to get correct deduced type
-                      { 
+            .def("__setitem__", [](wrapped_type & self, pybind11::args const & args)
```
tigercosmos, line 289: support set items for arrayplex.

tests/test_buffer.py
```diff
@@ -774,12 +774,47 @@ def test_SimpleArrayPlex_properties(self):
...
+        self.assertEqual(sarr.nbytes, 2 * 3 * 4 * 8)
+        self.assertEqual(sarr.size, 2 * 3 * 4)
         self.assertEqual(sarr.itemsize, 8)
         self.assertEqual(sarr.stride, (12, 4, 1))
-        self.assertEqual(len(sarr), 2*3*4)  # number of elements
+        self.assertEqual(len(sarr), 2 * 3 * 4)  # number of elements
         self.assertEqual(sarr.nbody, 2 - 0)
         self.assertEqual(sarr.has_ghost, False)
 
+    def test_SimpleArrayPlex_set_item_simple(self):
```
tigercosmos, line 840: it works :)

tigercosmos:

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,307 @@
...
+#include <modmesh/buffer/SimpleArray.hpp>
+#include <modmesh/buffer/pymod/TypeBroadcast.hpp>
+
+namespace modmesh
+{
+
+namespace python
+{
+using slice_type = small_vector<int>;
+using shape_type = modmesh::detail::shape_type;
```
yungyuc: `shape_type` should not be in `modmesh::python` namespace as it has been in `modmesh::detail` namespace.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+#include <modmesh/buffer/SimpleArray.hpp>
+#include <modmesh/buffer/pymod/TypeBroadcast.hpp>
+
+namespace modmesh
+{
+
+namespace python
+{
+using slice_type = small_vector<int>;
```
yungyuc: I don't like `slice_type` to be a global alias, but I am not sure how to deal with it yet.  You can leave it here.  I will think more about it.

yungyuc: It's good to factor out the standalone common file.  It makes issues like this stand out.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,307 @@
...
+    }
+    catch (const pybind11::cast_error &)
+    {
+        shape = shape_in.cast<std::vector<size_t>>();
+    }
+    return shape;
+}
+
+template <typename T>
+inline std::vector<slice_type> make_default_slices(SimpleArray<T> const & arr)
```
yungyuc: The multitude of free function (templates) made me think that maybe classes for shape and slice would further simplify the implementation.  What do you think?

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: This (and alike) are good.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -285,43 +258,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
                     return execute_callback_with_typed_array(
                         self, [](const auto & array)
                         {
-                            pybind11::tuple ret(array.shape().size());
-                            for (size_t i = 0; i < array.shape().size(); ++i)
-                            {
-                                ret[i] = array.shape()[i];
-                            }
-                            return ret; });
+                            using data_type = typename std::remove_const_t<std::remove_reference_t<decltype(array[0])>>;
```
yungyuc: This is too magical.  The original lambda is much more readable.  Please don't use this overly tricky template code.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -285,43 +258,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
+                        self, [](const auto & array)
                         {
-                            pybind11::tuple ret(array.stride().size());
-                            for (size_t i = 0; i < array.stride().size(); ++i)
-                            {
-                                ret[i] = array.stride()[i];
-                            }
-                            return ret;
-                        });
+                            using data_type = typename std::remove_const_t<std::remove_reference_t<decltype(array[0])>>;
```
yungyuc: Lambda is clearer

tests/test_buffer.py
yungyuc: Fantastic

yungyuc (changes requested):
- [x] Limit the scope of `shape_type`
- [x] Discuss: whether or not to use classes for shape and slice
- [x] Avoid overly tricky template code

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: I agree with you. A helper class might be useful.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: I see. I thought it is readable if both wrapper class call the same function.
So, do you want to revert the change for both "shape" and "stride"?

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: maybe we can also put it `modmesh::detail`. I couldn't find a good place for it. Maybe `SimpleArray.hpp` is fine, but it might be better to have a header file for basic types.

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: However, if the existing functions lack a robust correlation, the helper class might just serve as a namespace.

yungyuc:
@tigercosmos any update?

tigercosmos:
> @tigercosmos any update?

@yungyuc  I've left few comments few days ago, and waiting for your response.

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: I am looking for a good design pattern for this these days, so far I haven't found it yet.

yungyuc: Do you want to use this PR to make them classes, or leave it here for a future PR?  I am fine either way.  We should choose one way to proceed quickly.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: I think the original lambda is good for this occurrence.  Specifically here reverting is good.

In other occurrences maybe more review is needed.  I am fine to use this PR or another one to handle the cases that need review.  A decision should be made quickly.

yungyuc:
Thanks, @tigercosmos .  I added comments inline.

Please always leave a global comment after you have done your commenting and wait for feedback, like what I am doing in this comment.

Inline comments are hard to track.  For example, in the following snapshot, the only clue for me to tell whether I should review again is the message you left 32 minutes ago:

image

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: let's make it in another PR.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: reverted.

tigercosmos:
@yungyuc  regarding the issues above:
 - Limit the scope of shape_type: Done
- Discuss: whether or not to use classes for shape and slice:  Done
- Avoid overly tricky template code: Reverted.

tigercosmos:
CI failed due to the temporary failure of macOS-13

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,288 @@
...
+    catch (const pybind11::cast_error &)
+    {
+        shape = shape_in.cast<std::vector<size_t>>();
+    }
+    return shape;
+}
+
+/// Helper class for array property in Python.
+template <typename T>
+class ArrayPropertyHelper
```
tigercosmos, line 58: new introduced.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,288 @@
...
+{
+public:
```
tigercosmos, line 60: only public APIs are here.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,288 @@
...
+            array.data(), /* Pointer to buffer */
+            sizeof(T), /* Size of one scalar */
+            pybind11::format_descriptor<T>::format(), /* Python struct-style format descriptor */
+            array.ndim(), /* Number of dimensions */
+            std::vector<size_t>(array.shape().begin(), array.shape().end()), /* Buffer dimensions */
+            stride /* Strides (in bytes) for each index */
+        );
+    }
+
+private:
```
tigercosmos, line 161: private functions within the class

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: no need to repeat template now.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: `shape_type` is still in `modmesh::python` namespace.  This should be fixed.  Please remove this alias When using `shape_type`, spelling out the full scope `modmesh::detail::shape_type`.

Perhaps `slice_type` should be treated in the same way.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -84,6 +84,7 @@ namespace detail
 
 using shape_type = small_vector<size_t>;
 using sshape_type = small_vector<ssize_t>;
+using slice_type = small_vector<int>;
```
yungyuc: While moved it here, it becomes obvious that `int` does not look right.  `slice_type` should be interchangeable with `sshape_type`.  Please review and make them consistent.

yungyuc (changes requested):
Thanks, @tigercosmos .  There are an remaining issue and a new one:

- [x] (Still need to address) Limit the scope of `shape_type`
- [x] Review `slice_type` and its consistency with other type aliases

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,289 @@
...
+    return shape;
+}
+
+/// Helper class for array property in Python.
+template <typename T>
+class ArrayPropertyHelper
+{
+public:
+
+    static inline void broadcast_array_using_ellipsis(SimpleArray<T> & arr_out, pybind11::array const & arr_in)
```
yungyuc: Why do you need `inline` for functions already in the class declaration?

tigercosmos: I want to make the function code to be inline within the caller. I think static and inline can be used at the same time?

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: thanks for catching this. fixed.

tigercosmos:
@yungyuc fixed the issue of `slice_type` and `shape_type`, and added comments.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -84,6 +84,7 @@ namespace detail
 
 using shape_type = small_vector<size_t>;
 using sshape_type = small_vector<ssize_t>;
+using slice_type = small_vector<size_t>;
```
yungyuc: Why do you use an unsized integer rather than a signed?  Slice can be negative-valued.

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -40,11 +40,9 @@ namespace python
 template <typename T /* original type */, typename D /* for destination type */>
 struct TypeBroadcastImpl
 {
-    using slice_type = small_vector<int>;
```
yungyuc, line 43: It's OK to have type aliases in the class.  Why removing them?

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: They can, but the `inline` should be no op, because code in the class declaration is automatically inlined, IIRC.

yungyuc (changes requested):
One more issue in your update:

- [ ] Signness of the value_type of slice_type

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: I see, then I think your previous comment, https://github.com/solvcon/modmesh/pull/303#discussion_r1557555171, should not be an issue?

shape is always positive, and slice can be negative, so they don't need to be interchangeable.

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
tigercosmos: I misunderstood. I thought you were asking to remove all aliases.

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: sure, removed.

tigercosmos:
@yungyuc please check the comments.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: I misunderstood, fixed.

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
tigercosmos: reverted.

tigercosmos:
CI failed due to unrelated issues for all macOS

tigercosmos:
@yungyuc  fixed all issues, ready for review.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -84,6 +84,7 @@ namespace detail
 
 using shape_type = small_vector<size_t>;
 using sshape_type = small_vector<ssize_t>;
+using slice_type = small_vector<ssize_t>;
```
yungyuc, line 87: Great

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+/// Helper class for array property in Python.
+template <typename T>
+class ArrayPropertyHelper
+{
+public:
+    using shape_type = modmesh::detail::shape_type;
+    using slice_type = modmesh::detail::slice_type;
+
+    static void broadcast_array_using_ellipsis(SimpleArray<T> & arr_out, pybind11::array const & arr_in)
```
yungyuc, line 64: Looks good

cpp/modmesh/buffer/pymod/TypeBroadcast.hpp
yungyuc: No problem

yungyuc (approved):
LGTM.  The CI mac runs are unstable.  I will merge once the CI are back online.

PR #304, opened 2024-04-07
First chunk of code for array-compatible collecting container

### LINKED ISSUE #305: Refactor to consolidate common parts between `ConcreteBuffer` and `BufferExpander`
There are common code between `ConcreteBuffer` (fixed-size buffer manager) and `BufferExpander` (growing buffer manager), e.g., the [`validate_range()`](https://github.com/solvcon/modmesh/blob/3d1944e2342118b62e3349f425b3585a89002b6c/cpp/modmesh/buffer/ConcreteBuffer.hpp#L279C10-L279C24).  Refactoring can make the code more maintainable.  There are 3 types of tasks:

1. Extract the common code to base classes (mixins), or use other ways to share the code.  (No need to overdo it.  Some duplicated code is good for maintenance and debugging.)
2. Unify the type aliases, e.g., [`size_type`](https://github.com/solvcon/modmesh/blob/3d1944e2342118b62e3349f425b3585a89002b6c/cpp/modmesh/buffer/BufferExpander.hpp#L58)
3. Move lengthy code from the header file `ConcreteBuffer.hpp` to the implementation file `ConcreteBuffer.cpp` to save compile time

They may use different PRs for clarity.  This issue is to follow up the discussion at https://github.com/solvcon/modmesh/pull/304#discussion_r1555105830

yungyuc:
This is preparatory work for developing code/algorithms for mesh generation, which needs resizable buffers.

The class template `SimpleCollector` is a fundamental-type container that holds data in a contiguous buffer.  Unlike `SimpleArray`, the buffer size in `SimpleCollector` may grow.  At this early stage, I am not adding appending methods.  I haven't made my mind to call them like `std::vector::push_back` or use a different name.  `SimpleCollector::expand()` and `SimpleCollector::reserve()` are provided to verify the buffer growth works as expected.

In addition to make it expandable, a key reason to make `SimpleCollector` instead of using `std::vector` is that the latter disallows safe use of its internal buffer.  `SimpleCollector` uses a new class, `BufferExpander`, to ensure that the memory buffer created by `SimpleCollector::as_array()` has correct lifecycle while being used as the internal buffer in `SimpleCollector`.  When the buffer needs to be expanded, `BufferExpander` will use another holder for the newly created memory chunk and release the shared pointer to the `ConcreteBuffer` (which belongs to the `SimpleArray`).

Just like `ConcreteBuffer` is a (low-level) companion to `SimpleArray` for memory management, `BufferExpander` is the companion to `SimpleCollector`.  We may need more tailored memory management in the future, and the memory-managing code should be put in `ConcreteBuffer` and `BufferExpander` while leaving the data manipulation code in `SimpleArray` and `SimpleCollector`.


yungyuc:
Inline review comments WIP.

contrib/standalone_buffer/Makefile
```diff
@@ -44,21 +44,26 @@ $(SETUPROOT)/wrap_SimpleArrayPlex.o: $(SETUPROOT)/modmesh/buffer/pymod/wrap_Simp
...
-	c++ $(LDFLAGS) \
-		$(SETUPROOT)/SimpleArray.o \
-		$(SETUPROOT)/wrap_ConcreteBuffer.o \
-		$(SETUPROOT)/wrap_SimpleArray.o \
-		$(SETUPROOT)/wrap_SimpleArrayPlex.o \
-		$(SETUPROOT)/buffer_pymod.o \
-		$(SETUPROOT)/modbuf.o \
-		-o $@
+OBJS = \
+	$(SETUPROOT)/BufferExpander.o \
```
yungyuc, line 57: This is the new guy to the family.  It's too long for duplication in the prerequisite and the command so I create a new variable to hold the list.

cpp/modmesh/buffer/BufferExpander.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace modmesh
+{
+
+void BufferExpander::reserve(size_type cap)
+{
+    if (cap > capacity())
+    {
+        size_type const old_size = size();
+        // Create new data holder and copy data.
+        std::unique_ptr<int8_t> new_data_holder = allocate(cap);
```
yungyuc, line 40: When requesting more memory, always use the (alternate) data holder instead of the `m_concrete_buffer`.  We may play (after learning) more tricks of memory management with the data holder.

`ConcreteBuffer` is optimized for reading.  I don't want it to look too tricky regarding managing memory.

cpp/modmesh/buffer/BufferExpander.cpp
```diff
@@ -0,0 +1,84 @@
...
+        m_end = m_begin + old_size;
+        m_end_cap = m_begin + cap;
+    }
+    else
+    {
+        return;
+    }
+}
+
+std::shared_ptr<ConcreteBuffer> BufferExpander::copy_concrete(size_type cap)
```
yungyuc: `copy_concrete()` simply create a new `ConcreteBuffer` without touching the existing one.  I forgot to make it `const` and will fix it later.

cpp/modmesh/buffer/BufferExpander.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+    size_type const old_size = size();
+    size_type const csize = cap > old_size ? cap : old_size;
+    auto buf = ConcreteBuffer::construct(csize);
+    std::copy_n(m_begin, old_size, buf->data());
+    return buf;
+}
+
+std::shared_ptr<ConcreteBuffer> const & BufferExpander::as_concrete(size_type cap)
```
yungyuc, line 68: `as_concrete()` modifies the internal buffer.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+namespace modmesh
+{
+
+/**
+ * Untyped and growing memory buffer for contiguous data storage.  The internal
+ * expandable memory buffer cannot be used externally.
+ */
+class BufferExpander
+    : public std::enable_shared_from_this<BufferExpander>
```
yungyuc, line 47: Use a shared pointer to control the lifecycle of `BufferExpander`.  It makes it possible to share one `BufferExpander` among multiple `SimpleCollector`.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -0,0 +1,173 @@
...
+    {
+        std::unique_ptr<int8_t> ret(nullptr);
+        if (0 != nbytes)
+        {
+            ret = std::unique_ptr<int8_t>(new int8_t[nbytes]);
+        }
+        return ret;
+    }
+
+    std::unique_ptr<int8_t> m_data_holder = nullptr;
```
yungyuc, line 162: `m_data_holder` may be use more frequently than `m_concrete_buffer` because this class focuses on managing memory for collecting data.

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -274,6 +274,8 @@ class ConcreteBuffer
     // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays)
     using unique_ptr_type = std::unique_ptr<int8_t, data_deleter_type>;
 
+private:
```
yungyuc, line 277: This is a horrible bug.  I did not put data members in private!  Fix it.

cpp/modmesh/buffer/SimpleCollector.hpp
```diff
@@ -0,0 +1,143 @@
...
+        {
+            m_expander = std::move(other.m_expander);
+        }
+        return *this;
+    }
+
+    size_t size() const { return expander().size() / ITEMSIZE; }
+    size_t capacity() const { return expander().capacity() / ITEMSIZE; }
+
+    value_type const & operator[](size_t it) const noexcept { return data(it); }
```
yungyuc, line 96: Data collecting / accumulating is usually linear.  I think we only need 1D content access.

cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp
```diff
@@ -110,9 +110,53 @@ WrapConcreteBuffer::WrapConcreteBuffer(pybind11::module & mod, char const * pyna
...
+
+    (*this)
+        .def_timed(
+            py::init(
+                [](size_t length)
+                { return wrapped_type::construct(length); }),
+            py::arg("length"))
+        .def_timed(py::init([]()
+                            { return wrapped_type::construct(); }))
+        .def_timed("reserve", &wrapped_type::reserve, py::arg("cap"))
```
yungyuc, line 136: The argument name "`cap`" is to look like the argument name of `std::vector::reserve` in https://en.cppreference.com/w/cpp/container/vector/reserve

tests/test_buffer.py
```diff
@@ -86,6 +86,60 @@ def test_ConcreteBuffer_from_ndarray(self):
         self.assertTrue((ndarr == 0).all())
 
 
+class BufferExpanderBasicTC(unittest.TestCase):
+
+    def test_BufferExpander(self):
```
yungyuc, line 91: Is the testing code self-explaining?

tests/test_buffer.py
```diff
@@ -634,7 +688,8 @@ def test_SimpleArray_casting(self):
 
         # check if arrayplex can cast to simplearray
         self.assertEqual(
-            helper.test_load_arrayfloat64_from_arrayplex(arrayplex_float64), True)  # noqa: E501
+            helper.test_load_arrayfloat64_from_arrayplex(arrayplex_float64),
```
yungyuc, line 691: Linter made me do it.  No change of logic in this function.

tests/test_buffer.py
```diff
@@ -774,12 +829,64 @@ def test_SimpleArrayPlex_properties(self):
...
         self.assertEqual(sarr.stride, (12, 4, 1))
-        self.assertEqual(len(sarr), 2*3*4)  # number of elements
+        self.assertEqual(len(sarr), 2 * 3 * 4)  # number of elements
         self.assertEqual(sarr.nbody, 2 - 0)
         self.assertEqual(sarr.has_ghost, False)
 
+
+class SimpleCollectorTC(unittest.TestCase):
+
+    def test_construct(self):
```
yungyuc, line 843: Is the testing code self-explaining?

cpp/modmesh/buffer/BufferExpander.cpp
yungyuc: Fixed in https://github.com/solvcon/modmesh/pull/304/commits/67761889471d10cd853d4db923494d20ca5b6c2c

yungyuc:
@tigercosmos it's ready for review.

If I have subsequent work needing to use code in this PR I will go ahead merging it.  Feedbacks will be addressed in a follow up PR.  If I don't have time for the depending work, this PR will be open for commenting for a little bit longer time.

tests/test_buffer.py
tigercosmos: Seems like there is some flexibility in flake8. When I use different platforms, the format is sometimes changed due to the linter as well.

tests/test_buffer.py
```diff
@@ -728,17 +783,17 @@ def test_SimpleArrayPlex_constructor(self):
...
             ):
                 modmesh.SimpleArray((2, 3, 4), dtype=dtype, value=3.3)
 
         dtype_list_float = ["float32", "float64"]
         for dtype in dtype_list_float:
             modmesh.SimpleArray((2, 3, 4), dtype=dtype, value=3.0)
             with self.assertRaisesRegex(
-                TypeError,
-                r"Data type mismatch, expected Python float"
+                    TypeError,
```
tigercosmos, line 795: this format looks weird, 8 spaces?

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: question, why not reuse `using unique_ptr_type = std::unique_ptr<int8_t, data_deleter_type>;` in ConcreteBuffer.hpp?

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /* Backdoor */
+    int8_t data(size_type it) const { return data()[it]; }
+    int8_t & data(size_type it) { return data()[it]; }
+    int8_t const * data() const noexcept { return data<int8_t>(); }
+    int8_t * data() noexcept { return data<int8_t>(); }
+    // clang-format off
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    template <typename T> T const * data() const noexcept { return reinterpret_cast<T *>(m_begin); }
```
tigercosmos, line 137: maybe we can have `assert` here to ensure the size is the multiple of the size of `T`?

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -0,0 +1,173 @@
...
+    BufferExpander(ctor_passkey const &) {}
+
+    BufferExpander() = delete;
+    BufferExpander(BufferExpander const &) = delete;
+    BufferExpander(BufferExpander &&) = delete;
+    BufferExpander & operator=(BufferExpander const &) = delete;
+    BufferExpander & operator=(BufferExpander &&) = delete;
+    ~BufferExpander() = default;
+
+    size_type size() const noexcept
```
tigercosmos, line 94: in `ConcreteBuffer`, we also have `nbytes()` method, not sure what is the original intention, but maybe we can also add it.

cpp/modmesh/buffer/SimpleCollector.hpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+    value_type & at(size_t it)
+    {
+        validate_range(it);
+        return data(it);
+    }
+
+    SimpleArray<T> as_array()
+    {
+        return SimpleArray<T>(expander().as_concrete());
```
tigercosmos, line 115: perhaps have assert have to ensure the underlying buffer size is correct?

cpp/modmesh/buffer/SimpleCollector.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include <modmesh/buffer/SimpleArray.hpp>
+
+#include <limits>
+#include <stdexcept>
+
+namespace modmesh
+{
+
+template <typename T>
+class SimpleCollector
```
tigercosmos, line 41: question, can we use `SimpleCollector` for `SimpleArray`'s underlying buffer?

tigercosmos:
@yungyuc  Overall it seems good, and I left some comments as feedback.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: `using unique_ptr_type = std::unique_ptr<int8_t, data_deleter_type>;` is not used in `BufferExpander` because I do not want the deleter yet.  We may need to write sophisticated reallocation code for the buffer managed by `m_data_holder`, and a deleter complicates it.

Maybe eventually we will need a deleter, but let's wait for a while.  The deleter in `ConcreteBuffer` wasn't added until it was needed.  And in the case of `BufferExpander`, maybe we will never need a deleter.  I hope we minimize the chance of overengineering.

tests/test_buffer.py
yungyuc: It could be configurable.  Reformatting the small piece in the function is fine.  We can fix the configuration when the linter is causing problems.

tests/test_buffer.py
yungyuc: Yes, 2 levels of indentation.  It's required by [PEP-8](https://peps.python.org/pep-0008/#indentation).  I am fine either way as long as the linted code does not look horrible.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: It may not be a good idea to add `assert` at a helper that should be performant and optimized away.

Are you talking about `m_begin`?  Having alignment check sounds like a good idea and it should be enforced all the time, not just `assert` (at debugging).  The allocation helper is the place to add the enforcement.  It may be tracked in a new issue.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: It's a good idea!  There are other helpers available in `ConcreteBuffer` but not in `BufferExpander`.  Some of them may be shared between the two types.

After the PR is merged, we may consider two things.  One is to refactor `ConcreteBuffer` and `BufferExpander` by providing mixins to collect common (and simple) code together.  The other is to supplement missing helpers.

I guess it's good for the two things to be done separately for clarity.

yungyuc (commented):
Thanks, @tigercosmos , for the swift feedback.  Let me merge the code.  We can follow up the points made in the discussions.

cpp/modmesh/buffer/SimpleCollector.hpp
yungyuc: Negative.  For sharing a buffer, especially across Python and C++, the buffer cannot change sizes.  That is why we cannot take the pointer of `std::vector::data()` and give it to `ndarray`.  `std::vector` may reallocate and make the pointer dangling any time `push_back()` is called.  Using `SimpleCollector` or `BufferExpander` inside `SimpleArray` results in the same problem.

And semantically variable-size container differs from fixed-size one (array).  It is productive to tell which semantic is in use by reading (and writing) the code.

cpp/modmesh/buffer/SimpleCollector.hpp
yungyuc: I don't think you can check for the size correctness here.  The size information is stored and calculated from the `expander()` inside.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: Issues #305 and #306 are created.

tests/test_buffer.py
tigercosmos: I didn't know there a special rule for it, thanks for telling me.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: @yungyuc  Yes, `m_begin` should be checked. I agree with tracking this with a new issue.

cpp/modmesh/buffer/SimpleCollector.hpp
tigercosmos: I am thinking about this:
```
size_t size buffer_size = expander().as_concrete().nbytes();
assert(buffer_size % sizeof(T) == 0)
```

tigercosmos:
@yungyuc  I replied to a few comments.

cpp/modmesh/buffer/SimpleCollector.hpp
yungyuc: It is redundant code, because it is guaranteed to be divisible in `reserve()` and `expand()` https://github.com/solvcon/modmesh/blob/3d1944e2342118b62e3349f425b3585a89002b6c/cpp/modmesh/buffer/SimpleCollector.hpp#L99

Adding the calculation increases unnecessary instructions and misleads maintainers in a way that they might think we don't need performance here.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: Feel free to create an issue for the alignment support in `ConcreteBuffer` and `BufferExpander`.  The alignment support should not use template argument because I want the two classes to remain to be classes.

yungyuc:
Thanks, @tigercosmos .  I replied to them.

PR #309, opened 2024-04-13
Fix Github Actions mac failure

yungyuc:
GHA CI mac runs failed with [inability to find `pybind11/pybind11.h` include file](https://github.com/solvcon/modmesh/actions/runs/8665855764/job/23765404198#step:9:65).  The root cause could be the change of the GHA mac image.  This PR fixes the failure by the following measures:

* pybind11 installs in `/usr/local/include` which MacOS Github Actions does not include.
  * Use `-DUSE_PYTHON_INCLUDE_DIR=ON` to change the installation path to Python include directory which is already picked up by the modmesh cmake configuration.
  * Create `/usr/local/include` for mac
* Improve `contrib/install.sh`
  * pybind11 uses `-DUSE_PYTHON_INCLUDE_DIR=ON`
  * Upgrade pybind11 from 2.11.1 to 2.12.0
  * Allow skipping `INSTALL_PREFIX`
  * Print cmakeargs

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -57,8 +57,14 @@ jobs:
           echo "cmake version: $(cmake --version)"
           echo "python3 path: $(which python3)"
           echo "python3 version: $(python3 --version)"
+          echo "python3-config --prefix: $(python3-config --prefix)"
+          echo "python3-config --exec-prefix: $(python3-config --exec-prefix)"
+          echo "python3-config --includes: $(python3-config --includes)"
+          echo "python3-config --libs: $(python3-config --libs)"
+          echo "python3-config --cflags: $(python3-config --cflags)"
+          echo "python3-config --ldflags: $(python3-config --ldflags)"
           echo "pip3 path: $(which pip3)"
-          python3 -c 'import numpy ; print("numpy.__version__:", numpy.__version__)'
+          python3 -c 'import numpy as np; print("np.__version__:", np.__version__, np.get_include())'
           echo "pytest path: $(which pytest)"
           echo "pytest version: $(pytest --version)"
           echo "clang-tidy path: $(which clang-tidy)"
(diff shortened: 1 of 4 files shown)


.github/workflows/devbuild.yml
```diff
@@ -57,8 +57,14 @@ jobs:
           echo "cmake version: $(cmake --version)"
           echo "python3 path: $(which python3)"
           echo "python3 version: $(python3 --version)"
+          echo "python3-config --prefix: $(python3-config --prefix)"
```
yungyuc, line 60: Add more prints to help diagnosis.

.github/workflows/devbuild.yml
```diff
@@ -249,15 +261,13 @@ jobs:
       - name: event name
         run: |
           echo "github.event_name: ${{ github.event_name }}"
+          # Some mac runner does not have /usr/local/include and cmake sometimes crashes
```
yungyuc, line 264: At the time being, macos-14 image does not have the directory `/usr/local/include`.  It is not clear whether macos-13 image has it.  Create the directory regardless.

If the directory is not there, `make gtest` on macos-14 fails with the error messages saying cmake cannot find that directory.

contrib/dependency/install.sh
```diff
@@ -38,28 +38,36 @@ install() {
 
 pybind11() {
 
-  cmakeargs=("-DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX}")
-  cmakeargs+=("-DCMAKE_BUILD_TYPE=Release")
+  cmakeargs=("-DCMAKE_BUILD_TYPE=Release")
+  if [ -z "${NO_INSTALL_PREFIX}" ] ; then
+  cmakeargs+=("-DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX}")
```
yungyuc, line 43: Allow to skip setting `CMAKE_INSTALL_PREFIX`.

contrib/dependency/install.sh
```diff
@@ -38,28 +38,36 @@ install() {
...
+  fi
+  cmakeargs+=("-DUSE_PYTHON_INCLUDE_DIR=ON")
```
yungyuc, line 45: Use the Python include path instead of the standard cmake install path.

yungyuc (commented):
This update should fix the mac CI runs.  I'll merge once the runs turn green.

.github/workflows/lint.yml
```diff
@@ -162,15 +168,13 @@ jobs:
       - name: event name
         run: |
           echo "github.event_name: ${{ github.event_name }}"
+          # Some mac runner does not have /usr/local/include and cmake sometimes crashes
+          sudo mkdir -p /usr/local/include
```
tigercosmos, line 172: surprise

PR #310, opened 2024-04-14
remove DEBUG_BUILD_OPTIMIZATION_ONE

### LINKED ISSUE #283: Bad assembly with SimpleArray caster
In the [SimpleArrayCaster.hpp](https://github.com/solvcon/modmesh/pull/266) file, a custom caster for **SimpleArray** was defined to cast from **SimpleArrayPlex** to **SimpleArray**. By default, the debug build uses the **-O0** flag and the binary is incorrect. The compiler ignores the custom caster and opts for the generic caster instead. Setting the optimization level to **-O1** or higher leads to a correct binary and the custom caster works correctly.

This anomaly was identified on Windows (tested with MSVC 2022) and macOS (tested with LLVM 15 and GCC 13 toolchains). Notably, Ubuntu 22.04 with GCC 11 did not exhibit this issue.

The issue was initially observed in [PR#266](https://github.com/solvcon/modmesh/pull/266), and a concise investigation report is available [here](https://github.com/solvcon/modmesh/pull/266#issuecomment-1946362573).

As a temporary solution, a workaround involves adding a flag, **DEBUG_BUILD_OPTIMIZATION_ONE**, in CMake to ensure successful CI passes on Windows and macOS by applying **-O1** for the debug build. By default, this flag should be set to **OFF** to reproduce the issue. Developers are advised to apply this CMake flag to ensure correct behavior in their local environments, aligning with CI expectations.

tigercosmos:
To address #283, two steps should be applied.
- put the `SimpleArrayCaster.hpp` into `array_common.hpp` (check more at the comments in the code)
- refactored `wrap_modifiers()`, which leads some type issues for pybind11 when we introduce customized caster for the array.

close #283

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -477,26 +477,24 @@ jobs:
           Get-Command clang-tidy
           Get-Command flake8
 
-      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=OFF DEBUG_BUILD_OPTIMIZATION_ONE=ON
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=OFF
         run: |
           cmake `
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
-            -DDEBUG_BUILD_OPTIMIZATION_ONE=ON `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
           cmake --build ${{ github.workspace }}/build `
             --config ${{ matrix.cmake_build_type }} `
             --target ALL_BUILD
 
-      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=ON DEBUG_BUILD_OPTIMIZATION_ONE=ON
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=ON
         run: |
           Remove-Item -Path ${{ github.workspace }}/build -Recurse -Force
           cmake `
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
             -DUSE_PYTEST_HELPER_BINDING=ON `
-            -DDEBUG_BUILD_OPTIMIZATION_ONE=ON `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
           cmake --build ${{ github.workspace }}/build `
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,11 +31,6 @@ option(LINT_AS_ERRORS "clang-tidy warnings as errors" OFF)
 # - Address Sanitization: Protecting against address-related vulnerabilities. See https://clang.llvm.org/docs/AddressSanitizer.html
 option(USE_SANITIZER "use sanitizer (undefined, leak, address)" OFF)
 
-# FIXME: A workaround flag, check more on https://github.com/solvcon/modmesh/issues/283
-# For macOS and Windows development, please turn on this flag to avoid the issue.
-option(DEBUG_BUILD_OPTIMIZATION_ONE "use -O1 for debug build. This is a workaround for issue#283." OFF)
-message(STATUS "DEBUG_BUILD_OPTIMIZATION_ONE: ${DEBUG_BUILD_OPTIMIZATION_ONE}")
-
 option(USE_PYTEST_HELPER_BINDING "use helper bindings to run pytest. should be ON if the build is for pytest" OFF)
 if(USE_PYTEST_HELPER_BINDING)
     add_definitions(-D USE_PYTEST_HELPER_BINDING)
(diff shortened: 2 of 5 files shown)


.github/workflows/devbuild.yml
```diff
@@ -477,26 +477,24 @@ jobs:
...
             -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
-            -DDEBUG_BUILD_OPTIMIZATION_ONE=ON `
             -S${{ github.workspace }} `
             -B${{ github.workspace }}/build
           cmake --build ${{ github.workspace }}/build `
             --config ${{ matrix.cmake_build_type }} `
             --target ALL_BUILD
 
-      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=ON DEBUG_BUILD_OPTIMIZATION_ONE=ON
+      - name: cmake ALL_BUILD USE_PYTEST_HELPER_BINDING=ON
```
tigercosmos, line 491: removed all `DEBUG_BUILD_OPTIMIZATION_ONE` in the CI

CMakeLists.txt
```diff
@@ -97,20 +92,8 @@ message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
 if(DEBUG_SYMBOL)
     if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /DEBUG")
-        if(DEBUG_BUILD_OPTIMIZATION_ONE)
-            # MSVC optimization flags are different from GCC/Clang, so there is no -O1 in MSVC.
```
tigercosmos, line 101: Deleting these is a bit regrettable

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -33,6 +33,14 @@
 #include <modmesh/buffer/SimpleArray.hpp>
 #include <modmesh/buffer/pymod/TypeBroadcast.hpp>
 
+// We faced an issue where the template specialization for the caster of
+// SimpleArray<T> doesn't function correctly on both macOS and Windows.
+// While the root cause of the problem remains unclear, a workaround is
+// available by including the caster header in this file, impacting
+// wrap_SimpleArray.cpp.
+// See more details in the issue: https://github.com/solvcon/modmesh/issues/283
+#include <modmesh/buffer/pymod/SimpleArrayCaster.hpp>
```
tigercosmos, line 42: please see the comments.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -141,24 +141,15 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def_property_readonly("nbody", &wrapped_type::nbody)
             .def_property_readonly("plex", [](wrapped_type const & arr)
                                    { return pybind11::cast(SimpleArrayPlex(arr)); })
-            .wrap_modifiers()
+            .def(
+                "fill", [](wrapped_type & arr, value_type const value)
```
tigercosmos: it's interesting that we need to write in this way to make pybind11 work correctly.

cpp/modmesh/python/common.hpp
```diff
@@ -285,15 +285,15 @@ WrapBase
     wrapper_type & METHOD(Args &&... args)                                    \
     {                                                                         \
         m_cls.METHOD(std::forward<Args>(args)...);                            \
-        return *static_cast<wrapper_type *>(this);                            \
+        return *static_cast<std::add_pointer_t<wrapper_type>>(this);          \
```
tigercosmos: use modern C++

tigercosmos:
@yungyuc ready for review.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: This is interesting, and I used some similar workarounds before with boost.python.

Does using lambda resolve the bad assembly?  If yes, please add comments here to explain to future maintainers.  Also, try to keep `wrap_modifiers()`.

cpp/modmesh/python/common.hpp
yungyuc: If this does nothing to the resolution of the bad assembly, please avoid the refactoring change in this **changeset**.

You may use another refactoring PR.

yungyuc (commented):
Thanks for the update.

- [x] Discuss lambda.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
tigercosmos: sure, kept `wrap_modifiers` and used lambda inside.

cpp/modmesh/python/common.hpp
tigercosmos: sure

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
tigercosmos: 1. The lambda isn't responsible for the problematic assembly; instead, putting `SimpleArrayCaster.hpp` into `array_common.hpp` addresses it (I don't know why this works, to be honested)
2. Upon integrating `SimpleArrayCaster.hpp` into wrap_`SimpleArray.cpp`, a lambda becomes necessary to rectify the erroneous implicit casting in Pybind11. Without the lambda, Pybind11 fails to build due to certain type issues, albeit not explored in depth.

tigercosmos:
@yungyuc  fixed the issues above.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -152,7 +152,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         namespace py = pybind11; // NOLINT(misc-unused-alias-decls)
 
         (*this)
-            .def("fill", &wrapped_type::fill, py::arg("value"))
+            .def(
+                // Use lambda to avoid bad implicit conversion of SimpleArray<T> in pybind11
```
yungyuc: Please add the issue #283 in the comment for easy reference.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -152,7 +152,13 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         namespace py = pybind11; // NOLINT(misc-unused-alias-decls)
 
         (*this)
-            .def("fill", &wrapped_type::fill, py::arg("value"))
+            .def(
+                // Use lambda to avoid bad implicit conversion of SimpleArray<T>in pybind11
+                // (see https://github.com/solvcon/modmesh/issues/283)
```
yungyuc, line 157: I tweaked the comment with the link to the issue #283 added.

yungyuc (approved):
LGTM.  Waiting for CI and I will merge.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Oh this is even more interesting.  There is definitely something we missed or misunderstood in the language and/or the compiler.  But pragmatism beats purity.  Let's get this to work and keep the experience.  Maybe someday the Lord will shed lights on this matter.

This is good finding and description, and will be a good reference in the future.

tigercosmos:
thanks @yungyuc

PR #311, opened 2024-04-14
use modern C++ std::add_pointer_t in `cpp/modmesh/python/common.hpp`

tigercosmos:
instead of using `T*` in the template, use the modern `std::add_pointer_t` in `cpp/modmesh/python/common.hpp` file.

### DIFF
--- a/cpp/modmesh/python/common.hpp
+++ b/cpp/modmesh/python/common.hpp
@@ -285,7 +285,7 @@ public:
     wrapper_type & METHOD(Args &&... args)                                    \
     {                                                                         \
         m_cls.METHOD(std::forward<Args>(args)...);                            \
-        return *static_cast<wrapper_type *>(this);                            \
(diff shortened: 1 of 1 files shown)


tigercosmos:
@yungyuc  ready for review.

cpp/modmesh/python/common.hpp
```diff
@@ -285,15 +285,15 @@ WrapBase
     wrapper_type & METHOD(Args &&... args)                                    \
     {                                                                         \
         m_cls.METHOD(std::forward<Args>(args)...);                            \
-        return *static_cast<wrapper_type *>(this);                            \
```
yungyuc, line 288: `wrapper_type *` is more readable.  But it's fine to use `add_pointer_t`.

yungyuc (approved):
It's not wrong.

PR #313, opened 2024-04-16
implement remaining methods in the wrapper array

tigercosmos:
- nghost
- min
- max
- sum
- abs

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -128,8 +128,9 @@ public:
 
     using value_type = typename internal_types::value_type;
 
-    value_type min(value_type initial = std::numeric_limits<value_type>::max()) const
+    value_type min() const
     {
+        value_type initial = std::numeric_limits<value_type>::max();
         auto athis = static_cast<A const *>(this);
         for (size_t i = 0; i < athis->size(); ++i)
         {
@@ -141,8 +142,9 @@ public:
         return initial;
     }
 
-    value_type max(value_type initial = std::numeric_limits<value_type>::lowest()) const
+    value_type max() const
     {
+        value_type initial = std::numeric_limits<value_type>::lowest();
         auto athis = static_cast<A const *>(this);
         for (size_t i = 0; i < athis->size(); ++i)
         {
@@ -154,8 +156,9 @@ public:
         return initial;
     }
 
-    value_type sum(value_type initial = 0) const
+    value_type sum() const
     {
+        value_type initial = 0;
         auto athis = static_cast<A const *>(this);
         if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
         {
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -143,7 +143,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                                    { return pybind11::cast(SimpleArrayPlex(arr)); })
             .wrap_modifiers()
             .wrap_calculators()
-            //
+            // ATTENTION: always keep the same interface between WrapSimpleArrayPlex and WrapSimpleArray
             ;
     }
 
@@ -170,9 +170,9 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         namespace py = pybind11; // NOLINT(misc-unused-alias-decls)
 
         (*this)
-            .def("min", &wrapped_type::min, py::arg("initial") = std::numeric_limits<value_type>::max())
-            .def("max", &wrapped_type::max, py::arg("initial") = std::numeric_limits<value_type>::lowest())
-            .def("sum", &wrapped_type::sum, py::arg("initial") = 0)
+            .def("min", &wrapped_type::min)
+            .def("max", &wrapped_type::max)
+            .def("sum", &wrapped_type::sum)
             .def("abs", &wrapped_type::abs)
             //
             ;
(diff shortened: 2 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -128,8 +128,9 @@ class SimpleArrayMixinCalculators
 
     using value_type = typename internal_types::value_type;
 
-    value_type min(value_type initial = std::numeric_limits<value_type>::max()) const
```
tigercosmos, line 131: the old function increases the difficulty of implementing the method in WrapSimpleArrayPlex.cpp. I checked the usage cases in modmesh, I don't think we need to have an argument here. We can add the argument in the future if needed.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -128,8 +128,9 @@ class SimpleArrayMixinCalculators
...
+    value_type min() const
     {
+        value_type initial = std::numeric_limits<value_type>::max();
```
tigercosmos, line 133: instead, putting `initial` in the function.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -196,6 +196,10 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
 
     friend root_base_type;
 
+#define DECL_MM_EXECUTE_TYPED_ARRAY_METHOD(typed_array_method)          \
```
tigercosmos, line 199: this macro can saves more lines.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -238,18 +242,9 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
-                                         { return array.nbytes(); }); })
-            .def_property_readonly("size", [](wrapped_type & self)
-                                   { return execute_callback_with_typed_array(
-                                         self, [](auto & array)
-                                         { return array.size(); }); })
-            .def_property_readonly("itemsize", [](wrapped_type & self)
-                                   { return execute_callback_with_typed_array(
-                                         self, [](auto & array)
-                                         { return array.itemsize(); }); })
+            .def_property_readonly("nbytes", DECL_MM_EXECUTE_TYPED_ARRAY_METHOD(nbytes))
```
tigercosmos, line 245: much more cleaner

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -300,16 +292,64 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
...
+            ;
+
+        return *this;
+    }
+
+    wrapper_type & wrap_calculators()
+    {
+#define DECL_MM_EXECUTE_TYPED_ARRAY_METHOD_RETUN_TYPED_VALUE(typed_array_method) \
+    [](wrapped_type & self) { return execute_callback_with_typed_array(          \
+                                  self, [](auto & array) { return pybind11::cast(array.typed_array_method()); }); }
```
tigercosmos, line 341: the key is using `pybind11::cast` for the return type to avoid type differences.

tests/test_buffer.py
```diff
@@ -836,6 +836,28 @@ def test_SimpleArrayPlex_properties(self):
         self.assertEqual(len(sarr), 2 * 3 * 4)  # number of elements
         self.assertEqual(sarr.nbody, 2 - 0)
         self.assertEqual(sarr.has_ghost, False)
+        self.assertEqual(sarr.nghost, 0)
+        
+    def test_minmaxsum(self):
```
tigercosmos, line 841: borrow from `SimpleArrayTC`, just to make sure the methods can work.

tigercosmos:

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: The default argument is irrelevant in most use cases.  When we need it, it's also trivial to add an argument back through overloading or something.

The change here is benign and minor.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -143,7 +143,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                                    { return pybind11::cast(SimpleArrayPlex(arr)); })
             .wrap_modifiers()
             .wrap_calculators()
-            //
+            // ATTENTION: always keep the same interface between WrapSimpleArrayPlex and WrapSimpleArray
```
yungyuc, line 146: I love this.  Useful comment.  Sweet.

In the future, the words will be turned into a structure/pattern by the power of open source.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: It's not easy to tell if you are overusing the macros.  We'll see.  Disassemble them back into repeating characters is not difficult, anyway.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
yungyuc: Could you please double check if clang-format is doing a right thing here?  If it is, leave it here.  The format does not look nice, while tolerable.

tests/test_buffer.py
yungyuc: It is absolutely the right thing to do to copy (testing) code here.

yungyuc (approved):
LGTM.  Merging.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: @yungyuc  clang-format seems not working well for all the cases in which we put the lambda in the pybind11 class definition.

Issue #302, opened 2024-04-02
Plan to improve the profiler


yungyuc:
It has been [discussed](https://discord.com/channels/730297880140578906/1220627986965856277) that modmesh can use enhancements to [the current prototype of the profiling systems](https://github.com/solvcon/modmesh/blob/69b7e899bcb554cf6068951f6c79048017e2b7fe/cpp/modmesh/toggle/RadixTree.hpp#L178).

Some working items are identified:

1. We should start with integrating the profiling of Python and C++.  If possible, we can try to merge the call stacks from both Python and C++ in the profiling entries.
2. To assist the development, we will visualize the profile results held in the data structures in memory.
3. Develop serialization (and deserialization) of the profiling data structures.  It allows preserving the results for future inspection (visualization in UI), and sharing among contributors.

These ideas are contributed by @tigercosmos @q40603 @kuomc and others.  An older discussion can be found at #135 .  We will need more discussions and prototypes to move it forward.

PR #315, opened 2024-04-20
Add Python wrap to `CallProfiler`

### LINKED ISSUE #344: Wrap the C++ profile object (CallProfile) to Python
This issue derives from https://github.com/solvcon/modmesh/issues/302.

Make the C++ [`CallProfile`](https://github.com/solvcon/modmesh/blob/69b7e899bcb554cf6068951f6c79048017e2b7fe/cpp/modmesh/toggle/RadixTree.hpp#L178) accessible from Python.

q40603:
Following the issue : https://github.com/solvcon/modmesh/issues/302
This PR includes :
1. Wrap CallProfiler to python
2. Use decorator to wrap the python function with call_profiler_probe to get the profiling result
3. Convert RadixTree to python dict so user can better access the profiling result.


cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -225,17 +225,16 @@ class CallProfiler
     ~CallProfiler() = default;
 
     // Called when a function starts
-    void start_caller(const std::string & caller_name, std::function<void()> cancel_callback)
+    void start_caller(const std::string & caller_name)
```
q40603, line 228: remove unused parameter

yungyuc:
Thanks, @q40603 .  Before the PR should be reviewed, please make all CI checks passing.  After that, please annotate (using PR in-line comments) and make a global comment stating the readiness of review.

If there are issues resolving the CI failures, we may discuss here or on [discord](https://discord.gg/P9U7PFv).

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -51,8 +51,8 @@ class RadixTreeNode
     static_assert(std::is_integral_v<key_type> && std::is_signed_v<key_type>, "signed integral required");
 
     RadixTreeNode(std::string const & name, key_type key)
-        : m_name(name)
-        , m_key(key)
+        : m_key(key)
```
q40603, line 57: fix Wreorder-ctor error

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -225,17 +225,16 @@ class CallProfiler
     ~CallProfiler() = default;
 
     // Called when a function starts
-    void start_caller(const std::string & caller_name, std::function<void()> cancel_callback)
```
q40603, line 242: remove unused arg

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -265,14 +264,10 @@ class CallProfilerProbe
 {
 public:
     CallProfilerProbe(CallProfiler & profiler, const char * caller_name)
-        : m_profiler(profiler)
-        , m_caller_name(caller_name)
+        : m_caller_name(caller_name)
```
q40603, line 307: fix Wreorder-ctor error

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -225,17 +225,16 @@ class CallProfiler
...
     {
-        auto start_time = std::chrono::high_resolution_clock::now();
         m_radix_tree.entry(caller_name);
         CallerProfile & callProfile = m_radix_tree.get_current_node()->data();
         callProfile.caller_name = caller_name;
         callProfile.start_stopwatch();
     }
 
     // Called when a function ends
-    void end_caller(const std::string & caller_name)
```
q40603, line 253: remove unused arg

cpp/modmesh/toggle/pymod/wrap_profile.cpp
```diff
@@ -163,12 +163,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapTimeRegistry
 
 }; /* end class WrapTimeRegistry */
 
+class WrapCallProfiler : public WrapBase<WrapCallProfiler, CallProfiler>
```
q40603: Expose CallProfiler to python so profiling util can be shared.
RadixTree is internal data structure, hence no need to expose this to python

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+
+        @profile_function
+        def foo1():
+            time.sleep(1)
+
+        modmesh.call_profiler.reset()
+        foo1()
+        result = modmesh.call_profiler.print_profiling_result()
+        pattern = r"([-+]?\d*\.\d+|\d+) ms"
+        match = re.search(pattern, result)
```
q40603: Not sure if this is a good practice to extract value from the string and assert the value

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+            time.sleep(3)
+
+        foo1()
+        modmesh.call_profiler.reset()
+        result = modmesh.call_profiler.print_profiling_result()
+        self.assertIn("Profiling Result", result)
+        self.assertNotIn("foo1", result)
+
+    @unittest.skipUnless("nt" != os.name,
+                         "timing code on windows has resoltion issue")
```
q40603: Notice there's time resolution issue in windows, causing the CI to fail as the measured time is always less than what it should be.
Wonder if there's a better way to handle this on windows.

q40603 (commented):
@yungyuc THE PR is ready for review.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+
+import unittest
+import time
+import re
+import os
+from functools import wraps
+import modmesh
+
+
+# https://docs.python.org/3/library/functools.html#functools.wraps
```
yungyuc: Please remove the comment pointing to Python document.  This is a simple helper in Python standard library that every Python developer should be familiar with, so I expect all contributors to modmesh to be familiar with it.  We should not add comment for people who do not know how to learn easy things themselves.

On the other hand, if there are tricky interaction among popular API, or obscure bug, we usually add comments pointing to bug trackers.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+def profile_function(func):
```
yungyuc, line 43: This will be a helpful utility function for developing modmesh code, e.g., GUI (FYI @j8xixo12).  Eventually I think a more tailored implementation should be made available somewhere under the `modmesh` Python package.

At the time being, could you please add a comment for the above-mentioned argument?

When it comes to the to-be-useful-and-mature helper function in the future, I don't like the function name (`profile_function`).  But it is OK for a local helper in the test module.

tests/test_callprofiler.py
yungyuc: Much much less it is in windows than in other systems?

Preferably this can be fixed (or worked around) in this PR.  We use the `skipUnless` trick if can't do it soon enough.

tests/test_callprofiler.py
yungyuc: No it's not.  Please add relevant data accessors in `CallProfiler` for the profiling data.

By writing tests, we will discover many things for improving the API (system) omitted when during the initial design.  This "design-for-testing" is a powerful techniques for software development too, although it is not exactly the same as how it is used in IC design.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+
+        foo1()
+        modmesh.call_profiler.reset()
+        result = modmesh.call_profiler.print_profiling_result()
+        self.assertIn("Profiling Result", result)
+        self.assertNotIn("foo1", result)
+
+    @unittest.skipUnless("nt" != os.name,
+                         "timing code on windows has resoltion issue")
+    def test_single_fun_profile(self):
```
yungyuc: I suggest to use `func` as a shorthand for `function`, but it's just my personal preference.  I kind of feel `fun` is for fun.

You don't have to change if you don't feel it the same way.

yungyuc (changes requested):
- [x] Remove functool comment in the test code.
- [ ] Add comment for the future plan of `profile_function`.
- [x] Use a little bit of time to resolve the windows timing issue for testing.
- [x] Implement basic data access in `CallProfiler` for testing.

Again, please add in-line annotation to mark the points that you address or would like to discuss, and when everything is ready for review, leave a global comment.  Combining the in-line comments and the global using the *code review* section/feature/whatever may be a good idea.  Let's use this protocol from now on.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -265,14 +264,10 @@ class CallProfilerProbe
...
-        : m_profiler(profiler)
-        , m_caller_name(caller_name)
+        : m_caller_name(caller_name)
+        , m_profiler(profiler)
     {
-        auto cancel_callback = [&]()
-        {
-            cancel();
-        };
-        m_profiler.start_caller(m_caller_name, cancel_callback);
```
tigercosmos: why do you remove the `cancel_ballback`?

cpp/modmesh/toggle/pymod/wrap_profile.cpp
```diff
@@ -163,12 +163,43 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapTimeRegistry
...
+        namespace py = pybind11;
+
+        (*this)
+            .def_property_readonly_static(
+                "instance",
+                [](py::object const &) -> wrapped_type &
+                { return wrapped_type::instance(); })
+            .def("start_caller", &wrapped_type::start_caller, py::arg("caller_name"))
+            .def("end_caller", &wrapped_type::end_caller)
+            .def("print_profiling_result", [](const CallProfiler & profiler)
```
tigercosmos: maybe consider taking a stream object as the argument.

cpp/modmesh/toggle/toggle.hpp
```diff
@@ -31,6 +31,7 @@
 #include <modmesh/base.hpp>
 #include <modmesh/buffer/buffer.hpp>
 #include <modmesh/toggle/profile.hpp>
+#include <modmesh/toggle/RadixTree.hpp>
```
tigercosmos, line 34: do we need to put the include here?

tests/test_callprofiler.py
tigercosmos: can we have a workaround for Windows? e.g. increasing the time or enlarging the error tolerance?

tests/test_callprofiler.py
tigercosmos: `func` makes more sense. If possible, I prefer spelling out the full name.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,106 @@
...
+
+        modmesh.call_profiler.reset()
+        foo3()
+        result = modmesh.call_profiler.print_profiling_result()
+
+        foo3_pattern = r"foo3 - Total Time: ([-+]?\d*\.\d+|\d+) ms"
+        match = re.search(foo3_pattern, result)
+        self.assertIsNotNone(match, result)
+        foo3_total_time = float(match.group(1))
+        self.assertGreater(foo3_total_time, 3000)
```
tigercosmos: the number can be fixed. please refer to `test_nopython_radixtree.cpp`

cpp/modmesh/toggle/RadixTree.hpp
q40603: @tigercosmos 
Because the compiler raised unused variable error.
Also, I am not sure about the true behavior of this callback.
Could you please elaborate more or maybe kindly add some tests in gtests/test_nopython_callprofiler.cpp?

tigercosmos: @q40603 Originally, `cancel_callback` is for canceling the profiling. For example, if we restart the profiler, we need to cancel the previous profiling, otherwise the profiler will have unexpected behavior. Surprisingly, I didn't fully support the cancel mechanism before. Let me file another PR for this.

tests/test_callprofiler.py
tigercosmos: sorry I am wrong here, should be `test_nopython_callprofiler.cpp`

cpp/modmesh/toggle/RadixTree.hpp
tigercosmos: please check the PR:
https://github.com/solvcon/modmesh/pull/318

q40603:
I will try to finalize this PR in 2 days.

@yungyuc @tigercosmos 
As cancel callback feature has been implemented by @tigercosmos  in PR [318](https://github.com/solvcon/modmesh/pull/318). I found it hard to wrap the call_profiler using pybind11 because of the cancell_callback argument.

Hence, I tried to wrap the [CallProfilerProbe](https://github.com/solvcon/modmesh/pull/315/files#diff-9bb5cf9d21b507f6d20b5ee1c2b45ac4669440b0ca1e0d135ef8ea819c98bdd8R200) instead, but I ran into a issue.

Consider the following 2 code snippets:
with the difference modmesh.CallProfilerProbe is assigning to a variable or not

```python
def CallProfileProbeA(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        modmesh.CallProfilerProbe(func.__name__)
        result = func(*args, **kwargs)
        return result
    return wrapper
```

```python
def CallProfileProbeB(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        tmp = modmesh.CallProfilerProbe(func.__name__)
        result = func(*args, **kwargs)
        return result
    return wrapper
```

and a simple python code that drives the Profiler

```python
@CallProfileProbeA
def foo1():
    time.sleep(2)

@CallProfileProbeA
def foo2():
    time.sleep(3)
    foo1()


if __name__ == '__main__':
    foo2()
    print(modmesh.call_profiler.print_profiling_result())
```

The two different implementations gave different result

1. CallProfileProbeA - unexpected
```
Profiling Result
  foo2 - Total Time: 0.002378 ms, Call Count: 1
  foo1 - Total Time: 0.018734 ms, Call Count: 1
```

2. CallProfileProbeB - expected
```
Profiling Result
  foo2 - Total Time: 5002.27 ms, Call Count: 1
    foo1 - Total Time: 2001.02 ms, Call Count: 1
```

For the CallProfileProbeA, the CallProfilerProbe object is deleted almost immediately because of the python garbage collection, reference counting of CallProfilerProbe is zero.  (Spent a few hours debugging why CallProfileProbeA doesn't work 😂, also tried to implement CallProfileProbe as context manager, but give up.)

I just wonder if there's a more elegant way for the CallProfilerProbe object to be persisted before the wrapper function return?

yungyuc:
@tigercosmos why did a helper in a profiler need to use callback?  Callback is generally hard to understand and maintain.  It does not sound like necessary.

Is it possible to implement the similar capability without a callback?  Or is there a strong reason to use a callback?

tigercosmos:
@q40603  @yungyuc
I utilized a callback in this context to invoke the cancellation function of the probe, without exposing its entirety. However, if the callback presents an issue, we can modify the logic to pass the probe pointer to `CallProfiler`, so it can call `probe->cancel` later.

yungyuc:
Thanks, @q40603 @tigercosmos .  Now there seem to be two things:

1. Avoid using callback in the profiler.
2. Wrap the profiler to Python while there is the cancellation callback.

Can they be handled independently?  If so, let's do it this way.

If not, what is the dependency?

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: Please don't mark conversation as resolved.  If you want to say it is resolved, leave a comment saying so.

Marking conversation as resolved makes it hard to keep track of discussion history.

tigercosmos:
If they are dependent, in my opinion, we can refactor the callback to another design, for now maybe we can skip supporting `cancel()` in Python if it has an issue.

yungyuc:
> If they are dependent, in my opinion, we can refactor the callback to another design, for now maybe we can skip supporting `cancel()` in Python if it has an issue.

Thanks, @tigercosmos .  Does that mean that @q40603 can continue the profiler wrapper work by skipping the wrapper of `cancel()`?

tigercosmos:
@yungyuc yes, I think the cancel function is not the priority.

yungyuc:
> @yungyuc yes, I think the cancel function is not the priority.

Thanks for the confirmation.  I think I know what you meant now.

@q40603 do you have sufficient information to proceed?

q40603:
@yungyuc @tigercosmos 
Thanks for the comments. 
I will skip cancel callback at this moment and implement the CallProfiler data access ASAP, which will need to expose the RadixTree to python as well.

tigercosmos:
btw, can we rename the file to `CallProfiler.h` rather than `RadixTree.h`? RadixTree is the internal data structure, and the CallProfiler is the public API. @q40603  @yungyuc

yungyuc:
> btw, can we rename the file to `CallProfiler.h` rather than `RadixTree.h`? RadixTree is the internal data structure, and the CallProfiler is the public API. @q40603 @yungyuc

@tigercosmos feel free to create a refactoring PR for discussing the proposal.  It may conflict with what @q40603 is doing, but discussing with code is still better than only words.

yungyuc:
@q40603 Is there progress on this PR?

q40603:
I am revising the data accessor of profiling result by making it returned in nested python dictionary format.
The dict can then be used with better assertion. 
Some fields are also required to add to distinguish where is profiling probe called, like in python or in c++.

yungyuc:
It could be an issue to use Python dict for storing the data in the long run.  It is slow.  But we can do it in the short term like in this PR.  Performance issues need to be addressed by profiling.

q40603:
The python dict is only used for the user to retrieve the profiling result on the python side.
This PR has been opened for a while and I will make the code review ready by this Saturday (2024/06/15).

yungyuc:
> The python dict is only used for the user to retrieve the profiling result on the python side. This PR has been opened for a while and I will make the code review ready by this Saturday (2024/06/15).

"The python dict is ... used for the user to retrieve the profiling result" is the problem.  It is slow and will bite us soon.  But at this moment let us focus on getting things done.  Then it will become a baseline for improvement.

We just need to keep in mind that a slow profiler (including data retrieval) will be a useless profiler because it will distort the performance by too much.

tests/data/profiler_python_schema.json
```diff
@@ -0,0 +1,41 @@
+{
```
q40603, line 1: Add schema validation in unittest using jsonschema library.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+    def wrapper(*args, **kwargs):
+        _ = modmesh.CallProfilerProbe(func.__name__)
+        result = func(*args, **kwargs)
+        return result
+    return wrapper
+
+
+def busy_loop(duration):
+    end_time = time.perf_counter() + duration
+    while time.perf_counter() < end_time:
```
q40603, line 55: use time.perf_counter() to get higher time resolution instead of time.time()

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+        @profile_function
+        def foo1():
+            busy_loop(0.01)
+
+        foo1()
+        modmesh.call_profiler.reset()
+        result = modmesh.call_profiler.result()
+        self.assertEqual(result, {})
+
+    def test_profiler_result_schema(self):
```
q40603, line 75: add schema validation using jsoncshema library.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+        self.assertGreaterEqual(foo_result["total_time"], 600)
+        self.assertEqual(len(foo_result["children"]), 1)
+        self.assertEqual(foo_result["children"][0]["name"], "bar")
+
+        second_bar_result = foo_result["children"][0]
+        self.assertEqual(len(second_bar_result["children"]), 0)
+        self.assertEqual(second_bar_result["count"], 1)
+        self.assertGreaterEqual(second_bar_result["total_time"], 500)
+
+    def test_get_result_during_profiling(self):
```
q40603, line 174: As the previous unit test all assert the result after profiling a series of function calls.
But user should be able to get the profiling result within the function too, so add this test.

tests/data/profiler_python_schema.json
q40603, line 1: Simple json schema that regulates the output python dict format.

cpp/modmesh/toggle/pymod/wrap_profile.cpp
```diff
@@ -163,16 +165,99 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapTimeRegistry
...
+
+        (*this)
+            .def_property_readonly_static(
+                "instance",
+                [](py::object const &) -> wrapped_type &
+                { return wrapped_type::instance(); })
+            .def("result", [](CallProfiler & profiler)
+                 {
+            const RadixTreeNode<CallerProfile> * root = profiler.radix_tree().get_root();
+            if (root->no_children()) {
```
q40603: To avoid root only dict is returned when the radix tree is empty.

q40603 (commented):
> * [x]  Remove functool comment in the test code.
> * [ ]  Add comment for the future plan of `profile_function`.
> * [x]  Use a little bit of time to resolve the windows timing issue for testing.
> * [x]  Implement basic data access in `CallProfiler` for testing.

@yungyuc, @tigercosmos  The issues are addressed and the code is ready for review.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -72,6 +73,11 @@ class RadixTreeNode
     const T & data() const { return m_data; }
     const child_list_type & children() const { return m_children; }
 
+    bool no_children() const
```
yungyuc: There are two options for this getter: `no_children()` or `has_children()`.

`no_children()` acts like the implementation detail `m_children.empty()`.  If we choose this, what is the reason that we do not name the function as `empty_children()`?  `empty` is more consistent to the style of STL.

Please consider the options and we should discuss.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
```
yungyuc, line 25: Need two blank lines separating comment block and code.

tests/test_callprofiler.py
yungyuc: I am not seeing the descriptions in comment.  Please point it out.

tests/test_callprofiler.py
yungyuc: Sounds good.

yungyuc (changes requested):
We may use some discussions, which can take place here or on the discord.

- [x] Review and correct coding style
- [x] Missing comments for the future plan of `profile_function()`.

tests/test_callprofiler.py
tigercosmos: cc @ThreeMonth03

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+        _ = modmesh.CallProfilerProbe(func.__name__)
+        result = func(*args, **kwargs)
+        return result
+    return wrapper
+
+
+def busy_loop(duration):
+    end_time = time.perf_counter() + duration
+    while time.perf_counter() < end_time:
+        time.sleep(0.00001)
```
tigercosmos: `sleep` is usually not precise in a very short time. why not just use `pass` for the busy loop?

tests/test_callprofiler.py
```diff
@@ -0,0 +1,185 @@
...
+def profile_function(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        _ = modmesh.CallProfilerProbe(func.__name__)
+        result = func(*args, **kwargs)
+        return result
+    return wrapper
+
+
+def busy_loop(duration):
```
tigercosmos, line 53: let's have a comment to explain the time unit.

tests/data/profiler_python_schema.json
tigercosmos: not sure if the schema should be put in this path.

we should also mention this schema in the cpp implementation.

cpp/modmesh/toggle/RadixTree.hpp
q40603: I agree that no_chlidren should be renamed to empty_children, no_children is too verbal.
But I am curios, in practical, how to choose between negative check and positive check?
like empty_children or has_children in this case

tests/test_callprofiler.py
q40603: Yes, I think pass would be more straightforward.
Originally, the sleep is just for reducing the CPU usage.

cpp/modmesh/toggle/RadixTree.hpp
yungyuc: Time will tell.

q40603:
> We may use some discussions, which can take place here or on the discord.
> 
> * [x]  Review and correct coding style
> * [x]  Missing comments for the future plan of `profile_function()`.

Issues addressed.
Regarding profile_function(), here is the future plan:
- Implement profile_funcion somewhere else under python modmesh lib
- Add a filed in each node in the profiler to distinguish where is the profiler run, like from: C++, from: Python
- Try profile a real app that has both python and C++ codes, and add a unit test for this scenario.

tests/test_callprofiler.py
q40603: added

tests/test_callprofiler.py
yungyuc: > This will be a helpful utility function for developing modmesh code, e.g., GUI (FYI @j8xixo12). Eventually I think a more tailored implementation should be made available somewhere under the `modmesh` Python package.
> 
> At the time being, could you please add a comment for the above-mentioned argument?

By "add a comment for ...", I meant to add the comment in the code, so that a future maintainer will read it and perform the refactoring.

yungyuc (changes requested):
* [ ]  Still miss comments for the future plan of `profile_function()`.

yungyuc:
@q40603 Could you please find @terrychan999 for handling [the failure of creating artifacts in the windows CI run](https://github.com/solvcon/modmesh/actions/runs/9575702162/job/26400964857?pr=315#step:13:190)?  I think we should use some discussions to deal with it.  Please bring it to the discord channel.

q40603:
> * [x]  Still miss comments for the future plan of `profile_function()`.

Do you mean leaving comments in the code?

yungyuc:
Correct

q40603:
Comment added.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,194 @@
...
+import os
+import unittest
+import time
+import json
+from jsonschema import validate, ValidationError
+from functools import wraps
+import modmesh
+
+
+# TODO
```
q40603, line 37: Comment added.

q40603 (commented):
@yungyuc The code is ready for next review.

tests/test_callprofiler.py
yungyuc: LGTM.  Thanks.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,194 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+import os
+import unittest
+import time
+import json
+from jsonschema import validate, ValidationError
```
yungyuc: Please import the module and use `.` to access the functions in it, like:

```python
import jsonschema

...

jsonschema.validate(...)
```

Please review other Python files in the test and code directories and follow the style.  @ThreeMonth03 this may be captured in the style guide.

yungyuc (changes requested):
Almost there (while I might find more to change later):

- [x] Import the module and use . to access the functions in it.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,194 @@
...
+        foo1()
+        result = modmesh.call_profiler.result()
+
+        path = os.path.join(os.path.abspath(os.path.dirname(__file__)),
+                            "data", "profiler_python_schema.json")
+        with open(path, 'r') as schema_file:
+            schema = json.load(schema_file)
+
+        try:
+            jsonschema.validate(instance=result, schema=schema)
```
q40603, line 91: Fix module import and use dot operator.

tests/test_callprofiler.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+# TODO
+# 1. Implement profile_funcion somewhere else under python modmesh lib
+# 2. Add a filed in each node in profiler to distinguish where profiler run
+#    like from: C++, from: Python
+# 3. Try profile a real app that has both python and C++ codes
+#    and add a unit test for this scenario.
+def profile_function(func):
+    @functools.wraps(func)
```
q40603, line 44: Fix module import and use dot operator.

q40603 (commented):
> Almost there (while I might find more to change later):
> 
> * [x]  Import the module and use . to access the functions in it.

Issues addressed, the code is ready for next review.

yungyuc (approved):
Thanks a lot.  Merging.

PR #316, opened 2024-04-21
`BufferExpander::m_end` is used before initialization

yungyuc:
`BufferExpander::size()` is wrongly called when initializing `m_end`

### DIFF
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -72,7 +72,7 @@ public:
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
         : m_concrete_buffer(clone ? buf->clone() : buf)
         , m_begin(m_concrete_buffer->data())
-        , m_end(m_begin + size())
+        , m_end(m_begin + m_concrete_buffer->size())
         , m_end_cap(m_begin + m_concrete_buffer->size())
     {
     }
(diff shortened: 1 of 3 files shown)


cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -72,7 +72,7 @@ class BufferExpander
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
         : m_concrete_buffer(clone ? buf->clone() : buf)
         , m_begin(m_concrete_buffer->data())
-        , m_end(m_begin + size())
+        , m_end(m_begin + m_concrete_buffer->size())
```
yungyuc, line 75: `m_end` should use `m_concrete_buffer->size()`.  It is wrong to call `size()`, which reads `m_end` at line 96!

```cpp
    size_type size() const noexcept
    {
        return static_cast<size_type>(this->m_end - this->m_begin);
    }
```

cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp
```diff
@@ -133,6 +133,8 @@ WrapBufferExpander::WrapBufferExpander(pybind11::module & mod, char const * pyna
             py::arg("length"))
         .def_timed(py::init([]()
                             { return wrapped_type::construct(); }))
+        .def_timed(py::init([](std::shared_ptr<ConcreteBuffer> const & buf)
```
yungyuc, line 136: Expose the factory method taking a `ConcreteBuffer` to Python.

tests/test_buffer.py
```diff
@@ -139,6 +139,20 @@ def test_BufferExpander(self):
...
         self.assertEqual(list(i + 10 for i in range(10)), list(ep))
 
+    def test_BufferExpanderFromConcreteBuffer(self):
+        buf = modmesh.ConcreteBuffer(10)
+        for it in range(len(buf)):
+            buf[it] = it
+
+        ep = modmesh.BufferExpander(buf)
+        self.assertEqual(10, ep.capacity)
+        self.assertEqual(10, len(ep))
```
yungyuc, line 149: This (sometimes) fails before I add the fix in C++.

yungyuc (commented):
Fix an embarrassing bug.  @q40603 sorry I am not waiting for you.

PR #317, opened 2024-04-22
support array broadcast with list/tuple

tigercosmos:
support array broadcast with list/tuple

### DIFF
--- a/cpp/modmesh/buffer/pymod/array_common.hpp
+++ b/cpp/modmesh/buffer/pymod/array_common.hpp
@@ -95,8 +95,10 @@ public:
 
         if (args.size() == 2)
         {
+            const bool isNumber = py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) || py::isinstance<py::float_>(args[1]);
+
             // sarr[K] = V
-            if (py::isinstance<py::int_>(args[0]) && !py::isinstance<py::array>(args[1]))
+            if (py::isinstance<py::int_>(args[0]) && isNumber)
             {
                 const auto key = args[0].cast<ssize_t>();
 
@@ -104,16 +106,19 @@ public:
                 return;
             }
             // sarr[K1, K2, K3] = V
-            if (py::isinstance<py::tuple>(args[0]) && !py::isinstance<py::array>(args[1]))
+            if (py::isinstance<py::tuple>(args[0]) && isNumber)
             {
                 const auto key = args[0].cast<std::vector<ssize_t>>();
 
                 arr_out.at(key) = args[1].cast<T>();
                 return;
             }
+
+            const bool isArray = py::isinstance<py::list>(args[1]) || py::isinstance<py::array>(args[1]) || py::isinstance<py::tuple>(args[1]);
+
             // multi-dimension with slice and ellipsis
             // sarr[slice, slice, ellipsis] = ndarr
-            if (py::isinstance<py::tuple>(args[0]) && py::isinstance<py::array>(args[1]))
+            if (py::isinstance<py::tuple>(args[0]) && isArray)
             {
                 const py::tuple tuple_in = args[0];
                 const py::array arr_in = args[1];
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -95,25 +95,30 @@ class ArrayPropertyHelper
...
-            if (py::isinstance<py::tuple>(args[0]) && !py::isinstance<py::array>(args[1]))
+            if (py::isinstance<py::tuple>(args[0]) && isNumber)
             {
                 const auto key = args[0].cast<std::vector<ssize_t>>();
 
                 arr_out.at(key) = args[1].cast<T>();
                 return;
             }
+
+            const bool isArray = py::isinstance<py::list>(args[1]) || py::isinstance<py::array>(args[1]) || py::isinstance<py::tuple>(args[1]);
```
tigercosmos: missing support for tuple and list.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -95,25 +95,30 @@ class ArrayPropertyHelper
 
         if (args.size() == 2)
         {
+            const bool isNumber = py::isinstance<py::bool_>(args[1]) || py::isinstance<py::int_>(args[1]) || py::isinstance<py::float_>(args[1]);
```
tigercosmos: better way to identify a number

tests/test_buffer.py
```diff
@@ -666,6 +666,42 @@ def test_SimpleArray_broadcast_slice_ghost_md(self):
                 for k in range(4):
                     self.assertEqual(ndarr2[i, j, k], sarr[i - G, j, k])
 
+    def test_SimpleArray_broadcast_from_list_list(self):
```
tigercosmos, line 669: Is the testing code self-explaining?

tigercosmos:
@yungyuc PR is ready for review.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: It might run faster by caching `args[1]` locally in a `py::object`.  Every time you do the indexing, some Python operations incur.  But Python is already slow so this is not really a big deal.

yungyuc: But I do want local variables (e.g., `isNumber`) to use `snake_case`, which is consistent to the existing code in this function.  Please change the style.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: The variable name is misleading.  Do you meant to say `is_sequence`?  See https://docs.python.org/3/c-api/sequence.html and https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range .

Note that the sequence is not an iterable, which is related to [iterator](https://docs.python.org/3/c-api/iter.html).

tests/test_buffer.py
yungyuc: It's pretty good.  You can use 3 test functions or combine them into one.  Both are OK.  Sometimes the latter is easier to maintain.

yungyuc (changes requested):
- [x] Use `snake_case` for local variables.
- [x] Use a better name for is-sequence.

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: I thought `py::args` underlying is `py::object`, but actually it's `py::argument_record`.

Fixed in f9f90021927345266d1465e7c6169eeac2b58aad

cpp/modmesh/buffer/pymod/array_common.hpp
tigercosmos: fixed in f9f90021927345266d1465e7c6169eeac2b58aad

tigercosmos:
@yungyuc  all issues are addressed. please have a look.

yungyuc (approved):
LGTM.  Thanks, @tigercosmos .

PR #333, opened 2024-05-19
add NASA plot3d mesh format loader

### LINKED ISSUE #331: Add Plot3d mesh format loader
Add [Plot3d](https://github.com/nasa/Plot3D_utilities) mesh format loader in modmesh.
The following are the working items need to be finished:

- [x] Add Plot3d mesh file parsing logic.
- [x] Add logic to construct mesh element from Plot3d mesh file.

j8xixo12:
Adding NASA plot3d mesh format loading logic into modmesh.
The demo file can be downloaded from [here](https://turbmodels.larc.nasa.gov/NACA0012numerics_grids/n0012familyI.4.p3dfmt.gz)

**NACA0012 airfoil mesh**:
image


cpp/modmesh/inout/inout_util.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    small_vector<std::string> output;
+    std::stringstream ss(str);
+    std::string token;
+    while (std::getline(ss, token, delim))
+    {
+        output.push_back(token);
+    }
+    return output;
+}
```
j8xixo12: I moved function `tokenize` out of the `Gmsh` class and made it a general function of `inout` because I believe it will be used in parsing different mesh files.

tests/test_plot3d.py
```diff
@@ -0,0 +1,39 @@
...
+class Plot3dTC(unittest.TestCase):
+
+    def test_plot3d_parsing(self):
+
+        data = """1
+2 2 2
+0 0 0 0 1 1 1 1
+0 0 1 1 0 0 1 1
+0 1 0 1 0 1 0 1
+"""
```
j8xixo12, line 43: Adding testing p3d mesh file by string input instead of adding mesh file.

cpp/modmesh/inout/plot3d.cpp
```diff
@@ -0,0 +1,148 @@
...
+                {
+                    std::getline(stream, line);
+                    auto tokens = tokenize(line, "(-?\\d+(\\.\\d+)?([eE][+-]?\\d+)?|\\b\\d+\\b)");
+                    for (auto token : tokens) parsing_q.push(std::stod(token));
+                }
+                m_nds(j + base_idx, k) = parsing_q.front();
+                parsing_q.pop();
+            }
+        }
+    }
```
j8xixo12: I think there may be opportunities to optimize this code because there is some redundant code here.

cpp/modmesh/inout/inout_util.cpp
```diff
@@ -0,0 +1,32 @@
+#include <modmesh/inout/inout_util.hpp>
+
+namespace modmesh
+{
+namespace inout
+{
+small_vector<std::string> tokenize(const std::string & str, const std::string delim)
```
yungyuc: `small_vector` may not be performant with a large amount of elements or pointer-likes elements.  I think the return type should use STL `std::vector` in this case.

yungyuc: In the future, we can also consider develop a custom container class for the tokens.  String processing is usually slow and a custom class allows us to optimize.  (We do not need to do it right now.)

cpp/modmesh/inout/inout_util.cpp
```diff
@@ -0,0 +1,32 @@
...
+    std::sregex_iterator iter(str.begin(), str.end(), regex_delim);
+    std::sregex_iterator end;
+    while (iter != end)
+    {
+        output.push_back(iter->str());
+        ++iter;
+    }
+    return output;
+}
+small_vector<std::string> tokenize(const std::string & str, const char delim)
```
yungyuc: Don't forget a blank line to separate structural entities (functions in this case).

cpp/modmesh/inout/inout_util.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <string>
+#include <sstream>
+
+#include <modmesh/buffer/buffer.hpp>
+
+namespace modmesh
+{
+namespace inout
+{
+small_vector<std::string> tokenize(const std::string & str, const std::string delim);
```
yungyuc: In term of coding style, I prefer to leave a blank line between the namespace braces and the contents (functions, classes, etc.)  But it is just what I usually do, not a requirement.

cpp/modmesh/inout/plot3d.cpp
```diff
@@ -0,0 +1,148 @@
...
+    size_t x_base_idx = 0, y_base_idx = 0, z_base_idx = 0;
+    uint_type idx = 0;
+    uint_type nblocks = 0;
+    uint_type base_idx = 0;
+    // hexahedron has 8 nodes + 1 for store nodes number
+    small_vector<uint_type> nds_temp(9, 0U);
+    // These vector of x, y, z coordinate shift is used to build the hexahedron element
+    // the nodes order of the hexahedron element is clock of the xy face
+    // Referece: https://github.com/nasa/Plot3D_utilities/blob/main/colab/Plot3D_SplitBlocksExample.ipynb
+    small_vector<small_vector<int_type>> shift = {
```
yungyuc: `std::vector<std::vector<int_type>>` is good enough because there are only initial memory allocation.

You can also try `constexpr` if the contents are not changed later.

cpp/modmesh/inout/plot3d.cpp
yungyuc: Please leave a code comment to remind the future maintainer.

cpp/modmesh/inout/gmsh.hpp
```diff
@@ -228,18 +230,6 @@ class Gmsh
         }
     }
 
-    std::vector<std::string> tokenize(const std::string & str, const char delim)
```
yungyuc, line 231: It's good to make it a common utility.

yungyuc (changes requested):
- [x] Use `std::vector` for token return.
- [x] Need to resolve all CI failures.

cpp/modmesh/inout/plot3d.cpp
```diff
@@ -0,0 +1,150 @@
...
+        m_y_shape(i) = std::stoul(tokens[1]);
+        m_z_shape(i) = std::stoul(tokens[2]);
+        m_blk_sizes(i) = m_x_shape(i) * m_y_shape(i) * m_z_shape(i);
+        total_blk_size += m_blk_sizes(i);
+    }
+
+    m_nds.remake(small_vector<size_t>{total_blk_size, 3}, 0);
+
+    parseCoordinates(nblocks);
+    buildHexahedronElements(nblocks);
```
j8xixo12, line 69: I separated the parsing logic and element building logic into 2 functions to by pass the clang-tidy warning 
`readability-function-cognitive-complexity`.

cpp/modmesh/inout/inout_util.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <modmesh/buffer/buffer.hpp>
+
+namespace modmesh
+{
+
+namespace inout
+{
+
+std::vector<std::string> tokenize(const std::string & str, const std::regex & regex_delim);
+std::vector<std::string> tokenize(const std::string & str, const char delim);
```
j8xixo12, line 45: I have changed the return type to `std::vector<std::string>`.

j8xixo12 (commented):
- [X] Use std::vector for token return.
- [X] Need to resolve all CI failures.

@yungyuc this PR is ready for review!

cpp/modmesh/inout/inout_util.hpp
yungyuc: This is nice.

In the future, we can further enhance to use a custom string I/O facility.  String I/O is usually a performance hotspot.  A naive implementation using `std::vector` and `std::string` would not be fast.  But we should not do it without proper profiling.  Let's just keep it in mind, but refrain from doing anything now.

cpp/modmesh/inout/plot3d.cpp
yungyuc: Sounds great!

cpp/modmesh/inout/plot3d.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+void Plot3d::parseCoordinates(const uint_type nblocks)
+{
+    uint_type base_idx = 0;
+    std::string line;
+    std::queue<real_type> parsing_q;
+    const std::regex plot3d_coord_delim(R"(-?\d+(\.\d+)?([eE][+-]?\d+)?|\b\d+\b)");
+
+    // TODO: The nested for-loop needs to be enhanced
```
yungyuc, line 79: It would be better to be specific why the nested for loop should be enhanced.  But it's OK for now because it is short enough.

cpp/modmesh/inout/inout_util.cpp
```diff
@@ -0,0 +1,36 @@
+#include <modmesh/inout/inout_util.hpp>
```
yungyuc, line 29: Any license declaration in new files.  You need to use the same license term as the existing files.

tests/test_plot3d.py
```diff
@@ -0,0 +1,39 @@
+import unittest
```
yungyuc, line 27: You need a license declaration block in the Python test file too.

tests/test_plot3d.py
```diff
@@ -0,0 +1,39 @@
...
+                                        [1.0, 0.0, 0.0],
+                                        [1.0, 0.0, 1.0],
+                                        [1.0, 1.0, 0.0],
+                                        [1.0, 1.0, 1.0],
+                                        ])
+        # Check cells information
+        self.assertEqual(blk.ncell, 1)
+        self.assertEqual(blk.cltpn.ndarray[6:].tolist(), [5])
+        self.assertEqual(blk.clnds.ndarray[6:, :].tolist(),
+                         [[8, 0, 2, 6, 4, 1, 3, 7, 5]])
```
yungyuc, line 65: Could you please add vim modelines in the Python file too?

cpp/modmesh/inout/inout_util.cpp
```diff
@@ -0,0 +1,36 @@
...
+    while (std::getline(ss, token, delim))
+    {
+        output.push_back(token);
+    }
+    return output;
+}
+
+} // namespace inout
+
+} // namespace modmesh
```
yungyuc, line 64: Could you please add a vim modelines like that in the existing C++ files?

yungyuc (changes requested):
Many thanks, @j8xixo12 , it's a great patch.  Could you please fix some very minor formatting issues?

- [x] Add license comment blocks in the beginning of the new files.
- [x] Add modelines at the end of the new files

j8xixo12:
@yungyuc, I have add license and vim modelines in new files.
- [x] Add license comment blocks in the beginning of the new files.
- [x]  Add modelines at the end of the new files

yungyuc:
> @yungyuc, I have add license and vim modelines in new files.
> 
> * [x]  Add license comment blocks in the beginning of the new files.
> * [x]   Add modelines at the end of the new files

Thanks, @j8xixo12 .  You missed `cpp/modmesh/inout/pymod/wrap_Plot3d.cpp`.  Please add the comments next time changing the file.

yungyuc (approved):

yungyuc:
Ah-ha, we need to make some enhancements for the user experience:

```python
help_plot3d_viewer(path, set_command=True)
NameError: name 'path' is not defined
traceback:
  File "/Users/yungyuc/work/modmesh/modmesh/system.py", line 106, in enter_main
    ret = _run_viewer(argv)
  File "/Users/yungyuc/work/modmesh/modmesh/system.py", line 81, in _run_viewer
    return view.launch()
  File "/Users/yungyuc/work/modmesh/modmesh/view.py", line 116, in launch
    return app.exec()
  File "/Users/yungyuc/work/modmesh/modmesh/system.py", line 124, in exec_code
    traceback.print_stack()
```

Let's use some new issues.

j8xixo12:
Ah..., the path should be replace by the path to p3d file, ex: `/Users/...../modmesh/xxxx.p3d`
I think I should improve the description so that user can easily understand how to use it.

yungyuc:
It's OK for now.  I think we need an overall UI redesign.  Keeping a file loader in an "app" is absolutely a wrong design.  We are just doing a prototype and let us think more about what we want for everything in modmesh (and in the future solvcon).

And it is good that we did not put a large p3d fixture file.  We should not.

yungyuc:
Forgot to note: this PR resolves issue #331

PR #337, opened 2024-05-31
Use the NRBC type 2 with $\lambda=0$ #1

Gene0315:
Following the issue [#330](https://github.com/solvcon/modmesh/issues/330) 

This PR includes:
1. the NRBC type 2
2. the sequences of "march_half_so0", "treat_boundary_so0", and "treat_boundary_so1".

Obviously, the result shows that there must be something wrong in the code. 
截圖 2024-05-31 中午12 41 15

PR #338, opened 2024-06-01
Adjust the boundary plotting

Gene0315:
Following the issue https://github.com/solvcon/modmesh/discussions/335
The PR sets the value at boundary point equal to the inside one.

Features: The plotting looks more reasonable than the original one.
Strength: The adjustment is simple and makes the result look nicer.
Weakness: This way is too artificial and simplified and lack of physical basis. The result shows that there is still something wrong on the boundary, especially the left boundary.

截圖 2024-06-01 下午2 42 33

### DIFF
--- a/cpp/modmesh/onedim/Euler1DCore.cpp
+++ b/cpp/modmesh/onedim/Euler1DCore.cpp
@@ -170,6 +170,22 @@ void Euler1DCore::treat_boundary_so0()
         m_so0(ic, 1) = m_so0(ic - 1, 1);
         m_so0(ic, 2) = m_so0(ic - 1, 2);
     }
+
+    // Set the boundary point eqaul to the inside point.
+    {
+        // Left boundary.
+        size_t const ic = 0;
+        m_so0(ic, 0) = m_so0(ic + 2, 0);
+        m_so0(ic, 1) = m_so0(ic + 2, 1);
+        m_so0(ic, 2) = m_so0(ic + 2, 2);
+    }
+    {
+        // Right boundary.
+        size_t const ic = ncoord() - 1;
+        m_so0(ic, 0) = m_so0(ic - 2, 0);
+        m_so0(ic, 1) = m_so0(ic - 2, 1);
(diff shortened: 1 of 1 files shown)


yungyuc (changes requested):
Thanks, @Gene0315 .  This is a good starting point.  I agree this is a very each patch but it may be too easy to be useful.  The modification changes the outside points in a purely artificial way and one cannot predict the approach based on the understandings of the CESE method nor the boundary-condition treatment.  It makes it hard to maintain (understand).

This modification changes the solution array instead of changing the way we plot the array.  A more reasonable way is to just change the plotting code.

There may be follow-up work for enhancing the solution array size once the plotting is made more beautiful.

yungyuc:
@j8xixo12 comments?

j8xixo12:
> @j8xixo12 comments?

I agree with @yungyuc viewpoint, we should modify the plotting code not the solution data, because modifying solution data may effect the computation result, but plotting code not.
I think we can work together to enhance the plotting code !

Gene0315:
> > @j8xixo12 comments?
> 
> I agree with @yungyuc viewpoint, we should modify the plotting code not the solution data, because modifying solution data may effect the computation result, but plotting code not. I think we can work together to enhance the plotting code !

Ok, I know what I did improperly. I will try to revise the code in plotting section without changing the solution data.
The first thing I plan to do is to remove the boundary point on the figure.

yungyuc:
That sounds like a good next step, @Gene0315 .  When do you plan to be ready for the next review?

When you are ready, please annotate the important lines like what was done in https://github.com/solvcon/modmesh/pull/337#discussion_r1622530961 and other PRs.  And also leave a global comment explicitly saying that you are ready for the next review.  If anything is not clear to you, go to the discord channel to ask for help.

Gene0315:
@yungyuc
I think I can finish it this Friday or Saturday.

yungyuc:
> @yungyuc I think I can finish it this Friday or Saturday.

@Gene0315 We forgot to track this.  Any update?

PR #349, opened 2024-06-06
add missing function in BufferExpander

### LINKED ISSUE #306: Supplement missing helpers in `BufferExpander`
Some useful helper functions, e.g., [`nbytes()`](https://github.com/solvcon/modmesh/blob/3d1944e2342118b62e3349f425b3585a89002b6c/cpp/modmesh/buffer/ConcreteBuffer.hpp#L231), are available on `ConcreteBuffer` but not `BufferExpander`.  We should review and add them to `BufferExpander`.

This is to follow up the discussion at https://github.com/solvcon/modmesh/pull/304#discussion_r1555105830

tigercosmos:
fixes #306

### DIFF
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -91,7 +91,7 @@ public:
     BufferExpander & operator=(BufferExpander &&) = delete;
     ~BufferExpander() = default;
 
-    explicit operator bool() const noexcept { return bool(m_data_holder); }
+    explicit operator bool() const noexcept { return bool(m_begin); }
 
     size_type size() const noexcept
     {
@@ -134,12 +134,12 @@ public:
     using iterator = int8_t *;
     using const_iterator = int8_t const *;
 
-    iterator begin() noexcept { return data(); }
-    iterator end() noexcept { return data() + size(); }
-    const_iterator begin() const noexcept { return data(); }
-    const_iterator end() const noexcept { return data() + size(); }
-    const_iterator cbegin() const noexcept { return begin(); }
-    const_iterator cend() const noexcept { return end(); }
+    iterator begin() noexcept { return m_begin; }
+    iterator end() noexcept { return m_end; }
+    const_iterator begin() const noexcept { return m_begin; }
+    const_iterator end() const noexcept { return m_end; }
+    const_iterator cbegin() const noexcept { return m_begin; }
+    const_iterator cend() const noexcept { return m_end; }
 
     /* Backdoor */
     int8_t data(size_type it) const { return data()[it]; }
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -31,10 +31,10 @@
 #include <modmesh/base.hpp>
 #include <modmesh/buffer/ConcreteBuffer.hpp>
 
-#include <stdexcept>
-#include <memory>
 #include <algorithm>
+#include <memory>
 #include <sstream>
+#include <stdexcept>
 
 namespace modmesh
 {
@@ -91,10 +91,14 @@ public:
     BufferExpander & operator=(BufferExpander &&) = delete;
     ~BufferExpander() = default;
 
+    explicit operator bool() const noexcept { return bool(m_data_holder); }
+
     size_type size() const noexcept
     {
         return static_cast<size_type>(this->m_end - this->m_begin);
     }
+
+    size_t nbytes() const noexcept { return size(); }
 
     size_type capacity() const noexcept
     {
@@ -126,6 +130,16 @@ public:
         validate_range(it);
         return data(it);
     }
+
+    using iterator = int8_t *;
+    using const_iterator = int8_t const *;
+
+    iterator begin() noexcept { return data(); }
+    iterator end() noexcept { return data() + size(); }
+    const_iterator begin() const noexcept { return data(); }
+    const_iterator end() const noexcept { return data() + size(); }
+    const_iterator cbegin() const noexcept { return begin(); }
+    const_iterator cend() const noexcept { return end(); }
 
     /* Backdoor */
     int8_t data(size_type it) const { return data()[it]; }


cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -31,10 +31,10 @@
 #include <modmesh/base.hpp>
 #include <modmesh/buffer/ConcreteBuffer.hpp>
 
-#include <stdexcept>
-#include <memory>
 #include <algorithm>
+#include <memory>
```
tigercosmos, line 35: clang-format did it.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -91,11 +91,15 @@ class BufferExpander
     BufferExpander & operator=(BufferExpander &&) = delete;
     ~BufferExpander() = default;
 
+    explicit operator bool() const noexcept { return bool(m_data_holder); }
```
tigercosmos: please confirm if `m_data_holder` is the one to check.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -91,11 +91,15 @@ class BufferExpander
...
+
     size_type size() const noexcept
     {
         return static_cast<size_type>(this->m_end - this->m_begin);
     }
 
+    size_t nbytes() const noexcept { return size(); }
```
tigercosmos, line 101: maybe we should remove `nbytes` in both `ConcreteBuffer` and `BufferExpander`, it's redundant to `size()`

tigercosmos:
@yungyuc ready for review. the PR is small and straightforward so I don't think tests are required.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: Smart gadget.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -31,10 +31,10 @@
 #include <modmesh/base.hpp>
 #include <modmesh/buffer/ConcreteBuffer.hpp>
 
-#include <stdexcept>
-#include <memory>
 #include <algorithm>
+#include <memory>
 #include <sstream>
+#include <stdexcept>
```
yungyuc, line 37: I don't like clang-format to reorder the include file.  If you have time please find a switch to turn off the reordering (later).

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: No.  The object is valid when `m_concrete_buffer` is non-null.  Please check for `m_begin`.

This is a good point.  Thanks for highlighting.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: `nbytes()` is intuitive.  It's good redundancy.  It's hard to tell `size()` returns the number of bytes in the buffer managers.

The redundancy also works as document.  A maintainer immediately knows that `size()` is number of bytes by reading the alias.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -127,6 +131,16 @@ class BufferExpander
         return data(it);
     }
 
+    using iterator = int8_t *;
+    using const_iterator = int8_t const *;
+
+    iterator begin() noexcept { return data(); }
```
yungyuc: Use `m_begin`, `m_end` for iterator.

yungyuc (changes requested):
- [x] `bool()` to be true when `m_begin` is non-null.
- [x] Use `m_begin`, `m_end` for iterator.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: I followed what `ConcreteBuffer` does, shall I also fix `ConcreteBuffer`?

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: I think reordering is a good feature, why do you not like it?

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: I checked the code, actually I think checking `m_data_holder`, `m_concrete_buffer ` or `m_begin` are the same thing, but I can change to just checking `m_begin`, which looks more straightforward.

tigercosmos:
@yungyuc  please check the comment.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -127,6 +131,16 @@ class BufferExpander
         return data(it);
     }
 
+    using iterator = int8_t *;
+    using const_iterator = int8_t const *;
+
+    iterator begin() noexcept { return m_begin; }
```
yungyuc, line 137: This can use google-test.  But we can do it with this PR or in a future one.

`ConcreteBuffer` might be more complex than this, but I can't be sure at the moment.  You can also choose to do it with this PR or in a future PR.

yungyuc (approved):
I think it's OK to merge.  Let me know if you want to further change it or have it merged.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: @yungyuc I think it's ready to merge.

I think a test for testing the iterator is trivial, what do you think?

yungyuc:
> fixes #306

Next time please summarize what functions you add, remove, and modify, and avoid issue commands like this.  There is remaining work and I have to reopen the issue.  It is not considerate.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: > @yungyuc I think it's ready to merge.
> 
> I think a test for testing the iterator is trivial, what do you think?

Yes it can be trivial but should be added.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: Because the order has meaning.

tigercosmos: @yungyuc  to prevent reordering, you can add a new empty line, then clang-format will not do reordering.

yungyuc: That is ugly

PR #352, opened 2024-06-07
improve the comment in gtests

tigercosmos:
~~The current gtest files include `nopython` string in the filename, which is redundant as we always  put non-python test in gtest and python related tests in pytest. Therefore renaming all gtest files to make their names shorter.~~

add a comment to point out that the `test_python` target is only for testing non-Python code.


yungyuc:
No we may have C++ Google-test including Python header in the future.  Please do not rename the existing `nopython` gtest files.  Instead, you can add comment in the files and/or the `CMakeLists.txt` to indicate the future work.

yungyuc (changes requested):
No we may have C++ Google-test including Python header in the future.  Please do not rename the existing `nopython` gtest files.  Instead, you can add comment in the files and/or the `CMakeLists.txt` to indicate the future work.

tigercosmos:
@yungyuc I reverted the change and added a comment in `gtests/CMakefile.txt`.

yungyuc (approved):
Thanks.  LGTM.

PR #358, opened 2024-06-08
add gtest for ConcreteBuffer, SimpleArray and BufferExpander

### LINKED ISSUE #306: Supplement missing helpers in `BufferExpander`

tigercosmos:
This PR aims to close https://github.com/solvcon/modmesh/issues/306.

- After reading the source code of `ConcreteBuffer`, I think we don't need to align the iterator implementation between `ConcreteBuffer` and `BufferExpander`. In `ConcreteBuffer`, there are no `m_begin` and `m_end`, and only has a `std::unique_ptr<int8_t, data_deleter_type>` type `m_data`, so I think we can keep it as it is.
- Added iterator tests in gtest for `ConcreteBuffer`, `SimpleArray`, and `BufferExpander`.

p.s. Seems that `SimpleCollector.hpp` also misses the implementation of iterator, but let's make it in another PR.

### DIFF
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -62,6 +62,24 @@ TEST(SimpleArray, abs)
     EXPECT_EQ(brr.sum(), 10.0);
 }
 
+TEST(SimpleArray, iterator)
+{
+    using namespace modmesh;
+
+    SimpleArray<double> arr(10);
+    int8_t i = 0;
+    for (auto & it : arr)
+    {
+        it = i++;
+    }
+
+    i = 0;
+    for (const auto it : arr)
+    {
+        EXPECT_EQ(it, i++);
+    }
+}
+
 TEST(SimpleArray_DataType, from_type)
 {
     modmesh::DataType dt_double = modmesh::DataType::from<double>();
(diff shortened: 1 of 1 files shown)


gtests/test_nopython_buffer.cpp
```diff
@@ -6,9 +6,22 @@
...
 
-TEST(nopython_buffer, dummy)
+TEST(ConcreteBuffer, iterator)
 {
-    EXPECT_TRUE(true);
+    using namespace modmesh;
+
+    auto buffer = ConcreteBuffer::construct(10);
+    int8_t i = 0;
+    for (auto & it : *buffer)
```
tigercosmos, line 15: testing iterator

gtests/test_nopython_buffer.cpp
```diff
@@ -6,9 +6,22 @@
...
+    {
+        it = i++;
+    }
+
+    i = 0;
+    for (const auto it : *buffer)
```
tigercosmos, line 21: testing const iterator

yungyuc: Why the const iterator testing does not use reference while the non-const iterator testing does?

yungyuc (changes requested):
Thanks, @tigercosmos .  This is basic and useful testing.  There is a thing to clarify:

- [x] Clarify the consistency of testing between const and non-const iterator.

yungyuc:
> This PR aims to close #306.

I am not sure using the PR directives like "close", "fix" is a good idea.  But let's try.  If we need to do a lot of manual reopen, maybe it's not good.

> p.s. Seems that `SimpleCollector.hpp` also misses the implementation of iterator, but let's make it in another PR.

Makes sense.

gtests/test_nopython_buffer.cpp
tigercosmos: I think both works, but my original thought is that the value here is `int8_t`, so copying it is more efficient than getting the reference from it?

tigercosmos:
@yungyuc  do you think the issue is not ready to close after this PR? also, please check my comment.

gtests/test_nopython_buffer.cpp
yungyuc: I see, the explanation is clear.  This is good.  I just wanted to check that it is intentional.

yungyuc (approved):
Good for merging.

Issue #334, opened 2024-05-20
Viewer euler1d No Response in Windows


terrychan999:
When running the **euler1d** with Viewer on Windows and pressing the start button, there is no response.
However, when piping the viewer output to echo, it works properly.
Additionally, it functions properly when commenting out the `self.log` related code in `modmesh/app/euler1d.py`
https://github.com/solvcon/modmesh/blob/master/modmesh/app/euler1d.py#L631

![image](https://github.com/solvcon/modmesh/assets/14991138/8feb64ef-93ac-4cee-9d23-182f9b098ede)

#### Reference Build Process:
https://hackmd.io/XhxU2XnXQnqT4Of37kshVg

#### Build Commands:
```pwsh
cmake -Dpybind11_DIR="$(pybind11-config --cmakedir)" -S . -B build
cmake --build build --config Release --target viewer
```
#### Execute Commands:
Non-functioning Viewer:
```pwsh
.\build\cpp\binary\viewer\Release\viewer.exe
```
Functioning Viewer:
```pwsh
.\build\cpp\binary\viewer\Release\viewer.exe | echo
```
#### System Information:
- **OS:** Windows 11 23H2 22631.3593
- **modmesh git commit:** 880ea79
- **Compiler:** Visual Studio 17.9
- **Python:** 3.11.9
- **Qt:** 6.7.0
- **PySide6 / Shiboken6:** 6.7.0
- **CLI:** Windows Terminal + pwsh 7.4.2

yungyuc:
Thanks @terrychan999 for identifying the issue!

> When running the **euler1d** with Viewer on Windows and pressing the start button, there is no response. However, when piping the viewer output to echo, it works properly. Additionally, it functions properly when commenting out the `self.log` related code in `modmesh/app/euler1d.py` https://github.com/solvcon/modmesh/blob/master/modmesh/app/euler1d.py#L631

I wonder how does it behave by commenting out the following two lines https://github.com/solvcon/modmesh/blob/f9c053e28f4212b5ab2956016ee0b81c16503e92/modmesh/app/euler1d.py#L732

```cpp
        view.mgr.pycon.writeToHistory(msg)
        view.mgr.pycon.writeToHistory('\n')
```

`view.mgr.pycon` is a Qt object.

terrychan999:
> I wonder how does it behave by commenting out the following two lines
> ```c++
>         view.mgr.pycon.writeToHistory(msg)
[...]

PR #359, opened 2024-06-09
Fix euler1d not responding on Windows

tychuang1211:
This PR is a simple fix for #334.

According to #334, under some conditions `sys.stdout` can be `None`, which is usually the case for Windows GUI apps that aren't connected to a console.

Added a single line to check the value of `sys.stdout` before writing to fix the issue.

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -727,6 +727,7 @@ class Euler1DApp():
 
         :return: None
         """
+        # stdout can be None under some conditions, ref: https://github.com/solvcon/modmesh/issues/334
         if sys.stdout is not None:
             sys.stdout.write(msg)
             sys.stdout.write('\n')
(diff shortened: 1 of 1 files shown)


modmesh/app/euler1d.py
```diff
@@ -727,8 +727,9 @@ def log(msg):
 
         :return: None
         """
-        sys.stdout.write(msg)
-        sys.stdout.write('\n')
+        if sys.stdout != None:
```
yungyuc: What about adding a comment pointing to https://github.com/solvcon/modmesh/issues/334 ?  We may need some refactoring before being able to make a regression (unit) test.  Some explanation in code will help reduce the chance of regression before that.

yungyuc (changes requested):
Thanks @tychuang1211 .  A minor request:

- [x] Add comments.

modmesh/app/euler1d.py
tychuang1211: Added!

tychuang1211:
@yungyuc I added the comment.

yungyuc (approved):
Thanks, @tychuang1211 !

PR #361, opened 2024-06-10
Hide the boundary points

Gene0315:
Following the issue https://github.com/solvcon/modmesh/issues/338
The PR hides the points on boundary.

Features: The plotting hides the unreasonable points on the boundary.
Strength: It makes the figure look nicer without changing the solution data.
Weakness: It is better to calculate the correct values on the boundary and show them on the boundary.
截圖 2024-06-10 上午10 20 31


modmesh/app/euler1d.py
```diff
@@ -530,7 +530,8 @@ def build_grid_figure(self):
 
         :return: FigureCanvas
         """
-        x = self.st.svr.coord[::2]
+        pt_mid = np.linspace(2,len(self.st.svr.coord) - 3,num=((len(self.st.svr.coord) - 3) // 2),dtype=int)
```
j8xixo12, line 533: This code appears many times, would you like to optimize it ?

.DS_Store
j8xixo12, line 1: Please remove this binary file, we should not include unrelated file in a PR.

cpp/.DS_Store
j8xixo12, line 1: ditto.

modmesh/.DS_Store
j8xixo12, line 1: ditto.

j8xixo12 (changes requested):
Thanks @Gene0315 for the great work!

I left some comment in changed files and the following is the checking list before merging this PR:

- [ ] Fix all Github Action error
- [ ] Simplify the redundant code in this PR
- [ ] Clean up unrelated binary files

yungyuc:
And when cleaning up the blob files, please remove them and squash.  I do not want any blob files appearing in git history.

modmesh/app/euler1d.py
Gene0315: Yes, I think I can do it.

Gene0315:
> And when cleaning up the blob files, please remove them and squash. I do not want any blob files appearing in git history.

ok, I won't do it again.

.DS_Store
Gene0315: ok, I will remove them.

Issue #347, opened 2024-06-05
gtests/test_nopython_callprofiler.cpp timing is unstable


yungyuc:
See https://github.com/solvcon/modmesh/actions/runs/9378360062/job/25821452032#step:9:275 :

```
/Users/runner/work/modmesh/modmesh/gtests/test_nopython_callprofiler.cpp:175: Failure
Value of: diff_time(node1->data().total_time, uniqueTime1 + uniqueTime2)
  Actual: false
Expected: true
```

The test should use a more stable way to test for time.

tigercosmos:
@yungyuc  I checked all scheduled actions, no errors before, it's probably the PR issue.
https://github.com/solvcon/modmesh/actions?page=3&query=event%3Aschedule

yungyuc:
Right, it is a random failure.  Please take a look at my quote and link in the issue description.

tigercosmos:
@yungyuc  No, it's not a random failure, it never happens in the past few months

yungyuc:
I see what you meant.  It does not happen with scheduled runs, but only happens with PR runs?

tigercosmos:
@yungyuc  yes, that's what I meant. The current code should be very stable in the master branch, so the PR owner need to check why it failed.

yungyuc:
This is good finding.  Thanks, @tigercosmos .  Let me unassign this issue from you and move it out in progress.  We should revisit when we find another occasion.

tigercosmos:
We don't have this issue now.

PR #363, opened 2024-06-10
Implement `BufferBase` for the common part between `ConcreteBuffer` and `BufferExpander`

### LINKED ISSUE #305: Refactor to consolidate common parts between `ConcreteBuffer` and `BufferExpander`

tigercosmos:
part of https://github.com/solvcon/modmesh/issues/305

- Implemented the new `BufferBase` as the base class for all buffer-like type
- `ConcreteBuffer` and `BufferExpander` inherit the interface from `BufferBase` to ensure the interface's consistency.

### DIFF
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -102,4 +102,4 @@ protected:
 }; /* end class BufferBase */
 } /* end namespace modmesh */
 
-/* vim: set et ts=4 sw=4: */
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -41,17 +41,7 @@ public:
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
-    BufferBase() = default;
-    BufferBase(int8_t * begin, int8_t * end)
-        : m_begin(begin)
-        , m_end(end)
-    {
-    }
-
-    explicit operator bool() const
-    {
-        return bool(m_begin);
-    }
+    explicit operator bool() const { return bool(m_begin); }
     size_type size() const
     {
         return static_cast<size_type>(this->m_end - this->m_begin);
@@ -102,6 +92,8 @@ public:
     }
 
 protected:
+    virtual ~BufferBase() = default;
+
     void validate_range(size_t it) const
     {
         if (it >= size())
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -40,9 +40,7 @@ public:
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
-    virtual ~BufferBase() = default;
-
-    virtual explicit operator bool() const { return bool(m_begin); }
+    explicit operator bool() const { return bool(m_begin); }
     size_type size() const
     {
         return static_cast<size_type>(this->m_end - this->m_begin);
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -34,7 +34,7 @@ namespace modmesh
 
 /// Base class for buffer-like objects.
 template <typename Derived>
-class BufferBase : public std::enable_shared_from_this<BufferBase<Derived>>
+class BufferBase
 {
 public:
 
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -38,7 +38,11 @@ template <typename Derived>
 class BufferBase
 {
 public:
-    BufferBase() = default;
+    BufferBase(int8_t * begin, int8_t * end)
+        : m_begin(begin)
+        , m_end(end)
+    {
+    }
 
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
@@ -102,8 +106,8 @@ protected:
         }
     }
 
-    int8_t * m_begin = nullptr;
-    int8_t * m_end = nullptr;
+    int8_t * m_begin; // don't initialize, must be set by derived class
+    int8_t * m_end; // don't initialize, must be set by derived class
 
 }; /* end class BufferBase */
 
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -93,6 +93,11 @@ public:
 
 protected:
     BufferBase() = default;
+
+    BufferBase(int8_t * start, int8_t * end)
+        : m_begin(start)
+        , m_end(end){};
+
     BufferBase(BufferBase const &) = delete;
     BufferBase(BufferBase &&) = delete;
     BufferBase & operator=(BufferBase const &) = delete;
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -38,12 +38,6 @@ template <typename Derived>
 class BufferBase
 {
 public:
-    BufferBase(int8_t * begin, int8_t * end)
-        : m_begin(begin)
-        , m_end(end)
-    {
-    }
-
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -88,6 +88,8 @@ protected:
         }
     }
 
-    virtual constexpr const char* name() const { return "BufferBase"; }
-};
-} // namespace modmesh
+    virtual constexpr const char * name() const { return "BufferBase"; }
+}; /* end class BufferBase */
+} /* end namespace modmesh */
+
+/* vim: set et ts=4 sw=4: */
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -92,6 +92,13 @@ public:
     }
 
 protected:
+    BufferBase() = default;
+    BufferBase(BufferBase const &) = delete;
+    BufferBase(BufferBase &&) = delete;
+    BufferBase & operator=(BufferBase const &) = delete;
+    BufferBase & operator=(BufferBase &&) = delete;
+    ~BufferBase() = default;
+
     void validate_range(size_t it) const
     {
         if (it >= size())
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -92,7 +92,7 @@ public:
     }
 
 protected:
-    // virtual ~BufferBase() = default;
+    virtual ~BufferBase() = default;
 
     void validate_range(size_t it) const
     {
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -88,6 +88,6 @@ protected:
         }
     }
 
-    virtual constexpr std::string const & name() const { return "BufferBase"; }
+    virtual constexpr const char* name() const { return "BufferBase"; }
 };
 } // namespace modmesh
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -92,8 +92,6 @@ public:
     }
 
 protected:
-    virtual ~BufferBase() = default;
-
     void validate_range(size_t it) const
     {
         if (it >= size())
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -98,7 +98,7 @@ public:
 
     constexpr const char * name() const
     {
-        return static_cast<const Derived *>(this)->name();
+        return Derived::name();
     }
 
 protected:
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -33,7 +33,8 @@ namespace modmesh
 {
 
 /// Base class for buffer-like objects.
-class BufferBase : public std::enable_shared_from_this<BufferBase>
+template <typename Derived>
+class BufferBase : public std::enable_shared_from_this<BufferBase<Derived>>
 {
 public:
 
@@ -85,7 +86,14 @@ public:
     template <typename T> T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
     // clang-format on
 
+    constexpr const char * name() const
+    {
+        return static_cast<const Derived *>(this)->name();
+    }
+
 protected:
+    // virtual ~BufferBase() = default;
+
     void validate_range(size_t it) const
     {
         if (it >= size())
@@ -93,9 +101,6 @@ protected:
             throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
         }
     }
-
-    // TODO: make this constexpr virtual once C++20 is available
-    virtual const char * name() const { return "BufferBase"; }
 
     int8_t * m_begin = nullptr;
     int8_t * m_end = nullptr;
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -41,7 +41,17 @@ public:
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
-    explicit operator bool() const { return bool(m_begin); }
+    BufferBase() = default;
+    BufferBase(int8_t * begin, int8_t * end)
+        : m_begin(begin)
+        , m_end(end)
+    {
+    }
+
+    explicit operator bool() const
+    {
+        return bool(m_begin);
+    }
     size_type size() const
     {
         return static_cast<size_type>(this->m_end - this->m_begin);
@@ -92,8 +102,6 @@ public:
     }
 
 protected:
-    virtual ~BufferBase() = default;
-
     void validate_range(size_t it) const
     {
         if (it >= size())
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -96,7 +96,8 @@ protected:
         }
     }
 
-    virtual constexpr const char * name() const { return "BufferBase"; }
+    // TODO: make this constexpr virtual once C++20 is available
+    virtual const char * name() const { return "BufferBase"; }
 
     int8_t * m_begin = nullptr;
     int8_t * m_end = nullptr;
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -1,4 +1,5 @@
 #pragma once
+
 /*
  * Copyright (c) 2024, An-Chi Liu <phy.tiger@gmail.com>
  *
@@ -37,6 +38,7 @@ template <typename Derived>
 class BufferBase
 {
 public:
+    BufferBase() = default;
 
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
@@ -102,7 +104,9 @@ protected:
 
     int8_t * m_begin = nullptr;
     int8_t * m_end = nullptr;
+
 }; /* end class BufferBase */
+
 } /* end namespace modmesh */
 
 // vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -46,8 +46,6 @@ public:
 
     virtual size_type size() const noexcept = 0;
     virtual size_t nbytes() const noexcept = 0;
-    virtual size_type capacity() const noexcept = 0;
-    virtual void reserve(size_type cap) = 0;
 
     int8_t operator[](size_t it) const { return data(it); }
     int8_t & operator[](size_t it) { return data(it); }
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -42,10 +42,12 @@ public:
 
     virtual ~BufferBase() = default;
 
-    virtual explicit operator bool() const noexcept = 0;
-
-    virtual size_type size() const noexcept = 0;
-    virtual size_t nbytes() const noexcept = 0;
+    virtual explicit operator bool() const { return bool(m_begin); }
+    size_type size() const
+    {
+        return static_cast<size_type>(this->m_end - this->m_begin);
+    }
+    size_t nbytes() const { return size() * sizeof(int8_t); }
 
     int8_t operator[](size_t it) const { return data(it); }
     int8_t & operator[](size_t it) { return data(it); }
@@ -65,17 +67,25 @@ public:
     using iterator = int8_t *;
     using const_iterator = int8_t const *;
 
-    virtual iterator begin() noexcept = 0;
-    virtual iterator end() noexcept = 0;
-    virtual const_iterator begin() const noexcept = 0;
-    virtual const_iterator end() const noexcept = 0;
-    virtual const_iterator cbegin() const noexcept = 0;
-    virtual const_iterator cend() const noexcept = 0;
+    iterator begin() noexcept { return m_begin; }
+    iterator end() noexcept { return m_end; }
+    const_iterator begin() const noexcept { return m_begin; }
+    const_iterator end() const noexcept { return m_end; }
+    const_iterator cbegin() const noexcept { return m_begin; }
+    const_iterator cend() const noexcept { return m_end; }
 
-    virtual int8_t data(size_type it) const = 0;
-    virtual int8_t & data(size_type it) = 0;
-    virtual int8_t const * data() const noexcept = 0;
-    virtual int8_t * data() noexcept = 0;
+    /* Backdoor */
+    int8_t data(size_type it) const { return data()[it]; }
+    int8_t & data(size_type it) { return data()[it]; }
+    int8_t const * data() const noexcept { return data<int8_t>(); }
+    int8_t * data() noexcept { return data<int8_t>(); }
+
+    // clang-format off
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    template <typename T> T const * data() const noexcept { return reinterpret_cast<T *>(m_begin); }
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    template <typename T> T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
+    // clang-format on
 
 protected:
     void validate_range(size_t it) const
@@ -87,6 +97,9 @@ protected:
     }
 
     virtual constexpr const char * name() const { return "BufferBase"; }
+
+    int8_t * m_begin = nullptr;
+    int8_t * m_end = nullptr;
 }; /* end class BufferBase */
 } /* end namespace modmesh */
 
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -33,7 +33,7 @@ namespace modmesh
 {
 
 /// Base class for buffer-like objects.
-class BufferBase
+class BufferBase : public std::enable_shared_from_this<BufferBase>
 {
 public:
 
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -88,8 +88,6 @@ protected:
         }
     }
 
-    virtual std::string const & name() const { return m_name; }
-
-    std::string m_name = "BufferBase"; ///< Name of the buffer object.
+    virtual constexpr std::string const & name() const { return "BufferBase"; }
 };
 } // namespace modmesh
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -45,8 +45,8 @@ namespace modmesh
  * expandable memory buffer cannot be used externally.
  */
 class BufferExpander
-    : public BufferBase<BufferExpander>
-    , public std::enable_shared_from_this<BufferExpander>
+    : public std::enable_shared_from_this<BufferExpander>
+    , public BufferBase<BufferExpander>
 {
 
 private:
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -72,7 +72,8 @@ public:
     }
 
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
-        : m_concrete_buffer(clone ? buf->clone() : buf)
+        : BufferBase<BufferExpander>(nullptr, nullptr) //  initialize m_begin and m_end, but since we don't have the data yet, we set them to nullptr
+        , m_concrete_buffer(clone ? buf->clone() : buf)
     {
         m_begin = m_concrete_buffer->data();
         m_end = m_begin + m_concrete_buffer->size();
@@ -80,11 +81,15 @@ public:
     }
 
     BufferExpander(size_type nbyte, ctor_passkey const &)
+        : BufferBase<BufferExpander>(nullptr, nullptr) //  initialize m_begin and m_end, but since we don't have the data yet, we set them to nullptr
     {
         expand(nbyte);
     }
 
-    BufferExpander(ctor_passkey const &) {}
+    BufferExpander(ctor_passkey const &)
+        : BufferBase<BufferExpander>(nullptr, nullptr) //  initialize m_begin and m_end, but since we don't have the data yet, we set them to nullptr
+    {
+    }
 
     BufferExpander() = delete;
     BufferExpander(BufferExpander const &) = delete;
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -72,9 +72,10 @@ public:
 
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
         : m_concrete_buffer(clone ? buf->clone() : buf)
-        , BufferBase<BufferExpander>(m_concrete_buffer->data(), m_begin + m_concrete_buffer->size())
-        , m_end_cap(m_begin + m_concrete_buffer->size())
     {
+        m_begin = m_concrete_buffer->data();
+        m_end = m_begin + m_concrete_buffer->size();
+        m_end_cap = m_begin + m_concrete_buffer->size();
     }
 
     BufferExpander(size_type nbyte, ctor_passkey const &)
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -102,12 +102,12 @@ public:
 
     size_t nbytes() const noexcept override { return size(); }
 
-    size_type capacity() const noexcept override
+    size_type capacity() const noexcept
     {
         return static_cast<size_type>(this->m_end_cap - this->m_begin);
     }
 
-    void reserve(size_type cap) override;
+    void reserve(size_type cap);
 
     void expand(size_type length)
     {
@@ -150,7 +150,7 @@ private:
         return ret;
     }
 
-    constexpr const char* name() const override { return "BufferExpander"; }
+    constexpr const char * name() const override { return "BufferExpander"; }
 
     std::unique_ptr<int8_t> m_data_holder = nullptr;
     std::shared_ptr<ConcreteBuffer> m_concrete_buffer = nullptr;
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -119,7 +119,7 @@ private:
         return ret;
     }
 
-    constexpr const char * name() const override { return "BufferExpander"; }
+    const char * name() const override { return "BufferExpander"; }
 
     std::unique_ptr<int8_t> m_data_holder = nullptr;
     std::shared_ptr<ConcreteBuffer> m_concrete_buffer = nullptr;
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -81,13 +81,13 @@ public:
     }
 
     BufferExpander(size_type nbyte, ctor_passkey const &)
-        : BufferBase<BufferExpander>() // don't delegate m_begin and m_end, which will be overwritten later
+        : BufferBase<BufferExpander>(nullptr, nullptr) // initialize m_begin and m_end to nullptr, which will be overwritten later
     {
         expand(nbyte); // override m_begin and m_end once we have the data
     }
 
     BufferExpander(ctor_passkey const &)
-        : BufferBase<BufferExpander>(nullptr, nullptr)
+        : BufferBase<BufferExpander>(nullptr, nullptr) // initialize m_begin and m_end to nullptr, which will be overwritten later
     {
     }
 
@@ -100,6 +100,10 @@ public:
 
     size_type capacity() const noexcept
     {
+        if (!m_begin) // no data hence no capacity
+        {
+            return 0;
+        }
         return static_cast<size_type>(this->m_end_cap - this->m_begin);
     }
 
--- a/cpp/modmesh/buffer/BufferExpander.hpp
+++ b/cpp/modmesh/buffer/BufferExpander.hpp
@@ -107,7 +107,7 @@ public:
     std::shared_ptr<ConcreteBuffer> const & as_concrete(size_type cap = 0);
     bool is_concrete() const { return bool(m_concrete_buffer); }
 
-    constexpr const char * name() const { return "BufferExpander"; }
+    static constexpr const char * name() { return "BufferExpander"; }
 
 private:
     static std::unique_ptr<int8_t> allocate(size_type nbytes)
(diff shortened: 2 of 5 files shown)


cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -91,21 +93,21 @@ class BufferExpander
     BufferExpander & operator=(BufferExpander &&) = delete;
     ~BufferExpander() = default;
 
-    explicit operator bool() const noexcept { return bool(m_begin); }
+    explicit operator bool() const noexcept override { return bool(m_begin); }
```
tigercosmos: please check all `override`

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+    using difference_type = std::ptrdiff_t;
+
+    virtual ~BufferBase() = default;
+
+    virtual explicit operator bool() const noexcept = 0;
+
+    virtual size_type size() const noexcept = 0;
+    virtual size_t nbytes() const noexcept = 0;
+    virtual size_type capacity() const noexcept = 0;
+    virtual void reserve(size_type cap) = 0;
```
tigercosmos: do we want `reserve` in the base class?

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    int8_t operator[](size_t it) const { return data(it); }
```
tigercosmos: put the implementation of `[]` and `at()` in base class, as they are the same in different buffer classes.

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+    int8_t & at(size_t it)
+    {
+        validate_range(it);
+        return data(it);
+    }
+
+    using iterator = int8_t *;
+    using const_iterator = int8_t const *;
+
+    virtual iterator begin() noexcept = 0;
```
tigercosmos: let `begin`, `end()`, etc, as the virtual function since the implementations are different in the buffer classes.

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+    virtual const_iterator cbegin() const noexcept = 0;
+    virtual const_iterator cend() const noexcept = 0;
+
+    virtual int8_t data(size_type it) const = 0;
+    virtual int8_t & data(size_type it) = 0;
+    virtual int8_t const * data() const noexcept = 0;
+    virtual int8_t * data() noexcept = 0;
+
+protected:
+    void validate_range(size_t it) const
```
tigercosmos: same implementation, put in the base class.

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+        if (it >= size())
+        {
+            throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
+        }
+    }
+
+    virtual std::string const & name() const { return m_name; }
```
tigercosmos: basically only for `validate_range` to print the correct name.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -29,6 +29,7 @@
  */
 
 #include <modmesh/base.hpp>
+#include <modmesh/buffer/bufferbase.hpp>
```
yungyuc: Why the file is not using a camel case name `BufferBase.hpp` like others and the class `BufferBase`?

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+/// Base class for buffer-like objects.
+class BufferBase
+{
+public:
+
+    using size_type = std::size_t;
+    using difference_type = std::ptrdiff_t;
+
+    virtual ~BufferBase() = default;
```
yungyuc: Do not make `ConcreteBuffer` and `BufferExpander` polymorphic.  Most classes in modmesh should not use polymorphism.  Polymorphism slows down the system and the cost only makes sense when complex conditional branches are required, like in GUI.

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/base.hpp>
+
+namespace modmesh
+{
+
+/// Base class for buffer-like objects.
+class BufferBase
```
yungyuc: Make it inherit from `std::enable_shared_from_this` class template so that the derived classes do not need to explicitly inherit from `std::enable_shared_from_this`.

cpp/modmesh/buffer/bufferbase.hpp
yungyuc: No.  `reserve` does not make sense for `ConcreteBuffer` whose capacity and size should always be the same.

cpp/modmesh/buffer/bufferbase.hpp
yungyuc: I like it.

cpp/modmesh/buffer/bufferbase.hpp
yungyuc: Aligning the implementation of `begin()` and `end()` across buffers is good.

yungyuc: `begin()` and `end()` is a good example why polymorphism should not be used for a buffer (and data) object.  Every time you call the iterator getters you need a pointer redirect.  It is totally a waste of cycles.

cpp/modmesh/buffer/bufferbase.hpp
yungyuc: Could you please try if this can be made `constexpr`?

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+        if (it >= size())
+        {
+            throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
+        }
+    }
+
+    virtual std::string const & name() const { return m_name; }
+
+    std::string m_name = "BufferBase"; ///< Name of the buffer object.
```
yungyuc: I don't like this useless member in the base class.

More importantly, if we have to use a member datum, it has to be `static const` because it is for the class, not instances.

cpp/modmesh/buffer/bufferbase.hpp
```diff
@@ -0,0 +1,95 @@
...
+};
+} // namespace modmesh
```
yungyuc: Add the vim modeline like other files.  Even if you do not use vim, the modeline is to specify the formatting for humans.

yungyuc (changes requested):
It's in a good direction, but needs a lot of changes.

- [x] Clarify the file name of `BufferBase.hpp`
- [x] Avoid polymorphism
- [x] Make `BufferBase` inherit from `std::enable_shared_from_this`
- [x] `reserve()` is not common to all buffers.  Remove it from `BufferBase`.
- [x] `constexpr name()`?
- [x] Do not use instance member datum for the class name for the error message.  Use `static const` and/or `constexpr`, if possible.
- [x] Add correct vim modeline

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in c2a9f4e96b2a85c16fa5ba9dfe4b34434bef97c8

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in 88a5bc3797cc5860c851552cdbc9427001925e03

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in 5c7b51144ce4e152035c41bc48c64a64c1370df9

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: fixed in c035e21710366aa42bafb80ffba166e32a5ec064

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in 88a5bc3797cc5860c851552cdbc9427001925e03

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in 28737d85a2958209307467410fb38fbfc5f37ef8

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in 5c7b51144ce4e152035c41bc48c64a64c1370df9

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: fixed in bc36d4d0a1c7ddcbdaf62673adbeb9fb671a9a55

cpp/modmesh/buffer/bufferbase.hpp
tigercosmos: Ubuntu is fine, but macOS fails, the reason is that virtual constexpre requires c++20, see https://stackoverflow.com/questions/34828161/can-virtual-functions-be-constexpr

tigercosmos: I remove `constexpr` for now: 01721af7f8deb7869300c4a807121e884dfcc294

tigercosmos:
@yungyuc  all issues addressed, please take a look

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+/// Base class for buffer-like objects.
+class BufferBase : public std::enable_shared_from_this<BufferBase>
+{
+public:
+
+    using size_type = std::size_t;
+    using difference_type = std::ptrdiff_t;
+
+    virtual ~BufferBase() = default;
```
yungyuc: `BufferBase` is still polymorphic.  Please carefully remove polymorphism from `BufferBase` and all buffer containers.

cpp/modmesh/buffer/bufferbase.hpp
yungyuc: You have not fixed it yet.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,107 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/base.hpp>
+
+namespace modmesh
+{
+
+/// Base class for buffer-like objects.
+class BufferBase : public std::enable_shared_from_this<BufferBase>
```
yungyuc: The template argument `T` to `std::enable_shared_from_this<T>` has to be the class inheriting `BufferBase`.  As such, `BufferBase` needs to be a template employing CRTP.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: remove all the remaining `virtual` except for `name()`, which is only used for exceptions.

tigercosmos:
@yungyuc  fixed the issues. please have a look.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,105 @@
...
+    const_iterator cbegin() const noexcept { return m_begin; }
+    const_iterator cend() const noexcept { return m_end; }
+
+    /* Backdoor */
+    int8_t data(size_type it) const { return data()[it]; }
+    int8_t & data(size_type it) { return data()[it]; }
+    int8_t const * data() const noexcept { return data<int8_t>(); }
+    int8_t * data() noexcept { return data<int8_t>(); }
+
+    // clang-format off
```
yungyuc, line 82: Could you please show how it looks with clang-format on in a comment?

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,105 @@
...
+    void validate_range(size_t it) const
+    {
+        if (it >= size())
+        {
+            throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
+        }
+    }
+
+    // TODO: make this constexpr virtual once C++20 is available
+    virtual const char * name() const { return "BufferBase"; }
```
yungyuc: You cannot have any `virtual` function because we must not use polymorphism with buffer containers.  And please try `constexpr` after removing `virtual`.

cpp/modmesh/buffer/BufferBase.hpp
yungyuc: No virtual can appear in a non-polymorphic class.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+    int8_t * m_begin = nullptr;
+    int8_t * m_end = nullptr;
+}; /* end class BufferBase */
+} /* end namespace modmesh */
+
+/* vim: set et ts=4 sw=4: */
```
yungyuc: Please reference the modeline in other files.  `ff=unix fenc=utf8` and `sts=4` are needed.

yungyuc (changes requested):
Thanks, @tigercosmos .  Remaining issues:

- [ ] Avoid polymorphism.  `BufferBase` is still polymorphic in the third review request.
- [x] `BufferBase` has to be CRTP for `std::enable_shared_from_this`.
- [x] Correct modeline.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: I copied from existing files.

without "clang-format off"
```
    template <typename T>
    T const * data() const noexcept { return reinterpret_cast<T *>(m_begin); }
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
    template <typename T>
    T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
```

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
...
+    template <typename T> T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
+    // clang-format on
+
+    constexpr const char * name() const
+    {
+        return static_cast<const Derived *>(this)->name();
+    }
+
+protected:
+    // virtual ~BufferBase() = default;
```
tigercosmos: @yungyuc  if uncommenting this line, the code can build.

if without the polymorphism, the build encounter the error related to `dynamic_cast`, which seems weird to me, could you take a look and provide some hints?

```log
[ 88%] Building CXX object cpp/modmesh/CMakeFiles/modmesh_primary.dir/testhelper/pymod/wrap_TestSimpleArrayHelper.cpp.o
In file included from /usr/include/c++/10/memory:84,
                 from /usr/local/include/pybind11/detail/common.h:315,
                 from /usr/local/include/pybind11/attr.h:13,
                 from /usr/local/include/pybind11/detail/class.h:12,
                 from /usr/local/include/pybind11/pybind11.h:13,
                 from /home/mujin/modmesh/cpp/modmesh/buffer/pymod/buffer_pymod.hpp:31,
                 from /home/mujin/modmesh/cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp:29:
/usr/include/c++/10/bits/shared_ptr.h: In instantiation of ‘std::shared_ptr<_Tp> std::dynamic_pointer_cast(const std::shared_ptr<_Tp>&) [with _Tp = modmesh::ConcreteBuffer; _Up = modmesh::BufferBase<modmesh::ConcreteBuffer>]’:
/usr/local/include/pybind11/pybind11.h:1858:80:   required from ‘static void pybind11::class_<type_, options>::init_holder(pybind11::detail::instance*, pybind11::detail::value_and_holder&, const holder_type*, const std::enable_shared_from_this<T>*) [with T = modmesh::BufferBase<modmesh::ConcreteBuffer>; type_ = modmesh::ConcreteBuffer; options = {std::shared_ptr<modmesh::ConcreteBuffer>}; pybind11::class_<type_, options>::holder_type = std::shared_ptr<modmesh::ConcreteBuffer>]’
/usr/local/include/pybind11/pybind11.h:1910:20:   required from ‘static void pybind11::class_<type_, options>::init_instance(pybind11::detail::instance*, const void*) [with type_ = modmesh::ConcreteBuffer; options = {std::shared_ptr<modmesh::ConcreteBuffer>}]’
/usr/local/include/pybind11/pybind11.h:1584:30:   required from ‘pybind11::class_<type_, options>::class_(pybind11::handle, const char*, const Extra& ...) [with Extra = {const char*, pybind11::buffer_protocol}; type_ = modmesh::ConcreteBuffer; options = {std::shared_ptr<modmesh::ConcreteBuffer>}]’
/home/mujin/modmesh/cpp/modmesh/python/common.hpp:394:45:   required from ‘modmesh::python::WrapBase<Wrapper, Wrapped, Holder, WrappedBase>::WrapBase(pybind11::module&, const char*, const char*, const Extra& ...) [with Extra = {pybind11::buffer_protocol}; Wrapper = modmesh::python::WrapConcreteBuffer; Wrapped = modmesh::ConcreteBuffer; Holder = std::shared_ptr<modmesh::ConcreteBuffer>; WrappedBase = modmesh::ConcreteBuffer; pybind11::module = pybind11::module_]’
/home/mujin/modmesh/cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp:49:69:   required from here
/usr/include/c++/10/bits/shared_ptr.h:602:23: error: cannot ‘dynamic_cast’ ‘(& __r)->std::shared_ptr<modmesh::BufferBase<modmesh::ConcreteBuffer> >::<anonymous>.std::__shared_ptr<modmesh::BufferBase<modmesh::ConcreteBuffer>, __gnu_cxx::_S_atomic>::get()’ (of type ‘using element_type = std::remove_extent<modmesh::BufferBase<modmesh::ConcreteBuffer> >::type*’ {aka ‘class modmesh::BufferBase<modmesh::ConcreteBuffer>*’}) to type ‘using element_type = using element_type = std::remove_extent<modmesh::ConcreteBuffer>::type*’ {aka ‘class modmesh::ConcreteBuffer*’} (source type is not polymorphic)
  602 |       if (auto* __p = dynamic_cast<typename _Sp::element_type*>(__r.get()))
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/10/bits/shared_ptr.h: In instantiation of ‘std::shared_ptr<_Tp> std::dynamic_pointer_cast(const std::shared_ptr<_Tp>&) [with _Tp = modmesh::BufferExpander; _Up = modmesh::BufferBase<modmesh::BufferExpander>]’:
/usr/local/include/pybind11/pybind11.h:1858:80:   required from ‘static void pybind11::class_<type_, options>::init_holder(pybind11::detail::instance*, pybind11::detail::value_and_holder&, const holder_type*, const std::enable_shared_from_this<T>*) [with T = modmesh::BufferBase<modmesh::BufferExpander>; type_ = modmesh::BufferExpander; options = {std::shared_ptr<modmesh::BufferExpander>}; pybind11::class_<type_, options>::holder_type = std::shared_ptr<modmesh::BufferExpander>]’
/usr/local/include/pybind11/pybind11.h:1910:20:   required from ‘static void pybind11::class_<type_, options>::init_instance(pybind11::detail::instance*, const void*) [with type_ = modmesh::BufferExpander; options = {std::shared_ptr<modmesh::BufferExpander>}]’
/usr/local/include/pybind11/pybind11.h:1584:30:   required from ‘pybind11::class_<type_, options>::class_(pybind11::handle, const char*, const Extra& ...) [with Extra = {const char*}; type_ = modmesh::BufferExpander; options = {std::shared_ptr<modmesh::BufferExpander>}]’
/home/mujin/modmesh/cpp/modmesh/python/common.hpp:394:45:   required from ‘modmesh::python::WrapBase<Wrapper, Wrapped, Holder, WrappedBase>::WrapBase(pybind11::module&, const char*, const char*, const Extra& ...) [with Extra = {}; Wrapper = modmesh::python::WrapBufferExpander; Wrapped = modmesh::BufferExpander; Holder = std::shared_ptr<modmesh::BufferExpander>; WrappedBase = modmesh::BufferExpander; pybind11::module = pybind11::module_]’
/home/mujin/modmesh/cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp:124:40:   required from here
/usr/include/c++/10/bits/shared_ptr.h:602:23: error: cannot ‘dynamic_cast’ ‘(& __r)->std::shared_ptr<modmesh::BufferBase<modmesh::BufferExpander> >::<anonymous>.std::__shared_ptr<modmesh::BufferBase<modmesh::BufferExpander>, __gnu_cxx::_S_atomic>::get()’ (of type ‘using element_type = std::remove_extent<modmesh::BufferBase<modmesh::BufferExpander> >::type*’ {aka ‘class modmesh::BufferBase<modmesh::BufferExpander>*’}) to type ‘using element_type = using element_type = std::remove_extent<modmesh::BufferExpander>::type*’ {aka ‘class modmesh::BufferExpander*’} (source type is not polymorphic)
gmake[4]: *** [cpp/modmesh/CMakeFiles/modmesh_primary.dir/build.make:118: cpp/modmesh/CMakeFiles/modmesh_primary.dir/buffer/pymod/wrap_ConcreteBuffer.cpp.o] Error 1
gmake[4]: *** Waiting for unfinished jobs....
```

tigercosmos: the build difference can be checked in https://github.com/solvcon/modmesh/pull/363/commits/45ce936eb8ea2963420deee3937ec292180773f7 and https://github.com/solvcon/modmesh/pull/363/commits/e5b8d34a321560f9052c79c6fd5ba772bfeb4467

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -71,10 +71,10 @@ class BufferExpander
 
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
```
yungyuc, line 74: You need to call the constructor of the parent class!

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -71,10 +71,10 @@ class BufferExpander
...
         : m_concrete_buffer(clone ? buf->clone() : buf)
-        , m_begin(m_concrete_buffer->data())
-        , m_end(m_begin + m_concrete_buffer->size())
-        , m_end_cap(m_begin + m_concrete_buffer->size())
     {
+        m_begin = m_concrete_buffer->data();
```
yungyuc: Keep them in the initialization list.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
...
+    template <typename T> T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
+    // clang-format on
+
+    constexpr const char * name() const
+    {
+        return static_cast<const Derived *>(this)->name();
+    }
+
+protected:
+    virtual ~BufferBase() = default;
```
yungyuc: `virtual` needs to go away.  Please figure it out.

yungyuc (changes requested):
- [x] Parent constructor delegate.
- [ ] Do not use polymorphism.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: I don't think it's possible, due to `when initialized here [-Werror=reorder]`

`m_begin` and `m_end` need to rely on `m_data`, however the `m_data`'s type are different in the derived class, which result that we cannot put `m_data` into the base class.

so the order need to be:
- derived class `m_data` initialized
- `m_start` and `m_end` initialized.

if we use parent constructor, then the compiler will complain that the base class member needs to be initialized first. so, unless we turn off the "reorder warning", we cannot use parent constructor.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: removed all virtual.

tigercosmos:
@yungyuc  ready for review, note that:
- we need to put `enable_shared_from_this` in the derived class due to our wrapper implementation.
- we cannot use the parent constructor due to compiler's "Werror=reorder"
- removed all `virtual`

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: figured out, we cannot put `enable_shared_from_this` in the base class.

tigercosmos: If we want to put enable_shared_from_this  in the base class, then polymorphism is needed.

cpp/modmesh/buffer/BufferBase.hpp
yungyuc: `virtual` is still here.  Please check before asking for review.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    void validate_range(size_t it) const
+    {
+        if (it >= size())
+        {
+            throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
+        }
+    }
+
+    int8_t * m_begin = nullptr;
```
yungyuc: Remove initial value to save runtime, and add comments to explain why initial value should not be provided (because it must be overridden in derived class).

Although it is possible for compiler to optimize out the initialization here, I'd like the code to self-explain: `BufferBase::` `m_begin` and `m_end` must be overridden in the derived classes.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -43,8 +44,8 @@ namespace modmesh
  * Untyped and growing memory buffer for contiguous data storage.  The internal
  * expandable memory buffer cannot be used externally.
  */
-class BufferExpander
-    : public std::enable_shared_from_this<BufferExpander>
+class BufferExpander : public BufferBase<BufferExpander>
```
yungyuc: `std::enable_shared_from_this` should be the first parent.

Use a standalone line for `public BufferBase<BufferExpander>` to make the block like:

```cpp
class BufferExpander
    : public BufferBase<BufferExpander>
    , public std::enable_shared_from_this<BufferExpander>
```

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -117,8 +118,8 @@ struct ConcreteBufferDataDeleter
 /**
  * Untyped and unresizeable memory buffer for contiguous data storage.
  */
-class ConcreteBuffer
-    : public std::enable_shared_from_this<ConcreteBuffer>
+class ConcreteBuffer : public BufferBase<ConcreteBuffer>
```
yungyuc: `std::enable_shared_from_this` should be the first parent.

Format:

```cpp
class ConcreteBuffer
     : public BufferBase<ConcreteBuffer>
     , public std::enable_shared_from_this<ConcreteBuffer>
```

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
+/*
```
yungyuc, line 3: Use a blank line.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
...
+        if (it >= size())
+        {
+            throw std::out_of_range(Formatter() << name() << ": index " << it << " is out of bounds with size " << size());
+        }
+    }
+
+    int8_t * m_begin = nullptr;
+    int8_t * m_end = nullptr;
+}; /* end class BufferBase */
+} /* end namespace modmesh */
```
yungyuc, line 120: Use a blank line between class and namespace.  Please follow the style in the existing files.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: We must obey the initialization order.  It's not an option to initialize member data without following the order they are declared.

The type of derived class `m_data` probably does not matter.  What matters is their address.  Since `m_data` is after `m_begin` and `m_end`, the address of `m_data` cannot be known when `m_begin` and `m_end` appears in the initialization list.

Thus, `m_begin` and `m_end` need to be overridden after the derived classes are constructed.  We cannot initialize `m_begin` and `m_end` from the derived class initialization list through constructor delegate.

yungyuc (changes requested):
Do you still think making the common `BufferBase` class template is a good idea?  There are a lot to consider to align the common parts of `ConcreteBuffer` and `BufferExpander` in a base class.  I do not see it to have particularly high value, although it does no harm.

If there are more valuable work to do, we do not need to finish this now.

---

Please carefully check before asking for review.

- [x] `virtual` is still there.  Please remove all `virtual` to avoid polymorphism.
- [x] Do not provide initial value to `m_begin` and `m_end` in `BufferBase`.
- [x] Make `std::enable_shared_from_this` be the first parent.

tigercosmos:
> Do you still think making the common `BufferBase` class template is a good idea? 

I think we reduce a lot of duplicated code, which is better than before.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: fixed

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: fixed

tigercosmos: cc @ThreeMonth03

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: cc @ThreeMonth03

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: sorry... it appear when I revert commit... removed.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: sure, fixed.

cpp/modmesh/buffer/ConcreteBuffer.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: fixed.

tigercosmos:
- [x] Do not use polymorphism.
- [x]  virtual is still there. Please remove all virtual to avoid polymorphism.
- [x]  Do not provide initial value to m_begin and m_end in BufferBase.
- [x] Make std::enable_shared_from_this be the first parent.

tigercosmos:
@yungyuc i think all issues should be correctly fixed.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -70,19 +72,24 @@ class BufferExpander
     }
 
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
-        : m_concrete_buffer(clone ? buf->clone() : buf)
-        , m_begin(m_concrete_buffer->data())
-        , m_end(m_begin + m_concrete_buffer->size())
-        , m_end_cap(m_begin + m_concrete_buffer->size())
+        : BufferBase<BufferExpander>(nullptr, nullptr) //  initialize m_begin and m_end, but since we don't have the data yet, we set them to nullptr
```
yungyuc: Don't delegate to the parent constructor setting `nullptr`.  The pointers will be overridden and explicitly initializing here wastes cycles.

Please review delegates and remove unnecessary ones.

It will be good to add a comment line explaining why not delegate.

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -168,9 +170,12 @@ class ConcreteBuffer
      *      Size of the memory buffer in bytes.
      */
     ConcreteBuffer(size_t nbytes, const ctor_passkey &)
-        : m_nbytes(nbytes)
+        : BufferBase<ConcreteBuffer>(nullptr, nullptr) //  initialize m_begin and m_end, but since we don't have the data yet, we set them to nullptr
```
yungyuc: Don't delegate to the parent constructor setting `nullptr`.  The pointers will be overridden and explicitly initializing here wastes cycles.

yungyuc (changes requested):
- [x] Don't waste cycles to initialize `nullptr` which will be overridden.

cpp/modmesh/buffer/ConcreteBuffer.hpp
tigercosmos: fixed.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: fixed.

tigercosmos:
@yungyuc the last issue is addressed.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -70,19 +72,24 @@ class BufferExpander
     }
 
     BufferExpander(std::shared_ptr<ConcreteBuffer> const & buf, bool clone, ctor_passkey const &)
-        : m_concrete_buffer(clone ? buf->clone() : buf)
-        , m_begin(m_concrete_buffer->data())
-        , m_end(m_begin + m_concrete_buffer->size())
-        , m_end_cap(m_begin + m_concrete_buffer->size())
+        : BufferBase<BufferExpander>() // don't delegate m_begin and m_end, which will be overwritten later
```
yungyuc, line 75: The comment is clear.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -117,52 +115,9 @@ class BufferExpander
...
-    int8_t & data(size_type it) { return data()[it]; }
-    int8_t const * data() const noexcept { return data<int8_t>(); }
-    int8_t * data() noexcept { return data<int8_t>(); }
-    // clang-format off
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    template <typename T> T const * data() const noexcept { return reinterpret_cast<T *>(m_begin); }
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    template <typename T> T * data() noexcept { return reinterpret_cast<T *>(m_begin); }
-    // clang-format on
+    static constexpr const char * name() { return "BufferExpander"; }
```
yungyuc, line 122: For your information.  In my memory writing this in a header file may result in multiple symbols if the header is used in multiple build units.  But maybe the behavior is changed in a newer standard.  It's fine as long as the code builds.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -70,19 +72,24 @@ class BufferExpander
...
+        : BufferBase<BufferExpander>() // don't delegate m_begin and m_end, which will be overwritten later
+        , m_concrete_buffer(clone ? buf->clone() : buf)
     {
+        m_begin = m_concrete_buffer->data(); // override m_begin and m_end once we have the data
+        m_end = m_begin + m_concrete_buffer->size();
+        m_end_cap = m_begin + m_concrete_buffer->size();
     }
 
     BufferExpander(size_type nbyte, ctor_passkey const &)
+        : BufferBase<BufferExpander>() // don't delegate m_begin and m_end, which will be overwritten later
```
yungyuc: This looks fishy.  IIUC, `expand()` needs `m_begin` to be valid in an execution path.  So `m_begin` needs to be initialized.

cpp/modmesh/buffer/BufferExpander.hpp
```diff
@@ -70,19 +72,24 @@ class BufferExpander
...
     {
         expand(nbyte);
     }
 
-    BufferExpander(ctor_passkey const &) {}
+    BufferExpander(ctor_passkey const &)
+        : BufferBase<BufferExpander>() // don't delegate m_begin and m_end, which will be overwritten later
```
yungyuc: I think this needs `m_begin` and `m_end` to be initialized.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+#include <modmesh/base.hpp>
+
+namespace modmesh
+{
+
+/// Base class for buffer-like objects.
+template <typename Derived>
+class BufferBase
```
yungyuc, line 38: Explicitly write all default implementations of constructors and destructor.

yungyuc (changes requested):
- [x] Review `m_begin` and `m_end`.
- [x] List all constructors and destructor of `BufferBase`.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: it's `static`, so should be only one instance.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: no, `expand()` calls `reserve()`, which initialized `m_begin` and `m_end`.

cpp/modmesh/buffer/BufferBase.hpp
tigercosmos: sure, added.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: no, `expand()` calls `reserve()`, which initialized `m_begin` and `m_end`.

cpp/modmesh/buffer/BufferExpander.hpp
tigercosmos: sure, fixed the logic.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: `reserve()` calls [`capacity()`](https://github.com/solvcon/modmesh/blob/17ce957c13acde8d40963be6b4b56af6bf0fa037/cpp/modmesh/buffer/BufferExpander.hpp#L103) which reads `m_begin` before initialization:

```cpp
    size_type capacity() const noexcept
    {
        return static_cast<size_type>(this->m_end_cap - this->m_begin);
    }
```

tigercosmos: thanks for catching this lol, I am surprised the test passed on macOS, seems it's a UB.

tigercosmos:
the callprofiler timing issue (https://github.com/solvcon/modmesh/issues/347) happens again: https://github.com/solvcon/modmesh/actions/runs/9535397400/job/26281064195?pr=363

tigercosmos:
@yungyuc  I fixed the incorrect `m_begin` issue. please check it.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: If this function `name()` has a symbol in the binary, the function implementation needs to be outside the header file.  Otherwise, each compilation unit will have a symbol and we will end up with multiple symbols in a binary.  Most of the time we do not allow multiple symbols during linking and the linker will error out.

It this line builds without a screaming linker, maybe it's not taking an address because of `constexpr`.  I would guess this is the case.  Another possibility is that the standard/compiler is updated to accommodate this.  But it's just my argumentum ad ignorantiam.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,122 @@
...
+    {
+        return Derived::name();
+    }
+
+protected:
+    BufferBase() = default;
+
+    BufferBase(int8_t * start, int8_t * end)
+        : m_begin(start)
+        , m_end(end){};
```
yungyuc, line 99: `{}` should use a new line, and it does not need `;` at the end.  This is minor.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    BufferBase(BufferBase const &) = delete;
```
yungyuc, line 101: It's good to delete unwanted constructors.

cpp/modmesh/buffer/BufferExpander.hpp
yungyuc: Or we are just lucky (or unlucky).  I think no test is catching this on any platform.  It's commonplace so it's important to code with rigor.

yungyuc (approved):
LGTM.  I'd like to merge to get the code living for a while.

PR #364, opened 2024-06-15
use systematic way to locate third-party lib

### LINKED ISSUE #280: The way of finding PUI needs to be improved
The way of finding PUI in [`modmesh/view.py`](https://github.com/solvcon/modmesh/blob/master/modmesh/view.py#L67) is using hard-code import path, which is not a good approach.
This is a point that needs improvement.

j8xixo12:
I added a python script in `thirdparty` folder that provides `register_library` and `load_library` for modmesh to select which  library to load.

### DIFF
--- a/modmesh/core.py
+++ b/modmesh/core.py
@@ -35,7 +35,7 @@ three-dimensional space.
 
 import os
 
-from .pylibmgr import search_library_root
+from . import pylibmgr
 
 __all__ = [  # noqa: F822
     'WrapperProfilerStatus',
@@ -110,7 +110,7 @@ def _load():
 
     # Walk through the thirdparty folder and register all library
     # into a dictionary.
-    search_library_root(os.getcwd(), 'thirdparty')
+    pylibmgr.search_library_root(os.getcwd(), 'thirdparty')
 
 
 _load()
--- a/modmesh/core.py
+++ b/modmesh/core.py
@@ -33,6 +33,8 @@ three-dimensional space.
 
 # Use flake8 http://flake8.pycqa.org/en/latest/user/error-codes.html
 
+import sys
+import os
 
 __all__ = [  # noqa: F822
     'WrapperProfilerStatus',
@@ -104,6 +106,34 @@ except ImportError:
 def _load():
     for name in __all__:
         globals()[name] = getattr(_impl, name)
+
+    # Try to find the thirdparty, if failed to find it
+    # modmesh will raise ImportError and remind the user.
+    filename = os.path.join('thirdparty')
+    path = os.getcwd()
+    try:
+        while True:
+            if os.path.exists(os.path.join(path, filename)):
+                break
+            if path == os.path.dirname(path):
+                path = None
+                break
+            else:
+                path = os.path.dirname(path)
+        if path is None or not os.path.exists(os.path.join(path,
+                                                           filename)):
+            raise ImportError
+    except ImportError:
+        sys.stderr.write('Can not find thirdparty.\n')
+        sys.stderr.write('No python 3rd-party library will be imported\n')
+
+    path = os.path.join(path, filename)
+    sys.path.append(path)
+
+    # Walk through the thirdparty folder and register all library
+    # into a dictionary.
+    from python_lib_management import register_library
+    register_library()
 
 
 _load()
--- a/modmesh/core.py
+++ b/modmesh/core.py
@@ -33,8 +33,9 @@ three-dimensional space.
 
 # Use flake8 http://flake8.pycqa.org/en/latest/user/error-codes.html
 
-import sys
 import os
+
+from .pylibmgr import search_library_root
 
 __all__ = [  # noqa: F822
     'WrapperProfilerStatus',
@@ -107,33 +108,9 @@ def _load():
     for name in __all__:
         globals()[name] = getattr(_impl, name)
 
-    # Try to find the thirdparty, if failed to find it
-    # modmesh will raise ImportError and remind the user.
-    filename = os.path.join('thirdparty')
-    path = os.getcwd()
-    try:
-        while True:
-            if os.path.exists(os.path.join(path, filename)):
-                break
-            if path == os.path.dirname(path):
-                path = None
-                break
-            else:
-                path = os.path.dirname(path)
-        if path is None or not os.path.exists(os.path.join(path,
-                                                           filename)):
-            raise ImportError
-    except ImportError:
-        sys.stderr.write('Can not find thirdparty.\n')
-        sys.stderr.write('No python 3rd-party library will be imported\n')
-
-    path = os.path.join(path, filename)
-    sys.path.append(path)
-
     # Walk through the thirdparty folder and register all library
     # into a dictionary.
-    from python_lib_management import register_library
-    register_library()
+    search_library_root(os.getcwd(), 'thirdparty')
 
 
 _load()
--- a/modmesh/pylibmgr.py
+++ b/modmesh/pylibmgr.py
@@ -26,11 +26,12 @@
 
 import sys
 import os
+import time
 
 lib_path = {}
 
 
-def register_library():
+def search_library_root(curr_path, lib_root_name, timeout=1):
     """
     Walk through the thridparty library and register all third-party
     library folder path into a dictionary with library name as the key,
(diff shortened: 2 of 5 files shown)


modmesh/core.py
```diff
@@ -105,6 +107,34 @@ def _load():
...
+    try:
+        while True:
+            if os.path.exists(os.path.join(path, filename)):
+                break
+            if path == os.path.dirname(path):
+                path = None
+                break
+            else:
+                path = os.path.dirname(path)
+        if path is None or not os.path.exists(os.path.join(path,
```
yungyuc: It is harder to read the lines when they are broken like:

```python
if path is None or not os.path.exists(os.path.join(path,
                                                   filename)):
```

In this case, breaking the line using a local variable is more readable:

```python
_pf = os.path.join(path, filename)
if path is None or not os.path.exists(_pf):
```

modmesh/core.py
```diff
@@ -105,6 +107,34 @@ def _load():
...
+                                                           filename)):
+            raise ImportError
+    except ImportError:
+        sys.stderr.write('Can not find thirdparty.\n')
+        sys.stderr.write('No python 3rd-party library will be imported\n')
```
yungyuc: Do you need the trailing `\n`?

modmesh/core.py
```diff
@@ -105,6 +107,34 @@ def _load():
     for name in __all__:
         globals()[name] = getattr(_impl, name)
 
+    # Try to find the thirdparty, if failed to find it
+    # modmesh will raise ImportError and remind the user.
+    filename = os.path.join('thirdparty')
+    path = os.getcwd()
+    try:
+        while True:
```
yungyuc: You need a unit test for the loop so that when it goes wrong and into infinite, there is a place to add regression.

modmesh/view.py
```diff
@@ -53,6 +52,7 @@
 enable = False
 try:
     from _modmesh import view as _vimpl  # noqa: F401
+    from python_lib_management import load_library
```
yungyuc: It is very hard to know where to find `python_lib_management`.  A maintainer needs to find everywhere in the source package.  It will be hopeless when `modmesh` is deployed to system.

If you use it this way, I think the module needs to be placed under `modmesh` namespace.  If it's placed under `modmesh`, the module needs a better (more precise) name.

modmesh/core.py
```diff
@@ -105,6 +107,34 @@ def _load():
     for name in __all__:
         globals()[name] = getattr(_impl, name)
 
+    # Try to find the thirdparty, if failed to find it
```
yungyuc: The whole logic should be placed in a function and documented in the docstring.  It's hard to understand.

thirdparty/python_lib_management.py
yungyuc, line 1: You are not making this file a script but a module, and make it unreasonable to place the file under `thirdparty/`.

yungyuc (changes requested):
- [x] Make `thirdparty/` path searching testable and add the docstrings and tests.
- [x] Beautify line breaking.
- [x] Discuss where to place the module `python_lib_management` and its name.

yungyuc:
This PR is for #280 .  @j8xixo12 , next time, please note the association to the issue (write "ref xxx" or "for xxx", not "resolve" or "address" xxx which may close the associated issue.)

modmesh/view.py
j8xixo12: I will move the `python_lib_management` from `thirdparty` to `modmesh`.
How about rename it to `libutil` ?

modmesh/core.py
j8xixo12: Yes, without tailing `\n`, the message on the screen will become:
```
Can not find thirdparty.No python 3rd-party library will be imported
```

modmesh/view.py
yungyuc: I would like to have more descriptive name.  Let's try `pylibmgr`?

modmesh/core.py
yungyuc: I meant the `\n` in `... imported\n`.

j8xixo12: Oh, it still needs `\n` after `....import`, without `\n`, the shell prompt will not start on a  newline:
```
Can not find thirdparty.
No python 3rd-party library will be imported$
```
it will not look not good.

modmesh/view.py
j8xixo12: LGTM, let's try `pylibmgr`!

modmesh/core.py
yungyuc: I see.  Thanks.

modmesh/pylibmgr.py
```diff
@@ -0,0 +1,85 @@
...
+    user can select which library need to be imported by library name.
+
+    :return: None
+    """
+    # Try to find the library root, if failed to find it
+    # modmesh will raise ImportError and remind the user.
+    folder_name = os.path.join(lib_root_name)
+    start_time = time.time()
+    try:
+        while time.time() - start_time < timeout:
```
j8xixo12, line 47: Use a timeout here to ensure `modmesh` does not get stuck in an infinite loop.

modmesh/pylibmgr.py
```diff
@@ -0,0 +1,83 @@
...
+    user can select which library need to be imported by library name.
+
+    :return: None
+    """
+    # Try to find the library root, if failed to find it
+    # modmesh will raise ImportError and remind the user.
+    folder_name = os.path.join(lib_root_name)
+    _path = curr_path
+    start_time = time.time()
+    while time.time() - start_time < timeout:
```
j8xixo12, line 53: Use a timeout here to ensure `modmesh` does not stuck in an infinite loop.

modmesh/core.py
```diff
@@ -33,6 +33,9 @@
 
 # Use flake8 http://flake8.pycqa.org/en/latest/user/error-codes.html
 
+import os
+
+from .pylibmgr import search_library_root
```
yungyuc: Use `from . import pylibmgr` and `pylibmgr.search_library_root()` later.  It will be easier to find module use outside this file during maintenance.

modmesh/pylibmgr.py
```diff
@@ -0,0 +1,83 @@
...
+
+lib_path = {}
+
+
+def search_library_root(curr_path, lib_root_name, timeout=1):
+    """
+    Walk through the thridparty library and register all third-party
+    library folder path into a dictionary with library name as the key,
+    user can select which library need to be imported by library name.
+
```
yungyuc, line 39: Add the docstring for `timeout`:

```
:ivar timeout: Maximum time for root path searching.  Default is 1.0 second.
```

modmesh/pylibmgr.py
```diff
@@ -0,0 +1,83 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+import os
+import time
+
+lib_path = {}
+
+
+def search_library_root(curr_path, lib_root_name, timeout=1):
```
yungyuc: Since `timeout` is in second, it's clearer to write the default value as a float.

modmesh/pylibmgr.py
yungyuc: This is nice.

modmesh/pylibmgr.py
```diff
@@ -0,0 +1,83 @@
...
+def load_library(lib_name):
+    """
+    Append library path into python path by user input library name
+
+    :param lib_name: library name that need to be imported.
+    :return: None
+    """
+    try:
+        if lib_name in lib_path:
+            sys.path.append(lib_path[lib_name])
```
yungyuc, line 82: Appending `sys.path` deteriorate maintainability because it is hard to find this line and the value input to `append()` later during maintenance.  But I agree this is a straight-forward way to make the thirdparty importable.

If we have to do this, please create an issue to track for the improvement right before this is to be merged, and add a code comment to point to the issue.

modmesh/view.py
```diff
@@ -53,6 +52,7 @@
 enable = False
 try:
     from _modmesh import view as _vimpl  # noqa: F401
+    from .pylibmgr import load_library
```
yungyuc: Ditto. Use module import.

tests/test_pylibmgr.py
```diff
@@ -0,0 +1,47 @@
...
+
+import os
+
+import modmesh
+
+
+class pylibmgrTC(unittest.TestCase):
+
+    def test_pylibmgr_search_library_root(self):
+        modmesh.pylibmgr.search_library_root(os.getcwd(), 'thirdparty')
```
yungyuc, line 39: Please add a code comment stating that this line assumes the test directory is in the source package, in which the `modmesh` Python directory exists.

yungyuc (changes requested):
- [x] Use module import and access functions inside, instead of import the function.
- [x] Use float for `timeout` argument default and add docstring.
- [x] Deal with the difficulty of maintenance with `sys.path` entry insertion.

tests/test_pylibmgr.py
j8xixo12: I'm thinking this test case assume that the third-party library root folder name is `thirdparty`. If the `thirdparty` is renamed, we will need to modify this test case.
If I create a folder in `tests` by `os.mkdir` in `setUp`, this folder can be used for `test_pylibmgr_search_library_root` and this folder will be removed in `tearDown`.

I'm not sure if it's appropriate that create / remove folder in unit-test code ?

yungyuc: I don't think we need anything more than a code comment at the moment.

modmesh/pylibmgr.py
j8xixo12: I have created an issue https://github.com/solvcon/modmesh/issues/378 to track the improvement for importing python third-party library.

j8xixo12:
- [x] Use module import and access functions inside, instead of import the function.
- [x] Use float for `timeout` argument default and add docstring.
- [x] Deal with the difficulty of maintenance with `sys.path` entry insertion.

@yungyuc This PR is ready to review!

yungyuc:
LGTM.  Merging.

yungyuc:
ref #280

PR #370, opened 2024-06-18
Viewer Windows Nightly Build

### LINKED ISSUE #194: viewer Windows executable should include dll for portability
Currently the viewer itself (whether user execute it via Visual Studio or not) is unable to perform as intended without set `$QT6_DIR\$VERSION\$PLATFORM\bin` to environment path first.

Not only it confuses to new user (As one needs to modify system path obviously is an unspoken rule), it also decreases the portability of the project.

After discussion, @tychuang1211 and I proposed three possible solutions: (other solutions are also welcomed!)
 
1. Add Qt binary paths to VS projects
    * pros: 
        * Easy to implement
    * cons: 
        * Only solves the case when the viewer is launched via Visual Studio
        * Poor portability 
 
1. Copy dll & necessary files into the folder of the executable when generate VS project
    * pros: 
        * Easy to implement
        * Good portability 
    * cons: 
        * Increase executable folder size
        * Developers had to aware of that the dlls are copied when CMake generates project.
    
1. Stay same for now. We'll built release build into single executable in future anyway.
    * pros: 
        * No extra work
    * cons: 
        * The unspoken rule remains
  
Let's discuss which solution fits us most in the moment.

terrychan999:
In this PR, the POC for a portable viewer on Windows https://github.com/solvcon/modmesh/issues/194#issuecomment-2158610271 has been implemented into the `devbuild` CI, enabling the nightly build process.

The artifact size is now ~210MB compressed and ~530MB after extraction, with the PySide6 shipped Qt DLLs removed.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -552,3 +552,61 @@ jobs:
           cmake --build ${{ github.workspace }}/build `
             --config ${{ matrix.cmake_build_type }} `
             --target run_viewer_pytest
+
+      - name: generate portable
+        if: ${{ matrix.cmake_build_type == 'Release' }}
+        run: |
+          # Get the Python version installed and extract the major+minor version components
+          $py_ver=$(python3 -V | awk '{print $2}')
+          $py_main_ver=$(echo $py_ver | awk -F. '{print $1$2}')
+          $destination=".\modmesh-viewer-win64\modmesh-viewer-win64"
+          $py_exec="$destination\python.exe"
+
+          # Create the destination directory
+          New-Item -ItemType Directory -Path $destination
+
+          # Download and extract the Python embeddable package
+          $pyembed_url="https://www.python.org/ftp/python/$py_ver/python-$py_ver-embed-amd64.zip"
+          $pyembed_dl="python-$py_ver-embed-amd64.zip"
+          Invoke-WebRequest -Uri $pyembed_url -OutFile $pyembed_dl
+          Expand-Archive -Path $pyembed_dl -DestinationPath $destination
+
+          # Patch the .pth file to enable 'site' package (required for using pip)
+          $pth_file="$destination\python$py_main_ver._pth"
+          (Get-Content $pth_file) -replace '#import site', 'import site' | Set-Content $pth_file
+
+          # Download and setup pip installation script
+          $get_pip_url="https://bootstrap.pypa.io/get-pip.py"
+          $get_pip_script="$destination\get-pip.py"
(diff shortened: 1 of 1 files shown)


.github/workflows/devbuild.yml
```diff
@@ -552,3 +552,61 @@ jobs:
           cmake --build ${{ github.workspace }}/build `
             --config ${{ matrix.cmake_build_type }} `
             --target run_viewer_pytest
+
+      - name: generate portable
+        if: ${{ matrix.cmake_build_type == 'Release' }}
+        run: |
+          # Get the Python version installed and extract the major+minor version components
```
yungyuc, line 559: It is good enough to have the commands listed in the Github Actions yaml file for now.  In the future, would it be useful to absorb the commands to a standalone script file.  Or even make a Python module to do it?  (This is only for discussion.  No action is needed for this PR.)

yungyuc (approved):
This looks good, @terrychan999 , and the built portable works great on my _arm_ (_apple silicons_) windows VM!

image

@j8xixo12 @Gene0315 please take a look at it.

---

@terrychan999 I wonder if it is possible to provide a download link to the nightly built artifact in [`README.md`](https://github.com/solvcon/modmesh/blob/4d1a1bbe2ea93118f3ec00fe3b7692a8caeb1941/README.md?plain=1#L1)?

If a link is not feasible, maybe provide some descriptions in the `README.rst` for how to find the artifact to download?

yungyuc:
@terrychan999 Do you think I should merge it right away?  (I can.)  Or do you want to use this PR to add something to `README.md`?

.github/workflows/devbuild.yml
terrychan999: Sure, this is a great suggestion. Integrating these commands into a standalone script or Python module would make on-premise development and testing more competent.
Thanks, I'll consider this for future improvements.

terrychan999:
> @terrychan999 Do you think I should merge it right away? (I can.) Or do you want to use this PR to add something to `README.md`?

I would like to merge it right away.
I am considering generating the artifacts into [Releases](https://github.com/solvcon/modmesh/releases) so that users can easily download the nightly builds.

yungyuc:
Merged.  @terrychan999 Using releases sounds good.  Could you please create an issue for it if you haven't done so?  Issue #194 can be fixed by this PR.

yungyuc:
> Merged. @terrychan999 Using releases sounds good. Could you please create an issue for it if you haven't done so? Issue #194 can be fixed by this PR.

@terrychan999 An issue #371 is created and assigned to you.

PR #373, opened 2024-06-22
move ConcreteBuffer implementation to CPP file

tigercosmos:
part of https://github.com/solvcon/modmesh/issues/305

Move `ConcreteBuffer` implementation to CPP file.

I only moved the functions that have more than 1 line. If the function is too short as 1 line, keep it in the HPP file.


cpp/modmesh/buffer/ConcreteBuffer.cpp
```diff
@@ -0,0 +1,107 @@
+/*
```
tigercosmos, line 1: nothing new, only copy and paste

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -134,6 +134,8 @@ class ConcreteBuffer
 public:
 
     using remover_type = detail::ConcreteBufferRemover;
+    // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays)
+    using unique_ptr_type = std::unique_ptr<int8_t, data_deleter_type>;
```
tigercosmos, line 138: I decided to put this declaration to the front.

cpp/modmesh/buffer/ConcreteBuffer.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * Copyright (c) 2024, An-Chi Liu <phy.tiger@gmail.com>
```
yungyuc, line 2: When copying and paste, please retain the original copyright tag.  It is required by the license pasted here.

cpp/modmesh/buffer/ConcreteBuffer.cpp
yungyuc: This file is too short to make distinct implementation file useful.  It does not make code easier to read.  In contrary, I think it makes it harder to read the code because function bodies are moved away from the header file.

And I think the code split in the new file is too short to significantly shorten the compilation time.  It is so short that measuring it rigorously probably does not make sense.

yungyuc (commented):
The separate code is too short (only less than 80 lines) to justify the need for the separation.  

A defect: 
- Keep copyright declaration intact when copying and pasting.

yungyuc:
I don't think we need this separation.  Close the PR but we can discuss.

tigercosmos:
If you don't this it's needed, then the issue can be closed. @yungyuc

yungyuc:
Thanks, @tigercosmos .  I closed the refactoring issue #305.

PR #375, opened 2024-06-23
Add Reset within View menu

YenPeiChen07:
Hi @yungyuc and @tychuang1211,

This PR is related to camera improvement in #70. 

I added a Reset menu item under View menu. When clicking it (or pressing `esc` key), objects in all subwindows will be reset to initial state. 

Here are some demo images:

Screenshot 2024-06-23 at 3 44 22 PM

The initial mesh object:
Screenshot 2024-06-23 at 3 16 34 PM

Reset to initial state:
Screenshot 2024-06-23 at 3 16 56 PM

### DIFF
--- a/cpp/modmesh/view/R3DWidget.cpp
+++ b/cpp/modmesh/view/R3DWidget.cpp
@@ -103,11 +103,14 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
     m_container->resize(event->size());
 }
 
-void R3DWidget::resetCamera(Qt3DRender::QCamera * camera)
+void R3DWidget::resetCamera(Qt3DRender::QCamera * camera,
+                            float positionX,
+                            float positionY,
+                            float positionZ)
 {
     // Set up the camera.
     camera->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
-    camera->setPosition(QVector3D(0.0f, 0.0f, 10.0f));
+    camera->setPosition(QVector3D(positionX, positionY, positionZ));
     camera->setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
 
     // Set up the camera control.
--- a/cpp/modmesh/view/R3DWidget.cpp
+++ b/cpp/modmesh/view/R3DWidget.cpp
@@ -103,6 +103,20 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
     m_container->resize(event->size());
 }
 
+void R3DWidget::resetCamera(Qt3DRender::QCamera * camera)
+{
+    // Set up the camera.
+    camera->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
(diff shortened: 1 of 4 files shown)


yungyuc (changes requested):
This looks good.  Could you please provide a Python wrapper to the reset function?  The Python wrapper should also take an optional argument for the value that the camera is reset to.

And don't forget to run linter before pushing.

yungyuc:
@j8xixo12 Could you please also help take a look?

YenPeiChen07:
> This looks good. Could you please provide a Python wrapper to the reset function? The Python wrapper should also take an optional argument for the value that the camera is reset to.
> 
> And don't forget to run linter before pushing.

Ok, I would take a look how to implement a Python wrapper to the reset function.

cpp/modmesh/view/R3DWidget.cpp
```diff
@@ -103,6 +103,20 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
     m_container->resize(event->size());
 }
 
+void R3DWidget::resetCamera(Qt3DRender::QCamera * camera)
+{
+    // Set up the camera.
+    camera->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
+    camera->setPosition(QVector3D(0.0f, 0.0f, 10.0f));
```
j8xixo12: Is possible to determine the z-axis position of the camera based on the size of the displayed object ?
If the object is too large, the camera might too close then the entire object will not be fitted in the screen.

yungyuc:
> Ok, I would take a look how to implement a Python wrapper to the reset function.

Thanks, @YenPeiChen07 .  Let's sync up again later this week.

cpp/modmesh/view/R3DWidget.cpp
YenPeiChen07: Thanks for feedback. I will research it further.

YenPeiChen07:
@yungyuc, 

I added a python wrapper to the reset function in `wrap_view.cpp`. 

By default, the camera is reset to (0.0, 0.0, 10.0) when calling `.resetCamera()`. User can pass a coordinate value they want by calling `.resetCamera(positionX, positionY, positionZ)` in python console.

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -111,6 +111,10 @@ class R3DWidget
     Qt3DExtras::Qt3DWindow * view() { return m_view; }
     RScene * scene() { return m_scene; }
     Qt3DRender::QCamera * camera() { return m_view->camera(); }
+    void resetCamera(Qt3DRender::QCamera * camera,
+                     float positionX = 0.0f,
```
yungyuc, line 115: I am not sure using default parameters here is good for maintenance.  If no one objects, we can leave it here and observe.  (My hunch feeling is that it's gonna bite us soon.  We'll see.)

yungyuc (approved):
LGTM.  This is a simple addition.  Let's add more enhancements with or on top of it.

PR #377, opened 2024-06-25
SerializableItem: implement JSON from string

tigercosmos:
Support JSON from string
- add `from_json` into `MM_DECL_SERIALIZABLE`
- implemented `JsonNode`, which is the token of JSON
- implemented JSON parser: `parse_array`, `parse_object`
- added a helper class: `JsonHelper`

### DIFF
--- a/cpp/modmesh/base.hpp
+++ b/cpp/modmesh/base.hpp
@@ -72,6 +72,11 @@ struct is_specialization_of<Template, Template<Args...>> : std::true_type
 template <template <typename...> class Template, typename T>
 inline constexpr bool is_specialization_of_v = is_specialization_of<Template, T>::value;
 
+inline bool is_whitespace(char c)
+{
+    return c == ' ' || c == '\t' || c == '\n' || c == '\r' || c == '\f' || c == '\v';
+}
+
 } /* end namespace detail */
 
 using real_type = double;
--- a/cpp/modmesh/serialization/SerializableItem.cpp
+++ b/cpp/modmesh/serialization/SerializableItem.cpp
@@ -34,6 +34,16 @@ namespace modmesh
 namespace detail
 {
 
+inline bool is_json_number(char c)
+{
+    return std::isdigit(c) || c == '.' || c == 'e' || c == 'E' || c == '+' || c == '-';
+}
+
+inline void throw_serialization_error(Formatter & formatter, int line, int column)
+{
+    throw std::runtime_error(formatter << " (line: " << line << ", column: " << column << ")");
+}
+
 std::string escape_string(std::string_view str_view)
 {
     std::ostringstream oss;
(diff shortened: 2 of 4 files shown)


cpp/modmesh/base.hpp
```diff
@@ -72,6 +72,30 @@ struct is_specialization_of<Template, Template<Args...>> : std::true_type
 template <template <typename...> class Template, typename T>
 inline constexpr bool is_specialization_of_v = is_specialization_of<Template, T>::value;
 
+inline bool is_alpha(char c)
```
tigercosmos: some common helper functions

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -27,13 +27,17 @@
...
 
 #include <modmesh/serialization/SerializableItem.hpp>
-
 namespace modmesh
 {
 
 namespace detail
 {
 
+inline void throw_serialization_error(Formatter & formatter, int line, int column)
```
tigercosmos, line 42: print readable error message for parsing json

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,460 @@ std::string escape_string(std::string_view str_view)
...
 
+std::string trim_string(const std::string & str)
+{
+    const char * whitespace = " \t\n\r\f\v";
+    size_t start = str.find_first_not_of(whitespace);
+    size_t end = str.find_last_not_of(whitespace);
+    return (start == std::string::npos) ? "" : str.substr(start, end - start + 1);
+}
+
+#define MM_DECL_CACULATE_LINE_COLUMN(CH) \
```
tigercosmos, line 97: for the error message

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,460 @@ std::string escape_string(std::string_view str_view)
...
+    {                                    \
+        line += 1;                       \
+        column = 1;                      \
+    }                                    \
+    else                                 \
+    {                                    \
+        column += 1;                     \
+    }
+
+JsonArray JsonNode::parse_array(const std::string & json)
```
tigercosmos, line 110: a state machine parsing JSON array, same for the `parse_object`

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,460 @@ std::string escape_string(std::string_view str_view)
...
+                    MM_DECL_CACULATE_LINE_COLUMN(json[index])
+                    index += 1;
+                }
+
+                if (depth != 0)
+                {
+                    throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+                }
+
+                json_array.emplace_back(std::make_unique<JsonNode>(JsonType::Object, value_expression));
```
tigercosmos, line 180: parse the child, which is a JSON object here.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,460 @@ std::string escape_string(std::string_view str_view)
...
+                    MM_DECL_CACULATE_LINE_COLUMN(json[index])
+                    index += 1;
+                }
+
+                if (depth != 0)
+                {
+                    throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+                }
+
+                json_array.emplace_back(std::make_unique<JsonNode>(JsonType::Array, value_expression));
```
tigercosmos: the child is array.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,460 @@ std::string escape_string(std::string_view str_view)
...
+                    {
+                        type = JsonType::Number;
+                    }
+                    else
+                    {
+                        throw_serialization_error(Formatter() << "Invalid JSON format: invalid value expression: " << value_expression, line, column);
+                    }
+                }
+
+                json_array.emplace_back(std::make_unique<JsonNode>(type, value_expression));
```
tigercosmos, line 246: the basic types go to here.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,433 @@ std::string escape_string(std::string_view str_view)
...
+                {
+                    throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+                }
+
+                json_array.emplace_back(std::make_unique<JsonNode>(JsonType::Object, value_expression));
+            }
+            else if (c == '[')
+            {
+                // array in array is valid in JSON, but we do not support it
+                throw_serialization_error(Formatter() << "Invalid JSON format: unexpected array in array.", line, column);
```
tigercosmos, line 185: not support array in array for now.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,433 @@ std::string escape_string(std::string_view str_view)
...
+
+    if (state != JsonState::End)
+    {
+        throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+    }
+
+    return json_array;
+}
+
+JsonMap JsonNode::parse_object(const std::string & json)
```
tigercosmos, line 288: state machine for JSON object.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,433 @@ std::string escape_string(std::string_view str_view)
...
+                    MM_DECL_CACULATE_LINE_COLUMN(json[index])
+                    index += 1;
+                }
+
+                if (depth != 0)
+                {
+                    throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+                }
+
+                json_map.emplace(key, std::make_unique<JsonNode>(JsonType::Object, value_expression));
```
tigercosmos, line 395: nested object.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,433 @@ std::string escape_string(std::string_view str_view)
...
+                    MM_DECL_CACULATE_LINE_COLUMN(json[index])
+                    index += 1;
+                }
+
+                if (depth != 0)
+                {
+                    throw_serialization_error(Formatter() << "Invalid JSON format: missing closing bracket.", line, column);
+                }
+
+                json_map.emplace(key, std::make_unique<JsonNode>(JsonType::Array, value_expression));
```
tigercosmos, line 428: the child is array.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+    Start,
+    ObjectKey,
+    Column,
+    Comma,
+    ObjectValue,
+    End,
+}; /* end enum class JsonState */
+
+/// Type of JSON token.
+enum class JsonType
```
tigercosmos, line 74: support common types for now.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+    static JsonArray parse_array(const std::string & json);
+
+}; /* end struct JsonNode */
+
+/// Helper class for JSON serialization and deserialization.
+class JsonHelper
+{
+public:
+    template <typename T>
+    static std::string to_json_string(const T & value)
```
tigercosmos: move this function into the helper class.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+        }
+        else
+        {
+            return std::to_string(value);
+        }
     }
-    else if constexpr (is_specialization_of<std::vector, T>::value)
+
+    template <typename T>
+    static void from_json_string(const detail::JsonNodePtr & node, T & value)
```
tigercosmos: save the value into the class.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+            separator = ",";                                                                        \
+        };                                                                                          \
+        __VA_ARGS__                                                                                 \
+        oss << "}";                                                                                 \
+        return oss.str();                                                                           \
+    }                                                                                               \
+                                                                                                    \
+    void from_json(const std::string & json) override                                               \
+    {                                                                                               \
+        auto jsonNode = std::make_unique<detail::JsonNode>(detail::JsonType::Object, json);         \
```
tigercosmos, line 302: parse the JSON first.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+        auto register_member = [&](const char * name, auto && value)                                \
+        {                                                                                           \
+            if (jsonNode->type == detail::JsonType::Object)                                         \
+            {                                                                                       \
+                auto & obj = std::get<detail::JsonMap>(jsonNode->value);                            \
+                auto it = obj.find(name);                                                           \
+                if (it != obj.end())                                                                \
+                {                                                                                   \
+                    detail::JsonHelper::from_json_string(it->second, value);                        \
```
tigercosmos, line 311: then we can save the values from the parsed nodes into class's values.

cpp/modmesh/base.hpp
```diff
@@ -72,6 +72,30 @@ struct is_specialization_of<Template, Template<Args...>> : std::true_type
...
+{
+    return c >= '0' && c <= '9';
+}
+
+inline bool is_whitespace(char c)
+{
+    return c == ' ' || c == '\t' || c == '\n' || c == '\r';
+}
+
+template <typename T>
```
yungyuc: Use STL `is_integral`: https://en.cppreference.com/w/cpp/types/is_integral

cpp/modmesh/base.hpp
```diff
@@ -72,6 +72,30 @@ struct is_specialization_of<Template, Template<Args...>> : std::true_type
...
+constexpr bool is_integer_v = std::is_same_v<T, int8_t> || std::is_same_v<T, int16_t> ||
+                              std::is_same_v<T, int32_t> || std::is_same_v<T, int64_t> ||
+                              std::is_same_v<T, uint8_t> || std::is_same_v<T, uint16_t> ||
+                              std::is_same_v<T, uint32_t> || std::is_same_v<T, uint64_t>;
+
+template <typename T>
+constexpr bool is_float_v = std::is_same_v<T, float> || std::is_same_v<T, double>;
```
yungyuc: `is_floating_point`: https://en.cppreference.com/w/cpp/types/is_floating_point

cpp/modmesh/base.hpp
```diff
@@ -72,6 +72,30 @@ struct is_specialization_of<Template, Template<Args...>> : std::true_type
 template <template <typename...> class Template, typename T>
 inline constexpr bool is_specialization_of_v = is_specialization_of<Template, T>::value;
 
+inline bool is_alpha(char c)
+{
+    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
+}
+
+inline bool is_digit(char c)
```
yungyuc: `isdigit`: https://en.cppreference.com/w/cpp/string/byte/isdigit

cpp/modmesh/base.hpp
yungyuc: `isalpha`: https://en.cppreference.com/w/cpp/string/byte/isalpha

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -27,13 +27,17 @@
  */
 
 #include <modmesh/serialization/SerializableItem.hpp>
```
yungyuc: Thanks for adding the blank line.

cpp/modmesh/serialization/SerializableItem.cpp
yungyuc: It'll be good to append "yet" in the code comment.  Nested arrays will be needed.

I agree we do not need it badly at the moment.

cpp/modmesh/serialization/SerializableItem.cpp
yungyuc: Sweet. I like it. It's the kind of code I cannot quickly write.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -76,6 +80,433 @@ std::string escape_string(std::string_view str_view)
...
+        switch (state)
+        {
+        case JsonState::Start:
+            if (index > 0 || c != '{')
+            {
+                throw_serialization_error(Formatter() << "Invalid JSON format: missing opening bracket.", line, column);
+            }
+
+            state = JsonState::ObjectKey;
+            break; /* end case JsonState::Start */
```
yungyuc, line 323: I like the case ending marks.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+    Array,
+    String,
+    Number,
+    Boolean,
+    Null,
+    Unknown,
+}; /* end enum class JsonType */
+
+struct JsonNode; // Forward declaration
+typedef std::unique_ptr<JsonNode> JsonNodePtr;
```
yungyuc: Do not add type aliases to smart pointers (`JsonNodePtr`).  Use the modern style `using` (for style consistency):

```cpp
using std::unordered_map<std::string, std::unique_ptr<JsonNode>> JsonMap;
using JsonArray = std::vector<std::unique_ptr<JsonNode>>;
```

yungyuc: Aliases to smart pointer types always confuse people.  We will need to look up the definition.  It's better to always spell out the smart pointer type.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+        parse(expression);
     }
-    else if constexpr (std::is_convertible_v<T, std::string>)
+
+private:
+
+    void parse(const std::string & expression)
     {
-        return "\"" + escape_string(value) + "\"";
+        if (type == JsonType::Object)
```
yungyuc, line 106: The function body of `parse` is a little bit too long for declaration, but it's fine for now.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+            const char * separator = "";
+            for (const auto & item : value)
+            {
+                oss << separator << to_json_string(item); /* recursive here */
+                separator = ",";
+            }
+            oss << "]";
+            return oss.str();
+        }
+        else
```
yungyuc: I am not familiar with `constexpr if` enough.  Do you need the keyword here?

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: I don't like the output argument.  But it looks like necessary evil here.  Is it so?  If yes, is it possible to make it private?

yungyuc: And this function `from_json_string` is certainly too long for declaration.  Please make it outside the class declaration.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -55,65 +56,265 @@ namespace detail
...
+        {
+            throw std::runtime_error("Invalid JSON format: invalid type.");
         }
-        oss << "]";
-        return oss.str();
     }
-    else
+
+    template <typename T>
+    static void from_json_string(const JsonNodePtr & node, std::vector<T> & vec)
```
yungyuc: This overload can be outside the declaration too.

gtests/test_nopython_serializable.cpp
```diff
@@ -163,4 +167,111 @@ TEST(Json, serialization_with_object)
                          "\"pets\":[{\"name\":\"Fluffy\",\"age\":3,\"is_dog\":true,\"is_cat\":false},{\"name\":\"Whiskers\",\"age\":8,\"is_dog\":false,\"is_cat\":true}]}";
 }
 
+TEST(Json, deserialize_private_member_partial_exposure)
```
yungyuc, line 170: We will need Python tests for these things.  It's not productive to test for strings in C++.

It can use a new PR or even a new issue.

yungyuc (changes requested):
It's an exciting PR.  Things to fix:

- [x] Use STL `isalpha`, `isdigit`, `is_integral`, and `is_floating_point`.
- [x] Do not alias smart pointer types.
- [x] Use the modern `using` syntax.
- [x] Clarify how to improve the output argument of `from_json_string()`.
- [x] `from_json_string()` is too long for class declaration.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: `constexpr if` is doing the condition in compilation time, which will not have a runtime branch.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: > And this function from_json_string is certainly too long for declaration. Please make it outside the class declaration.

I put it in the class because L228 (`value.from_json(std::get<JsonMap>(node->value))`) uses the private method of the class that inherits `SerializableItem`, and that's why I declare `friend class detail::JsonHelper;`

tigercosmos: > I don't like the output argument. But it looks like necessary evil here. Is it so? If yes, is it possible to make it private?

based on the current design, it is necessary.
What do you mean "private"? this is a helper class, and the function is static.
`SerializableItem::from_json` calls this function, and feeds the class member throw the output argument.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: My question is against `else`.  Does it need `constexpr`?

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: In addition, since it's a helper class (or basically a helper function), I think the length should be fine? we have much longer helper class in `array_common.hpp`

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: therefore, I think it's necessary here.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: > > And this function from_json_string is certainly too long for declaration. Please make it outside the class declaration.
> 
> I put it in the class because L228 (when the parser doing recursion) uses the private method of `SerializableItem`, and that's why I declare `friend class detail::JsonHelper;`

The access is OK.  Long functions are not easy to read when putting in class declaration.  They should be taken outside the declaration while still made inline.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: this function is okay to put outside, since it doesn't use the private function. However, as `static void from_json_string(const detail::JsonNodePtr & node, T & value)` must be put in the helper class, I think we should put this function in the helper class for the consistency?

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: > > I don't like the output argument. But it looks like necessary evil here. Is it so? If yes, is it possible to make it private?
> 
> based on the current design, it is necessary. What do you mean "private"? this is a helper class, and the function is static. `SerializableItem::from_json` calls this function, and feeds the class member throw the output argument.

It will be easier to maintain a function of output argument when it is private.  It does not matter the function to be static or not.  But if you use the class as a namespace for collecting static member functions and call them from other class, making them private is not realistic.

Please leave a code comment to warn maintainers of the output argument.  We should get rid of it in the future.

gtests/test_nopython_serializable.cpp
tigercosmos: sure, should be in the future PRs.

cpp/modmesh/serialization/SerializableItem.cpp
tigercosmos: fixed.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: By putting outside the (class) declaration, I meant to move the function body, like:

```cpp
class C
{
    void F();
};
void C::F()
{
    // do something
}
```

It has nothing to do with function access.

cpp/modmesh/serialization/SerializableItem.cpp
tigercosmos: sorry, fixed.

tigercosmos: I hope clang-format can help for this.

cpp/modmesh/base.hpp
tigercosmos: thanks for telling me

cpp/modmesh/base.hpp
tigercosmos: note that `is_integral` doesn't not support `long long it`, and the support is until C++20 `std::integral`, but I think should be fine here.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: however,  the function is a template function, I don't think we can do it?

tigercosmos: I see, I realize it's possible to do it even for templates.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: > My question is against else. Does it need constexpr?

No need.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: fixed

tigercosmos:
@yungyuc 
> Use STL isalpha, isdigit, is_integral, and is_floating_point.

all fixed.

>  Do not alias smart pointer types.

fixed.

> Use the modern using syntax.

fixed.

> Clarify how to improve the output argument of from_json_string().

output argument is needed by design.

> from_json_string() is too long for class declaration.

splitted.

All issues are addressed, please also check the comments.

cpp/modmesh/serialization/SerializableItem.cpp
yungyuc: This is good. Thanks.

cpp/modmesh/serialization/SerializableItem.cpp
yungyuc: If it does not, we can consider to craft a linter for ourselves.  In the future, of course.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -89,31 +174,160 @@ std::string to_json_string(const T & value)
     }
 }
 
-}; /* end namespace detail */
+template <typename T>
+void JsonHelper::from_json_string(const std::unique_ptr<JsonNode> & node, T & value)
```
yungyuc, line 180: Need code comment for the plan to mitigate the output arguments.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -89,31 +174,160 @@ std::string to_json_string(const T & value)
...
+        value.from_json(std::get<JsonMap>(node->value)); /* recursive here */
+    }
+    else
+    {
+        throw std::runtime_error("Invalid JSON format: invalid type.");
+    }
+}
+
+template <typename T>
+void JsonHelper::from_json_string(const std::unique_ptr<JsonNode> & node, std::vector<T> & vec)
```
yungyuc, line 252: Ditto.  Output argument.

yungyuc (changes requested):
- [x] Leave code comments for a plan to mitigate the output argument.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: (I hit review changes button too quickly.)  We can discuss here or discord for how to author the plan.  It may be a simple one.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: @yungyuc  based on the current design:
in `from_json_string`:
```cpp
    else if constexpr (std::is_base_of_v<SerializableItem, T>)
    {
        if (node->type != detail::JsonType::Object)
        {
            throw std::runtime_error("Invalid JSON format: invalid object type.");
        }
        // NOLINTNEXTLINE(misc-no-recursion)
        value.from_json(std::get<JsonMap>(node->value)); /* recursive here */
    }
 ```
 
Note that we call `value`'s mutable method `from_json`. Therefore, we must pass the **mutable** reference of `value` as the argument of `from_json_string`.
 
 I don't think it can be an easy change to get rid of the output argument unless we change the whole design.
 
 Also,  I didn't fully understand why you wanted to remove the output argument. This function is live in the `detail` namespace, and the implementation is hidden under the macro, I don't think the user has chance to use this function.

**Comment by tigercosmos** [comment]
@yungyuc  please see my comment here: https://github.com/solvcon/modmesh/pull/377#discussion_r1653774032

File: cpp/modmesh/serialization/SerializableItem.hpp (hunk 14 above)
Comment (yungyuc) [reply cpp/modmesh/serialization/SerializableItem.hpp]: It's a good point that the static member function is in `detail`.  I missed it.  In this case I agree the output argument is not so severe.  Could you please add a simple code comment indicating the output arguments are bad for maintenance and should be fixed in the future?  (A concrete plan does not need to be mentioned.)

It's quite possible for a maintainer to move a useful helper class outside `detail`, in the future.  Then it will become an issue.

**Comment by tigercosmos** [comment]
@yungyuc  comments for the output argument is added.

File: cpp/modmesh/serialization/SerializableItem.hpp [hunk 29]
```diff
@@ -55,11 +56,94 @@ namespace detail
...
+
+/// Helper class for JSON serialization and deserialization.
+class JsonHelper
+{
+public:
+    template <typename T>
+    static std::string to_json_string(const T & value);
+
+    template <typename T>
+    static void from_json_string(const std::unique_ptr<JsonNode> & node, T & value); // TODO: have a design that can remove the output argument to increase the maintainability
```
yungyuc: Thanks, but please beautify the comments.  Appending after the signature makes the lines too line.  Please add the comments before the signature in distinct lines.

yungyuc (changes requested):
Thanks, @tigercosmos .  We need some cosmetics.

- [x] Cosmetic changes for the comments.

yungyuc:
I saw the change and will merge once CI completes.

yungyuc (approved):

PR #385, opened 2024-07-07
Camera: use WASD+EQ and arrows+Page Up/Down for movement

### LINKED ISSUE #70: Improve camera control

Tucchhaa:
Now instead of arrows and Page Up/Down, WASD+EQ can be used.

It was impossible to just reassign keys for camera controllers, so we had to re-implement some parts. Such as, moveCamera and input listeners. 

Original code of input listeners here: [link](https://github.com/qt/qt3d/blob/7d57b46f381671215ee323f4f1d04a01deafa71c/src/extras/defaults/qabstractcameracontroller.cpp)
Original code of first person camera movement: [link](https://github.com/qt/qt3d/blob/7d57b46f381671215ee323f4f1d04a01deafa71c/src/extras/defaults/qfirstpersoncameracontroller.cpp#L61-L79)
Original code of orbit camera movement: [link](https://github.com/qt/qt3d/blob/7d57b46f381671215ee323f4f1d04a01deafa71c/src/extras/defaults/qorbitcameracontroller.cpp#L213-L258)


cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include <modmesh/view/common_detail.hpp> // Must be the first include.
```
yungyuc, line 31: It's good to keep this.  It is a must for Python.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include <modmesh/view/RCameraController.hpp>
+
+#include <Qt3DInput/QKeyboardDevice>
+#include <Qt3DInput/QMouseDevice>
+#include <Qt3DRender/QCamera>
+
+namespace modmesh {
+
+/*
+ * CameraInputListener
```
yungyuc: Do not add obvious code remarks.  They are redundant and hurt maintainability.

yungyuc (changes requested):
Thanks, @Tucchhaa .  Before we start review, please make all CI runs clean.  You can turn it on in your fork.  (Github does not run Actions for a PR whose creator has not got a PR merged.)

When you have everything ready, please do explicitly add a global comment.

Other items that are expected in your next request for review:

- [x] Add copyright notice in all files.  We use BSD license.  See [`R3DWidget.hpp`](https://github.com/solvcon/modmesh/blob/ba1e4cb16b6281d2991b98d34736284b892f718a/cpp/modmesh/view/R3DWidget.hpp)
- [x] Add inline annotations (in the PR) to highlight the key additions/changes.
- [x] Make sure you write all the code in your PR.  You may reference an example, but may not copy its code.
- [x] Review your code comments.  Add useful ones and remove redundant ones.

cpp/modmesh/view/RCameraController.cpp
Tucchhaa: I added them because there're several classes in the file, and it's easier to distinguish between them if there're these comments. But if you feel that they're redundant ok, I'm gonna remove them

Tucchhaa:
> * [ ]  Make sure you write all the code in your PR.  You may reference an example, but may not copy its code.


Well, there's a lot of code copied from Qt repository, what should I do with that?

Like [RFirstPersonCameraController.updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R199-R221) and [ROrbitCameraController.updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R253-R312)

yungyuc:
> > * [ ]   Make sure you write all the code in your PR.  You may reference an example, but may not copy its code.
> 
> Well, there's a lot of code copied from Qt repository, what should I do with that?
> 
> Like [RFirstPersonCameraController.updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R199-R221) and [ROrbitCameraController.updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R253-R312)

We may not copy from Qt which is a commercial code with an optional GPL license.  So you should fully understand what it does and code it up yourself.

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -39,27 +39,13 @@
...
-{
-    using Qt3DExtras::QFirstPersonCameraController::QFirstPersonCameraController;
-}; /* end class RFirstPersonCameraController */
-
-class ROrbitCameraController
-    : public Qt3DExtras::QOrbitCameraController
-{
-    using Qt3DExtras::QOrbitCameraController::QOrbitCameraController;
-}; /* end class ROrbitCameraController */
-
```
Tucchhaa, line 62: Key difference in this PR, is that we don't use Qt original camera controller function ([moveCamera](https://github.com/solvcon/modmesh/pull/385/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR134)), but instead implement our own controller function ([updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR136)).

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,343 @@
...
+    m_keyboardTzNegInput->setButtons(QList<int> { Qt::Key_S });
+    m_keyboardTzNegInput->setScale(-1.0f);
+    m_keyboardTzNegInput->setSourceDevice(m_keyboardDevice);
+    m_tzAxis->addInput(m_keyboardTzNegInput);
+
+    // keyboard negative y translation
+    m_keyboardTyNegInput->setButtons(QList<int> { Qt::Key_Q });
+    m_keyboardTyNegInput->setScale(-1.0f);
+    m_keyboardTyNegInput->setSourceDevice(m_keyboardDevice);
+    m_tyAxis->addInput(m_keyboardTyNegInput);
```
Tucchhaa: arrows were replaced for WASD here

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,343 @@
...
+        state.tzAxisValue = m_tzAxis->value();
+
+        state.leftMouseButtonActive = m_leftMouseButtonAction->isActive();
+        state.middleMouseButtonActive = m_middleMouseButtonAction->isActive();
+        state.rightMouseButtonActive = m_rightMouseButtonAction->isActive();
+
+        state.altKeyActive = m_altButtonAction->isActive();
+        state.shiftKeyActive = m_shiftButtonAction->isActive();
+
+        m_callback(state, dt);
```
Tucchhaa: This is called every frame and it reads the input, and calls m_callback. m_callback is actually [updateCameraPositions](https://github.com/solvcon/modmesh/pull/385/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR136) from camera controllers (orbit, and first person)

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,161 @@
...
+class CameraControllerMixin
+{
+public:
+    virtual ~CameraControllerMixin() = default;
+
+    virtual void updateCameraPosition(const Qt3DExtras::QAbstractCameraController::InputState &state, float dt) = 0;
+
+protected:
+    RCameraInputListener* m_listener = nullptr;
+};
```
Tucchhaa: RFirstPersonCameraController and ROrbitCameraController extend this mixin, they define m_listener in their constructors

Tucchhaa:
@yungyuc applied all your review notes, you can check PR again now. Btw, how can I enable CI in my fork?

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,343 @@
...
+    m_shiftButtonAction->addInput(m_shiftButtonInput);
+
+    // alt button
+    m_altButtonInput->setButtons(QList<int> { Qt::Key_Alt });
+    m_altButtonInput->setSourceDevice(m_keyboardDevice);
+    m_altButtonAction->addInput(m_altButtonInput);
+
+    //// Axes
+
+    // mouse x rotation
```
yungyuc: Could you share why mouse has only x/y rotation without z, and why only x/z translation without y?

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,343 @@
...
+    m_mouseTzXInput->setAxis(Qt3DInput::QMouseDevice::WheelX);
+    m_mouseTzXInput->setSourceDevice(m_mouseDevice);
+    m_tzAxis->addInput(m_mouseTzXInput);
+
+    // mouse z translation
+    m_mouseTzYInput->setAxis(Qt3DInput::QMouseDevice::WheelY);
+    m_mouseTzYInput->setSourceDevice(m_mouseDevice);
+    m_tzAxis->addInput(m_mouseTzYInput);
+
+    // keyboard positive x translation
```
yungyuc: Could you share why the code of keyboard translation is in the order of x, z, y?

yungyuc:
> how can I enable CI in my fork?

Please go to the corresponding section (for actions) in the settings tab in your github repository.

cpp/modmesh/view/RCameraController.cpp
yungyuc: Keyboard arrow keys are useful in moving cameras (translation) too.  Could you also add them?

(When we mentioned "WSAD" in #70, we also implied the arrow keys.  But it is not explicitly mentioned in the issue.)

Later, after the basic control feature is in place, we will need to add configurations for users to select how they want to map the control keys.  But in this PR we focus on the basic feature.  Two reasons: (i) simplify the code to make future maintenance easier, and (ii) keep PR short enough for review.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,343 @@
...
+        state.altKeyActive = m_altButtonAction->isActive();
+        state.shiftKeyActive = m_shiftButtonAction->isActive();
+
+        m_callback(state, dt);
+    });
+}
+
+void RCameraInputListener::init()
+{
+    // left mouse button
```
yungyuc: Breaking the setup code for mouse and keyboard in two helper functions could make the logic read more modular.  The other setup code can remain in this function.  The modularization could help a future developer to add code on top of it.  What do you think about it?

cpp/modmesh/view/RCameraController.hpp
yungyuc: Let's not resolve a conversation in PR code review.  The resolution makes reviewing harder.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,161 @@
...
+namespace modmesh {
+
+class RCameraInputListener : public Qt3DCore::QEntity
+{
+    Q_OBJECT
+
+public:
+    using callback_type = std::function<void (const Qt3DExtras::QAbstractCameraController::InputState&, float)>;
+
+    RCameraInputListener(Qt3DInput::QKeyboardDevice* keyboardDevice, Qt3DInput::QMouseDevice* mouseDevice,
```
yungyuc: The indentation does not look right and I am not sure clang-format can detect it.  Please align the indentation similar to PEP-8, like:

```cpp
    RCameraInputListener(
        Qt3DInput::QKeyboardDevice* keyboardDevice, Qt3DInput::QMouseDevice* mouseDevice,
        callback_type callback, QNode *parent = nullptr);
```

yungyuc (changes requested):
Something needs to be added:

- [x] Need Python wrapper for changing the camera controller and testing it from Python.
- [x] Try to create Python unit tests for the code.  If it is too hard to test, show why it is.

Something needs to be changed or discussed:

- [x] Use both "WSAD" and arrow keys.
- [x] Discussion: should we break the `init()` function into smaller helper functions?
- [x] Code style formatting.
- [x] Question: mouse rotation and translation.
- [x] Question: keyboard translation in the order of x, z, y?


----

Note on PR size.  A rule of thumb is for the code diff in a PR is 300-500 lines.  It helps us efficiently review and discuss.  If a batch of development takes more than 500 lines of diff, we do our best to break it into smaller pieces.

cpp/modmesh/view/RCameraController.cpp
Tucchhaa: mouse x/y rotation is listening for mouse's movement, which can be only in two dimensions.

mouse x/z translation is listening for mouse's wheels. Afaik, there are mouses with trackballs, which can rotate in two dimensions. 

Actually, we can remove mouse x translation, since such mouses aren't common. I decided to keep it, because qt had it, and I wanted as little changes to camera's behavior as possible.

cpp/modmesh/view/RCameraController.cpp
Tucchhaa: There's no specific reason 🙂, I will change the order to xyz

cpp/modmesh/view/RCameraController.cpp
Tucchhaa: I think that's a good idea. I will refactor init() into smaller parts, and will add a loop for assigning keys to remove duplication of code, if it'll be more easy to read

cpp/modmesh/view/RCameraController.hpp
Tucchhaa: If I move every parameter to a separated line, do you think it'll be good?

Tucchhaa:
> Note on PR size. A rule of thumb is for the code diff in a PR is 300-500 lines. It helps us efficiently review and discuss. If a batch of development takes more than 500 lines of diff, we do our best to break it into smaller pieces

It'll be hard to separate this PR, since every class depends on each other. And if PR just RCameraInputListener, then it'll not make a lot sense imho. Moreover, a big part of the PR is assigning to input which is duplicated parts of code

cpp/modmesh/view/RCameraController.cpp
Tucchhaa: Yeah, of course, I will add arrows and Page Up/Down

cpp/modmesh/view/RCameraController.cpp
yungyuc: The repeating (duplicated) code for setting value is fine.  I prefer to keep them.  It is not overly long, especially after broken into sub-functions, and makes debugging easier.

cpp/modmesh/view/RCameraController.cpp
yungyuc: This is worth mentioning in the code remark.  I agree keeping support for trackballs is good.  Professional CAD may even use more moving modes for pointing devices.

cpp/modmesh/view/RCameraController.cpp
yungyuc: I see.  I agree that if there is not a specific reason for xzy, changing to xyz order will make the code more straight-forward.

cpp/modmesh/view/RCameraController.hpp
yungyuc: For this specific function, which does not have a long list of arguments, it does not matter.

Style is subjective.  I will exercise my preference to how the code should look like and make the linter force it as much as possible.  But everything can be discussed.

yungyuc (commented):
@Tucchhaa When ready for the next code review, leave a global comment like what you did.  The global comment helps us keep track of the review progress.

----

> > Note on PR size. A rule of thumb is for the code diff in a PR is 300-500 lines. It helps us efficiently review and discuss. If a batch of development takes more than 500 lines of diff, we do our best to break it into smaller pieces
> 
> It'll be hard to separate this PR, since every class depends on each other. And if PR just RCameraInputListener, then it'll not make a lot sense imho. Moreover, a big part of the PR is assigning to input which is duplicated parts of code

This one is OK.  300-500 is not a strict number.  But if you have a PR consisting of 1,300 lines of diff, there is certainly something wrong.

yungyuc:
> > Note on PR size. A rule of thumb is for the code diff in a PR is 300-500 lines. It helps us efficiently review and discuss. If a batch of development takes more than 500 lines of diff, we do our best to break it into smaller pieces
> 
> It'll be hard to separate this PR, since every class depends on each other. And if PR just RCameraInputListener, then it'll not make a lot sense imho. Moreover, a big part of the PR is assigning to input which is duplicated parts of code

This one is OK.  300-500 is not a strict number.  But if you have a PR consisting of 1,300 lines of diff, there is certainly something wrong.

Tucchhaa:
> [ ] Need Python wrapper for changing the camera controller and testing it from Python.

May I ask a question. Why do we need a python wrapper for the camera controller? Is it only for tests? @yungyuc

yungyuc:
> > [ ] Need Python wrapper for changing the camera controller and testing it from Python.
> 
> May I ask a question. Why do we need a python wrapper for the camera controller? Is it only for tests? @yungyuc

I like the question.  It will work for testing in Python but much more than that.  Everything in GUI should be scriptable.

Tucchhaa:
> > > [ ] Need Python wrapper for changing the camera controller and testing it from Python.
> > 
> > 
> > May I ask a question. Why do we need a python wrapper for the camera controller? Is it only for tests? @yungyuc
> 
> I like the question. It will work for testing in Python but much more than that. Everything in GUI should be scriptable.

Actually, I'm not sure if we need python wrapper for this case. All functions are private. Except [CameraControllerMixin::updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR124), but it has parameter dt (delta time), so calling this function makes sense only after every frame (which is done is [cpp](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R221-R228))

And for tests, with that function exposed we actually can test how camera position and rotation are changed. But then we'll also need to expose camera's props (like position, rotation, maybe others)

So rn I see two options.
1) we don't wrap the function, since we don't have practical usage of it in python
2) we wrap this function only for python tests, and wrap some camera's props too. But we do it in another PR, since there are going to be a lot of changes: python wrappers + a lot of tests (translation, rotation, some key combinations for fps and orbit camera types)

What do you think @yungyuc

yungyuc:
> Actually, I'm not sure if we need python wrapper for this case. All functions are private. Except [CameraControllerMixin::updateCameraPosition](https://github.com/solvcon/modmesh/pull/385/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR124), but it has parameter dt (delta time), so calling this function makes sense only after every frame (which is done is [cpp](https://github.com/solvcon/modmesh/pull/385/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R221-R228))

We can start from it.  We will expose as much as possible.  PySide6 will serve as a guideline for what functions and data to expose.

> And for tests, with that function exposed we actually can test how camera position and rotation are changed. But then we'll also need to expose camera's props (like position, rotation, maybe others)

Yes, we should expose properties too.

> So rn I see two options.
> 
> 1. we don't wrap the function, since we don't have practical usage of it in python

Controlling cameras from Python will be a must-have.  You can take a look at https://kitware.github.io/paraview-docs/v5.9.0/python/paraview.python_view.html and https://docs.vtk.org/en/latest/api/python.html .  I hope solvcon will work more convenient, in the long run.  But it's definitely not easy work.

> 2. we wrap this function only for python tests, and wrap some camera's props too. But we do it in another PR, since there are going to be a lot of changes: python wrappers + a lot of tests (translation, rotation, some key combinations for fps and orbit camera types)

There should be basic Python wrapper for testing in this PR.  Tests almost always go with fix/feature checkins.  Additional functions and data exposure can be made with later PRs or issues.

Tucchhaa:
@yungyuc ready for review

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -440,6 +440,93 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRManager
...
+                py::arg("dt"))
+            .def(
+                "position",
+                [](wrapped_type & self)
+                {
+                    const auto position = self.position();
+                    return py::make_tuple(position.x(), position.y(), position.z());
+                })
+            .def(
+                "linear_speed", [](wrapped_base_type & self)
```
yungyuc: Please move the lambda to the next line like in line 502 to help clang-format to make the code look nicer.

```cpp
"linear_speed",
[](wrapped_base_type & self)
{ return self.getLinearSpeed(); })
```

tests/test_view_camera.py
```diff
@@ -0,0 +1,149 @@
+import unittest
```
yungyuc: Please add copyright notice.  You may copy it from the files in the same directory.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+RCameraInputListener::RCameraInputListener(
+    Qt3DInput::QKeyboardDevice * keyboardDevice,
+    Qt3DInput::QMouseDevice * mouseDevice,
+    callback_type callback,
+    QNode * parent)
+    : QEntity(parent)
+    , m_frameAction(new Qt3DLogic::QFrameAction)
+    , m_logicalDevice(new Qt3DInput::QLogicalDevice)
+
```
yungyuc: Please do not add the blank lines.  They do not make the code clearer.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,347 @@
...
+    , m_keyboardTxPosInput(new Qt3DInput::QButtonAxisInput)
+    , m_keyboardTyPosInput(new Qt3DInput::QButtonAxisInput)
+    , m_keyboardTzPosInput(new Qt3DInput::QButtonAxisInput)
+    , m_keyboardTxNegInput(new Qt3DInput::QButtonAxisInput)
+    , m_keyboardTyNegInput(new Qt3DInput::QButtonAxisInput)
+    , m_keyboardTzNegInput(new Qt3DInput::QButtonAxisInput)
+{
+    init();
+
+    connect(m_frameAction, &Qt3DLogic::QFrameAction::triggered, this, [this](const float dt)
```
yungyuc: Move the lambda to a new line so that it indents with the following open curly brace:

```cpp
    connect(m_frameAction, &Qt3DLogic::QFrameAction::triggered, this,
            [this](const float dt)
            {
                ...
            });
```

This is a limitation of clang-format.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,347 @@
...
+        state.tzAxisValue = m_tzAxis->value();
+
+        state.leftMouseButtonActive = m_leftMouseButtonAction->isActive();
+        state.middleMouseButtonActive = m_middleMouseButtonAction->isActive();
+        state.rightMouseButtonActive = m_rightMouseButtonAction->isActive();
+
+        state.altKeyActive = m_altButtonAction->isActive();
+        state.shiftKeyActive = m_shiftButtonAction->isActive();
+
+        m_callback(state, dt); });
```
yungyuc: The closing curly brace should use a new line from the last line of code.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,347 @@
...
+    };
+
+    m_listener = new RCameraInputListener(keyboardDevice(), mouseDevice(), callback, this);
+}
+
+void RFirstPersonCameraController::updateCameraPosition(const InputState & input, const float dt)
+{
+    constexpr auto upVector = QVector3D(0.f, 1.f, 0.f);
+
+    Qt3DRender::QCamera * _camera = camera();
```
yungyuc: Never name any variable with leading underscore in solvcon/modmesh.  See https://stackoverflow.com/a/228797 .  The style in solvcon/modmesh is more strict than required.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    RCameraInputListener(
+        Qt3DInput::QKeyboardDevice * keyboardDevice,
+        Qt3DInput::QMouseDevice * mouseDevice,
+        callback_type callback,
+        QNode * parent = nullptr);
+
+private:
+
+    Qt3DLogic::QFrameAction * m_frameAction;
```
yungyuc: Member data naming convention is `m_snake_case`.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    // keyboard translation input
+    Qt3DInput::QButtonAxisInput * m_keyboardTxPosInput;
+    Qt3DInput::QButtonAxisInput * m_keyboardTyPosInput;
+    Qt3DInput::QButtonAxisInput * m_keyboardTzPosInput;
+    Qt3DInput::QButtonAxisInput * m_keyboardTxNegInput;
+    Qt3DInput::QButtonAxisInput * m_keyboardTyNegInput;
+    Qt3DInput::QButtonAxisInput * m_keyboardTzNegInput;
+
+    void init();
```
yungyuc: I prefer member functions to be placed before member data.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    void initMouseListeners() const;
+
+    void initKeyboardListeners() const;
+};
+
+class CameraController
+{
+
+protected:
```
yungyuc: Please order members in the order of public, protected, private, unless there is a logical reason not to.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -0,0 +1,194 @@
...
+    void zoom(float zoomValue) const;
+
+    void orbit(float pan, float tilt) const;
+
+    static float clamp(float value);
+
+    static float zoomDistanceSquared(QVector3D firstPoint, QVector3D secondPoint);
+};
+
+} // namespace modmesh
```
yungyuc: Use "end namespace xxx" for the namespace ending mark.

Add the vim modeline like other files in this directory.  Even if a maintainer does not use vim, it makes the file format clear.

tests/test_view_camera.py
```diff
@@ -0,0 +1,149 @@
...
+        self.controller.updateCameraPosition(pitch=rotation, dt=dt, right_mouse_button=True)
+
+        rotation_matrix = angle_axis(rotation * dt * look_speed, (1, 0, 0))
+        rotated_vector = np.array(rotated_vector) @ rotation_matrix
+        view_vector = self.controller.view_vector()
+
+        self.assertAlmostEqual(rotated_vector[0], view_vector[0], delta=1e-2)
+        self.assertAlmostEqual(rotated_vector[1], view_vector[1], delta=1e-2)
+        self.assertAlmostEqual(rotated_vector[2], view_vector[2], delta=1e-2)
+
```
yungyuc: Use the vim modeline like other files in this directory.

yungyuc (changes requested):
It's good.  Now the review can start to consider the quality of code.  Before requesting for the next review, please make sure CI is clean.  Please pay attention to linter.

- [x] Format lambda code with proper indentation.
- [x] Copyright notice in test code.
- [ ] Remove unnecessary blank lines.
- [x] Never name any variable with leading underscore in solvcon/modmesh.
- [x] Member data naming convention is `m_snake_case`.
- [x] Use the vim modelines.
- [x] Use "end namespace xxx" for the namespace ending mark.

Tucchhaa:
@yyc CI should be green. I skipped camera tests for now, since we don't have GUI in our CI.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -440,6 +439,95 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRManager
...
+        : root_base_type(mod, pyname, pydoc)
+    {
+        namespace py = pybind11;
+
+        (*this)
+            .def(
+                "updateCameraPosition",
+                [](
+                    wrapped_type & self,
+                    float const & x,
```
yungyuc: Please use (non-const) `float` for arguments of fundamental types instead of `const &`.  A `float` takes 4 bytes and a `float const &` or `float const *` takes 8 bytes (pointer).

The generated assembly might be the same between `float` and `float const &` but the semantics differ.  Arguments of a const reference type cannot be reused in the callee.  But the most significant issue is the conflict with what we expect the compiler to do.

yungyuc: It is unfortunate that clang-format forces you to use a single line per argument for this wrapping lambda.  I hope there is an option to ask clang-format to format like what I want:

```cpp
[](wrapped_type & self, float x, float y, float z, float pitch, float yaw,
   bool left_mouse_button, bool right_mouse_button,
   bool alt_key, bool shift_key, float dt)
```

But this is future work.  We'll follow what clang-format does for us even if it is not perfect.  It's more important to have a consistent style.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -440,6 +439,95 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRManager
...
+
+                    input.leftMouseButtonActive = left_mouse_button;
+                    input.rightMouseButtonActive = right_mouse_button;
+
+                    input.altKeyActive = alt_key;
+                    input.shiftKeyActive = shift_key;
+
+                    self.updateCameraPosition(input, dt);
+                },
+
```
yungyuc, line 488: Please remove all unnecessary blank lines.

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -0,0 +1,341 @@
...
+    , m_keyboard_ty_neg_input(new Qt3DInput::QButtonAxisInput)
+    , m_keyboard_tz_neg_input(new Qt3DInput::QButtonAxisInput)
+{
+    init();
+
+    connect(
+        m_frame_action,
+        &Qt3DLogic::QFrameAction::triggered,
+        this,
+        [this](const float dt)
```
yungyuc, line 81: You don't really need the `const` for the argument of fundamental types.  It usually does more harm than good.  But it's fine at this point.

tests/test_view.py
```diff
@@ -59,4 +60,156 @@ def test_pycon(self):
         self.assertFalse(view.mgr.pycon.python_redirect)
 
 
+def angle_axis(angle_deg, axis):
```
yungyuc: Make a base class for the test case classes `ViewFPSCameraTC` and `ViewOrbitCameraTC` for holding `angle_axis()`.  An example can be found in https://github.com/solvcon/modmesh/blob/975f7c5730247f8e4ce770634ecb692fc3a70c0e/tests/test_universe.py#L35

tests/test_view.py
```diff
@@ -59,4 +60,156 @@ def test_pycon(self):
...
+        [a[2], 0, -a[0]],
+        [-a[1], a[0], 0]
+    ]) * sin_angle
+
+    rotation += m + n
+
+    return rotation
+
+
+@unittest.skip("GUI is not yet available for testing")
```
yungyuc: Only skip tests when GUI is not build.  Test for `modmesh.HAS_VIEW`.  See https://github.com/solvcon/modmesh/blob/d92add6ee4f07caca20b52c943dd57b672e89a5a/tests/test_view.py#L45

tests/test_view.py
```diff
@@ -59,4 +60,156 @@ def test_pycon(self):
...
+class ViewFPSCameraTC(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
yungyuc, line 70: To confirm: do you need corresponding teardown?

tests/test_view.py
```diff
@@ -59,4 +60,156 @@ def test_pycon(self):
...
+
+    def test_translation(self):
+        dt = 0.01
+        linear_speed = self.controller.linear_speed()
+
+        delta = dt * linear_speed
+        new_pos = np.array(self.get_pos()) + [delta, delta, -delta]
+
+        self.move(x=1, dt=dt)
+        self.assertAlmostEqual(self.get_pos()[0], new_pos[0], delta=1e-2)
```
yungyuc: Shouldn't the position be exact?

yungyuc (changes requested):
> @yyc CI should be green. I skipped camera tests for now, since we don't have GUI in our CI.

Yes, the CI is green.  Thanks.  The camera tests should be turned on with the builds with Qt.

- [x] Don't use `const &` for arguments of fundamental types.
- [x] Review and remove all unnecessary blank lines.
- [x] Use a base class for holding the test helper `angle_axis()`.
- [x] Only skip tests when GUI is not build.

----

Discussions:
- [x] Do we need teardown for the camera tests?
- [x] Clarify whether the position in camera can be made exact for testing.

tests/test_view.py
Tucchhaa: I've tried to do that one and I got an error: https://github.com/Tucchhaa/modmesh/actions/runs/10027785212/job/27713902533#step:14:2013

Seems that renderer can't be used in CI :/ At this point I don't know how to solve this problem, ahhh I hate testing so much 😣, but I will try to figure out what we can do with that

yungyuc: Does it work with `make run_viewer_pytest`?

----

Nonetheless, we need a way to turn on the test without having to edit the source code in the test file.  If `modmesh.HAS_VIEW` does not work, we need another way for `skipUnless`.

Tucchhaa: Yes, `make run_viewer_pytest` works on my machine, but looks like it wouldn't work on CI

tests/test_view.py
Tucchhaa: Thank you for pointing to this. I'm gonna add it

tests/test_view.py
Tucchhaa: I wish it was possible, but there's always a little error in calculations, because qt takes cross products and magnitude of vectors in [qcamera.translate](https://github.com/qt/qt3d/blob/82b59c89e282dcc04a01727e355d284c128752ab/src/render/frontend/qcamera.cpp). But we don't need very precise value here, delta=1e-2 is pretty enough

tests/test_view.py
yungyuc: Perhaps you can use [`GITHUB_ACTIONS`](https://docs.github.com/en/actions/learn-github-actions/variables) environment variable to turn it off temporarily.

tests/test_view.py
yungyuc: A smaller tolerance close to the round-off done by the helper should be used.  1 percent does not look like a small tolerance.  If possible, please use the smallest possible tolerance.  Alternately, please elaborate why 1 percent is sufficiently small.

yungyuc: Exact equality will prevent a lot of maintenance issues.  Inexact comparison is merely tolerated.

tests/test_view.py
Tucchhaa: you mean to turn off the camera tests?

tests/test_view.py
Tucchhaa: I agree that 1 percent doesn't look like a small tolerance, but it's actually enough for what we are testing here. If we decrease tolerance we will get unstable tests, which is much worse. 

I know that exact equality is always better choice than approximate equality. And I've spent a lot of time to do that, but couldn't get the desired result. And only after looking to source code of camera translation, I realized that it should be due to this [line](https://github.com/qt/qt3d/blob/82b59c89e282dcc04a01727e355d284c128752ab/src/render/frontend/qcamera.cpp#L607) and possibly we can't do much here or it will take a lot of time.

To prove my point here's the log data that I got:

Before translating camera towards X+ coordinate:
camera position: (0, 0, 10)
view center: (0, 0, 0)
view vector: (0, 0, -10)

After translate by (0.5, 0, 0):
camera position: (0.5, 0.000123389, 10)

tests/test_view.py
yungyuc: Yes.  Instead of always turning off the new camera tests, only turn them off when they are on GitHub Actions.

The environment-dependent behavior causes inconsistency and is hard to debug in the future. But it is better than having no test coverage for the new code.

Please leave a code comment stating a plan of improvement.  We will create an issue tracking it after merging the PR.

tests/test_view.py
yungyuc: We need to find a balance between stability and sensitivity.  1 percent is so large that the stability may be false.  Graphical code uses single precision and has 1e-6 precision.  1 percent is 4 orders of magnitude larger than that.

I need more arguments to be convinced to use 1 percent for the tolerance. For example, you did not make it clear why 1 percent is used as the tolerance while the error in your test is 1.23e-4.

Developing the arguments will help us clarify what is a proper tolerance.

Tucchhaa: We can decrease tolerance to 1e-3 then. What do you think?

Tucchhaa: We move camera by the value of 0.5, so tolerance of 0.01 will always detect if camera did/didn't move. Of course 0.5 can be changed in the future to a bigger or smaller value, but I am pretty sure that we will never need to test if translation work on such small values as 0.01.

But yeah the smaller the better. We can decrease to 1e-3. And if we want to make it even more precise, it will take some time, that I think won't really benefit the project

yungyuc: When saying something like "I am pretty sure ...", we usually mean that "from our experience, it must ..."  But based on what we have discussed so far, we probably do not have sound experience to serve as a background reference to say how a camera or Qt should behave.  But please let me know if you have a hidden argument with that sentence.

When we do not have sufficient experience to make decision, the best too is our reasoning faculty.  We need good (logical) arguments to exercise that.  I know the round-off of single precision floating point is 1e-6 and 1.e-2 and 1.e-4 are 2 to 4 orders of magnitude larger than that.  If the error in Qt is resulted from round-off, it should not be as big as 1.e-4.

I hope to learn something I didn't know from your arguments.  And based on the arguments we decide what to do next.  (I only need arguments.  I am fine for you to do more experiments, but that is not my point.)

Tucchhaa: So I found the reason why that's happening. When camera's created for some reason it's upVector is not (0, 1, 0) but some strange value like (0.0001, 0.9999, 0). That's why when we translate camera, these strange values appear...

I prepared a commit solve this, I'm going to push it soon. This commit also will solve the issue when camera was reset, but it's rotation is strange

yungyuc: That adds up.  If the input has an error of 1.e-4, the error of 1.23e-4 in the result makes sense.  We can wait for your change, but feel free to discuss more if you want.

Tucchhaa:
@yungyuc in these commits:
* I have improved camera reset, now camera's rotation is correct when it's reset
* Added tests for camera reset
* For translation tests removed tolerance and just check the exact value
* For rotation tests decreased tolerance from 1e-2 to 1e-5
* Added checks for camera's view center after translation and rotation
* Applied your review code changes and did some refactoring
* Skipped tests for Github actions

CI is green. ready for review 🙂

tests/test_view.py
```diff
@@ -59,4 +63,238 @@ def test_pycon(self):
...
+
+    def test_translation(self):
+        dt = 0.01
+        delta = dt * self.linear_speed()
+        delta_vec = np.array([delta, delta, -delta])
+        new_pos = np.array(self.pos()) + delta_vec
+        new_view_center = np.array(self.view_center()) + delta_vec
+
+        self.move(x=1, dt=dt)
+        self.assertEqual(self.pos()[0], new_pos[0])
```
yungyuc, line 151: Exact comparison tells a lot.  This is great.

It is a productive practice for the author of a code review (pull request) to add review comments at the place that worth reading.  The author is the most familiar with the code and they can make the comments in the fastest way.  Without the annotation it will take a review more time to understand the code.

tests/test_view.py
```diff
@@ -59,4 +63,238 @@ def test_pycon(self):
...
+        self.assertEqual(self.view_center(), initial_view_center)
+
+        # test camera rotates around y-axis
+        self.move(yaw=1, dt=dt, right_mouse_button=True)
+
+        rotation_matrix = self.angle_axis(angle, (0, -1, 0))
+        rotated_vector = np.array(initial_view_vector) @ rotation_matrix
+        view_vector = self.view_vector()
+
+        self.assertAlmostEqual(rotated_vector[0], view_vector[0], delta=1e-5)
```
yungyuc, line 281: [`numpy.testing.assert_allclose()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) can merge the 3 lines into 1, but it is not wrong to use [`assertAlmostEqual()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual).

yungyuc (approved):
This looks great.

tests/test_view.py
Tucchhaa: Thank you for the feedback) Will take into account

Tucchhaa:
Let's merge it then. Looks like I don't have the permission to do that. Will you do that?

PR #386, opened 2024-07-12
add clone for SimpleArray

tigercosmos:
per conversation in [Discord](https://discord.com/channels/730297880140578906/978989059844214784/1261223700770848768), adding the missing `clone` method for `SimpArray` and `SimpleArrayPlex`

the usage:
```py
        sarr = modmesh.SimpleArrayFloat64((2, 3, 4))
        sarr_ref = sarr
        sarr_clone = sarr.clone()

        self.assertTrue(sarr_ref is sarr)
        self.assertFalse(sarr_clone is sarr)
```

### DIFF
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -87,6 +87,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     }),
                 py::arg("array"))
             .def_buffer(&property_helper::get_buffer_info)
+            .def("clone", [](wrapped_type const & self)
+                 { return wrapped_type(self); }) // cloning the object using the copy constructor. never add the clone method to the C++ class.
             .def_property_readonly(
                 "ndarray",
                 [](wrapped_type & self)
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -87,7 +87,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     }),
                 py::arg("array"))
             .def_buffer(&property_helper::get_buffer_info)
-            .def("clone", [](wrapped_type const & self)
+            .def("clone",
+                 [](wrapped_type const & self)
                  { return wrapped_type(self); }) // cloning the object using the copy constructor. never add the clone method to the C++ class.
             .def_property_readonly(
                 "ndarray",
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -205,6 +205,14 @@ class SimpleArrayBasicTC(unittest.TestCase):
         self.assertEqual((1, 24), sarr.reshape((1, 24)).shape)
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
 
+    def test_SimpleArray_clone(self):
+        sarr = modmesh.SimpleArrayFloat64((2, 3, 4))
+        sarr_ref = sarr
+        sarr_clone = sarr.clone()
+
+        self.assertTrue(sarr_ref is sarr)
+        self.assertFalse(sarr_clone is sarr)
+
     def test_SimpleArray_ghost_1d(self):
 
         sarr = modmesh.SimpleArrayFloat64(4 * 3 * 2)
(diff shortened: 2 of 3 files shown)


tests/test_buffer.py
```diff
@@ -205,6 +205,14 @@ def test_SimpleArray(self):
         self.assertEqual((1, 24), sarr.reshape((1, 24)).shape)
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
 
+    def test_SimpleArray_clone(self):
```
tigercosmos, line 208: the test should be self-explained

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -364,6 +364,11 @@ class SimpleArray
 
     ~SimpleArray() = default;
 
+    SimpleArray clone() const
+    {
+        return SimpleArray(*this);
```
yungyuc: This is clearly a redundant helper function for C++ consumer, which obviously should use the copy constructor.

The fact that `SimpleArray` is not managed by a `std:shared_ptr` asserts that the copy constructor should be used for copying all contents of the array.

Do not add `SimpleArray::clone()`.  Please remove the helper.  But maybe you can add a line of comment around the copy constructor to explain that we should never add a helper like `clone()`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -748,6 +753,11 @@ class SimpleArrayPlex
 
     ~SimpleArrayPlex();
 
+    SimpleArrayPlex clone() const
```
yungyuc: No `SimpleArrayPlex::clone()` either.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,6 +87,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     }),
                 py::arg("array"))
             .def_buffer(&property_helper::get_buffer_info)
+            .def("clone", &wrapped_type::clone)
```
yungyuc: You may add `SimpleArray.clone()` (a Python function) in the Python wrapper.

yungyuc (changes requested):
Python does not have copy constructor (or the versatile constructors) like C++ has.  For the data management (and lifecycle management), we intentionally avoid the same API between Python and C++.

- [x] Never add `SimpleArray::clone()` or `SimpleArrayPlex::clone()` helpers.  Consider to add a comment explaining why around the copy constructor.

yungyuc:
I once made a C++ library/system (that is not publicly available) supporting exactly the same API between C++ and Python.  That was a bad idea.

yungyuc:
You can add `clone()` in the Python wrapper for `SimpleArray`.

Do not add `clone()` to the C++ `SimpleArray` class template or its dynamic multiplexor.  Users will get confused with the copy constructor.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: I see the point. fixed.

tigercosmos:
@yungyuc  The C++ clone helpers are removed. Please check.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,6 +87,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     }),
                 py::arg("array"))
             .def_buffer(&property_helper::get_buffer_info)
+            .def("clone", [](wrapped_type const & self)
+                 { return wrapped_type(self); }) // cloning the object using the copy constructor. never add the clone method to the C++ class.
```
yungyuc, line 92: When the second argument is a lambda, please move it to the next line, so that the indentation of the lambda body will be aligned with the character the lambda starts, like:

```cpp
            .def("clone",
                 [](wrapped_type const & self)
                 { return wrapped_type(self); }) // cloning the object using the copy constructor. never add the clone method to the C++ class.
```

tests/test_buffer.py
yungyuc: Yes it is.

tests/test_buffer.py
```diff
@@ -205,6 +205,14 @@ def test_SimpleArray(self):
...
+        sarr = modmesh.SimpleArrayFloat64((2, 3, 4))
+        sarr_ref = sarr
+        sarr_clone = sarr.clone()
+
+        self.assertTrue(sarr_ref is sarr)
+        self.assertFalse(sarr_clone is sarr)
```
yungyuc, line 217: Please also assert for content equality, like:

```python
# Test content equality through the numpy interface.
np.testing.assert_equal(sarr_clone.ndarray[...], sarr.ndarray[...])
# Alternately, test content equality through SimpleArray element interface.
np.testing.assert_equal(sarr_clone, sarr)
```

yungyuc (changes requested):
Thanks.  The fix looks good.  Other things to change including code formatting and testing.

- [x] Format lambda to align indentation.
- [x] Test content equality between the original and the clone.

tigercosmos:
@yungyuc the issues above are fixed. please check.

yungyuc (approved):
LGTM.  Thanks.

PR #388, opened 2024-07-16
support std::unordered_map for JSON Serializable

tigercosmos:
The PR adds the support of `std::unordered_map` for JSON Serializable, including the serialization and deserialization.

### DIFF
--- a/cpp/modmesh/serialization/SerializableItem.hpp
+++ b/cpp/modmesh/serialization/SerializableItem.hpp
@@ -179,7 +179,7 @@ std::string JsonHelper::to_json_string(const T & value)
         {
             keys.push_back(kv.first);
         }
-        std::sort(keys.begin(), keys.end());
+        std::sort(keys.begin(), keys.end()); // TODO: the sorting may not be necessary. This is more for the testing purpose.
 
         std::ostringstream oss;
         oss << "{";
--- a/cpp/modmesh/serialization/SerializableItem.hpp
+++ b/cpp/modmesh/serialization/SerializableItem.hpp
@@ -156,7 +156,7 @@ std::string JsonHelper::to_json_string(const T & value)
     {
         return value ? "true" : "false";
     }
-    else if constexpr (is_specialization_of<std::vector, T>::value)
+    else if constexpr (is_specialization_of_v<std::vector, T>)
     {
         std::ostringstream oss;
         oss << "[";
@@ -168,6 +168,29 @@ std::string JsonHelper::to_json_string(const T & value)
             separator = ",";
         }
         oss << "]";
+        return oss.str();
+    }
+    else if constexpr (is_specialization_of_v<std::unordered_map, T>)
+    {
+        static_assert(std::is_same_v<typename T::key_type, std::string>, "Only support std::unordered_map<std::string, ...>.");
+
+        std::vector<std::string> keys;
+        for (const auto & kv : value)
+        {
+            keys.push_back(kv.first);
+        }
+        std::sort(keys.begin(), keys.end());
+
+        std::ostringstream oss;
+        oss << "{";
+        const char * separator = "";
+        for (const auto & key : keys)
+        {
+            // NOLINTNEXTLINE(misc-no-recursion)
+            oss << separator << "\"" << key << "\":" << to_json_string(value.at(key)); /* recursive here */
+            separator = ",";
+        }
+        oss << "}";
         return oss.str();
     }
     else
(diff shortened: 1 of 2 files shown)


cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -170,6 +170,29 @@ std::string JsonHelper::to_json_string(const T & value)
         oss << "]";
         return oss.str();
     }
+    else if constexpr (is_specialization_of_v<std::unordered_map, T>)
+    {
+        static_assert(std::is_same_v<typename T::key_type, std::string>, "Only support std::unordered_map<std::string, ...>.");
```
tigercosmos, line 175: the key of the unordered_map should only be a string.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -170,6 +170,29 @@ std::string JsonHelper::to_json_string(const T & value)
...
+
+        std::vector<std::string> keys;
+        for (const auto & kv : value)
+        {
+            keys.push_back(kv.first);
+        }
+        std::sort(keys.begin(), keys.end());
```
tigercosmos: sort the key to ensure output consistency in all platforms.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -170,6 +170,29 @@ std::string JsonHelper::to_json_string(const T & value)
...
+
+        std::ostringstream oss;
+        oss << "{";
+        const char * separator = "";
+        for (const auto & key : keys)
+        {
+            // NOLINTNEXTLINE(misc-no-recursion)
+            oss << separator << "\"" << key << "\":" << to_json_string(value.at(key)); /* recursive here */
```
tigercosmos, line 190: get the string of the value of the map.

cpp/modmesh/serialization/SerializableItem.hpp
```diff
@@ -242,6 +265,19 @@ void JsonHelper::from_json_string(const std::unique_ptr<JsonNode> & node, T & va
...
+    {
+        if (node->type != detail::JsonType::Object)
+        {
+            throw std::runtime_error("Invalid JSON format: invalid object type.");
+        }
+        auto & obj = std::get<detail::JsonMap>(node->value);
+        for (const auto & [key, jsonNode] : obj)
+        {
+            // NOLINTNEXTLINE(misc-no-recursion)
+            from_json_string(jsonNode, value[key]); /* recursive here */
```
tigercosmos, line 278: load the string json into the map's value.

gtests/test_nopython_serializable.cpp
```diff
@@ -111,6 +111,16 @@ struct EscapeItem : SerializableItem
         register_member("escape_string", escape_string);)
 }; // end struct EscapeItem
 
+struct TestUnorderedMapItem : SerializableItem
```
tigercosmos, line 114: the test should be self-explained.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: I think something convertible to a string should also work, but that may complicate the code.  Type-check alone looks good right now.

cpp/modmesh/serialization/SerializableItem.hpp
yungyuc: It's not unusual to use a specific order (like that of key insertion).  Could you please add a code comment indicating the possibility (for future enhancement)?

gtests/test_nopython_serializable.cpp
yungyuc: Please also test for invalid key types like integer.

yungyuc (changes requested):
LGTM.  Just a minor enhancement for testing.

- [x] Test for invalid key types.

cpp/modmesh/serialization/SerializableItem.hpp
tigercosmos: Indeed, sorting seems to be unnecessary.  This is more for testing. For now, it shouldn't be a bottleneck, so let me just add a comment here.

gtests/test_nopython_serializable.cpp
tigercosmos: The check for for invalid key types is not required, since we already have the static check, which prevents the wrong usage:
```
/home/mujin/modmesh/cpp/modmesh/serialization/SerializableItem.hpp:175:28: error: static assertion failed: Only support std::unordered_map<std::string, ...>.
  175 |         static_assert(std::is_same_v<typename T::key_type, std::string>, "Only support std::unordered_map<std::string, ...>.");
  ```

tigercosmos:
@yungyuc Please check the comment [here](https://github.com/solvcon/modmesh/pull/388#discussion_r1680300962).

gtests/test_nopython_serializable.cpp
yungyuc: I see.  It does not make sense to test the invalid key from C++.  But we may need to test for it from Python.  We have not had the Python wrapper for the serializer.  We should make it in a followup PR.

yungyuc (approved):
Thanks.  This looks good.

PR #391, opened 2024-07-21
Hide boundary points in figure

Gene0315:
Following the issue https://github.com/solvcon/modmesh/pull/361
The PR hides the boundary points.

Features: The plotting hides the unreasonable points on the boundary.
Strength: It makes the figure look nicer without changing the solution data.
Weakness: It is better to calculate the correct values on the boundary and show them on the boundary.

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -743,19 +743,30 @@ class Euler1DApp():
         :return: None
         """
         if self.use_grid_layout:
-            self.density.update(adata=self.st.density_field,
-                                ndata=self.st.svr.density[self.st.svr.xindices])
-            self.pressure.update(adata=self.st.pressure_field,
-                                 ndata=self.st.svr.pressure[self.st.svr.xindices])
-            self.velocity.update(adata=self.st.velocity_field,
-                                 ndata=self.st.svr.velocity[self.st.svr.xindices])
-            self.temperature.update(adata=self.st.temperature_field,
-                                    ndata=self.st.svr.temperature[self.st.svr.xindices])
-            self.internal_energy.update(adata=(self.st.internal_energy_field),
-                                        ndata=(self.st.svr.
-                                               internal_energy[self.st.svr.xindices]))
-            self.entropy.update(adata=self.st.entropy_field,
-                                ndata=self.st.svr.entropy[self.st.svr.xindices])
+            self.density.update(
+                adata=self.st.density_field,
+                ndata=self.st.svr.density[self.st.svr.xindices]
+            )
+            self.pressure.update(
+                adata=self.st.pressure_field,
+                ndata=self.st.svr.pressure[self.st.svr.xindices]
+            )
+            self.velocity.update(
+                adata=self.st.velocity_field,
+                ndata=self.st.svr.velocity[self.st.svr.xindices]
+            )
+            self.temperature.update(
+                adata=self.st.temperature_field,
+                ndata=self.st.svr.temperature[self.st.svr.xindices]
+            )
+            self.internal_energy.update(
+                adata=(self.st.internal_energy_field),
+                ndata=(self.st.svr.internal_energy[self.st.svr.xindices])
+            )
+            self.entropy.update(
+                adata=self.st.entropy_field,
+                ndata=self.st.svr.entropy[self.st.svr.xindices]
+            )
         else:
             for name, is_selected, *_ in self.plot_config.state:
                 if is_selected:
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -530,7 +530,7 @@ class Euler1DApp():
 
         :return: FigureCanvas
         """
-        x = self.st.svr.coord[self.st.xindices]
+        x = self.st.svr.coord[self.st.svr.xindices]
         fig = Figure()
         canvas = FigureCanvas(fig)
         ax = canvas.figure.subplots(3, 2)
@@ -574,7 +574,7 @@ class Euler1DApp():
 
         :return: FigureCanvas
         """
-        x = self.st.svr.coord[self.st.xindices]
+        x = self.st.svr.coord[self.st.svr.xindices]
         fig = Figure()
         canvas = FigureCanvas(fig)
         ax = canvas.figure.subplots()
@@ -744,23 +744,23 @@ class Euler1DApp():
         """
         if self.use_grid_layout:
             self.density.update(adata=self.st.density_field,
-                                ndata=self.st.svr.density[self.st.xindices])
+                                ndata=self.st.svr.density[self.st.svr.xindices])
             self.pressure.update(adata=self.st.pressure_field,
-                                 ndata=self.st.svr.pressure[self.st.xindices])
+                                 ndata=self.st.svr.pressure[self.st.svr.xindices])
             self.velocity.update(adata=self.st.velocity_field,
-                                 ndata=self.st.svr.velocity[self.st.xindices])
+                                 ndata=self.st.svr.velocity[self.st.svr.xindices])
             self.temperature.update(adata=self.st.temperature_field,
-                                    ndata=self.st.svr.temperature[self.st.xindices])
+                                    ndata=self.st.svr.temperature[self.st.svr.xindices])
             self.internal_energy.update(adata=(self.st.internal_energy_field),
                                         ndata=(self.st.svr.
-                                               internal_energy[self.st.xindices]))
+                                               internal_energy[self.st.svr.xindices]))
             self.entropy.update(adata=self.st.entropy_field,
-                                ndata=self.st.svr.entropy[self.st.xindices])
+                                ndata=self.st.svr.entropy[self.st.svr.xindices])
         else:
             for name, is_selected, *_ in self.plot_config.state:
                 if is_selected:
                     eval(f'(self.{name}.update(adata=self.st.{name}_field,'
-                         f' ndata=self.st.svr.{name}[self.st.xindices]))')
+                         f' ndata=self.st.svr.{name}[self.st.svr.xindices]))')
 
 
 class PlotArea(PuiInQt):
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -511,6 +511,8 @@ class Euler1DApp():
         self.interval = self.solver_config.get_var("timer_interval", "value")
         self.max_steps = self.solver_config.get_var("max_steps", "value")
         self.profiling = self.solver_config.get_var("profiling", "value")
+        ncoord = self.solver_config.get_var("ncoord", "value")
+        self.plot_point = np.linspace(2, (ncoord - 3), num=((ncoord - 3) // 2), dtype=int)
 
     def setup_timer(self):
         """
@@ -530,7 +532,7 @@ class Euler1DApp():
 
         :return: FigureCanvas
         """
-        x = self.st.svr.coord[::2]
+        x = self.st.svr.coord[self.plot_point]
         fig = Figure()
         canvas = FigureCanvas(fig)
         ax = canvas.figure.subplots(3, 2)
@@ -574,7 +576,7 @@ class Euler1DApp():
 
         :return: FigureCanvas
         """
-        x = self.st.svr.coord[::2]
+        x = self.st.svr.coord[self.plot_point]
         fig = Figure()
         canvas = FigureCanvas(fig)
         ax = canvas.figure.subplots()
@@ -744,23 +746,23 @@ class Euler1DApp():
         """
         if self.use_grid_layout:
             self.density.update(adata=self.st.density_field,
-                                ndata=self.st.svr.density[::2])
+                                ndata=self.st.svr.density[self.plot_point])
             self.pressure.update(adata=self.st.pressure_field,
-                                 ndata=self.st.svr.pressure[::2])
+                                 ndata=self.st.svr.pressure[self.plot_point])
             self.velocity.update(adata=self.st.velocity_field,
-                                 ndata=self.st.svr.velocity[::2])
+                                 ndata=self.st.svr.velocity[self.plot_point])
             self.temperature.update(adata=self.st.temperature_field,
-                                    ndata=self.st.svr.temperature[::2])
+                                    ndata=self.st.svr.temperature[self.plot_point])
             self.internal_energy.update(adata=(self.st.internal_energy_field),
                                         ndata=(self.st.svr.
-                                               internal_energy[::2]))
+                                               internal_energy[self.plot_point]))
             self.entropy.update(adata=self.st.entropy_field,
-                                ndata=self.st.svr.entropy[::2])
+                                ndata=self.st.svr.entropy[self.plot_point])
         else:
             for name, is_selected, *_ in self.plot_config.state:
                 if is_selected:
                     eval(f'(self.{name}.update(adata=self.st.{name}_field,'
-                         f' ndata=self.st.svr.{name}[::2]))')
+                         f' ndata=self.st.svr.{name}[self.plot_point]))')
 
 
 class PlotArea(PuiInQt):
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -744,30 +744,19 @@ class Euler1DApp():
         """
         if self.use_grid_layout:
             _s = self.st.svr.xindices
-            self.density.update(
-                adata=self.st.density_field,
-                ndata=self.st.svr.density[_s]
-            )
-            self.pressure.update(
-                adata=self.st.pressure_field,
-                ndata=self.st.svr.pressure[_s]
-            )
-            self.velocity.update(
-                adata=self.st.velocity_field,
-                ndata=self.st.svr.velocity[_s]
-            )
-            self.temperature.update(
-                adata=self.st.temperature_field,
-                ndata=self.st.svr.temperature[_s]
-            )
-            self.internal_energy.update(
-                adata=(self.st.internal_energy_field),
-                ndata=(self.st.svr.internal_energy[_s])
-            )
-            self.entropy.update(
-                adata=self.st.entropy_field,
-                ndata=self.st.svr.entropy[_s]
-            )
+            self.density.update(adata=self.st.density_field,
+                                ndata=self.st.svr.density[_s])
+            self.pressure.update(adata=self.st.pressure_field,
+                                 ndata=self.st.svr.pressure[_s])
+            self.velocity.update(adata=self.st.velocity_field,
+                                 ndata=self.st.svr.velocity[_s])
+            self.temperature.update(adata=self.st.temperature_field,
+                                    ndata=self.st.svr.temperature[_s])
+            self.internal_energy.update(adata=(self.st.internal_energy_field),
+                                        ndata=(self.st.svr.
+                                               internal_energy[_s]))
+            self.entropy.update(adata=self.st.entropy_field,
+                                ndata=self.st.svr.entropy[_s])
         else:
             for name, is_selected, *_ in self.plot_config.state:
                 if is_selected:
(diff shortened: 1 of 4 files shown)


Gene0315:
This PR removed the unrelated files in https://github.com/solvcon/modmesh/pull/361 and set the shown points as a variable of Euler1DApp.
However, I still failed to make the plot_point as an universal variable for modmesh/app/euler1d.py and modmesh/onedim/euler1d.py. Is it necessary to do this?

modmesh/app/euler1d.py
```diff
@@ -511,6 +511,8 @@ def set_solver_config(self):
         self.interval = self.solver_config.get_var("timer_interval", "value")
         self.max_steps = self.solver_config.get_var("max_steps", "value")
         self.profiling = self.solver_config.get_var("profiling", "value")
+        ncoord = self.solver_config.get_var("ncoord", "value")
```
yungyuc: @j8xixo12 With this PR I realized that `self.solver_config.get_var()` is too wordy.  The lengthy API makes it hard to quickly write and understand code.

Please propose a new design of [`SolverConfig`](https://github.com/solvcon/modmesh/blob/8ccfa3229e23a1e6d5340c5314cfd782eeeccb5c/modmesh/app/euler1d.py#L107) and [`PlotConfig`](https://github.com/solvcon/modmesh/blob/8ccfa3229e23a1e6d5340c5314cfd782eeeccb5c/modmesh/app/euler1d.py#L230) in the [discussions](https://github.com/solvcon/modmesh/discussions).  We can probably consider `__getattr__()` or `__getitem__()`

----

@Gene0315 please provide an option to keep existing behavior.

modmesh/app/euler1d.py
```diff
@@ -511,6 +511,8 @@ def set_solver_config(self):
...
+        self.plot_point = np.linspace(2, (ncoord - 3), num=((ncoord - 3) // 2), dtype=int)
```
yungyuc: A convention of short-living temporary variables is `_`.  The two lines can be changed to:

```python
_ = self.solver_config.get_var("ncoord", "value")
self.xindices = np.linspace(2, (_ - 3), num=((_ - 3) // 2), dtype=int)
```

`plot_point` is not a good attribute name because the content is not points but indices of the coordinate (1D point) arrays.

yungyuc (changes requested):
- [x] Provide an option to select the amount of data to plot.  It can serve as a way to fall back to the original behavior.
- [x] Rename `plot_point` to `xindices`.
- [x] Fix CI errors.

----

Please fix the CI failures.  A prominent one is test failures on the array size: https://github.com/solvcon/modmesh/actions/runs/10025477375/job/27708689299?pr=391#step:11:285 :

```
=================================== FAILURES ===================================
____________________ ShockTubeTC.test_field_with_numerical _____________________

self = <test_onedim_euler.ShockTubeTC testMethod=test_field_with_numerical>

    def test_field_with_numerical(self):
        self.st.build_numerical(xmin=-1, xmax=1, ncoord=21,
                                time_increment=0.05)
        self.st.build_field(t=0.0)
        self.assertIsInstance(self.st.svr, euler1d.Euler1DSolver)
>       self.assertEqual(len(self.st.coord), 11)
E       AssertionError: 9 != 11

tests/test_onedim_euler.py:110: AssertionError
FAILED tests/test_onedim_euler.py::ShockTubeTC::test_field_with_numerical - AssertionError: 9 != 11
=================== 1 failed, 181 passed, 5 skipped in 6.72s ===================
make: *** [pytest] Error 1
```

The test fails because you changed the length of the array.

Gene0315:
@yungyuc 

> * [ ]  Provide an option to select the amount of data to plot.  It can serve as a way to fall back to the original behavior.
> * [ ]  @Gene0315 please provide an option to keep existing behavior.

Are these two the same questions? Do you mean to provide it as an option such as "ncoord" or "max step" in qt window?  
> * [ ]  Rename `plot_point` to `xindices`.

It is ok.
> * [ ]  Fix CI errors.

But I don't know how to check whether I solve the CI error correctly. Maybe I can create a PR in my own repo again before I upload the new code here?

yungyuc:
> @yungyuc
> 
> > * [ ]   Provide an option to select the amount of data to plot.  It can serve as a way to fall back to the original behavior.
> > * [ ]   @Gene0315 please provide an option to keep existing behavior.
> 
> Are these two the same questions? Do you mean to provide it as an option such as "ncoord" or "max step" in qt window?

Yes they are similar.  No, the options only need to be in the code.  We do not need them to be in the GUI yet.

> 
> > * [ ]   Rename `plot_point` to `xindices`.
> 
> It is ok.
> 
> > * [ ]   Fix CI errors.
> 
> But I don't know how to check whether I solve the CI error correctly. Maybe I can create a PR in my own repo again before I upload the new code here?

You can turn on Github Actions in your own repository and you will see the CI runs when you push to your repository.  You do not need a PR for the CI runs in your repository.  You should also run the tests locally to reproduce the failures on your development machine for fixing them.

modmesh/app/euler1d.py
yungyuc: @Gene0315 You can use the inline comment for focused discussions.

Gene0315:
> Yes they are similar. No, the options only need to be in the code. We do not need them to be in the GUI yet.

I'm so confused how to set the `xindices` as an option in the code. Can you give me an example? Just like which variable is an option in the code.
> You can turn on Github Actions in your own repository and you will see the CI runs when you push to your repository. You do not need a PR for the CI runs in your repository. You should also run the tests locally to reproduce the failures on your development machine for fixing them.

How can I run the check tests in my laptop? Is it feasable to run the `test_onedim_euler1d.py` or other test python files directly?

yungyuc:
> > Yes they are similar. No, the options only need to be in the code. We do not need them to be in the GUI yet.
> 
> I'm so confused how to set the `xindices` as an option in the code. Can you give me an example? Just like which variable is an option in the code.

Something like:

```python
def set_solver_config(self, keep_edge=False):
    # ...
    if keep_edge:
        # create self.xindices including the edge points.
    else:
        _ = self.solver_config.get_var("ncoord", "value")
        self.xindices = np.linspace(2, (_ - 3), num=((_ - 3) // 2), dtype=int)
```

In this way you can allow a configuring function to turn on `keep_edge`.  The consumer of the solver can decide whether to plot the edge points through that argument.

If you want to make the option available in GUI, the flag `keep_edge` can be moved where the PUI control can recognize.  No matter what you do, you should make unit tests pass.

> > You can turn on Github Actions in your own repository and you will see the CI runs when you push to your repository. You do not need a PR for the CI runs in your repository. You should also run the tests locally to reproduce the failures on your development machine for fixing them.
> 
> How can I run the check tests in my laptop? Is it feasable to run the `test_onedim_euler1d.py` or other test python files directly?

The `Makefile` has many targets exercising the tests: https://github.com/solvcon/modmesh/blob/6a9888dfd30f7e36473c821f89997ab68b772f0c/Makefile#L116

They are used in Github Actions too.

modmesh/app/euler1d.py
yungyuc: @j8xixo12 Any comments?

j8xixo12: Ok, I will create a discussion and propose a new design in the discussion.

yungyuc: > Ok, I will create a discussion and propose a new design in the discussion.

Thanks, @j8xixo12 .  Maybe creating an issue describing what we know and may be done can help.

Gene0315:
@yungyuc I passed all the checks finally! But, I don't know what I have to do next. Could you give me some tips?

startup.py
yungyuc, line 1: This is redundant file, please remove.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +302,27 @@ def calc_internal_energy(self, pressure, density):
...
+            _ = self.svr.ncoord
+            if keep_edge:
+                self.svr.xindices = np.linspace(
+                    0, (_ - 1), num=((_ + 1) // 2), dtype=int
+                )
+            else:
+                self.svr.xindices = np.linspace(
+                    2, (_ - 3), num=((_ - 3) // 2), dtype=int
+                )
+            # Use the numerical solver.
```
yungyuc: Please be more specific about what you meant.  "Use the numerical solver" is not a clear statement.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +302,27 @@ def calc_internal_energy(self, pressure, density):
...
-    def build_field(self, t, coord=None):
+    def build_field(self, t, coord=None, keep_edge=False):
         """
         Populate the field data using the analytical solution.
 
         :param t:
         :param coord: If None, take the coordinate from the numerical solver.
         :return: None
         """
+
```
yungyuc: This insertion of a blank line is not necessary.  Do you introduce unnecessary/unrelated change of blank lines and whitespaces in a patch (i.e., checkins/PR).

tests/test_onedim_euler.py
```diff
@@ -105,7 +105,7 @@ def test_field_without_numerical(self):
     def test_field_with_numerical(self):
         self.st.build_numerical(xmin=-1, xmax=1, ncoord=21,
                                 time_increment=0.05)
-        self.st.build_field(t=0.0)
+        self.st.build_field(t=0.0, keep_edge=True)
```
yungyuc: The fix of the unit tests by the addition of the single argument is good.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +302,27 @@ def calc_internal_energy(self, pressure, density):
...
         if None is coord:
-            coord = self.svr.coord[::2]  # Use the numerical solver.
+            _ = self.svr.ncoord
```
yungyuc: This 9 lines of code is mouthy.  Please use the follow code instead (please make proper adjustment if needed because I didn't test it myself):

```python
            _ = self.svr.ncoord
            start = 0 if keep_edge else 2
            stop = _ - stop - 1
            num =  # It makes more sense to calculate from start and stop
            xindices = np.linspace(start, stop, num=num, dtype='int32')
```

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +302,27 @@ def calc_internal_energy(self, pressure, density):
...
+            if keep_edge:
+                self.svr.xindices = np.linspace(
+                    0, (_ - 1), num=((_ + 1) // 2), dtype=int
```
yungyuc: It makes more sense to calculate `num` from `start` and `stop`.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +302,27 @@ def calc_internal_energy(self, pressure, density):
...
         :param t:
         :param coord: If None, take the coordinate from the numerical solver.
         :return: None
         """
+
         if None is coord:
-            coord = self.svr.coord[::2]  # Use the numerical solver.
+            _ = self.svr.ncoord
+            if keep_edge:
+                self.svr.xindices = np.linspace(
```
yungyuc: The `xindices` attribute should be created in the initialization helper `Euler1DSolver.init_solver()` of the object `self.svr`: https://github.com/solvcon/modmesh/blob/214b4743bae1995cf7c451855880a397eae559fc/modmesh/onedim/euler1d.py#L36

yungyuc: Do not add attributes outside the class.  It will be very difficult for a maintainer to keep track of where attributes are added, if they are not created in an organized way.

modmesh/app/euler1d.py
```diff
@@ -743,24 +743,35 @@ def update_lines(self):
...
-            self.velocity.update(adata=self.st.velocity_field,
-                                 ndata=self.st.svr.velocity[::2])
-            self.temperature.update(adata=self.st.temperature_field,
-                                    ndata=self.st.svr.temperature[::2])
-            self.internal_energy.update(adata=(self.st.internal_energy_field),
-                                        ndata=(self.st.svr.
-                                               internal_energy[::2]))
-            self.entropy.update(adata=self.st.entropy_field,
-                                ndata=self.st.svr.entropy[::2])
+            self.density.update(
```
yungyuc: Please simplify the code like:

```python
            _s = self.st.svr.xindices
            self.density.update(adata=self.st.density_field,
                                ndata=self.st.svr.density[_s])
```

In this way, you do not change the original code by too much.  It will be easier to keep track of the change.

yungyuc (changes requested):
> @yungyuc I passed all the checks finally! But, I don't know what I have to do next. Could you give me some tips?

Congratulations! Passing all checks means your code starts to work. Then we can start code review.

----

Please address the following points:

- [x] Remove the redundant `startup.py`.
- [x] Use a more specific code comment than "Use the numerical solver".
- [x] Remove all unnecessary add/removal/change of blank lines and white spaces.
- [x] It makes more sense to calculate `num` from `start` and `stop`.
- [ ] There should be only one call to `np.linspace()`.  The conditional (if/else) should be used to calculate the arguments to the helper.
- [x] The `xindices` attribute should be created in the initialization helper `Euler1DSolver.init_solver()` of the object `self.svr`.
- [ ] Reduce the change to the code in the function `update_lines()`.

----

@j8xixo12 Could you please also help take a look?

modmesh/onedim/euler1d.py
Gene0315: > Do not add attributes outside the class. It will be very difficult for a maintainer to keep track of where attributes are added, if they are not created in an organized way.

@yungyuc I agree that we shouldn't add attributes outside the class, but I failed to create the `xindices` in the Euler1DCore https://github.com/solvcon/modmesh/blob/214b4743bae1995cf7c451855880a397eae559fc/modmesh/onedim/euler1d.py#L38
I tried to add the `xindices` in 
https://github.com/solvcon/modmesh/blob/214b4743bae1995cf7c451855880a397eae559fc/cpp/modmesh/onedim/Euler1DCore.cpp#L45 
like `m_coord` in
https://github.com/solvcon/modmesh/blob/214b4743bae1995cf7c451855880a397eae559fc/cpp/modmesh/onedim/Euler1DCore.cpp#L52
but I'm confused how to add it because I cannot decide its size in the begining. 

Or, maybe I misunderstand what you want me to do? Can you give me some suggestion?

yungyuc: `self.svr` is a `Euler1DSolver` object.  You can create `xindices` after https://github.com/solvcon/modmesh/blob/214b4743bae1995cf7c451855880a397eae559fc/modmesh/onedim/euler1d.py#L28

The code may be like:

```python
    def __init__(self, xmin, xmax, ncoord, time_increment=0.05):
        self._core = self.init_solver(xmin, xmax, ncoord, time_increment,
                                      gamma=1.4)
        self.xindices = ...
        # gamma is 1.4 for air.
```

You may consider to let `Euler1DSolver.__init__()` take a new argument for `keep_edge`, if necessary.

Gene0315:
@yungyuc I tried to revise the code like what you mentioned above. Could you take a look at it? I'm not sure whether the code is fine now. Thank you!

modmesh/app/euler1d.py
yungyuc: Change of line breaking is not necessary for updating the slicing (from `[::2]` to `[_s]`).  Please revert it to the original style.

That is, since the original style is:

```python
            self.pressure.update(adata=self.st.pressure_field,
                                 ndata=self.st.svr.pressure[::2])
```

instead of writing (with the different line breaking):

```python
            self.pressure.update(                                
                adata=self.st.density_field,
                ndata=self.st.svr.density[_s]
             )
```

write

```python
            self.pressure.update(adata=self.st.pressure_field,
                                 ndata=self.st.svr.pressure[_s])
```

The unnecessary change will baffle future maintainers.  Style matters.  Don't change the style when changing the logic, and vice versa.  Change of logic and that of style should not take place at the same time.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +308,22 @@ def calc_internal_energy(self, pressure, density):
...
         :return: None
         """
-        if None is coord:
-            coord = self.svr.coord[::2]  # Use the numerical solver.
+        if coord is None:
+            _ = self.svr.ncoord - 1
+            start = 0 if keep_edge else 2
+            stop = _ if keep_edge else (_ - 2)
+            num = (stop - start) // 2 + 1
+            self.svr.xindices = np.linspace(start, stop, num, dtype='int32')
```
yungyuc: Write `self.svr.xindices[...] = np.linspace(start, stop, num, dtype='int32')`.  Note that the first `...` in `[]` is [ellipsis](https://numpy.org/doc/stable/user/basics.indexing.html#slicing-and-striding).  By using the ellipsis, you are using the `xindices` created in the solver object instead of creating a new one and replacing the existing one.

modmesh/onedim/euler1d.py
```diff
@@ -302,16 +308,22 @@ def calc_internal_energy(self, pressure, density):
...
         Populate the field data using the analytical solution.
 
         :param t:
         :param coord: If None, take the coordinate from the numerical solver.
         :return: None
         """
-        if None is coord:
-            coord = self.svr.coord[::2]  # Use the numerical solver.
+        if coord is None:
+            _ = self.svr.ncoord - 1
```
yungyuc: Why do you need to repeat the same code here and in [line 32](https://github.com/solvcon/modmesh/pull/391/files#diff-00396195422e4a46224688f7c4b3aa59a6ceedd1d4c37cc274d4e86597737561R32)?

yungyuc (changes requested):
Some old issues were not handled and there are also new issues.  Please fix.

- [x] [`app/euler1d.py:747`](https://github.com/solvcon/modmesh/pull/391/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R747): Do not change the line breaking (style).  This has been request in the [previous review](https://github.com/solvcon/modmesh/pull/391#discussion_r1712611030).
- [x] [`onedim/euler1d.py:324`](https://github.com/solvcon/modmesh/pull/391/files#diff-00396195422e4a46224688f7c4b3aa59a6ceedd1d4c37cc274d4e86597737561R324): Do not create a new object for `xindices` in `self.svr` when updating the value.  Instead, force to use the existing one by writing `self.svr.xindices[...]`.
- [x] [`onedim/euler1d.py:320`](https://github.com/solvcon/modmesh/pull/391/files#diff-00396195422e4a46224688f7c4b3aa59a6ceedd1d4c37cc274d4e86597737561R320): Why duplicate code?

Gene0315:
> Some old issues were not handled and there are also new issues. Please fix.
> 
> * [ ]  [`app/euler1d.py:747`](https://github.com/solvcon/modmesh/pull/391/files#diff-4ebeb3148c683103bf030637fdf520154a4b12d9b660a82610861aa5250a0be1R747): Do not change the line breaking (style).  This has been request in the [previous review](https://github.com/solvcon/modmesh/pull/391#discussion_r1712611030).

Sorry, in my understanding, passing `make flake8` means the writing style is fine. I will notice it more carefully next time.

> * [ ]  [`onedim/euler1d.py:324`](https://github.com/solvcon/modmesh/pull/391/files#diff-00396195422e4a46224688f7c4b3aa59a6ceedd1d4c37cc274d4e86597737561R324): Do not create a new object for `xindices` in `self.svr` when updating the value.  Instead, force to use the existing one by writing `self.svr.xindices[...]`.
> * [ ]  [`onedim/euler1d.py:320`](https://github.com/solvcon/modmesh/pull/391/files#diff-00396195422e4a46224688f7c4b3aa59a6ceedd1d4c37cc274d4e86597737561R320): Why duplicate code?

In previous code, I want to know the size of `xindices` when init, but I still need to reset it in different conditions with `keep_edge`. Thus, the code seems to be repeated. 
Now, I tried to move the option `keep_edge` from function `build_field` to `build_numerical` and create the `xindices` in `Euler1DSolver` directly to avoid the code repeat.

yungyuc (approved):
Thanks.  It now looks perfect!

PR #392, opened 2024-07-22
Add README content: make it explicit modmesh is modules using meshes

yungyuc:
Just change README


tigercosmos:
I suggest putting a picture of the QT application.

yungyuc:
> I suggest putting a picture of the QT application.

That will be later.  I do not want to add blob files to the repository.  Need to figure out a way to add images.

yungyuc:
> I suggest putting a picture of the QT application.

I intend to put documentation in the parent project [solvcon](https://github.com/solvcon/solvcon) which has documents already: https://doc.solvcon.net/en/latest/ .

tigercosmos:
@yungyuc But I didn't see any picture of the application in the document that you mentioned.

yungyuc:
> @yungyuc But I didn't see any picture of the application in the document that you mentioned.

https://doc.solvcon.net/en/latest/mesh.html @tigercosmos here you go.

Screenshots for modmesh viewer need to wait.  I don't like how it is named.  "Viewer" is too generic and we must give it a better name.  The code structure should also change after the rename is done.

@tigercosmos It seems to me you want to have screenshots right now and I am interested in why?  But please let me know if I guessed wrongly.

tigercosmos:
> > @yungyuc But I didn't see any picture of the application in the document that you mentioned.
> 
> https://doc.solvcon.net/en/latest/mesh.html @tigercosmos here you go.
> 
> Screenshots for modmesh viewer need to wait. I don't like how it is named. "Viewer" is too generic and we must give it a better name. The code structure should also change after the rename is done.
> 
> @tigercosmos It seems to me you want to have screenshots right now and I am interested in why? But please let me know if I guessed wrongly.

@yungyuc  People become interested when they can see something visually, which is why I believe including some screenshots would be beneficial.

yungyuc:
> @yungyuc People become interested when they can see something visually, which is why I believe including some screenshots would be beneficial.

I agree without reservation.  But the code is not even in a shape good enough for a release.  For a release, I think the minimal requirement is https://github.com/solvcon/modmesh/milestone/2 .  For promotional screenshots, I think https://github.com/solvcon/modmesh/milestone/1 may be the minimal requirement.

Those are my 2 cents.  @j8xixo12 @terrychan999 , what do you think?

Issue #320, opened 2024-04-28
Use Bezier curve to fit NACA 4-digit airfoil 


yungyuc:
In the current prototype of the `naca` app, the airfoil is described [using sampled edges](https://github.com/solvcon/modmesh/blob/bc9e26f07585d74b2fb4d96c194ff209e7c78996/modmesh/app/naca.py#L104).  I'd like to try to describe the airfoil using Bezier curves, so that generating meshes for the airfoil shapes ca use arbitrarily high resolution.

### Progress

- [x] PR #384 adds a prototype that samples the 4th-order polynomial profile of the airfoil using both cubic Bezier curves and line segments, but has not done fitting.  More details in the comment: https://github.com/solvcon/modmesh/issues/320#issuecomment-2281235786
- [x] PR #406 refactors the code added in PR #384.  More details in the comment: https://github.com/solvcon/modmesh/issues/320#issuecomment-2282744365
- [ ] Use cubic Bezier curves and line segments to fit the airfoil profile.

### Open for discussions

Add GUI control for changing the airfoil drawing and fitting parameters.

yungyuc:
PR #384 adds a prototype that samples the 4th-order polynomial profile of the airfoil using both cubic bezier curves: https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/modmesh/gui/naca.py#L113-L127

And straight lines: https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/modmesh/gui/naca.py#L106-L110

But fitting is not done yet.

yungyuc:
@YenPeiChen07 Could you please help completing all tasks for fitting the airfoil?  I may add more as we go.  But if the work lasts too long I may create a new issue to track.

If you are not familiar with an optimization algorithm like linear least square, perhaps discussing with @Gene0315 can help.

yungyuc:
PR #406 refactors the code added in PR #384 and moved the calculating and drawing code from `modmesh/gui/naca.py` to `modmesh/pilot/airfoil/__init__.py`.  `modmesh/gui/naca.py` is left for adding interactive GUI code in the future.

[...]

PR #397, opened 2024-07-28
Rebuild the serializer of the profiler with general serializer

### LINKED ISSUE #343: Prototype profiler object serialization using JSON
This issue derives from #302.

Develop a helper class or some helper classes to serialize [the `CallProfile` object](https://github.com/solvcon/modmesh/blob/69b7e899bcb554cf6068951f6c79048017e2b7fe/cpp/modmesh/toggle/RadixTree.hpp#L178) used for recording the profile data (in memory).  Remarks:

1. The helper class(es) should use proper encapsulation.
2. The serialized data should be testable in Python.  It may also be tested in C++ (google-test) but the C++ tests are not mandatory.
3. The serialized JSON string should be human readable.  The readability should be easy and use decent indentation.

There should be a capability to deserialize the data back to `CallProfile`.  We may use this issue to track the deserialization, or use a different issue.

---

- [ ] Make more member data for serialization.  (See https://github.com/solvcon/modmesh/issues/343#issuecomment-2308388291)
- [ ] Implement deserializer.  (See https://github.com/solvcon/modmesh/issues/343#issuecomment-2308388291)

ThreeMonth03:
According to the [discussion in issue343](https://github.com/solvcon/modmesh/issues/343#issuecomment-2191837756), I rebuild the serializer of the profiler with general serializer, but I haven't finished the unit test.

By the way, @q40603 has already implemented [the similar function](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/pymod/wrap_profile.cpp#L201C1-L201C65), should I keep the related code when wrapping the serializer and writing the pytest?


cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -121,8 +121,7 @@ namespace detail
 class CallProfilerTest; // for gtest
 } /* end namespace detail */
 
-class CallProfilerSerializer; // for declaration
-
+class SerializableRadixTree; // for declaration
```
yungyuc: Typo: forward declaration.

cpp/modmesh/toggle/SerializableProfiler.hpp
```diff
@@ -0,0 +1,162 @@
...
+        register_member("call_count", m_call_count);
+        register_member("is_running", m_is_running);)
+
+private:
+
+    std::string m_caller_name;
+    std::chrono::nanoseconds m_total_time;
+    int m_call_count;
+    bool m_is_running;
+};
```
yungyuc: Ending markers.

gtests/test_nopython_callprofilerserializer.cpp
```diff
@@ -1,554 +0,0 @@
-#include <gtest/gtest.h>
```
yungyuc, line 1: Do you need to port some of the C++ testing code to Python?

yungyuc (changes requested):
It took me a while to find time to review the PR.  @ThreeMonth03 if think I am responding too slowly don't hesitate to add a comment and tag to remind me.

@q40603 We need your review too.

> By the way, @q40603 has already implemented [the similar function](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/pymod/wrap_profile.cpp#L201C1-L201C65), should I keep the related code when wrapping the serializer and writing the pytest?

@ThreeMonth03 by "keep the related code", do you mean to keep @q40603 's code in "[the similar function](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/pymod/wrap_profile.cpp#L201C1-L201C65)"?  Is it correct that another possible treatment is to replace @q40603 's `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` with @tigercosmos 's JSON serializer?

----

To fix:
- [x] Unit tests must be included in the PR.  The unit tests should serialize the profiler using the general JSON serializer and test some basic data.

We should not merge PRs without unit tests unless the change is too difficult to change.  An example of hard-to-test change is graphical hardware or UI interaction.

In this PR, the code should be easy to change.  I do not see a reason to miss tests, but please correct me if I am wrong.

----

Discussions:
- [x] Do you need to port some of the C++ testing code from `test_nopython_callprofilerserializer.cpp` to Python?

ThreeMonth03:
> It took me a while to find time to review the PR. @ThreeMonth03 if think I am responding too slowly don't hesitate to add a comment and tag to remind me.

It's fine.

> It took me a while to find time to review the PR. @ThreeMonth03 if think I am responding too slowly don't hesitate to add a comment and tag to remind me.
> 
> @q40603 We need your review too.
> 
> > By the way, @q40603 has already implemented [the similar function](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/pymod/wrap_profile.cpp#L201C1-L201C65), should I keep the related code when wrapping the serializer and writing the pytest?
> 
> @ThreeMonth03 by "keep the related code", do you mean to keep @q40603 's code in "[the similar function](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/pymod/wrap_profile.cpp#L201C1-L201C65)"? Is it correct that another possible treatment is to replace @q40603 's `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` with @tigercosmos 's JSON serializer?

Yes, I'm not sure whether I should replace @q40603 's `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` with @tigercosmos 's JSON serializer, or whether I should create another function like  `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`.

Currently, the function `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` can return the important result clearly, and I'm satisfied with the function. However, it doesn't return some minor information like [this](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/RadixTree.hpp#L200), and it may become difficult when deserialize the object. Thus, I prefer wrapping the  @tigercosmos 's JSON serializer by creating another function like  `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`. How do you think?

yungyuc:
> Yes, I'm not sure whether I should replace @q40603 's `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` with @tigercosmos 's JSON serializer, or whether I should create another function like `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`.

I feel that is what you should do.  Could you please ask @q40603 to take a look at the discussions here?  (It seems that he is occupied and did not notice here, but a gentle reminder should be OK.)
 
> Currently, the function `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` can return the important result clearly, and I'm satisfied with the function. However, it doesn't return some minor information like [this](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/RadixTree.hpp#L200), and it may become difficult when deserialize the object. Thus, I prefer wrapping the @tigercosmos 's JSON serializer by creating another function like `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`. How do you think?

Your proposal looks nice, and we can update the Python tests that consumed `result()` to use the new `serialize()`.  Python has JSON library built in and you can use it to convert the serialized JSON to Python dictionary.  It's possible that the existing tests only need minor update.

I think your `serialize()` should return a JSON string, but please correct me if I am wrong:

```cpp
std::string WrapCallProfiler::serialize(CallProfiler & profiler)
{
    // return the serialized JSON string.
}
```

If @q40603 does not have time to join the discussion, please go ahead execute your plan.  If `result()` should be there, we can add it back in a new PR.

ThreeMonth03:
> > Yes, I'm not sure whether I should replace @q40603 's `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` with @tigercosmos 's JSON serializer, or whether I should create another function like `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`.
> 
> I feel that is what you should do. Could you please ask @q40603 to take a look at the discussions here? (It seems that he is occupied and did not notice here, but a gentle reminder should be OK.)
> 
> > Currently, the function `pybind11::dict WrapCallProfiler::result(CallProfiler & profiler) { ... }` can return the important result clearly, and I'm satisfied with the function. However, it doesn't return some minor information like [this](https://github.com/solvcon/modmesh/blob/fde363640c959cb43493012885d361ebb8c423dd/cpp/modmesh/toggle/RadixTree.hpp#L200), and it may become difficult when deserialize the object. Thus, I prefer wrapping the @tigercosmos 's JSON serializer by creating another function like `pybind11::dict WrapCallProfiler::serialize(CallProfiler & profiler) { ... }`. How do you think?
> 
> Your proposal looks nice, and we can update the Python tests that consumed `result()` to use the new `serialize()`. Python has JSON library built in and you can use it to convert the serialized JSON to Python dictionary. It's possible that the existing tests only need minor update.
> 
> I think your `serialize()` should return a JSON string, but please correct me if I am wrong:
> 
> ```c++
> std::string WrapCallProfiler::serialize(CallProfiler & profiler)
> {
>     // return the serialized JSON string.
> }
> ```
> 

Yes, you are right.

> If @q40603 does not have time to join the discussion, please go ahead execute your plan. If `result()` should be there, we can add it back in a new PR.

I temporaily keep the `result()` function, and in the follow, there is a minor issue I want to discuss.

gtests/test_nopython_callprofilerserializer.cpp
ThreeMonth03: Yes, I test the serialize result by python.

cpp/modmesh/toggle/RadixTree.hpp
```diff
@@ -121,8 +121,7 @@ namespace detail
 class CallProfilerTest; // for gtest
 } /* end namespace detail */
 
-class CallProfilerSerializer; // for declaration
-
+class SerializableRadixTree; // forward declaration
```
ThreeMonth03, line 125: Fix the typo.

cpp/modmesh/toggle/SerializableProfiler.hpp
```diff
@@ -0,0 +1,138 @@
...
+        auto radix_tree_root = profiler.radix_tree().get_root();
+        if (radix_tree_curent_node != radix_tree_root)
+        {
+            throw std::runtime_error("The callprofiler cannot be serialized because it is measuring some functions.");
+        }
+        SerializableRadixTree serializable_radix_tree(profiler.radix_tree());
+        return serializable_radix_tree.to_json();
+    }
+
+}; /* end class CallProfilerSerializer */
```
ThreeMonth03, line 133: Add the ending markers.

tests/test_callprofiler.py
```diff
@@ -284,3 +284,92 @@ def baz():
...
+        @profile_function
+        def foo():
+            busy_loop(0.1)
+            serialization_string = modmesh.call_profiler.serialize()
+            bar()
+
+        modmesh.call_profiler.reset()
+        self.assertRaises(RuntimeError, foo)
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
```
ThreeMonth03: It is the unit test of the function  `modmesh.call_profiler.serialize()`.

ThreeMonth03: It is worth noting that I ban the function `modmesh.call_profiler.serialize()`  during profiling, because both the motivation and the serialization result are wierd. For example, if we serialize during profiling and deserialize after profiling, then the profiler will wait some exited functions.

But in line174, the function `result` can print part of the result during profiling. I'm not sure whether it is reasonable, and whether I should rebuild the function `result()` with `serialize()`.

tests/test_callprofiler.py
```diff
@@ -284,3 +284,91 @@ def baz():
...
+            busy_loop(0.5)
+
+        @profile_function
+        def foo():
+            busy_loop(0.1)
+            modmesh.call_profiler.serialize()
+            bar()
+
+        modmesh.call_profiler.reset()
+        self.assertRaises(RuntimeError, foo)
```
ThreeMonth03: Here is the unit test of the function `serializtion`.

ThreeMonth03: It is worth noting that I ban the function `modmesh.call_profiler.serialize()` during profiling, because both the motivation and the serialization result are wierd. For example, if we serialize the profiler during profiling and deserialize the result after profiling, the profiler will wait the exited function.

However, in line 174, the function `result` support printing the information during profiling. I have no idea whether I should rebuild the function `result` by `modmesh.call_profiler.serialize()`.

ThreeMonth03 (commented):
- [x] Fix the typo.
- [x] Add the unit test.
- [x] Add the ending markers.

ThreeMonth03:
> * [x]  Fix the typo.
> * [x]  Add the unit test.
> * [x]  Add the ending markers.

@yungyuc You could review the code if you are available. Thank you.

q40603:
@ThreeMonth03  @yungyuc Sorry for the late reply.
I think the purpose of serialize is to record and replay the profiling runtime.
And sometimes the user just want to access the clear profiling result without some private data being exposed, such as unique_id, id_map.

So maybe the viewable result and the serialize function can be co-exist, what do you think?

BTW, I am fine with replacing the profiling result wrapper with JSON @tigercosmos  's JSON serializer.

ThreeMonth03:
> @ThreeMonth03 @yungyuc Sorry for the late reply. I think the purpose of serialize is to record and replay the profiling runtime. And sometimes the user just want to access the clear profiling result without some private data being exposed, such as unique_id, id_map.
> 
> So maybe the viewable result and the serialize function can be co-exist, what do you think?

It looks good to me.

> 
> BTW, I am fine with replacing the profiling result wrapper with JSON @tigercosmos 's JSON serializer.

Although [I ban the function `serialize()` during profiling](https://github.com/solvcon/modmesh/pull/397#discussion_r1704746507),  in order to replacing the profiling result with @tigercosmos 's JSON serializer, should the JSON serializer return `is_current_node`?

tests/test_callprofiler.py
```diff
@@ -284,3 +284,91 @@ def baz():
         self.assertGreaterEqual(baz_dict["total_per_call"], ref_per_call)
         self.assertGreaterEqual(baz_dict["cumulative_time"], time3)
         self.assertGreaterEqual(baz_dict["cumulative_per_call"], time3)
+
+    def test_zero_callers_serializing(self):
+        modmesh.call_profiler.reset()
+        serialization_string = modmesh.call_profiler.serialize()
```
yungyuc: It looks nicer in the test code by chaining this line with the next, like:

```python
        sdict = json.loads(modmesh.call_profiler.serialize())
```

It's OK to name the return as `serialization_dict`, but it is unnecessarily long.  The key point is that the return is a dictionary after serialization.  If you can accept a short-hand, something like `sdict` is OK too.

yungyuc: But do not name it as `dict` which is the Python `dict` class name.

tests/test_callprofiler.py
yungyuc: Pretty good

yungyuc: Please use [`assertRaisesRegex()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) to test for the exception message as well.

yungyuc: It sounds like a bug or misbehavior to fail to serialize while profiling.  If it is, we should create an issue to fix it.

@q40603 @tigercosmos what do you think?

yungyuc (changes requested):
> > @ThreeMonth03 @yungyuc Sorry for the late reply. I think the purpose of serialize is to record and replay the profiling runtime. And sometimes the user just want to access the clear profiling result without some private data being exposed, such as unique_id, id_map.

In my experience it is rare to take only part of the data from profiler when we want to analyze the profiling data.  We usually pull all data and trim off what we do not want.

The ID information seems to be important that I cannot think of a scenario that I do not want it.  If you have a use case, could you please elaborate, @q40603 ?

But I doubt it exists.  I think we should always return all data before a reasonable use case of partial return is identified.

> > So maybe the viewable result and the serialize function can be co-exist, what do you think?
> 
> It looks good to me.

It is fairly straight-forward to take the serialized data and turn it into a viewable format.  Besides, JSON is pretty human-readable.  (It is much more human-readable than XML and plain text without a hierarchy.)  Two functions returning the same data in two formats are a [code smell](https://en.wikipedia.org/wiki/Code_smell).  It is unnecessary duplicate of code.

If we want the "viewable result", please write a function to reinterpret the serialized JSON.  We should remove the `result()` function.  But I am fine removing it in this PR or using another one.

> > BTW, I am fine with replacing the profiling result wrapper with JSON @tigercosmos 's JSON serializer.
> 
> Although [I ban the function `serialize()` during profiling](https://github.com/solvcon/modmesh/pull/397#discussion_r1704746507), in order to replacing the profiling result with @tigercosmos 's JSON serializer, should the JSON serializer return `is_current_node`?

It makes sense to serialize profile data when the profiler is running.  It also helps debugging the profiler.  `serialize()` should be allowed during profiling, as I commented at the line you mentioned.

----

- [x] Please use [`assertRaisesRegex()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) to test for the exception message as well.

----

Quite some progress.  Thanks a lot.

tests/test_callprofiler.py
tigercosmos: I'm curious about why it failed at this point. It's important to note that if we perform serialization during profiling, the resulting serialized data will be incomplete, which poses another problem, so I believe the function should be called after profiling. However, streaming serialization could be an option, but I'm uncertain about its feasibility within the current architecture.

cpp/modmesh/toggle/RadixTree.cpp
```diff
@@ -116,169 +116,6 @@ void CallProfiler::print_statistics(const RadixTreeNode<CallerProfile> & node, s
     TimeRegistry::me().clear();
 }
 
-void CallProfilerSerializer::serialize_call_profiler(const CallProfiler & profiler, std::ostream & outstream)
```
tigercosmos, line 119: I think the test is now clearer than the before one. :)

tests/test_callprofiler.py
ThreeMonth03: @tigercosmos In detail, the profiler starts to measure the function when `CallProfilerProbe` is initialized, and it ends to measure the function when `CallProfilerProbe` is destructed.

https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/cpp/modmesh/toggle/RadixTree.hpp#L310-L346

When it starts to measure function, the function `CallProfiler::start_caller(const std::string & caller_name, std::function<void()> cancel_callback);` calls `RadixTree::entry(const std::string & name);` and moves `RadixTree::m_current_node` to its child.

https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/cpp/modmesh/toggle/RadixTree.hpp#L254-L261

https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/cpp/modmesh/toggle/RadixTree.hpp#L137-L153

When it ends to measure function, the function `CallProfiler::end_caller();` updates `total_time` in `RadixTree::m_current_node` and moves `RadixTree::m_current_node` to its parent.

https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/cpp/modmesh/toggle/RadixTree.hpp#L264-L269

https://github.com/solvcon/modmesh/blob/6e634cdb499d4e14a2e0b9dc81c67ed862d4319c/cpp/modmesh/toggle/RadixTree.hpp#L213-L219

Thus, if the profiler is serialized during profiling, the following issues arise:
1. The `total_time` in `RadixTree::m_current_node` is not updated. Additionally, if `m_current_node` visits child nodes for the first time, the serialized result will show that the execution time and count of the function nodes are both zero.
2. `m_current_node` should move to the root node after profiling. Although `m_current_node` is not the root node, we can consider forcefully setting `m_current_node` to the root node during deserialization.

tigercosmos: Then I wonder why we can't do the serialization/deserialization when the profiler is not running?

ThreeMonth03: @tigercosmos I think the profiler is running regularly in this testcase. I refered [this post](https://stackoverflow.com/a/65210404) to test the exception.

The correct way to test the Exception is raised:
`self.assertRaises(RuntimeError, foo)`

The wrong way:
`self.assertRaises(RuntimeError, foo())`

I'm not sure whether I answered your question.

ThreeMonth03: > @tigercosmos I think the profiler is running regularly in this testcase. I refered [this post](https://stackoverflow.com/a/65210404) to test the exception.
> 
> The correct way to test the Exception is raised: `self.assertRaises(RuntimeError, foo)`
> 
> The wrong way: `self.assertRaises(RuntimeError, foo())`
> 
> I'm not sure whether I answered your question.

In this testcase, the profiler only measures the start time of function `foo()`, and it doesn't measure the end time of the function `foo()`. Then, the result of the serialization is not complete. Thus, in the line 128 of `cpp/modmesh/toggle/SerializableProfiler.hpp`, I write a simple exception to prevent the circumstances temporarily.

yungyuc: #### @ThreeMonth03 

> In this testcase, the profiler only measures the start time of function `foo()`, and it doesn't measure the end time of the function `foo()`. Then, the result of the serialization is not complete. Thus, in the line 128 of `cpp/modmesh/toggle/SerializableProfiler.hpp`, I write a simple exception to prevent the circumstances temporarily.

If I understood correctly, you are describing a reentrant problem in the profiler that manifests with the serialization.  Perhaps the stack overflow example may help: https://stackoverflow.com/a/33445858

#### @tigercosmos 

> Then I wonder why we can't do the serialization/deserialization when the profiler is not running?

Other test code tests for serialization/deserialization when the profiler is not running already.  The test function `test_serialize_during_profiling()` here is specifically testing serialization when the profiler is running.

I fail to understand the purpose of your question.  Could you please elaborate?

tigercosmos: > The test function test_serialize_during_profiling() here is specifically testing serialization when the profiler is running.

Sorry for the confusing. My question is why do we need to have ` test_serialize_during_profiling`? Does it make sense to do the serialization while the profiler is running?

yungyuc: > Sorry for the confusing. My question is why do we need to have ` test_serialize_during_profiling`? Does it make sense to do the serialization while the profiler is running? 

No problem.  Yes it does.  One reason is for debugging/testing as I mentioned in https://github.com/solvcon/modmesh/pull/397#pullrequestreview-2221406932 .

There are two more.  The second and more important reason is to pave the road for thread safety, which is required for sample-based profiling.  Without the reentrance, it will be hard to analyze issues from multi-threading.  Although the current profiler is scope-based, nothing prevents us from upgrading it to be also sample-based.  It should be very interesting to have a profiler supporting both scope- and sample-based profiling.  The capability will be powerful in performance tuning.

The third reason is also important: for completeness.  If we special-case serialization, there will be more "special cases" for profilable functions.  We don't know how "special" should qualify special cases.  It's better to make the profiler reentrant to avoid ambiguity.

tests/test_callprofiler.py
```diff
@@ -284,3 +284,100 @@ def baz():
...
+
+            radix_tree = sdict["radix_tree"]
+            self.assertEqual(radix_tree["key"], -1)
+            self.assertEqual(radix_tree["name"], "")
+            self.assertEqual(radix_tree["call_count"], 0)
+            self.assertEqual(radix_tree["total_time"], 0)
+            self.assertEqual(radix_tree["children"], [])
+
+        modmesh.call_profiler.reset()
+        foo()
```
ThreeMonth03, line 381: In this test, the serialization result should be empty because the root function `foo()` hasn't exited.

tests/test_callprofiler.py
```diff
@@ -284,3 +284,100 @@ def baz():
...
+
+        @profile_function
+        def bar():
+            busy_loop(0.5)
+
+        @profile_function
+        def foo():
+            bar()
+            busy_loop(0.1)
+            sdict = json.loads(modmesh.call_profiler.serialize())
```
ThreeMonth03, line 366: Fix the excessively long variable naming.

cpp/modmesh/toggle/pymod/wrap_profile.cpp
```diff
@@ -191,8 +192,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapCallProfiler : public WrapBase<WrapC
                     return ss.str();
                 })
             .def("result", &result)
```
ThreeMonth03, line 194: I'm not planning to delete function `result` in this pr.

cpp/modmesh/toggle/RadixTree.cpp
```diff
@@ -47,6 +47,30 @@ void CallProfiler::reset()
...
+    call_profile.stop_stopwatch(); // Update profiling information to the pending time and count
+    m_pending_nodes.insert(m_radix_tree.get_current_node());
+    m_radix_tree.move_current_to_parent(); // Pop the caller from the call stack
+
+    if (m_radix_tree.get_current_node() == m_radix_tree.get_root()) // If the root function ends, update all pending nodes and stable items
+    {
+        update_pending_nodes();
+        m_radix_tree.update_stable_items();
+    }
+}
```
ThreeMonth03, line 74: To address the re-entrancy issue, I added some additional members to the profiler. These members are updated only when the root function completes or when the profiler resets. As a result, if the profiler is serialized during profiling, the recorded data will reflect the state before the root function's execution was measured.

cpp/modmesh/toggle/SerializableProfiler.hpp
```diff
@@ -0,0 +1,137 @@
...
+    SerializableRadixTree() = default;
+    SerializableRadixTree(SerializableRadixTree const &) = default;
+    SerializableRadixTree(SerializableRadixTree &&) = default;
+    SerializableRadixTree & operator=(SerializableRadixTree const &) = default;
+    SerializableRadixTree & operator=(SerializableRadixTree &&) = default;
+    ~SerializableRadixTree() = default;
+    SerializableRadixTree(const RadixTree<CallerProfile> & radix_tree)
+        : m_root(radix_tree.get_root())
+        , m_id_map(radix_tree.get_stable_id_map())
+        , m_unique_id(radix_tree.get_stable_unique_node())
```
ThreeMonth03, line 102: The members are re-entrant.

cpp/modmesh/toggle/SerializableProfiler.hpp
```diff
@@ -0,0 +1,137 @@
...
+    ~SerializableRadixTreeNode() = default;
+    SerializableRadixTreeNode(const RadixTreeNode<CallerProfile> * node)
+        : m_key(node->key())
+        , m_name(node->name())
+        , m_total_time(node->data().stable_total_time)
+        , m_call_count(node->data().stable_call_count)
+    {
+        for (const auto & child : node->children())
+        {
+            if (child->data().stable_call_count > 0)
```
ThreeMonth03, line 59: The node is valid only if the root function has completed at least once.

ThreeMonth03 (commented):
- [x] Fix the excessively long variable naming.
- [x] Address the re-entrancy issue.

ThreeMonth03:
> * [x] Fix the excessively long variable naming.
> * [x] Address the re-entrancy issue.

@yungyuc You could review the code if you are available. Thank you very much.

cpp/modmesh/toggle/RadixTree.cpp
yungyuc: Could you please summarize the variables you added for the reentrance support, and how it works, under this review comment?

Once it is clarified, we discuss whether or not the conclusion should be added in code comments.

yungyuc: https://github.com/solvcon/modmesh/issues/320#issuecomment-2282744365 is an example for how to summarize the code mechanism.

cpp/modmesh/toggle/SerializableProfiler.hpp
yungyuc: Could you please elaborate more on reentrance?  What members are related to reentrance support and how they work for it?

tests/test_callprofiler.py
yungyuc: This is nice.

cpp/modmesh/toggle/SerializableProfiler.hpp
tigercosmos: How about putting the state in the serialization result? such as `"status: "processing"`, `"status": "done"`.

cpp/modmesh/toggle/pymod/wrap_profile.cpp
tigercosmos: You meant there will be a followup cleanup PR?

ThreeMonth03:
> * [x] Fix the excessively long variable naming.
> * [x] Address the re-entrancy issue.

By the way, there exists the re-entrancy problem when deserializating the json. For example:

```python
def deserialize_during_profiling(self):
        @profile_function
        def bar():
             modmesh.call_profiler.deserialize(json_str)
             busy_loop(0.1)

        @profile_function
        def foo():
            bar()
            busy_loop(0.1)

        foo()
```

Maybe we could make the upper case act like the following case: 
```python
def deserialize_during_profiling(self):
        @profile_function
        def bar():
             busy_loop(0.1)

        @profile_function
        def foo():
            bar()
            busy_loop(0.1)

        foo()
        modmesh.call_profiler.deserialize(json_str)
```
In other words, we could postpone the deserialization instruction until the measurement of the root function ends. How do you think?

yungyuc (commented):
> > * [x]  Fix the excessively long variable naming.
> > * [x]  Address the re-entrancy issue.
> 
> @yungyuc You could review the code if you are available. Thank you very much.

It is wonderful that you are making the reentrance work!

----

Reentrance support is important for the profiler.  We probably should spend a little bit more time to discuss the implementation, and discuss how to document it (the possible forms of the documentation are review comments and code comments).

- [x] Discuss the reentrance support.

yungyuc (approved):

yungyuc:
> In other words, we could postpone the deserialization instruction until the measurement of the root function ends. How do you think?

The reentrance issue with the deserialization is a good finding, and we may discuss it along with the summary of the reentrance support at https://github.com/solvcon/modmesh/pull/397#discussion_r1713108125 . 

I agree we can work on the reentrance support for the deserialization in the future.  In this regard, it is a good thing to add the speculated test cases in the test file but disable (skip) them for now.  Development plan can be added in the code comments in the skipped test functions.

ThreeMonth03:
> > > * [x]   Fix the excessively long variable naming.
> > > * [x]   Address the re-entrancy issue.
> > 
> > 
> > @yungyuc You could review the code if you are available. Thank you very much.
> 
> It is wonderful that you are making the reentrance work!
> 
> Reentrance support is important for the profiler. We probably should spend a little bit more time to discuss the implementation, and discuss how to document it (the possible forms of the documentation are review comments and code comments).
> 
> * [ ]  Discuss the reentrance support.

ThreeMonth03:
[#PR397](https://github.com/solvcon/modmesh/pull/397) rebuild the serializer of the profiler with general serializer, and it makes the serializer support the reentrance.

To support the reentrance, if the profiler is serialized during profiling, I make the serializer return the profiler state before profiling.

For example:

```python
def serialize_during_profiling(self):
        @profile_function
        def bar():
            busy_loop(0.5)

        @profile_function
        def foo():
            bar()
            busy_loop(0.1)
            sdict = json.loads(modmesh.call_profiler.serialize())
        
        modmesh.call_profiler.reset()        
        foo()
```
Because the `foo()` hasn't completed before serialization, the `sdict` should be same as the following `sdict`.
```python
def serialize_during_profiling(self):
        @profile_function
        def bar():
            busy_loop(0.5)

        @profile_function
        def foo():
            bar()
            busy_loop(0.1)

        modmesh.call_profiler.reset()            
        sdict = json.loads(modmesh.call_profiler.serialize())
        foo()
```

In order to achieve the trait, it is necessary to create additional members to record the profiler information before profiling, such as `m_stable_id_map`, `m_stable_unique_id`, `stable_total_time` and `stable_call_count`.

https://github.com/solvcon/modmesh/blob/ce352bd4ba1102908587c6033fadf30590d5e05c/cpp/modmesh/toggle/RadixTree.hpp#L201-L204

https://github.com/solvcon/modmesh/blob/ce352bd4ba1102908587c6033fadf30590d5e05c/cpp/modmesh/toggle/RadixTree.hpp#L232-L235

By the way, to trace and update the `stable_total_time` and `stable_call_count` in some nodes handy, I create a set `m_pending_nodes` that records the nodes which haven't updated the `stable_total_time` and `stable_call_count`.
https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L313
`m_pending_nodes` is only updated when a function ends.
https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.cpp#L66-L66

`m_stable_id_map`, `m_stable_unique_id`, `stable_total_time` and `stable_call_count` are updated only when
1. The profiler resets.
https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.cpp#L46-L48
https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L167-L170

2. The root function ends.
https://github.com/solvcon/modmesh/blob/ce352bd4ba1102908587c6033fadf30590d5e05c/cpp/modmesh/toggle/RadixTree.cpp#L68-L73

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L301-L308

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L224-L229

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L178-L182

Because `m_stable_id_map`, `m_stable_unique_id`, `stable_total_time` and `stable_call_count` are updated when the root function ends, the serializer could work regularly during profiling.

Last but not least, the general serializer should read the `m_stable_id_map`, `m_stable_unique_id`, `stable_total_time` and `stable_call_count`.

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/SerializableProfiler.hpp#L101-L102

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/RadixTree.hpp#L186-L189

https://github.com/solvcon/modmesh/blob/26f0f911745cd20933d49ce4a7e2221d9178d337/cpp/modmesh/toggle/SerializableProfiler.hpp#L54-L55

cpp/modmesh/toggle/SerializableProfiler.hpp
ThreeMonth03: > How about putting the state in the serialization result? such as `"status: "processing"`, `"status": "done"`.

While this is certainly doable, I'm unsure if it's necessary. When deserializing the result, the `status` may not be particularly useful since the profiler should already reflect the state before or after processing. However, including the `status` in the serialization result could make it easier to monitor the serializer.

cpp/modmesh/toggle/pymod/wrap_profile.cpp
ThreeMonth03: Yes, there will be a followup cleanup PR. The function `result()` would be replaced with the function `viewable_result()`, which is implemented by `serialize()`.

cpp/modmesh/toggle/SerializableProfiler.hpp
yungyuc: We do not need to make a decision now.  It can be done if a use case surfaces.  That is, we do not need to do anything now.

yungyuc:
> [#PR397](https://github.com/solvcon/modmesh/pull/397) rebuild the serializer of the profiler with general serializer, and it makes the serializer support the reentrance.
> 
> To support the reentrance, if the profiler is serialized during profiling, I make the serializer return the profiler state before profiling.

It is detailed explanation.  I think a more concise description will be easier for a new maintainer, but it is certainly good to show everything.

I am not very sure it is good to keep a pair of stable and (seemingly) unstable variables:

```cpp
 std::unordered_map<std::string, key_type> m_id_map; 
 std::unordered_map<std::string, key_type> m_stable_id_map; // Re-entrant safe 
 key_type m_unique_id = 0; 
 key_type m_stable_unique_id = 0; // Re-entrant safe 
 // ...
 std::chrono::nanoseconds total_time = std::chrono::nanoseconds(0); /// use nanoseconds to have higher precision 
 std::chrono::nanoseconds stable_total_time = std::chrono::nanoseconds(0); // Re-entrant safe 
 int call_count = 0; 
 int stable_call_count = 0; // Re-entrant safe 
```

The concern of redundancy should be addressed in the future.  Let's keep it for now.  The profiler has not been well exercised.  Even if it is not using the best implementation, we should review it after exercising it more.

Another concern is the use of STL hash table (`std::unordered_map`) and tree (`std::set`) extensively, e.g.,

```cpp
 std::set<RadixTreeNode<CallerProfile> *> m_pending_nodes; /// The nodes that are not yet updated 
```

STL uses general but slow implementation.  A profiler is supposed to be little overhead and fast.  We will need to examine later when putting the profiler into real use.

But this is good.  Everything necessary is clarified.  I am merging it.

PR #402, opened 2024-08-03
Use the repository variable MMGH_NIGHTLY to control scheduled Github Actions run

### LINKED ISSUE #286: Enhance the CI runs with secrets
In a forked repository the Github Actions CI also performs nightly runs.  It is usually undesirable because the duplicated runs may be noisy to a contributor, and they waste resources for no purpose.

Maybe we can use a simple [secret](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) check in the nightly jobs.  The upstream has the secret turned on.  The forked does not have it (by default) and the nightly does not run.  If a contributor wants it in the forked repository, they can add it manually.

yungyuc:
If the repository variable `MMGH_NIGHTLY` is set to `enable`, the Github Actions will run at the scheduled time point.  Otherwise the scheduled runs will not take place.

This PR is to implement the enhancement requested in issue #286.

References:
* Repository configuration variables: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/variables#defining-configuration-variables-for-multiple-workflows
* Access variables from actions: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#vars-context

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -10,7 +10,7 @@ jobs:
 
   standalone_buffer:
 
-    if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
+    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
 
     name: standalone_buffer_${{ matrix.os }}
 
@@ -79,7 +79,7 @@ jobs:
 
   build_ubuntu:
 
-    if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
+    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
 
     name: build_${{ matrix.os }}_${{ matrix.cmake_build_type }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -10,7 +10,7 @@ jobs:
 
   standalone_buffer:
 
-    if: ${{ github.event_name != '' || (github.event_name == '' && github.repository_owner == 'solvcon') }}
+    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
```
tigercosmos, line 13: Could you explain to me when you'd like to build the nightly? I don't understand the logic here.

yungyuc: My apologies for missing the reference to the issue #286.  The fencing was updated so that a forked repository will not run the workflows nightly.

Without the fencing, a forked repository will waste a lot of Github Actions cycles to duplicate the scheduled nightly workflow runs in the upstream repository.  To avoid the waste, the forked repository should turn off the Github Actions, and thus does not execute any workflow when a developer pushes to it.

With the fencing, the nightly runs are skipped unless the repository variable `MMGH_NIGHTLY` is set to `enable` in the forked repository.  The variable has been set in the upstream repository https://github.com/solvcon/modmesh.

tigercosmos: @yungyuc  But I wonder why it also builds when the event is not "schedule"? I think it should be built only once a day.

yungyuc: `schedule` is just one of the many values of [`github.event_name`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#github-context): https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows .  When the triggering event is not `schedule`, the workflow should just run.

tigercosmos: Okay, I think I got your intention now.

PR #414, opened 2024-08-19
GUI configuration supporting string based access

j8xixo12:
I found that the content of the table in `SolverConfig` and `PlotConfig` is only used internally within the class, and `QtTableView` interacts with the data solely through data and setData methods. Users will only access the data in a string-based manner. In this PR, I have only implemented the string-based accessing part with `__getitem__`

### DIFF
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -104,7 +104,41 @@ class QuantityLine:
         self.num.figure.canvas.draw()
 
 
-class GUIConfig:
+class _Accessor:
+    """
+    Helper calss to access data within the configuration table using
+    multiple dimensions, currenlty only support 2-dimensions table.
+
+    This class allows users to access data by chaining multiple [],
+    supporting string-based.
+
+    Attributes:
+        :ivar: _data: handle input data that allow user can using
+            index or string to access it.
+        :ivar: _header (list): list of data's column header name.
+        :ivar: _dimIdx (int): using to indicate which dimension is
+            currenlty being accessed by __getitem__.
+    """
+    def __init__(self, data, dimIdx=0, header=None):
+        self._data = data
+        self._header = header
+        self._dimIdx = dimIdx
+
+    def __getitem__(self, key):
+        if self._dimIdx == 0:
+            for row_idx, row in enumerate(self._data):
+                if key == row[0]:
+                    return _Accessor(self._data[row_idx],
+                                     self._dimIdx+1,
+                                     self._header)
+        else:
+            for idx, ele in enumerate(self._header):
+                if key == ele:
+                    return self._data[idx]
+        raise KeyError(f'Invaild key: {key} not found in table')
+
+
+class GUIConfig(object):
     """
     Configuration class for the GUI.
 
@@ -112,52 +146,27 @@ class GUIConfig:
     users to set and retrieve parameters related to the simulation.
 
     Attributes:
-        - `state` (:class:`State`): The state object holding configuration
-          data.
-        - `_tbl_content` (:class:`State`): The content of the
-          configuration table.
-        - `_col_header` (list): The header for the configuration
-          table columns.
+        :ivar: state (:class:`State`): The state object holding configuration
+            data.
+        :ivar: _tbl_content (:class:`State`): The content of the
+            configuration table.
+        :ivar: _col_header (list): The header for the configuration
+            table columns.
 
     Methods:
-        - :meth:`data(row, col)`: Get the value at a specific row and column
-          in the configuration table.
-        - :meth:`setData(row, col, value)`: Set the value at a specific row
-          and column in the configuration table.
-        - :meth:`columnHeader(col)`: Get the header for a specific column
-          in the configuration table.
-        - :meth:`editable(row, col)`: Check if a cell in the configuration
-          table is editable.
-        - :meth:`rowCount()`: Get the number of rows in the configuration
-          table.
-        - :meth:`columnCount()`: Get the number of columns in the
-          configuration table.
+        :meth:`data(row, col)`: Get the value at a specific row and column
+            in the configuration table.
+        :meth:`setData(row, col, value)`: Set the value at a specific row
+            and column in the configuration table.
+        :meth:`columnHeader(col)`: Get the header for a specific column
+            in the configuration table.
+        :meth:`editable(row, col)`: Check if a cell in the configuration
+            table is editable.
+        :meth:`rowCount()`: Get the number of rows in the configuration
+            table.
+        :meth:`columnCount()`: Get the number of columns in the
+            configuration table.
     """
-    class Accessor:
-        """
-        Helper calss to access data within the configuration table using
-        multiple dimensions, currenlty only support 2-dimensions table.
-
-        This class allows users to access data by chaining multiple [],
-        supporting string-based.
-        """
-        def __init__(self, data, dimIdx=0, header=None):
-            self._data = data
-            self._header = header
-            self._dimIdx = dimIdx
-
-        def __getitem__(self, key):
-            if self._dimIdx == 0:
-                for row_idx, row in enumerate(self._data):
-                    if key == row[0]:
-                        return GUIConfig.Accessor(self._data[row_idx],
-                                                  self._dimIdx+1,
-                                                  self._header)
-            else:
-                for idx, ele in enumerate(self._header):
-                    if key == ele:
-                        return self._data[idx]
-            raise KeyError(f'Invaild key: {key} not found in table')
 
     def __init__(self, input_data, col_headers):
         self.state = State(input_data)
@@ -165,7 +174,7 @@ class GUIConfig:
         self._col_header = col_headers
 
     def __getitem__(self, key):
-        return self.Accessor(self._tbl_content, 0, self._col_header)[key]
+        return _Accessor(self._tbl_content, 0, self._col_header)[key]
 
     def data(self, row, col):
         """
--- a/modmesh/app/euler1d.py
+++ b/modmesh/app/euler1d.py
@@ -104,16 +104,16 @@ class QuantityLine:
         self.num.figure.canvas.draw()
 
 
-class SolverConfig():
+class GUIConfig:
     """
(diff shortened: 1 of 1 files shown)


modmesh/app/euler1d.py
```diff
@@ -104,16 +104,16 @@ def update(self, adata, ndata):
         self.num.figure.canvas.draw()
 
 
-class SolverConfig():
+class GUIConfig:
```
j8xixo12: Considering that `SolverConfig` and `PlotConfig` have a lot of similar code, I created a parent class to implement the shared parts.

modmesh/app/euler1d.py
```diff
@@ -317,42 +292,6 @@ def editable(self, row, col):
             return True
         return False
```
j8xixo12, line 293: Since `PlotConfig` table has multiple columns allowing user to modified it's value via `QtTableView`. Therefore I overwrite the `editable` method.

modmesh/app/euler1d.py
```diff
@@ -132,13 +132,40 @@ class SolverConfig():
...
+        This class allows users to access data by chaining multiple [],
+        supporting string-based.
+        """
+        def __init__(self, data, dimIdx=0, header=None):
+            self._data = data
+            self._header = header
+            self._dimIdx = dimIdx
+
+        def __getitem__(self, key):
+            if self._dimIdx == 0:
```
j8xixo12: This is used to identify whether the `[]` operator is currently handling row accessing, because row access requires checking the row name in the first column.

modmesh/app/euler1d.py
```diff
@@ -242,66 +271,12 @@ class PlotConfig():
...
-        :type col: int
-        :return: The header for the specific column.
-        """
-        return self._col_header[col]
+        super().__init__(input_data, ["variable",
+                                      "line_selection",
+                                      "y_axis_upper_limit",
+                                      "y_axis_bottom_limit"])
 
     def editable(self, row, col):
```
j8xixo12, line 290: Since `PlotConfig` table has multiple columns allowing user to modified its value via `QtTableView`. Therefore I overwrite the editable method.

modmesh/app/euler1d.py
```diff
@@ -132,13 +132,40 @@ class SolverConfig():
...
+            raise KeyError(f'Invaild key: {key} not found in table')
+
+    def __init__(self, input_data, col_headers):
         self.state = State(input_data)
         self._tbl_content = self.state
-        self._col_header = ["variable", "value", "description"]
+        self._col_header = col_headers
+
+    def __getitem__(self, key):
+        return self.Accessor(self._tbl_content, 0, self._col_header)[key]
```
tigercosmos: Seems that the logic is simple enough to be a function. I am not sure if using a function can reduce the overhead of creating a class?

modmesh/app/euler1d.py
```diff
@@ -493,24 +432,21 @@ def set_solver_config(self):
...
+                         xmin=self.solver_config["xmin"]["value"],
+                         xmax=self.solver_config["xmax"]["value"],
+                         ncoord=self.solver_config["ncoord"]["value"],
+                         time_increment=(self.solver_config["time_increment"][
+                                                            "value"]))
         self.current_step = 0
-        self.interval = self.solver_config.get_var("timer_interval", "value")
-        self.max_steps = self.solver_config.get_var("max_steps", "value")
-        self.profiling = self.solver_config.get_var("profiling", "value")
+        self.interval = self.solver_config["timer_interval"]["value"]
```
tigercosmos, line 456: The original `get_var` handles the case if the key doesn't exist.
Perhaps write it as  `self.solver_config.get("timer_interval", {}). get("value", None)`

modmesh/app/euler1d.py
yungyuc: I would hope we keep a minimal Python 2/3 compatibility for classes (although I don't think anyone would use Python 2).  That is to make a class derived from `object`:

```python
class GUIConfig(object):
```

In Python 3 it is the same as writing `class GUIConfig:`.  But in Python 2, `class GUIConfig:` is the old-style (pre-2.2) class, and incompatible to the new-style class (2.3 and beyond).

Having the minimal compatibility makes it slight easier for people to copy code to Python 2.

modmesh/app/euler1d.py
```diff
@@ -104,16 +104,16 @@ def update(self, adata, ndata):
...
+class GUIConfig:
     """
-    Configuration class for the solver.
+    Configuration class for the GUI.
 
-    This class provides a configuration interface for the solver, allowing
+    This class provides a configuration interface for the GUI, allowing
     users to set and retrieve parameters related to the simulation.
 
     Attributes:
```
yungyuc, line 148: Please use Sphinx format for the attributes in the docstring.

modmesh/app/euler1d.py
```diff
@@ -132,13 +132,40 @@ class SolverConfig():
           table.
         - :meth:`columnCount()`: Get the number of columns in the
           configuration table.
-        - :meth:`get_var(key_row, key_col)`: Get the value of a configuration
-          variable based on its key.
     """
-    def __init__(self, input_data):
+    class Accessor:
```
yungyuc: Why do you use an inner class?  It's less testable.  What about moving it outside in the module?

modmesh/app/euler1d.py
yungyuc: I have the same question.  But I care more about the maintainability.  Both Python functions and classes are slow.  And I don't think we plan to profile for it any time soon.

yungyuc (approved):
It's significantly simpler.  I don't mind merging it now.  @j8xixo12, please let me know if you want to discuss before merging, or the other way around.

- [x] Discuss whether `Accessor` should be moved outside from inner?

modmesh/app/euler1d.py
j8xixo12: But the `get_var` way was already quite wordy, and the new way seems to make it even wordy.

yungyuc: API conciseness is hard, and that is why I developed the toggle code:
 https://github.com/solvcon/modmesh/blob/194a313d8bd6c898d959841d8288f69ac85c0692/tests/test_toggle.py#L258

It may take some efforts to connect the toggle library with PUI, so I am not promoting it now.  But we will need it at some point.

yungyuc: At this point, I'd say `__getitem__()` may be a a reasonable balance.

modmesh/app/euler1d.py
j8xixo12: My original idea is to achieve `__getitem__` chaining by returning an `Accessor` instance and `Accessor` is specialize for accessing `GUIConfig` data. That's why I put this class inside `GUIConfig`.

If we move `Accessor` outside, it would imply that `Accessor` needs to become a more generic class.
I will try to re-design `Accessor` class.

yungyuc: Being specific or general has nothing to do with making it an inner class or a normal (in-module) class.  When making a help class for a specific use inside another class, a common style is to mark it as a private class like:

```python
class _Accessor:
    ....
class Consumer:
    def func(self):
        a = _Accessor(...)
        a.helper_function()
```

yungyuc:
> @j8xixo12, please let me know if you want to discuss before merging, or the other way around.

@j8xixo12, could you please let me know which way do you want to go?  If you cannot make your mind now, don't hesitate to just say you need some time to think about it.

j8xixo12:
> > @j8xixo12, please let me know if you want to discuss before merging, or the other way around.
> 
> @j8xixo12, could you please let me know which way do you want to go? If you cannot make your mind now, don't hesitate to just say you need some time to think about it.

I would like to discuss more before merging, and https://github.com/solvcon/modmesh/pull/414#discussion_r1724145035 seems like a good way to proceed.

And I also found that the way of accessing data using indexes and strings seems very similar to a dataframe, and it looks like we could use a dataframe to store table contents in the future.

modmesh/app/euler1d.py
j8xixo12: I have fixed it in https://github.com/solvcon/modmesh/pull/414/commits/d25d5c2e7a37da1a74cd142ebc39ee0bf43b76bd.

j8xixo12:
In latest patch I have completed the following items:
- [x] `GUIConfig` inherit from `object` for python2 compatible.
- [x] Fix docstring style in `GUIConfig`
- [x] Move `Accessor` out of `GUIConfig` and mark it as a private class.

modmesh/app/euler1d.py
```diff
@@ -104,41 +104,77 @@ def update(self, adata, ndata):
         self.num.figure.canvas.draw()
 
 
-class SolverConfig():
+class _Accessor:
```
yungyuc: Every class (added/changed in the PR) should follow the Python 2/3 compatible declaration (inherit from `object`).

yungyuc (changes requested):
> And I also found that the way of accessing data using indexes and strings seems very similar to a dataframe, and it looks like we could use a dataframe to store table contents in the future.

If you have hundreds of thousands of parameters to manage, a data frame is a good container.  Otherwise, it hurts more than gains.

But we can use more improvement of the GUI code after this PR.

----

- [x] (Since you decide to do it), make every class you added/changed to be Python 2/3 compatible.

modmesh/app/euler1d.py
j8xixo12: Done!

j8xixo12:
> > And I also found that the way of accessing data using indexes and strings seems very similar to a dataframe, and it looks like we could use a dataframe to store table contents in the future.
> 
> If you have hundreds of thousands of parameters to manage, a data frame is a good container. Otherwise, it hurts more than gains.
> 
> But we can use more improvement of the GUI code after this PR.
> 
> * [ ]  (Since you decide to do it), make every class you added/changed to be Python 2/3 compatible.

I have fixed it.
* [x]  (Since you decide to do it), make every class you added/changed to be Python 2/3 compatible.

yungyuc:
LGTM.  Thanks.

PR #417, opened 2024-08-23
Edit style: always add curly braces

yungyuc:
Always add curly braces and always add them in standalone lines:

```cpp
  if (condition)
  {
      return;
  }
```

That is, never drop curly braces even when you can:

```cpp
  // NEVER DROP CURLY BRACES
  if (condition)
      return;
```

### DIFF
--- a/STYLE.rst
+++ b/STYLE.rst
@@ -440,6 +440,26 @@ hundreds of lines) to keep track of.
 
   } /* end namespace modmesh */
 
+C++ Curly Braces
+====
+
+Always add curly braces and always add them in standalone lines:
+
+.. code-block:: cpp
+
+  if (condition)
+  {
+      return;
+  }
+
+That is, never drop curly braces even when you can:
+
+.. code-block:: cpp
+
+  // NEVER DROP CURLY BRACES
+  if (condition)
+      return;
+
 Copyright Notice
 ====
 


yungyuc:
Style enforcement like this is always controversial.  This will be a good topic for face-to-face discussions.  That is, online discussions do not work very well for this, unless we are asking for flame wars.

If contributors need clarification for this style, feel free to add your arguments, and I will try to address.

tigercosmos:
I believe this is one standard of the clean code.

ThreeMonth03:
> Style enforcement like this is always controversial. This will be a good topic for face-to-face discussions. That is, online discussions do not work very well for this, unless we are asking for flame wars.
> 
> If contributors need clarification for this style, feel free to add your arguments, and I will try to address.

I don't mind either style, but I'm curious why we prefer to add curly braces even when the code has only one line.

yungyuc:
> > Style enforcement like this is always controversial. This will be a good topic for face-to-face discussions. That is, online discussions do not work very well for this, unless we are asking for flame wars.
> > If contributors need clarification for this style, feel free to add your arguments, and I will try to address.
> 
> I don't mind either style, but I'm curious why we prefer to add curly braces even when the code has only one line.

A concrete example is that the curly braces make comments more readable:

```cpp
if (condition)
{
    /* Comments that take at least one line.
       Sometimes two or more. */
    return;
}
```

It looks confusing without the curly braces:

```cpp
if (condition)
    /* Comments that take at least one line.
       Sometimes two or more. */
    return;
```

In addition, [clang-tidy checks for it](https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html):

```
...: error: statement should be inside braces
[google-readability-braces-around-statements,
readability-braces-around-statements,
-warnings-as-errors]
        if (condition)
                     ^
```

PR #418, opened 2024-08-23
Camera Controller: expose complete API to python

### LINKED ISSUE #70: Improve camera control

Tucchhaa:
In this PR:
* New props added to camera controller to control it's reset behaviour: defaultPosition, defaultUpVector, defaultViewCenter, defaultLookSpeed, defaultLinearSpeed
* All these props are exposed to python
* Camera controller python API was rewritten to use `def_property` using macros. Tests also were updated
* Minor issue was fixed: When reset camera, aspect ratio was incorrect

### DIFF
--- a/cpp/modmesh/view/R3DWidget.cpp
+++ b/cpp/modmesh/view/R3DWidget.cpp
@@ -41,7 +41,8 @@ R3DWidget::R3DWidget(Qt3DExtras::Qt3DWindow * window, RScene * scene, QWidget * 
 {
     m_view->setRootEntity(m_scene);
 
-    resetCamera();
+    cameraController()->setCamera(m_view->camera());
+    cameraController()->reset();
 
     if (Toggle::instance().fixed().get_show_axis())
     {
@@ -97,23 +98,6 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
     QWidget::resizeEvent(event);
     m_view->resize(event->size());
     m_container->resize(event->size());
-}
-
-void R3DWidget::resetCamera() const
-{
-    Qt3DRender::QCamera * camera = m_view->camera();
-
-    // Set up the camera.
-    camera->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
-    camera->setPosition(QVector3D(0.0f, 0.0f, 10.0f));
-    camera->setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
-    camera->setUpVector(QVector3D(0.f, 1.f, 0.f));
-
-    // Set up the camera control.
-    auto * control = m_scene->controller();
-    control->setCamera(camera);
-    control->setLinearSpeed(50.0f);
-    control->setLookSpeed(180.0f);
 }
 
 } /* end namespace modmesh */
--- a/cpp/modmesh/view/R3DWidget.hpp
+++ b/cpp/modmesh/view/R3DWidget.hpp
@@ -59,7 +59,8 @@ public:
     CameraController * controller() const { return m_controller; }
     Qt3DExtras::QAbstractCameraController * qtController() const { return m_controller->asQtCameraController(); }
 
-    void setCameraController(CameraController * controller) {
+    void setCameraController(CameraController * controller)
+    {
         qtController()->deleteLater();
         m_controller = controller;
     }
@@ -70,7 +71,7 @@ public:
 
 private:
 
-    CameraController* m_controller;
+    CameraController * m_controller;
 
 }; /* end class RScene */
 
--- a/cpp/modmesh/view/R3DWidget.hpp
+++ b/cpp/modmesh/view/R3DWidget.hpp
@@ -46,7 +46,8 @@
 namespace modmesh
 {
 
-class RScene : public Qt3DCore::QEntity
+class RScene
+    : public Qt3DCore::QEntity
 {
 public:
 
@@ -57,11 +58,10 @@ public:
     }
 
     CameraController * controller() const { return m_controller; }
-    Qt3DExtras::QAbstractCameraController * qtController() const { return m_controller->asQtCameraController(); }
 
     void setCameraController(CameraController * controller)
     {
-        qtController()->deleteLater();
+        m_controller->deleteLater();
         m_controller = controller;
     }
 
@@ -97,7 +97,6 @@ public:
     Qt3DRender::QCamera * camera() { return m_view->camera(); }
 
     CameraController * cameraController() const { return m_scene->controller(); }
-    Qt3DExtras::QAbstractCameraController * qtCameraController() const { return m_scene->qtController(); }
 
     QPixmap grabPixmap() const { return m_view->screen()->grabWindow(m_view->winId()); }
 
--- a/cpp/modmesh/view/R3DWidget.hpp
+++ b/cpp/modmesh/view/R3DWidget.hpp
@@ -46,24 +46,21 @@
 namespace modmesh
 {
 
-class RScene
-    : public Qt3DCore::QEntity
+class RScene : public Qt3DCore::QEntity
 {
-
 public:
 
-    explicit RScene(Qt3DCore::QNode * parent = nullptr)
-        : Qt3DCore::QEntity(parent)
-        , m_controller(new ROrbitCameraController(this))
+    explicit RScene(QNode * parent = nullptr)
+        : QEntity(parent)
     {
+        m_controller = new ROrbitCameraController(this);
     }
 
-    Qt3DExtras::QAbstractCameraController const * controller() const { return m_controller; }
-    Qt3DExtras::QAbstractCameraController * controller() { return m_controller; }
+    CameraController * controller() const { return m_controller; }
+    Qt3DExtras::QAbstractCameraController * qtController() const { return m_controller->asQtCameraController(); }
 
-    void setCameraController(Qt3DExtras::QAbstractCameraController * controller)
-    {
-        m_controller->deleteLater();
+    void setCameraController(CameraController * controller) {
+        qtController()->deleteLater();
         m_controller = controller;
     }
 
@@ -73,7 +70,7 @@ public:
 
 private:
 
-    Qt3DExtras::QAbstractCameraController * m_controller = nullptr;
+    CameraController* m_controller;
 
 }; /* end class RScene */
 
@@ -97,8 +94,9 @@ public:
     Qt3DExtras::Qt3DWindow * view() { return m_view; }
     RScene * scene() { return m_scene; }
     Qt3DRender::QCamera * camera() { return m_view->camera(); }
-    Qt3DExtras::QAbstractCameraController * qtCameraController() { return m_scene->controller(); }
-    CameraController * cameraController() { return dynamic_cast<CameraController *>(m_scene->controller()); }
+
+    CameraController * cameraController() const { return m_scene->controller(); }
+    Qt3DExtras::QAbstractCameraController * qtCameraController() const { return m_scene->qtController(); }
 
     QPixmap grabPixmap() const { return m_view->screen()->grabWindow(m_view->winId()); }
 
--- a/cpp/modmesh/view/R3DWidget.hpp
+++ b/cpp/modmesh/view/R3DWidget.hpp
@@ -57,9 +57,9 @@ public:
         m_controller = new ROrbitCameraController(this);
     }
 
-    CameraController * controller() const { return m_controller; }
+    RCameraController * controller() const { return m_controller; }
 
-    void setCameraController(CameraController * controller)
+    void setCameraController(RCameraController * controller)
     {
         m_controller->deleteLater();
         m_controller = controller;
@@ -71,7 +71,7 @@ public:
 
 private:
 
-    CameraController * m_controller;
+    RCameraController * m_controller;
 
 }; /* end class RScene */
 
@@ -96,7 +96,7 @@ public:
     RScene * scene() { return m_scene; }
     Qt3DRender::QCamera * camera() { return m_view->camera(); }
 
-    CameraController * cameraController() const { return m_scene->controller(); }
+    RCameraController * cameraController() const { return m_scene->controller(); }
 
     QPixmap grabPixmap() const { return m_view->screen()->grabWindow(m_view->winId()); }
 
--- a/cpp/modmesh/view/R3DWidget.hpp
+++ b/cpp/modmesh/view/R3DWidget.hpp
@@ -100,8 +100,6 @@ public:
     Qt3DExtras::QAbstractCameraController * qtCameraController() { return m_scene->controller(); }
     CameraController * cameraController() { return dynamic_cast<CameraController *>(m_scene->controller()); }
 
-    void resetCamera() const;
-
     QPixmap grabPixmap() const { return m_view->screen()->grabWindow(m_view->winId()); }
 
     void showMark();
--- a/cpp/modmesh/view/RCameraController.cpp
+++ b/cpp/modmesh/view/RCameraController.cpp
@@ -214,7 +214,10 @@ void RCameraInputListener::initKeyboardListeners() const
 
 void CameraController::reset()
 {
-    camera()->lens()->setPerspectiveProjection(45.0f, camera()->lens()->aspectRatio(), 0.1f, 1000.0f);
+    constexpr float fov_deg = 45.0f;
+    constexpr float nearPlane = 0.1f;
+    constexpr float farPlane = 1000.0f;
+    camera()->lens()->setPerspectiveProjection(fov_deg, camera()->lens()->aspectRatio(), nearPlane, farPlane);
 
     setPosition(m_default_position);
     setViewCenter(m_default_view_center);
@@ -225,17 +228,17 @@ void CameraController::reset()
 }
 
 RFirstPersonCameraController::RFirstPersonCameraController(QNode * parent)
-    : QFirstPersonCameraController(parent)
+    : CameraController(parent)
 {
     auto callback = [this](const CameraInputState & state, const float dt)
     {
-        updateCameraPosition(state, dt);
+        moveCamera(state, dt);
     };
 
     m_listener = new RCameraInputListener(keyboardDevice(), mouseDevice(), callback, this);
 }
 
-void RFirstPersonCameraController::updateCameraPosition(const CameraInputState & input, const float dt)
+void RFirstPersonCameraController::moveCamera(const CameraInputState & input, const float dt)
 {
     constexpr auto positiveY = QVector3D(0.f, 1.f, 0.f);
 
@@ -256,17 +259,17 @@ void RFirstPersonCameraController::updateCameraPosition(const CameraInputState &
 }
 
 ROrbitCameraController::ROrbitCameraController(QNode * parent)
-    : QOrbitCameraController(parent)
+    : CameraController(parent)
 {
     auto callback = [this](const CameraInputState & state, const float dt)
     {
-        updateCameraPosition(state, dt);
+        moveCamera(state, dt);
     };
 
     m_listener = new RCameraInputListener(keyboardDevice(), mouseDevice(), callback, this);
 }
 
-void ROrbitCameraController::updateCameraPosition(const CameraInputState & input, const float dt)
+void ROrbitCameraController::moveCamera(const CameraInputState & input, const float dt)
 {
     if (camera() == nullptr)
         return;
@@ -317,7 +320,9 @@ void ROrbitCameraController::updateCameraPosition(const CameraInputState & input
 
 void ROrbitCameraController::zoom(const float zoomValue) const
 {
-    const float limitSquared = zoomInLimit() * zoomInLimit();
+    constexpr float zoomInLimit = 2.0f; // Default Qt zoomInLimit value
+
+    const float limitSquared = zoomInLimit * zoomInLimit;
     const float distanceSquared = zoomDistanceSquared(camera()->position(), camera()->viewCenter());
 
     const float z = distanceSquared > limitSquared ? zoomValue : -0.5f;
--- a/cpp/modmesh/view/RCameraController.cpp
+++ b/cpp/modmesh/view/RCameraController.cpp
@@ -216,12 +216,12 @@ void CameraController::reset()
 {
     camera()->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
 
-    setPosition(QVector3D(0.0f, 0.0f, 10.0f));
-    setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
-    setUpVector(QVector3D(0.f, 1.f, 0.f));
+    setPosition(m_default_position);
+    setViewCenter(m_default_view_center);
+    setUpVector(m_default_up_vector);
 
-    setLinearSpeed(50.0f);
-    setLookSpeed(180.0f);
+    setLinearSpeed(m_default_linear_speed);
+    setLookSpeed(m_default_look_speed);
 }
 
 RFirstPersonCameraController::RFirstPersonCameraController(QNode * parent)
--- a/cpp/modmesh/view/RCameraController.cpp
+++ b/cpp/modmesh/view/RCameraController.cpp
@@ -214,7 +214,7 @@ void RCameraInputListener::initKeyboardListeners() const
 
 void CameraController::reset()
 {
-    camera()->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
+    camera()->lens()->setPerspectiveProjection(45.0f, camera()->lens()->aspectRatio(), 0.1f, 1000.0f);
 
     setPosition(m_default_position);
     setViewCenter(m_default_view_center);
--- a/cpp/modmesh/view/RCameraController.cpp
+++ b/cpp/modmesh/view/RCameraController.cpp
@@ -212,6 +212,18 @@ void RCameraInputListener::initKeyboardListeners() const
     m_ty_axis->addInput(m_keyboard_ty_neg_input);
 }
 
+void CameraController::reset()
+{
+    camera()->lens()->setPerspectiveProjection(45.0f, 16.0f / 9.0f, 0.1f, 1000.0f);
+
+    setPosition(QVector3D(0.0f, 0.0f, 10.0f));
+    setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
+    setUpVector(QVector3D(0.f, 1.f, 0.f));
+
+    setLinearSpeed(50.0f);
+    setLookSpeed(180.0f);
+}
+
 RFirstPersonCameraController::RFirstPersonCameraController(QNode * parent)
     : QFirstPersonCameraController(parent)
 {
--- a/cpp/modmesh/view/RCameraController.cpp
+++ b/cpp/modmesh/view/RCameraController.cpp
@@ -212,7 +212,7 @@ void RCameraInputListener::initKeyboardListeners() const
     m_ty_axis->addInput(m_keyboard_ty_neg_input);
 }
 
-void CameraController::reset()
+void RCameraController::reset()
 {
     constexpr float fov_deg = 45.0f;
     constexpr float nearPlane = 0.1f;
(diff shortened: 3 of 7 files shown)


cpp/modmesh/view/R3DWidget.cpp
```diff
@@ -99,23 +100,6 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
...
-    camera->setUpVector(QVector3D(0.f, 1.f, 0.f));
-
-    // Set up the camera control.
-    auto * control = m_scene->controller();
-    control->setCamera(camera);
-    control->setLinearSpeed(50.0f);
-    control->setLookSpeed(180.0f);
-}
-
 } /* end namespace modmesh */
```
Tucchhaa, line 103: resetCamera moved to CameraController

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -73,7 +71,7 @@ class RScene
 
 private:
 
-    Qt3DExtras::QAbstractCameraController * m_controller = nullptr;
+    CameraController * m_controller;
```
Tucchhaa, line 74: small refactoring to `controller()` getters. Logic didn't change, but because our `CameraController` class is used more often than Qt's less `dynamic_cast`'s will happen

cpp/modmesh/view/R3DWidget.cpp
```diff
@@ -99,23 +100,6 @@ void R3DWidget::resizeEvent(QResizeEvent * event)
...
-    camera->setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
-    camera->setUpVector(QVector3D(0.f, 1.f, 0.f));
-
-    // Set up the camera control.
-    auto * control = m_scene->controller();
-    control->setCamera(camera);
-    control->setLinearSpeed(50.0f);
-    control->setLookSpeed(180.0f);
-}
-
```
Tucchhaa, line 118: `resetCamera` moved to `CameraController`

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -46,24 +46,22 @@
...
-    Qt3DExtras::QAbstractCameraController const * controller() const { return m_controller; }
-    Qt3DExtras::QAbstractCameraController * controller() { return m_controller; }
+    CameraController * controller() const { return m_controller; }
+    Qt3DExtras::QAbstractCameraController * qtController() const { return m_controller->asQtCameraController(); }
 
-    void setCameraController(Qt3DExtras::QAbstractCameraController * controller)
+    void setCameraController(CameraController * controller)
     {
-        m_controller->deleteLater();
+        qtController()->deleteLater();
```
Tucchhaa: small refactoring to `controller()` getters. Logic didn't change, but because our `CameraController` class is used more often than Qt's less `dynamic_cast`'s will happen

cpp/modmesh/view/RCameraController.cpp
```diff
@@ -212,6 +212,18 @@ void RCameraInputListener::initKeyboardListeners() const
     m_ty_axis->addInput(m_keyboard_ty_neg_input);
 }
 
+void CameraController::reset()
+{
+    camera()->lens()->setPerspectiveProjection(45.0f, camera()->lens()->aspectRatio(), 0.1f, 1000.0f);
```
Tucchhaa: Now don't reset `aspectRatio`. Because if resize a window and reset, then mesh will look strange. Qt automatically already updates `aspectRatio` when window is resized

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -135,30 +135,59 @@ class CameraController
...
+    void setDefaultViewCenter(QVector3D value) { m_default_view_center = value; }
+
+    QVector3D defaultUpVector() const { return m_default_up_vector; }
+    void setDefaultUpVector(QVector3D value) { m_default_up_vector = value; }
+
+    float defaultLinearSpeed() const { return m_default_linear_speed; }
+    void setDefaultLinearSpeed(float value) { m_default_linear_speed = value; }
+
+    float defaultLookSpeed() const { return m_default_look_speed; }
+    void setDefaultLookSpeed(float value) { m_default_look_speed = value; }
```
Tucchhaa, line 183: new props in `CameraController` to control reset behaviour

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -460,8 +432,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRCameraController
                     bool left_mouse_button,
                     bool right_mouse_button,
                     bool alt_key,
-                    bool shift_key,
-                    float dt)
```
Tucchhaa, line 464: remove `dt` parameter, because no need of it in python.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -196,39 +196,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapR3DWidget
...
-
-        (*this)
-            // clang-format off
-            DECL_QVECTOR3D_PROPERTY(position, position, setPosition)
-            DECL_QVECTOR3D_PROPERTY(up_vector, upVector, setUpVector)
-            DECL_QVECTOR3D_PROPERTY(view_center, viewCenter, setViewCenter)
-            // clang-format on
-            ;
-
-#undef DECL_QVECTOR3D_PROPERTY
```
Tucchhaa, line 231: this is now available through `widget.camera`

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -487,44 +459,64 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRCameraController
...
+
+        (*this)
+            // clang-format off
+                DECL_FLOAT_PROPERTY(linear_speed, linearSpeed, setLinearSpeed)
+                DECL_FLOAT_PROPERTY(look_speed, lookSpeed, setLookSpeed)
+                DECL_FLOAT_PROPERTY(default_linear_speed, defaultLinearSpeed, setDefaultLinearSpeed)
+                DECL_FLOAT_PROPERTY(default_look_speed, defaultLookSpeed, setDefaultLookSpeed)
+            // clang-format on
+            ;
+#undef DECL_FLOAT_PROPERTY
```
Tucchhaa: Now use macro instead of defining each getter and setter

tests/test_view.py
```diff
@@ -111,190 +103,231 @@ def normalize(self, vec):
...
+
+        c.reset()
+
+        self.assertEqual(c.position, (1, 2, 3))
+        self.assertEqual(c.view_center, (4, 5, 6))
+        self.assertEqual(c.up_vector, (7, 8, 9))
+        self.assertEqual(c.linear_speed, 123.0)
+        self.assertEqual(c.look_speed, 456.0)
+
+
```
Tucchhaa: test common things of both camera types. For now it's default properties

tests/test_view.py
```diff
@@ -111,190 +103,231 @@ def normalize(self, vec):
...
-        self.widget.resetCamera()
+        self.camera.reset()
 
     def test_reset(self):
-        dt = 0.01
-        initial_pos = self.pos()
-        initial_view_vector = self.view_vector()
-        initial_view_center = self.view_center()
-        initial_up_vector = self.up_vector()
+        c = self.camera
```
Tucchhaa, line 206: refactored all camera tests:
* removed usage of `dt` in `camera.move` function
* use properties instead of getters and setters

Tucchhaa:
@yungyuc ready for review

cpp/modmesh/view/RCameraController.cpp
tigercosmos: Better to have a comment to explain what is 45.0, 0.1, and 1000.0. Or, you can have `constexpr float angle = 45.0f`, which can better explain the code.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -135,30 +135,59 @@ class CameraController
...
     Qt3DExtras::QAbstractCameraController * asQtCameraController()
     {
         return dynamic_cast<Qt3DExtras::QAbstractCameraController *>(this);
     }
+
+protected:
+    RCameraInputListener * m_listener = nullptr;
+
+private:
+    QVector3D m_default_position = QVector3D(0.0f, 0.0f, 10.0f);
```
tigercosmos, line 189: Why (0, 0, 10)?

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -487,44 +459,64 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRCameraController
...
+                    DECL_QVECTOR3D_PROPERTY(up_vector, upVector, setUpVector)
+                    DECL_QVECTOR3D_PROPERTY(view_center, viewCenter, setViewCenter)
+                    DECL_QVECTOR3D_PROPERTY(default_position, defaultPosition, setDefaultPosition)
+                    DECL_QVECTOR3D_PROPERTY(default_view_center, defaultViewCenter, setDefaultViewCenter)
+                    DECL_QVECTOR3D_PROPERTY(default_up_vector, defaultUpVector, setDefaultUpVector)
+            // clang-format on
+            ;
+#undef DECL_QVECTOR3D_PROPERTY
+
+#define DECL_FLOAT_PROPERTY(NAME, GETTER, SETTER) \
```
tigercosmos: The macro naming doesn't follow the style rules.
https://github.com/solvcon/modmesh/blob/master/STYLE.rst#c-pre-processor-macro

tests/test_view.py
```diff
@@ -111,190 +103,231 @@ def normalize(self, vec):
...
-        self.assertEqual(view_center[1], new_view_center[1])
-        self.assertEqual(view_center[2], new_view_center[2])
+        self.assertEqual(c.view_center[0], new_view_center[0])
+        self.assertEqual(c.view_center[1], new_view_center[1])
+        self.assertEqual(c.view_center[2], new_view_center[2])
 
     def test_rotation(self):
-        dt = 0.01
-        angle = dt * self.look_speed()
+        c = self.camera
```
tigercosmos, line 226: Personally, I don't like a shorthand like `c`, but maybe yyc has a different thought.

cpp/modmesh/view/R3DWidget.cpp
yungyuc: Moving it to `CameraController` makes it more testable.  I like it.

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -46,24 +46,22 @@
 namespace modmesh
 {
 
-class RScene
-    : public Qt3DCore::QEntity
+class RScene : public Qt3DCore::QEntity
```
yungyuc: Did clang-format force you to make it in one line: `class RScene : public Qt3DCore::QEntity`?

If not, please change it back to use the two lines and avoid this unnecessary change of code.

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -46,24 +46,22 @@
...
 {
-
 public:
 
-    explicit RScene(Qt3DCore::QNode * parent = nullptr)
-        : Qt3DCore::QEntity(parent)
-        , m_controller(new ROrbitCameraController(this))
+    explicit RScene(QNode * parent = nullptr)
+        : QEntity(parent)
```
yungyuc, line 55: It's good to separate the `CameraController` from the scene.

cpp/modmesh/view/R3DWidget.hpp
```diff
@@ -46,24 +46,22 @@
...
     {
+        m_controller = new ROrbitCameraController(this);
     }
 
-    Qt3DExtras::QAbstractCameraController const * controller() const { return m_controller; }
-    Qt3DExtras::QAbstractCameraController * controller() { return m_controller; }
+    CameraController * controller() const { return m_controller; }
+    Qt3DExtras::QAbstractCameraController * qtController() const { return m_controller->asQtCameraController(); }
```
yungyuc: Why do we need an helper `qtController()` for casting?  Is there a reason that the consumers cannot do it themselves?

```cpp
Qt3DExtras::QAbstractCameraController * qtc = scene->controller()->asQtCameraController();
```

If you want to avoid `Qt3DExtras::QAbstractCameraController const *`, an overload of `CameraController * controller()` should work.

cpp/modmesh/view/R3DWidget.hpp
yungyuc: Is there a reason why we do not make (our) `CameraController` a sub class of `QAbstractCameraController`?

cpp/modmesh/view/RCameraController.cpp
yungyuc: It is a good idea to add local `constexpr` for constants.  It will be better document than comments.

Angles need special treatment.  If it's in degree, the (local) name should contain it like `front_deg`.  If it's in radian, the name should contain it accordingly.

It's impossible to exclusively use either degree or radian for angles.  Both are needed by different libraries.  Conversion is a must.  Local variable names should make it clear which unit is used.

---

Names for member data and function parameters of angles are trickier.

cpp/modmesh/view/wrap_view.cpp
yungyuc: Long live C pre-processor.  It is very handy in C++.

tests/test_view.py
```diff
@@ -30,6 +30,7 @@
 import os
 
 import modmesh
+
```
yungyuc: Please remove the unnecessary change of blank line.

tests/test_view.py
```diff
@@ -111,190 +103,231 @@ def normalize(self, vec):
...
+        c.move(yaw=1, left_mouse_button=True)
 
         rotation_matrix = self.angle_axis(angle, (0, 1, 0))
         rotated_vector = np.array(initial_view_vector) @ rotation_matrix
-        view_vector = self.view_vector()
 
-        self.assertAlmostEqual(rotated_vector[0], view_vector[0], delta=1e-5)
-        self.assertAlmostEqual(rotated_vector[1], view_vector[1], delta=1e-5)
-        self.assertAlmostEqual(rotated_vector[2], view_vector[2], delta=1e-5)
+        self.assertAlmostEqual(rotated_vector[0], c.view_vector[0], places=5)
```
yungyuc: `delta` is clearer than `places`.  Why change to the latter?

tests/test_view.py
```diff
@@ -111,190 +103,231 @@ def normalize(self, vec):
         return vec / np.linalg.norm(vec)
 
 
+@unittest.skipIf(GITHUB_ACTIONS, "GUI is not available in GitHub Actions")
+class ViewCommonCameraTC(ViewCameraTB, unittest.TestCase):
+    camera_type = "fps"  # no difference when use orbit camera
```
yungyuc: I believe this comment suggests they behave the same in theory.  If that is true, both fps and orbit cameras should be tested.  Testing is required to make sure the theory is always the reality.

You may use the TB/TC pattern to test both.

yungyuc (changes requested):
> In this PR:
> 
> * New props added to camera controller to control it's reset behaviour: defaultPosition, defaultUpVector, defaultViewCenter, defaultLookSpeed, defaultLinearSpeed
> * All these props are exposed to python
> * Camera controller python API was rewritten to use `def_property` using macros. Tests also were updated
> * Minor issue was fixed: When reset camera, aspect ratio was incorrect

The new Python wrapper and tests are great.

It is unfortunate we cannot execute the tests with github actions.  It's an improvement we should make in the future.

---

- [x] Clarify why [`Qt3DExtras::QAbstractCameraController * qtController() const`](https://github.com/solvcon/modmesh/pull/418/files#diff-f940faded53f4d9cc5445a00082fe9d330c8afddcafef0ba4a42b51d6a77f0e5R60) is needed.
- [x] Is there a reason why we do not make (our) `CameraController` a sub class of `QAbstractCameraController`?  (See discussion at [`setCameraController()`](https://github.com/solvcon/modmesh/pull/418/files#diff-f940faded53f4d9cc5445a00082fe9d330c8afddcafef0ba4a42b51d6a77f0e5R64)).
- [x] Fix the naming of [`#define DECL_FLOAT_PROPERTY(NAME, GETTER, SETTER)`](https://github.com/solvcon/modmesh/pull/418/files#diff-88bff1c42c8197e364f9cbfb23de0593b395fc90d08bd52b8c0f0a7a05ab03beR499): Follow the macro naming style https://github.com/solvcon/modmesh/blob/master/STYLE.rst#c-pre-processor-macro
- [x] Clarify why change `delta` to `places` for [`assertAlmostEqual()`](https://github.com/solvcon/modmesh/pull/418/files#diff-f50af349a1f116824ed7af1180dc8a287127e38fca8ae87d9b556053aafbba44R220)?
- [x] Test both [fps and orbit](https://github.com/solvcon/modmesh/pull/418/files#diff-f50af349a1f116824ed7af1180dc8a287127e38fca8ae87d9b556053aafbba44R108) cameras.

tests/test_view.py
Tucchhaa: Because flake8's limitation to 80 characters makes me do weird things 😵. `places=5` is two characters less than`delta=1e-5`

cpp/modmesh/view/R3DWidget.hpp
Tucchhaa: I like your question. Before I intended to not repeat functionality of Qt in our code. But now since I already duplicated almost all properties to expose it to python, we can remove mixin and make it as you suggest.

cpp/modmesh/view/R3DWidget.hpp
Tucchhaa: I'll make `CameraController` inherit `QAbstractCameraController`, so `qtController` will be removed

cpp/modmesh/view/RCameraController.hpp
Tucchhaa: Because meshes are placed at `(0, 0, 0)`, and we don't want the camera to be inside the mesh

Tucchhaa:
In last commit:
* Used constexpr for 'magic values' in `reset` function ([code](https://github.com/solvcon/modmesh/pull/418/files#diff-be9511fa234431d29c196ed57eaf68dbf67a702bc5400c605cdb9d5450a48d33R217-R219))
* Made `CameraController` to inherit from `QAbstractCameraController` ([code](https://github.com/solvcon/modmesh/pull/418/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR129)) and removed all now unnecessary functions, like `qtCameraControler`. And getters and setters from FPS and Orbit subclasses moved to `CameraController`
* Fixed formatting issues
* Test for both fps and orbit camera types in `ViewCommonCameraTC` ([code](https://github.com/solvcon/modmesh/pull/418/files#diff-f50af349a1f116824ed7af1180dc8a287127e38fca8ae87d9b556053aafbba44R108))

@yungyuc @tigercosmos ready for your review 🙂

cpp/modmesh/view/RCameraController.hpp
tigercosmos: I can guess that, but my point is, better to have some comments for it.

yungyuc: >> Because meshes are placed at `(0, 0, 0)`, and we don't want the camera to be inside the mesh

To put the camera outside the mesh, it is not good to hard-code the default value of the member `m_default_position`, because the mesh may be placed anywhere.

We do not have a reasonably large number of mesh examples for the default location to show disadvantage.

> I can guess that, but my point is, better to have some comments for it.

I concur, but it's OK to leave it here and deal with it in the future.  As we make the code more useful, some meshes will cover the position sooner or later.  We will immediately know and fix it.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -128,77 +126,95 @@ class RCameraInputListener : public Qt3DCore::QEntity
     Qt3DInput::QButtonAxisInput * m_keyboard_tz_neg_input;
 }; /* end class RCameraInputListener */
 
-class CameraController
+class CameraController : public Qt3DExtras::QAbstractCameraController
```
yungyuc: This needs to be named as `RCameraController` because it is a Qt class customized (derived) in modmesh.  See https://github.com/solvcon/modmesh/blob/master/STYLE.rst#qt for the modmesh Qt style.

cpp/modmesh/view/RCameraController.hpp
```diff
@@ -128,77 +126,95 @@ class RCameraInputListener : public Qt3DCore::QEntity
...
+
 public:
-    virtual ~CameraController() = default;
+    explicit CameraController(Qt3DCore::QNode * parent = nullptr)
+        : QAbstractCameraController(parent)
+    {
+    }
 
-    virtual void updateCameraPosition(const CameraInputState & state, float dt) = 0;
+    // Do nothing in QAbstractCameraController's moveCamera
```
yungyuc: Thanks for the informative comment.  It will work even better to put the comment in the function body:

```cpp
{
    // Do nothing in QAbstractCameraController's moveCamera
}
```

tests/test_view.py
yungyuc: That's a clear and valid reason.  But it is also unfortunate 🤣.

Could you please rename `rotated_vector` as `rot_vec` or even `rvec`, and change back to use `delta=1.e-5`?  Using `delta` gives us more correct behaviors.  We will copy and paste the code for future testing.  Behaviorally correct code gives us longer mileage.

We cannot change the argument name nor the float literals, but modmesh developers have the constitutive right to rename local variables.  (We have not written how to treat local variable names in the [style guide](https://github.com/solvcon/modmesh/blob/4166c1aa6ffca26f5cd5865606d495d366ddc9e2/STYLE.rst), but we should.)

tests/test_view.py
```diff
@@ -111,190 +104,235 @@ def normalize(self, vec):
...
+            self.assertEqual(c.position, (1, 2, 3))
+            self.assertEqual(c.view_center, (4, 5, 6))
+            self.assertEqual(c.up_vector, (7, 8, 9))
+            self.assertEqual(c.linear_speed, 123.0)
+            self.assertEqual(c.look_speed, 456.0)
+
+
 @unittest.skipIf(GITHUB_ACTIONS, "GUI is not available in GitHub Actions")
 class ViewFPSCameraTC(ViewCameraTB, unittest.TestCase):
     camera_type = "fps"
```
yungyuc, line 200: The testing code in `ViewFPSCameraTC` and `ViewOrbitCameraTC` looks very similar if not the same.  Is it possible to merge to a base class (`*TB`)?

yungyuc (changes requested):
Thanks for clarifying the many things swiftly.  You updated pretty quickly!  Not sure what time zone you are in 🙂

Additional requests after that:
- [x] [`RCameraController.hpp:129`](https://github.com/solvcon/modmesh/pull/418/files#diff-d22d8f512997b8a57a09a8179b7446c525a92a5797d54e5e8dfa0922555814bbR129): Rename `CameraController` to `RCameraController`.
- [x] [`test_view.py:225`](https://github.com/solvcon/modmesh/pull/418/files#diff-f50af349a1f116824ed7af1180dc8a287127e38fca8ae87d9b556053aafbba44R225): Rename `rotated_vector` as `rot_vec` or something like `rvec`, and change back to use `delta=1.e-5`.
- [x] [`test_view.py:158`](https://github.com/solvcon/modmesh/pull/418/files#diff-f50af349a1f116824ed7af1180dc8a287127e38fca8ae87d9b556053aafbba44R158): Is it possible to merge the testing code in `ViewFPSCameraTC` and `ViewOrbitCameraTC` to a base class (`*TB`)?  (If it is too cumbersome to do it now or you don't like a `*TB`, we can leave the duplicated code here for now.)

cpp/modmesh/view/R3DWidget.hpp
yungyuc: Either way works for me.  Qt prefers inheritance, but HPC code usually uses templates and/or mixins.  There is probably not a definite answer for everything.  But for the Qt code here, inheritance might not be wrong.

tests/test_view.py
Tucchhaa: `test_rest`, `test_rotation` are actually different and `test_translation` indeed is the same in both fps and orbit. But it's not easy to move common tests to the base class, because unittest can't understand that structure.

I will move `test_translation` to `ViewCommonCameraTC`

Tucchhaa:
> You updated pretty quickly! Not sure what time zone you are in 🙂

It's only 2 hours earlier for me. Btw, I will arrive to Taiwan very soon)

tests/test_view.py
Tucchhaa: I did that due to flake8's rule to limit lines to 80 characters. If I write `self.camera` or `camera` everywhere I will get a lot of errors and I will have to shorten more other variables :/

yungyuc: A local `c` in test code is good.  It can't be right to skim through testing code for a complex structure for 3D camera with only a glance.  Writing `c = self.camera` forces a maintainer to pause for a second to keep the information in their brain's short-term memory block, and improves their understanding of the code.

yungyuc:
> > You updated pretty quickly! Not sure what time zone you are in 🙂
> 
> It's only 2 hours earlier for me. Btw, I will arrive to Taiwan very soon)

Around the end of the summer vacation one can be anywhere.  I hope you had a good time this summer.  Having said that, we will be more than happy to share some good ideas of traveling!

tests/test_view.py
yungyuc: It's good to move identical testing code to a single place.  Thanks.

---

A side note.  For testing, it is not rare to duplicate code intentionally.  Compared to "feature" code, we more often change the testing code for experimenting without checking in the change (we want the testing code to be slightly more stable than the feature code).

In this PR, I do not see a reason to duplicate testing code.  This is just FYI.

Tucchhaa:
In last PR:
* Renamed `CameraController` to `RCameraController`
* moved duplicated test function `test_translate` to `ViewCommonCameraTC`
* Replaced `places=5` to `delta=1e-5`

Tucchhaa:
@yungyuc 
I have a question. Does it matter how we name commits? I try to give them meaningful names, but sometimes it's hard to do that. Like last to commits are just 'apply review' and 'apply review 2'. Is it ok for modmesh?

yungyuc:
I'll review the code after the CI completes.

> In last PR:

I think you meant to say commit (or update to the PR).  But I am nitpicking.  If you like to write accurately, it's a reminder.  Whatever term you use does not bother me.

> @yungyuc I have a question. Does it matter how we name commits? I try to give them meaningful names, but sometimes it's hard to do that. Like last to commits are just 'apply review' and 'apply review 2'. Is it ok for modmesh?

It depends.  For a non-trivial enhancement like this PR, some mouthy commits are fine.  But in general I do not want to be picky on commit logs.  I guess I am "just reasonably picky".  We'll see.

cpp/modmesh/view/wrap_view.cpp
```diff
@@ -487,44 +459,64 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRCameraController
...
+            // clang-format off
+                MM_DECL_FLOAT_PROPERTY(linear_speed, linearSpeed, setLinearSpeed)
+                MM_DECL_FLOAT_PROPERTY(look_speed, lookSpeed, setLookSpeed)
+                MM_DECL_FLOAT_PROPERTY(default_linear_speed, defaultLinearSpeed, setDefaultLinearSpeed)
+                MM_DECL_FLOAT_PROPERTY(default_look_speed, defaultLookSpeed, setDefaultLookSpeed)
+            // clang-format on
+            ;
+#undef MM_DECL_FLOAT_PROPERTY
     }
 };
```
yungyuc, line 521: Class ending mark is not here.  It should be added in the future.

yungyuc (approved):
LGTM to me.  Missed class ending marks (e.g., [`wrap_view.cpp:521`](https://github.com/solvcon/modmesh/pull/418/files#diff-88bff1c42c8197e364f9cbfb23de0593b395fc90d08bd52b8c0f0a7a05ab03beR521)) may be addressed in the future.

PR #426, opened 2024-09-11
Use ccache to save CI build time

csinrn:
This PR make use of `ccache` to accelerate the CI build time.

I did not add `ccache` on Windows platform. In my test, `ccache` does not work well on Windows. It views all calls as uncacheable call. I guess it is because windows is [not fully supported](https://ccache.dev/platform-compiler-language-support.html) ([Demo run](https://github.com/csinrn/modmesh/actions/runs/10796610111/job/29945995598), uncachable calls is 34/34).
As a result, I did not add ccache in windows build.

---
### Test result:
* Devbuild.yml
   * Before cache ([Demo run](https://github.com/csinrn/modmesh/actions/runs/10812926198))
MacOS ~12min
Ubuntu ~10min
    * An Ideal cache hit (0 code change) ([Demo run](https://github.com/csinrn/modmesh/actions/runs/10814176494))
MacOS ~5min
Ubuntu ~2min
    * Cached, but with code change (I rebase my branch to modmesh main with 41 new commits ) ([Demo run](https://github.com/csinrn/modmesh/actions/runs/10813894066))
MacOS ~9min
Ubuntu ~8min

* Lint.yml
     * Before cache: [Demo run](https://github.com/csinrn/modmesh/actions/runs/10814276420)
MacOS: ~14min
Ubuntu: ~9min
    * After cache: [Demo run](https://github.com/csinrn/modmesh/actions/runs/10814875763)
MacOS: ~13min
Ubuntu: ~7min

ccache does not help a lot in lint.yml.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -72,6 +72,13 @@ jobs:
           echo "flake8 path: $(which flake8)"
           echo "flake8 version: $(flake8 --version)"
 
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
+          restore-keys: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
+          create-symlink: true
+
       - name: make standalone_buffer
         run: |
           make standalone_buffer_setup
@@ -175,6 +182,13 @@ jobs:
         echo "clang-tidy version: $(clang-tidy -version)"
         echo "flake8 path: $(which flake8)"
         echo "flake8 version: $(flake8 --version)"
+
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        key: ${{ runner.os }}-${{ matrix.cmake_build_type }}
+        restore-keys: ${{ runner.os }}-${{ matrix.cmake_build_type }}
+        create-symlink: true
 
     - name: make gtest BUILD_QT=OFF
       run: |
@@ -351,6 +365,13 @@ jobs:
           echo "clang-tidy version: $(clang-tidy -version)"
           echo "flake8 path: $(which flake8)"
           echo "flake8 version: $(flake8 --version)"
+
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: ${{ runner.os }}-${{ matrix.cmake_build_type }}
+          restore-keys: ${{ runner.os }}-${{ matrix.cmake_build_type }}
+          create-symlink: true
 
       - name: make gtest BUILD_QT=OFF
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -135,6 +135,13 @@ jobs:
         echo "flake8 path: $(which flake8)"
         echo "flake8 version: $(flake8 --version)"
 
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        key: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
+        restore-keys: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
+        create-symlink: true
+
     - name: make cinclude (check_include)
       run: make cinclude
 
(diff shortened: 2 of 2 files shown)


csinrn:
@tigercosmos :)

tigercosmos:
Sweet. It's now faster. :)

.github/workflows/devbuild.yml
```diff
@@ -72,6 +72,13 @@ jobs:
           echo "flake8 path: $(which flake8)"
           echo "flake8 version: $(flake8 --version)"
 
+      - name: ccache
```
yungyuc, line 75: The job standalone_buffer runs fairly fast.  If it uses [ccache](https://ccache.dev), the PR implies all building jobs is supposed to use ccache.  Is my assumption correct?

Is it a common practice to apply ccache to all building jobs in Github Actions?

yungyuc (approved):
This looks good to me.  I am merging the PR.  Follow-up discussions can stay in this PR or use a new issue/conversation.

---

How reliable is [ccache](https://ccache.dev)?

> ccache does not help a lot in lint.yml.

Maybe it does not support clang-tidy very well?

---

In addition to compiler cache, is there a way to cache dependencies / environments?

csinrn:
@yungyuc Thanks for the feedbacks.

> How reliable is [ccache](https://ccache.dev/)?

Considering that it is a 2.3k star [repo](https://github.com/ccache/ccache?tab=readme-ov-file), and I have seen `ccache` in some business code building, I would say it should be quite reliable.

> Maybe it does not support clang-tidy very well?

That's a good point and I have never thought of that. I found several posts online that point out `clang-tidy` does not work well with ccache. It would be good to find a better way to accelerate clang-tidy in github action.

> In addition to compiler cache, is there a way to cache dependencies / environments?

Yes. `Github cache action` can do that. Theoretically, pip install and qt install can benefit from that. I did not apply this in the PR because:
1. Our qt install action has already made use of Github cache action deep in the `jurplel/install-qt-action@v4` action. So our qt install has already been cached.
2. Generally, our pip install process in all platforms take 10-30s. It is quick. Considering that `Github cache action` has its own execution overhead, we cannot save a lot of time using `Github cache action` for pip install.

> The job standalone_buffer runs fairly fast. If it uses [ccache](https://ccache.dev/), the PR implies all building jobs is supposed to use ccache. Is my assumption correct?

Yes. I think so.

> Is it a common practice to apply ccache to all building jobs in Github Actions?

There are about 5 custom github actions about ccache so I would say it is usable, but I would not say it is popular.

yungyuc:
> @yungyuc Thanks for the feedbacks.
> 
> > How reliable is [ccache](https://ccache.dev/)?
> 
> Considering that it is a 2.3k star [repo](https://github.com/ccache/ccache?tab=readme-ov-file), and I have seen `ccache` in some business code building, I would say it should be quite reliable.

Number of stars (fame) is an hint for, but does not constitute an evidence or argument for its reliability.  If you have time, it benefits the team to formulate an analysis for how reliable the tool is.

In the same time, since it is installed in the CI flow, we can experience its reliability ourselves.  But we should be careful to avoid making baseless assumption (on the reliability) before an analysis is available.

> > Maybe it does not support clang-tidy very well?
> 
> That's a good point and I have never thought of that. I found several posts online that point out `clang-tidy` does not work well with ccache. It would be good to find a better way to accelerate clang-tidy in github action.

This sounds like an interesting topic.  clang-tidy performs rather complicated static analysis.  Maybe ccache is not working well for all static analyzers?

If we continue to discuss this, a separate conversation will be productive.

> > In addition to compiler cache, is there a way to cache dependencies / environments?
> 
> Yes. `Github cache action` can do that. Theoretically, pip install and qt install can benefit from that. I did not apply this in the PR because:
> 
> 1. Our qt install action has already made use of Github cache action deep in the `jurplel/install-qt-action@v4` action. So our qt install has already been cached.
> 2. Generally, our pip install process in all platforms take 10-30s. It is quick. Considering that `Github cache action` has its own execution overhead, we cannot save a lot of time using `Github cache action` for pip install.

Yeah, qt could be one of the hotspot before install-qt-action was employed.  This is a good analysis.

> > The job standalone_buffer runs fairly fast. If it uses [ccache](https://ccache.dev/), the PR implies all building jobs is supposed to use ccache. Is my assumption correct?
> 
> Yes. I think so.
> 
> > Is it a common practice to apply ccache to all building jobs in Github Actions?
> 
> There are about 5 custom github actions about ccache so I would say it is usable, but I would not say it is popular.

What do you mean by '5 custom github actions about ccache'?

csinrn:
Thanks @yungyuc.
I'll research more on ccache and clang-tidy, and try to find out whether we can improve the CI again.

> What do you mean by '5 custom github actions about ccache'?

There is a [marketplace](https://github.com/marketplace?query=ccache) for github actions. There are 5 ccache related actions there. I find the ccache action there.

Issue #425, opened 2024-09-08
Long functions in `modmesh::inout::Gmsh` should be defined outside the class declaration


yungyuc:
Some long functions, e.g., [`Gmsh::is_valid_transition()`](https://github.com/solvcon/modmesh/blob/da4328f3fbfd4cd21fde72a1f6abfd41625daa7c/cpp/modmesh/inout/gmsh.hpp#L82) (and more), should be moved outside the class declaration and header file:

```cpp
    bool is_valid_transition(const std::string s)
    {
        if (last_fmt_state == FormatState::BEGIN)
        {
            return s == "$MeshFormat";
        }
        else if (last_fmt_state == FormatState::META_END || last_fmt_state == FormatState::PYHSICAL_NAME_END)
        {
            return s == "$PhysicalNames" || s == "$Nodes";
        }
        else if (last_fmt_state == FormatState::NODE_END)
        {
            return s == "$Elements";
        }

        return false;
    }
```

(In this specific case, the argument should use a const reference `const std::string & s` to save a copy.)

This issue tracks the work to refactor the `Gmsh` class.

ThreeMonth03:
@meng5113 Do you want to solve this issue? I think it may be suitable for you.

ThreeMonth03:
@KHLee529 Are you interested in refactoring the code?

sadla001:
Hi @yungyuc , @ThreeMonth03 is this issue still being worked on? I'd be interested in taking it if it's available.

PR #429, opened 2024-10-05
moved long function definitions out of Gmsh class declaration and header file

meng5113:
The PR is to move long function definitions out of Gmsh class declaration and header file in issue #425.


cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+// Check the finite state machine state transition is valid or not to check msh file format is correct
+bool Gmsh::is_valid_transition(const std::string s)
+{
+    if (last_fmt_state == FormatState::BEGIN)
+    {
+        return s == "$MeshFormat";
+    }
+    else if (last_fmt_state == FormatState::META_END || last_fmt_state == FormatState::PYHSICAL_NAME_END)
+    {
+        return s == "$PhysicalNames" || s == "$Nodes";
```
meng5113, line 107: ij0j0j

yungyuc:
@j8xixo12 could you please help take a look too?

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
     blk->build_boundary();
     blk->build_ghost();
 }
+
+// Check the finite state machine state transition is valid or not to check msh file format is correct
+bool Gmsh::is_valid_transition(const std::string s)
```
yungyuc, line 99: Could you please help change the argument to be a const reference?  It's a simple change but needs to be applied in both the declaration (in header file) and the definition (here).

To be specific, it is to change

```cpp
bool Gmsh::is_valid_transition(const std::string s)
```

with

```cpp
bool Gmsh::is_valid_transition(const std::string & s)
```

That is, change the argument of the function from `const std::string s` to `const std::string & s`.

The change will speed up the code by saving unnecessary string copying.

yungyuc (changes requested):
Thanks, @meng5113 .  Please follow the directives of the linter to fix the issues: https://github.com/solvcon/modmesh/actions/runs/11191984882/job/31118518269?pr=429#step:11:389 .  You can turn on Github Actions in your forked repository to get the same results.

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+{
+    if (last_fmt_state == FormatState::BEGIN)
+    {
+        return s == "$MeshFormat";
+    }
+    else if (last_fmt_state == FormatState::META_END || last_fmt_state == FormatState::PYHSICAL_NAME_END)
```
tigercosmos, line 105: It looks odd to me that you use "or" here. Why not split them into two `else if`?

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+    }
+    else if (last_fmt_state == FormatState::NODE_END)
+    {
+        return s == "$Elements";
+    }
+
+    return false;
+}
+
+void Gmsh::load_meta(void)
```
tigercosmos, line 117: `Gmsh::load_meta()`?

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+{
+    std::string line;
+
+    while (std::getline(stream, line) && line.find('$') == std::string::npos)
+    {
+        auto tokens = tokenize(line, ' ');
+        msh_ver = std::stod(tokens[0]);
+
+        // The parse only support ver 2.2 msh file.
+        if (msh_ver != 2.2)
```
tigercosmos, line 127: I will choose to use `constexpr` for the number.

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+    while (std::getline(stream, line) && line.find('$') == std::string::npos)
+    {
+        auto tokens = tokenize(line, ' ');
+        // gmsh node index is 1-based index
+        m_nds(std::stoul(tokens[0]) - 1, 0) = std::stod(tokens[1]);
+        m_nds(std::stoul(tokens[0]) - 1, 1) = std::stod(tokens[2]);
+        m_nds(std::stoul(tokens[0]) - 1, 2) = std::stod(tokens[3]);
+    }
+
+    if (!line.compare("$EndNodes"))
```
tigercosmos, line 173: I think you should maintain a table to keep these token strings.

cpp/modmesh/inout/gmsh.cpp
```diff
@@ -94,6 +94,159 @@ void Gmsh::build_interior(const std::shared_ptr<StaticMesh> & blk)
...
+        small_vector<uint_type> nds_temp(nds.size() + 1, 0);
+        auto mmcl = eldef.mmcl();
+
+        for (size_t i = 0; i < mmcl.size(); ++i)
+        {
+            nds_temp[mmcl[i] + 1] = nds[i] - 1;
+        }
+        usnds.insert(usnds.end(), nds_temp.begin() + 1, nds_temp.end());
+        nds_temp[0] = mmcl.size();
+        m_elems.insert(std::pair{idx, nds_temp});
```
tigercosmos, line 228: I think `emplace` will be better?

cpp/modmesh/inout/gmsh.hpp
```diff
@@ -78,158 +78,11 @@ class Gmsh
...
-        // put used node id to m_ndmap
-        m_ndmap.remake(small_vector<size_t>{usnds.size()}, -1);
-        for (size_t i = 0; i < usnds.size(); ++i)
-        {
-            m_ndmap(usnds[i]) = i;
-        }
-    }
-
+    bool is_valid_transition(const std::string s);
+    void load_meta(void);
```
tigercosmos, line 82: No need to have `void` in the argument.

cpp/modmesh/inout/gmsh.cpp
j8xixo12: This is my bad, I have checked the [gmsh manual](https://gmsh.info/dev/doc/texinfo/gmsh.pdf) section 10.3.1.
According to the manual, after `MeshFormat`, which is the `META_END` state, it can either be followed by `$PhysicalNames` or `$Nodes`. That is why the function returns an "or" expression. However the problem is that after physical name section, it only can be followed by `Node`. Therefore this code should be changed to 
```
...
else if (last_fmt_state == FormatState::META_END)
{
    return s == "$PhysicalNames" || s == "$Nodes";
}
else if (last_fmt_state == FormatState::PYHSICAL_NAME_END)
{
    return s == "$Nodes";
}
...
```
Thanks @tigercosmos to point out this problem.
I think we can use another PR to fix this problem, it should add some test case to make sure this logic is correct!

I will file an issue to track this.

cpp/modmesh/inout/gmsh.cpp
j8xixo12: Why not just use a macro ?

yungyuc: Macro does not allow compilers to check type. When macros and constexpr both work, choose constexpr.

yungyuc: For version number, integer should be used.  Floating could be inexact and should be avoided.

tigercosmos:
Close due to lack activity.

PR #431, opened 2024-10-13
Feature/time series dataframe

xingularity:
This PR contains the first version of a prototype, TimeSeriesDataFrame which is to preserve and extract time series data for analysis. This feature is related to https://github.com/solvcon/modmesh/discussions/380. Current features as following:
1. Create timeseries_dataframe.py module under modmesh.
2. Create TimeSeriesDataFrame class in the module
3. The TimeSeriesDataFrame utilize columnar format to store data.
4. It uses SimpleArrayUint64 to store index column which is expected to be epoch or incremental unsigned integer.
5. It uses a list of SimpleArrayFloat64 to store every column (except epoch column) in the CSV.
6. It provides "columns" attribute which returns a list of string contains column names.
7. It provides get_column method to extract data in a particular column. The return type is a ndarray.


modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,101 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+import os
+import numpy as np
+
+from modmesh import SimpleArrayUint64, SimpleArrayFloat64
```
yungyuc: When inside `modmesh`, use relative import.  You need the `.core` module for the relative import.

modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+__all__ = [
+    'TimeSeriesDataFrame'
+]
+
+
+class TimeSeriesDataFrame(object):
```
yungyuc: For the very early stage prototype code, put it in unit tests.  It also allows you to add basic tests for your prototype.

modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(self):
+
+        self._columns = list() # Column names
+        self._index_column = None
+        self._index_column_name = None
+        self._data = list()  # List of SimpleArrayFloat64
+
+    def read_from_text_file(
+        self,
+        txt_path: str,
```
yungyuc: Don't use type annotation in Python.  It's bad smell.  If something needs typing information, it should go to C++.

yungyuc (changes requested):
Please leave in-line comments for what to review.  Have all CI passing before requesting for review.  (You can turn on Github Actions in your fork.  It is also productive to run linter locally.)

- [x] For a simple change like this, please use a single commit.  Squash properly.
- [x] [`timeseries_dataframe.py:L31`](https://github.com/solvcon/modmesh/pull/431/files#diff-d04a304bd0afbe837ec9945f0b0d03d428dc9f2f12da1f3c8bdfd1fe6cedb626R31): Use relative import inside `modmesh`.
- [x] [`timeseries_dataframe.py:L39`](https://github.com/solvcon/modmesh/pull/431/files#diff-d04a304bd0afbe837ec9945f0b0d03d428dc9f2f12da1f3c8bdfd1fe6cedb626R39): Move simple prototype to a test file.
- [x] [`timeseries_dataframe.py:L50`](https://github.com/solvcon/modmesh/pull/431/files#diff-d04a304bd0afbe837ec9945f0b0d03d428dc9f2f12da1f3c8bdfd1fe6cedb626R50): Do not use type annotation in Python.

modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,105 @@
...
+
+        nd_arr = np.genfromtxt(txt_path, delimiter=delimiter)[1:]
+        index_column_num = 0 if timestamp_in_file else None
+
+        with open(txt_path, 'r') as f:
+            table_header = [x for x in f.readline().strip().split(delimiter)]
+            if timestamp_in_file:
+                if timestamp_column in table_header:
+                    index_column_num = table_header.index(timestamp_column)
+                self._index_column = SimpleArrayUint64(
```
tigercosmos: I wonder why it's `SimpleArrayUint64`? Can't it be a floating number?

xingularity: Timestamp should be an integer (like epoch), can you provide any other case that we need a floating point?

modmesh/timeseries_dataframe.py
xingularity: Removed and already contained in my latest commit

modmesh/timeseries_dataframe.py
xingularity: Test file and data have been added into the repo.

modmesh/timeseries_dataframe.py
xingularity: The latest commit contains this modification

xingularity (commented):
The spaces are removed in my latest commit. Similar change is also made to Line No. 96 in my latest commit.

modmesh/timeseries_dataframe.py
yungyuc: This class needs to be put in the test file because it is a prototype.  The prototype is not mature enough to be put in `modmesh` namespace.

modmesh/__init__.py
```diff
@@ -41,6 +41,8 @@
 from . import onedim  # noqa: F401
 from . import system  # noqa: F401
 from . import toggle  # noqa: F401
+from . import timeseries_dataframe  # noqa: F401, F403
```
yungyuc: Do not surpress F403.

modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,115 @@
...
+    ):
+        """
+        Generate dataframe from a text file.
+
+        :param txt_path: path to the text file.
+        :type txt_path: str
+        :param delimiter: delimiter.
+        :type delimiter: str
+        :param timestamp_in_file: If the text file containing index column,
+        expected to be integer.
```
yungyuc: Wrong indentation.  Refer Sphinx style guide and PEP-8.

yungyuc: A linter/formatter for docstrings will help.  We should consider add it if it's available.

modmesh/timeseries_dataframe.py
```diff
@@ -0,0 +1,115 @@
...
+            if index_column_num is not None:
+                self._columns.pop(index_column_num)
+
+        for i in range(nd_arr.shape[1]):
+            if i != index_column_num:
+                self._data.append(
+                    SimpleArrayFloat64(array=nd_arr[:, i].copy())
+                )
+
+    def __getitem__(self, column_name):
```
yungyuc: Is it a convention of Pandas or any popular package to name the argument this long?

It should only be the name for column so `name` alone seems to suffice.  But I can be wrong.

yungyuc (changes requested):
You have not addressed [all points in the previous review](https://github.com/solvcon/modmesh/pull/431#pullrequestreview-2365003509).  When requesting for the next review, please address all points.

When requesting for the next review, a global comment should be added and summarize the changes.  Inline comments should be added to highlight what to review.

It is ideal for each review to use a commit.  Sometimes a couple of commits are OK.  But so far you are using 16.  They are too many.  Please do squash for this batch.  Be kind to your reviewer.

Additional points:
- [x] Since the time series prototype should go to test file, no change should be made in `__init__.py`.

modmesh/timeseries_dataframe.py
tigercosmos: I think it depends on the unit? for example, if the unit is ms, and the precision is ns, then the timestamp can be a floating number.

tests/test_timeseries_dataframe.py
```diff
@@ -0,0 +1,126 @@
...
+            list(tsdf.index), list(nd_arr[:, 0].astype(np.uint64))
+        )
+
+    def test_dataframe_get_column(self):
+        tsdf = TimeSeriesDataFrame()
+        tsdf.read_from_text_file(
+            os.path.join(self.DATADIR, "dlc_trimmed.csv")
+        )
+
+        one_column_data = tsdf['DATA_DELTA_VEL[1] ']
```
j8xixo12: I’m wondering if the user needs to include a space when entering the column name?

xingularity: Hi @j8xixo12
My design is to preserve whatever there is in the CSV. I do not want to assume what the data "should be". If there is a space in the column name in the original file, I will read as is. This is also the behavior of Pandas if you use it to read my test CSVs.

yungyuc: It is not user friendly to require them remember the whitespaces in text file.  It is especially difficult when the whitespaces are at the end of a line.  Trailing whitespaces are invisible in many editors and web browsers.

In this early prototype it's OK to keep the space in key names.  But I don't think it's a good design to force keeping whitespaces.  We need to think about it in the future.

modmesh/timeseries_dataframe.py
xingularity: Hi @yungyuc 
May I ask what is the suggested indentation in the docstring? It seems not easy to find suitable examples which has a long text (for a parameter) and need to have a new line in the guide documents. But I just found this in the cpython code. Do you think if this is what you expect to be (4 spaces indentation in the docstring).
https://github.com/python/cpython/blob/main/Tools/build/stable_abi.py#L73C27-L80

xingularity: 4 spaces indentation (from the start of the parameter name) in the docstring.

modmesh/timeseries_dataframe.py
xingularity: This is a good point. Since in the normal practice of a DataFrame, the "[]" operator usually extracts a column, we can just use "name" instead of column_name. I will change in my next commit.

modmesh/timeseries_dataframe.py
yungyuc: 4-space indentation is reasonable, but [any indentation](https://stackoverflow.com/questions/32566466/multi-line-description-of-a-parameter-description-in-python-docstring) is acceptable.

tests/test_timeseries_dataframe.py
xingularity: Hi @yungyuc 

I think we can discuss further when we have F2F time, and I still want to keep the trailing white spaces since user can access the list of columns through "columns" property. I think there could be scenario where user wants to keep the column names as is. Let's say user compose an automatic script to read the column names from another file (maybe the config of the data source instruments) and the trailing spaces is there in the beginning. Under this scenario the user may want the column name stays the same.

modmesh/timeseries_dataframe.py
xingularity: Hi @tigercosmos 
The reason I designed in this way is because of the data from the model problem we are trying to use is integer, and it seems normal to treat timestamp an integer. I think the scenario you refer to is that the meaning of the time is the delta from certain starting point (1.2s, 2.0s, ...). This is not in our model problem so I will keep as is for this PR. But we can definitely open an issue when we encounter this kind of data.

tigercosmos: Sure, it's clearer to me.

tests/test_timeseries_dataframe.py
yungyuc: You can keep it for now.

.github/workflows/devbuild.yml
xingularity, line 1: These changes are pulled from upstream master branch to resolve CI/CD pip install issue.

modmesh/__init__.py
xingularity, line 1: __init__.py in modmesh module is restored since TimeSeriesDataFrame class has been moved to test file.

xingularity:
Hi @yungyuc and reviewers

My latest commit should have made following changes, and therefore, should have already addressed above issues.

1. I have moved the TimeSeriesDataFrame prototype from modmesh module into the test file.
2. Removed timeseries_dataframe.py in modmesh module from repo.
3. Since 1 is completed, the __init__.py of the modmesh module is restored.
4. Previous 16 commits are squashed into one, and force pushed onto github branch.
5. Changed parameter name in the TimeSeriesDataFrame. using "name" instead of column name.
6. Fixed indentation in docstring.

All the CIs are passed, please help to review this PR.

modmesh/__init__.py
xingularity: Noted. This line is removed since I moved prototype into test file.

.github/workflows/devbuild.yml
yungyuc, line 1: Just for this time.  Please rebase, do not duplicate change.  The former is clearer than the latter.

tests/data/dlc_trimmed.csv
yungyuc, line 1: Please put the contents of the two short fixture csv files `dlc_trimmed.csv` and `dlc_trimmed_header_changed.csv` in the test file `test_timeseries_dataframe.py`.  They are short and distinct files are harder to read and maintain.

Distinct fixture files can be created later after the fixture (data) become larger/longer.

tests/test_timeseries_dataframe.py
yungyuc: Do not mark conversation resolved.  Doing it makes it harder to track.

yungyuc (changes requested):
Thanks for the update, @xingularity .  Additional changes are requested:

- [ ] Rebase instead of duplicate changes for CI update.
- [x] Move the fixture contents to the test (code) file because there are not yet many lines.

tests/data/dlc_trimmed.csv
xingularity: Hi @yungyuc 

I found the numpy function I used to generate data from text supported StringIO. I will change my function to support  streaming text input, and I will also put the CSV data into test file. I estimate this can be done on 11/3.

xingularity: Need more time to verify my new implementation, push the estimated time to 11/6.

tests/test_timeseries_dataframe.py
```diff
@@ -62,36 +66,50 @@ def read_from_text_file(
...
+            fhd = iter(fid)
+
+            idx_col_num = 0 if timestamp_in_file else None
+
+            table_header = [
+                x.strip() for x in next(fhd).strip().split(delimiter)
+            ]
+            nd_arr = np.genfromtxt(fhd, delimiter=delimiter)
+
+            self._init_members()
```
xingularity, line 89: Re-initiate object members when reading a new file using same object.

xingularity (commented):
1. Moved test data into the test file and removed CSV files. 
2. Refactored read_from_text_file method to accept file path (as string), iterable of strings or StringIO.
3. Shorten several variable names to make code cleaner.
4. column name now strips white spaces.
5. re-initiate class members when reading a new file.
6. Modify all tests to read CSV using StringIO in test file.

tests/test_timeseries_dataframe.py
```diff
@@ -0,0 +1,223 @@
...
+
+import numpy as np
+
+from modmesh import SimpleArrayUint64, SimpleArrayFloat64
+
+
+class TimeSeriesDataFrame(object):
+
+    def __init__(self):
+        self._init_members()
```
yungyuc, line 40: The code will be more readable to have member data listed in `__init__()`, but keeping them in a separate function is OK for now in the prototype.

yungyuc (changes requested):
The code looks much cleaner.  Thank you, @xingularity .

- [x] Please rebase to get rid of the change in the directory of `.github/workflows`.

yungyuc (approved):
It's [rebased](https://github.com/solvcon/modmesh/pull/431#event-15112768704).  I'll merge once CI finishes running.

xingularity:
> It's [rebased](https://github.com/solvcon/modmesh/pull/431#event-15112768704). I'll merge once CI finishes running.

It looks like one of them has failed. But the all CI under devbuild have passed in my fork repo. Quiet weird.

yungyuc:
It's a sensitive profiling test and passed after rerun.  Merged.

Issue #432, opened 2024-10-19
Incorrect construction of SimpleArray from a discontiguous slice of ndarray


xingularity:
When a SimpleArray is constructed through the "array" argument in Python and the numpy array passed to this argument is a subset using ndarray accessor, the SimpleArray is constructed with wrong section of data from the original ndarray. Following code is a demonstration.

```
from modmesh import SimpleArrayFloat64
import numpy as np
a = np.array([[1,2,3,4,5], [10,20,30,40,50],[100,200,300,400,500]])
```
Now we have 

```
array([[  1,   2,   3,   4,   5],
       [ 10,  20,  30,  40,  50],
       [100, 200, 300, 400, 500]])
```
If we construct a SimpleArray with
```
b = SimpleArrayFloat64(array=a[:, 4])
print(b.ndarray)
```
The print result is [ 5., 10., 20.].
We should expect a SimpleArray contains [5, 50, 500] instead of [ 5., 10., 20.]

yungyuc:
The `SimpleArray` constructor expects a contiguous memory buffer and disregard the strides set in `ndarray`.  Before `SimpleArray` supports the same stride as `ndarray`, it should throw an exception when the input `ndarray` is not contiguous.

For now, please use `b = SimpleArrayFloat64(array=a[:,4].copy())` to work around.

xingularity:
Thanks @yungyuc, I already used that workaround in my code.

yungyuc:
@tigercosmos Would you have time to take a look?

tigercosmos:
I can help support this starting from the beginning of November.

yungyuc:
Fixed by #438

Issue #437, opened 2024-11-28
[Bug] Incorrect construction of SimpleArray from a high-dimension ndarray


ThreeMonth03:
I notice that when the number of the dimension > 3, the construction of SimpleArray from the ndarray would fail.

How to reproduce the bug:
```python3
import unittest
from modmesh import SimpleArrayUint64, SimpleArrayFloat64
import numpy as np

class TimeBuffer(unittest.TestCase):
    a = np.ndarray(shape= (5,5,5,5), dtype=float)
    for i in range(5):
        for j in range(5):
            for k in range(5):
                for l in range(5):
                    a[i,j,k,l] = i*1000 + j*100 + k*10 + l

    b_copy = SimpleArrayFloat64(array=a)
    print(b_copy.ndarray)
```

[Error log](https://github.com/ThreeMonth03/modmesh/actions/runs/12073106861/job/33668534857):
```console
============================= test session starts ==============================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/runner/work/modmesh/modmesh
double free or corruption (out)
Fatal Python error: Aborted

Current thread 0x00007f6be743e000 (most recent call first):
  File "/home/runner/work/modmesh/modmesh/tests/test_array.py", line 26 in TimeBuffer
  File "/home/runner/work/modmesh/modmesh/tests/test_array.py", line 5 in <module>
  File "/usr/local/lib/python3.10/dist-packages/_pytest/assertion/rewrite.py", line 184 in exec_module
  File "<frozen importlib._bootstrap>", line 688 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1006 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1027 in _find_and_load
  File "<frozen importlib._bootstrap>", line 1050 in _gcd_import
  File "/usr/lib/python3.10/importlib/__init__.py", line 126 in import_module
  File "/usr/local/lib/python3.10/dist-packages/_pytest/pathlib.py", line 582 in import_path
  File "/usr/local/lib/python3.10/dist-packages/_pytest/python.py", line 493 in importtestmodule
  File "/usr/local/lib/python3.10/dist-packages/_pytest/python.py", line 546 in _getobj
  File "/usr/local/lib/python3.10/dist-packages/_pytest/python.py", line 284 in obj
[...]

PR #438, opened 2024-11-28
Fix the construction of SimpleArray from slicing ndarray

ThreeMonth03:
In this pull request, I fix the issue #432 by passing the stride vector to SimpleArray.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -32,6 +32,8 @@
 
 #include <limits>
 #include <stdexcept>
+#include <functional>
+#include <numeric>
 
 #if defined(_MSC_VER)
 #include <BaseTsd.h>
@@ -304,6 +306,65 @@ public:
                 throw std::runtime_error(Formatter() << "SimpleArray: shape byte count " << nbytes
                                                      << " differs from buffer " << buffer->nbytes());
             }
+        }
+    }
+
+    explicit SimpleArray(small_vector<size_t> const & shape,
+                         small_vector<size_t> const & stride,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         bool is_c_contiguous = true,
+                         bool is_f_contiguous = false)
+        : SimpleArray(buffer)
+    {
+        if (buffer)
+        {
+            if (shape.size() != stride.size())
+            {
+                throw std::runtime_error("SimpleArray: shape and stride size mismatch");
+            }
+
+            if (is_c_contiguous)
+            {
+                if (stride[stride.size() - 1] != 1)
+                {
+                    throw std::runtime_error("SimpleArray: C contiguous stride must end with 1");
+                }
+                for (size_t it = 0; it < shape.size() - 1; ++it)
+                {
+                    if (stride[it] != shape[it + 1] * stride[it + 1])
+                    {
+                        throw std::runtime_error("SimpleArray: C contiguous stride must match shape");
+                    }
+                }
+            }
+            if (is_f_contiguous)
+            {
+                if (stride[0] != 1)
+                {
+                    throw std::runtime_error("SimpleArray: Fortran contiguous stride must start with 1");
+                }
+                for (size_t it = 0; it < shape.size() - 1; ++it)
+                {
+                    if (stride[it + 1] != shape[it] * stride[it])
+                    {
+                        throw std::runtime_error("SimpleArray: Fortran contiguous stride must match shape");
+                    }
+                }
+            }
+
+            const size_t nbytes = ITEMSIZE *
+                                  std::accumulate(shape.begin(),
+                                                  shape.end(),
+                                                  static_cast<size_t>(1),
+                                                  std::multiplies<size_t>());
+            if (nbytes != buffer->nbytes())
+            {
+                throw std::runtime_error(Formatter() << "SimpleArray: shape byte count " << nbytes
+                                                     << " differs from buffer " << buffer->nbytes());
+            }
+
+            m_shape = shape;
+            m_stride = stride;
         }
     }
 
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -74,16 +74,24 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                         {
                             throw std::runtime_error("dtype mismatch");
                         }
+
                         modmesh::detail::shape_type shape;
+                        modmesh::detail::shape_type stride;
+                        constexpr size_t itemsize = wrapped_type::itemsize();
                         for (ssize_t i = 0; i < arr_in.ndim(); ++i)
                         {
                             shape.push_back(arr_in.shape(i));
+                            stride.push_back(arr_in.strides(i) / itemsize);
                         }
+
+                        const bool is_c_contiguous = (arr_in.flags() & py::array::c_style) == py::array::c_style;
+                        const bool is_f_contiguous = (arr_in.flags() & py::array::f_style) == py::array::f_style;
+
                         std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
                             arr_in.nbytes(),
                             arr_in.mutable_data(),
                             std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
-                        return wrapped_type(shape, buffer);
+                        return wrapped_type(shape, stride, buffer, is_c_contiguous, is_f_contiguous);
                     }),
                 py::arg("array"))
             .def_buffer(&property_helper::get_buffer_info)
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -450,6 +450,27 @@ class SimpleArrayBasicTC(unittest.TestCase):
         sarr.ndarray.fill(100)
         self.assertTrue((ndarr == 100).all())
 
+    def test_SimpleArray_from_ndarray_slice(self):
+        ndarr = np.arange(1000, dtype='float64').reshape((10, 10, 10))
+        parr = ndarr[1:7:3, 6:2:-1, 3:9]
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr[1:7:3, 6:2:-1, 3:9])
+
+        for i in range(2):
+            for j in range(4):
+                for k in range(6):
+                    self.assertEqual(parr[i, j, k], sarr[i, j, k])
+
+    def test_SimpleArray_from_ndarray_transpose(self):
+        ndarr = np.arange(350, dtype='float64').reshape((5, 7, 10))
+        # The following array is F contiguous.
+        parr = ndarr[2:4].T
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr[2:4].T)
+
+        for i in range(10):
+            for j in range(7):
+                for k in range(2):
+                    self.assertEqual(parr[i, j, k], sarr[i, j, k])
+
     def test_SimpleArray_broadcast_ellipsis_shape(self):
         sarr = modmesh.SimpleArrayFloat64((2, 3, 4))
         ndarr = np.arange(2 * 3 * 4, dtype='float64').reshape((2, 3, 4))


cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -75,15 +75,18 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                             throw std::runtime_error("dtype mismatch");
                         }
                         modmesh::detail::shape_type shape;
+                        modmesh::detail::shape_type stride;
+                        constexpr size_t itemsize = wrapped_type::itemsize();
                         for (ssize_t i = 0; i < arr_in.ndim(); ++i)
                         {
                             shape.push_back(arr_in.shape(i));
+                            stride.push_back(arr_in.strides(i) / itemsize);
```
ThreeMonth03, line 84: The unit of the `py::array::strides()` is bytes, and the unit of the `SimpleArray::strides()` is the elements.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -307,6 +307,16 @@ class SimpleArray
...
 
+    explicit SimpleArray(small_vector<size_t> const & shape, small_vector<size_t> const & stride, std::shared_ptr<buffer_type> const & buffer)
+        : SimpleArray(buffer)
+    {
+        if (buffer)
+        {
+            m_shape = shape;
+            m_stride = stride;
+        }
+    }
```
ThreeMonth03: I create this function according to the function `explicit SimpleArray(small_vector<size_t> const & shape, std::shared_ptr<buffer_type> const & buffer)`, but I'm not sure whether it would create the potential bug without checking `nbytes == buffer->nbytes()`.

tests/test_buffer.py
```diff
@@ -450,6 +450,15 @@ def test_SimpleArray_from_ndarray_content(self):
...
         self.assertTrue((ndarr == 100).all())
 
+    def test_SimpleArray_from_ndarray_slice(self):
+        ndarr = np.arange(1000, dtype='float64').reshape((10, 10, 10))
+        parr = ndarr[1:7, 2:6, 3:9]
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr[1:7, 2:6, 3:9])
+        for i in range(6):
+            for j in range(4):
+                for k in range(6):
+                    self.assertEqual(parr[i, j, k], sarr[i, j, k])
```
ThreeMonth03: This is the unittest of constructing SimpleArray from slicing ndarray. When the dimension is less than 4, the constructor should work regularly.

ThreeMonth03 (commented):
Instead of allocating the contiguous memory and copying the elements from py::array, I try to fix the bug by passing the stride vector.

ThreeMonth03:
> Instead of allocating the contiguous memory and copying the elements from py::array, I try to fix the bug by passing the stride vector.

@yungyuc @tigercosmos Could you please review the pull request?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: We should check the shape and stride are within the buffer range, in the same way it is done in [the overload `explicit SimpleArray(small_vector<size_t> const & shape, std::shared_ptr<buffer_type> const & buffer);`](https://github.com/solvcon/modmesh/blob/ad32dd95c59d1eac22941682e938868951a0db0a/cpp/modmesh/buffer/SimpleArray.hpp#L294)

tests/test_buffer.py
yungyuc: This is why you filed issue #437.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: The code is clear enough.  It is correct to use no comment.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -75,15 +75,18 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
                         }
                         std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
                             arr_in.nbytes(),
                             arr_in.mutable_data(),
                             std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
-                        return wrapped_type(shape, buffer);
+                        return wrapped_type(shape, stride, buffer);
```
yungyuc: The wrapper should check if the input array is contiguous.  You can use [the ndarray flags](https://numpy.org/doc/2.1/reference/c-api/array.html#array-flags) and/or the source striding value.

yungyuc (changes requested):
> Instead of allocating the contiguous memory and copying the elements from `py::array`, I try to fix the bug by passing the stride vector.

This is a good fix.  When taking a memory buffer, `SimpleArray` should reuse instead of copying it.  Sanity check should be done during the simple array construction so that the array operating code can follow the assumption made by the class template `SimpleArray`.

Some enhancements are needed in the PR:

- [x] [`SimpleArray.hpp:310`](https://github.com/solvcon/modmesh/pull/438/files#diff-bf43ff74a749046af0d1aca3542c01ba73e98d3065d0cabb4fc3f6ebc1f85f7dR310): Check the shape and stride are within the buffer range, in the same way it is done in [the overload `explicit SimpleArray(small_vector<size_t> const & shape, std::shared_ptr<buffer_type> const & buffer);`](https://github.com/solvcon/modmesh/blob/ad32dd95c59d1eac22941682e938868951a0db0a/cpp/modmesh/buffer/SimpleArray.hpp#L294)
- [x] [`wrap_SimpleArray.cpp:89`](https://github.com/solvcon/modmesh/pull/438/files#diff-260aa2069e17fc23d433c409a69e235ab3ed90bbf0362f2f88a7dcb3037b223eR89): The wrapper should check if the input array is contiguous.  You can use [the ndarray flags](https://numpy.org/doc/2.1/reference/c-api/array.html#array-flags) and/or the source striding value.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
ThreeMonth03: Why should the wrapper check if the input array is contiguous?

yungyuc: Because only the wrapper sees the `PyObject` of the incoming `nearray`, on which the contiguous bit is available.  When the buffer pointer is passed in the `SimpleArray` C++ code, no information of memory contiguity is available.  The constructor can only trust the shape and stride passed in blindly.

Checking for contiguity in the wrapper is more robust.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -307,6 +309,26 @@ class SimpleArray
...
+            const size_t nbytes = std::accumulate(m_shape.begin(),
+                                                  m_shape.end(),
+                                                  static_cast<size_t>(1),
+                                                  std::multiplies<size_t>()) *
+                                  ITEMSIZE;
+            if (nbytes != buffer->nbytes())
+            {
+                throw std::runtime_error(Formatter() << "SimpleArray: shape byte count " << nbytes
+                                                     << " differs from buffer " << buffer->nbytes());
+            }
```
ThreeMonth03: Check the `nbytes()`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -75,15 +75,19 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
                         {
                             shape.push_back(arr_in.shape(i));
+                            stride.push_back(arr_in.strides(i) / itemsize);
                         }
                         std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
                             arr_in.nbytes(),
                             arr_in.mutable_data(),
                             std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
-                        return wrapped_type(shape, buffer);
+                        return is_c_contiguous ? wrapped_type(shape, buffer) : wrapped_type(shape, stride, buffer);
```
ThreeMonth03: If the array is c contiguous, the wrapper use the original constructor.

tests/test_buffer.py
```diff
@@ -450,6 +450,16 @@ def test_SimpleArray_from_ndarray_content(self):
         sarr.ndarray.fill(100)
         self.assertTrue((ndarr == 100).all())
 
+    def test_SimpleArray_from_ndarray_slice(self):
+        ndarr = np.arange(1000, dtype='float64').reshape((10, 10, 10))
+        parr = ndarr[1:7:3, 6:2:-1, 3:9]
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr[1:7:3, 6:2:-1, 3:9])
```
ThreeMonth03, line 456: I modify the step to make the testcase complicated.

ThreeMonth03 (commented):
* [x]  [`SimpleArray.hpp:310`](https://github.com/solvcon/modmesh/pull/438/files#diff-bf43ff74a749046af0d1aca3542c01ba73e98d3065d0cabb4fc3f6ebc1f85f7dR310): Check the shape and stride are within the buffer range, in the same way it is done in [the overload `explicit SimpleArray(small_vector<size_t> const & shape, std::shared_ptr<buffer_type> const & buffer);`](https://github.com/solvcon/modmesh/blob/ad32dd95c59d1eac22941682e938868951a0db0a/cpp/modmesh/buffer/SimpleArray.hpp#L294)
 * [x]  [`wrap_SimpleArray.cpp:89`](https://github.com/solvcon/modmesh/pull/438/files#diff-260aa2069e17fc23d433c409a69e235ab3ed90bbf0362f2f88a7dcb3037b223eR89): The wrapper should check if the input array is contiguous.  You can use [the ndarray flags](https://numpy.org/doc/2.1/reference/c-api/array.html#array-flags) and/or the source striding value.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: f_contiguous also needs to be handled.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -307,6 +309,26 @@ class SimpleArray
...
+    {
+        if (buffer)
+        {
+            m_shape = shape;
+            m_stride = stride;
+            const size_t nbytes = std::accumulate(m_shape.begin(),
+                                                  m_shape.end(),
+                                                  static_cast<size_t>(1),
+                                                  std::multiplies<size_t>()) *
+                                  ITEMSIZE;
```
yungyuc: Would the expression look nicer by placing `ITEMSIZE` before `std::accumulate`?

yungyuc (changes requested):
One more point to address.

- [x] [`wrap_SimpleArray.cpp:90`](https://github.com/solvcon/modmesh/pull/438/files#diff-260aa2069e17fc23d433c409a69e235ab3ed90bbf0362f2f88a7dcb3037b223eR90): f_contiguous also needs to be handled.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: That's a good idea.

tests/test_buffer.py
```diff
@@ -450,6 +450,28 @@ def test_SimpleArray_from_ndarray_content(self):
...
+    def test_SimpleArray_from_ndarray_transpose(self):
+        ndarr = np.arange(350, dtype='float64').reshape((5, 7, 10))
+        # The following array is F contiguous.
+        parr = ndarr[2:4].T 
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr[2:4].T)
+
+        for i in range(10):
+            for j in range(7):
+                for k in range(2):
+                    self.assertEqual(parr[i, j, k], sarr[i, j, k])
```
ThreeMonth03: Unittest for F contiguous ndarray.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -307,6 +309,65 @@ class SimpleArray
...
+                    throw std::runtime_error("SimpleArray: Fortran contiguous stride must start with 1");
+                }
+                for (size_t it = 0; it < shape.size() - 1; ++it)
+                {
+                    if (stride[it + 1] != shape[it] * stride[it])
+                    {
+                        throw std::runtime_error("SimpleArray: Fortran contiguous stride must match shape");
+                    }
+                }
+            }
```
ThreeMonth03, line 353: Check the C contiguous and F contiguous array.

ThreeMonth03 (commented):
* [x] Check the F contiguous array.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: The current implementation is good.

We should refactor it to use a distinct helper (class or function) in a later PR (may use a separate issue to track).

yungyuc (approved):
Looks good!

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: By the way I have a questions:
1. How could we pass and receive the `bool is_c_contiguous` and `bool is_f_contiguous`  elegantly? 
I have considered passing the numpy flags directly, but the constructor may not only deal with the numpy array. Is it a good way to design the customed flags by  `enum class types` to deal with every possible input?

yungyuc: I don't think the class template `SimpleArray` needs to know f/c continuity.  The sanity check may happen in the Python wrapper which has the information from Numpy.  Standalone helper (static or free function) allows you to check in C++ and pybind11 wrapper.  It could make sense to create a stride object to organize the logics.

It is ok to add more metadata in `SimpleArray`, including f/c continuity, but it is a separate topic.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -307,6 +309,65 @@ class SimpleArray
         }
     }
 
+    explicit SimpleArray(small_vector<size_t> const & shape,
+                         small_vector<size_t> const & stride,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         bool is_c_contiguous = true,
```
tigercosmos, line 315: why not use an enum here? @ThreeMonth03

ThreeMonth03: I realized I could use an enum class inside the `class SimpleArray` after I asked the question.

tigercosmos: @ThreeMonth03  Maybe you can consider opening another PR to fix this.

PR #439, opened 2024-12-04
Refactor: refactor the constructor of SimpleArray

ThreeMonth03:
In the pull request #438, @yungyuc mentioned that I could refactor the constructor.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -312,46 +312,31 @@ public:
     explicit SimpleArray(small_vector<size_t> const & shape,
                          small_vector<size_t> const & stride,
                          std::shared_ptr<buffer_type> const & buffer,
-                         bool is_c_contiguous = true,
-                         bool is_f_contiguous = false)
+                         bool c_contiguous,
+                         bool f_contiguous)
+        : SimpleArray(shape, stride, buffer)
+    {
+        if (shape.size() != stride.size())
+        {
+            throw std::runtime_error("SimpleArray: shape and stride size mismatch");
+        }
+        if (c_contiguous)
+        {
+            check_c_contiguous(shape, stride);
+        }
+        if (f_contiguous)
+        {
+            check_f_contiguous(shape, stride);
+        }
+    }
+
+    explicit SimpleArray(small_vector<size_t> const & shape,
+                         small_vector<size_t> const & stride,
+                         std::shared_ptr<buffer_type> const & buffer)
         : SimpleArray(buffer)
     {
         if (buffer)
         {
-            if (shape.size() != stride.size())
-            {
-                throw std::runtime_error("SimpleArray: shape and stride size mismatch");
-            }
-
-            if (is_c_contiguous)
-            {
-                if (stride[stride.size() - 1] != 1)
-                {
-                    throw std::runtime_error("SimpleArray: C contiguous stride must end with 1");
-                }
-                for (size_t it = 0; it < shape.size() - 1; ++it)
-                {
-                    if (stride[it] != shape[it + 1] * stride[it + 1])
-                    {
-                        throw std::runtime_error("SimpleArray: C contiguous stride must match shape");
-                    }
-                }
-            }
-            if (is_f_contiguous)
-            {
-                if (stride[0] != 1)
-                {
-                    throw std::runtime_error("SimpleArray: Fortran contiguous stride must start with 1");
-                }
-                for (size_t it = 0; it < shape.size() - 1; ++it)
-                {
-                    if (stride[it + 1] != shape[it] * stride[it])
-                    {
-                        throw std::runtime_error("SimpleArray: Fortran contiguous stride must match shape");
-                    }
-                }
-            }
-
             const size_t nbytes = ITEMSIZE *
                                   std::accumulate(shape.begin(),
                                                   shape.end(),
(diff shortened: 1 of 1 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -312,46 +312,31 @@ class SimpleArray
     explicit SimpleArray(small_vector<size_t> const & shape,
                          small_vector<size_t> const & stride,
                          std::shared_ptr<buffer_type> const & buffer,
-                         bool is_c_contiguous = true,
-                         bool is_f_contiguous = false)
+                         bool c_contiguous,
+                         bool f_contiguous)
+        : SimpleArray(shape, stride, buffer)
```
ThreeMonth03, line 317: In order to avoid 
```console
error: call to constructor of 'SimpleArray<bool>' is ambiguous
  330 |         return SimpleArray(shape, stride, buffer);
```
I don't assign the default value to `c_contiguous` and `f_contiguous`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -312,46 +312,31 @@ class SimpleArray
...
+        }
+        if (c_contiguous)
+        {
+            check_c_contiguous(shape, stride);
+        }
+        if (f_contiguous)
+        {
+            check_f_contiguous(shape, stride);
+        }
+    }
```
ThreeMonth03, line 331: @jteng2127 suggest that I could create a constructor with `c_contiguous` and `f_contiguous`, and this constructor could check the contiguous property and call another construcotr without `c_contiguous` and `f_contiguous`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -84,14 +84,20 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-                        const bool is_c_contiguous = (arr_in.flags() & py::array::c_style) == py::array::c_style;
-                        const bool is_f_contiguous = (arr_in.flags() & py::array::f_style) == py::array::f_style;
+                        auto && is_c_contiguous = [&]() -> bool
+                        {
+                            return (arr_in.flags() & py::array::c_style) == py::array::c_style;
+                        };
+                        auto && is_f_contiguous = [&]() -> bool
+                        {
+                            return (arr_in.flags() & py::array::f_style) == py::array::f_style;
+                        };
```
ThreeMonth03: I'm not sure whether wrapping with lambda function is better than initializing the boolean variable.

ThreeMonth03 (commented):
@yungyuc @tigercosmos Could you please review this pull request?

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Since we only have C type or F type, why not just:

```c++
    explicit SimpleArray(small_vector<size_t> const & shape,
                         small_vector<size_t> const & stride,
                         std::shared_ptr<buffer_type> const & buffer,
                         bool is_c_contiguous = true)
```

I prefer using enum more, just for your reference:

```c++
enum class SimpleArrayMemoryType {
    CType
    FType
};

    explicit SimpleArray(small_vector<size_t> const & shape,
                         small_vector<size_t> const & stride,
                         std::shared_ptr<buffer_type> const & buffer,
                         SimpleArrayMemoryType is_c_contiguous = SimpleArrayMemoryType::CType)
```

yungyuc: Boolean flag does not need to introduce new symbols at the Python side and is preferred ATM.

The input buffer may be F or C contiguous or discontiguous.  A single Boolean flag is not enough.

yungyuc: And it's good to remove default arguments in C++.  They usually do harm more than good.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: These lambda implementations work the same as direct evaluation.  They do not simplify anything either.  Please revert them.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -623,6 +608,37 @@ class SimpleArray
     value_type * body() { return m_body; }
 
 private:
+    void check_c_contiguous(small_vector<size_t> const & shape,
```
yungyuc, line 611: I like the helper functions.

They do not need to be in the class template because they do not use the template argument.  But it's OK to keep them here for now.

There are other member functions that can be moved outside the class template.  We only need to refactor when we run out of interesting things to do.

yungyuc (changes requested):
Good work!  One thing to revert:

- [x] [`wrap_SimpleArray.cpp:87`](https://github.com/solvcon/modmesh/pull/439/files#diff-260aa2069e17fc23d433c409a69e235ab3ed90bbf0362f2f88a7dcb3037b223eR87): Revert the unnecessary lambdas.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: On the Python side, the argument can be a string; we already have the example from `DataType`.

yungyuc: It can, but using a string for a C++ enum is not the most robust implementation.  It can be tolerated if the API is necessary.  We do not need the enum so it's not the case here.

ThreeMonth03: By the way, if the array is 1 dimension, it could be c-contiguous and f-contiguous simultaneously, and I'm not sure enum is a good manner.

ThreeMonth03 (commented):
* [x]  [`wrap_SimpleArray.cpp:87`](https://github.com/solvcon/modmesh/pull/439/files#diff-260aa2069e17fc23d433c409a69e235ab3ed90bbf0362f2f88a7dcb3037b223eR87): Revert the unnecessary lambdas.

Fixed. @yungyuc Could you please review this pull request?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: > By the way, if the array is 1 dimension, it could be c-contiguous and f-contiguous simultaneously, and I'm not sure enum is a good manner.

Good point.  I overlooked the 1D case.  But 1D should be a simple case.  We can leave it here and revisit for a more complete solution in the future.

yungyuc (approved):
Perfect

PR #440, opened 2024-12-14
remove the unused dockerfile and readme

kuomc:
I'm wondering if I need to delete dockerfiles that are no longer in use.


tigercosmos:
Dockerfile could be useful. Instead of deleting it, would you consider to help updating it?

yungyuc:
> Dockerfile could be useful. Instead of deleting it, would you consider to help updating it?

We discourage use of containers because they complicate debugging and deploying on bare metal.  The docker files were introduced with https://github.com/solvcon/modmesh/commit/5a0243dcb6afe07051ad8cada255151a59cbb8b9 but I do not see how they help the development of modmesh.  @dboyliao may comment.

I am fine removing them.

yungyuc (approved):

yungyuc:
Merged the PR.  If we find containers useful, we can add updated version.

ThreeMonth03:
@yungyuc By the way, should we remove some scripts in `/contrib`? It seems that we don't use **apt.ubuntu-1804-xenial.sh** in CI.

yungyuc:
That's a good idea.  Automation for dependency may go to https://github.com/solvcon/devenv

PR #441, opened 2024-12-15
add radix 2 fourier transform cpp implementation and test

### LINKED ISSUE #454: Prototype for Fourier transform
Add prototype code for Fourier transform based on `SimpleArray`.

### LINKED ISSUE #485: add Fourier transform Python wrapper and unit test
modmesh now has Fourier transform but only in C++, we need to add Python wrapper and unit test.

j8xixo12:
In this PR, I added dft and fft c++ implementation.
Also, add total energy test and delta function test for correctness checking.

### DIFF
--- a/cpp/modmesh/math.hpp
+++ b/cpp/modmesh/math.hpp
@@ -35,30 +35,33 @@ struct ComplexImpl
     T real_v;
     T imag_v;
 
-    explicit ComplexImpl(T r = 0.0, T i = 0.0)
+    explicit ComplexImpl()
+        : ComplexImpl(0.0, 0.0)
+    {
+    }
+
+    explicit ComplexImpl(T r, T i)
         : real_v(r)
         , imag_v(i)
     {
     }
+
     explicit ComplexImpl(const ComplexImpl & other)
         : real_v(other.real_v)
         , imag_v(other.imag_v)
     {
     }
 
-    // ComplexImpl addition
     ComplexImpl operator+(const ComplexImpl & other) const
     {
         return ComplexImpl(real_v + other.real_v, imag_v + other.imag_v);
     }
 
-    // ComplexImpl subtraction
     ComplexImpl operator-(const ComplexImpl & other) const
     {
         return ComplexImpl(real_v - other.real_v, imag_v - other.imag_v);
     }
 
-    // ComplexImpl multiplication
     ComplexImpl operator*(const ComplexImpl & other) const
     {
         return ComplexImpl(real_v * other.real_v - imag_v * other.imag_v, real_v * other.imag_v + imag_v * other.real_v);
@@ -69,7 +72,6 @@ struct ComplexImpl
         return ComplexImpl(real_v / other, imag_v / other);
     }
 
-    // Assignment operator
     ComplexImpl & operator=(const ComplexImpl & other)
     {
         if (this != &other) // Check for self-assignment
@@ -87,7 +89,6 @@ struct ComplexImpl
         return *this;
     }
 
-    // In-place subtraction
     ComplexImpl & operator-=(const ComplexImpl & other)
     {
         real_v -= other.real_v;
@@ -97,7 +98,7 @@ struct ComplexImpl
 
     T real() const { return real_v; }
     T imag() const { return imag_v; }
-};
+}; /* end class ComplexImpl */
 
 } /* end namespace detail */
 
--- a/cpp/modmesh/math.hpp
+++ b/cpp/modmesh/math.hpp
@@ -98,6 +98,7 @@ struct ComplexImpl
 
     T real() const { return real_v; }
     T imag() const { return imag_v; }
+    T norm() const { return real_v * real_v + imag_v * imag_v; }
 }; /* end class ComplexImpl */
 
 } /* end namespace detail */
@@ -114,8 +115,6 @@ constexpr T pi = detail::pi_v<T>;
 template <typename T>
 using Complex = detail::ComplexImpl<T>;
 
-template <template <typename> typename T1, typename T2>
-inline T2 norm(const T1<T2> & input) { return input.real() * input.real() + input.imag() * input.imag(); }
 } /* end namespace modmesh */
 
 // vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/math.hpp
+++ b/cpp/modmesh/math.hpp
@@ -26,6 +26,79 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
     return pow(base, std::integral_constant<size_t, N - 1>()) * base;
 }
 
+template <typename T>
+inline constexpr T pi_v = std::enable_if_t<std::is_floating_point_v<T>, T>(3.141592653589793238462643383279502884L);
+
+template <typename T, typename = std::enable_if_t<std::is_arithmetic<T>::value>>
+struct ComplexImpl
+{
+    T real_v;
+    T imag_v;
+
+    explicit ComplexImpl(T r = 0.0, T i = 0.0)
+        : real_v(r)
+        , imag_v(i)
+    {
+    }
+    explicit ComplexImpl(const ComplexImpl & other)
+        : real_v(other.real_v)
+        , imag_v(other.imag_v)
+    {
+    }
+
+    // ComplexImpl addition
+    ComplexImpl operator+(const ComplexImpl & other) const
+    {
+        return ComplexImpl(real_v + other.real_v, imag_v + other.imag_v);
+    }
+
+    // ComplexImpl subtraction
+    ComplexImpl operator-(const ComplexImpl & other) const
+    {
+        return ComplexImpl(real_v - other.real_v, imag_v - other.imag_v);
+    }
+
+    // ComplexImpl multiplication
+    ComplexImpl operator*(const ComplexImpl & other) const
+    {
+        return ComplexImpl(real_v * other.real_v - imag_v * other.imag_v, real_v * other.imag_v + imag_v * other.real_v);
+    }
+
+    ComplexImpl operator/(const T & other) const
+    {
+        return ComplexImpl(real_v / other, imag_v / other);
+    }
+
+    // Assignment operator
+    ComplexImpl & operator=(const ComplexImpl & other)
+    {
+        if (this != &other) // Check for self-assignment
+        {
+            real_v = other.real_v;
+            imag_v = other.imag_v;
+        }
+        return *this;
+    }
+
+    ComplexImpl & operator+=(const ComplexImpl & other)
+    {
+        real_v += other.real_v;
+        imag_v += other.imag_v;
+        return *this;
+    }
+
+    // In-place subtraction
+    ComplexImpl & operator-=(const ComplexImpl & other)
+    {
+        real_v -= other.real_v;
+        imag_v -= other.imag_v;
+        return *this;
+    }
+
+    T real() const { return real_v; }
+    T imag() const { return imag_v; }
+};
+
 } /* end namespace detail */
 
 template <size_t N, typename T>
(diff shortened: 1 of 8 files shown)


cpp/modmesh/math.hpp
```diff
@@ -26,6 +26,79 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
     return pow(base, std::integral_constant<size_t, N - 1>()) * base;
 }
 
+template <typename T>
+inline constexpr T pi_v = std::enable_if_t<std::is_floating_point_v<T>, T>(3.141592653589793238462643383279502884L);
```
yungyuc: This gives us a reason to move on to C++20: http://eel.is/c++draft/numbers

cpp/modmesh/math.hpp
```diff
@@ -26,6 +26,79 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
...
+    ComplexImpl & operator-=(const ComplexImpl & other)
+    {
+        real_v -= other.real_v;
+        imag_v -= other.imag_v;
+        return *this;
+    }
+
+    T real() const { return real_v; }
+    T imag() const { return imag_v; }
+};
```
yungyuc: Class ending mark.

cpp/modmesh/math.hpp
yungyuc: Do you need the `inline`?

cpp/modmesh/math.hpp
```diff
@@ -26,6 +26,79 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
...
+template <typename T>
+inline constexpr T pi_v = std::enable_if_t<std::is_floating_point_v<T>, T>(3.141592653589793238462643383279502884L);
+
+template <typename T, typename = std::enable_if_t<std::is_arithmetic<T>::value>>
+struct ComplexImpl
+{
+    T real_v;
+    T imag_v;
+
+    explicit ComplexImpl(T r = 0.0, T i = 0.0)
```
yungyuc: I prefer to avoid the default arguments and pass them in the default constructor:

```cpp
ComplexImpl() : ComplexImpl(0.0, 0.0) {}
```

cpp/modmesh/math.hpp
```diff
@@ -26,6 +26,79 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
...
+        , imag_v(i)
+    {
+    }
+    explicit ComplexImpl(const ComplexImpl & other)
+        : real_v(other.real_v)
+        , imag_v(other.imag_v)
+    {
+    }
+
+    // ComplexImpl addition
```
yungyuc: Please remove the unnecessary comments.  The operator overloads are self-explaining.

cpp/modmesh/math.hpp
```diff
@@ -34,6 +107,14 @@ inline constexpr T pow(T base)
...
 }
 
+template <typename T>
+inline constexpr T pi = detail::pi_v<T>;
+
+template <typename T>
+using Complex = detail::ComplexImpl<T>;
+
+template <template <typename> typename T1, typename T2>
+inline T2 norm(const T1<T2> & input) { return input.real() * input.real() + input.imag() * input.imag(); }
```
yungyuc: Why not making `norm()` a member function?

cpp/modmesh/transform/fourier.cpp
```diff
@@ -0,0 +1,29 @@
...
+        }
+    }
+    return reversed;
+}
+
+} /* end namespace detail */
+
+} /* end namespace transform */
+
+} // namespace modmesh
```
yungyuc, line 29: Should be `/* end namespace modmesh */`

cpp/modmesh/transform/fourier.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace detail
+{
+
+size_t bit_reverse(size_t n, const size_t bits);
+
+} /* end namespace detail */
+
+template <template <typename> typename T1, typename T2>
+void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
```
yungyuc, line 21: The template of template `<T1<T2>>` syntax looks so remote to me, although I understand why it is used.  Is this the only way to support complex number?

gtests/test_nopython_transform.cpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+};
+
+typedef ::testing::Types<float, double> TestTypes;
+TYPED_TEST_SUITE(ParsevalTest, TestTypes);
+TYPED_TEST_SUITE(DeltaFunctionTest, TestTypes);
+
+TYPED_TEST(ParsevalTest, fft)
+{
+    modmesh::transform::fft<modmesh::Complex, TypeParam>(this->signal, this->out);
```
yungyuc, line 90: @j8xixo12 could you please point out where `TypeParam` comes from?  I can't find it.

yungyuc (changes requested):

cpp/modmesh/math.hpp
j8xixo12: I learned from the https://en.cppreference.com/w/cpp/language/inline that `constexpr` variables with external linkage or in multiple translation units should have `inline` to avoid multiple definition error.
Although the cppreference does not mention whether template variables need inline, but I guessed it needed, so I added it.

cpp/modmesh/math.hpp
j8xixo12: Does this have additional advantages?

cpp/modmesh/math.hpp
j8xixo12: I just want to mimic the style of the c++ stdlib.

cpp/modmesh/transform/fourier.hpp
j8xixo12: hmm...., I need more time to think about it... 
This way was the first thing that came to my mind, although it looks quite complicated.

gtests/test_nopython_transform.cpp
j8xixo12: `TypeParam` is a google test type parameter used within `TYPED_TEST`, allowing user to run test cases with different type!

Ref: https://google.github.io/googletest/reference/testing.html#TYPED_TEST

gtests/test_nopython_transform.cpp
```diff
@@ -0,0 +1,111 @@
...
+        T psd_sig = 0.0, psd_out = 0.0;
+
+        for (unsigned int i = 0; i < VN; ++i)
+        {
+            psd_sig += modmesh::norm<modmesh::Complex, T>(signal[i]);
+            psd_out += modmesh::norm<modmesh::Complex, T>(out[i]);
+        }
+
+        // Expect the total energy in the time and frequency domains to be equal
+        EXPECT_NEAR(psd_sig, psd_out, (std::is_same<T, float>::value ? (T)1e-3 : (T)1e-10));
```
j8xixo12: Actually, I’m not sure how to decide the tolerance for the FFT in both `float` and `double`.

I also noticed that when the FFT uses `float`, the accumulation error increases rapidly. While I know that float loses precision, this makes me worry if there might be an issue with the implementation.

cpp/modmesh/math.hpp
yungyuc: Could you please try without `inline` here?  I think a template does not need it.  If it works without `inline`, it's better to avoid it for consistency.

cpp/modmesh/math.hpp
yungyuc: The explicit constructor delegate provides a clear place for break point when debugging.

cpp/modmesh/math.hpp
yungyuc: We don't need the C++ stdlib style for this.  The helpers may need to be exposed to Python which cannot use the C++ stdlib style.  Please make it a member function.

cpp/modmesh/transform/fourier.hpp
yungyuc: OK, we can live with it in this PR and improve later.

gtests/test_nopython_transform.cpp
yungyuc: Thanks.  New trick to me.

gtests/test_nopython_transform.cpp
yungyuc: Leave a comment for what you found and worry about it later.

cpp/modmesh/math.hpp
j8xixo12: I have tested it without `inline` and it is working, therefore I have removed it in  https://github.com/solvcon/modmesh/pull/441/commits/5b17c51fb0ae9e67e692e8dc5576b5328336c795

cpp/modmesh/math.hpp
```diff
@@ -26,6 +26,81 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
...
+    ComplexImpl & operator-=(const ComplexImpl & other)
+    {
+        real_v -= other.real_v;
+        imag_v -= other.imag_v;
+        return *this;
+    }
+
+    T real() const { return real_v; }
+    T imag() const { return imag_v; }
+    T norm() const { return real_v * real_v + imag_v * imag_v; }
```
j8xixo12, line 101: Making `norm` as a member function of `ComplexImpl`.

cpp/modmesh/transform/fourier.hpp
```diff
@@ -0,0 +1,84 @@
...
+{
+
+namespace detail
+{
+
+size_t bit_reverse(size_t n, const size_t bits);
+
+} /* end namespace detail */
+
+// TODO: The template of template is too complicate, we should find a way to make it easier.
```
j8xixo12, line 19: I left a TODO here to mention that we need to find a way to make it easier.

gtests/test_nopython_transform.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        for (unsigned int i = 0; i < VN; ++i)
+        {
+            psd_sig += signal[i].norm();
+            psd_out += out[i].norm();
+        }
+
+        // TODO: When FFT / DFT is using float, the accumulation error grows rapidly, we should fix it.
+        // Expect the total energy in the time and frequency domains to be equal
+        EXPECT_NEAR(psd_sig, psd_out, (std::is_same<T, float>::value ? (T)1e-2 : (T)1e-10));
```
j8xixo12, line 49: I left a TODO here to mention that the accumulation error grows rapidly when fft / dft are using `float`.
Also, relax the tolerance of `float` test cases, otherwise it may fail randomly.

j8xixo12:
@yungyuc this pr is ready for review!

gtests/test_nopython_transform.cpp
yungyuc: Please be explicit for how the error grows.  It is unclear what "grows rapidly" means.

cpp/modmesh/transform/fourier.hpp
yungyuc: This is informative.  Thanks.

yungyuc (changes requested):
One minor point to fix:

- [x] [`test_nopython_transform.cpp:46`](https://github.com/solvcon/modmesh/pull/441/files#diff-469d0d25b3170ce548707a94a8496d705af78368558fcb92791b34412c96b93bR46): Comment is not clear about "rapid".

gtests/test_nopython_transform.cpp
j8xixo12: I have updated the comment to include the error ranges for `float` and `double`, describing the difference between the two data types.

yungyuc:
@j8xixo12 Are you ready for the next review?

j8xixo12:
> @j8xixo12 Are you ready for the next review?

Yes, I'm ready for next review !

yungyuc:
Thanks.  When being ready for the next review, please remember to leave a global comment.  A simple comment will make it much easier for collaborators to keep track of the progress.

The updated comment at [`test_nopython_transform.cpp:44`](https://github.com/solvcon/modmesh/pull/441/files#diff-469d0d25b3170ce548707a94a8496d705af78368558fcb92791b34412c96b93bR44) can be further simplified because it is expected.

```cpp
         // TODO: When FFT / DFT uses float, the accumulation error is around 1e-3,
         // but when using double, the accumulation error is around 1e-12.
         // We need to investigate the root cause to determine whether this is an issue
         // or if there is a way to optimize it.
```

The accuracy of single precision floating point is around 1.e-6 and that of double is around 1.e-15.  Both single and double accumulate error by 3 orders of magnitude.  It's consistent.  I think 3 orders of magnitude for error are a little bit too large, but like you mentioned, we can investigate later.

I'll merge now to facilitate further development.

yungyuc (approved):

PR #448, opened 2024-12-26
Rename test_view.py to test_pilot.py

### LINKED ISSUE #400: Change the "viewer" to use an official name
The "viewer" GUI app was named like that because "viewer" is a placeholder name.  After fixing the Mac app icon (PR #398), we can start to give the GUI app an official name.

It could take a while.  This issue is created to track the progress of naming.

---

Update on 20th December 2024: I decide to use "pilot" as the new name.  Start to change the code.  (Ref: discussion #443)

yungyuc:


j8xixo12 (approved):
LGTM, but the profiler unit test failed again...

tigercosmos (approved):
LGTM

yungyuc:
> LGTM, but the profiler unit test failed again...

It should be false alarm and I will rerun to confirm.  @ThreeMonth03 we need to fix the unstable profiler tests.  We are not using a right way for timing.

PR #452, opened 2025-01-02
Complex py wrap

### LINKED ISSUE #454: Prototype for Fourier transform

### LINKED ISSUE #485: add Fourier transform Python wrapper and unit test

j8xixo12:
In this I have added modmesh `complex` implementation python wrapper, and containing python unit test code.
And it is in preparation for further fft GUI related development.

### DIFF
--- a/cpp/modmesh/python/module.cpp
+++ b/cpp/modmesh/python/module.cpp
@@ -30,9 +30,12 @@
 #include <modmesh/spacetime/pymod/spacetime_pymod.hpp>
 #include <modmesh/toggle/pymod/toggle_pymod.hpp>
 #include <modmesh/universe/pymod/universe_pymod.hpp>
+#include <modmesh/math/pymod/math_pymod.hpp>
+
 #ifdef USE_PYTEST_HELPER_BINDING
 #include <modmesh/testhelper/pymod/testbuffer_pymod.hpp>
 #endif // USE_PYTEST_HELPER_BINDING
+
 #ifdef QT_CORE_LIB
 #include <modmesh/pilot/wrap_pilot.hpp>
 #endif // QT_CORE_LIB
@@ -51,6 +54,7 @@ void initialize(pybind11::module_ mod)
     initialize_mesh(mod);
     initialize_multidim(mod);
     initialize_inout(mod);
+    initialize_math(mod);
     pybind11::module_ spacetime_mod = mod.def_submodule("spacetime", "spacetime");
     initialize_spacetime(spacetime_mod);
     pybind11::module_ onedim_mod = mod.def_submodule("onedim", "onedim");
--- a/modmesh/core.py
+++ b/modmesh/core.py
@@ -51,6 +51,8 @@ __all__ = [  # noqa: F822
     'BufferExpander',
     'Gmsh',
     'Plot3d',
+    'ComplexFloat32',
+    'ComplexFloat64',
     'SimpleArray',
     'SimpleArrayBool',
     'SimpleArrayInt8',
@@ -96,7 +98,7 @@ __all__ = [  # noqa: F822
     'Bezier3dFp64',
     'WorldFp32',
     'WorldFp64',
-    'testhelper',
+    'testhelper'
 ]
 
 
--- a/tests/test_math.py
+++ b/tests/test_math.py
@@ -1,0 +1,184 @@
+# Copyright (c) 2025, Chun-Hsu Lai <as2266317@gmail.com>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of the copyright holder nor the names of its contributors
+#   may be used to endorse or promote products derived from this software
+#   without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import unittest
+
+import numpy as np
+
+import modmesh as mm
+
+
+class ComplexTC(unittest.TestCase):
+
+    def setUp(self):
+        self.real1_32 = np.float32(0.7)
+        self.imag1_32 = np.float32(1.6)
+        self.real2_32 = np.float32(2.5)
+        self.imag2_32 = np.float32(3.4)
+
+        self.real1_64 = np.float64(4.3)
+        self.imag1_64 = np.float64(5.2)
+        self.real2_64 = np.float64(6.1)
+        self.imag2_64 = np.float64(7.0)
+
+    def test_construct_float32_default(self):
+        cplx = mm.ComplexFloat32()
+        self.assertAlmostEqual(cplx.real, 0.0)
+        self.assertAlmostEqual(cplx.imag, 0.0)
+
+    def test_construct_float64_default(self):
+        cplx = mm.ComplexFloat64()
+        self.assertAlmostEqual(cplx.real, 0.0)
+        self.assertAlmostEqual(cplx.imag, 0.0)
+
+    def test_construct_float32_random(self):
+        cplx = mm.ComplexFloat32(self.real1_32, self.imag1_32)
+        self.assertAlmostEqual(cplx.real, self.real1_32)
+        self.assertAlmostEqual(cplx.imag, self.imag1_32)
+
+    def test_construct_float64_random(self):
+        cplx = mm.ComplexFloat64(self.real1_64, self.imag1_64)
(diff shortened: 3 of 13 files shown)


cpp/modmesh/math.hpp
```diff
@@ -29,78 +30,6 @@ inline constexpr T pow(T base, std::integral_constant<size_t, N> /*unused*/)
...
-        real_v -= other.real_v;
-        imag_v -= other.imag_v;
-        return *this;
-    }
-
-    T real() const { return real_v; }
-    T imag() const { return imag_v; }
-    T norm() const { return real_v * real_v + imag_v * imag_v; }
-}; /* end class ComplexImpl */
-
```
j8xixo12: Move `ComplexImpl` into `mathtype` folder for creating python wrapper.

cpp/modmesh/mathtype/Complex.hpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        real_v -= other.real_v;
+        imag_v -= other.imag_v;
+        return *this;
+    }
+
+    T real() const { return real_v; }
+    T imag() const { return imag_v; }
+    T norm() const { return real_v * real_v + imag_v * imag_v; }
+}; /* end class ComplexImpl */
```
j8xixo12: Meanwhile, I'm trying to integrate with SimpleArray and create an array type Python wrapper and make it compatible with the NumPy buffer protocol, but I got a static assert https://github.com/solvcon/modmesh/blob/c9594da86f25b12dcacb68444b1badf8f181b1b0/cpp/modmesh/python/common.hpp#L63
Which is that `ComplexImpl<T>` is not a POD type. Currently, I don't have a firm plan to move forward.

cpp/modmesh/mathtype/pymod/wrap_Complex.cpp
```diff
@@ -0,0 +1,95 @@
...
+            .def(py::self * py::self) // NOLINT(misc-redundant-expression)
+            .def(py::self / value_type()) // NOLINT(misc-redundant-expression)
+            .def(py::self += py::self) // NOLINT(misc-redundant-expression)
+            .def(py::self -= py::self) // NOLINT(misc-redundant-expression)
+            .def_property_readonly("real", &wrapped_type::real)
+            .def_property_readonly("imag", &wrapped_type::imag)
+            .def("norm", &wrapped_type::norm)
+            .def("__complex__",
+                 [](wrapped_type const & self)
+                 { return std::complex<T>(self.real(), self.imag()); });
```
j8xixo12: Making it compatible with python built-in `complex` type and `cmath` function.

j8xixo12:
@yungyuc This PR is ready for review!

tests/test_mathtype.py
```diff
@@ -0,0 +1,186 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import unittest
+
+import numpy as np
+
+import modmesh as mm
+
+from random import random
```
yungyuc: This is in stdlib and should be grouped with `import unittest` at [`L27`](https://github.com/solvcon/modmesh/pull/452/files#diff-9e349809b47baed57bf24c28e82e3e2713396b9c63c333bc8207e23c5d92fe3fR27).

tests/test_mathtype.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class ComplexTC(unittest.TestCase):
+
+    def setUp(self):
+        self.real1_32 = np.float32(random())
```
yungyuc: Please avoid random number in unit tests unless there is not another way.  Using random number introduces a hidden dependency.  If random number goes wrong, the tests here will issue false alarm.  The false alarm is noise that burdens maintenance.

The random number is not printed in code and maintainers need to print it themselves when troubleshooting.  This is another burden.

---

In this case, there are ways other than generating random number here, so random number should not be used.

You can randomly generate fixture and print the number here.

cpp/modmesh/math.hpp
yungyuc, line 1: You should move `cpp/modmesh/math.hpp` to `cpp/modmesh/math/math.hpp`.  The directory should be named as `cpp/modmesh/math/`, not `cpp/modmesh/mathtype/`.

tests/test_mathtype.py
yungyuc, line 1: After changing the new directory to be `cpp/modmesh/math/`, this should be renamed as `test_math.py`.

modmesh/core.py
```diff
@@ -97,6 +97,8 @@
     'WorldFp32',
     'WorldFp64',
     'testhelper',
+    'ComplexFloat32',
```
yungyuc: The two lines should be placed before `SimpleArray` because they are more fundamental.

cpp/modmesh/mathtype/mathtype.hpp
yungyuc, line 1: This should be replaced by moving `cpp/modmesh/math.hpp` here.

cpp/modmesh/mathtype/Complex.hpp
yungyuc: The ending mark should be `/* end structure ComplexImpl */`.

yungyuc: `ComplexImpl<T>` looks like POD.  What does it mean to say t is not a POD type?

yungyuc (changes requested):
Good move.  I would like to have tightened structure of the code:

- [x] [`cpp/modmesh/math.hpp`](https://github.com/solvcon/modmesh/pull/452/files#diff-78108ba95c2dee0342122eb89fdbf30026664c0db8a14837f0a3de5a2236927b): Use `cpp/modmesh/math/` to conform to the existing name `cpp/modmesh/math.hpp`.  Do not create another directory name (`mathtype`).
- [x] [`test_mathtype.py`](https://github.com/solvcon/modmesh/pull/452/files#diff-9e349809b47baed57bf24c28e82e3e2713396b9c63c333bc8207e23c5d92fe3fR39): Random number is not necessary.  Don't use it.

j8xixo12:
@yungyuc This PR is ready for the next review.

I have made the following change:

- [x] Rename `mathtype` folder to `math` and move `cpp/modmesh/math.hpp` to `cpp/modmesh/math`
- [x] Rename `test_mathtype.py` to `test_math.py`
- [x] Using fix number instead of random number in `test_math.py`
- [x] Moving `ComplexFloat32` and `ComplexFloat64` to before `SimpleArray`
- [x] Fix typo in `cpp/modmesh/math/Complex.hpp` struct end mark

For the array type of complex number, I plan to use another PR.

cpp/modmesh/python/module.cpp
```diff
@@ -30,6 +30,7 @@
 #include <modmesh/spacetime/pymod/spacetime_pymod.hpp>
 #include <modmesh/toggle/pymod/toggle_pymod.hpp>
 #include <modmesh/universe/pymod/universe_pymod.hpp>
+#include <modmesh/math/pymod/math_pymod.hpp>
```
yungyuc, line 33: Need a blank line.  (This issue appears at multiple locations.  Please fix them all.)

tests/test_math.py
```diff
@@ -0,0 +1,184 @@
...
+
+import modmesh as mm
+
+
+class ComplexTC(unittest.TestCase):
+
+    def setUp(self):
+        self.real1_32 = np.float32(1.0)
+        self.imag1_32 = np.float32(1.0)
+        self.real2_32 = np.float32(1.0)
```
yungyuc: real/imag 1/2 are both 1+i.  Please use use different value.

yungyuc (changes requested):
> For the array type of complex number, I plan to use another PR.

Sounds like a plan.  Still having issues to be worked out:

- [x] [`cpp/modmesh/python/module.cpp:33`](https://github.com/solvcon/modmesh/pull/452/files#diff-f1a599ca51b05efce46b38c48ddcce743d6b88d674294bfe9da2e2eb47b69609R33): Formatting: missing blank lines after inclusion.
- [x] [`tests/test_math.py:39`](https://github.com/solvcon/modmesh/pull/452/files#diff-1066dfb0a00e170b5e4209e4d390334859589c6b0acda171b325ced1a11e28c7R39): Fixture should not use duplicate value.
- [x] Please squash.  This PR does not need so many commits.

cpp/modmesh/python/module.cpp
j8xixo12: I also found that some include order do not follow the [LLVM include style](https://llvm.org/docs/CodingStandards.html#include-style), for example https://github.com/solvcon/modmesh/blob/c9594da86f25b12dcacb68444b1badf8f181b1b0/cpp/modmesh/spacetime/io.hpp#L8

should I fix them ?

yungyuc: Thanks for spotting the inclusion order.  It should not be done with the enhancement PR.

We are not strictly following inclusion order.  It's not important atm, but we should optimize the inclusion at some point.  If anyone cares about it, it can be done in a standalone PR.

tests/test_math.py
```diff
@@ -0,0 +1,184 @@
...
+    def setUp(self):
+        self.real1_32 = np.float32(0.7)
+        self.imag1_32 = np.float32(1.6)
+        self.real2_32 = np.float32(2.5)
+        self.imag2_32 = np.float32(3.4)
+
+        self.real1_64 = np.float64(4.3)
+        self.imag1_64 = np.float64(5.2)
+        self.real2_64 = np.float64(6.1)
+        self.imag2_64 = np.float64(7.0)
```
j8xixo12, line 45: Using different number for fixtures.

j8xixo12 (commented):
@yungyuc This PR is ready for the next review.

I have squashed commits, fixed missing blank lines after inclusion and using different number for test fixtures.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -29,6 +29,7 @@
 #include <modmesh/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
 
 #include <modmesh/buffer/pymod/array_common.hpp>
+
```
yungyuc: Do not change formatting in unrelated files.  Adding/removing blank lines is refactoring.  Refactoring change should not be included in the same PR as feature development.

tests/test_math.py
yungyuc: This is good.

yungyuc (changes requested):
- [ ] Do not change blank lines in unrelated files.

Please revert all unnecessary style changes and squash.

tigercosmos:
Do you have any plan to support complex array?

yungyuc:
> Do you have any plan to support complex array?

It has been [discussed](https://github.com/solvcon/modmesh/pull/452#issuecomment-2571266849) and the conclusion is to handle it in another PR.

> [...]
> For the array type of complex number, I plan to use another PR.

j8xixo12:
@yungyuc This PR is ready for the next review, I have reverted all unnecessary style changes and only applied style change within related files.

modmesh/core.py
```diff
@@ -96,7 +98,7 @@
     'Bezier3dFp64',
     'WorldFp32',
     'WorldFp64',
-    'testhelper',
+    'testhelper'
```
yungyuc, line 101: There should be a comma after each line in a Python list.  Please be careful about it next time.

yungyuc (approved):
It now looks good.  Thanks for the work.

modmesh/core.py
j8xixo12: Thanks, I will be careful about this next time.

PR #453, opened 2025-01-05
[ci] upgrade Qt version to 6.8.1 LTS

terrychan999:
In this PR, we upgrade the Qt version to 6.8.1 LTS for Ubuntu and macOS, aligning all platforms to the latest LTS version.


yungyuc:
Thanks, @terrychan999 .

Issue #435, opened 2024-11-05
Develop helpers for TimeSeriesDataFrame to sort data and reorder data


xingularity:
(Updated on 9th Nov)

To provide ordered data from a data frame potentially storing large volume of data, efficient sorting capability needs to be built. It can be built by providing `sort()` and `argsort()` helper functions on `SimpleArray`, and a `reorder()` function is provided in the data frame class in Python. A proposed sequence of ordering data is [here](https://github.com/solvcon/modmesh/issues/435#issuecomment-2466003695).

The `SimpleArray.sort()` and `SimpleArray.argsort()` should be provided in both Python and C++, and the Python functions are just a wrapper of C++ implementation. The sorting function should work like [numpy.ndarray.sort](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.sort.html#numpy.ndarray.sort) but provide both in-place and out-of-place options. `SimpleArray.argsort()` should be provided only out-of-place.

The reordering helper should shuffle one, multiple selected, or all columns in a data frame.

- [x] `SimpleArray.sort()`
- [x] `SimpleArray.argsort()`
- [ ] `DataFrame.reorder()`

---

(Initiated on 6th Nov)

# Original statement

TimeSeriesDataFrame is to provide data in correct sequence order by index/timestamp. Current prototype implementation only reads text and organizes data in a columnar format, it does not guarantee the sequence of the data when retrieving the data. A sorting algorithm is required to guarantee the order of data.

yungyuc:
This issue is to continue the discussion in https://github.com/solvcon/modmesh/discussions/380#discussioncomment-11154329 .  @j8xixo12 , I am not entirely sure what does _sequence of data_ mean.  Could you please provide a definition to help specification?

xingularity:
> This issue is to continue the discussion in [#380 (comment)](https://github.com/solvcon/modmesh/discussions/380#discussioncomment-11154329) . @j8xixo12 , I am not entirely sure what does _sequence of data_ mean. Could you please provide a definition to help specification?

Hi @yungyuc 

I think I might not explain clear enough in the beginning of this issue. The data provided by this DataFrame should be ordered by index or timestamp, this is the correct sequence which user can expect. Current implementation does not guarantee this. We need to enhance it.

yungyuc:
> The data provided by this DataFrame should be ordered by index or timestamp, this is the correct sequence which user can expect. Current implementation does not guarantee this. We need to enhance it.

[...]

PR #456, opened 2025-01-22
Implement `sort`, `argsort` and `take_along_axis` for one dimensional `SimpleArray`

KHLee529:
# Description

Basic sorting functions, `sort` and `argsort`, requested in issue #435 are implemented in this PR. 

### `SimpleArray::sort`

This C++ function is the inplace sorting function for `SimpleArray`. It currently supports only 1 dimensional array. 
It basically just wrap the `std::sort` from C++ STL with dimension check.

### `SimpleArray::argsort`

This is the function that returns an array in which the indices are stored in the order that the array is sorted. 
For example, if the array is `arr = [1, 9, 7, 3, 5]`, the returned array will be `[0, 3, 4, 2, 1]` indicating that `arr[0]->arr[3]->arr[4]->arr[2]->arr[1]` is the sorted order of `arr`.

### `SimpleArray::take_along_axis`

This function is to apply the result that returned by `argsort` of another `SimpleArray` to make it sorted based on another array.
A concrete example is shown below.

```c++
constexpr size_t N = 10;
SimpleArray<uint64_t> array(N);
SimpleArray<uint64_t> other(N);

// ... add data to arrays

other.take_along_axis(array.argsort());
```

## Current Problem

- [ ] How to implement a wrapped python function returns `Option[SimpleArray]` in C++. as the prototype of `sort` are intended to be the following as mentioned in issue #435?
  ```python
  sarr = SimpleArrayUint32()
  assert(sarr.sort(inplace=True) is None)
  assert(sarr.sort(inplace=False).shape == sarr.shape)
  ```
- [x] How should the python prototype of `argsort` be like?
- [ ] How should the python prototype of `TimeSeriesDataFrame.reorder` be like?

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -34,6 +34,7 @@
 #include <stdexcept>
 #include <functional>
 #include <numeric>
+#include <algorithm>
 
 #if defined(_MSC_VER)
 #include <BaseTsd.h>
@@ -42,6 +43,9 @@ typedef SSIZE_T ssize_t;
 
 namespace modmesh
 {
+
+template <typename T>
+class SimpleArray; // forward declaration
 
 namespace detail
 {
@@ -194,6 +198,38 @@ public:
     }
 }; /* end class SimpleArrayMixinCalculators */
 
+template <typename A, typename T>
+class SimpleArrayMixinSort
+{
+
+private:
+
+    using internal_types = detail::SimpleArrayInternalTypes<T>;
+
+public:
+
+    using value_type = typename internal_types::value_type;
+
+    void sort(void);
+    SimpleArray<uint64_t> argsort(void);
+    template <typename I>
+    A take_along_axis(SimpleArray<I> const & indices);
+
+}; /* end class SimpleArrayMixinSort */
+
+template <typename A, typename T>
+void SimpleArrayMixinSort<A, T>::sort(void)
+{
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray::sort(): currently only support 1D array but the array is "
+                                             << athis->ndim() << " dimension");
+    }
+
+    std::sort(athis->begin(), athis->end());
+}
+
 } /* end namespace detail */
 
 /**
@@ -205,6 +241,7 @@ template <typename T>
 class SimpleArray
     : public detail::SimpleArrayMixinModifiers<SimpleArray<T>, T>
     , public detail::SimpleArrayMixinCalculators<SimpleArray<T>, T>
+    , public detail::SimpleArrayMixinSort<SimpleArray<T>, T>
 {
 
 private:
@@ -524,6 +561,28 @@ public:
         return data(offset);
     }
 
+    shape_type first_sidx() const noexcept
+    {
+        return shape_type(shape().size(), 0);
+    }
+
+    bool next_sidx(shape_type & sidx) const noexcept
+    {
+        ssize_t dim = shape().size() - 1;
+        while (dim >= 0)
+        {
+            sidx[dim] += 1;
+            if (sidx[dim] == shape()[dim])
+            {
+                sidx[dim] = 0;
+                dim -= 1;
+                continue;
+            }
+            return true;
+        }
+        return false;
+    }
+
     size_t ndim() const noexcept { return m_shape.size(); }
     shape_type const & shape() const { return m_shape; }
     size_t shape(size_t it) const noexcept { return m_shape[it]; }
@@ -728,6 +787,68 @@ private:
     size_t m_nghost = 0;
     value_type * m_body = nullptr;
 }; /* end class SimpleArray */
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSort<A, T>::argsort(void)
+{
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray::argsort(): currently only support 1D array"
+                                             << " but the array is " << athis->ndim() << " dimension");
+    }
+
+    SimpleArray<uint64_t> ret(athis->shape());
+
+    { // Return array initialization
+        uint64_t cnt = 0;
+        std::for_each(ret.begin(), ret.end(), [&cnt](uint64_t & v)
+                      { v = cnt++; });
+    }
+
+    value_type const * buf = athis->body();
+    auto cmp = [buf](uint64_t a, uint64_t b)
+    {
+        return buf[a] < buf[b];
+    };
+    std::sort(ret.begin(), ret.end(), cmp);
+    return ret;
+}
+
+template <typename A, typename T>
+template <typename I>
+A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & indices)
+{
+    static_assert(std::is_integral_v<I>, "I must be integral type");
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray::take_along_axis(): currently only support 1D array"
+                                             << " but the array is " << athis->ndim() << " dimension");
+    }
+
+    A ret(indices.shape());
+
+    auto sidx = indices.first_sidx();
+    do {
+        I idx = indices.at(sidx);
+        if (idx < 0 || idx > athis->shape()[0])
+        {
+            Formatter err_msg;
+            err_msg << "SimpleArray::take_along_axis(): indices[" << sidx[0];
+            for (size_t dim = 1; dim < sidx.size(); ++dim)
+            {
+                err_msg << ", " << sidx[dim];
+            }
+            err_msg << "] is " << idx << ", which is out of range of the array size "
+                    << athis->shape()[0];
+
+            throw std::out_of_range(err_msg);
+        }
+        ret.at(sidx) = (*athis)[static_cast<size_t>(idx)];
+    } while (indices.next_sidx(sidx));
+    return ret;
+}
 
 template <typename S>
 using is_simple_array = std::is_same<
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -154,6 +154,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                                    { return pybind11::cast(SimpleArrayPlex(arr)); })
             .wrap_modifiers()
             .wrap_calculators()
+            .wrap_sort()
             // ATTENTION: always keep the same interface between WrapSimpleArrayPlex and WrapSimpleArray
             ;
     }
@@ -185,6 +186,25 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("max", &wrapped_type::max)
             .def("sum", &wrapped_type::sum)
             .def("abs", &wrapped_type::abs)
+            //
+            ;
+
+        return *this;
+    }
+
+    wrapper_type & wrap_sort()
+    {
+        namespace py = pybind11; // NOLINT(misc-unused-alias-decls)
+
+        (*this)
+            .def("sort", &wrapped_type::sort)
+            .def(
+                "argsort",
+                [](wrapped_type & self)
+                { return pybind11::cast(self.argsort()); })
+            .def("take_along_axis",
+                 [](wrapped_type & self, py::object const & indices)
+                 { return pybind11::cast(self.take_along_axis(indices.cast<SimpleArrayUint64>())); })
             //
             ;
 
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -824,6 +824,80 @@ class SimpleArrayBasicTC(unittest.TestCase):
         self.assertEqual(
             str(type(arrayplex_int32_2)), "<class '_modmesh.SimpleArray'>")
 
+    def test_sort(self):
+        # Note: tests on array with ghost indices should be added
+        #       in the future
+
+        test_data = [
+            [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+            [10, 9, 8, 7, 6, 5, 4, 3, 2, 1],
+            [1, 5, 10, 2, 6, 9, 7, 8, 4, 3],
+            [1,  0,  1, -3, -4, -1,  1,  9,  5, -4],
+            [-1.3, -4.8,  1.5,  0.3,  7.1,  2.5,  4.8, -0.1,  9.4,  7.6]
+        ]
+
+        def _check(arr, use_float=False):
+            if use_float:
+                narr = np.array(arr, dtype='float64')
+                sarr = modmesh.SimpleArrayFloat64(array=narr)
+            else:
+                narr = np.array(arr, dtype='int32')
+                sarr = modmesh.SimpleArrayInt32(array=narr)
+
+            args = sarr.argsort()
+            for i in range(1, len(args)):
+                self.assertLessEqual(sarr[args[i]], sarr[args[i]])
+
+            sorted_arr = sarr.take_along_axis(args)
+            for i in range(1, len(sorted_arr)):
+                self.assertLessEqual(sorted_arr[i - 1], sorted_arr[i])
+
+            sarr.sort()
+            for i in range(1, len(sarr)):
+                self.assertLessEqual(sarr[i - 1], sarr[i])
+
+        _check(test_data[0])
+        _check(test_data[1])
+        _check(test_data[2])
+        _check(test_data[3])
+        _check(test_data[4], True)
+
+    def test_talk_along_axis(self):
+        data = [1, 5, 10, 2, 6, 9, 7, 8, 4, 3]
+        narr = np.array(data, dtype='int32')
+        data_arr = modmesh.SimpleArrayInt32(array=narr)
+
+        # test 1-D indices
+        idx = [0, 2, 4, 6, 1, 3, 5, 7]
+        narr = np.array(idx, dtype='uint64')
+        idx_arr = modmesh.SimpleArrayUint64(array=narr)
+
+        ret_arr = data_arr.take_along_axis(idx_arr)
+        for i in range(len(idx)):
+            self.assertEqual(ret_arr[i], data[idx[i]])
+
+        # test 2-D indices
+        idx = [[0, 2, 4, 6], [1, 3, 5, 7]]
+        narr = np.array(idx, dtype='uint64')
+        idx_arr = modmesh.SimpleArrayUint64(array=narr)
+
+        ret_arr = data_arr.take_along_axis(idx_arr)
+        for i in range(len(idx)):
+            for j in range(len(idx[i])):
+                self.assertEqual(ret_arr[i, j], data[idx_arr[i, j]])
+
+        # test out-of-range index
+        idx = [[0, 1], [2, 3], [4, 20]]
+        narr = np.array(idx, dtype='uint64')
+        idx_arr = modmesh.SimpleArrayUint64(array=narr)
+
+        with self.assertRaisesRegex(
+            IndexError,
+            r"SimpleArray::take_along_axis\(\): indices\[2, 1\] is 20, " +
+            "which is out of range of the array size 10"
+        ):
+            ret_arr = data_arr.take_along_axis(idx_arr)
+
 
 class SimpleArrayCalculatorsTC(unittest.TestCase):
 


KHLee529:
@yungyuc Here's the basic implementation of the sorting functions. Please take a look and give me some feedback when you are available.

yungyuc:
> * [ ]  How to implement a wrapped python function returns `Option[SimpleArray]` in C++. as the prototype of `sort` are intended to be the following as mentioned in issue [Develop helpers for TimeSeriesDataFrame to sort data and reorder data #435](https://github.com/solvcon/modmesh/issues/435)?
>   ```python
>   sarr = SimpleArrayUint32()
>   assert(sarr.sort(inplace=True) is None)
>   assert(sarr.sort(inplace=False).shape == sarr.shape)
>   ```

You can first focus on the in-place version of the wrapper for sorting and worry about the out-of-place version that returns an array.

> * [ ]  How should the python prototype of `argsort` and `TimeSeriesDataFrame.reorder` be like?

Please review how numpy implements `argsort()` API.

KHLee529:
> Please review how numpy implements `argsort()` API.

@yungyuc Thanks for advice, I've wrapped the `argsort` function.

yungyuc:
Please pass all CI runs and add annotations inline.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
         }
     }
 
+    void sort(void)
```
KHLee529: The inplace sorting function that currently support only 1-D `SimpleArray`

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+    {
+        if (ndim() != 1)
+        {
+            throw std::runtime_error("SimpleArray: Sorting is only supported in 1D array.");
+        }
+
+        std::sort(begin(), end());
+    }
+
+    SimpleArray<uint64_t> argsort(void)
```
KHLee529: The function that returns a SimpleArray containing the indices which will sort the array

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+        if (ndim() != 1)
+        {
+            throw std::runtime_error("SimpleArray: Sorting is only supported in 1D array.");
+        }
+
+        SimpleArray<uint64_t> ret(shape());
+
+        { // Return array initialization
+            uint64_t cnt = 0;
+            std::for_each(ret.begin(), ret.end(), [&cnt](uint64_t & v)
```
KHLee529: Initialize the array to be returned to the ascending indices

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+                          { v = cnt++; });
+        }
+
+        value_type const * buf = body();
+        auto cmp = [buf](uint64_t a, uint64_t b)
```
KHLee529: A comparison function that take the input as indices of the SimpleArray and compare the value of the SimpleArray at the given index.
Use this function as the comparison function for `std::sort` can make the return index array sorted by the data stored in the SimpleArray

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+        {
+            return buf[a] < buf[b];
+        };
+        std::sort(ret.begin(), ret.end(), cmp);
+        return ret;
+    }
+
+    void apply_argsort(SimpleArray<uint64_t> const & sorted_args)
```
KHLee529: A function that takes the return value of `argsort` of another SimpleArray and reorder current SimpleArray

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+        if (shape()[0] != sorted_args.shape()[0])
+        {
+            throw std::runtime_error("SimpleArray: argsort only support same shape");
+        }
+        if (shape()[0] < 2)
+        {
+            return;
+        }
+
+        std::vector<bool> applied_arg(shape()[0], false);
```
KHLee529: A data storage that stores whether the indices in `sorted_args` is already applied.
When this array is all `true`, the index application is done.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+        {
+            for (auto i : vec)
+            {
+                if (i == false)
+                    return false;
+            }
+            return true;
+        };
+
+        auto next = [](std::vector<bool> & vec, ssize_t last)
```
KHLee529: A helper function that returns the next index not yet applied

KHLee529:
@yungyuc @ThreeMonth03 please review this PR. 
I think the implementation of `TimeSeriesDataFrame.reorder` function might be better to move to next stage, so that this PR can be treated as done with current functionality.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: You might consider using the composition design like ```SimpleArrayMixinCalculators```.
https://github.com/solvcon/modmesh/blob/b6adfc94e8cc89ab0cf22d375110ace195432c0e/cpp/modmesh/buffer/SimpleArray.hpp#L122

yungyuc: I concur.  This helper should be put in the calculator mixin.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
         }
     }
 
+    void sort(void)
+    {
+        if (ndim() != 1)
+        {
+            throw std::runtime_error("SimpleArray: Sorting is only supported in 1D array.");
```
yungyuc: Add the dimension number in the error message.  See the example: https://github.com/solvcon/modmesh/blob/74a3a0f7e91f85638303457bc755cd0fd51edc59/cpp/modmesh/buffer/SimpleArray.hpp#L306

Do not capitalize after `:`.  That is, write `SimpleArray: sorting`.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: I concur.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Could you please point me to the document link to the corresponding API in numpy?

yungyuc: The implementation of `apply_argsort()` has too many lines and need to be taken outside the class declaration, like:

```cpp
class A {
void helper();
};

A::helper()
{
    // long implementation.
}
```

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Please add a comment to indicate that this only support 1D array.  In the same time, please prepare for implementing for multi-dimensional array sorting.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+            return;
+        }
+
+        std::vector<bool> applied_arg(shape()[0], false);
+
+        auto all = [](std::vector<bool> & vec)
+        {
+            for (auto i : vec)
+            {
+                if (i == false)
```
yungyuc: Always use curly braces for blocking, e.g.,:

```cpp
if (i == false)
{
    return false;
}
```

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+            throw std::runtime_error("SimpleArray: argsort only support same shape");
+        }
+        if (shape()[0] < 2)
+        {
+            return;
+        }
+
+        std::vector<bool> applied_arg(shape()[0], false);
+
+        auto all = [](std::vector<bool> & vec)
```
yungyuc: Use `std::any_of()`: https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+            {
+                if (i == false)
+                    return false;
+            }
+            return true;
+        };
+
+        auto next = [](std::vector<bool> & vec, ssize_t last)
+        {
+            for (ssize_t i = last; i < static_cast<ssize_t>(vec.size()); i++)
```
yungyuc: Prefer pre-increment `++i` instead of post-increment `i++`.  They mean differently.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +586,102 @@ class SimpleArray
...
+            {
+                if (vec[i] == false)
+                    return i;
+            }
+            return static_cast<ssize_t>(-1);
+        };
+
+        ssize_t idx = 0;
+        while (!all(applied_arg))
```
yungyuc: No! This call to `all()` makes $O(n^2)$.  Calling to `next()` gives unnecessary big complexity too.  Applying index should be $O(n)$.

yungyuc (changes requested):
You do not have a test case for `apply_argsort()` and the time complexity is horrible.  If you are not ready to have the function reviewed, remove it from the PR.

- [x] Move the `sort()` helper to `SimpleArrayMixinCalculators`.
- [x] Add dimension number to the error message.
- [x] Confirm API consistency of `apply_argsort()` to numpy.
- [x] The implementation of `apply_argsort()` has too many lines and need to be taken outside the class declaration.
- [x] Always use curly braces for blocking.
- [x] Use `std::any_of()`: https://en.cppreference.com/w/cpp/algorithm/all_any_none_of
- [x] Prefer pre-increment `++i` instead of post-increment `i++`.  They mean differently.
- [ ] `apply_argsort()` applies sorted index and should be $O(n)$.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: > Could you please point me to the document link to the corresponding API in numpy?

This function is intend to achieve the functionality of [`numpy.take_along_axis`](https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html).

KHLee529: > The implementation of `apply_argsort()` has too many lines and need to be taken outside the class declaration, like:

What I consider moving the implementation out from the declaration is to move the implementation to `SimpleArray.cpp`, but I'm wondering if I can move this out because it is still a template function.

If you mean just move the implementation from the class declaration to somewhere beneath in the `SimpleArray.hpp`. Then I have no problem about it.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: OK

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: Got it

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: I found that `all()` is $O(n^2)$, I'll fix it.

However, I consider `next()` a $O(n)$ function. Since it always start searching from the position got last time, every element will only be checked once.

yungyuc: I think apply argsort indices should be $O(n)$, just like how it behaves in numpy.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: > If you mean just move the implementation from the class declaration to somewhere beneath in the `SimpleArray.hpp`. Then I have no problem about it.

This is what I meant: to move the implementation outside the class declaration but keep it in the header file.

yungyuc: > > Could you please point me to the document link to the corresponding API in numpy?
> 
> This function is intend to achieve the functionality of [`numpy.take_along_axis`](https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html).

What is the rationale to name it otherwise but not `take_along_axis`?

KHLee529:
@yungyuc all the problems you mentioned has been fixed except for the one related to complexity.
I'll add comment about my question on your comment on that code.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: I've remove the `all()` judgement and use `next()` for the only termination condition of the loop. And I think this function then achieve $O(n)$ time complexity. Could you please explain more detail if it is not what I thought

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: `argsort()` is of the same group as `sort()` and should also go to `SimpleArrayMixinCalculators`.

In PR we do not add comments for every detail.  Please be comprehensive when addressing review comments.

yungyuc: And the function is too long and should be moved outside the class declaration.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +769,64 @@ class SimpleArray
     value_type * m_body = nullptr;
 }; /* end class SimpleArray */
 
+template <typename T>
+void SimpleArray<T>::take_along_axis(SimpleArray<uint64_t> const & indices)
+{
+    if (indices.ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray: sorting is only supported in 1D array. "
```
yungyuc: Please reword to be like `SimpleArray::take_along_axis() supports only 1D array but the array is N dimension` (replacing N as the variable).

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +769,64 @@ class SimpleArray
...
+                                             << "Array size " << shape()[0] << " != indices shape " << indices.shape()[0]);
+    }
+    if (shape()[0] < 2)
+    {
+        return;
+    }
+
+    std::vector<bool> applied_arg(shape()[0], false);
+
+    auto next = [](std::vector<bool> & vec, ssize_t last)
```
yungyuc: Do not use `next()`.  It prevents you from making $O(n)$ complexity.  Please ask for help from @ThreeMonth03 .

tests/test_buffer.py
```diff
@@ -824,6 +824,17 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
         self.assertEqual(
             str(type(arrayplex_int32_2)), "<class '_modmesh.SimpleArray'>")
 
+    def test_sort(self):
```
yungyuc, line 827: Test `take_along_axis()` too, since you implemented it.

yungyuc (changes requested):
- [x] `argsort()` is of the same group as `sort()` and should also go to `SimpleArrayMixinCalculators`.
- [x] Please reword to be like `SimpleArray::take_along_axis() supports only 1D array but the array is N dimension` (replacing N as the variable).
- [x] `apply_argsort()` should be $O(n)$.
- [x] Test `take_along_axis()` too, since you implemented it.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: I think although the time complexity of the `next()` is O(n), the total time complexiy of `void SimpleArray<T>::take_along_axis(SimpleArray<uint64_t> const & indices)` is O(n).
It is because `while ((idx = next(applied_arg, idx)) != -1)` only visit every number once from 0 to n.

By the way, here is the analysis of the gpt-o3-mini.
https://chatgpt.com/share/67a1fc66-a350-800a-910b-73f4a845664b

ThreeMonth03: It is prefered to use  `std::vector<char> & vec` or `std::bitset<n> vec` instead of `std::vector<bool> & vec`. The performance of `std::vector<bool> & vec` is awful.

yungyuc: Agree the overall complexity should still be $O(n)$, but the split of nested loop with the inner one in the lambda is hard to read.

@KHLee529 , please avoid the lambda.  It will make the code clearer and better optimized.  The use of lambda also makes it difficult if not impossible for SIMD.

Please also avoid bound-checking as much as possible.

Is there a way to reduce the loop and/or avoid the nesting?

ThreeMonth03: > Agree the overall complexity should still be O ( n ) , but the split of nested loop with the inner one in the lambda is hard to read.
> 
> @KHLee529 , please avoid the lambda. It will make the code clearer and better optimized. The use of lambda also makes it difficult if not impossible for SIMD.
> 
> Please also avoid bound-checking as much as possible.
> 
> Is there a way to reduce the loop and/or avoid the nesting?

I notice that the space complexity of this function is O(n). If we directly create the new buffer `vector<T> buffer` and rearrange the elements in the `vector<T> buffer`, the space complexity is also O(n). Additionally, we could make the code clear and straight forward.

KHLee529:
@yungyuc Problems mentioned have been fixed. Please review the modification.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -585,6 +593,9 @@ class SimpleArray
         }
     }
 
+    template <typename I>
+    SimpleArray take_along_axis(SimpleArray<I> const & indices);
```
yungyuc: It should not be too difficult to infer that `take_along_axis()` should go to `SimpleArrayMixinCalculators` with `sort()` and `argsort()` because `take_along_axis()` takes the results produced by `argsort()`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +740,81 @@ class SimpleArray
...
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray: take_along_axis() supports same "
+                                                "shape of indices and array. Array size "
+                                             << shape()[0] << " != indices array size " << indices.shape()[0]);
+    }
+
+    SimpleArray<T> ret(*this);
+    for (size_t i = 0; i < shape()[0]; ++i)
+    {
+        ret.at(i) = at(static_cast<size_t>(indices[i]));
```
yungyuc: The bound-checking is not necessary.  Please use a smarter way to implement the element accessing without the bound-checking.

tests/test_buffer.py
```diff
@@ -824,6 +824,21 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
...
+        narr = np.random.randint(0, 100, 20, dtype='int32')
```
yungyuc: Never use random number in testing.  Test with random number lacks reproducibility unless you are very careful about seeding (but you didn't).

This is a simple algorithm and you can test with carefully crafted fixture.

yungyuc (changes requested):
- [x] Put things in relevant mixin classes.
- [ ] Careful about bound-checking, which is slow.
- [x] Carefully devise fixture for testing.  Never use random number.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -194,6 +198,25 @@ class SimpleArrayMixinCalculators
     }
 }; /* end class SimpleArrayMixinCalculators */
 
+template <typename A, typename T>
+class SimpleArrayMixinSorters
```
KHLee529: In order to keep the cleanness of interfaces used by `SimpleArrayMixinCalculators`, Sorting functions are isolated to another mixin.

KHLee529:
@yungyuc I fixed the problem you mentioned. Please review them.

In pytest, I have no idea how to make good test cases so I manually made some trivial cases. Please tell me if you have any idea how better test cases should be designed.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: I like a standalone mixin for sorting.  Could you please name it `SimpleArrayMixinSort`?  Sorter sounds like a clerk to me.  I know this is subjective, but I like it to use just "sort".

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -194,6 +198,25 @@ class SimpleArrayMixinCalculators
...
+public:
+
+    using value_type = typename internal_types::value_type;
+
+    void sort(void);
+    SimpleArray<uint64_t> argsort(void);
+    template <typename I>
+    A take_along_axis(SimpleArray<I> const & indices);
+
+}; /* end class SimpleArrayMixinCalculators */
```
yungyuc: Correct the name in the class ending mark.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -205,6 +228,7 @@ template <typename T>
 class SimpleArray
     : public detail::SimpleArrayMixinModifiers<SimpleArray<T>, T>
     , public detail::SimpleArrayMixinCalculators<SimpleArray<T>, T>
+    , public detail::SimpleArrayMixinSorters<SimpleArray<T>, T>
```
yungyuc: Beautiful.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +753,68 @@ class SimpleArray
     value_type * m_body = nullptr;
 }; /* end class SimpleArray */
 
+template <typename A, typename T>
```
yungyuc, line 791: Please make the function implementation/definition adjacent to the class declaration of the mixin.  This new mixin is the only one that has function definition outside the class declaration.

tests/test_buffer.py
```diff
@@ -824,6 +824,28 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
         self.assertEqual(
             str(type(arrayplex_int32_2)), "<class '_modmesh.SimpleArray'>")
 
+    def test_sort(self):
+        test_data = [
+            [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+            [10, 9, 8, 7, 6, 5, 4, 3, 2, 1],
```
yungyuc, line 833: Please add one fixture (array) with negative value and zero, and another with real number (floating point).

yungyuc: And please avoid using a for loop in the test code.  Use inner function with unrolled loop instead:

```python
def _check(arr):
    narr = np.array(arr, dtype='int32')
    ...

# Unroll the loop.
_check(test_data[0])
_check(test_data[1])
_check(test_data[2])
```

Loops in test make troubleshooting hard.  If one of the fixture fails, the stack trace does not tell you which one.  But using the unrolled loop like above, the stack trace tells you exactly which fixture is bad.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +753,68 @@ class SimpleArray
...
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray: take_along_axis() supports currently only in 1D array "
+                                                " but the array is "
+                                             << athis->ndim() << " dimension");
+    }
+
+    SimpleArray<T> ret(indices.shape());
+    std::transform(indices.begin(), indices.end(), ret.begin(), [athis](I idx)
```
yungyuc: The use of transform is less clear than the previous implementation that uses an explicit loop.  If there is not runtime benefit of transform (with benchmark to prove), please revert to the explicit loop.

And don't do bound checking (`athis->at()`).  You know exactly how many elements to process.

yungyuc (changes requested):
> In pytest, I have no idea how to make good test cases so I manually made some trivial cases. Please tell me if you have any idea how better test cases should be designed.

Your test cases are not trivial.  They are simple and work well.

Things to change:

- [x] Name the mixin as `SimpleArrayMixinSort`.
- [ ] Make the function implementation/definition to immediately follow the class declaration of the mixin.
- [x] Add one more fixture (array) with negative value and zero, and another with real number (floating point).
- [x] Avoid the for loop in the test code.
- [x] Don't use `std::transform` in `take_along_axis()`.  The previous loop code looks clearer.
- [ ] You still have unnecessary bound checking (`athis->at()`).    Don't waste cycles.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: In my opinion, the bound checking in indexing `athis` (`athis->at()`) is necessary, because this function can take not only the result of `argsort` but also arrays with any possible values.

For example
```c++
SimpleArrayInt32 arr({1,3,5});
SimpleArrayUint64 indices1({2,1,0,1,2});
SimpleArrayUint64 indices2({1,2,3,4,5});

SimpleArrayInt32 res1 = arr.take_along_axis(indices1) // res1 = {5,3,1,3,5}
SimpleArrayInt32 res2 = arr.take_along_axis(indices2) // buffer overflow
```

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: Since the function implementation does use the implementation of `SimpleArray`, it must be implemented after the implementation of `SimpleArray`.

I saw some other people add a new `SimpleArray.tpp` file to write the implementation of long template functions and include it at the end of `SimpleArray.hpp`. Do you think this is a good idea for us to better organize the codes?

KHLee529:
@yungyuc I fixed some of the problem you mentioned except for the follow two. I left comments of the reason why I didn't fix them below your origin comment on the codes.

- Make the function implementation/definition to immediately follow the class declaration of the mixin.
- You still have unnecessary bound checking (athis->at()). Don't waste cycles.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +753,72 @@ class SimpleArray
...
+    static_assert(std::is_integral_v<I>, "I must be integral type");
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray: take_along_axis() supports currently only in 1D array "
+                                                " but the array is "
+                                             << athis->ndim() << " dimension");
+    }
+
+    SimpleArray<T> ret(indices.shape());
```
yungyuc: `SimpleArray<T>` should be `A` here for consistency to the returning type `A`.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: I see, you are talking about [line 781](https://github.com/solvcon/modmesh/pull/456/files#diff-bf43ff74a749046af0d1aca3542c01ba73e98d3065d0cabb4fc3f6ebc1f85f7dR781).

What about keeping only that function after `SimpleArray`?

No, we are not ready to make the template structure so complex.  We may consider that in the future, but not now.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -190,6 +191,25 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
 
         return *this;
     }
+
+    wrapper_type & wrap_sort()
```
yungyuc, line 195: This is good.

tests/test_buffer.py
```diff
@@ -824,6 +824,41 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
...
+
+            sorted_arr = sarr.take_along_axis(args)
+            for i in range(1, len(sorted_arr)):
+                self.assertLessEqual(sorted_arr[i - 1], sorted_arr[i])
+
+            sarr.sort()
+            for i in range(1, len(sarr)):
+                self.assertLessEqual(sarr[i - 1], sarr[i])
+
+        _check(test_data[0])
```
yungyuc, line 859: Nice.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +753,72 @@ class SimpleArray
...
+
+    auto val_iter = ret.begin();
+    for (auto idx : indices)
+    {
+        *val_iter = athis->at(static_cast<size_t>(idx));
```
yungyuc: Don't use the bound-checking `at()`.  Use `operator[]()` that does not check bounds.

And the iterator will soon bite.  Keep in mind that you need to implement this function for multi-dimensional arrays.  Just use indices.

When improving, leave a code comment indicating this part will need to be enhanced for multi-dimensional arrays.

tests/test_buffer.py
yungyuc: Please leave a code comment that there will need another test for ghost indices in the future.

yungyuc (changes requested):
- [x] Make the function implementation/definition to immediately follow the class declaration of the mixin as much as possible.
- [x] Avoid bound-checking `at()` in `take_along_axis()`.
- [x] Just use indices in `take_along_axis()`.  You could use iterator simply because you only implement the function for 1D array.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: In my opinion, the bound checking in indexing athis (athis->at()) is necessary, because this function can take not only the result of argsort but also arrays with any possible values.

For example

```c++
SimpleArrayInt32 arr({1,3,5});
SimpleArrayUint64 indices1({2,1,0,1,2});
SimpleArrayUint64 indices2({1,2,3,4,5});

SimpleArrayInt32 res1 = arr.take_along_axis(indices1); // res1 = {5,3,1,3,5}
SimpleArrayInt32 res2 = arr.take_along_axis(indices2); // buffer overflow
```

Besides, the iterator actually are taken advantage in the current stage because it go through every element in the SimpleArray no matter which dimension it is. It can, therefore,  already support multi-dimentional `indices`.

For example (this is just pseudo code, the initializer list might not be usable in real code)

```c++
SimpleArrayInt32 arr({1,3,5});
SimpleArrayUint64 indices3( {{0,1,2}, {2,1,0}} );
SimpleArrayInt32 res3 = arr.take_along_axis(indices3); // res3 = { {1, 3, 5}, {5, 3, 1} }
```

yungyuc: In this case, add multi-dimensional support for the function and corresponding test cases.  The implementation should be consistent to test.

If you need bound-checking, do it manually, so that you can customize the error message, and the code looks much clearer.  The error message of `at()` will be cryptic and does not help users fix an out-of-range index.

You should add a test case for the out-of-range scenario too.

KHLee529:
@yungyuc all the change requests are fixed. Please review the modification.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +767,63 @@ class SimpleArray
...
+    }
+
+    A ret(indices.shape());
+
+    auto val_iter = ret.begin();
+    for (auto idx : indices)
+    {
+        if (idx < 0 || idx > athis->shape()[0])
+        {
+            throw std::out_of_range(Formatter() << "SimpleArray: index " << idx << " in indices "
```
yungyuc: The error message should writes like:

```
SimpleArray::take_along_axis() indices[i1] is i2, which is out of range of the array size v_size
```

You need 3 values:
* `i1` is the serial number of the input indices whose index is out of range
* `i2` is the index out of range
* `vsize` is the range (size)

yungyuc (changes requested):
- [x] `SimpleArray::take_along_axis()` need clearer out of range error message.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: OK, now I know why iterator will bite.

KHLee529:
@yungyuc change request is fixed, please review.

yungyuc (changes requested):
You seem to forget adding a unit test for the out-of-range error message.  Please add one.

yungyuc:
And please squash your commits.  We need only one for this straight-forward enhancement.

KHLee529:
@yungyuc I added the unit test but not do message assertion. I've fixed it, please review.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +789,70 @@ class SimpleArray
...
+
+template <typename A, typename T>
+template <typename I>
+A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & indices)
+{
+    static_assert(std::is_integral_v<I>, "I must be integral type");
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray: take_along_axis() supports currently only in 1D array "
```
yungyuc: It should be `SimpleArray::take_along_axis()`.  There are two `::`, just like how you write C++ scoping.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +789,70 @@ class SimpleArray
...
+    do {
+        I idx = indices.at(sidx);
+        if (idx < 0 || idx > athis->shape()[0])
+        {
+            Formatter err_msg = Formatter();
+            err_msg << "SimpleArray: take_along_axis() indices[" << sidx[0];
+            for (size_t dim = 1; dim < sidx.size(); ++dim)
+            {
+                err_msg << ", " << sidx[dim];
+            }
```
yungyuc, line 842: Use `::take_along_axis()`, not `:take_along_axis()`

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -729,6 +789,70 @@ class SimpleArray
...
+    }
+
+    A ret(indices.shape());
+
+    auto sidx = indices.first_sidx();
+    do {
+        I idx = indices.at(sidx);
+        if (idx < 0 || idx > athis->shape()[0])
+        {
+            Formatter err_msg = Formatter();
```
yungyuc: Just write

```cpp
Formatter err_msg;
```

tests/test_buffer.py
```diff
@@ -824,6 +824,80 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
...
+                self.assertEqual(ret_arr[i, j], data[idx_arr[i, j]])
+
+        # test out-of-range index
+        idx = [[0, 1], [2, 3], [4, 20]]
+        narr = np.array(idx, dtype='uint64')
+        idx_arr = modmesh.SimpleArrayUint64(array=narr)
+
+        with self.assertRaisesRegex(
+            IndexError,
+            r"SimpleArray: take_along_axis\(\) indices\[2, 1\] is 20, " +
```
yungyuc: Great!

yungyuc (changes requested):
Please fix error message formatting.

KHLee529:
@yungyuc changes requested are done, please review.

yungyuc (approved):
Look good.

tests/test_buffer.py
```diff
@@ -824,6 +824,80 @@ def test_SimpleArray_SimpleArrayPlex_type_switch(self):
...
+            if use_float:
+                narr = np.array(arr, dtype='float64')
+                sarr = modmesh.SimpleArrayFloat64(array=narr)
+            else:
+                narr = np.array(arr, dtype='int32')
+                sarr = modmesh.SimpleArrayInt32(array=narr)
+
+            args = sarr.argsort()
+            for i in range(1, len(args)):
+                self.assertLessEqual(sarr[args[i]], sarr[args[i]])
```
pchsu-hsupc, line 849: Should be `sarr[args[i-1]]`, right?

KHLee529: Oh it's right. I wrote it wrongly.

PR #465, opened 2025-02-01
Allow `modmesh.World` to keep vertices

yungyuc:
For `modmesh.World`:

* Implement `vertex()`, `add_vertex()`, `nvertex` on `World`.
* Use `SimpleCollector` to implement the vertex container in `World`.
* Add `push_back` to `SimpleCollector`.

To improve the relevant testing:

* Create `modmesh.testing.TestBase` to house common code for testing.

PR #466, opened 2025-02-01
Fix the issue that freezes `SimpleCollector::push_back` from 2 -> 3

yungyuc:
A logical bug in `SimpleCollector::push_back` prevents the size from increasing from 2 to 3.

The bug was introduced with the initial implementation in PR #465 .

### DIFF
--- a/cpp/modmesh/buffer/SimpleCollector.hpp
+++ b/cpp/modmesh/buffer/SimpleCollector.hpp
@@ -156,17 +156,16 @@ private:
         if (capacity() == 0)
         {
             reserve(1);
-            m_expander->push_size(ITEMSIZE);
         }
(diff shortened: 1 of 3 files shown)


cpp/modmesh/buffer/SimpleCollector.hpp
```diff
@@ -156,17 +156,16 @@ class SimpleCollector
...
         else if (size() == capacity())
         {
             reserve(size() * 2);
-            m_expander->push_size(ITEMSIZE);
         }
         else
         {
             // do nothing
         }
+        m_expander->push_size(ITEMSIZE);
```
yungyuc, line 168: This is the culprit.

PR #469, opened 2025-02-15
Implement `TimeSeriesDataFrame.sort`

KHLee529:
## Abstract

As mentioned in #435, `TimeSeriesDataFrame` should have data sorted along the index, also known as the time series. 
This PR implement the member function `reorder` for `TimeSeriesDataFrame`

## Functionality

`TimeSeriesDataFrame.reorder(columns)` takes an array of name of the columns in the data frame as argument and returns a new data frame with index sorted and the data also reordered along the index.

### DIFF
--- a/tests/test_timeseries_dataframe.py
+++ b/tests/test_timeseries_dataframe.py
@@ -128,6 +128,69 @@ class TimeSeriesDataFrame(object):
     def index(self):
         return self._index_data.ndarray
 
+    def sort(self, columns=None, index_column=None, inplace=True):
+        """
+        Sort the dataframe along the given index column
+
+        :param columns: column names required in reordered TimeSeriesDataFrame
+        :type columns: Option[List[str]]
+        :param index_column: column name treated as the index, if None is
+                                given, sort along the index
+        :type index_column: Option[str]
+        :param inplace: flag indicates whether to sort inplace or out-of-place
+        :type inplace: bool
+        :return: sorted TimeSeriesDataFrame (return self if inplace is set to
+                 True)
+        """
+
+        if index_column is None and self._index_data is None:
+            raise ValueError("TimeSeriesDataFrame: data frame has no index, "
+                             "please provide index column")
+
+        index_data = self._index_data if (
+            index_column is None
+            ) else self._data[self._columns.index(index_column)]
+        indices = index_data.argsort()
+
+        if inplace:
+            for i, col in enumerate(self._data):
+                self._data[i] = col.take_along_axis(indices)
+
+            if self._index_data is not None:
+                self._index_data = self._index_data.take_along_axis(indices)
+
+            return self
+        else:
+            if columns is None:
+                columns = []
+
+            ret = TimeSeriesDataFrame()
+            ret._index_name = self._index_name
+            ret._index_data = self._index_data.take_along_axis(indices)
+
+            for name in columns:
+                if name not in self._columns:
+                    raise ValueError("Column '{}' does not exist".format(name))
+                idx = self._columns.index(name)
+                new_col = self._data[idx].take_along_axis(indices)
+                ret._columns.append(name)
+                ret._data.append(new_col)
+
+            return ret
+
+    def sort_by_index(self, columns=None, inplace=True):
+        """
+        Sort the dataframe along the index column
+
+        :param columns: column names required in reordered TimeSeriesDataFrame
+        :type columns: List[str]
+        :param inplace: flag indicates whether to sort inplace or out-of-place
+        :type inplace: bool
+        :return: sorted TimeSeriesDataFrame (return self if inplace is set to
+                True)
+        """
+        return self.sort(columns=columns, index_column=None, inplace=inplace)
+
 
 class TimeSeriesDataFrameTC(unittest.TestCase):
 
@@ -157,6 +220,18 @@ class TimeSeriesDataFrameTC(unittest.TestCase):
 -0.1895751953125,-0.00128173828125,1602596010369299968,-0.04925537109375
 -0.18841552734375,6.103515625e-05,1602596010389309952,-0.0489501953125
 -0.1884765625,-0.00042724609375,1602596010409309952,-0.04840087890625
+"""
+    unsorted_dlc_data = """EPOCH ,DELTA_VEL[1] ,DELTA_VEL[2] ,DELTA_VEL[3]
+1.60259601024931e+18,-0.1903076171875,-0.0009765625,-0.0489501953125
+1.60259601034931e+18,-0.18902587890625,-0.000732421875,-0.0489501953125
+1.60259601040931e+18,-0.1884765625,-0.00042724609375,-0.04840087890625
+1.60259601032931e+18,-0.18951416015625,-0.000732421875,-0.0489501953125
+1.6025960102293e+18,-0.18792724609375,-0.00048828125,-0.0478515625
+1.6025960103693e+18,-0.1895751953125,-0.00128173828125,-0.04925537109375
+1.60259601030931e+18,-0.188720703125,-0.00103759765625,-0.0504150390625
+1.60259601026931e+18,-0.18743896484375,0.0006103515625,-0.0498046875
+1.60259601028932e+18,-0.18927001953125,-0.0009765625,-0.04840087890625
+1.60259601038931e+18,-0.18841552734375,6.103515625e-05,-0.0489501953125
 """
 
     def test_read_from_text_file_basic(self):
@@ -220,4 +295,31 @@ class TimeSeriesDataFrameTC(unittest.TestCase):
 
         nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
 
+        self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+    def test_dataframe_sort(self):
+        tsdf = TimeSeriesDataFrame()
+        tsdf.read_from_text_file(StringIO(self.unsorted_dlc_data))
+
+        # Test out-of-place sort
+        reordered_tsdf = tsdf.sort(tsdf.columns, index_column=None,
+                                   inplace=False)
+        col_data = reordered_tsdf['DELTA_VEL[1]']
+        nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
+        self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+        # Test inplace sort_by_index
+        tsdf.sort_by_index()
+        col_data = tsdf['DELTA_VEL[1]']
+        nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
+        self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+        # Test out-of-place sort with index_column
+        tsdf.read_from_text_file(StringIO(self.unsorted_dlc_data),
+                                 timestamp_in_file=False)
+
+        reordered_tsdf = tsdf.sort(['EPOCH', 'DELTA_VEL[1]'],
+                                   index_column='EPOCH', inplace=False)
+        col_data = reordered_tsdf['DELTA_VEL[1]']
+        nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
         self.assertEqual(list(col_data), list(nd_arr[:, 1]))


KHLee529:
@xingularity @yungyuc This PR implements the `TimeSeriesDataFrame.reorder` mentioned in issue #435. Please review it and check whether it fits the requirement.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,35 @@ def shape(self):
...
+    def reorder(self, columns):
+        """
+        Reorder the data frame based on the index
+
+        :param columns: column names required in reordered TimeSeriesDataFrame
+        :type columns: Iterable[str]
+        
+        :return: reordered TimeSeriesDataFrame
+        """
+        if self._index_data is None:
```
KHLee529: Since the data frame should be reordered according to the index. Exception should be raised when this function is called by a data frame without index data.

tests/test_timeseries_dataframe.py
```diff
@@ -157,6 +186,18 @@ class TimeSeriesDataFrameTC(unittest.TestCase):
 -0.1895751953125,-0.00128173828125,1602596010369299968,-0.04925537109375
 -0.18841552734375,6.103515625e-05,1602596010389309952,-0.0489501953125
 -0.1884765625,-0.00042724609375,1602596010409309952,-0.04840087890625
+"""
+    unsorted_dlc_data = """EPOCH ,DELTA_VEL[1] ,DELTA_VEL[2] ,DELTA_VEL[3]
```
KHLee529, line 224: Shuffled DLC data for testing reorder functionality.

tests/test_timeseries_dataframe.py
```diff
@@ -221,3 +262,14 @@ def test_dataframe_get_column(self):
...
+    def test_dataframe_reorder(self):
+        tsdf = TimeSeriesDataFrame()
+        tsdf.read_from_text_file(StringIO(self.unsorted_dlc_data))
+        reordered_tsdf = tsdf.reorder(tsdf.columns)
+
+        col_data = reordered_tsdf['DELTA_VEL[1]']
+
+        nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
+
+        self.assertEqual(list(col_data), list(nd_arr[:, 1]))
```
KHLee529, line 315: Compare the reordered data with data originally in order.

tests/test_timeseries_dataframe.py
yungyuc: @xingularity, please comment.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,34 @@ def shape(self):
...
+        :param columns: column names required in reordered TimeSeriesDataFrame
+        :type columns: Iterable[str]
+        :return: reordered TimeSeriesDataFrame
+        """
+        if self._index_data is None:
+            raise Exception("TimeSeriesDataFrame: data frame has no index")
+
+        sorted_indices = self._index_data.argsort()
+
+        ret = TimeSeriesDataFrame()
```
yungyuc: I think `reorder` should perform in-place by default.  @KHLee529 could you please check for how it behaves in pandas and make this consistent with it?

yungyuc (changes requested):
- ~~[ ] `TimeSeriesDataFrame.reorder()` should use consistent behaviors with pandas counterpart.~~

tests/test_timeseries_dataframe.py
xingularity: Hi @yungyuc 

This is time series data frame, the data should be sorted against index. So, the reorder function here should be more like "sort_index" instead of "reorder_levels" in the Pandas, and sort_index is default out-of-place.

The description I put in #435 is indeed to reorder out-of-place and return ordered data of selected columns, to avoid sorting all data. I think we can add an option to let user decide if the want to sort in-place or out-of-place. When user chooses in-place, them all columns should be sorted against index.

tests/test_timeseries_dataframe.py
xingularity: I recommend to do this. Because a data frame should ALWAYS have index data. So if there is no index data here, we should throw error.

tests/test_timeseries_dataframe.py
yungyuc: Is `reorder()` still a good name for the function?  If not, should it use a different name here?

I think a keyword argument `inplace` (but please advise for a name compatible to pandas or similar convention) should be used regardless of the name of the function.  The default should follow pandas convention.

xingularity: 1. sort_time or sort_timestamp is better naming. Since in the "read_from_text_file" function, the name of the parameter is  "timestamp_column". And the sorting is always against index (storing timestamp) column.
2. By using above naming, and API design, this dataframe would be specifically for time series data. If we want to build a more general-purpose dataframe, and treat time series as a special case on top of it, we can use another issue to address it and perform refactoring.
3. We also need to consider one point, many of Pandas features are out-of-place (inplace=False) by default now, including sort_values, sort_index and reorder_levels. So if we follow Pandas default setting, it is not memory-efficient.

yungyuc: Insightful analysis.  So maybe we should diverge from the pandas default value of `inplace` and make `inplace=True` as the default, because speed is the king.

@xingularity could you make a single recommendation for the function name?  It is hard for @KHLee529 to choose between naming options.

@j8xixo12 could you please comment?

xingularity: sort_time is shorter. I recommend this.

xingularity: Hi @KHLee529 

Let me summarize above discussion for now. We originally allow user to input column names and we return a copied dataframe with those columns and sorted data. Now we need two optional parameters columns, and inplace. 

- columns is an array of strings which specify the column names the user wanted with sorted data.
- inplace is a boolean value. If it is true, then ignore "columns" and sorted all columns against index columns, and return None. Otherwise, sorting selected columns (or all columns if columns=None) and returns a sorted dataframe.

Let's wait @j8xixo12 's comment before you make changes.

j8xixo12: As @xingularity mentioned, epoch sorting is a specific case of time series dataframes, but sorting other columns is indeed a convenient feature. How about `reorder` or `sort` as a general function, and `sort_time` can call `reorder` to sort epochs?
I agree with @yungyuc’s point that adding an argument `inplace` with a default value of `True`, because speed is the king!

yungyuc: Sounds like a plan.  @KHLee529 , please follow the conclusion.  If there is anything unclear, please bring it to discussions.

KHLee529: Let me summarize the plan. @yungyuc @xingularity @j8xixo12 Please check if I understand anything wrong.
1. The name of function `reorder` should be renamed to `sort_time`
2. A argument `inplace` should be added to function `sort_time` and set to `True` by default.
    ```python
    def sort_time(columns: List[str] = list(), inplace: bool = True) 
    ```
    * if `inplace` is `True`, it will ignore the columns chosen and sort every columns existing.
    * if `inplace` is `False`, it will sort only the columns chosen by parameter `columns`
3. Another function `sort` will be added with prototype 
    ```python
    def sort(columns: List[str], index_column: Option[str] = None, inplace: bool = True)
    ```
    * if the index_column is set to `None`, by default it will be sorted with respect to the index column (identical to the behavior of `sort_time`, `sort_time` will call this function with this setting)
    * if the index_column is any of the names of columns, it will be sorted with respected the specified column.

===== My opinion =====
However, the modification above might actually not improve speeds because currently the `take_along_axis` function of `SimpleArray` which is used for apply the sorted arguments to other `SimpleArray` is by default out-of-place (which follows the Numpy convention). Inplace setting for data frame, at this moment, will therefore implemented by copying the sorted buffer back to the original buffer, or by replacing the original buffer with the new one returned. This implementation will indeed not improve the speed and even increase time spent.
If the key consideration is speed, I think keep the original out-of-place version might be fit better. Otherwise, the inplace `take_along_axis` function might need implemented first.

yungyuc: It looks good.  Except that since the function will sort against an index column, I wonder if the name `sort_by_index()` makes more sense than `sort_time()`?

Regardless, I suggest to name it as `sort_by_index()` for now.  It may be changed if a better opinion comes.

It is OK to postpone the speedup.  For high-performance design, what matters is that the design will _allow_ speedup.  The actual speedup should be done after profiling.

KHLee529: @yungyuc Thanks for clarification. I'll name the function `sort_by_index()` at this moment and keep the function prototype as I summarized.

```python
def sort(columns: List[str], index_column: Option[str] = None, inplace: bool = True)
```

```python
def sort_by_index(columns: List[str] = list(), inplace: bool = True) 
```

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,66 @@ def shape(self):
...
+            raise Exception("TimeSeriesDataFrame: data frame has no index")
+
+        index_data = self._index_data if (
+            index_column is None
+            ) else self._data[self._columns.index(index_column)]
+        sorted_indices = index_data.argsort()
+
+        if inplace:
+            for i, col in enumerate(self._data):
+                self._data[i] = col.take_along_axis(sorted_indices)
```
KHLee529: Update data if `inplace` set to `True`

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,66 @@ def shape(self):
...
+
+            if self._index_data is not None:
+                self._index_data = self._index_data.take_along_axis(
+                    sorted_indices)
+
+            return self
+        else:
+            ret = TimeSeriesDataFrame()
```
KHLee529, line 167: Setup a new data frame when `inplace` is set to `False`

KHLee529:
@yungyuc @xingularity @j8xixo12 I've modified the `TimeSeriesDataFrame` according to our discussion above. Please review it.

yungyuc:
@xingularity @j8xixo12 please let me know what do you think.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,66 @@ def shape(self):
     def index(self):
         return self._index_data.ndarray
 
+    def sort(self, columns=list(), index_column=None, inplace=True):
```
yungyuc: Do not use a mutable container as Python argument default value.  It does not behave as what you want, see:

* https://www.jetbrains.com/help/inspectopedia/PyDefaultArgument.html
* https://stackoverflow.com/questions/41686829/why-does-pycharm-warn-about-mutable-default-arguments-how-can-i-work-around-the 

TL;DR, make `columns=None`.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,66 @@ def shape(self):
...
+
+            return self
+        else:
+            ret = TimeSeriesDataFrame()
+            ret._index_name = self._index_name
+            ret._index_data = self._index_data.take_along_axis(sorted_indices)
+
+            for name in columns:
+                if name not in self._columns:
+                    raise Exception("Column '{}' does not exist".format(name))
```
yungyuc: Use [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError)

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,66 @@ def shape(self):
...
+                new_col = self._data[self._columns.index(name)
```
yungyuc: You can break the line in a nicer way:

```python
_ = self._columns.index(name)
new_col = self._data[_].take_along_axis(sorted_indices)
```

`_` is a idiomatic name for temporary surrogate for line-breaking.

Alternately, you can give the temporary variable a meaningful name:

```python
i = self._columns.index(name)
new_col = self._data[i].take_along_axis(sorted_indices)
```

yungyuc (changes requested):
- [x] Fix the argument default value being mutable
- [x] Use `ValueError` for input argument
- [x] Make line breaking nicer

tests/test_timeseries_dataframe.py
KHLee529: Thanks for this important feedback.

KHLee529:
> * [ ]  Fix the argument default value being mutable
> * [ ]  Use `ValueError` for input argument
> * [ ]  Make line breaking nicer

@yungyuc The problems you mentioned here are solved, please review it.

cpp/modmesh/spacetime/pymod/wrap_spacetime.hpp
```diff
@@ -361,7 +361,7 @@ class WrapSolverBase
                 py::arg("odd_plane") = false)
             .def("get_so0p", &wrapped_type::get_so0p, py::arg("iv"), py::arg("odd_plane") = false);
 
-// clang-format off
+        // clang-format off
```
yungyuc: Don't make unrelated changes.  Please revert the change of spaces.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,68 @@ def shape(self):
...
+                                given, sort along the index
+        :type index_column: Option[str]
+        :param inplace: flag indicates whether to sort inplace or out-of-place
+        :type inplace: bool
+        :return: sorted TimeSeriesDataFrame (return self if inplace is set to
+                 True)
+        """
+
+        if index_column is None and self._index_data is None:
+            raise Exception("TimeSeriesDataFrame: data frame has no index")
```
yungyuc: `ValueError` too.  It is very rare to use `Exception` in Python.

yungyuc (changes requested):
- [x] Revert unrelated changes
- [x] `ValueError` for exception about arguments.

@xingularity @j8xixo12 please help review too.  @KHLee529 please find them on discord or google space.

KHLee529:
@yungyuc requested changes have been fixed. Please review them.

tests/test_timeseries_dataframe.py
```diff
@@ -128,6 +128,69 @@ def shape(self):
...
+                                given, sort along the index
+        :type index_column: Option[str]
+        :param inplace: flag indicates whether to sort inplace or out-of-place
+        :type inplace: bool
+        :return: sorted TimeSeriesDataFrame (return self if inplace is set to
+                 True)
+        """
+
+        if index_column is None and self._index_data is None:
+            raise ValueError("TimeSeriesDataFrame: data frame has no index, " +
```
yungyuc: The trailing `+` is not necessary.  It will be good to be removed.

yungyuc (approved):

tests/test_timeseries_dataframe.py
```diff
@@ -221,3 +296,20 @@ def test_dataframe_get_column(self):
         nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
 
         self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+    def test_dataframe_sort(self):
+        tsdf = TimeSeriesDataFrame()
+        tsdf.read_from_text_file(StringIO(self.unsorted_dlc_data))
+
+        # Test out-of-place sort
+        reordered_tsdf = tsdf.sort(tsdf.columns, index_column=None,
```
j8xixo12, line 305: It would be good to have a test case for `index_column`.

KHLee529:
@j8xixo12 The test case for sort with `index_column` is added, please review it.

tests/test_timeseries_dataframe.py
yungyuc: Agree

yungyuc:
> @j8xixo12 The test case for sort with `index_column` is added, please review it.

@KHLee529 please fix test/linter failure

tests/test_timeseries_dataframe.py
```diff
@@ -221,3 +296,30 @@ def test_dataframe_get_column(self):
...
+        col_data = tsdf['DELTA_VEL[1]']
+        nd_arr = np.genfromtxt(StringIO(self.dlc_data), delimiter=',')[1:]
+        self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+        # Test out-of-place sort with index_column
+        tsdf.read_from_text_file(StringIO(self.unsorted_dlc_data),
+                                 timestamp_in_file=False)
+
+        reordered_tsdf = tsdf.sort(['EPOCH', 'DELTA_VEL[1]'],
+                                   index_column='EPOCH', inplace=False)
```
j8xixo12, line 322: LGTM.

j8xixo12 (approved):

PR #474, opened 2025-02-26
Refactor: replace the implementation about config widget in One/Euler windows from PUI to PySide6 directly

Sean20405:
In this PR, it changes how the config widget in One/Euler windows in GUI is implemented. It was developed using PUI, a third party package, and now it transfers into using PySide6 directly.

It also modified the data structure in `GUIConfig`, from `PUI.state` to a 2D list, so that it will not depend on `PUI`.

This PR is ready to be reviewed!

PR #476, opened 2025-03-09
refactor: Completely remove PUI in One/Euler

### LINKED ISSUE #462: Use native PySide6 for `modmesh.pilot`
The Euler 1D solver GUI currently uses [`PUI`](https://github.com/buganini/PUI) for GUI controls, but the layout should be improved.  See the red box in the screenshot:

Image

Consider to copy-n-paste code from `PUI` to `modmesh.pilot` (in which `modmesh.pilot._euler1d` contains the most `PUI`-based code) for the GUI visual improvement.

Sean20405:
This PR finishes the remaining part in #474. It removes PUI in plot area and change where the plot canva places.

This PR is ready to be reviewed!

### DIFF
--- a/modmesh/pilot/_euler1d.py
+++ b/modmesh/pilot/_euler1d.py
@@ -42,11 +42,6 @@ from PySide6.QtWidgets import (QDockWidget, QLabel, QVBoxLayout, QHBoxLayout,
                                QComboBox, QPushButton, QSpacerItem,
                                QSizePolicy, QDialog, QWidget, QTableView)
 
-from PUI.state import State
-from PUI.PySide6.base import PuiInQt, QtInPui
-from PUI.PySide6.window import Window
-from PUI.PySide6.toolbar import ToolBar
-
 from .. import core as mcore
 from ..onedim import euler1d
 from ._gui_common import PilotFeature
@@ -356,18 +351,18 @@ class Euler1DApp(PilotFeature):
 
     def run(self):
         self.setup_app()
-        plotting_area = PlotArea(Window(), self)
 
+        # A new dock widget for showing config
         config_window = ConfigWindow(self)
         config_widget = QDockWidget("config")
         config_widget.setWidget(config_window)
-
         self._mgr.mainWindow.addDockWidget(Qt.LeftDockWidgetArea,
                                            config_widget)
-        _subwin = self._mgr.addSubWindow(plotting_area.ui.ui)
-        _subwin.showMaximized()
 
-        plotting_area.redraw()
+        # A new sub-window (`QMdiSubWindow`) for the plot area
+        _subwin = self._mgr.addSubWindow(QWidget())
+        _subwin.setWidget(PlotArea_wo_Pui(self))
+        _subwin.showMaximized()
         _subwin.show()
 
     def setup_app(self):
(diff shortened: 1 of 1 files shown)


modmesh/pilot/_euler1d.py
```diff
@@ -356,18 +351,18 @@ def populate_menu(self):
...
         config_widget.setWidget(config_window)
-
         self._mgr.mainWindow.addDockWidget(Qt.LeftDockWidgetArea,
                                            config_widget)
-        _subwin = self._mgr.addSubWindow(plotting_area.ui.ui)
-        _subwin.showMaximized()
 
-        plotting_area.redraw()
+        # A new sub-window (`QMdiSubWindow`) for the plot area
+        _subwin = self._mgr.addSubWindow(QWidget())
```
Sean20405, line 363: The current code can run properly. But I'm wonder why `_subwin = self._mgr.addSubWindow(PlotArea_wo_Pui(self))` may not work.

modmesh/pilot/_euler1d.py
```diff
@@ -730,38 +724,30 @@ def update_lines(self):
...
-        - :meth:`content()`: Method for defining the content of the plot area,
-          including a toolbar and the actual plot.
+        - :meth:`init_UI()`: Define the GUI layout of the window.
     """
-
-    def __init__(self, parent, app):
+    def __init__(self, app, parent=None):
         super().__init__(parent)
         self.app = app
+        self.init_UI()
```
Sean20405: In the discussion of #474, it mentioned that the naming convention is mixed. I'm wonder whether I can resolve it in this PR. If so, which one of the convention would be better.

I think `snake_case` may be a better one. (?)

modmesh/pilot/_euler1d.py
yungyuc: Maybe `addSubWindow` does not own the return of `PlotArea_wo_Pui(self)`.  Just a wild guess and need study for a proof.

modmesh/pilot/_euler1d.py
```diff
@@ -730,38 +724,30 @@ def update_lines(self):
                          f' ndata=self.st.svr.{name}[self.st.svr.xindices]))')
 
 
-class PlotArea(PuiInQt):
+class PlotArea_wo_Pui(QWidget):
```
yungyuc: Naming convention should not mix `CamelCase` and `snake_case`.  Class uses `CamelCase`.

Just retain the original name `PlotArea`.

modmesh/pilot/_euler1d.py
yungyuc: If you use `snake_case`, why don't you name this function `init_ui`?  Never mix `Camel_Snake_cases` which only confuses people without any help.

yungyuc (changes requested):
- [x] Keep the original class name `PlotArea`.  It does not make sense to name a class based on the use of PUI like `PlotArea_wo_Pui`.
- [x] Don't use mix names like `init_UI`.  Very confusing.

@j8xixo12 please help review.

modmesh/pilot/_euler1d.py
yungyuc: @buganini any thought?

modmesh/pilot/_euler1d.py
j8xixo12: Agree, since `PUI` will be phased out, having `wo_Pui` does not make much sense anymore.

j8xixo12:
@Sean20405 Could you please paste the screenshot of the `euler1d` application actually running ?

Sean20405:
> @Sean20405 Could you please paste the screenshot of the `euler1d` application actually running ?

Sure! Here it is.
image

modmesh/pilot/_euler1d.py
Sean20405: Ohhh that is a mistake. This class name is used to not conflict the original one when I first start. I forgot to change it.

Already fix it!

Sean20405:
Hi @yungyuc,
I've changed the class and function name. Please review it again. Thanks a lot!

yungyuc (approved):
LGTM

yungyuc:
@j8xixo12 After merging, it stops from working on mac.  Need your help fix on the mac platform in a followup PR.

PR #486, opened 2025-03-19
Improve efficiency of `take_along_axis`

KHLee529:
It is found that the efficiency is incredibly low comparing the the counterpart of Numpy, after profiling the performance of current implementation of `take_along_axis`.

(unit: ms/call) | Ascending | Descending | Random
-- | -- | -- | --
SimpleArray.take_along_axis() | 63 | 63 | 67
np.take_along_axis(ndarray) | 1.7 | 1.7 | 2.5

According to the implementation of `take_along_axis` of numpy, the function is refactored to lower level manipulation, which makes the function about 10x faster.

The profile result after this modification is the following.

(unit: ms/call) | Ascending | Descending | Random
-- | -- | -- | --
SimpleArray.take_along_axis() | 3.6 | 3.5 | 5.9
np.take_along_axis(ndarray) | 1.7 | 1.7 | 2.5

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -837,26 +837,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
                                              << " but the array is " << athis->ndim() << " dimension");
     }
 
-    A ret(indices.shape());
-
-    auto sidx = indices.first_sidx();
-    do {
-        I idx = indices.at(sidx);
-        if (idx < 0 || idx > athis->shape()[0])
+    size_t max_idx = athis->shape()[0];
+    I const * src = indices.begin();
+    I const * const end = indices.end();
+    while (src < end)
+    {
+        if (*src < 0 || *src > max_idx)
         {
+            size_t offset = src - indices.begin();
+            shape_type const & stride = indices.stride();
             Formatter err_msg;
-            err_msg << "SimpleArray::take_along_axis(): indices[" << sidx[0];
-            for (size_t dim = 1; dim < sidx.size(); ++dim)
+            err_msg << "SimpleArray::take_along_axis(): indices[" << offset / stride[0];
+            offset %= stride[0];
+            for (size_t dim = 1; dim < stride.size(); ++dim)
             {
-                err_msg << ", " << sidx[dim];
+                err_msg << ", " << offset / stride[dim];
+                offset %= stride[dim];
             }
-            err_msg << "] is " << idx << ", which is out of range of the array size "
-                    << athis->shape()[0];
+            err_msg << "] is " << *src << ", which is out of range of the array size "
+                    << max_idx;
 
             throw std::out_of_range(err_msg);
         }
-        ret.at(sidx) = (*athis)[static_cast<size_t>(idx)];
-    } while (indices.next_sidx(sidx));
+        src++;
+    }
+
+    src = indices.begin();
+    A ret(indices.shape());
+    T * data = athis->begin();
+    T * dst = ret.begin();
+    while (src < end)
+    {
+        T * valp = data + static_cast<size_t>(*src);
+        *dst = *valp;
+        ++dst;
+        ++src;
+    }
     return ret;
 }
 


KHLee529:
@yungyuc This PR is to improve the performance of `take_along_axis`, please review it.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -837,26 +837,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
-    A ret(indices.shape());
-
-    auto sidx = indices.first_sidx();
-    do {
-        I idx = indices.at(sidx);
-        if (idx < 0 || idx > athis->shape()[0])
+    size_t max_idx = athis->shape()[0];
+    I const * src = indices.begin();
+    I const * const end = indices.end();
+    while (src < end)
```
KHLee529, line 843: Check indices validity first.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -837,26 +837,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
-        ret.at(sidx) = (*athis)[static_cast<size_t>(idx)];
-    } while (indices.next_sidx(sidx));
+        src++;
+    }
+
+    src = indices.begin();
+    A ret(indices.shape());
+    T * data = athis->begin();
+    T * dst = ret.begin();
+    while (src < end)
```
KHLee529, line 869: Copy values to the output array

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -837,26 +837,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    {
+        T * valp = data + static_cast<size_t>(*src);
+        *dst = *valp;
+        dst++;
```
yungyuc: Please use the pre-increment operator `++dst`, which avoids a copy when the type is not fundamental: https://en.cppreference.com/w/cpp/language/operator_incdec 

Here it makes no difference to use either pre or post increment.  But if someone copies the code elsewhere for non-fundamental type, it will become a defect.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: It is wise to factor unchanged parts outside the loop.

yungyuc (changes requested):
LGTM.  Please improve the maintainability.

- [ ] Always use pre-increment unless the copy behavior in post-increment is really needed.

yungyuc:
While addressing the maintainability, what do you think about the next steps to make it faster than numpy?

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: Because I found this actually is a factor that slows down the function.

KHLee529:
> While addressing the maintainability, what do you think about the next steps to make it faster than numpy?

I, to be honest, have no concrete idea about how to further improve it now. I'm thinking about parallel programming or speed up using hardware feature but I still have no idea about either whether it will work or how to implement it.

KHLee529:
> LGTM. Please improve the maintainability.
> 
> * [ ]  Always use pre-increment unless the copy behavior in post-increment is really needed.

@yungyuc This issue is fixed, please review.

yungyuc:
> > While addressing the maintainability, what do you think about the next steps to make it faster than numpy?
> 
> I, to be honest, have no concrete idea about how to further improve it now. I'm thinking about parallel programming or speed up using hardware feature but I still have no idea about either whether it will work or how to implement it.

Could you check if the assembly is already compiled with SIMD?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: It should be.

yungyuc (approved):
> > LGTM. Please improve the maintainability.
> > 
> > * [ ]   Always use pre-increment unless the copy behavior in post-increment is really needed.
> 
> @yungyuc This issue is fixed, please review.

Beautiful

yungyuc:
> > > While addressing the maintainability, what do you think about the next steps to make it faster than numpy?
> > 
> > 
> > I, to be honest, have no concrete idea about how to further improve it now. I'm thinking about parallel programming or speed up using hardware feature but I still have no idea about either whether it will work or how to implement it.
> 
> Could you check if the assembly is already compiled with SIMD?

I wonder if you can provide a quick analysis for the difference between numpy and your implementation?

KHLee529:
> > > > While addressing the maintainability, what do you think about the next steps to make it faster than numpy?
> > > 
> > > 
> > > I, to be honest, have no concrete idea about how to further improve it now. I'm thinking about parallel programming or speed up using hardware feature but I still have no idea about either whether it will work or how to implement it.
> > 
> > 
> > Could you check if the assembly is already compiled with SIMD?
> 
> I wonder if you can provide a quick analysis for the difference between numpy and your implementation?

From the assembly, there's no SIMD instructions in both numpy library and my implementation.
As for the comparison between numpy and my implementation, I couldn't find any big difference between these two when considering only the loop part.

yungyuc:
> From the assembly, there's no SIMD instructions in both numpy library and my implementation. As for the comparison between numpy and my implementation, I couldn't find any big difference between these two when considering only the loop part.

Then you probably found a good point to improve.  Could you please summarize your analysis of the assembly here?

PR #487, opened 2025-03-26
Add Burgers' equation analytical solution

Gene0315:
Add 1d Burgers' equation analytical solution.

It is in "One -> Burgers equation".
image

### DIFF
--- a/modmesh/pilot/_burgers1d.py
+++ b/modmesh/pilot/_burgers1d.py
@@ -1,0 +1,261 @@
+import sys
+from dataclasses import dataclass
+
+import numpy as np
+
+import matplotlib
+import matplotlib.pyplot
+from matplotlib.backends.backend_qtagg import FigureCanvas
+from matplotlib.figure import Figure
+from matplotlib.pyplot import setp
+
+from PySide6.QtCore import QTimer, Slot, Qt
+from PySide6.QtWidgets import QDockWidget
+
+from PUI.state import State
+from PUI.PySide6.window import Window
+
+from .. import core as mcore
+from ._gui_common import PilotFeature
+from ._euler1d import SolverConfig, PlotConfig, PlotArea, ConfigWindow
+
+
+@dataclass
+class QuantityLine(object):
+    ana: matplotlib.lines.Line2D = None
+    axis: matplotlib.pyplot.axis = None
+    y_upper_lim: float = 0.0
+    y_bottom_lim: float = 0.0
+    name: str = ""
+    unit: str = ""
+
+    def update(self, x, y):
+        self.ana.set_xdata(x)
+        self.ana.set_ydata(y)
+        self.ana.figure.canvas.draw()
+
+
+class BurgersEquation:
+
+    R = 8.31446261815324
+
+    def __init__(self):
+        # Initial condition
+        self.coord = None
+        self.velocity = None
+
+        # Field data of the analytical solution.
+        self.coord_field = None
+        self.velocity_field = None
+
+    def build_constant(self, x1, x2, x3, velocity12, velocity23):
+        self.coord = np.array([x1, x2, x3], dtype='float64')
+        self.velocity = np.array([velocity12, velocity23], dtype='float64')
+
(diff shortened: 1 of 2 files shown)


modmesh/pilot/_burgers1d.py
```diff
@@ -0,0 +1,260 @@
+import sys
```
yungyuc, line 1: You should add a copyright comment block like other files.  Please do it in the next PR.

modmesh/pilot/_burgers1d.py
```diff
@@ -0,0 +1,260 @@
...
+from PySide6.QtCore import QTimer, Slot, Qt
+from PySide6.QtWidgets import QDockWidget, QWidget
+
+from .. import core as mcore
+from ._gui_common import PilotFeature
+from ._euler1d import SolverConfig, PlotConfig, PlotArea, ConfigWindow
+
+
+@dataclass
+class QuantityLine(object):
```
yungyuc, line 21: You should use some docstrings.  Please do it in a follow-up PR.

yungyuc: This is common code shared between multiple physical problems.  It should be factored out.

yungyuc (approved):
There are some minor points to fix but it looks nice!

image

Since you are using the same GUI code, it may be hard to change the GUI alone with the addition of the new solution.  Let's use follow-up PRs to improve.

To be improved in follow-up PRs (you can combine items in a PR or use multiple ones; it's up to you):
* Copyright block
* Docstrings
* Refactor to remove the unnecessarily duplicated code
* Improve the GUI code

yungyuc:
@Sean20405 please take a look at this PR.  Feel free to comment inline for discussions (even though it is merged already).

Sean20405:
I'm wonder whether it is possible to write a generate GUI for multiple functions. For xample,  we only need to override some functions or pass some variables or config, instead of writing a new one?

This may make the interface more consistent, as well as easy to maintain if we want to change the layout.

yungyuc:
> I'm wonder whether it is possible to write a generate GUI for multiple functions. For xample, we only need to override some functions or pass some variables or config, instead of writing a new one?

I think so, and it will be nice.

Gene0315:
Hi @yungyuc @Sean20405, I try to write a general GUI for 1DApp solver in issue https://github.com/solvcon/modmesh/pull/491

If I understand correctly, there are only some attributes in 1DApp relative to Solver, such as
- `st`: Solver
- `solver_config`: Parameters for solver initailization
- `plot_data`, `plot_config`: Properties in solver
- `plot_ana`, `plot_num`: Flag indicating whether to plot analytical and numerical data

Therefore, we only need to redefine the following functions for different solver:
- `populate_menu`: Set menu item for GUI.
- `init_solver_config`: Set `solver_config`.
- `set_plot_data`: Set `plot_data`, `plot_ana` and `plot_num`.
- `init_solver`: Set `st` and initialize analytical and numerical data.
- `update_step`: update analytical and numerical data at current step.

yungyuc:
> Hi @yungyuc @Sean20405, I try to write a general GUI for 1DApp solver in issue #491
> 
> If I understand correctly, there are only some attributes in 1DApp relative to Solver, such as
> 
> * `st`: Solver
> * `solver_config`: Parameters for solver initailization
> * `plot_data`, `plot_config`: Properties in solver
> * `plot_ana`, `plot_num`: Flag indicating whether to plot analytical and numerical data
> 
> Therefore, we only need to redefine the following functions for different solver:
> 
> * `populate_menu`: Set menu item for GUI.
> * `init_solver_config`: Set `solver_config`.
> * `set_plot_data`: Set `plot_data`, `plot_ana` and `plot_num`.
> * `init_solver`: Set `st` and initialize analytical and numerical data.
> * `update_step`: update analytical and numerical data at current step.

@Gene0315 this is good.  Could you please add the comments above in the PR #491 global comment section as well as the inline annotation?  It is not easy to track comments in different issues/PRs.

PR #489, opened 2025-04-03
Improve `take_along_axis` using SIMD

KHLee529:
# Description

The `take_along_axis` function improved in PR #486 can be further improved using SIMD.
This PR use Neon (Arm SIMD Instruction Set) to speed up `take_along_axis` in some data type.

> NOTE: Currently only the index check part of `take_along_axis` is improved. The data copy part of it is still ongoing.

## Profiling

Profiling is necessary to observe the efficacy of this PR, so a Makefile target pyprof is added, which simply runs all the python scripts in directory profiles and output the result to directory profiles/results. The mechanism of profiling in modmesh should also be discussed.

# Profile Result

From the following results, it is obvious that SIMD actually did some improvement on `take_along_axis`.

## N = 128 type: uint8

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 5.037E-03       | 1.000           | 3.425           |
| sa         | 1.471E-03       | 0.292           | 1.000           |
| simd       | 9.293E-04       | 0.184           | 0.632           |

## N = 1K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 3.758E-03       | 1.000           | 2.284           |
| sa         | 1.646E-03       | 0.438           | 1.000           |
| simd       | 1.179E-03       | 0.314           | 0.716           |

## N = 8K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.242E-02       | 1.000           | 1.960           |
| sa         | 6.334E-03       | 0.510           | 1.000           |
| simd       | 3.817E-03       | 0.307           | 0.603           |

## N = 64K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 9.501E-02       | 1.000           | 1.602           |
| sa         | 5.930E-02       | 0.624           | 1.000           |
| simd       | 3.811E-02       | 0.401           | 0.643           |

## N = 512K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 8.673E-01       | 1.000           | 1.247           |
| sa         | 6.953E-01       | 0.802           | 1.000           |
| simd       | 5.478E-01       | 0.632           | 0.788           |

## N = 4M type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.445E+01       | 1.000           | 1.999           |
| sa         | 7.230E+00       | 0.500           | 1.000           |
| simd       | 6.305E+00       | 0.436           | 0.872           |

## N = 32M type: uint64

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 3.115E+02       | 1.000           | 1.600           |
| sa         | 1.947E+02       | 0.625           | 1.000           |
| simd       | 1.864E+02       | 0.599           | 0.958           |


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -224,6 +228,10 @@ class SimpleArrayMixinSort
     SimpleArray<uint64_t> argsort(void);
     template <typename I>
     A take_along_axis(SimpleArray<I> const & indices);
+#if defined(__aarch64__)
+    template <typename I>
+    A take_along_axis_simd(SimpleArray<I> const & indices);
```
KHLee529, line 229: SIMD integration is currently parallel to the current implementation in order to do profiling and compare between.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +894,108 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    static_assert(std::is_integral_v<I>, "I must be integral type");
+    auto athis = static_cast<A *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(Formatter() << "SimpleArray::take_along_axis(): currently only support 1D array"
+                                             << " but the array is " << athis->ndim() << " dimension");
+    }
+
+    size_t max_idx = athis->shape()[0];
+    I const * oor_ptr = check_index_range(indices, max_idx);
```
KHLee529, line 1006: First check every elements in the indices array is in the valid range.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +894,108 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+
+        throw std::out_of_range(err_msg);
+    }
+
+    I const * src = indices.begin();
+    I const * const end = indices.end();
+    A ret(indices.shape());
+    T * data = athis->begin();
+    T * dest = ret.begin();
+    detail::buffer_cpy(dest, data, src, end);
```
KHLee529: Second move the data to correct place.

KHLee529:
@yungyuc Please review this SIMD integration in `take_along_axis` function. As mentioned in the inline comments, the modification will be merged into `take_along_axis` function instead of being standalone if this implementation is acceptable.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -36,6 +36,77 @@ namespace modmesh
 
 namespace detail
 {
+
+#if defined(__aarch64__)
```
KHLee529: This is only supported by Arm machines.

Makefile
```diff
@@ -118,6 +118,21 @@ pytest: buildext
 	env $(RUNENV) \
 		$(PYTEST) $(PYTEST_OPTS) tests/
 
+PROFFILES = $(shell find profiles -type f -name 'profile_*.py' | sort)
+PROFRESDIR = profiles/results
+
+.PHONY: pyprof
+pyprof: buildext $(PROFFILES)
```
KHLee529, line 125: New Makefile target for profiling python scripts.

KHLee529:
My profiling scripts and Makefile target for profiling is also added to this PR.

j8xixo12:
Why does the effect of SIMD become less as N gets larger ？

Makefile
yungyuc: I like the name `pyprof`.

cpp/modmesh/buffer/SimpleArray.cpp
yungyuc: How difficult it is to separate the SIMD code to distinct files?  Please make a guess for the difficulties, maybe in work hours.  We will use the guess to evaluate how and when to make the separation.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Rename `take_along_axis_simd()` to `take_along_axis_neon()` because the added SIMD code is specific to ARM NEON so far.

yungyuc: This is an API in `SimpleArray` and cannot go to a separate SIMD file.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Good annotation here.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -201,15 +201,111 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+#if defined(__aarch64__)
+        const char * py_typename = pybind11::detail::obj_class_name(indices.ptr());
+        const char * type = strstr(py_typename, "_modmesh.SimpleArray");
+        if (type == nullptr)
+        {
+            return pybind11::cast(std::move(self));
+        }
+        type += strlen("_modmesh.SimpleArray");
+
+        if (strcmp(type, "Int8") == 0)
```
yungyuc: Please add a "TODO" comment here to remind us to refactor this by following the way `SimpleArrayPlex` is implemented: https://github.com/solvcon/modmesh/blob/3854842b759fb055230332f3e35ccf4df47f262c/cpp/modmesh/buffer/SimpleArray.cpp#L172

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -201,15 +201,111 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        {
+            return pybind11::cast(self.take_along_axis(indices.cast<SimpleArrayUint64>()));
+        }
+
+        return pybind11::cast(std::move(self));
+    }
+
+    static pybind11::object take_along_axis_simd(wrapped_type & self, pybind11::object const & indices)
+    {
+#if defined(__aarch64__)
```
yungyuc: This should be a runtime check, not compile time.

profiles/profile_sort.py
```diff
@@ -0,0 +1,178 @@
+import functools
```
yungyuc: Rename the directory to be `profiling`.

yungyuc (changes requested):
It may take a little bit of time to complete, but this is a good patch.

- [x] Evaluate whether to separate SIMD code to distinct files.
- [x] Rename `take_along_axis_simd()` to `take_along_axis_neon()` because the added SIMD code is specific to ARM NEON so far.
- [ ] Runtime check for existence of SIMD and dispatch to the proper helper.
- [x] Rename the directory `profiles` to be `profiling`.

---

> Why does the effect of SIMD become less as N gets larger ？

@j8xixo12 Good question.  It should be answered by more study after this PR is merged.

KHLee529:
> Why does the effect of SIMD become less as N gets larger ？

@j8xixo12 @yungyuc  Here is my inference. 
1. The gap between norma non-SIMD version of `SimpleArray` and `ndarray` decrease while N increases. The reason might be that there's more pre-processing codes to identify, for example, the type of index etc. in `ndarray` which takes a constant amount of time, and this amount of time stands for higher ratio to the total time when N is less the iteration takes less time also. Therefore, the `SimpleArray` version can be quite faster due to the lack of these pre-processing.
2. The effect of SIMD decrease while N increases, which is actually the result I expected. This is because that the SIMD instructions treats a fixed width register as the unit of manipulation. When the size of the data type is smaller, for example, `uint8_t` only takes 1 byte,  The SIMD processing could parallelize more operations. When it comes to wider datatype, `uint64_t` for example, the number of instructions run is almost the same as pure iteration without SIMD integration.

The 2nd inference could be better confirmed by parallelizing arithmetic operation such as addition, subtraction, etc.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -36,6 +36,77 @@ namespace modmesh
...
+#define vec_typ(typ, N) typ##x##N##_t
+#define t_typ(typ) typ##_t
+
+#define IMPL_CHECK_IDX_RNG_SIMD(typ, N_vec, typ_symbol)                                                       \
```
tigercosmos: nickpick: `DECL_MM_XXX`

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -36,6 +36,77 @@ namespace modmesh
 
 namespace detail
 {
+
+#if defined(__aarch64__)
+#define vec_typ(typ, N) typ##x##N##_t
```
tigercosmos: I would recommend to not use `typ`. Instead spelling out will be better `_type`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +894,108 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+        if (idx < 0 || idx > max_idx)
+        {
+            return src;
+        }
+        ++src;
+    }
+    return nullptr;
+}
+
+#define DECL_CHECK_IDX_RNG_SIMD(typ) \
```
tigercosmos: nickpick: `DECL_MM_XXX`

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -201,15 +201,111 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        const char * py_typename = pybind11::detail::obj_class_name(indices.ptr());
+        const char * type = strstr(py_typename, "_modmesh.SimpleArray");
+        if (type == nullptr)
+        {
+            return pybind11::cast(std::move(self));
+        }
+
+        type += strlen("_modmesh.SimpleArray");
+
+        if (strcmp(type, "Int8") == 0)
```
tigercosmos: It seems that we can use macro here.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -36,6 +36,77 @@ namespace modmesh
...
+        t_typ(typ) const * const end = indices.end();                                                         \
+                                                                                                              \
+        for (; src <= end - N_vec; src += N_vec)                                                              \
+        {                                                                                                     \
+            data_vec = vld1q_##typ_symbol(src);                                                               \
+            cmp_vec = vcgeq_##typ_symbol(data_vec, max_vec);                                                  \
+            if (vgetq_lane_##typ_symbol(cmp_vec, 0) ||                                                        \
+                vgetq_lane_##typ_symbol(cmp_vec, 1))                                                          \
+            {                                                                                                 \
+                goto OUT_OF_RANGE;                                                                            \
```
tigercosmos: @yungyuc  do you accept this?

yungyuc: GOTO in asm is OK.  But I need this to be moved in separate files before get a right feeling for the code structure. In short, I don't know yet.

KHLee529:
@yungyuc I've done some research on [xsimd](https://github.com/xtensor-stack/xsimd) and runtime dispatching. Currently I still couldn't make precise work load estimation on separating the SIMD into a separate file.
Here is my rough plan.

First, wrapping the SIMD function calls on a continuous buffer into helper functions according to the basic functionality.
- e.g. `add_simd` to add two buffer element-wise, `sub_simd` to subtract instead.
- e.g. `check_between_simd` to check whether all the values stored in buffer is within a certain range [min, max].

This might not be too difficult and it might take possibly 4 hr to rearrange the SIMD function calls used in this PR into a separate file. I would like to also wrap some basic arithmetic operation as well if it takes not too long.
Then, add the runtime dispatcher which takes the boolean values determined in compile time as the support flag. This, however, still confuse me a lot since I still have no idea what will happen when I compile the intrinsic functions on another platform without cross-compiling, e.g. compile neon intrinsic library on x86 compiler. From this on, I, therefore, cannot estimate the work load neither make further plan.

yungyuc:
> * [ ]  Evaluate whether to separate SIMD code to distinct files.

@KHLee529 this is to evaluate. Please focus on evaluation.

KHLee529:
> > * [ ]   Evaluate whether to separate SIMD code to distinct files.
> 
> @KHLee529 this is to evaluate. Please focus on evaluation.

I think separating SIMD code to distinct file is better and worthwhile due to better maintainability and readability. I estimate that the time cost for separating current codes will be about 8 hr.

yungyuc:
> I think separating SIMD code to distinct file is better and worthwhile due to better maintainability and readability. I estimate that the time cost for separating current codes will be about 8 hr.

Then let's try to separate the SIMD code to a distinct file.  The current mixture is hard to review and maintain.

KHLee529:
Progress record
- [x] Evaluate whether to separate SIMD code to distinct files.
- [x]  Rename take_along_axis_simd() to take_along_axis_neon() because the added SIMD code is specific to ARM NEON so far.
- [ ]  Runtime check for existence of SIMD and dispatch to the proper helper.
- [x]  Rename the directory profiles to be profiling.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -36,6 +36,77 @@ namespace modmesh
 
 namespace detail
 {
+
+#if defined(__aarch64__)
+#define vec_typ(__type, __N) __type##x##__N##_t
```
yungyuc: `_` prefix is reserved for compilers.  Please don't name anything starting with `_`.

cpp/modmesh/buffer/SimpleArray.cpp
yungyuc: This seems to be for NEON.  Why is it still in `SimpleArray.cpp` but not in the new `neon.hpp`, or yet another `neon.cpp`?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +895,98 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+#if defined(__aarch64__)
+#define check_type_range(_type, max_val)                                \
+    do {                                                                \
+        constexpr _type __type_max = std::numeric_limits<_type>::max(); \
```
yungyuc: Not `_` prefix in names.

cpp/modmesh/simd/neon.hpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
```
yungyuc: Add copyright.

cpp/modmesh/simd/wrap/neon_wrap.hpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
```
yungyuc: What are the `neon_wrap.*` files for?  Is there a reason the code should not be included in `neon.hpp` and/or `neon.cpp`?

yungyuc (changes requested):
@KHLee529 I would need some elaboration of the new code, but it seems to be in a right direction to separate NEON code from the basic array code.

Items must be changed:
- [x] `_` prefix is reserved for compilers.  Please don't name anything starting with `_`.
- [x] Clarify why NEON code is still in `SimpleArray.cpp`.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: This function is renamed back to `take_along_axis_simd()` since its implementation now possibly support different SIMD.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -894,51 +885,24 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+        if (MaxVal >= DataTypeMax__ && DataTypeMin__ == 0)                       \
+        {                                                                        \
+            return nullptr;                                                      \
+        }                                                                        \
     } while (0)
 
 template <typename T>
 T const * detail::check_index_range(SimpleArray<T> const & indices, size_t max_idx)
 {
     check_type_range(T, max_idx);
```
KHLee529: Check if the maximum of the data type containing indices already smaller than the maximal index. If true, every element in the array containing indices must be legal so no further check needs to be done.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -894,51 +885,24 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
-    {
-        T const idx = *src;
-        if (idx < 0 || idx > max_idx)
-        {
-            return src;
-        }
-        ++src;
-    }
-    return nullptr;
+    return simd::check_between<T>(indices.begin(), indices.end(), 0, max_idx);
```
KHLee529, line 975: Check if every index in `indices` is in the range [0, max_idx) using SIMD library.

cpp/modmesh/simd/simd.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
```
KHLee529, line 1: This library contain the main interface of SIMD functions. Users should call functions only in this library.

cpp/modmesh/simd/simd.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+
+namespace simd
+{
+
+// Check if each element from start to end (excluded end) is within the range [min_val, max_val)
+template <typename T>
+const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
+{
+#if defined(__aarch64__)
```
KHLee529: Currently still support compile time dispatching, runtime dispatching will be done later.

cpp/modmesh/simd/simd_generic.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
KHLee529, line 1: This library implements the SIMD function called when no SIMD ISA is available.

cpp/modmesh/simd/neon/neon_wrap.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
KHLee529: This header wraps the NEON intrinsic functions to general template function because they're original different functions for different vector data type.

cpp/modmesh/simd/neon/neon_type.hpp
```diff
@@ -0,0 +1,104 @@
+#pragma once
```
KHLee529, line 1: This header bounds the basic data type to its corresponding NEON vector data type using template.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
```
KHLee529, line 1: This header implement the SIMD library functions with NEON instruction set.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -212,98 +214,72 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
 
     static pybind11::object take_along_axis(wrapped_type & self, pybind11::object const & indices)
     {
-        const char * py_typename = pybind11::detail::obj_class_name(indices.ptr());
-        const char * type = strstr(py_typename, "_modmesh.SimpleArray");
-        if (type == nullptr)
+        std::string py_typename(pybind11::detail::obj_class_name(indices.ptr()));
```
KHLee529, line 241: Take out type name from python object and extract the data type.

KHLee529:
@yungyuc I've separated the SIMD implementations to other files. Some minor fix will be done later includes

- [ ] Runtime dispatching SIMD ISA
- [x] Copy right comments

Please review them.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +885,81 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+template <typename T>
+T const * detail::check_index_range(SimpleArray<T> const & indices, size_t max_idx)
+{
+    check_type_range(T, max_idx);
+
+    return simd::check_between<T>(indices.begin(), indices.end(), 0, max_idx);
+}
+
+template <typename T, typename I>
+void detail::buffer_cpy(T * dest, T const * data, I const * begin, I const * const end)
```
yungyuc: Name it in full as `buffer_copy`.  It is confusing to use `cpy` as an API name.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +885,81 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+{
+    T * dst = dest;
```
yungyuc, line 981: Check `dest == begin` and do nothing when it's true.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -201,15 +201,85 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        if (found == std::string::npos)
+        {
+            return pybind11::cast(std::move(self));
+        }
+
+        py_typename.replace(0, strlen("_modmesh.SimpleArray"), "");
+        py_typename[0] = tolower(py_typename[0]);
+        const DataType dt(py_typename);
+
+#define DECL_MM_TAKE_ALONG_AXIS_TYPED(DataType__) \
```
yungyuc: Don't postfix the unnecessary underscore `_`.  In general, don't prefix or postfix underscore `_` in names in modmesh unless you have a very good reason and spell it out in review.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+/*
+ * Copyright (c) 2025, Yung-Yu Chen <yyc@solvcon.net>
```
yungyuc: Use your name and copy the full BSD license.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -0,0 +1,91 @@
...
+namespace modmesh
+{
+
+namespace simd
+{
+
+namespace neon
+{
+
+template <typename T, typename std::enable_if_t<!type::has_vectype<T>, int> = 0>
```
yungyuc: Move `typename std::enable_if_t<!type::has_vectype<T>, int> = 0` (I think this is our good old friend of SFINAE) to be inside the function template using `static_assert`:

```cpp
static_assert(!type::has_vectype<T>);
```

SFINAE works but is very hard to read.  More information in https://en.cppreference.com/w/cpp/language/sfinae

cpp/modmesh/simd/neon/neon_type.hpp
yungyuc: You grew fond of template meta programming.

cpp/modmesh/simd/neon/neon_wrap.hpp
yungyuc: Rename the files to be `neon_alias.*`.  "wrap" has been used to mean Python/C++ wrappers.

yungyuc (changes requested):
Reviewing 1000 lines of diff is hard, but this is a wonderful leap of maturity.  Can't wait to see its merging after polishing the details.

- [x] Add full license texts and your name and email to the files you solely created.
- [x] Clarify `buffer_copy`
- [ ] Check `dest == begin` and do nothing when it's true.
- [x] Do no prefix or postfix underscore `_` in name in modmesh.
- [ ] Use [static_assert](https://en.cppreference.com/w/cpp/language/static_assert) to avoid [SFINAE](https://en.cppreference.com/w/cpp/language/sfinae).
- [x] Rename the files for NEON function aliases as `*_alias.*`.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: If `dest == begin`, it only set two local variables, since the whole while loop will be skipped. Is the cost of making local variables more than branching?

cpp/modmesh/simd/neon/neon.hpp
KHLee529: I got stuck when I'm trying to make it into `static_assert` since there're two declaration for different types. Compile error indicating duplicated function declaration appears when I move the same condition into the function body using `static_assert`. Are there any way to get through it?

yungyuc: Please post the error messages.  I do not have a crystal ball.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: The `while` loop is not skipped when `dest == begin`, but please show me if I am wrong.

contrib/standalone_buffer/Makefile
```diff
@@ -50,6 +50,9 @@ $(SETUPROOT)/BufferExpander.o: $(SETUPROOT)/modmesh/buffer/BufferExpander.cpp Ma
 $(SETUPROOT)/SimpleArray.o: $(SETUPROOT)/modmesh/buffer/SimpleArray.cpp Makefile
 	c++ $(CFLAGS) -c $< -o $@
 
+$(SETUPROOT)/neon_wrap.o: $(SETUPROOT)/modmesh/simd/neon/neon_wrap.cpp Makefile
```
yungyuc: Thank you for taking care of the special build.  I forgot to mention it in the first place.

yungyuc (changes requested):
Good progress.  Let's keep moving.  Remaining points:

- [x] Check `dest == begin` and do nothing when it's true.
- [x] Use [static_assert](https://en.cppreference.com/w/cpp/language/static_assert) to avoid [SFINAE](https://en.cppreference.com/w/cpp/language/sfinae).

cpp/modmesh/simd/neon/neon.hpp
KHLee529: Here are the error messages.
```
/Users/khlee/Documents/dev/modmesh/cpp/modmesh/simd/neon/neon.hpp:58:11: error: redefinition of 'check_between'
          ^
/Users/khlee/Documents/dev/modmesh/cpp/modmesh/simd/neon/neon.hpp:49:11: note: previous definition is here
          ^
```

yungyuc: I see.  In this case, you do need SFINAE and it cannot be replaced by `static_assert`.  My bad.  And I wonder if we should use `void` instead of `int`?  I do not understand what purpose is `int` for.

yungyuc (commented):
I see, you do need SFINAE in `neon.hpp`.  My bad.

cpp/modmesh/simd/simd.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace simd
+{
+
+// Check if each element from start to end (excluded end) is within the range [min_val, max_val)
+template <typename T>
+const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
+{
+    using namespace detail;
+    switch (detectSIMD())
```
KHLee529: Runtime SIMD dispatching

cpp/modmesh/simd/simd_support.cpp
```diff
@@ -0,0 +1,150 @@
...
+namespace modmesh
+{
+
+namespace simd
+{
+
+namespace detail
+{
+
+SIMDFeature detectSIMD(void)
```
KHLee529: This file is to detect hardware SIMD feature.

KHLee529:
@yungyuc The runtime SIMD dispatching is finally done, please review them. Other change requests are also modified.

cpp/modmesh/simd/simd_support.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace modmesh
+{
+
+namespace simd
+{
+
+namespace detail
+{
+
+SIMDFeature detectSIMD()
```
KHLee529: This implements the cpu feature detection, currently only support NEON.

Besides, I originally wrote `SIMDFeature detectSIMD(void)` which I think more reasonable but that will cause linting CI failure.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -876,6 +885,81 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+#define check_type_range(DataType, MaxVal)                                       \
+    do {                                                                         \
+        constexpr DataType DataTypeMax__ = std::numeric_limits<DataType>::max(); \
```
yungyuc: Please remember to not prefix nor postfix `_` 🙂

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: We concluded that no same-location check should be implemented here.  @KHLee529 next time please leave a message to record what we discussed in a side channel.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+
+namespace simd
+{
+
+namespace neon
+{
+
+#if defined(__aarch64__)
+template <typename T, typename std::enable_if_t<!type::has_vectype<T>> * = nullptr>
```
yungyuc, line 49: SFINAE is needed.  This comment is to record the conclusion of previous discussions.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    if (ptr != end)
+    {
+        ret = generic::check_between<T>(ptr, end, min_val, max_val);
+    }
+
+    return ret;
+
+OUT_OF_RANGE:
```
yungyuc, line 94: It just occurred to me that I do not know what will happen when using `goto` within a template.

If it is UB (undefined behavior) or forbidden, we will need to move the implementation elsewhere.  We'll see.

cpp/modmesh/simd/simd_support.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace modmesh
+{
+
+namespace simd
+{
+
+namespace detail
+{
+
+enum SIMDFeature
```
yungyuc: Treat an acronym like a word for naming, so it should be named as `SimdFeature` here.  For snake casing, write `simd_something`.

cpp/modmesh/simd/simd_support.hpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    SIMD_NONE = 0,
```
yungyuc, line 42: It is correct to use ALL_CAPS for enum constants.

cpp/modmesh/simd/simd_support.hpp
```diff
@@ -0,0 +1,62 @@
...
+    SIMD_SSSE3,
+    SIMD_SSE41,
+    SIMD_SSE42,
+    SIMD_AVX,
+    SIMD_AVX2,
+    SIMD_AVX512,
+    SIMD_UNKNOWN
+};
+
+SIMDFeature detectSIMD(void);
```
yungyuc: Function uses snake case, so `detect_simd()`

cpp/modmesh/simd/simd_support.cpp
yungyuc: Please leave a comment in the helper function that the detection result should be cached.  The detection needs only to be done once because it will not be changed throughout the process.

yungyuc (changes requested):
We are close to the finalization.

- [x] Treat an acronym like a word for naming, so it should be named as `SimdFeature` here.  For snake casing, write `simd_something`.
- [x] Function uses snake case, so `detect_simd()`
- [x] Leave a comment in the `detect_simd()` helper function that the detection result should be cached.  The detection needs only to be done once because it will not be changed throughout the process.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -953,6 +962,76 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+template <typename T>
+T const * detail::check_index_range(SimpleArray<T> const & indices, size_t max_idx)
+{
+    constexpr T DataTypeMax = std::numeric_limits<T>::max();
```
KHLee529, line 968: The unncessary `check_type_range` is removed

cpp/modmesh/simd/simd_support.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace simd
+{
+
+namespace detail
+{
+
+SimdFeature detect_simd()
+{
+    static SimdFeature CurrentFeature = SIMD_UNKNOWN;
```
KHLee529, line 54: SIMD feature is cached as static variable because platform feature must not change during runtime.

KHLee529:
@yungyuc all the change requests are resolved, please review.

cpp/modmesh/simd/simd_support.cpp
yungyuc: Sweet

yungyuc (approved):
Nothing I can complain about.

KHLee529:
I've run profiling again after all these are finished. The results shows that the SIMD version of uint32 is actually slower than normal version. This should be improve afterward. 

# Profiling results

## N = 128 type: uint8

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 5.063E-03       | 1.000           | 3.606           |
| sa         | 1.404E-03       | 0.277           | 1.000           |
| simd       | 9.710E-04       | 0.192           | 0.692           |

## N = 1K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 3.921E-03       | 1.000           | 2.168           |
| sa         | 1.808E-03       | 0.461           | 1.000           |
| simd       | 1.475E-03       | 0.376           | 0.816           |

## N = 8K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.266E-02       | 1.000           | 1.919           |
| sa         | 6.596E-03       | 0.521           | 1.000           |
| simd       | 5.571E-03       | 0.440           | 0.845           |

## N = 64K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 9.536E-02       | 1.000           | 1.673           |
| sa         | 5.701E-02       | 0.598           | 1.000           |
| simd       | 6.755E-02       | 0.708           | 1.185           |

## N = 512K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 8.498E-01       | 1.000           | 1.234           |
| sa         | 6.888E-01       | 0.811           | 1.000           |
| simd       | 7.662E-01       | 0.902           | 1.112           |

## N = 4M type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.450E+01       | 1.000           | 1.980           |
| sa         | 7.323E+00       | 0.505           | 1.000           |
| simd       | 7.916E+00       | 0.546           | 1.081           |

## N = 32M type: uint64

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 3.151E+02       | 1.000           | 1.605           |
| sa         | 1.964E+02       | 0.623           | 1.000           |
| simd       | 2.164E+02       | 0.687           | 1.102           |

yungyuc:
> I've run profiling again after all these are finished. The results shows that the SIMD version of uint32 is actually slower than normal version. This should be improve afterward.

Could you please file an issue ticket to track the improvement?

The runtime is actually mixed.  SIMD is only slower when then array gets larger.

KHLee529:
> > I've run profiling again after all these are finished. The results shows that the SIMD version of uint32 is actually slower than normal version. This should be improve afterward.
> 
> Could you please file an issue ticket to track the improvement?
> 
> The runtime is actually mixed. SIMD is only slower when then array gets larger.

No problem, issue #532 is created to track this issue.

PR #490, opened 2025-04-03
refactor: Completely remove PUI in modmesh

### LINKED ISSUE #481: Make PUI an optional dependency
Review modmesh code and to make it independent from PUI.  Keep the third-party installation script so that devs can optionally include it when they want.

Remove PUI from the CI run.

---

Relevant discussion is in https://github.com/solvcon/modmesh/issues/462#issuecomment-2709993963 .  Quote here (there is still `PUI` used in the code base):

```
$ ag PUI modmesh/ tests/
modmesh/params.py
32:import PUI
33:from PUI.PySide6 import State, PuiInQt, Window, VBox, Table, TextField
36:if tuple([int(x) for x in PUI.__version__.split(".")]) != (0, 3):
37:    raise ValueError("PUI too old")
99:            from PUI.PySide6 import StateDict

tests/test_pilot.py
38:    import PUI.PySide6
40:    # Bypass PUI import error if modmesh is built without Qt PUI may not be
42:    # If modmesh is built with Qt, the PilotTC will check if PUI is working
54:        self.assertTrue(hasattr(PUI.PySide6, "PUINode"))
55:        self.assertTrue(hasattr(PUI.PySide6, "PUIView"))
56:        self.assertEqual(PUI.PySide6.PUI_BACKEND, "PySide6",
57:                         "PUI backebd mismatch")
```

Sean20405:
This PR is related to https://github.com/solvcon/modmesh/issues/481. It removes the import test for PUI,  `RParameter` and its related file.


yungyuc (changes requested):
LGTM.  Could you please also change the comment at https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/.github/workflows/devbuild.yml#L620

from 

```
# PUI is necessary
Copy-Item -Path ".\thirdparty" -Destination $destination -Recurse
```

to 

```
# Also include potential thirdparty
Copy-Item -Path ".\thirdparty" -Destination $destination -Recurse
```

---

A side note, there is code to handling thirdparty Python file importing in https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/tests/test_pylibmgr.py#L41 , but it does not need to be changed.  Only FYI.

Sean20405:
> LGTM. Could you please also change the comment at
> 
> https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/.github/workflows/devbuild.yml#L620
> 
> from
> 
> ```
> # PUI is necessary
> Copy-Item -Path ".\thirdparty" -Destination $destination -Recurse
> ```
> 
> to
> 
> ```
> # Also include potential thirdparty
> Copy-Item -Path ".\thirdparty" -Destination $destination -Recurse
> ```

This has been done in:
https://github.com/solvcon/modmesh/blob/c32c673954cb8b4fd26cc190ba6641a2ace86a81/.github/workflows/devbuild.yml#L620

Since the CI may need some time to run, I re-request a review in advance.

Sean20405:
Also, I found that: 
- Our [`CMakeLists.txt`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/CMakeLists.txt#L44) will install PUI automatically.
- In GitHub workflow [`devbuild.yml`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/.github/workflows/devbuild.yml#L620), we also wrap PUI into Windows portable file.

According to https://github.com/solvcon/modmesh/issues/481#issuecomment-2777297096, they would be kept the same.

yungyuc (changes requested):
> Also, I found that:
> 
> * Our [`CMakeLists.txt`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/CMakeLists.txt#L44) will install PUI automatically.
> * In GitHub workflow [`devbuild.yml`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/.github/workflows/devbuild.yml#L620), we also wrap PUI into Windows portable file.
> 
> According to [#481 (comment)](https://github.com/solvcon/modmesh/issues/481#issuecomment-2777297096), they would be kept the same.

`CMakeLists.txt` needs some change.  Please see  https://github.com/solvcon/modmesh/issues/481#issuecomment-2777297096

> > * Our [`CMakeLists.txt`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/CMakeLists.txt#L44) will install PUI automatically.
> 
> **Installation of PUI should be changed to be opt-in.** That is, the code to install PUI in thirdparty can stay, but it should not be executed by default.
> 
> > * In GitHub workflow [`devbuild.yml`](https://github.com/solvcon/modmesh/blob/cd676e8d1feffde42951273522a0ed0c254ac9c5/.github/workflows/devbuild.yml#L620), we also wrap PUI into Windows portable file.
> 
> This should be treated in the same way as the PUI installation in `CMakeLists.txt`.

Sean20405:
> CMakeLists.txt needs some change. Please see https://github.com/solvcon/modmesh/issues/481#issuecomment-2777297096

Oops. Sorry for misunderstanding.

Then instead of just comment it, how about adding an option in CMake to let users choose whether to install when they prompt a command. Like:

```
option(INSTALL_PUI "install PUI" OFF)
message(STATUS "INSTALL_PUI : ${INSTALL_PUI}")
# ...
if(INSTALL_PUI)
    execute_process(
        COMMAND python3 PUI_install.py
        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/thirdparty 
    )
endif()
```

Or we don't want users to install it since it indeed has not been used, so it should be removed.

I think the first one will be more like an "optional choose"(?)

Sean20405:
Hi @yungyuc! What about your thought? Or it can depend on me.

yungyuc:
> Hi @yungyuc! What about your thought? Or it can depend on me.

You can make a decision.  Both works.  But I like the cmake option.

yungyuc:
@Sean20405 is it ready for review again?

Sean20405:
Hi @yungyuc,

Yes! This is ready for review again. Thanks for kindly reminder!

CMakeLists.txt
```diff
@@ -41,6 +41,13 @@ message(STATUS "USE_PYTEST_HELPER_BINDING: ${USE_PYTEST_HELPER_BINDING}")
 if(BUILD_QT)
     option(QT3D_USE_RHI "Qt use RHI" OFF)
     message(STATUS "QT3D_USE_RHI: ${QT3D_USE_RHI}")
+endif()
+
+# PUI-related code has been migrated to PySide6, so it doesn't need to be installed by default.
+# Yet, you can still install it by setting the INSTALL_PUI option to ON.
+option(INSTALL_PUI "install PUI" OFF)
+message(STATUS "INSTALL_PUI : ${INSTALL_PUI}")
+if(INSTALL_PUI)
```
yungyuc, line 50: Please check if `BUILD_QT` is turned on, and issue a warning message if it is not, because installing PUI without Qt GUI built is useless.

yungyuc (changes requested):
Need to improve a logical defect in CMakeLists.txt.

- [ ] Warn when `INSTALL_PUI` is turned on but `BUILD_QT` is not.

CMakeLists.txt
```diff
@@ -41,10 +41,21 @@ message(STATUS "USE_PYTEST_HELPER_BINDING: ${USE_PYTEST_HELPER_BINDING}")
...
-    )
+endif()
+
+# PUI-related code has been migrated to PySide6, so it doesn't need to be installed by default.
+# Yet, you can still install it by setting the INSTALL_PUI option to ON.
+option(INSTALL_PUI "install PUI" OFF)
+message(STATUS "INSTALL_PUI : ${INSTALL_PUI}")
+if(INSTALL_PUI)
+    if(NOT BUILD_QT)
+        message(WARNING "INSTALL_PUI is enabled, but BUILD_QT is not enabled. PUI depends on Qt, so this configuration may not work as expected.")
```
Sean20405: Add a detection here. If not `BUILD_QT`, then PUI will not be install even with `INSTALL_PUI`

Sean20405:
@yungyuc This PR is ready for next review!

Add a detection. If not `BUILD_QT`, then PUI will not be install even with `INSTALL_PUI`.

A change on CMakeLists:
https://github.com/solvcon/modmesh/blob/676aa5479968c5d62696f48b98c21c08bbf1704e/CMakeLists.txt#L50-L59

CMakeLists.txt
yungyuc: Do not resolve comments.  Resolving comments makes it hard to track.

CMakeLists.txt
yungyuc: This is incorrect.

When `NOT BUILD_QT`, cmake should still install PUI as requested, but warn that it is not executed with `BUILD_QT` disabled.

yungyuc (changes requested):
> @yungyuc This PR is ready for next review!
> 
> Add a detection. If not `BUILD_QT`, then PUI will not be install even with `INSTALL_PUI`.

No, it is not what I asked:

> Warn when `INSTALL_PUI` is turned on but `BUILD_QT` is not.

I asked to warn but I did not ask to skip the installation.  Please correct the behavior.

- [x] Warn when `INSTALL_PUI` is turned on but `BUILD_QT` is not.

CMakeLists.txt
yungyuc: I got it fixed.

yungyuc (approved):

CMakeLists.txt
Sean20405: Thanks!!!

yungyuc: No problem

PR #492, opened 2025-04-05
feat: add bluestein fft implementation and test cases

j8xixo12:
In this PR I have added:
- Bluestein Fourier transform algorithm for processing arbitrary length input data
- Add non radix-2 length test cases for testing Bluestein Fourier transform algorithm
- fft function will select different algorithm based on input data length

### DIFF
--- a/cpp/modmesh/transform/fourier.hpp
+++ b/cpp/modmesh/transform/fourier.hpp
@@ -13,8 +13,12 @@ namespace detail
 {
 
 size_t bit_reverse(size_t n, const size_t bits);
+size_t next_power_of_two(size_t n);
 
 } /* end namespace detail */
+
+template <template <typename> class T1, typename T2>
+void fft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out);
 
 // TODO: The template of template is too complicate, we should find a way to make it easier.
 template <template <typename> typename T1, typename T2>
@@ -25,7 +29,7 @@ void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
     {
         for (size_t j = 0; j < N; ++j)
         {
-            T2 tmp = -2.0 * pi<T2> * i * j / N;
+            T2 tmp = -2.0 * pi<T2> * i * j / static_cast<T2>(N);
             T1<T2> twiddle_factor{std::cos(tmp), std::sin(tmp)};
             out[i] += in[j] * twiddle_factor;
         }
@@ -33,7 +37,7 @@ void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
 }
 
 template <template <typename> class T1, typename T2>
-void fft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
+void fft_radix_2(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
 {
     size_t N = in.size();
     const unsigned int bits = static_cast<unsigned int>(std::log2(N));
@@ -48,7 +52,7 @@ void fft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
     for (size_t size = 2; size <= N; size *= 2)
     {
         size_t half_size = size / 2;
-        T2 angle_inc = -2.0 * pi<T2> / size;
+        T2 angle_inc = -2.0 * pi<T2> / static_cast<T2>(size);
 
         for (size_t i = 0; i < N; i += size)
         {
@@ -72,7 +76,7 @@ template <template <typename> class T1, typename T2>
 void ifft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
 {
     size_t N = in.size();
-    SimpleArray<T1<T2>> in_conj(N);
+    SimpleArray<T1<T2>> in_conj{modmesh::small_vector<size_t>{N}, T1<T2>{0.0, 0.0}};
 
     for (size_t i = 0; i < N; ++i)
     {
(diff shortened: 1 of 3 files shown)


cpp/modmesh/transform/fourier.hpp
```diff
@@ -25,15 +29,15 @@ void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
     {
         for (size_t j = 0; j < N; ++j)
         {
-            T2 tmp = -2.0 * pi<T2> * i * j / N;
+            T2 tmp = -2.0 * pi<T2> * i * j / static_cast<T2>(N);
```
j8xixo12, line 32: Preventing type conversion after dividing a `size_t`, which can cause loss of numerical precision.

cpp/modmesh/transform/fourier.hpp
```diff
@@ -72,7 +76,7 @@ template <template <typename> class T1, typename T2>
 void ifft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
 {
     size_t N = in.size();
-    SimpleArray<T1<T2>> in_conj(N);
+    SimpleArray<T1<T2>> in_conj{modmesh::small_vector<size_t>{N}, T1<T2>{0.0, 0.0}};
```
j8xixo12, line 79: Make sure the `in_conj` has the correct initialization value.

cpp/modmesh/transform/fourier.hpp
```diff
@@ -87,6 +91,67 @@ void ifft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
...
+
+    if ((N & (N - 1)) == 0)
+    {
+        fft_radix_2<T1, T2>(in, out);
+    }
+    else
+    {
+        fft_bluestein<T1, T2>(in, out);
+    }
+}
```
j8xixo12, line 153: If the input array length is power of 2 (radix-2) then using radix-2 fft otherwise using Bluestein fft.

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
 #error "Python.h should not be included."
 #endif
 
-template <typename T>
+template <typename T, bool IsPow2>
+struct FFTTestParams
+{
+    using Type = T;
+    static constexpr bool isPow2 = IsPow2;
+};
```
j8xixo12: Typed test parameter template declaration.
The first template argument specifies the floating point type (`double` or `float`), while the second indicates whether the input data length is a power of 2 (radix-2) or not.

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
...
+
+template <typename TestParam>
 class ParsevalTest : public ::testing::Test
 {
+public:
+    using T = typename TestParam::Type;
+    static constexpr bool isPow2 = TestParam::isPow2;
```
j8xixo12: Since the test parameter `isPow2` can be defined during compile time, so declare it as a `static constexpr`.

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
...
+
 protected:
-    const size_t VN = 1024;
+    static constexpr size_t VN = isPow2 ? 1024 : 1000;
```
j8xixo12: Ditto.

gtests/test_nopython_transform.cpp
```diff
@@ -116,51 +135,57 @@ class InverseTest : public ::testing::Test
     }
 };
 
-typedef ::testing::Types<float, double> TestTypes;
+typedef ::testing::Types<
+    FFTTestParams<float, true>,
+    FFTTestParams<float, false>,
+    FFTTestParams<double, true>,
+    FFTTestParams<double, false>>
+    TestTypes;
```
j8xixo12, line 143: Declare different test cases by different floating point types and different input data length.

j8xixo12 (commented):
@yungyuc This PR is ready for the review!

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
...
+struct FFTTestParams
+{
+    using Type = T;
+    static constexpr bool isPow2 = IsPow2;
+};
+
+template <typename TestParam>
 class ParsevalTest : public ::testing::Test
 {
+public:
```
j8xixo12, line 20: When I'm developing the test case, I encountered a weird compile error `C2248 cannot access private typedef declared in class` when I'm using MSVC, but this error does not occur on gcc or clang.

Windows CI failure:
https://github.com/j8xixo12/modmesh/actions/runs/14277469073/job/40022662576#step:9:421

I also found some people discussing similar issues:
1. https://stackoverflow.com/questions/77310307/msvc-bug-template-friend-function-cannot-access-private-members-despite-templat
2. https://stackoverflow.com/questions/47142451/msvc-fails-to-compile-an-explicit-template-instantiation-with-a-private-type-as

It looks like gcc and clang does not apply access checking during template instantiation but MSVC does.
So I declare these class members as public member to resolve the compiler compatibility issue.

cpp/modmesh/transform/fourier.hpp
yungyuc: Good catch

gtests/test_nopython_transform.cpp
yungyuc: Perhaps it is undefined behavior, or a compiler has a bug.  But it is minor and we do not need to pay attention to it.  It's good enough to use this inline annotation here as a note.

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
 #error "Python.h should not be included."
 #endif
 
-template <typename T>
+template <typename T, bool IsPow2>
+struct FFTTestParams
+{
+    using Type = T;
+    static constexpr bool isPow2 = IsPow2;
```
yungyuc: The `static` member should use `snake_case`.

gtests/test_nopython_transform.cpp
yungyuc: Member data should use `snake_case`.

yungyuc (changes requested):
It looks good.  Please fix the naming style for the member data in the testing code.

gtests/test_nopython_transform.cpp
```diff
@@ -7,11 +7,22 @@
 #error "Python.h should not be included."
 #endif
 
-template <typename T>
+template <typename T, bool IsPow2>
+struct FFTTestParams
+{
+    using Type = T;
+    static constexpr bool is_pow_2 = IsPow2;
```
j8xixo12, line 14: Modify `static` member name to `snake_name` style.

j8xixo12 (commented):
@yungyuc This PR is ready for next review.

yungyuc (approved):
Beautiful

PR #494, opened 2025-04-11
Add SVG parser into Python module and adjust camera control

YenPeiChen07:
The PR is main to render a [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG) (scaleable vector graphics). A SVG parser parse a SVG file and store all lines and curves using `SegmentPad` and `CurvePad`.  By clicking `File > Open SVG file`on the menu, user can select any SVG file and render it.

When rendering an example graphics ([tree.svg](https://github.com/YenPeiChen07/modmesh/blob/feature/svg/tests/data/tree.svg)), its size and position is out of view window initially. Therefore, I adjusted some code to automatically adjust the camera position based on a bounding box of an object. It looks OK now, but the revised code is quite messy, which requires more adjustment.

## Example (tree.svg)
The minimum vertex is at `(-452.424, -725.057, 0)` and the maximum vertex  is at `(201.036, 49.9452, 0)`, which is out of view window. After correcting the camera position and view center, the object can fit to view window at the beginning.
Screenshot 2025-04-11 at 6 51 00 PM


tests/data/tree.svg
yungyuc, line 1: Did you draw it yourself?  If not, is the data file allowed to be copied for other use?

modmesh/pilot/_svg.py
```diff
@@ -0,0 +1,454 @@
+# Copyright (c) 2021, Yung-Yu Chen <yyc@solvcon.net>
```
yungyuc: Please update copyright year and author.

modmesh/pilot/_svg.py
```diff
@@ -0,0 +1,454 @@
...
+    def __init__(self, dAttr, fillAttr):
+        """
+        :param closedPaths: list of closed paths in a <path>.
+        """
+        self.dAttr = dAttr
+        self.fillAttr = fillAttr
+        self.cmds = self.dAttr_parser()
+        self.closedPaths = self.cal_vertices()
+
+    def cal_arc2pnts(self, start_pt, end_pt, rx, ry, phi_deg, large_arc, sweep,
```
yungyuc: What does `cal` in the function name mean?  If it stands for calculate, please use `calc` which is the convention in the code base.

modmesh/pilot/_svg.py
```diff
@@ -0,0 +1,454 @@
...
+from .. import core
+from ._gui_common import PilotFeature
+
+__all__ = [  # noqa: F822
+    'SVGFileDialog',
+]
+
+
+class EPath(object):
+    def __init__(self, dAttr, fillAttr):
```
yungyuc: Use `snake_case` for argument names

modmesh/pilot/_svg.py
```diff
@@ -0,0 +1,454 @@
...
+        return commands
+
+    def getClosedPaths(self):
+        return self.closedPaths
+
+    def getCmds(self):
+        return self.cmds
+
+
+class _pathParser(object):
```
yungyuc: Class name should use `_CamelCase`.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,33 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
+    QVector3D bounding_min(box_minX, box_minY, box_minZ);
+    QVector3D bounding_max(box_maxX, box_maxY, box_maxZ);
+    QVector3D center = (bounding_min + bounding_max) * 0.5f;
+    float radius = (bounding_max - bounding_min).length() * 0.5f;
+
+    float fov = 45.0f;
+    float dist = radius / std::tan(qDegreesToRadians(fov) / 2.0f);
+
+    // adjust camera's position
+    Qt3DRender::QCamera * cam = camera();
```
yungyuc: There is already a camera control in https://github.com/solvcon/modmesh/blob/446f2ae7b458b8a5e284d8a4576aa4e4d676ca8c/cpp/modmesh/pilot/RCameraController.hpp#L146 .  It seems strange to create another camera here.  Can you try to use the existing camera control?

yungyuc (changes requested):
- [x] SVG file source and license.
- [x] New file year and author.
- [x] Clarify the meaning of `cal` in function names.  Use `calc` if necessary.
- [x] Use `snake_case` for variable and argument names.  `CamelCase` for class names.
- [x] Try to use existing camera and its control.

Also, please make sure CI is clean before updating the branch.

cpp/modmesh/pilot/R3DWidget.cpp
YenPeiChen07: You are right. I thought about reusing the existing camera control. But it could take some time and I also wanted to show it earlier. I forget to annotate it in the first place.

I am going to try it.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,31 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
             child->deleteLater();
         }
     }
+
     new RVertices(world, m_scene);
     new RLines(world, m_scene);
+
+    auto vertices = std::make_shared<RVertices>(world, m_scene);
```
YenPeiChen07: Use shared pointer to prevent the `error: Potential leak of memory pointed to by 'vertices'`.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,31 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
+
+    QVector3D bounding_min(box_minX, box_minY, box_minZ);
+    QVector3D bounding_max(box_maxX, box_maxY, box_maxZ);
+    QVector3D center = (bounding_min + bounding_max) * 0.5f;
+    float radius = (bounding_max - bounding_min).length() * 0.5f;
+
+    float fov = 45.0f;
+    float dist = radius / std::tan(qDegreesToRadians(fov) / 2.0f);
+
+    cameraController()->setPosition(center + QVector3D(0, 0, dist));
```
YenPeiChen07: Use the existing camera control to update the camera.

cpp/modmesh/pilot/R3DWidget.cpp
yungyuc: Qt objects may not use shared pointer but the Qt object tree: https://doc.qt.io/qt-6.8/objecttrees.html .  In short, you need to set parent properly.

cpp/modmesh/pilot/R3DWidget.cpp
yungyuc: This is good.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,31 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
         }
     }
+
     new RVertices(world, m_scene);
     new RLines(world, m_scene);
+
+    auto vertices = std::make_shared<RVertices>(world, m_scene);
+    auto lines = std::make_shared<RLines>(world, m_scene);
+
+    float box_minX = std::min(vertices->bounding_min().x(), lines->bounding_min().x());
```
yungyuc: Naming convention: `snake_case_x`.

Please add code comments to explain what the calculation does.  It is not crystal clear by simply reading the code.

yungyuc (changes requested):
- [ ] Use Qt object tree to manage Qt objects.  Do not use shared pointer for Qt objects.
- [x] Be careful on variable naming convention.
- [x] Add code comment to explain the bounding box calculation.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,50 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
             child->deleteLater();
         }
     }
-    new RVertices(world, m_scene);
-    new RLines(world, m_scene);
+
+    // Create standalone Qt objects first
+    auto vertices = std::make_unique<RVertices>(world, nullptr);
```
YenPeiChen07: Create a QObject without assigning parent object at first.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,50 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
+    float fov = 45.0f;
+    float half_height = (bounding_max - bounding_min).length() * 0.5f;
+    float dist = half_height / std::tan(qDegreesToRadians(fov) / 2.0f);
+
+    cameraController()->setPosition(box_center + QVector3D(0, 0, dist));
+    cameraController()->setViewCenter(box_center);
+    cameraController()->setFarPlane(dist + half_height * 2);
+
+    // Attach into the scene, being part of existing Qt object tree
+    vertices->setParent(m_scene);
```
YenPeiChen07: Attach the standalone objects into the scene by setting parent. They will become children node of the `m_scene`.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,50 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
+    lines->setParent(m_scene);
+
+    // Release ownership from unique_ptr. Qt will delete them.
+    vertices.release();
```
YenPeiChen07: The unique_ptr objects is released from the responsibility of deleting the objects. Let Qt object tree to manage them.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,50 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
...
+     *
+     * Using the relation: tan(fov / 2) = (half of vertical size) / (distance to center),
+     * we can rearrange and solve for the distance:
+     *
+     *     (distance to center) = (half of vertical size) / tan(fov / 2)
+     *
+     * This gives the minimum distance from the box center along the view direction within
+     * the specified FOV. After that, set the far plane properly to ensurce enclose whole
+     * bounding box.
+     */
```
YenPeiChen07, line 106: Comment for the calculation of camera position.

cpp/modmesh/pilot/R3DWidget.cpp
yungyuc: You are requesting `std::unique_ptr::release()` to delete the object and there is nothing Qt can delete after this line.  It is incorrect.  Do not use `unique_ptr`.

cpp/modmesh/pilot/R3DWidget.cpp
yungyuc: Why don't you let `vertices` and `lines` to set their parents (`m_scene`) here?

cpp/modmesh/pilot/R3DWidget.cpp
yungyuc: This looks good.

yungyuc (changes requested):
- [x] Do not use any smart pointer with Qt objects.  They conflict.

cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -84,8 +84,34 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
             child->deleteLater();
         }
     }
-    new RVertices(world, m_scene);
+
     new RLines(world, m_scene);
+    new RVertices(world, m_scene);
```
YenPeiChen07, line 87: Create a QObject and set parent. Discard using raw pointers or smart pointers to take over QObject's ownership.

cpp/modmesh/pilot/R3DWidget.hpp
```diff
@@ -69,9 +73,26 @@ class RScene
 
     void setFirstPersonCameraController() { setCameraController(new RFirstPersonCameraController(this)); }
 
+    void setBoundingBox(QVector3D buttom_left, QVector3D top_right)
```
YenPeiChen07: Move setter and getter function of bounding box from `RVertices` and `RLines` to `RScene`. 

Sine `RVertices` and `RLines` would set `RScene` as their parent, I decide to calculate bounding box in `RVertices` and `RLines` and then pass the value to their parent.

cpp/modmesh/pilot/RWorld.cpp
```diff
@@ -179,6 +211,12 @@ RLines::RLines(std::shared_ptr<WorldFp64> const & world, Qt3DCore::QNode * paren
             vertices->setCount(npoint);
 
             m_geometry->addAttribute(vertices);
+
+            RScene * parent = qobject_cast<RScene *>(this->parent());
```
YenPeiChen07, line 215: Find a `RScene` parent and set bounding box to it.

yungyuc:
I tried to load [an alien](https://www.svgrepo.com/svg/535115/alien) and this is what I got:

image

It is supposed to be:

image

But I did not observe any crash.  As a first PR it is not bad.  It would be better if a unit test case is added, but it can be done in a follow-up PR.

cpp/modmesh/pilot/R3DWidget.hpp
yungyuc: This is a good idea to add `setBoundingBox()`.  Please make sure there is not a function named like this in the Qt parent class, and make the argument to be `QVector3D const &`, instead of pass by value (`QVector3D`).

cpp/modmesh/pilot/R3DWidget.hpp
```diff
@@ -69,9 +73,26 @@ class RScene
...
+        m_min_pt = QVector3D(bbox_min_x, bbox_min_y, bbox_min_z);
+        m_max_pt = QVector3D(bbox_max_x, bbox_max_y, bbox_max_z);
+    }
+    QVector3D minPoint() const { return m_min_pt; }
+    QVector3D maxPoint() const { return m_max_pt; }
+
 private:
 
     RCameraController * m_controller;
+    QVector3D m_min_pt;
```
yungyuc, line 94: Good addition if the parent class does not have them.

yungyuc (changes requested):
We are almost there.

- [x] Use copy by reference instead of copy by value for larger objects like `QVector3D`.

cpp/modmesh/pilot/R3DWidget.hpp
```diff
@@ -69,9 +73,26 @@ class RScene
 
     void setFirstPersonCameraController() { setCameraController(new RFirstPersonCameraController(this)); }
 
+    void updateBoundingBox(QVector3D const & buttom_left, QVector3D const & top_right)
```
YenPeiChen07, line 76: Used copy by reference for `QVector3D`.

And I found that the same function existed in Qt 5.15 LTS under [`QGeoLocation::setBoundingBox()`](https://doc.qt.io/qt-5/qgeolocation.html#setBoundingBox). However, starting from Qt 6, this was replaced by [`QGeoLocation::setboundingShape()`](https://doc.qt.io/qtforpython-6.5/overviews/qtpositioning-changes-qt6.html#c). 

To prevent conflict with Qt 5.15 LTS, I use `updateBoundingBox()` here.

YenPeiChen07:
> I tried to load [an alien](https://www.svgrepo.com/svg/535115/alien) and this is what I got:
> 
> image
>
> It is supposed to be:
> 
> image
>
> But I did not observe any crash. As a first PR it is not bad. It would be better if a unit test case is added, but it can be done in a follow-up PR.

Thanks for testing. There are still some bugs in the SVG parser, which I will fix it. I will also correct the Y-axis direction of a rendering graphic and add a unit test in a follow-up PR.

cpp/modmesh/pilot/R3DWidget.hpp
yungyuc: This is good.  I meant to type "pass by reference", but got a typo and wrote "copy by reference".  My bad.  It is "pass by reference".

yungyuc (approved):
Good job.  The control of the camera got upgraded.

yungyuc:
> Thanks for testing. There are still some bugs in the SVG parser, which I will fix it. I will also correct the Y-axis direction of a rendering graphic and add a unit test in a follow-up PR.

@YenPeiChen07 Could you please create an issue or issues to track follow-up work?

yungyuc:
Oops.  I forgot to squash.  I will remember next time :)

Issue #506, opened 2025-04-16
Implement `SimpleArray::transpose` for a matrix


yungyuc:
Implement `SimpleArray::transpose` and expose it to Python for testing and profiling.  It should support in-place and out-of-place transpose.

Compare its performance with an optimized BLAS extension like [`?imatcopy` and `?omatcopy` in OpenBLAS](http://www.openmathlib.org/OpenBLAS/docs/extensions/).

If possible, provide optimized implementation.

ThreeMonth03:
Hi @yungyuc , I'm interested in this issue.

yungyuc:
Thank you, @ThreeMonth03 .  Would you think it is possible to have a pull request (may or may not be for everything mentioned in the issue) by Friday?

ThreeMonth03:
> Thank you, [@ThreeMonth03](https://github.com/ThreeMonth03) . Would you think it is possible to have a pull request (may or may not be for everything mentioned in the issue) by Friday?

I'm not sure. I'm not available this Friday.

yungyuc:
> > Thank you, [@ThreeMonth03](https://github.com/ThreeMonth03) . Would you think it is possible to have a pull request (may or may not be for everything mentioned in the issue) by Friday?
> 
> I'm not sure. I'm not available this Friday.

Let's give it a try.  The work should be straight-forward.  If you can't make it by the time, just update the progress and set the goal again.

yungyuc:
Implemented with PR #509

PR #509, opened 2025-04-17
Add the transpose function for SimpleArray

ThreeMonth03:
According to issue #506 , I implement the `void SimpleArray::reverse()` and wrap it to the python.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -654,6 +654,37 @@ public:
         }
     }
 
+    void transpose()
+    {
+        std::reverse(m_shape.begin(), m_shape.end());
+        std::reverse(m_stride.begin(), m_stride.end());
+    }
+
+    void transpose(shape_type const & axis)
+    {
+        if (axis.size() != m_shape.size())
+        {
+            throw std::runtime_error("SimpleArray: axis size mismatch");
+        }
+        shape_type new_shape(m_shape.size(), -1);
+        shape_type new_stride(m_stride.size());
+        for (size_t it = 0; it < m_shape.size(); ++it)
+        {
+            if (axis[it] >= m_shape.size() || axis[it] < 0)
+            {
+                throw std::runtime_error("SimpleArray: axis out of range");
+            }
+            if (new_shape[it] != -1)
+            {
+                throw std::runtime_error("SimpleArray: axis already set");
+            }
+            new_shape[it] = m_shape[axis[it]];
+            new_stride[it] = m_stride[axis[it]];
+        }
+        m_shape = new_shape;
+        m_stride = new_stride;
+    }
+
     template <typename... Args>
     value_type const & operator()(Args... args) const { return *vptr(args...); }
     template <typename... Args>
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -147,6 +147,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
                 { return self.reshape(make_shape(shape)); })
+            .def(
+                "transpose",
+                [](wrapped_type & self, py::object const & axis, bool const & inplace) -> py::object
+                {
+                    if (inplace)
+                    {
+                        if (axis.is_none())
+                        {
+                            self.transpose();
+                        }
+                        else
+                        {
+                            self.transpose(make_shape(axis));
+                        }
+                        return py::none();
+                    }
+
+                    wrapped_type self_copy(self);
+                    if (axis.is_none())
+                    {
+                        self_copy.transpose();
+                    }
+                    else
+                    {
+                        self_copy.transpose(make_shape(axis));
+                    }
+                    return py::cast(std::move(self_copy));
+                },
+                py::arg("axis") = py::none(),
+                py::arg("inplace") = true)
+
             .def_property_readonly("has_ghost", &wrapped_type::has_ghost)
             .def_property("nghost", &wrapped_type::nghost, &wrapped_type::set_nghost)
             .def_property_readonly("nbody", &wrapped_type::nbody)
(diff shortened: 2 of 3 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -654,6 +654,37 @@ class SimpleArray
...
+        for (size_t it = 0; it < m_shape.size(); ++it)
+        {
+            if (axis[it] >= m_shape.size() || axis[it] < 0)
+            {
+                throw std::runtime_error("SimpleArray: axis out of range");
+            }
+            if (new_shape[it] != -1)
+            {
+                throw std::runtime_error("SimpleArray: axis already set");
+            }
```
ThreeMonth03, line 680: Ensure the `small_vector axis;` is the permutation of the sequence from `0` to `n - 1`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -147,6 +147,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    wrapped_type self_copy(self);
+                    if (axis.is_none())
+                    {
+                        self_copy.transpose();
+                    }
+                    else
+                    {
+                        self_copy.transpose(make_shape(axis));
+                    }
+                    return py::cast(std::move(self_copy));
```
ThreeMonth03, line 176: I don't know why the parameter of `py::cast()`  should be rvalue.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -147,6 +147,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
                 { return self.reshape(make_shape(shape)); })
+            .def(
+                "transpose",
+                [](wrapped_type & self, py::object axis, bool inplace) -> py::object
```
ThreeMonth03: In order to deal with the in-place function, the return type of `transpose()` is `py::object`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -147,6 +147,37 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "reshape",
                 [](wrapped_type const & self, py::object const & shape)
                 { return self.reshape(make_shape(shape)); })
+            .def(
+                "transpose",
+                [](wrapped_type & self, py::object const & axis, bool const & inplace) -> py::object
```
ThreeMonth03, line 152: In order to return `None` when `inplace = True`, the return type could be py::object.

tests/test_buffer.py
```diff
@@ -222,6 +222,40 @@ def test_SimpleArray_clone(self):
         self.assertEqual(sarr_ref[3], 3.0)
         self.assertEqual(sarr_clone[3], 2.0)  # should be the original value
 
+    def test_SimpleArray_transpose(self):
+        def check(sarr, ndarr):
```
yungyuc: Please rename the inner function `check` to `check_identical` for readability.

tests/test_buffer.py
```diff
@@ -222,6 +222,40 @@ def test_SimpleArray_clone(self):
...
+
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr)
+        sarr.transpose()
+        check(sarr, ndarrT)
+
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr)
+        sarr = sarr.transpose(inplace=False)
+        check(sarr, ndarrT)
+
+        ndarrT = ndarr.transpose(0, 2, 1, 3)
```
yungyuc: Wow, numpy is so .... confusing.  I grow more fond of `SimpleArray`.

yungyuc (changes requested):
It looks good.  One minor enhancement to the tests and their readability:

- [x] Inner function name

tests/test_buffer.py
```diff
@@ -222,6 +222,40 @@ def test_SimpleArray_clone(self):
...
+                            sarrnum = sarr[idx1, idx2, idx3, idx4]
+                            self.assertEqual(ndnum, sarrnum)
+
+        ndarr = np.arange(2 * 3 * 4 * 5, dtype='float64')
+        ndarr = ndarr.reshape((2, 3, 4, 5))
+        ndarrT = ndarr.transpose()
+
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr)
+        sarr.transpose()
+        check(sarr, ndarrT)
```
yungyuc: Could you please also test `sarr.transpose().copy()`?

ThreeMonth03: The default paremeter `bool inplace` is true, and it would return nothing.

tests/test_buffer.py
```diff
@@ -222,6 +222,42 @@ def test_SimpleArray_clone(self):
...
+                            self.assertEqual(ndnum, sarrnum)
+
+        ndarr = np.arange(2 * 3 * 4 * 5, dtype='float64')
+        ndarr = ndarr.reshape((2, 3, 4, 5))
+        ndarrT = ndarr.transpose()
+
+        sarr = modmesh.SimpleArrayFloat64(array=ndarr)
+        noarr = sarr.transpose()
+        check_identical(sarr, ndarrT)
+        self.assertEqual(noarr, None)
```
ThreeMonth03, line 244: The return type of in-place transpose is None.

ThreeMonth03 (commented):
@yungyuc You could review the pull request when you are available.

tests/test_buffer.py
yungyuc: Oh, right.  It's OK to leave it behaving like this right now.  But I think we will later need to consider how to align the returning behaviors for various array functions that @Tommy-Hsu is planning in https://docs.google.com/document/d/1r6ESbTZPtzmYpnx-sepgp7kXWwwM48XoeVny0dqrg8s .

Also cc @KHLee529 .  We will have a lot of fun code to write.

tests/test_buffer.py
yungyuc: This will be an issue.  In that case, there will not be an equivalent of `ndarr.transpose.copy()`.

Let's give it more discussions and revise in a new PR.

yungyuc (approved):
LGFN

yungyuc:
See the follow-up issue #511

PR #521, opened 2025-04-22
Fix CI error on installing Qt

tychuang1211:
Fix CI error by setting aqtversion to '==3.2.*'

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -141,7 +141,7 @@ jobs:
     - name: install qt
       uses: jurplel/install-qt-action@v4
       with:
-        aqtversion: '==3.1.*'
+        aqtversion: '==3.2.*'
         version: '6.8.1'
         host: 'linux'
         target: 'desktop'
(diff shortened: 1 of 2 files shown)


yungyuc:
`aqtversion: '==3.1.*'` was added with PR #292 .  @terrychan999 , do you recall why it's fixed with version 3.1.*?

yungyuc:
Beautiful.  Thank you @tychuang1211 .

terrychan999:
> `aqtversion: '==3.1.*'` was added with PR #292 . @terrychan999 , do you recall why it's fixed with version 3.1.*?

@yungyuc @tychuang1211 I think it's because Windows used version `'==2.1.*'` before adding `install-qt-action` to Ubuntu and macOS, so I aligned and upgraded it.
Maybe we can try removing this config to keep it updated. 😄

yungyuc:
> @yungyuc @tychuang1211 I think it's because Windows used version `'==2.1.*'` before adding `install-qt-action` to Ubuntu and macOS, so I aligned and upgraded it.

It sounds like a reasonable treatment.

> Maybe we can try removing this config to keep it updated. 😄

Could you please make a PR for the upgrade?

PR #523, opened 2025-04-24
[ci] remove aqtversion of install-qt-action

terrychan999:
In this PR, we removed aqtversion: '==3.2.*' to allow the Qt installer `aqtinstall` to use the default version provided by `jurplel/install-qt-action@v4`.
This keeps us in sync with version updates.


yungyuc:
LGTM.  Thank you, @terrychan999 .

PR #528, opened 2025-04-27
Turn off pip upgrade on macos to its failure on macos 14

yungyuc:
Upgrading pip to 25.1 currently fails on macos-14 runner.  Turning it off temporarily.


yungyuc:
@terrychan999 Please help review if you have time.

PR #529, opened 2025-04-28
Fix wrong assertion in unit test

KHLee529:
Wrong assertion in unit test on `SimpleArray::argsort` function is found by @pchsu-hsupc and he left comment in #456.
This PR fixes this bug.


yungyuc:
LGTM

Issue #519, opened 2025-04-20
Implement `SimpleArray` member functions for searching


yungyuc:
Implement `SimpleArray` member functions for selected searching operations (both C++ and Python wrapper and test in Python):

- [x] [`numpy.argmax()`](https://numpy.org/doc/2.2/reference/generated/numpy.argmax.html#numpy.argmax) as `SimpleArray::argmax()`
- [x] [`numpy.argmin()`](https://numpy.org/doc/2.2/reference/generated/numpy.argmin.html#numpy.argmin) as `SimpleArray::argmin()`
- [x] [`numpy.argwhere()`](https://numpy.org/doc/2.2/reference/generated/numpy.argwhere.html#numpy.argwhere) as `SimpleArray::argwhere()`
- [ ] [`numpy.where()`](https://numpy.org/doc/2.2/reference/generated/numpy.where.html#numpy.where) as `SimpleArray::where()`

# Plan made on 2025/4/24

## Stage 1

- [x] Make Python unittest for `SimpleArray::argmax()` and `SimpleArray::argmin()`
- [x] Implement `SimpleArray::argmax()` and `SimpleArray::argmin()` w/o args `axis` in `SimpleArrayMixinCalculators`

## Stage 2

- [ ] Make Python unittest for `SimpleArray::argwhere()` and `SimpleArray::where()`
  - `argwhere` in PR #977
  - `where` in PR #1055
- [ ] Using new mixin class ~~`SimpleArrayMixinConditional`~~ `SimpleArrayMixinSearch` to collect methods `SimpleArray::argwhere()` and `SimpleArray::where()` 

## Stage 3

- [ ] Add `axis` args back to method `SimpleArray::argmax()` and `SimpleArray::argmin()`

yungyuc:
@pchsu0214 would you be interested in working on this issue?  For implementing each function, use at least one PR.

pchsu-hsupc:
For sure,I am glad to take these.

yungyuc:
Thank you, @pchsu0214 .  When do you think the first PR can be made?

pchsu-hsupc:
Before Friday, I hope.

yungyuc:
> Before Friday, I hope.

If you are not busy with other things, perhaps try to create a PR by Wednesday.  It may be incomplete and for discussion only.  You can reference the `sort()` function implemented with PRs #456 #486 as references.

pchsu-hsupc:
Sure, I see your point.

[...]

Issue #532, opened 2025-04-30
Improve SIMD performance when the array get longer


KHLee529:
From the profiling result of PR #489, the SIMD function `SimpleArray::take_along_axis_simd` get slower when the size of array grows. This issue tracks the improvement of this function.

# Profiling results

The profiling result of `take_along_axis` is listed as following.

## N = 128 type: uint8

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 5.063E-03       | 1.000           | 3.606           |
| sa         | 1.404E-03       | 0.277           | 1.000           |
| simd       | 9.710E-04       | 0.192           | 0.692           |

## N = 1K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 3.921E-03       | 1.000           | 2.168           |
| sa         | 1.808E-03       | 0.461           | 1.000           |
| simd       | 1.475E-03       | 0.376           | 0.816           |

## N = 8K type: uint16

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.266E-02       | 1.000           | 1.919           |
| sa         | 6.596E-03       | 0.521           | 1.000           |
| simd       | 5.571E-03       | 0.440           | 0.845           |

## N = 64K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 9.536E-02       | 1.000           | 1.673           |
| sa         | 5.701E-02       | 0.598           | 1.000           |
| simd       | 6.755E-02       | 0.708           | 1.185           |

## N = 512K type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 8.498E-01       | 1.000           | 1.234           |
| sa         | 6.888E-01       | 0.811           | 1.000           |
| simd       | 7.662E-01       | 0.902           | 1.112           |

## N = 4M type: uint32

| func       | per call (ms)   | cmp to np       | cmp to sa       |
| ---------- | --------------- | --------------- | --------------- |
| np         | 1.450E+01       | 1.000           | 1.980           |
| sa         | 7.323E+00       | 0.505           | 1.000           |
[...]

PR #535, opened 2025-05-05
Implement `SimpleArray` member functions for searching part 2

pchsu-hsupc:
### Completed Tasks
- [x] Make Python unittest for `SimpleArray::argwhere()` and `SimpleArray::where()`

- [x] Using mixin class `SimpleArrayMixinConditional` to collect methods `SimpleArray::argwhere()` and `SimpleArray::where()`

### Abstraction
As mentioned in #519 , this PR implements `argwhere()` and `where()` methods without overloading operators.

`numpy.argwhere(a)`: Find the indices of array elements that are non-zero by default, also could add condition as well.

`numpy.where(condition, [x, y, ]/)`: Return elements chosen from x or y depending on condition.

### Implement Method
Thanks to the `std::function` template, I could use lambdas as argument—letting the API mirror NumPy’s style in Python.

```
ret_np = np.argwhere(narr == 5)
ret_sa = sarr.argwhere(lambda x: x == 5)
```

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -292,6 +292,11 @@ public:
         }
         return max_index;
     }
+
+    SimpleArray<uint64_t> argwhere() const;
+
+    SimpleArray<uint64_t> argwhere(std::function<bool(value_type const &)> const & condition) const;
+    A where(std::function<bool(value_type const &)> const & condition, A const & other) const;
 }; /* end class SimpleArrayMixinSearch */
 
 } /* end namespace detail */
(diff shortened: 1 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1032,6 +1037,69 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis_simd(SimpleArray<I> const
...
+{
+    auto default_condition = [](value_type const & x)
+    {
+        return x != value_type();
+    };
+    return this->argwhere(default_condition);
+}
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere(std::function<bool(value_type const &)> const & condition) const
```
pchsu-hsupc, line 1051: - Use an `std::vector` to store the indices of elements that satisfy the condition.
- Creating the output `coordinates` and filling the dimensions's information.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1032,6 +1037,69 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis_simd(SimpleArray<I> const
...
+            *coord = remaining_index / product_of_dims[i];
+            remaining_index = remaining_index % product_of_dims[i];
+            ++coord;
+        }
+    }
+    return coordinates;
+}
+
+template <typename A, typename T>
+A detail::SimpleArrayMixinSearch<A, T>::where(std::function<bool(value_type const &)> const & condition, A const & other) const
```
pchsu-hsupc, line 1084: - Should output the same shape as input.
- If the condition is met, fill with the original value; otherwise, use the alternative.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -313,9 +313,23 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                {
+                    if (!condition)
+                    {
+                        return py::cast(self.argwhere());
+                    }
+                    return py::cast(self.argwhere(condition));
+                },
+                py::arg("condition") = py::none())
+            .def(
+                "where",
```
pchsu-hsupc, line 328: [Numpy's where()](https://numpy.org/doc/stable/reference/generated/numpy.where.html#numpy.where) says that when only condition is provided, `nonzero()` should be preferred. So I didn't give `SimpleArray::where` default operation.

pchsu-hsupc:
Thank you, @ThreeMonth03 and @KHLee529, for your valuable insights.
I would appreciate any additional feedback you might have.
@yungyuc, could you please take a look at this PR when you have a moment? Thanks!

yungyuc (changes requested):
Could you please profile the runtime performance of the searching functions like what issue #532 does?  If the runtime is much worse than numpy, we may not use the implementation in this PR and need to find another approach.

pchsu-hsupc:
Sure, I will add it in.

pchsu-hsupc:
# Profiling 
Use the built-in tools `modmesh.call_profiler` and `modmesh.time_registry` to profile the default behavior of `SimpleArray::argwhere()` from Python. This function is equivalent to `SimpleArray::argwhere(lambda x: x != 0)`.

Numpy is abbreviated to `np` ,SimpleArray is abbreviated to `sa`.
And Nums is the abbreviation of number of elements.
Speedup achieved by `SimpleArray` is measured relative to `NumPy`.

| | np.argwhere() per call | sa.argwhere() per call | Speedup |
| :---:   |     :---:      |          :---: |      :---: |
| Nums = 128, Data type = uint8   | 5.489 ms | 1.849  ms | 2.97x |
| Nums = 1K, Data type = uint16   |   3.150 ms |  3.795 ms |  0.83x |
| Nums = 8K, Data type = uint16 |  11.546 ms | 24.759 ms | 0.47x |
| Nums = 64K, Data type = uint32 | 80.181 ms | 198.060 ms | 0.40x |
| Nums = 512K, Data type = uint32 | 980.167 ms | 2165.842 ms | 0.45x |
| Nums = 4M, Data type = uint64 | 9604.355 ms | 17401.239 ms | 0.55x |
| Nums = 32M, Data type = uint64 | 102880.127 ms | 160952.633 ms | 0.64x |

With iterations = 50, the test cases were generated using `np.random.uniform()`.

yungyuc:
Please include your profile script in the PR.  @KHLee529 could you please help @pchsu-hsupc ?

The runtime is much less slower than I thought, but still too slow to be useful.  We may not implement the interface by using a Python callable and/or lambda.

KHLee529:
@pchsu-hsupc I think it 's worthy to first clarify whether it's the Python callable value that makes runtime slow or the C++ `std::function`. Maybe try to make the default argument `condition` a free function written in C++. If it is the Python callable object that slows down the whole process, we should implement this function in another way.

yungyuc:
@pchsu-hsupc In the discussions in the meetup, we concluded that it should use Boolean array to select elements in the way that numpy employs.  Please update your progress.

tigercosmos:
Close due to latest master already has the methods.

PR #531, opened 2025-04-29
Port the linear scalar wave solver to `modmesh.onedim` and make a GUI

tychuang1211:
This is a work-in-progress PR for #463.

I’ve started porting the linear scalar wave solver to `modmesh.onedim`, using `modmesh.onedim.euler1d` as a reference and making GUI based on `modmesh.pilot._burgers1d`. The layout and behavior seem to work correctly (at least visually), but I’m not entirely confident about the underlying logic or if this is the right direction.

I'd appreciate any feedback on whether this approach makes sense and what should be the next steps.

PR #536, opened 2025-05-06
Restore original copyright and license texts

tychuang1211:
As mentioned [here](https://github.com/solvcon/modmesh/pull/531#discussion_r2073250579).


yungyuc (approved):
Thanks.

PR #537, opened 2025-05-11
Updates GUI with Add/Delete region buttons

Gene0315:
This PR updates GUI with Add/Delete region buttons.
截圖 2025-05-11 上午11 22 30

Before this revision, the number of region is fixed by code and we can only change the value in GUI.
image

But now, by pressing the `Add Region` or `Delete Region` button, the number of region in GUI is adjusted.
截圖 2025-05-11 上午11 28 45

### DIFF
--- a/modmesh/pilot/_base_app.py
+++ b/modmesh/pilot/_base_app.py
@@ -210,6 +210,7 @@ class GUIConfig(object):
         :return None
         """
         self._tbl_content[row][col] = value
+        return True
 
     def columnHeader(self, col):
         """
@@ -322,6 +323,7 @@ class OneDimBaseApp(PilotFeature):
         - `plot_num` (bool): Flag indicating whether to plot numerical data.
         - `use_grid_layout` (bool): Flag indicating whether to use a grid
           layout.
+        - `adjust_region` (bool): Flag indicating whether to adjust the region.
 
     Methods:
         - :meth:`populate_menu()`: Set menu item for GUI.
@@ -361,6 +363,7 @@ class OneDimBaseApp(PilotFeature):
     plot_ana: bool = False
     plot_num: bool = False
     use_grid_layout: bool = False
+    adjust_region: bool = False
 
     def populate_menu(self):
         """
@@ -690,6 +693,8 @@ class ConfigWindow(QWidget):
           window is opened.
         - :meth:`on_close()`: Callback function when plot configure modal
           windows is closed.
+        - :meth:`add_region()`: Add data for a new region.
+        - :meth:`delete_region()`: Delete data for a region.
         - :meth:`init_ui()`: Define the GUI layout of the window.
     """
 
@@ -720,6 +725,28 @@ class ConfigWindow(QWidget):
         if self.plot_config_dialog:
             self.plot_config_dialog.close()
 
+    def add_region(self):
+        """
+        Add data for a new region to the solver configuration table.
+        """
+        table_model = ConfigTableModel(self.solver_config)
+        list_add = sorted(self.app.get_region_add(),
+                          key=lambda x: x[0],
+                          reverse=True)
+        for pos, data in list_add:
+            table_model.insertRow(data, pos)
+        self.table.setModel(table_model)
+
+    def delete_region(self):
+        """
+        Delete data for a region from the solver configuration table.
+        """
+        table_model = ConfigTableModel(self.solver_config)
+        list_del = sorted(self.app.get_region_delete(), reverse=True)
+        for pos in list_del:
+            table_model.deleteRow(pos)
+        self.table.setModel(table_model)
+
     def init_ui(self):
         """
         Define the GUI layout of the window for this class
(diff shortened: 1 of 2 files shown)


modmesh/pilot/_base_app.py
```diff
@@ -210,6 +210,7 @@ def setData(self, row, col, value):
         :return None
         """
         self._tbl_content[row][col] = value
+        return True
```
yungyuc, line 213: Why do you need this `return True`?  I do not see other functions returning and it stands out.  Please clarify the reason.

modmesh/pilot/_burgers1d.py
```diff
@@ -44,16 +44,16 @@ def __init__(self):
...
         self.velocity_field = None
 
-    def build_constant(self, x1, x2, x3, velocity12, velocity23):
+    def build_constant(self, x, velocity):
         """
         Set solver initial condition.
 
-        :param xn: the n-th point of x axis.
-        :param velocitymn: the velocity between point m and n.
+        :param x: the points of x axis.
```
yungyuc: If it is a list or a container type, please document it in the docstring.

yungyuc (changes requested):
In addition to the point below, could you please help evaluate how to add unit tests for the 1D GUI?  (To evaluate does not mean to do it right away.)

- [x] Clarify the reason to `return True` from `setData()`.

@tychuang1211 I'd like to solicit your comments as well.

modmesh/pilot/_base_app.py
```diff
@@ -322,6 +323,7 @@ class OneDimBaseApp(PilotFeature):
         - `plot_num` (bool): Flag indicating whether to plot numerical data.
         - `use_grid_layout` (bool): Flag indicating whether to use a grid
           layout.
+        - `adjust_region` (bool): Flag indicating whether to adjust the region.
```
tychuang1211, line 326: The current description sounds like the region will be automatically adjusted. However, based on the implementation, this flag seems to control whether the app allows the user to modify the number of regions (e.g., by showing Add/Delete buttons). Maybe we can consider rephrasing it to something like: "Flag indicating whether the user can modify the number of regions"

modmesh/pilot/_base_app.py
Gene0315: I saw the error message `modmesh/pilot/_gui.py:56: RuntimeWarning: Invalid return value in function 'QAbstractTableModel.setData', expected bool, got NoneType.` when I pressed the add/delete region button.

Thus, I added `return True` in function `setData` of class `GUIConfig` and `ConfigTableModel`.
Is it improper to return True?

modmesh/pilot/_burgers1d.py
Gene0315: ok, I will add it.

yungyuc:
@Gene0315 Could you please help feed back when you have time.

modmesh/pilot/_base_app.py
yungyuc: I see.  `QAbstractTableModel` requires `setData()` to return true.  Then it is OK.  Thank you for clarifying.

modmesh/pilot/_burgers1d.py
yungyuc: While we are not in a hurry, this is an easy fix.  @Gene0315 could you please help make an estimate when you will update for the next review?

Gene0315: I think I will update the next review this weekend.
The next step, I want to fix the analytical solution. After shock and expansion wave meet, the velocity looks abnormal like this:
截圖 2025-05-21 晚上8 41 00

Gene0315: @yungyuc I added the type of variables in docstring. Can you please merge this PR?

modmesh/pilot/_burgers1d.py
```diff
@@ -44,16 +44,16 @@ def __init__(self):
...
         self.velocity_field = None
 
-    def build_constant(self, x1, x2, x3, velocity12, velocity23):
+    def build_constant(self, x, velocity):
         """
         Set solver initial condition.
 
-        :param xn: the n-th point of x axis.
-        :param velocitymn: the velocity between point m and n.
+        :param x (type: list): the points of x axis.
```
yungyuc, line 51: This is better than the previous way, but still has room to improve.  Please do it in the future refactoring.

yungyuc (approved):
LGTM

modmesh/pilot/_burgers1d.py
yungyuc: > The next step, I want to fix the analytical solution. After shock and expansion wave meet, the velocity looks abnormal like this: 截圖 2025-05-21 晚上8 41 00

Please create an issue stating what you want to do.

Gene0315: Can I ask what issue stating is?

yungyuc: > Can I ask what issue stating is?

Create an issue to state what you want to do.

PR #538, opened 2025-05-18
Fix the initializer from numpy to SimpleArray

### LINKED ISSUE #527: Bug: `SimpleArrayMixinCalculators` needs to handle non-c-contiguous memory buffer
If the array is not c-contiguous, the return values of the functions in `class SimpleArrayMixinCalculators;` are wrong.

In other words, it is necessary to fix the function `SimpleArrayMixinCalculators::min()`, `SimpleArrayMixinCalculators::max()`, `SimpleArrayMixinCalculators::sum()`, `SimpleArrayMixinCalculators::abs()`.

For example:

Unittest:
```python3
    def test_sum2(self):

        nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
        nparr = nparr[:,:,::2] ## Slicing
        sarr = modmesh.SimpleArrayFloat64(array=nparr) 
        npsum = nparr.sum()
        ssum = sarr.sum()
        self.assertEqual(npsum, ssum)
```

Source code of `sum()`:
```c++
    value_type sum() const
    {
        value_type initial;
        if constexpr (is_complex_v<value_type>)
        {
            initial = value_type();
        }
        else
        {
            initial = 0;
        }

        auto athis = static_cast<A const *>(this);
        if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
        {
            for (size_t i = 0; i < athis->size(); ++i)
            {
                initial += athis->data(i);
            }
        }
        else
        {
            for (size_t i = 0; i < athis->size(); ++i)
            {
                initial |= athis->data(i);
            }
        }
        return initial;
    }
```

Console log:
```console

self = <test_buffer.SimpleArrayCalculatorsTC testMethod=test_sum2>

    def test_sum2(self):
    
        nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
        nparr = nparr[:,:,::2]
        sarr = modmesh.SimpleArrayFloat64(array=nparr)
        npsum = nparr.sum()
        ssum = sarr.sum()
>       self.assertEqual(npsum, ssum)
E       AssertionError: np.float64(132.0) != 66.0

tests/test_buffer.py:962: AssertionError
================================================================================================================================================================================== short test summary info ==================================================================================================================================================================================
FAILED tests/test_buffer.py::SimpleArrayCalculatorsTC::test_sum2 - AssertionError: np.float64(132.0) != 66.0
```

ThreeMonth03:
In this pull request, I fix the initializer in SimpleArray. 
This modification will help us solve the issue #527.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -152,11 +152,7 @@ public:
     value_type median() const
     {
         auto athis = static_cast<A const *>(this);
-        size_t n = 1;
-        for (size_t i = 0; i < athis->ndim(); ++i)
-        {
-            n *= athis->shape(i);
-        }
+        const size_t n = athis->size();
         small_vector<T> acopy(n);
         auto sidx = athis->first_sidx();
         size_t i = 0;
@@ -205,11 +201,7 @@ public:
     real_type var(size_t ddof) const
     {
         auto athis = static_cast<A const *>(this);
-        size_t n = 1;
-        for (size_t i = 0; i < athis->ndim(); ++i)
-        {
-            n *= athis->shape(i);
-        }
+        const size_t n = athis->size();
         if (n <= ddof)
         {
             throw std::runtime_error("SimpleArray::var(): ddof must be less than the number of elements");
@@ -723,17 +715,6 @@ public:
     {
         if (buffer)
         {
-            const size_t nbytes = ITEMSIZE *
-                                  std::accumulate(shape.begin(),
-                                                  shape.end(),
-                                                  static_cast<size_t>(1),
-                                                  std::multiplies<size_t>());
-            if (nbytes != buffer->nbytes())
-            {
-                throw std::runtime_error(Formatter() << "SimpleArray: shape byte count " << nbytes
-                                                     << " differs from buffer " << buffer->nbytes());
-            }
-
             m_shape = shape;
             m_stride = stride;
         }
@@ -834,8 +815,22 @@ public:
 
     explicit operator bool() const noexcept { return bool(m_buffer) && bool(*m_buffer); }
 
-    size_t nbytes() const noexcept { return m_buffer ? m_buffer->nbytes() : 0; }
-    size_t size() const noexcept { return nbytes() / ITEMSIZE; }
+    size_t nbytes() const noexcept { return size() * ITEMSIZE; }
+    /**
+     * In the following numpy documentaions,
+     * `size()` are calculated based on the number of the elements.
+     * https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.size.html
+     * SimpleArray matchs the behavior of numpy.ndarray.
+     */
+    size_t size() const noexcept
+    {
+        size_t size = 1;
+        for (size_t it = 0; it < m_shape.size(); ++it)
+        {
+            size *= m_shape[it];
+        }
+        return size;
+    }
 
     using iterator = T *;
     using const_iterator = T const *;
@@ -1090,11 +1085,11 @@ private:
         {
             throw std::out_of_range(Formatter() << "SimpleArray: index " << it << " < -nghost: " << -static_cast<ssize_t>(m_nghost));
         }
-        if (it >= static_cast<ssize_t>(size() - m_nghost))
+        if (it >= static_cast<ssize_t>((buffer().nbytes() / ITEMSIZE) - m_nghost))
         {
             throw std::out_of_range(
-                Formatter() << "SimpleArray: index " << it << " >= " << size() - m_nghost
-                            << " (size: " << size() << " - nghost: " << m_nghost << ")");
+                Formatter() << "SimpleArray: index " << it << " >= " << (buffer().nbytes() / ITEMSIZE) - m_nghost
+                            << " (buffer size: " << (buffer().nbytes() / ITEMSIZE) << " - nghost: " << m_nghost << ")");
         }
     }
 
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -78,6 +78,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                         modmesh::detail::shape_type shape;
                         modmesh::detail::shape_type stride;
                         constexpr size_t itemsize = wrapped_type::itemsize();
+                        constexpr size_t span = 0;
                         for (ssize_t i = 0; i < arr_in.ndim(); ++i)
                         {
                             shape.push_back(arr_in.shape(i));
@@ -87,10 +88,49 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                         const bool is_c_contiguous = (arr_in.flags() & py::array::c_style) == py::array::c_style;
                         const bool is_f_contiguous = (arr_in.flags() & py::array::f_style) == py::array::f_style;
 
-                        std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
-                            arr_in.nbytes(),
-                            arr_in.mutable_data(),
-                            std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
+                        py::array owner = arr_in;
+                        /*
+                         * In the following document, it introduces the base object in ndarray.
+                         * https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.base.html
+                         * The `array.base` is base object if memory is from some other object.
+                         * If object owns its memory, base is None.
+                         */
+                        while (true)
+                        {
+                            const py::object b = owner.attr("base");
+                            if (b.is_none() || !py::isinstance<py::array>(b))
+                            {
+                                break;
+                            }
+                            auto next = b.cast<py::array>();
+                            /*
+                             * Prevent the infinite loop.
+                             * For example, the following code will create a loop:
+                             * nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
+                             * nparr = nparr[::2, ::2, ::2]
+                             */
+                            if (next.ptr() == owner.ptr())
+                            {
+                                break;
+                            }
+                            owner = next;
+                        }
+
+                        char * base_ptr = static_cast<char *>(owner.mutable_data());
+                        char * view_ptr = static_cast<char *>(arr_in.mutable_data());
+                        const ptrdiff_t offset_bytes = view_ptr - base_ptr;
+                        if (offset_bytes < 0)
+                        {
+                            throw std::runtime_error("Unexpected negative offset!");
+                        }
+                        const size_t true_owner_nbytes = owner.nbytes();
+                        const size_t view_nbytes = true_owner_nbytes - static_cast<size_t>(offset_bytes);
+                        auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(owner);
+                        const std::shared_ptr<ConcreteBuffer> buffer =
+                            ConcreteBuffer::construct(
+                                view_nbytes,
+                                view_ptr,
+                                std::move(remover));
                         return wrapped_type(shape, stride, buffer, is_c_contiguous, is_f_contiguous);
                     }),
                 py::arg("array"))
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -320,7 +320,7 @@ class SimpleArrayBasicTC(unittest.TestCase):
 
         with self.assertRaisesRegex(
                 IndexError,
-                r"SimpleArray: index 14 >= 14 \(size: 24 - nghost: 10\)"
+                r"SimpleArray: index 14 >= 14 \(buffer size: 24 - nghost: 10\)"
         ):
             sarr[14]
 
@@ -332,7 +332,7 @@ class SimpleArrayBasicTC(unittest.TestCase):
 
         with self.assertRaisesRegex(
                 IndexError,
-                r"SimpleArray: index 14 >= 14 \(size: 24 - nghost: 10\)"
+                r"SimpleArray: index 14 >= 14 \(buffer size: 24 - nghost: 10\)"
         ):
             sarr[14] = 1
 
@@ -489,6 +489,8 @@ class SimpleArrayBasicTC(unittest.TestCase):
         np_sarr = np.empty(shape, dtype='float64')
         py_sarr = modmesh.SimpleArrayFloat64(array=np_sarr)
         self.assertTupleEqual(shape, py_sarr.shape)
+        self.assertEqual(np_sarr.nbytes, py_sarr.nbytes)
+        self.assertEqual(np_sarr.size, py_sarr.size)
 
         shape = (5, 5, 5, 5)
         np_sarr = np.empty(shape, dtype='float64')
@@ -500,6 +502,8 @@ class SimpleArrayBasicTC(unittest.TestCase):
 
         py_sarr = modmesh.SimpleArrayFloat64(array=np_sarr)
         self.assertTupleEqual(shape, py_sarr.shape)
+        self.assertEqual(np_sarr.nbytes, py_sarr.nbytes)
+        self.assertEqual(np_sarr.size, py_sarr.size)
 
     def test_SimpleArray_from_ndarray_content(self):
 
@@ -524,6 +528,8 @@ class SimpleArrayBasicTC(unittest.TestCase):
             for j in range(4):
                 for k in range(6):
                     self.assertEqual(parr[i, j, k], sarr[i, j, k])
+        self.assertEqual(parr.nbytes, sarr.nbytes)
+        self.assertEqual(parr.size, sarr.size)
 
     def test_SimpleArray_from_ndarray_transpose(self):
         ndarr = np.arange(350, dtype='float64').reshape((5, 7, 10))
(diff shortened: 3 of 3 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,18 @@ class SimpleArray
...
+    size_t nbytes() const noexcept { return size() * ITEMSIZE; }
+    size_t size() const noexcept
+    {
+        size_t size = 1;
+        for (size_t it = 0; it < m_shape.size(); ++it)
+        {
+            size *= m_shape[it];
+        }
+        return size;
+    }
```
ThreeMonth03: Differentiate the `buffer_size()`, `size()`, `buffer_nbytes()` and `nbytes()`.
In numpy, the function `size()` and `nbytes()` is calculated by shape.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1090,11 +1081,11 @@ class SimpleArray
...
             throw std::out_of_range(Formatter() << "SimpleArray: index " << it << " < -nghost: " << -static_cast<ssize_t>(m_nghost));
         }
-        if (it >= static_cast<ssize_t>(size() - m_nghost))
+        if (it >= static_cast<ssize_t>(buffer_size() - m_nghost))
         {
             throw std::out_of_range(
-                Formatter() << "SimpleArray: index " << it << " >= " << size() - m_nghost
-                            << " (size: " << size() << " - nghost: " << m_nghost << ")");
+                Formatter() << "SimpleArray: index " << it << " >= " << buffer_size() - m_nghost
+                            << " (buffer size: " << buffer_size() << " - nghost: " << m_nghost << ")");
```
ThreeMonth03: This condition is calculated by the `buffer_size()`, so I fix it.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,10 +88,49 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                             * For example, the following code will create a loop:
+                             * nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
+                             * nparr = nparr[::2, ::2, ::2]
+                             */
+                            if (next.ptr() == owner.ptr())
+                            {
+                                break;
+                            }
+                            owner = next;
+                        }
```
ThreeMonth03, line 117: The while loop will take the buffer and metadata of the root numpy array.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,10 +88,49 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                        if (offset_bytes < 0)
+                        {
+                            throw std::runtime_error("Unexpected negative offset!");
+                        }
+                        const size_t true_owner_nbytes = owner.nbytes();
+                        const size_t view_nbytes = true_owner_nbytes - static_cast<size_t>(offset_bytes);
+                        auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(owner);
+                        const std::shared_ptr<ConcreteBuffer> buffer =
+                            ConcreteBuffer::construct(
+                                view_nbytes,
```
ThreeMonth03, line 131: The buffer size should consider the offset.

tests/test_buffer.py
```diff
@@ -500,6 +502,8 @@ def test_SimpleArray_from_ndarray(self):
 
         py_sarr = modmesh.SimpleArrayFloat64(array=np_sarr)
         self.assertTupleEqual(shape, py_sarr.shape)
+        self.assertEqual(np_sarr.nbytes, py_sarr.nbytes)
+        self.assertEqual(np_sarr.size, py_sarr.size)
```
ThreeMonth03, line 506: Ensure that the information of `nbytes` and `size` are right when the SimpleArray is not sliced.

tests/test_buffer.py
```diff
@@ -524,6 +528,8 @@ def test_SimpleArray_from_ndarray_slice(self):
             for j in range(4):
                 for k in range(6):
                     self.assertEqual(parr[i, j, k], sarr[i, j, k])
+        self.assertEqual(parr.nbytes, sarr.nbytes)
+        self.assertEqual(parr.size, sarr.size)
```
ThreeMonth03, line 532: Ensure that the information of nbytes and size are right when the SimpleArray is sliced.

tests/test_buffer.py
```diff
@@ -1061,6 +1074,13 @@ def test_average(self):
         savg = sarr.average()
         self.assertEqual(npavg, savg)
 
+        nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
+        nparr = nparr[::2, ::2, ::2]
+        sarr = modmesh.SimpleArrayFloat64(array=nparr)
+        npavg = np.average(nparr)
+        savg = sarr.average()
+        self.assertEqual(npavg, savg)
```
ThreeMonth03, line 1082: Add the unittest of the statistic operation to ensure that the sliced array works well.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -152,11 +152,7 @@ class SimpleArrayMixinCalculators
     value_type median() const
     {
         auto athis = static_cast<A const *>(this);
-        size_t n = 1;
-        for (size_t i = 0; i < athis->ndim(); ++i)
-        {
-            n *= athis->shape(i);
-        }
+        size_t n = athis->size();
```
ThreeMonth03: Use member function `size()`.

ThreeMonth03 (commented):
@yungyuc You can review this pull request if you are available. Thank you.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Good catch.  Please also make it `const`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,18 @@ class SimpleArray
 
     explicit operator bool() const noexcept { return bool(m_buffer) && bool(*m_buffer); }
 
-    size_t nbytes() const noexcept { return m_buffer ? m_buffer->nbytes() : 0; }
-    size_t size() const noexcept { return nbytes() / ITEMSIZE; }
+    size_t buffer_nbytes() const noexcept { return m_buffer ? m_buffer->nbytes() : 0; }
```
yungyuc: Do not implement `buffer_nbytes()` and use `buffer().nbytes()` instead.  Note that `buffer()` is a public member function.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,18 @@ class SimpleArray
...
+    size_t buffer_size() const noexcept { return buffer_nbytes() / ITEMSIZE; }
```
yungyuc: Do not implement `buffer_size()`.  Use `buffer().size()/ITEMSIZE` instead.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: The condition should continue to use the buffer size, but use `buffer().size()` instead of the unnecessary helper function.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Please add code comment here to say this is to match numpy implementation.

I like the previous `SimpleArray` implementation more, but matching numpy behavior here will make everyone's life easier.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,10 +88,49 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                         const bool is_c_contiguous = (arr_in.flags() & py::array::c_style) == py::array::c_style;
                         const bool is_f_contiguous = (arr_in.flags() & py::array::f_style) == py::array::f_style;
 
-                        std::shared_ptr<ConcreteBuffer> const buffer = ConcreteBuffer::construct(
-                            arr_in.nbytes(),
-                            arr_in.mutable_data(),
-                            std::make_unique<ConcreteBufferNdarrayRemover>(arr_in));
+                        py::array owner = arr_in;
+                        /*
+                         * In the following document, it introduces the base object in ndarray.
```
yungyuc, line 93: Good note.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -87,10 +88,49 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                            }
+                            owner = next;
+                        }
+
+                        char * base_ptr = static_cast<char *>(owner.mutable_data());
+                        char * view_ptr = static_cast<char *>(arr_in.mutable_data());
+                        const ptrdiff_t offset_bytes = view_ptr - base_ptr;
+                        if (offset_bytes < 0)
+                        {
+                            throw std::runtime_error("Unexpected negative offset!");
```
yungyuc, line 124: Is it possible to add a unit test checking for this error message?

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: When there is time, I'd like to learn your thoughts on the design of remover.

tests/test_buffer.py
```diff
@@ -320,7 +320,7 @@ def test_SimpleArray_ghost_1d(self):
 
         with self.assertRaisesRegex(
                 IndexError,
-                r"SimpleArray: index 14 >= 14 \(size: 24 - nghost: 10\)"
+                r"SimpleArray: index 14 >= 14 \(buffer size: 24 - nghost: 10\)"
```
yungyuc, line 323: Good change of the message.

tests/test_buffer.py
yungyuc: Good check.

tests/test_buffer.py
yungyuc: Beautiful.

yungyuc (changes requested):
- [x] Add constness: `size_t n = athis->size();` to `size_t const n = athis->size();`  Note that there are more than 1 occurrence.
- [x] Add code comment here for matching the `size()` and `nbytes()` behavior with numpy.

Don't add the following unnecessary `SimpleArray` member functions.  They are duplicate code confusing people without adding value.

- [x] Do not implement `buffer_nbytes()` and use `buffer().nbytes()` instead.  Note that `buffer()` is a public member function.
- [x] Do not implement `buffer_size()`.  Use `buffer().size()/ITEMSIZE` instead.

Additional item:

- [ ] Add an issue to track profiling the statistical functions and comparing with numpy.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
ThreeMonth03: Here is the suggestion by gpt.
https://chatgpt.com/share/682c87e3-4d40-800a-807d-84e761427b3d
It seems that we should implement the data structure like np.array.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -205,11 +201,7 @@ class SimpleArrayMixinCalculators
     real_type var(size_t ddof) const
     {
         auto athis = static_cast<A const *>(this);
-        size_t n = 1;
-        for (size_t i = 0; i < athis->ndim(); ++i)
-        {
-            n *= athis->shape(i);
-        }
+        const size_t n = athis->size();
```
ThreeMonth03, line 204: Add `const`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,23 @@ class SimpleArray
...
 
-    size_t nbytes() const noexcept { return m_buffer ? m_buffer->nbytes() : 0; }
-    size_t size() const noexcept { return nbytes() / ITEMSIZE; }
+    /*
+     * In the following numpy documentaions,
+     * `nbytes()` and `size()` are calculated based on the number of the elements.
+     * 1. https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.size.html
+     * 2. https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.nbytes.html
+     * SimpleArray matchs the behavior of numpy.ndarray.
+     */
```
ThreeMonth03: Add comments.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,23 @@ class SimpleArray
...
+    size_t nbytes() const noexcept { return size() * ITEMSIZE; }
+    size_t size() const noexcept
+    {
+        size_t size = 1;
+        for (size_t it = 0; it < m_shape.size(); ++it)
+        {
+            size *= m_shape[it];
+        }
+        return size;
+    }
```
ThreeMonth03: Remoe the duplicated function `buffer_size()` and `buffer_nbytes()`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1090,11 +1086,11 @@ class SimpleArray
         {
             throw std::out_of_range(Formatter() << "SimpleArray: index " << it << " < -nghost: " << -static_cast<ssize_t>(m_nghost));
         }
-        if (it >= static_cast<ssize_t>(size() - m_nghost))
+        if (it >= static_cast<ssize_t>((buffer().nbytes() / ITEMSIZE) - m_nghost))
```
ThreeMonth03, line 1088: https://github.com/solvcon/modmesh/blob/6b1a4b953aa11f3975f7c0462022be36da8d20c4/cpp/modmesh/buffer/BufferBase.hpp#L45-L49
In `class buffebase;`, the function `size()` is `nbytes() / sizeof(int8_t`, so I don't use `buffer().size()` here.

ThreeMonth03 (commented):
@yungyuc You could review this pull request when you are available. Thank you.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: It is never a good idea to directly use AI assistant as your thought.  It is fine to use AI assistant, but do verify and organize the results before making response.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Please move the comment to be in front of `size()`, and use doxygen format:

```cpp
    /**
     * In the following numpy documentaions,
     * `nbytes()` and `size()` are calculated based on the number of the elements.
     * 1. https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.size.html
     * 2. https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.nbytes.html
     * SimpleArray matchs the behavior of numpy.ndarray.
     */
```

It is not necessary to explain `nbytes()` since it depends on `size()`.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: This is good.

yungyuc (changes requested):
- [x] Carefully analyze the way to test for the error message, or make it clear why you do not want to analyze.
- [x] Beautify comments for `SimpleArray::size()`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -834,8 +815,22 @@ class SimpleArray
...
 
-    size_t nbytes() const noexcept { return m_buffer ? m_buffer->nbytes() : 0; }
-    size_t size() const noexcept { return nbytes() / ITEMSIZE; }
+    size_t nbytes() const noexcept { return size() * ITEMSIZE; }
+    /**
+     * In the following numpy documentaions,
+     * `size()` are calculated based on the number of the elements.
+     * https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.size.html
+     * SimpleArray matchs the behavior of numpy.ndarray.
+     */
```
ThreeMonth03, line 824: - [x] Move the comment in front of `size()`.
 - [x] Use the doxygen format.
 - [x] Remove the comment about `nbytes()`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
ThreeMonth03: I think it is not possile to add a unit test checking for this error message.
In [pybind11 documentation](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays), it says:
```
By exchanging py::buffer with py::array in the above snippet, we can restrict the function so that it only accepts NumPy arrays (rather than any type of Python object satisfying the buffer protocol).
```
The initializer only accepts the numpy array, but it is impossible to throw the error message with the common numpy array.

ThreeMonth03 (commented):
- [x] Modify the comment.
- [x] Explain why I don't check the error message.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: I think you meant to say it is impossible to trigger the exception with a normally created ndarray from Python whose `offset_bytes < 0`.  In that case it makes sense.  Let me know if my understanding is incorrect.

yungyuc (approved):
Beautiful

PR #543, opened 2025-05-29
Add wave diagram in Burgers1DApp

Gene0315:
This PR adds wave diagram sub window in Burgers1DApp.

The new window looks like:
截圖 2025-05-29 中午12 40 05

But I hope the wave diagram becomes:
截圖 2025-05-29 中午12 46 45

Obviously, there are still some things to do in the wave diagram:
- plot the shock/expansion waves
- hide the wavefronts after they interset
- add a horizontal line which move from bottom to top along time

### DIFF
--- a/modmesh/pilot/_base_app.py
+++ b/modmesh/pilot/_base_app.py
@@ -385,12 +385,6 @@ class OneDimBaseApp(PilotFeature):
         self._mgr.mainWindow.addDockWidget(Qt.LeftDockWidgetArea,
                                            config_widget)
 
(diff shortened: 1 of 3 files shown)


yungyuc (commented):
Before discussing the code, we need to clarify how the feature should be implemented.  It is important to have schematics and it is good that the snapshots are shown.

It is unclear to me why do you want to draw the characteristics for the Burgers equation.  I do not see it makes the solver and solution clearer.  Let's discuss this on discord before moving forward.

I am closing this PR.

PR #546, opened 2025-06-07
Add missing break in the switch-case in `verify_python_value_datatype()`

yungyuc:


cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -125,13 +125,15 @@ static void verify_python_value_datatype(pybind11::object const & value, DataTyp
         {
             throw pybind11::type_error("Data type mismatch, expected Complex64");
         }
+        break;
```
yungyuc, line 128: We do not have unit tests covering this, but it is probably too detail for testing.

PR #549, opened 2025-06-17
Packaging: add PyInstaller script

terrychan999:
In this PR, we proposed a PyInstaller packaging script.
Currently only packaging the Python part of modmesh into a macOS app.
Maybe we could turn the line `make clean` into a toggle to decide whether we wants to build the C++ part or not before the packaging started.

### DIFF
--- a/contrib/packaging/macos.sh
+++ b/contrib/packaging/macos.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# make sure starting from project's root
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR/../.."
+
+python3 -m pip install pyinstaller
+
+make clean
+rm -rf build/portable/
+mkdir -p build/portable/
+cp -R ./modmesh build/portable/
+cp -R ./resources build/portable/
+cd build/portable/
+
+cat > modmesh.py <<'EOF'
+import os
+os.environ["QT3D_RENDERER"] = "opengl"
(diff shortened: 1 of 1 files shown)


yungyuc:
Would it make the total CI time longer to add an action for it?

contrib/packaging/macos.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# make sure starting from project's root
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR/../.."
+
+python3 -m pip install pyinstaller
```
yungyuc, line 7: Please check if pip install fails

contrib/packaging/macos.sh
```diff
@@ -0,0 +1,33 @@
...
+
+cat > modmesh.py <<'EOF'
+import os
+os.environ["QT3D_RENDERER"] = "opengl"
+from modmesh.pilot import launch
+if __name__ == "__main__":
+    launch()
+EOF
+
+pyinstaller --onedir \
```
yungyuc, line 24: Would `pilot` work with this installation?  `pilot` is the primary entry point of modmesh.

yungyuc (changes requested):
- [ ] Add more error handling.
- [ ] Clarify if `pilot` binary works.

yungyuc:
@terrychan999 Please also take it to [the discord #solvcon channel](https://discord.gg/P9U7PFv) for quick discussions.

PR #551, opened 2025-07-22
Fix: Duplicate dft operation the result should be the same

### LINKED ISSUE #547: dft function has implicit side effect
## What happened?

`dft()` function has implicit side effect when called multiple times.

https://github.com/solvcon/modmesh/blob/master/cpp/modmesh/transform/fourier.hpp#L100-L113
```
    template <template <typename> typename T1, typename T2>
    static void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
    {
        size_t N = in.size();
        for (size_t i = 0; i < N; ++i)
        {
            for (size_t j = 0; j < N; ++j)
            {
                T2 tmp = -2.0 * pi<T2> * i * j / static_cast<T2>(N);
                T1<T2> twiddle_factor{std::cos(tmp), std::sin(tmp)};
                out[i] += in[j] * twiddle_factor;
            }
        }
    }
```

## Steps to reproduce the issue

```
dft(in, out);
dft(in, out);
```

## Expected Behavior

Calling `dft()` multiple times with the same out buffer should produce consistent results, or at least document that the caller must clear out beforehand.

c1ydehhx:
## What's the problem?

Just take a look about #547, the expected output of duplicate `dft` function with the same input should be the same. Since output `SimpleArray` is not initialize before filled the value, the first `dft` operation and second `dft` operation may not have the same result. 

## How I fix this?

Initialize output `SimpleArray` by fill 0. I also write a testcase to check it work properly. 

## How to review it?

Check `test_numpy_duplicate_dft_comparison` is passed.

Resolved: #547

### DIFF
--- a/cpp/modmesh/transform/fourier.hpp
+++ b/cpp/modmesh/transform/fourier.hpp
@@ -103,6 +103,7 @@ public:
         size_t N = in.size();
         for (size_t i = 0; i < N; ++i)
         {
+            out[i] = 0;
             for (size_t j = 0; j < N; ++j)
             {
                 T2 tmp = -2.0 * pi<T2> * i * j / static_cast<T2>(N);
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -60,6 +60,25 @@ class FourierTransformTB(mm.testing.TestBase):
             self.assert_allclose(mm_output[i].real, np_output[i].real)
             self.assert_allclose(mm_output[i].imag, np_output[i].imag)
 
+    def test_numpy_duplicate_dft_comparison(self):
+        input_size = 100
+
+        mm_input = self.SimpleArray(input_size)
+        for i in range(input_size):
+            mm_input[i] = self.complex(self.real_rng(), self.imag_rng())
+
+        np_input = np.array(mm_input, copy=False)
+
+        mm_output = self.SimpleArray(input_size, self.complex())
+        mm.FourierTransform.dft(mm_input, mm_output)
+        mm.FourierTransform.dft(mm_input, mm_output)
+
+        np_output = np.fft.fft(np_input)
+
+        for i in range(input_size):
+            self.assert_allclose(mm_output[i].real, np_output[i].real)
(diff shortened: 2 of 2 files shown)


cpp/modmesh/transform/fourier.hpp
```diff
@@ -100,6 +100,8 @@ class FourierTransform
     template <template <typename> typename T1, typename T2>
     static void dft(SimpleArray<T1<T2>> const & in, SimpleArray<T1<T2>> & out)
     {
+        out.fill(0);
```
yungyuc: Why initialize the array using a distinct loop but not in the beginning of the outer loop over `i` below?  `out[i]` is read at line 112 so the initialization right before the loop over `j` is logically clearer.

```cpp
        size_t N = in.size();
        for (size_t i = 0; i < N; ++i)
        {
            /* PR note: Initialization right before the loop that reads out[i].
             * It is good to use a comment line here stating
             * that out[i] should be initialized trivially
             * here because it will be read later. */
            out[i] = 0.0;
            for (size_t j = 0; j < N; ++j)
            {
                T2 tmp = -2.0 * pi<T2> * i * j / static_cast<T2>(N);
                T1<T2> twiddle_factor{std::cos(tmp), std::sin(tmp)};
                // PR note: This is the first line reading out[i].
                out[i] += in[j] * twiddle_factor;
            }
        }
```

c1ydehhx: I see. Thanks. Fix in 8750003.

yungyuc:
Please squash to one commit because the PR is simple.

c1ydehhx:
> Please squash to one commit because the PR is simple.

No problem. I already squash to one commit (5560b84).

yungyuc:
@j8xixo12 @KHLee529 please take a look and leave a comment.

yungyuc:
I merged the PR.  Thank you @ntut-xuan .  @j8xixo12 @KHLee529 please still take a look and leave a comment

PR #553, opened 2025-07-23
Fancy Profiling Table

c1ydehhx:
## What's new?

In this PR, I create `ProfilingResultPrinter` to print the profiling result with markdown table format.

It support:
 - Fetch result with function name as key. You can use `printer["func_name"]` to get the profiling result.
    ```py
    printer["profile_sort_np"].total_time
    ```
 - Create column with lambda function. If you want to create a column that compare with specific value:
    ```py
    printer.add_column("cmp to np", lambda result : result.total_time / printer["profile_sort_np"].total_time)
    ```
 - Static-typing for all function. It will be helped with Pylance.
 - Adjustable column width.

I only consider the story in `profiling/profile_sort.py` to implement this feature. If you need more feature, please let me know.

## Result

Previous version:

```python
modmesh.call_profiler.reset()
for _ in range(it):
    test_data = np.arange(0, N, dtype='uint32')
    profile_sort_np(test_data)
    profile_sort_sa(make_container(test_data))
print_res(modmesh.call_profiler.result()["children"])
```

```md
| func                           | per call (ms)   | cmp to np       |
| ------------------------------ | --------------- | --------------- |
| profile_sort_np                | 1.043E+00       | 1.000           |
| profile_sort_sa                | 9.586E-02       | 0.092           |
```

Now:

```python
modmesh.call_profiler.reset()
for _ in range(it):
    test_data = np.arange(0, N, dtype="uint32")
    profile_sort_np(test_data)
    profile_sort_sa(make_container(test_data))

printer = ProfilingResultPrinter(modmesh.call_profiler.result()["children"])
printer.add_column("per call (ms)", lambda result: result.total_time)
printer.add_column(
    "cmp to np",
    lambda result: result.total_time / printer["profile_sort_np"].total_time,
)
printer.print_result(column_width=30)
```

```md
| func                           | per call (ms)                  | cmp to np                      |
| ------------------------------ | ------------------------------ | ------------------------------ |
| profile_sort_np                | 9.769251                       | 1.0                            |
| profile_sort_sa                | 0.904417                       | 0.09257792639374297            |
```

## How to review it?

 - Check `profiling/profile_sort.py` work properly.

### DIFF
--- a/modmesh/profiling/__init__.py
+++ b/modmesh/profiling/__init__.py
@@ -1,0 +1,31 @@
+# Copyright (c) 2025, Han-Xuan Huang <c1ydehhx@gmail.com>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of the copyright holder nor the names of its contributors
+#   may be used to endorse or promote products derived from this software
+#   without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from ._result import (  # noqa: F401
+    ProfilingResultPrinter
+)
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/modmesh/profiling/__main__.py
+++ b/modmesh/profiling/__main__.py
@@ -28,6 +28,8 @@ import functools
 import modmesh
 import numpy as np
 
+from ._result import ProfilingResultPrinter
+
 
 def profile_function(func):
     @functools.wraps(func)
@@ -35,6 +37,7 @@ def profile_function(func):
         _ = modmesh.CallProfilerProbe(func.__name__)
         result = func(*args, **kwargs)
         return result
+
     return wrapper
 
 
@@ -84,26 +87,6 @@ def profile_take_along_axis_simd(sarr, indices):
     return sarr.take_along_axis_simd(indices)
 
 
-def print_res(res):
-    out = {}
-    for r in res:
-        name = r["name"]
-        time = r["total_time"] / r["count"]
-        out[name] = time
-
-    def print_row(*cols):
-        print(str.format("| {:30s} | {:15s} | {:15s} |", *(cols[0:3])))
-
-    print_row('func', 'per call (ms)', 'cmp to np')
-    print_row('-' * 30, '-' * 15, '-' * 15)
-    npkey = [k for k in out.keys() if "np" in k][0]
-    npbase = out[npkey]
-    for k, v in out.items():
-        print_row(f"{k:8s}", f"{v:.3E}", f"{v/npbase:.3f}")
-
-    print()
-
-
 def main():
 
     N = 100000
@@ -115,7 +98,16 @@ def main():
         test_data = np.arange(0, N, dtype='uint32')
         profile_sort_np(test_data)
         profile_sort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_sort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `sort` Descending Data\n")
     modmesh.call_profiler.reset()
@@ -123,7 +115,16 @@ def main():
         test_data = np.arange(N, 0, -1, dtype='uint32')
         profile_sort_np(test_data)
         profile_sort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_sort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `sort` Random Data\n")
     modmesh.call_profiler.reset()
@@ -132,7 +133,16 @@ def main():
         np.random.shuffle(test_data)
         profile_sort_np(test_data)
         profile_sort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_sort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `argsort` Ascending Data\n")
     modmesh.call_profiler.reset()
@@ -140,7 +150,16 @@ def main():
         test_data = np.arange(0, N, dtype='uint32')
         profile_argsort_np(test_data)
         profile_argsort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_argsort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `argsort` Descending Data\n")
     modmesh.call_profiler.reset()
@@ -148,7 +167,16 @@ def main():
         test_data = np.arange(N, 0, -1, dtype='uint32')
         profile_argsort_np(test_data)
         profile_argsort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_argsort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `argsort` Random Data\n")
     modmesh.call_profiler.reset()
@@ -157,7 +185,16 @@ def main():
         np.random.shuffle(test_data)
         profile_argsort_np(test_data)
         profile_argsort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_argsort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `take_along_axis` Ascending Data\n")
     modmesh.call_profiler.reset()
@@ -169,7 +206,16 @@ def main():
         profile_take_along_axis_np(test_data, indices)
         profile_take_along_axis_sa(test_sa, idx_sa)
         profile_take_along_axis_simd(test_sa, idx_sa)
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_take_along_axis_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `take_along_axis` Descending Data\n")
     modmesh.call_profiler.reset()
@@ -181,7 +227,16 @@ def main():
         profile_take_along_axis_np(test_data, indices)
         profile_take_along_axis_sa(test_sa, idx_sa)
         profile_take_along_axis_simd(test_sa, idx_sa)
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_take_along_axis_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
     print("## `take_along_axis` Random Data\n")
     modmesh.call_profiler.reset()
@@ -195,7 +250,16 @@ def main():
         profile_take_along_axis_np(test_data, indices)
         profile_take_along_axis_sa(test_sa, idx_sa)
         profile_take_along_axis_simd(test_sa, idx_sa)
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_take_along_axis_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
 
 
 if __name__ == '__main__':
--- a/modmesh/profiling/_result.py
+++ b/modmesh/profiling/_result.py
@@ -1,0 +1,280 @@
+# Copyright (c) 2025, Han-Xuan Huang <c1ydehhx@gmail.com>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of the copyright holder nor the names of its contributors
+#   may be used to endorse or promote products derived from this software
+#   without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Any, Callable
+from dataclasses import dataclass
+
+
+@dataclass
+class ProfilingsFunctionResult:
+    name: str
+    total_time: float
+    count: float
+    children: list
+
+
+@dataclass
+class ProfilingColumnData:
+    column_name: str
+    column_value: list[float | str]
+
+
+class ProfilingTableBuilder:
+    """
+    The table builder for `ProfilingResultPrinter`.
+
+    It accept the list of `ProfilingColumnData` as attribute.
+    The column width is 30 in default,
+    the user can adjust it by modified `column_width`.
+
+    The `ProfilingColumnData` list will be convert as row datas
+    and used to generate rows by `generate_result_row_by_row`.
+
+    Attributes:
+        column_data (list[ProfilingColumnData]):
+            The column data represent list of `ProfilingColumnData`.
+        column_width (int):
+            The column width.
+    """
+
+    def __init__(
+        self,
+        column_data: list[ProfilingColumnData] = None,
+        column_width: int = 30,
+    ):
+        self.column_data = [] if column_data is None else column_data
+        self.column_width = column_width
+
+        self.row_data = []
+
+        row_count = len(self.column_data[0].column_value)
+        for i in range(row_count):
+            row_data = []
+
+            for j in range(len(self.column_data)):
+                row_data.append(self.column_data[j].column_value[i])
+
+            self.row_data.append(row_data)
+
+    def generate_header(self):
+        """
+        Generate the header with the width specific from user.
+
+        It will be the format like:
+        `| func     | column A | column B | ... |`
+        """
+        return (
+            "| "
+            + " | ".join(
+                [
+                    data.column_name.ljust(self.column_width)
+                    for data in self.column_data
+                ]
+            )
+            + " |\n"
+        )
+
+    def generate_horizontal_lines(self):
+        """
+        Generate the horizontal lines with the width specific from user.
+
+        It will be the format like:
+        `| -------- | -------- | -------- | ... |`
+        """
+        return (
+            "| "
+            + " | ".join(
+                ["".ljust(self.column_width, "-") for _ in self.column_data]
+            )
+            + " |\n"
+        )
+
+    def generate_row(self):
+        """
+        Generate the result row with the width specific from user.
+
+        It will be the format like:
+        `| func A   | 30       | 40       | ... |`
+        """
+        result = ""
+
+        if len(self.column_data) == 0:
+            return result
+
+        for row in self.row_data:
+            result += (
+                "| "
+                + " | ".join(
+                    [str(value).ljust(self.column_width) for value in row]
+                )
+                + " |\n"
+            )
+
+        return result
+
+    def generate_table_str(self):
+        """
+        Generate the table that combines header, horizontal lines and rows.
+
+        It will be the format like:
+        ```
+        | func     | column A | column B | ... |
+        | -------- | -------- | -------- | ... |
+        | func A   | 30       | 40       | ... |
+        ```
+        """
+        return (
+            self.generate_header()
+            + self.generate_horizontal_lines()
+            + self.generate_row()
+        )
+
+
+class ProfilingResultPrinter:
+    """
+    A printer that formats and displays profiling results in a table.
+
+    The profiling result should be a list of dictionaries.
+
+    Each dictionary must contain at least the following keys:
+        - "name": Name of the function (str)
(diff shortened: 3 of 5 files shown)


yungyuc (changes requested):
- [x] Please add example of the profiling results in the PR comments.

c1ydehhx:
> * [ ]  Please add example of the profiling results in the PR comments.

No problem. I added the result in PR comment.

yungyuc:
Thank you. Please also add the example of the results before the change to compare.

c1ydehhx:
> Thank you. Please also add the example of the results before the change to compare.

No problem. I add previous user story and result and new one in PR comment. I squash commit also.

yungyuc:
Thank you, @ntut-xuan .  @ThreeMonth03 @KHLee529 , please comment.

profiling/profile_sort.py
```diff
@@ -112,93 +95,183 @@ def main():
     print("## `sort` Ascending Data\n")
     modmesh.call_profiler.reset()
     for _ in range(it):
-        test_data = np.arange(0, N, dtype='uint32')
+        test_data = np.arange(0, N, dtype="uint32")
```
yungyuc: No need to change between `'` and `"`.  Please revert this unnecessary change from your patch.

profiling/profile_sort.py
```diff
@@ -112,93 +95,183 @@ def main():
...
         profile_sort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda result: result.total_time)
+    printer.add_column(
+        "cmp to np",
+        lambda result: result.total_time
```
yungyuc: It's better to keep lambda in a single line.  Please try if a shorter argument variable name may fit it in a line and an alias to `printer["profile_sort_np"].total_time`:

```python
tot = printer["profile_sort_np"].total_time
printer.add_column("cmp to np", lambda r: r.total_time / tot)
```

The two-line version is much more readable than the 5-line version.

profiling/profile_sort.py
```diff
@@ -112,93 +95,183 @@ def main():
...
-        test_data = np.arange(0, N, dtype='uint32')
+        test_data = np.arange(0, N, dtype="uint32")
         profile_sort_np(test_data)
         profile_sort_sa(make_container(test_data))
-    print_res(modmesh.call_profiler.result()["children"])
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda result: result.total_time)
```
yungyuc: Shorter lambda variable name makes the code more readable:

```python
printer.add_column("per call (ms)", lambda r: r.total_time)
```

profiling/util/profiler_table.py
```diff
@@ -0,0 +1,131 @@
+from typing import Any, Callable
```
yungyuc: Add copyright header.

profiling/util/profiler_table.py
```diff
@@ -0,0 +1,131 @@
...
+                [
+                    callable(profiling_result)
+                    for profiling_result in self.profiling_results
+                ],
+            )
+        )
+
+    def print_result(self, column_width=30):
+        builder = ProfilingTableBuilder(self.column_datas, column_width)
+        print(builder.generate_table_str())
```
yungyuc: Add vim modeline:

```
# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
```

yungyuc (changes requested):
- [x] Do not make unnecessary changes between `'` and `"`.
- [x] Try to use more concise and readable code.
- [x] Add copyright header.
- [x] Add modeline.

c1ydehhx:
No problem. I fixed all requested changes. Thanks.

profiling/profile_sort.py
yungyuc: Do you like the 2-line style?

c1ydehhx: Yeah. I think it will simplify the code. But I consider the name of lambda variable should be meaningful also. 
The user of `ProfilingResultPrinter` should know `r` present "result" in lambda variable.

yungyuc (changes requested):
LGTM.  But I just found that the new code does not have unit tests because it is not put under `modmesh` package.  Why don't you add the code under `modmesh` package so that more engineering facilities can be added?  We can discuss what is a good way to organize it under `modmesh.profiling`.

- [ ] Move utility code (`profiler_table.py`) to `modmesh.profiling` sub-package and add unit tests.

profiling/profile_sort.py
yungyuc: Lambda arguments do not need a meaningful name, while the variable inside it might, because a lambda is mostly useful in a single line.  It is commonplace to name the lambda arguments like `lambda *a, *kw: return a, kw` or `lambda a, b, c: return c, b, a`.

c1ydehhx:
> LGTM. But I just found that the new code does not have unit tests because it is not put under `modmesh` package. Why don't you add the code under `modmesh` package so that more engineering facilities can be added? We can discuss what is a good way to organize it under `modmesh.profiling`.
> 
> * [ ]  Move utility code (`profiler_table.py`) to `modmesh.profiling` sub-package and add unit tests.

No problem. I migrate `ProfilingTableBuilder` and `ProfilingResultPrinter` under `modmesh.profiling` by move the code into `modmesh/profiling/__init__.py`. The user can use `modmesh.profiling` to create printer.

```py
from modmesh.profiling import (
    ProfilingColumnData,
    ProfilingTableBuilder,
    ProfilingResultPrinter,
)
```

Also, I create 8 testcase to test printer and table builder.

```sh
tests/test_profiling_result_printer.py::TestProfilingResultPrinter::test_printer_getitem_with_valid_function_key_should_return_correct_data PASSED                                                                  [ 12%]
tests/test_profiling_result_printer.py::TestProfilingResultPrinter::test_printer_getitem_with_absent_function_key_should_raise_value_error PASSED                                                                   [ 25%]
tests/test_profiling_result_printer.py::TestProfilingResultPrinter::test_printer_add_column_should_have_correct_column PASSED                                                                                       [ 37%]
tests/test_profiling_result_printer.py::TestProfilingResultPrinter::test_printer_constructor_should_have_func_name_column_as_default PASSED                                                                         [ 50%]
tests/test_profiling_result_printer.py::TestProfilingTableBuilder::test_generate_header_should_have_correct_header PASSED                                                                                           [ 62%]
tests/test_profiling_result_printer.py::TestProfilingTableBuilder::test_generate_horizontal_lines_should_have_correct_horizontal_line PASSED                                                                        [ 75%]
tests/test_profiling_result_printer.py::TestProfilingTableBuilder::test_generate_row_data_should_have_correct_row_data PASSED                                                                                       [ 87%]
tests/test_profiling_result_printer.py::TestProfilingTableBuilder::test_generate_table_str_should_have_correct_table PASSED                                                                                         [100%]
```

Please take a look. I can add more tests if necessery. I'll squash my commit when this PR is ready-to-merge.

profiling/profile_sort.py
```diff
@@ -115,15 +98,33 @@ def main():
...
+
+    printer = ProfilingResultPrinter(
+        modmesh.call_profiler.result()["children"]
+    )
+    printer.add_column("per call (ms)", lambda r: r.total_time)
+
+    tot = printer["profile_sort_np"].total_time
+    printer.add_column("cmp to np", lambda r: r.total_time / tot)
+
+    printer.print_result(column_width=30)
```
ThreeMonth03, line 110: @c1ydehhx I have a quick question. Is there any default column in `printer`?
In other words, should we always add new column before print results?

c1ydehhx: It have default column to present function name. For the total time and call count, the user need to add it manually.

modmesh/profiling/__init__.py
```diff
@@ -0,0 +1,159 @@
+# Copyright (c) 2025, Han-Xuan Huang <xuan910625.cs13@nycu.edu.tw>
```
yungyuc: University email does not usually last long after graduation.  If you intend to be reachable through the email for more than 5 years, a non-university email address will work better.

modmesh/profiling/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Any, Callable
```
yungyuc: Please move implementation code to a private sub-module file `_report.py`.  In this file `__init__.py`, only keep import to expose API like:

```python
from ._report import ProfilingTableBuilder
```

`ProfilingTableBuilder` is assumed to be the API and other classes are considered implementation details.  Do NOT make the classes/functions private.  Only use the private sub-module to signify implementation detail.

API is code entities to be consumed from outside the package.  When creating a package, do not expose too many entities as API.  The fewer the better.

In unit testing, code can freely access the private sub-module.

Existing example for the API exposure structure: https://github.com/solvcon/modmesh/blob/0678f50f617ead08dc8e0c34aba47076708e1ac3/modmesh/pilot/__init__.py#L35

profiling/profile_sort.py
```diff
@@ -28,13 +28,16 @@
 import modmesh
 import numpy as np
 
+from modmesh.profiling import ProfilingResultPrinter
```
yungyuc: Why don't you move the script to `modmesh.profiling.__main__` so that we can just type `python3 -m modmesh.profiling` to run it?  See https://yyc.solvcon.net/en/latest/nsd/02numpy/numpy.html#make-a-package .

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
+from typing import Any
```
yungyuc: Copyright header and vim modeline.  Remember them for all new files.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    @pytest.fixture
+    def profiling_result_fixture(self) -> dict[str, Any]:
+        modmesh.call_profiler.reset()
+
+        self.run_profile_function()
+
+        result: dict[str, Any] = modmesh.call_profiler.result()["children"]
+        return result
+
+    def test_printer_getitem_with_valid_key_should_return_correct_data(
```
yungyuc: Rename to `test_getitem_valid_key`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+        self, profiling_result_fixture: dict[str, Any]
+    ) -> None:
+        printer: ProfilingResultPrinter = ProfilingResultPrinter(
+            profiling_result_fixture
+        )
+
+        assert printer["numpy_arange_100"].name == "numpy_arange_100"
+
+    def test_printer_getitem_with_absent_key_should_raise_value_error(
```
yungyuc: Rename to `test_getitem_absent_key`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+            profiling_result_fixture
+        )
+
+        try:
+            printer["numpy_arange_1000"]
+            assert False
+        except ValueError:
+            assert True
+
+    def test_printer_add_column_should_have_correct_column(
```
yungyuc: Rename to `test_add_column`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+
+        col = list(
+            filter(
+                lambda cols: cols.column_name == "tot_scale_10",
+                printer.column_datas,
+            )
+        )[0]
+        assert col.column_data[0] == tot * 10
+
+    def test_printer_constructor_should_have_func_name_column_as_default(
```
yungyuc: Rename to `test_default_column`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        printer: ProfilingResultPrinter = ProfilingResultPrinter(
+            profiling_result_fixture
+        )
+
+        assert printer["numpy_arange_100"].name == "numpy_arange_100"
+
+    def test_printer_getitem_with_absent_key_should_raise_value_error(
+        self, profiling_result_fixture: dict[str, Any]
+    ) -> None:
```
yungyuc: I don't like the specification of return type in Python.  It usually confuses more than clarifies.  You can still do it but keep it in mind.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    def expect_row_data(self) -> str:
+        return "".join(
+            [
+                f'| {"foo".ljust(30, " ")} | {"10".ljust(30, " ")} |\n',
+                f'| {"bar".ljust(30, " ")} | {"20".ljust(30, " ")} |\n',
+                f'| {"foobar".ljust(30, " ")} | {"200".ljust(30, " ")} |\n',
+            ]
+        )
+
+    def test_generate_header_should_have_correct_header(
```
yungyuc: Rename to `test_header`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        builder: ProfilingTableBuilder = ProfilingTableBuilder(
+            fake_column_datas
+        )
+
+        header: str = builder.generate_header()
+
+        assert expect_header == header
+
+    def test_generate_horizontal_lines_should_have_correct_horizontal_line(
```
yungyuc: Rename to `test_horizontal_lines`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        builder: ProfilingTableBuilder = ProfilingTableBuilder(
+            fake_column_datas
+        )
+
+        horizontal_line: str = builder.generate_horizontal_lines()
+
+        assert expect_horizontal_lines == horizontal_line
+
+    def test_generate_row_data_should_have_correct_row_data(
```
yungyuc: Rename to `test_row`.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+    ) -> None:
+        builder: ProfilingTableBuilder = ProfilingTableBuilder(
+            fake_column_datas
+        )
+
+        result_rows: str = builder.generate_result_row_by_row()
+
+        assert result_rows == expect_row_data
+
+    def test_generate_table_str_should_have_correct_table(
```
yungyuc: Rename to `test_generate_string`.

yungyuc (changes requested):
- [x] Reorganize `modmesh.profiling` with a private sub-module to provide an API-exposing structure.
- [ ] Allow to run the reporting script through `python3 -m modmesh.profiling`.
- [x] All new files need copyright header and vim modeline.
- [ ] Test default columns with null input to constructor to the reporter.
- [x] Do not use overly long function names even in tests.  Python is not Java.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
...
+                lambda cols: cols.column_name == "tot_scale_10",
+                printer.column_datas,
+            )
+        )[0]
+        assert col.column_data[0] == tot * 10
+
+    def test_printer_constructor_should_have_func_name_column_as_default(
+        self, profiling_result_fixture: dict[str, Any]
+    ) -> None:
+        printer: ProfilingResultPrinter = ProfilingResultPrinter(
```
yungyuc: You should also test the default columns with null input to the constructor.

profiling/profile_sort.py
yungyuc: And the default column should be covered in unit tests.

modmesh/profiling/__init__.py
c1ydehhx: OK. I only expose `ProfilingResultPrinter` since it should be used for user. For `ProfilingTableBuilder`, it should be internal class to provide utility for `ProfilingResultPrinter`.

profiling/profile_sort.py
c1ydehhx: Do `profile_sort.py` should be a package that can run? I treat `ProfilingResultPrinter` as a utility to create table for profiling result. I'm not sure why we need to make it as a package to run it.

tests/test_profiling_result_printer.py
c1ydehhx: OK. I add test for testing null value in constructor and handled in my profiling table.

profiling/profile_sort.py
yungyuc: You will see in the future.

c1ydehhx: Could you explain what `python3 -m modmesh.profiling` does? I'd like to understand exactly what I need to do.

c1ydehhx:
I already modified my email on copyright from organization email to my private email. 

I also create private sub-module and only expose `ProfilingResultPrinter` for user. But since `ProfilingTableBuilder` is treat as private class, I remove the tests since it's not expose and cannot be test.

In addition, I handle null value in `ProfilingResultPrinter` constructor and add test to check it work properly. I can add more tests if necessery.

The last things is trying to clearify that what `modmesh.profiling` suppose to do when `modmesh.profiling` is a package.

profiling/profile_sort.py
c1ydehhx: OK. I'll trying to load JSON files that contains result of profiling.

c1ydehhx:
Now we have a profiling packaging that can help us to make the table for profiling result. It have function name, total time, count as default column. It will read the file contains result JSON and print the table.

For example, if we have a JSON file that contains result:

```json
[
    {
        "name": "numpy_arange_100",
        "total_time": 0.002083,
        "count": 1,
        "children": []
    },
    {
        "name": "numpy_arange_1000",
        "total_time": 0.004096,
        "count": 1,
        "children": []
    }
]
```

We can use `python3 -m modmesh.profiling <result-file-path>` to create table. 
For example: `python3 -m modmesh.profiling result.json`.

```md
| func                           | total_time                     | count                          |
| ------------------------------ | ------------------------------ | ------------------------------ |
| numpy_arange_100               | 0.002083                       | 1                              |
| numpy_arange_1000              | 0.004096                       | 1                              |
```

I think it will be a good start that print the profiling result as a table. When one day profling result support output result as a JSON file, it will be good to using profiling packaging to print the result. However, since `add_column` support using lambda to create a new column, it will be a challenge that support add new column with only command.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,116 @@
...
+        col = list(
+            filter(
+                lambda cols: cols.column_name == "func", printer.column_datas
+            )
+        )[0]
+        assert col.column_data[0] == "numpy_arange_100"
+
+    def test_null_column(self) -> None:
+        try:
+            ProfilingResultPrinter(None)
```
yungyuc: "Null input to the constructor" means to call the constructor with nothing (`ProfilingResultPrinter()`), not calling with `None`.

`ProfilingResultPrinter()` should return a valid object without throwing an exception.

tests/test_profiling_result_printer.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_getitem_absent_key(
+        self, profiling_result_fixture: dict[str, Any]
+    ) -> None:
+        printer: ProfilingResultPrinter = ProfilingResultPrinter(
+            profiling_result_fixture
+        )
+
+        try:
+            printer["numpy_arange_1000"]
+            assert False
```
yungyuc: Never use `assert` in modmesh testing code.

Use helper functions in https://docs.python.org/3/library/unittest.html#unittest.TestCase .  Using [`assertRaisesRegex`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex), you can [test both the exception and error messages](https://github.com/solvcon/modmesh/blob/931e3a3bf7949da9d3e2f4952307e9ae1dae535b/tests/test_buffer.py#L51).  Your testing code here can be improved like:

```python
        with self.assertRaisesRegex(
                ValueError,
                "... (please find it out yourself) ..."
        ):
            printer["numpy_arange_1000"]
```

profiling/profile_sort.py
yungyuc: Just move this script to `modmesh.profiling.__main__`.

modmesh/profiling/__main__.py
```diff
@@ -0,0 +1,60 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+import json
+
+if __name__ == "__main__":
+    from modmesh.profiling._profiling_result_printer import (
```
yungyuc: Move this line in the module top-level without absolute import:

```python
from . import _report
```

Do not import class from the module.

modmesh/profiling/__main__.py
```diff
@@ -0,0 +1,60 @@
...
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+import json
+
+if __name__ == "__main__":
```
yungyuc: Move all code under the if to a function `main()` like:

```python
def main():
    # ... code ...

if __name__ == "__main__":
    main()
```

modmesh/profiling/_profiling_result_printer.py
```diff
@@ -0,0 +1,163 @@
+# Copyright (c) 2025, Han-Xuan Huang <c1ydehhx@gmail.com>
```
yungyuc: As I mentioned in https://github.com/solvcon/modmesh/pull/553#discussion_r2229884474 , this module should be named as `_report.py`.  Do not use long names.

The contents should use that of `profiling/profile_sort.py` and the copyright belongs to the original author.

yungyuc (changes requested):
> I also create private sub-module and only expose `ProfilingResultPrinter` for user. But since `ProfilingTableBuilder` is treat as private class, I remove the tests since it's not expose and cannot be test.

Keep the tests for `ProfilingTableBuilder`.  Testing implementation details is important.

- [x] Allow to run the reporting script through `python3 -m modmesh.profiling`.  Keep all script logic in the function `modmesh.profiling.__main__.main()` and the code under `if __name__ == "__main__":` trivial.  You should just move `profile_sort.py` to `modmesh/profiling/_report.py` and keep the original author's copyright.
- [x] Test default columns with null input to constructor (`Class()`) to the reporter.
- [x] Use moderate names for module: `modmesh.profiling._report`.  `modmesh.profiling._profiling_result_printer` is inconsiderate.
- [x] Keep the tests for `ProfilingTableBuilder`.
- [x] Do not use `assert`.  Use helper functions in https://docs.python.org/3/library/unittest.html#unittest.TestCase .

> Now we have a profiling packaging that can help us to make the table for profiling result. It have function name, total time, count as default column. It will read the file contains result JSON and print the table.

We do not need this feature now.  Please move it to another branch for a future PR.  @ThreeMonth03 , please help the serialization and deserialization of the profiling results for reporting, including the text-based table covered in this PR and the GUI part in the future.

c1ydehhx:
In this version of changes:

 - I short the file name from `_profiling_result_table.py` into `_result.py`.
 - Use `unittest.Testcase` assert function instead of python built-in assert syntax.
 - Migrate `profile_sort.py` into `modmesh.profiling.__main__.py`. The user can using `python3 -m modmesh.profiling` to run the script.
 - Add NULL testcase when construct argument is empty.
 - Keep the testcase of `ProfilingTableBuilder`.

tests/test_profiling_result.py
```diff
@@ -0,0 +1,160 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Any
+
+import numpy
+import unittest
+
+import modmesh
+from modmesh.profiling import (
```
yungyuc, line 33: A better style is to import the modules `profiling` and `_report` only, and access the classes as attributes of the modules in the testing code.  It will be more maintainable than importing the classes.

But I am fine with the current code.  This comment is only FYI.

modmesh/pilot/_profiling.py
```diff
@@ -0,0 +1,48 @@
+from PySide6 import QtCore, QtWidgets
```
yungyuc: This file does not seem to belong to this review.  Please remove.

modmesh/profiling/_result.py
```diff
@@ -0,0 +1,163 @@
...
+    children: list
+
+
+@dataclass
+class ProfilingColumnData:
+    column_name: str
+    column_data: list[float | str]
+
+
+class ProfilingTableBuilder:
```
yungyuc, line 45: It will be good to have a concise example for the table string (in the docstrings).

yungyuc (changes requested):
LGTM.  Please squash.  And I think you can use a little bit more comments.

Defect to be corrected:
- [x] Please remove the unrelated file `modmesh/pilot/_profiling.py`.

c1ydehhx:
I think it's good suggestion to have docstring in `ProfilingResultPrinter`.

For `ProfilingResultPrinter`, I add the docstring in:
 - Constructor: Contains example of profiler result struct. 
 - `__getitem__`: Mentioned absent key will raise `ValueError`
 - `add_column`: Contains example of lambda function.
 - `print_result`

I also add docstring for `ProfilingTableBuilder` to show the output format of table.

These docstring will be good to let user know how to use `ProfilingResultPrinter`. 

@KHLee529 @ThreeMonth03 could you please comment to let me know these docstring is helped you to realize how to use `ProfilingResultPrinter` or not?

modmesh/profiling/_result.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Attributes:
+        column_datas (list[ProfilingColumnData]):
+            The column data represent list of `ProfilingColumnData`.
+        column_width (int):
+            The column width.
+    """
+    def __init__(
+        self,
+        column_datas: list[ProfilingColumnData] = [],
```
yungyuc: I missed a serious issue: never use an empty container as the default value of an argument:

```python
>>> def f(a=[]): # wrong: the default value outlives the local scope
...     a.append(0)
...     print(a)
...
>>> f()
[0]
>>> f()
[0, 0]
```

This is because the lifecycle of the default argument is in the module scope, the same level as the function itself, and not inside the function scope.

A right way to handle the default value of an argument for a container is:

```python
>>> def f(a=None):
...     # build the container based on argument value
...     a = [] if a is None else a
...     a.append(0)
...     print(a)
...
>>> f()
[0]
>>> f()
[0]
```

So the code here should be corrected as:

```python
def __init__(self, column_data=None, column_width=30):
    self.column_data = [] if column_data is None else column_data
```

The argument name should be `data`, not `datas`.  `Data` is the plural form of `datum`.

---

This is one reason why I do not like typing in Python.  The right way of writing Python code should use dynamic types.

modmesh/profiling/_result.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        return (
+            "| "
+            + " | ".join(
+                ["".ljust(self.column_width, "-") for _ in self.column_datas]
+            )
+            + " |\n"
+        )
+
+    def generate_result_row_by_row(self):
```
yungyuc: The shorter name `generate_row` for the function would be easier to read.

yungyuc (changes requested):
- [x] Never use a container (like list or dict) as the default value of a keyword argument.
- [x] Fix the typo "datas".  Should be "data".

ThreeMonth03:
> I think it's good suggestion to have docstring in `ProfilingResultPrinter`.
> 
> For `ProfilingResultPrinter`, I add the docstring in:
> 
> * Constructor: Contains example of profiler result struct.
> * `__getitem__`: Mentioned absent key will raise `ValueError`
> * `add_column`: Contains example of lambda function.
> * `print_result`
> 
> I also add docstring for `ProfilingTableBuilder` to show the output format of table.
> 
> These docstring will be good to let user know how to use `ProfilingResultPrinter`.
> 
> @KHLee529 @ThreeMonth03 could you please comment to let me know these docstring is helped you to realize how to use `ProfilingResultPrinter` or not?

Thanks, @c1ydehhx. The example in [__main__.py](https://github.com/solvcon/modmesh/pull/553/files#diff-340b2c5de1b676cea89178601435e22bafb4932c06f942c85e8c08b87dc87591) is clear.

c1ydehhx:
I just fix the issue about default value and check all tests passed. 

Also, I adjust the argument name of `ProfilingColumnData` from `column_data` to `column_value`, since we usually call `list[ProfilingColumnData]` as `column_data`. It help to distinguish between the value in `column_data` and `column_data`. (That prevent the situation like `column_data.column_data`.)

```py
@dataclass
class ProfilingColumnData:
    column_name: str
    column_value: list[float | str] # I adjust it from column_data to column_value

column_data = ProfilingColumnData("abc", [10, 20, 30])

# By adjust it, it prevent the situation like `column_data.column_data[0]`, which is easy to confuse.
value = column_data.column_value[0]
```

Besides, I short the function name to make it easier to read. From `generate_result_row_by_row` to `generate_row`.

modmesh/profiling/_result.py
```diff
@@ -0,0 +1,280 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Any, Callable
+from dataclasses import dataclass
+
+
+@dataclass
```
yungyuc, line 31: The data classes will be moved to C++ in the future.  We can keep them here for now.

tests/test_profiling_result.py
```diff
@@ -0,0 +1,160 @@
...
+    def test_generate_header(self) -> None:
+        builder: ProfilingTableBuilder = ProfilingTableBuilder(
+            self.fake_column_data
+        )
+
+        header: str = builder.generate_header()
+
+        self.assertEqual(self.expect_header, header)
+
+    def test_generate_hr_lines(
```
yungyuc: This should use a single line like what you did elsewhere in this file.  It does not read smooth to use two different styles for similar signatures.  I'll change it for you.

yungyuc (approved):

PR #557, opened 2025-07-27
Feat: Speed up the median operation

ThreeMonth03:
In this pull request, I try to speed up the operation `median()`.
By the way, the work is in progress. 

Feature:
 - [x] Accelerate the function `partition` with SIMD.
 - [ ] Accelerate the process to move the elements from `SimpleArray` to `small_vector`.

Invalid attempt:
 - [x] Block partition.
 - [x] Floyd-Rivest Algorithm.

Result:
The effectiveness of `partition_simd` is slight.

# median (dtype=<class 'numpy.float64'>)
## N = 1000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 1.968E+00       | 1.000           |
| sa         | 9.629E-03       | 0.005           |
| simd_sa    | 9.400E-03       | 0.005           |

## N = 10000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 7.739E-02       | 1.000           |
| sa         | 1.107E-01       | 1.430           |
| simd_sa    | 1.089E-01       | 1.408           |

## N = 100000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 9.781E-01       | 1.000           |
| sa         | 1.650E+00       | 1.687           |
| simd_sa    | 1.648E+00       | 1.685           |

## N = 1000000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 9.245E+00       | 1.000           |
| sa         | 1.371E+01       | 1.483           |
| simd_sa    | 1.362E+01       | 1.473           |

## N = 10000000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 8.719E+01       | 1.000           |
| sa         | 1.299E+02       | 1.490           |
| simd_sa    | 1.303E+02       | 1.495           |

# median (dtype=<class 'numpy.float64'>) (non-contiguous)
## N = 1000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 1.668E-02       | 1.000           |
| sa         | 4.367E-03       | 0.262           |
| simd_sa    | 3.892E-03       | 0.233           |

## N = 10000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 6.396E-02       | 1.000           |
| sa         | 6.823E-02       | 1.067           |
| simd_sa    | 6.726E-02       | 1.052           |

## N = 100000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 4.492E-01       | 1.000           |
| sa         | 5.413E-01       | 1.205           |
| simd_sa    | 5.427E-01       | 1.208           |

## N = 1000000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 5.322E+00       | 1.000           |
| sa         | 6.207E+00       | 1.166           |
| simd_sa    | 6.216E+00       | 1.168           |

## N = 10000000
| func       | per call (ms)   | cmp to np       |
| ---------- | --------------- | --------------- |
| np         | 5.239E+01       | 1.000           |
| sa         | 8.148E+01       | 1.555           |
| simd_sa    | 8.133E+01       | 1.553           |

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -261,7 +261,22 @@ public:
             acopy[i] = athis->at(sidx);
             ++i;
         } while (athis->next_sidx(sidx));
-        return median_op(acopy);
+        return simd::generic::median(acopy.begin(), acopy.end());
+    }
+
+    value_type parallel_median() const
+    {
+        auto athis = static_cast<A const *>(this);
+        const size_t n = athis->size();
+        small_vector<T> acopy(n);
+        auto sidx = athis->first_sidx();
+        size_t i = 0;
+        do
+        {
+            acopy[i] = athis->at(sidx);
+            ++i;
+        } while (athis->next_sidx(sidx));
+        return simd::median(acopy.begin(), acopy.end());
     }
 
     value_type average_op(small_vector<value_type> & sv, small_vector<value_type> & weight) const
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -31,6 +31,8 @@
 #include <modmesh/simd/simd_generic.hpp>
 #include <modmesh/simd/neon/neon_type.hpp>
 #include <modmesh/simd/neon/neon_alias.hpp>
+#include <algorithm>
+#include <stdexcept>
 
 #if defined(__aarch64__)
 #include <arm_neon.h>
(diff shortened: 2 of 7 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -261,7 +261,22 @@ class SimpleArrayMixinCalculators
             acopy[i] = athis->at(sidx);
             ++i;
         } while (athis->next_sidx(sidx));
-        return median_op(acopy);
+        return simd::generic::median(acopy.begin(), acopy.end());
+    }
+
+    value_type parallel_median() const
```
yungyuc, line 267: It seems to me that you are not following the naming convention used in this file.  I am not sure whether or not you have read through the code.  If you have not, please check the related code first before adding inconsistent code.

If you have read through the code and decided to use a different style, please provide your argument and discuss.  It is not apparent to me why a different naming style is needed for your addition.

In my experience, it is not right to name a function using SIMD as "parallel" something, which usually implies to use multi cores or machines.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -222,6 +224,267 @@ void div(T * dest, T const * dest_end, T const * src1, T const * src2)
     }
 }
 
+/**
+ * @brief 使用SIMD加速的三数取中法选择pivot
```
yungyuc, line 228: Everything in the code base, including comments, should use English.

The comments do not look like your words.  If you copied the code from elsewhere, please close this PR and avoid doing that.  I do not want this repository to use code copied from elsewhere.

If you generated the code using AI, it is not right to directly take the comments from the generated code.  We should inspect the generated code line by line and fully understand it, including comments.

Any code generated from AI (or copied from elsewhere) should be explicitly indicated in PR inline annotations.

The comments in the code are over-explaining.  It suggests the code does not speak for itself, and should be improved.

yungyuc (commented):
> In this pull request, I try to speed up the operation `median()`. By the way, the work is in progress.

Thank you for your work, but I have serious concerns about it.  Please see my comments inline.

> Feature:
> 
> * [x]  Accelerate the function `partition` with SIMD.
> * [ ]  Accelerate the process to move the elements from `SimpleArray` to `small_vector`.
> 
> Invalid attempt:
> 
> * [x]  Block partition.
> * [x]  Floyd-Rivest Algorithm.
> 
> Result: The effectiveness of `partition_simd` is slight.

It may be too early to try SIMD for median.  Your profiling results suggested SimpleArray is generally slower than numpy.  It may be more fruitful to speed up SimpleArray without SIMD by studying how numpy did it.

yungyuc:
This should be reworked.  I am closing this PR because starting over will be more productive.  The speedup should use more detailed profiling.

ThreeMonth03:
> This should be reworked. I am closing this PR because starting over will be more productive. The speedup should use more detailed profiling.

Thanks, it sounds like a good idea. I will make another pull request about the detail analysis of the profiling with cpp.

c1ydehhx:
I'm trying to figure out how to do profiling on median function. The idea I trying to do is, insert the CallProfiler probe on SimpleArray C++ median function. Something like:

```C++
value_type median() const
    {
        PROFILE_PROBE_START("topic A");
        auto athis = static_cast<A const *>(this);
        const size_t n = athis->size();
        small_vector<T> acopy(n);
        auto sidx = athis->first_sidx();
        size_t i = 0;
        PROFILE_PROBE_END();

        PROFILE_PROBE_START("topic B"));
        do
        {
            acopy[i] = athis->at(sidx);
            ++i;
        } while (athis->next_sidx(sidx));
        PROFILE_PROBE_END()

        PROFILE_PROBE_START("topic C"));
        auto result = median_op(acopy);
        PROFILE_PROBE_END();
        return result;
    }
```

And call the median function on python script.

```py
def main():
    modmesh.call_profiler.reset()
    simple_array = make_container(numpy.arange(0, 1e6, dtype="uint8"))
    median = median_array(simple_array)
    res = modmesh.call_profiler.result()["children"]
    print(res)
```

And fetch the result to know what's going on.

Since we don't have `PROFILE_PROBE_START` and `PROFILE_PROBE_END`, I would like to implement one.

Is it a good motivation? @yungyuc

yungyuc:
There is an existing macro `MODMESH_PROFILE`. Do not invent a duplicate macro: https://github.com/solvcon/modmesh/blob/4b0c17a57f771acae9c42889d40b18b6f67e4ee1/cpp/modmesh/toggle/profile.hpp#L276

```cpp
value_type median() const
{
    MODMESH_TIME("SimpleArray<T>::median()"):
    auto athis = static_cast<A const *>(this);
    const size_t n = athis->size();
    small_vector<T> acopy(n);
    auto sidx = athis->first_sidx();
    size_t i = 0;

    do
    {
        acopy[i] = athis->at(sidx);
        ++i;
    } while (athis->next_sidx(sidx));

    auto result = median_op(acopy);
    return result;
}

auto median_op(SimpleArray)
{
    MODMESH_TIME("SimpleArray<T>::median_op()");
    // ... code
}
```

Then you should see `median()` calls `median_op()` in the profiler radix tree.

@q40603 , please correct me if I am wrong.

c1ydehhx:
> There is an existing macro `MODMESH_PROFILE`. Do not invent a duplicate macro:
> 
> https://github.com/solvcon/modmesh/blob/4b0c17a57f771acae9c42889d40b18b6f67e4ee1/cpp/modmesh/toggle/profile.hpp#L276
> 
> ```c++
> value_type median() const
> {
>     MODMESH_PROFILE("SimpleArray<T>::median()"):
>     auto athis = static_cast<A const *>(this);
>     const size_t n = athis->size();
>     small_vector<T> acopy(n);
>     auto sidx = athis->first_sidx();
>     size_t i = 0;
> 
>     do
>     {
>         acopy[i] = athis->at(sidx);
>         ++i;
>     } while (athis->next_sidx(sidx));
> 
>     auto result = median_op(acopy);
>     return result;
> }
> 
> auto median_op(SimpleArray)
> {
>     MODMESH_PROFILE("SimpleArray<T>::median_op()"):
> }
> ```
> 
> Then you should see `median()` calls `median_op()` in the profiler radix tree.
> 
> @q40603 , please correct me if I am wrong.

No problem. Thanks to let me know this. I'll trying to do it.

ThreeMonth03:
> There is an existing macro `MODMESH_PROFILE`. Do not invent a duplicate macro:
> 
> https://github.com/solvcon/modmesh/blob/4b0c17a57f771acae9c42889d40b18b6f67e4ee1/cpp/modmesh/toggle/profile.hpp#L276
> 
> ```c++
> value_type median() const
> {
>     MODMESH_TIME("SimpleArray<T>::median()"):
>     auto athis = static_cast<A const *>(this);
>     const size_t n = athis->size();
>     small_vector<T> acopy(n);
>     auto sidx = athis->first_sidx();
>     size_t i = 0;
> 
>     do
>     {
>         acopy[i] = athis->at(sidx);
>         ++i;
>     } while (athis->next_sidx(sidx));
> 
>     auto result = median_op(acopy);
>     return result;
> }
> 
> auto median_op(SimpleArray)
> {
>     MODMESH_TIME("SimpleArray<T>::median_op()");
>     // ... code
> }
> ```
> 
> Then you should see `median()` calls `median_op()` in the profiler radix tree.
> 
> @q40603 , please correct me if I am wrong.

I have a quick question. Is it possible to insert the profiler probe to function declaration in another cpp file?
It might make code messy to insert the probe in the source code.

yungyuc:
> I have a quick question. Is it possible to insert the profiler probe to function declaration in another cpp file? It might make code messy to insert the probe in the source code.

Manually adding profiling directives in source code is a practical way for scope-based profiler.  It is indeed ad hoc and can make code messier (I would say just slightly), but it also gives us absolute control over what to be profiled.

It is not a good idea to add profiling entry for a function that is supposed to be inlined in a tight loop.  Automatically inserting profiling entries cannot make the distinction.

In theory, it is possible for a runtime helper thread to insert the scoped-based profiling entry automatically when entering and exiting a call stack.  I read about a tool doing that a year or two ago, but cannot find it now.

yungyuc:
@ThreeMonth03 I just found something like what I read before: https://github.com/yosefk/funtrace

c1ydehhx:
> There is an existing macro `MODMESH_PROFILE`. Do not invent a duplicate macro:
> 
> https://github.com/solvcon/modmesh/blob/4b0c17a57f771acae9c42889d40b18b6f67e4ee1/cpp/modmesh/toggle/profile.hpp#L276
> 
> ```c++
> value_type median() const
> {
>     MODMESH_TIME("SimpleArray<T>::median()"):
>     auto athis = static_cast<A const *>(this);
>     const size_t n = athis->size();
>     small_vector<T> acopy(n);
>     auto sidx = athis->first_sidx();
>     size_t i = 0;
> 
>     do
>     {
>         acopy[i] = athis->at(sidx);
>         ++i;
>     } while (athis->next_sidx(sidx));
> 
>     auto result = median_op(acopy);
>     return result;
> }
> 
> auto median_op(SimpleArray)
> {
>     MODMESH_TIME("SimpleArray<T>::median_op()");
>     // ... code
> }
> ```
> 
> Then you should see `median()` calls `median_op()` in the profiler radix tree.
> 
> @q40603 , please correct me if I am wrong.

I would like to create a self-defined scope profiling result. Not per-function profiling result.

The reason I would like to do it is, in median function, that have three part that can measure the performance.

```cpp
value_type median() const
    {
        PROFILE_PROBE_START("topic A");
        auto athis = static_cast<A const *>(this);
        const size_t n = athis->size();
        small_vector<T> acopy(n);
        auto sidx = athis->first_sidx();
        size_t i = 0;
        PROFILE_PROBE_END();

        PROFILE_PROBE_START("topic B"));
        do
        {
            acopy[i] = athis->at(sidx);
            ++i;
        } while (athis->next_sidx(sidx));
        PROFILE_PROBE_END()

        PROFILE_PROBE_START("topic C"));
        auto result = median_op(acopy);
        PROFILE_PROBE_END();
        return result;
    }
```

So that we can have three part of profiling result that related to median, represent three different part to know which part is bottleneck.

Another reason why I wouldn't like to split the function is, the return value need to be handled and it will be a little annoying.

In summary, although we have a `MODMESH_TIME` macro, is function based and only can profile a whole function, not the part of function.

yungyuc:
> I would like to create a self-defined scope profiling result. Not per-function profiling result.
> 
> The reason I would like to do it is, in median function, that have three part that can measure the performance.

The need for manual sectioning for profiling suggests that the function is too long to be well organized.  It is a bad smell that you should refactor.

It is why I follow the principle to prefer to refactor when I need to speed up.  Profiling is the first step of speeding up.  If code is not easy to be profiled, it should be refactored.

> 
> ```c++
> value_type median() const
>     {
>         PROFILE_PROBE_START("topic A");
>         auto athis = static_cast<A const *>(this);
>         const size_t n = athis->size();
>         small_vector<T> acopy(n);
>         auto sidx = athis->first_sidx();
>         size_t i = 0;
>         PROFILE_PROBE_END();
> 
>         PROFILE_PROBE_START("topic B"));
>         do
>         {
>             acopy[i] = athis->at(sidx);
>             ++i;
>         } while (athis->next_sidx(sidx));
>         PROFILE_PROBE_END()
> 
>         PROFILE_PROBE_START("topic C"));
>         auto result = median_op(acopy);
>         PROFILE_PROBE_END();
>         return result;
>     }
> ```
> 
> So that we can have three part of profiling result that related to median, represent three different part to know which part is bottleneck.
> 
> Another reason why I wouldn't like to split the function is, the return value need to be handled and it will be a little annoying.

Properly separate functions will make code more organized and readable.  This is of course tricky.  It's OK to try it back and forth.

> 
> In summary, although we have a `MODMESH_TIME` macro, is function based and only can profile a whole function, not the part of function.

Please stick with the existing `MODMESH_TIME` macro without creating a new one.  Do not create `..._START/BEGIN()` and `..._END()` which are error-prone.  Everyone makes the mistake of forgetting about opening or closing macro.

If you insist to add a profiling entry for part of the code, use C++ scope.

```cpp
value_type median() const
{
    MODMESH_TIME("SimpleArray<T>::median()");
    auto athis = static_cast<A const *>(this);
    const size_t n = athis->size();
    small_vector<T> acopy(n);

    {
        MODMESH_TIME("SimpleArray<T>::median(): copy");
        // Move the code initializing loop variables here.
        // Profiling helps identify this structure. If this section appears in many places
        // It should be factored out to be a function.
        auto sidx = athis->first_sidx();
        size_t i = 0;
        do
        {
            acopy[i] = athis->at(sidx);
            ++i;
        } while (athis->next_sidx(sidx));
    }

    auto result = median_op(acopy);
    return result;
}

auto median_op(SimpleArray)
{
    MODMESH_TIME("SimpleArray<T>::median_op()");
    // ... code
}
```

However, adding sectioned profiling entry is confusing.  I discourage you from doing it.

Issue #561, opened 2025-07-31
Rename macros whose name contains `CALLPROFILER`


yungyuc:
The class `CallProfiler` is an implementation detail of the home brewed profiler in modmesh.  The CPP macro API using it should not contain the specific name.  Thus, the follow macros should be renamed accordingly:

1. `USE_CALLPROFILER_PROFILE_THIS_FUNCTION` to `MODMESH_PROFILE_FUNCTION`.
2. `USE_CALLPROFILER_PROFILE_THIS_SCOPE` to `MODMESH_PROFILE_SCOPE`.

SidoJain:
Please assign this to me. I would like to help.

yungyuc:
Thank you @SidoJain .  Look forward to your contribution.

yungyuc:
When is this planned to have a PR?

yungyuc:
When would a PR be planned for this?

SidoJain:
Im sorry I was caught up in some work. I'll have a PR ready within the hour.

SidoJain:
Just to clarify, I just need to replace `USE_CALLPROFILER_PROFILE_THIS_FUNCTION` with `MODMESH_PROFILE_FUNCTION` and `USE_CALLPROFILER_PROFILE_THIS_SCOPE` with `MODMESH_PROFILE_SCOPE`, right? Or do i need to consider something else as well?

yungyuc:
> Im sorry I was caught up in some work. I'll have a PR ready within the hour.

You are perfectly fine.  It's just regular reminder.  Any schedule is OK and the project accommodates.

> Just to clarify, I just need to replace `USE_CALLPROFILER_PROFILE_THIS_FUNCTION` with `MODMESH_PROFILE_FUNCTION` and `USE_CALLPROFILER_PROFILE_THIS_SCOPE` with `MODMESH_PROFILE_SCOPE`, right? Or do i need to consider something else as well?

Yes, that's the scope of this issue.  Consistency to the whole system should be considered, but that applies to any PR.  We'll handle it in PR.

yungyuc:
Deassign for lack of activity.  Will assign again when there is.

c1ydehhx:
I think this task is already solved with #581. Maybe we can close this issue? @yungyuc

yungyuc:
Thanks for reminding, @c1ydehhx .  Closing.

PR #559, opened 2025-07-30
Add profiling on median function

c1ydehhx:
## What's new?

In this PR, I add profiling macro on median function to measure the bottleneck of median function. I create `profiling/profile_median.py` also. 

The reason why I create this script is, `MODMESH_TIME` macro use `TimeRegistry` class, which is not tree structure. We need to migrate `MODMESH_TIME` from `TimeRegistry` to `ProfileCallProble`, which support tree struct. And I can use this script to generate call stack result.

## The result

I create 1000000 size of `SimpleArray` and do median. The following output shows the profiling result.

```
median() : count = 1 , time = 2.33615 (second)
median::copy() : count = 1 , time = 0.028013 (second)
median::init() : count = 1 , time = 1.292e-06 (second)
median::run() : count = 1 , time = 2.30812 (second)
median::select_kth_1() : count = 1 , time = 4.2e-08 (second)
median::select_kth_2() : count = 1 , time = 0.0166571 (second)
median::select_kth_3() : count = 1 , time = 2.29145 (second)
median_op() : count = 1 , time = 2.30812 (second)
```

The tree struct (I manually create by keyboard) should be:

```
median_op() : count = 1 , time = 2.30812 (second)
    median::select_kth_1() : count = 1 , time = 4.2e-08 (second)
    median::select_kth_2() : count = 1 , time = 0.0166571 (second)
    median::select_kth_3() : count = 1 , time = 2.29145 (second)
 
median() : count = 1 , time = 2.33615 (second)
    median::run() : count = 1 , time = 2.30812 (second)
    median::copy() : count = 1 , time = 0.028013 (second)
    median::init() : count = 1 , time = 1.292e-06 (second)
```

PR #563, opened 2025-08-01
refactor CALLPROFILER to MODMESH

SidoJain:
I have renamed the following macros:
1. `USE_CALLPROFILER_PROFILE_THIS_FUNCTION` to `MODMESH_PROFILE_FUNCTION`.
2. `USE_CALLPROFILER_PROFILE_THIS_SCOPE` to `MODMESH_PROFILE_SCOPE`.

For Issue #561

### DIFF
--- a/cpp/modmesh/toggle/RadixTree.hpp
+++ b/cpp/modmesh/toggle/RadixTree.hpp
@@ -362,11 +362,11 @@ private:
 // ref: https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html
 #define __CROSS_PRETTY_FUNCTION__ __PRETTY_FUNCTION__
 #endif
-#define USE_CALLPROFILER_PROFILE_THIS_FUNCTION() modmesh::CallProfilerProbe __profilerProbe##__COUNTER__(modmesh::CallProfiler::instance(), __CROSS_PRETTY_FUNCTION__)
-#define USE_CALLPROFILER_PROFILE_THIS_SCOPE(scopeName) modmesh::CallProfilerProbe __profilerProbe##__COUNTER__(modmesh::CallProfiler::instance(), scopeName)
+#define MODMESH_PROFILE_FUNCTION() modmesh::CallProfilerProbe __profilerProbe##__COUNTER__(modmesh::CallProfiler::instance(), __CROSS_PRETTY_FUNCTION__)
+#define MODMESH_PROFILE_SCOPE(scopeName) modmesh::CallProfilerProbe __profilerProbe##__COUNTER__(modmesh::CallProfiler::instance(), scopeName)
 #else
-#define USE_CALLPROFILER_PROFILE_THIS_FUNCTION() // do nothing
-#define USE_CALLPROFILER_PROFILE_THIS_SCOPE(scopeName) // do nothing
+#define MODMESH_PROFILE_FUNCTION() // do nothing
+#define MODMESH_PROFILE_SCOPE(scopeName) // do nothing
 #endif
 
 } /* end namespace modmesh */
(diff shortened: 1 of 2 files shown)


yungyuc:
@c1ydehhx Please help review.

c1ydehhx (commented):
[Discuss]
I'm consider the new name of macro. 

`MODMESH_PROFILE_FUNCTION` fetch function name automatically, but it will generate a long function name in name field. For example, in #561, `median()` will be 

`
value_type modmesh::detail::SimpleArrayMixinCalculators<modmesh::SimpleArray<unsigned char>, unsigned char>::median() const [A = modmesh::SimpleArray<unsigned char>, T = unsigned char]
`

Which is hard to read.

I think the new name of macro should guide user that don't use `MODMESH_PROFILE_FUNCTION` at most of time, it will destory the conciseness of result. Since we already want to rename all profiling macro, it may a good timing to introduce this idea.

I'll suggest:

```
USE_CALLPROFILER_PROFILE_THIS_FUNCTION -> MODMESH_PROFILE_PROBE_AUTO
USE_CALLPROFILER_PROFILE_THIS_SCOPE -> MODMESH_PROFILE_PROBE(scope_name)
```

Or, since we already in modmesh this project. It can be:

```
PROFILE_PROBE_AUTO
PROFILE_PROBE(scope_name)
```

By using `PROFILE_PROBE` rename to the second macro name, it may guide user that using `PROFILE_PROBE` as most of time.

Please comment your idea below and we can discuss.

yungyuc:
> [Discuss] I'm consider the new name of macro.
> 
> `MODMESH_PROFILE_FUNCTION` fetch function name automatically, but it will generate a long function name in name field. For example, in #561, `median()` will be
> 
> `value_type modmesh::detail::SimpleArrayMixinCalculators<modmesh::SimpleArray<unsigned char>, unsigned char>::median() const [A = modmesh::SimpleArray<unsigned char>, T = unsigned char]`
> 
> Which is hard to read.

I agree the automatically generate name is hard to read in profiling report.  It is a point to be improved.

> I think the new name of macro should guide user that don't use `MODMESH_PROFILE_FUNCTION` at most of time, it will destory the conciseness of result. Since we already want to rename all profiling macro, it may a good timing to introduce this idea.
> 
> I'll suggest:
> 
> ```
> USE_CALLPROFILER_PROFILE_THIS_FUNCTION -> MODMESH_PROFILE_PROBE_AUTO
> USE_CALLPROFILER_PROFILE_THIS_SCOPE -> MODMESH_PROFILE_PROBE(scope_name)
> ```

Different probing helper macros should not have an "embedding" name, that is, `NAME` should not be fully included in `NAME_AUTO`.  Thus, when grepping `NAME1` we get only `NAME1`, and so as `NAME2`.  The proposal of using `MODMESH_PROFILE_PROBE_AUTO` and `MODMESH_PROFILE_PROBE` does not meet the criteria.  When grepping `MODMESH_PROFILE_PROBE`, we will get both `MODMESH_PROFILE_PROBE_AUTO` and `MODMESH_PROFILE_PROBE`.

The new names should remain to use my original proposal:

1. `USE_CALLPROFILER_PROFILE_THIS_FUNCTION` to `MODMESH_PROFILE_FUNCTION`.

   The name `MODMESH_PROFILE_FUNCTION` is clear enough for a proficient C++ coder that it may take the function name automatically because it does not allow an argument for the profiling node name.  To use a manually set node name, use `MODMESH_PROFILE_SCOPE` instead.
2. `USE_CALLPROFILER_PROFILE_THIS_SCOPE` to `MODMESH_PROFILE_SCOPE`.

> Or, since we already in modmesh this project. It can be:
> 
> ```
> PROFILE_PROBE_AUTO
> PROFILE_PROBE(scope_name)
> ```

`MODMESH` should be prefix in the macro names to avoid collision with other macros.  Failure to avoid macro name collision makes is hard if not impossible to integrate with foreign C/C++ code base.

yungyuc (changes requested):
@SidoJain Please rebase to the latest `master` and also rename the new code using the old macro names introduced in #559 .

PR #570, opened 2025-08-08
Fix SIMD dispatching malfunctioning

KHLee529:
In the previous version, SIMD dispatching mechanism is buggy that makes all program flow go to generic flow which is designed for platforms without SIMD support. The program is therefore not accelerated at all.
This PR fix this problem.

### DIFF
--- a/cpp/modmesh/simd/simd_support.cpp
+++ b/cpp/modmesh/simd/simd_support.cpp
@@ -83,8 +83,10 @@ SimdFeature detect_simd()
 #endif
 #endif /* defined(__aarch64__) || defined(__arm__) */
 
-    CurrentFeature = SIMD_NONE;
-
+    if (CurrentFeature == SIMD_UNKNOWN)
+    {
+        CurrentFeature = SIMD_NONE;
+    }
     return CurrentFeature;
 }
 


KHLee529:
@yungyuc This PR fixes the malfunctioning SIMD dispatching mechanism. This part is easy but critical for SIMD functionality so I make it a standalone PR. Please review it.

yungyuc (approved):
> In the previous version, SIMD dispatching mechanism is buggy that makes all program flow not going to generic flow. The program is therefore not accelerated at all. This PR fix this problem.

It is not entirely clear what do you mean by "all program flow not going to generic flow", but I will just merge since the logic in the patch is clear.

Let us continue discuss in this PR after merging for what do you mean by the statement.

KHLee529:
@yungyuc I wrote the description wrong. I've fixed it and please check if it's clearer.

yungyuc:
> In the previous version, SIMD dispatching mechanism is buggy that makes
> all program flow not going to generic flow. The program is therefore not
> accelerated at all.
> This commit fix this problem.

Thank you, @KHLee529 .  Your current description is clear:

> all program flow go to generic flow which is designed for platforms without SIMD support

But it is the opposite to what you had in the commit log:

> ... all program flow not going to generic flow.

I write it down here to highlight, so that it will be easier to reference from the future.

In addition, our conclusion in the commit log remains unchanged:

> "The program is therefore not accelerated at all.  This commit fix this problem."

KHLee529:
Yes exactly, our conclusion stays that the same. The problem is fixed.

Issue #514, opened 2025-04-18
Implement arithmetic operations as `SimpleArray` member functions


yungyuc:
Implement the selected arithmetic member functions for `SimpleArray` (both C++ and Python wrapper and test in Python):

- [x] [`numpy.add()`](https://numpy.org/doc/stable/reference/generated/numpy.add.html#numpy.add) as `SimpleArray::add()`
- [x] [`numpy.subtract()`](https://numpy.org/doc/stable/reference/generated/numpy.subtract.html#numpy.subtract) as `SimpleArray::subtract()`
- [x] [`numpy.multiply()`](https://numpy.org/doc/stable/reference/generated/numpy.multiply.html#numpy.multiply) as `SimpleArray::multiply()`
- [x] [`numpy.divide()`](https://numpy.org/doc/stable/reference/generated/numpy.divide.html#numpy.divide) as `SimpleArray::divide()`
- [x] Allow function chaining (fluent API style) of in-place arithmetic functions.
- [x] Make the `SimpleArray` functions run faster than `numpy`.
- [x] Use function overloading instead of template when it applies.

yungyuc:
@KHLee529 Are you interested in this task?  Each function may use one or more PRs.

KHLee529:
@yungyuc no problem, I'll take this task after my SIMD PR is finished.

yungyuc:
Thank you, @KHLee529 .  Let me assign it to you and we will set a schedule later.

KHLee529:
PR #533 is the implementation for this issue.

yungyuc:
Thank you, @KHLee529 .  I will close this issue later.

KHLee529:
Don't we just use this issue to keep tracking the progress of making in-place operations support function chaining?

yungyuc:
That is a good idea.  Also, we may use this issue to make `SimpleArray` arithmetics run faster than `numpy`.

KHLee529:
After discussing with @ntut-xuan, the alias library of neon should be implemented in function overloading instead of template. This issue will track this problem as well.

yungyuc:
> * [ ]  Allow function chaining (fluent API style) of in-place arithmetic functions.
> * [ ]  Use function overloading instead of template when it applies.

@KHLee529 could you please help review the progress of the above 2 items?

KHLee529:
> > * [ ]   Use function overloading instead of template when it applies.
> 
[...]

PR #572, opened 2025-08-09
Improve performance of SimpleArray arithmetic operation 

KHLee529:
# Description

As mentioned in ISSUE #514, the arithmetic operation of `SimpleArray` needs to be improved. 

This PR implements the following improvement.

* Accelerate non-SIMD arithmetic operation
* Accelerate SIMD arithmetic operation
* Refine neon alias functions from template to function overloading and make it inline

### Non-SIMD operation acceleration

I found that non-SIMD arithmetic can be twice faster when implemented using raw pointer operation instead of the original operation that always get and set the value using `SimpleArray` public interface.
The non-SIMD operation therefore is improved using raw pointer operation.

### SIMD acceleration

Besides, the out-of-place arithmetic operations are found to be slow due to the buffer-copy when constructing new objects. The function flows are change a bit from "Copy constructing return array -> do in-place operation" to "Constructing emtpy return objects -> call out-of-place SIMD operation with return objects set as output buffer"

### Refine neon alias implementation

Since the SIMD implementation is first wrapping all the intrinsic functions into aliases to remove postfix from the function names, and then use the wrapped alias functions to implement template functions. 
However, the aliases which was defined in a standalone compilation unit actually slows down the program due to the overhead of function calls. The alias functions are therefore made into inline function with definition directly written in the header so that they can be better compiled as SIMD instructions in the program flow to decrease overhead.

Besides, alias functions which was implemented as template are converted to overloaded functions with same name to improve readability loss from strange template syntax. 

# Profiling Result

<details>
<summary>Addition</summary>

## add N = 4M range: $[0, 255]$ type: `uint8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 2.051E+00       | 1.000           | 0.516           | 1.042           |
| profile_add_sa                                     | 3.972E+00       | 1.936           | 1.000           | 2.018           |
| profile_add_simd                                   | 1.968E+00       | 0.959           | 0.495           | 1.000           |

## add N = 4M range: $[0, 65535]$ type: `uint16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 4.551E+00       | 1.000           | 0.544           | 1.029           |
| profile_add_sa                                     | 8.364E+00       | 1.838           | 1.000           | 1.891           |
| profile_add_simd                                   | 4.424E+00       | 0.972           | 0.529           | 1.000           |

## add N = 4M range: $[0, 4294967295]$ type: `uint32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 1.061E+01       | 1.000           | 0.927           | 1.022           |
| profile_add_sa                                     | 1.144E+01       | 1.078           | 1.000           | 1.102           |
| profile_add_simd                                   | 1.039E+01       | 0.979           | 0.908           | 1.000           |

## add N = 4M range: $[0, 18446744073709551615]$ type: `uint64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 2.264E+01       | 1.000           | 0.583           | 1.061           |
| profile_add_sa                                     | 3.886E+01       | 1.716           | 1.000           | 1.821           |
| profile_add_simd                                   | 2.135E+01       | 0.943           | 0.549           | 1.000           |

## add N = 4M range: $[-128, 127]$ type: `int8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 2.078E+00       | 1.000           | 0.553           | 0.989           |
| profile_add_sa                                     | 3.758E+00       | 1.808           | 1.000           | 1.788           |
| profile_add_simd                                   | 2.101E+00       | 1.011           | 0.559           | 1.000           |

## add N = 4M range: $[-32768, 32767]$ type: `int16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 4.892E+00       | 1.000           | 0.572           | 1.080           |
| profile_add_sa                                     | 8.560E+00       | 1.750           | 1.000           | 1.889           |
| profile_add_simd                                   | 4.531E+00       | 0.926           | 0.529           | 1.000           |

## add N = 4M range: $[-2147483648, 2147483647]$ type: `int32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 1.004E+01       | 1.000           | 0.942           | 1.017           |
| profile_add_sa                                     | 1.065E+01       | 1.061           | 1.000           | 1.080           |
| profile_add_simd                                   | 9.865E+00       | 0.983           | 0.926           | 1.000           |

## add N = 4M range: $[-9223372036854775808, 9223372036854775807]$ type: `int64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 2.202E+01       | 1.000           | 0.624           | 1.031           |
| profile_add_sa                                     | 3.528E+01       | 1.602           | 1.000           | 1.652           |
| profile_add_simd                                   | 2.136E+01       | 0.970           | 0.605           | 1.000           |

## add N = 4M range: $[-4294967296.0, 4294967296.0]$ type: `float32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 1.004E+01       | 1.000           | 0.636           | 1.033           |
| profile_add_sa                                     | 1.578E+01       | 1.572           | 1.000           | 1.623           |
| profile_add_simd                                   | 9.718E+00       | 0.968           | 0.616           | 1.000           |

## add N = 4M range: $[-1.8446744073709552e+19, 1.8446744073709552e+19]$ type: `float64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_add_np                                     | 2.070E+01       | 1.000           | 0.605           | 1.026           |
| profile_add_sa                                     | 3.422E+01       | 1.653           | 1.000           | 1.696           |
| profile_add_simd                                   | 2.018E+01       | 0.975           | 0.590           | 1.000           |

</details>

<details>
<summary>Subtraction</summary>

## sub N = 4M range: $[0, 255]$ type: `uint8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 2.132E+00       | 1.000           | 0.559           | 1.032           |
| profile_sub_sa                                     | 3.817E+00       | 1.790           | 1.000           | 1.848           |
| profile_sub_simd                                   | 2.066E+00       | 0.969           | 0.541           | 1.000           |

## sub N = 4M range: $[0, 65535]$ type: `uint16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 5.011E+00       | 1.000           | 0.575           | 1.129           |
| profile_sub_sa                                     | 8.718E+00       | 1.740           | 1.000           | 1.964           |
| profile_sub_simd                                   | 4.438E+00       | 0.886           | 0.509           | 1.000           |

## sub N = 4M range: $[0, 4294967295]$ type: `uint32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 1.044E+01       | 1.000           | 0.611           | 1.042           |
| profile_sub_sa                                     | 1.709E+01       | 1.637           | 1.000           | 1.706           |
| profile_sub_simd                                   | 1.002E+01       | 0.960           | 0.586           | 1.000           |

## sub N = 4M range: $[0, 18446744073709551615]$ type: `uint64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 2.155E+01       | 1.000           | 0.626           | 1.044           |
| profile_sub_sa                                     | 3.443E+01       | 1.598           | 1.000           | 1.668           |
| profile_sub_simd                                   | 2.064E+01       | 0.958           | 0.599           | 1.000           |

## sub N = 4M range: $[-128, 127]$ type: `int8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 2.181E+00       | 1.000           | 0.572           | 1.083           |
| profile_sub_sa                                     | 3.812E+00       | 1.748           | 1.000           | 1.893           |
| profile_sub_simd                                   | 2.013E+00       | 0.923           | 0.528           | 1.000           |

## sub N = 4M range: $[-32768, 32767]$ type: `int16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 4.397E+00       | 1.000           | 0.550           | 1.006           |
| profile_sub_sa                                     | 7.997E+00       | 1.819           | 1.000           | 1.830           |
| profile_sub_simd                                   | 4.370E+00       | 0.994           | 0.546           | 1.000           |

## sub N = 4M range: $[-2147483648, 2147483647]$ type: `int32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 1.072E+01       | 1.000           | 0.593           | 1.045           |
| profile_sub_sa                                     | 1.809E+01       | 1.687           | 1.000           | 1.764           |
| profile_sub_simd                                   | 1.026E+01       | 0.957           | 0.567           | 1.000           |

## sub N = 4M range: $[-9223372036854775808, 9223372036854775807]$ type: `int64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 2.210E+01       | 1.000           | 0.597           | 1.047           |
| profile_sub_sa                                     | 3.703E+01       | 1.676           | 1.000           | 1.754           |
| profile_sub_simd                                   | 2.111E+01       | 0.955           | 0.570           | 1.000           |

## sub N = 4M range: $[-4294967296.0, 4294967296.0]$ type: `float32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 9.940E+00       | 1.000           | 0.591           | 1.004           |
| profile_sub_sa                                     | 1.682E+01       | 1.692           | 1.000           | 1.699           |
| profile_sub_simd                                   | 9.900E+00       | 0.996           | 0.589           | 1.000           |

## sub N = 4M range: $[-1.8446744073709552e+19, 1.8446744073709552e+19]$ type: `float64`

</details>

<details>
<summary>Multiplication</summary>

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_sub_np                                     | 2.144E+01       | 1.000           | 0.606           | 1.020           |
| profile_sub_sa                                     | 3.540E+01       | 1.651           | 1.000           | 1.684           |
| profile_sub_simd                                   | 2.102E+01       | 0.980           | 0.594           | 1.000           |

## mul N = 4M range: $[0, 255]$ type: `uint8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 2.175E+00       | 1.000           | 0.562           | 1.074           |
| profile_mul_sa                                     | 3.873E+00       | 1.780           | 1.000           | 1.912           |
| profile_mul_simd                                   | 2.026E+00       | 0.931           | 0.523           | 1.000           |

## mul N = 4M range: $[0, 65535]$ type: `uint16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 4.213E+00       | 1.000           | 0.545           | 0.961           |
| profile_mul_sa                                     | 7.735E+00       | 1.836           | 1.000           | 1.765           |
| profile_mul_simd                                   | 4.384E+00       | 1.040           | 0.567           | 1.000           |

## mul N = 4M range: $[0, 4294967295]$ type: `uint32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 1.006E+01       | 1.000           | 0.592           | 1.035           |
| profile_mul_sa                                     | 1.699E+01       | 1.689           | 1.000           | 1.749           |
| profile_mul_simd                                   | 9.713E+00       | 0.966           | 0.572           | 1.000           |

## mul N = 4M range: $[0, 18446744073709551615]$ type: `uint64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 2.223E+01       | 1.000           | 0.571           | 0.994           |
| profile_mul_sa                                     | 3.893E+01       | 1.752           | 1.000           | 1.742           |
| profile_mul_simd                                   | 2.236E+01       | 1.006           | 0.574           | 1.000           |

## mul N = 4M range: $[-128, 127]$ type: `int8`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 2.100E+00       | 1.000           | 0.558           | 1.084           |
| profile_mul_sa                                     | 3.761E+00       | 1.791           | 1.000           | 1.942           |
| profile_mul_simd                                   | 1.937E+00       | 0.922           | 0.515           | 1.000           |

## mul N = 4M range: $[-32768, 32767]$ type: `int16`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 4.257E+00       | 1.000           | 0.559           | 1.014           |
| profile_mul_sa                                     | 7.621E+00       | 1.790           | 1.000           | 1.815           |
| profile_mul_simd                                   | 4.199E+00       | 0.986           | 0.551           | 1.000           |

## mul N = 4M range: $[-2147483648, 2147483647]$ type: `int32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 1.015E+01       | 1.000           | 0.583           | 0.992           |
| profile_mul_sa                                     | 1.741E+01       | 1.715           | 1.000           | 1.701           |
| profile_mul_simd                                   | 1.024E+01       | 1.008           | 0.588           | 1.000           |

## mul N = 4M range: $[-9223372036854775808, 9223372036854775807]$ type: `int64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 2.298E+01       | 1.000           | 0.590           | 1.054           |
| profile_mul_sa                                     | 3.897E+01       | 1.696           | 1.000           | 1.787           |
| profile_mul_simd                                   | 2.181E+01       | 0.949           | 0.560           | 1.000           |

## mul N = 4M range: $[-4294967296.0, 4294967296.0]$ type: `float32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 9.942E+00       | 1.000           | 0.609           | 0.994           |
| profile_mul_sa                                     | 1.632E+01       | 1.641           | 1.000           | 1.632           |
| profile_mul_simd                                   | 1.000E+01       | 1.006           | 0.613           | 1.000           |

## mul N = 4M range: $[-1.8446744073709552e+19, 1.8446744073709552e+19]$ type: `float64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_mul_np                                     | 2.143E+01       | 1.000           | 0.607           | 1.034           |
| profile_mul_sa                                     | 3.533E+01       | 1.649           | 1.000           | 1.704           |
| profile_mul_simd                                   | 2.073E+01       | 0.967           | 0.587           | 1.000           |

</details>

<details>
<summary>Division</summary>

## div N = 4M range: $[-4294967296.0, 4294967296.0]$ type: `float32`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_div_np                                     | 1.047E+01       | 1.000           | 0.609           | 1.045           |
| profile_div_sa                                     | 1.718E+01       | 1.642           | 1.000           | 1.716           |
| profile_div_simd                                   | 1.001E+01       | 0.957           | 0.583           | 1.000           |

## div N = 4M range: $[-1.8446744073709552e+19, 1.8446744073709552e+19]$ type: `float64`

| func                                               | per call (ms)   | cmp to np       | cmp to sa       | cmp to simd     |
| -------------------------------------------------- | --------------- | --------------- | --------------- | --------------- |
| profile_div_np                                     | 2.048E+01       | 1.000           | 0.605           | 0.980           |
| profile_div_sa                                     | 3.385E+01       | 1.653           | 1.000           | 1.621           |
| profile_div_simd                                   | 2.088E+01       | 1.020           | 0.617           | 1.000           |

</details>


KHLee529:
@yungyuc The performance improvement for `SimpleArray` arithmetic operations is done in this PR. Please review them.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -542,14 +542,21 @@ class SimpleArrayMixinCalculators
         auto athis = static_cast<A *>(this);
         if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
         {
-            for (size_t i = 0; i < athis->size(); ++i)
+            const value_type * const end = athis->end();
```
yungyuc, line 545: I guess this is what you meant by the comment:

> I found that non-SIMD arithmetic can be twice faster when implemented using raw pointer operation instead of the original operation that always get and set the value using `SimpleArray` public interface.
> 
> The non-SIMD operation therefore is improved using raw pointer operation.

It is interesting.  I would expect `SimpleArray::data()` to have zero runtime overhead.

A follow-up task is to study the assembly between the arithmetic operations using `data()` and raw pointers to understand why there is the runtime difference.  The knowledge will be useful for improving `SimpleArray` element-wise API.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -614,22 +640,42 @@ class SimpleArrayMixinCalculators
 
     A add_simd(A const & other) const
     {
-        return A(*static_cast<A const *>(this)).iadd_simd(other);
+        A const * athis = static_cast<A const *>(this);
```
yungyuc, line 643: This is what you meant by the following summary:

> Besides, the out-of-place arithmetic operations are found to be slow due to the buffer-copy when constructing new objects. The function flows are change a bit from "Copy constructing return array -> do in-place operation" to "Constructing emtpy return objects -> call out-of-place SIMD operation with return objects set as output buffer"

I agree that you do not need the copy constructor here.  It is interesting that the copy constructor makes such a difference.  It's hard to foresee.  Profiling matters.

But the current implementation is actually "constructing empty return object -> call in-place SIMD operation".  IIUC, all of your SIMD operation helpers are implemented as in-place.

---

Please think about if we need some profiling nodes in these C++ operation functions.  If we do, they can be added in the future.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -614,22 +640,42 @@ class SimpleArrayMixinCalculators
...
+        A ret(athis->shape());
```
yungyuc, line 644: I think RVO / copy elision works for this.  I am curious whether the older way allowed copy elision:

```cpp
A(*static_cast<A const *>(this)).iadd_simd(other);
```

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -614,22 +640,42 @@ class SimpleArrayMixinCalculators
...
+
+        simd::add<T>(ret.begin(), ret.end(), athis->begin(), other.begin());
```
yungyuc, line 646: This seems to be in-place operation, but please correct me if I am wrong.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -59,26 +59,26 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     using cmpvec_t = type::vector_t<uint64_t>;
     constexpr size_t N_lane = type::vector_lane<T>;
 
-    vec_t max_vec = vdupq<T>(max_val);
-    vec_t min_vec = vdupq<T>(min_val);
+    vec_t max_vec = vdupq(max_val);
```
yungyuc, line 62: Oh, type inferencing should work we do not need the template argument.  Clever.

cpp/modmesh/simd/neon/neon_alias.hpp
yungyuc, line 1: > ### Refine neon alias implementation
> Since the SIMD implementation is first wrapping all the intrinsic functions into aliases to remove postfix from the function names, and then use the wrapped alias functions to implement template functions. However, the aliases which was defined in a standalone compilation unit actually slows down the program due to the overhead of function calls. The alias functions are therefore made into inline function with definition directly written in the header so that they can be better compiled as SIMD instructions in the program flow to decrease overhead.
> 
> Besides, alias functions which was implemented as template are converted to overloaded functions with same name to improve readability loss from strange template syntax.

It's impressive that you managed to work out this inlining work within 24 hours.

yungyuc (approved):
> # Description
> [...snipped...]

This is an excellent summary.  Thank you.

And the PR itself is also excellent.  I'll just merge.

---

In your current profiling report, the reference is NEON SIMD.  It is not a common choice of reference.  People tend to expect the reference to be an established library like numpy.

If you do not have newer profiling results, it's OK to keep the table as is to use NEON SIMD as the reference.  But if you update the results later, please change the reference to be numpy.

Future profiling report should follow the convention of using the results of an established library as the reference.

---

The current profiling results seem to be for out-of-place operations.  If you have in-place profiling results, please also add in the PR preamble comment.

If there is not in-place profiling results, please plan for add it in the future.  It is informative to compare the runtime between in-place and out-of-place operations.

---

You did not add inline annotations in the PR review.  Please add annotations in the future to guide the review process.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: Exactly, I'll try to figure out the difference later.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: Probably our definition of "in-place" is not aligned. I say this is out-of-place because the object of both addends, `athis` and `other` in this case, are not modified. If we consider the point of view from object `ret`, it is modified "in-place" exactly. 
What I consider "in-place" is similar to the expression `a += b` where `a` is added in-place. When it comes to `c = a + b` I'll say it is a out-of-place addition. 
In this function call, I consider it similar to expression `ret = this + other`. That's why I said it is an out-of-place addition.

cpp/modmesh/buffer/SimpleArray.hpp
KHLee529: I did add some profiling node here to determine whether the bottleneck is the copy-constructor or the SIMD implementation. Currently we don't have any good idea where to add profiling nodes. Let's discuss about it in the future.

KHLee529:
@yungyuc Thanks for adding the annotation in the code for me. What you added is totally correct. 

---

> In your current profiling report, the reference is NEON SIMD. It is not a common choice of reference. People tend to expect the reference to be an established library like numpy.
>
> If you do not have newer profiling results, it's OK to keep the table as is to use NEON SIMD as the reference. But if you update the results later, please change the reference to be numpy.
>
> Future profiling report should follow the convention of using the results of an established library as the reference.

I did add the result with numpy as reference, but it's not in the last column. From the result you can see 4 columns containing numbers. The left-most one is the time of function call, and the rest are the multiple taking each of the data as reference. The 2nd column from is the multiple based on numpy. Probably I should make it the right-most so that it's easy to be find.

---

> The current profiling results seem to be for out-of-place operations. If you have in-place profiling results, please also add in the PR preamble comment.
>
> If there is not in-place profiling results, please plan for add it in the future. It is informative to compare the runtime between in-place and out-of-place operations.

For this part, I consider "in-place" a operation similar to `a += b` and "out-of-place" operation similar to `c = a + b`. Although they all share the same function, but when the output buffer is marked same with one of the input buffer, I'll consider it an in-place operation, otherwise "out-of-place". Please tell me the correct definition if my definition has anything wrong.

yungyuc:
> > In your current profiling report, the reference is NEON SIMD. It is not a common choice of reference. People tend to expect the reference to be an established library like numpy.
> > <...snip...>
> 
> I did add the result with numpy as reference, but it's not in the last column. From the result you can see 4 columns containing numbers. The left-most one is the time of function call, and the rest are the multiple taking each of the data as reference. The 2nd column from is the multiple based on numpy. Probably I should make it the right-most so that it's easy to be find.

Yes you did.  It is my fault not seeing the two columns before the last one.  Nothing needs to be changed in the PR profiling report.

When putting in a presentation there should be only one obvious column, but it is another topic.

> > If there is not in-place profiling results, please plan for add it in the future. It is informative to compare the runtime between in-place and out-of-place operations.
> 
> For this part, I consider "in-place" a operation similar to `a += b` and "out-of-place" operation similar to `c = a + b`. Although they all share the same function, but when the output buffer is marked same with one of the input buffer, I'll consider it an in-place operation, otherwise "out-of-place". Please tell me the correct definition if my definition has anything wrong.

I have the same understanding.  In-place and out-of-place operations should be compared in a future report.

KHLee529:
@yungyuc I summarize the profiling result from which I make this improvement in a [HackMD Note](https://hackmd.io/@KHLee529/ryN51xHulg) since the profiling result is quite a lot. You can take a look if you're interested in it.

yungyuc:
> @yungyuc I summarize the profiling result from which I make this improvement in a [HackMD Note](https://hackmd.io/@KHLee529/ryN51xHulg) since the profiling result is quite a lot. You can take a look if you're interested in it.

It is a great material.  Based on the results, I feel you can further speed up both the non-SIMD and SIMD arithmetic operations in `SimpleArray`.

Copy @ThreeMonth03 for his awareness.

PR #577, opened 2025-09-06
Accelerate median function without axis

### LINKED ISSUE #513: Implement some statistical member functions on `SimpleArray`
Implement basic statistical member functions on `SimpleArray`.  See https://numpy.org/doc/stable/reference/routines.statistics.html for possible API design.

- [x] [`numpy.median()`](https://numpy.org/doc/stable/reference/generated/numpy.median.html#numpy.median)
- [x] [`numpy.average()`](https://numpy.org/doc/stable/reference/generated/numpy.average.html#numpy.average)
- [x] [`numpy.mean()`](https://numpy.org/doc/stable/reference/generated/numpy.mean.html#numpy.mean)
- [x] [`numpy.std()`](https://numpy.org/doc/stable/reference/generated/numpy.std.html#numpy.std)
- [x] [`numpy.var()`](https://numpy.org/doc/stable/reference/generated/numpy.var.html#numpy.var)

(PRs #525 and #539)

In addition:

- [x] Support axes.  (PR #542)
- [x] Provide performance profiling.

ThreeMonth03:
This pull request accelerate the median operation in issue #513 .

## Method
1. By default, use [lomuto_branchless_cyclic_op](https://github.com/Voultapher/sort-research-rs/blob/main/writeup/lomcyc_partition/text.md) to reduce branch miss and cache miss.

2. If the element is small (uint8, int8, bool), calculate the median with frequency.

## Result
C-contiguous array:
image

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -237,18 +237,9 @@ public:
         return result;
     }
 
-    value_type median_op(small_vector<value_type> & sv) const
-    {
-        USE_CALLPROFILER_PROFILE_THIS_SCOPE("SimpleArray::median_op()");
-        const size_t n = sv.size();
-        if (n % 2 != 0)
-        {
-            return sv.select_kth(n / 2);
-        }
-        auto v1 = sv.select_kth(n / 2 - 1);
-        auto v2 = sv.select_kth(n / 2);
-        return static_cast<value_type>(v1 + v2) / static_cast<value_type>(2.0);
-    }
+    value_type median_op(small_vector<value_type> & sv) const;
+
+    value_type median_freq(small_vector<value_type> & sv) const;
 
     A median(const small_vector<size_t> & axis) const
     {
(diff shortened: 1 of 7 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -241,13 +241,70 @@ class SimpleArrayMixinCalculators
...
+        if (n == 0)
+        {
+            throw std::runtime_error("median_freq(): empty container");
+        }
+
+        uint32_t freq[256] = {};
+
+        if constexpr (std::is_same_v<value_type, uint8_t>)
+        {
+            for (uint8_t v : sv) ++freq[v];
```
yungyuc: I would like every block to have the surrounding curly braces `{}` but they can be added in the same line like:

```cpp
for (uint8_t v : sv) { ++freq[v]; }
```

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -241,13 +241,70 @@ class SimpleArrayMixinCalculators
...
+            {
+                return v1;
+            }
+            value_type v2 = simd::max(sv.begin(), sv.begin() + n / 2);
+            value_type result = static_cast<value_type>(v1 + v2) / static_cast<value_type>(2.0);
+            return result;
+        }
+    }
+
+    value_type median_freq(small_vector<value_type> & sv) const
```
yungyuc: Please add a comment block to describe this algorithm and include a link to reference.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -734,6 +791,27 @@ class SimpleArrayMixinCalculators
         }
     }
 
+private:
+    static inline void find_two_bins(const uint32_t freq[256], size_t n, int & bin1, int & bin2)
```
yungyuc: This function is a little bit too long to be put in the class declaration.  It's better to move it outside the class declaration.

Using the `inline` specifier in a header file should be no-op.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -734,6 +791,27 @@ class SimpleArrayMixinCalculators
...
+
+        uint32_t cumulative = 0;
+        bin1 = -1;
+        bin2 = -1;
+
+        for (int i = 0; i < 256; ++i)
+        {
+            cumulative += freq[i];
+            if (bin1 < 0 && cumulative > k1)
+                bin1 = i;
```
yungyuc: Curly braces `{}`.  It's important for consistent style in the code base.

I would be good to have clang-format to enforce it, but I have not figured out how to do it.

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -312,6 +312,7 @@ class small_vector
 
     iterator choose_pivot(iterator left, iterator right)
     {
+        USE_CALLPROFILER_PROFILE_THIS_SCOPE("small_vector::choose_pivot()");
```
yungyuc: Do not add profiling nodes for a function in tight loop (like this).

If you want to have it when debugging, use a comment block to direct a maintainer to add it manually like:

```cpp
// For performance debugging, uncomment below for an additional profiling node,
// but do not turn it on by default.
//USE_CALLPROFILER_PROFILE_THIS_SCOPE("small_vector::choose_pivot()");
```

Even if the macro is only effective when the CPP macro `MODMESH_PROFILE` is define, it can be noisy in most cases.

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -329,43 +330,86 @@ class small_vector
         }
         return mid;
     }
+    // reference: https://github.com/Voultapher/sort-research-rs/blob/main/writeup/lomcyc_partition/text.md
```
yungyuc: Thanks for the reference.  It will help maintainers.

Use a blank line between functions.  Also, use [doxygen](https://www.doxygen.nl) comment style to describe a function:

```cpp
/**
 * ...
 */
```

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -397,4 +441,4 @@ static_assert(sizeof(small_vector<size_t>) == 40, "small_vector<size_t> should u
 
 } /* end namespace modmesh */
 
-/* vim: set et ts=4 sw=4: */
+/* vim: set et ts=4 sw=4: */
```
yungyuc: Please revert this unnecessary change of EOF.

profiling/profile_statistic_ops.py
```diff
@@ -27,6 +27,7 @@
 import functools
 import numpy as np
 import modmesh
+import matplotlib.pyplot as plt
```
yungyuc, line 30: Import third-party should be above `modmesh`.  The order is stdlib, third-party, modmesh.

yungyuc: Is the mpl plotting happening in the modmesh pilot window?

tests/test_buffer.py
```diff
@@ -1095,6 +1095,104 @@ def test_median_with_axis(self):
         smed = smed.ndarray
         self.assertTrue(np.array_equal(npmed, smed))
 
+    def test_median_freq(self):
```
yungyuc, line 1098: Good job.

yungyuc (changes requested):
Thanks.  Good work.  Some points to address/discuss:

- [x] Add curly braces `{}` for all blocks including one-line code, which can be kept as one line like `if (true) { return; }`.
- [x] Add comment block to describe the calculation of median using frequency.
- [x] Do not add profiling nodes for a function in tight loop like `choose_pivot()`.
- [x] Use a blank line between functions and use doxygen style to describe a function.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -573,6 +573,44 @@ SimpleArrayPlex::~SimpleArrayPlex()
     m_has_instance_ownership = false;
 }
 
+/**
+ * Find the two bins that correspond to the median values for frequency-based median calculation.
+ * This function is used for small data types (uint8_t, int8_t, bool) where frequency counting
+ * is more efficient than sorting.
+ */
```
ThreeMonth03: Write comment with doxygen format.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -573,6 +573,44 @@ SimpleArrayPlex::~SimpleArrayPlex()
...
+            break;
+        }
+    }
+}
+
+// Explicit instantiations for all SimpleArray types that use find_two_bins
+template void detail::SimpleArrayMixinCalculators<SimpleArray<bool>, bool>::find_two_bins(const uint32_t freq[256], size_t n, int & bin1, int & bin2);
+template void detail::SimpleArrayMixinCalculators<SimpleArray<int8_t>, int8_t>::find_two_bins(const uint32_t freq[256], size_t n, int & bin1, int & bin2);
+template void detail::SimpleArrayMixinCalculators<SimpleArray<uint8_t>, uint8_t>::find_two_bins(const uint32_t freq[256], size_t n, int & bin1, int & bin2);
+
```
ThreeMonth03: Move the long function to `SimpleArray.cpp`.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fix.

profiling/profile_statistic_ops.py
ThreeMonth03: Yes, you could refer [_base_app.py](https://github.com/solvcon/modmesh/blob/master/modmesh/pilot/_base_app.py).

ThreeMonth03 (commented):
@yungyuc You could review this pull request again.

ThreeMonth03:
> @yungyuc You could review this pull request again.

Oh, I miss that the CI runs failed.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -573,6 +573,44 @@ SimpleArrayPlex::~SimpleArrayPlex()
...
+        }
+        if (bin1 >= 0 && bin2 >= 0)
+        {
+            break;
+        }
+    }
+}
+
+// Explicit instantiations for all SimpleArray types that use find_two_bins
+template void detail::SimpleArrayMixinCalculators<SimpleArray<bool>, bool>::find_two_bins(const uint32_t * freq, size_t n, int & bin1, int & bin2);
```
yungyuc: Instantiating templates in an implementation (cpp) file smells bad.  It could imply something goes wrong.  But let's do it now and see what will happen.

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -330,42 +333,92 @@ class small_vector
         return mid;
     }
 
+    /**
```
yungyuc: Please move the long function (and all long functions you add in this PR) outside the class declaration.

Long function bodies in class declaration make it hard to read.

profiling/profile_statistic_ops.py
yungyuc: It is great!  It's not obvious by reading the PR, but it's an architectural issue we should not solve right now.

yungyuc (changes requested):
This is good.  While reading it again I think we should make one more change:

- [x] Move all long functions added in this PR (e.g., `small_vector::partition`) outside the class declaration.  But do not touch long functions that have already been written before this PR.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -734,8 +725,131 @@ class SimpleArrayMixinCalculators
...
+        if (n % 2 != 0)
+        {
+            return v1;
+        }
+        value_type v2 = simd::max(sv.begin(), sv.begin() + n / 2);
+        value_type result = static_cast<value_type>(v1 + v2) / static_cast<value_type>(2.0);
+        return result;
+    }
+}
+
```
ThreeMonth03, line 757: Move some function definitions out of the class.

ThreeMonth03 (commented):
@yungyuc. Thanks for your suggestion. You could review this pull request again.

yungyuc (approved):
Good

PR #578, opened 2025-09-07
Fix: Reduce the latency of the unittest

ThreeMonth03:
In this pull request, I make the sleep time shorter in pytest.


yungyuc (approved):
Now the unit tests run as fast as it should.  This PR deserves immediate merge.

PR #582, opened 2025-09-13
Migrate process_attribute precall postcall into CallProfiler

### LINKED ISSUE #560: Replace the old `TimeRegistry` with the new `CallProfiler`
The [`TimeRegistry` code in `modmesh/toggle/profile.hpp:143`](https://github.com/solvcon/modmesh/blob/a55f639735ec616bd8bd8ffdcc162d017d510cc6/cpp/modmesh/toggle/profile.hpp#L143) is an early stage prototype for the hom-brewed profiler.  It should be integrated into the new [`CallProfiler` in `modmesh/toggle/RadixTree.hpp:240`](https://github.com/solvcon/modmesh/blob/a55f639735ec616bd8bd8ffdcc162d017d510cc6/cpp/modmesh/toggle/RadixTree.hpp#L240).

`CallProfiler` should also be turn on/off during compile-time using the `MODMESH_PROFILE` macro.

`CallProfiler` should be used to profile [the Python wrapper in `modmesh/python/common.hpp:127`](https://github.com/solvcon/modmesh/blob/a55f639735ec616bd8bd8ffdcc162d017d510cc6/cpp/modmesh/python/common.hpp#L127).

c1ydehhx:
In this PR, I migrate TimeRegistry into CallProfiler to remove TimeRegistry in the future.

## How to use it?

Since `precall()` and `postcall()` will be a hook when a module that enter and exit C++ function, you may add mmtag struct for your pybind11 function. 

For example, in file `wrap_SimpleArray.cpp:L257`, add mmtag structure:
```cpp
            .def(
                "median",
                [](wrapped_type const & self)
                {
                    return self.median();
                }, modmesh::python::mmtag{})
```

## How to review it?

After you add mmtag structure, you may run `./profiling/profile_median.py`, to check new record that caused by precall and postcall is appear in profiling JSON.

```json
[
    {
        "name": "SimpleArrayUint8.median", # New record caused by precall/postcall
        "total_time": 29.891959,
        "count": 1,
        "children": [
            {
                "name": "SimpleArray::median()",
                "total_time": 29.887417,
                "count": 1,
                "children": [
                    {
                        "name": "SimpleArray::median_op()",
                        "total_time": 0.891292,
                        "count": 1,
                        "children": [
                            {
                                "name": "SimpleArray::median_freq()",
                                "total_time": 0.888875,
                                "count": 1,
                                "children": []
                            }
                        ]
                    }
                ]
            }
        ]
    }
]
```

You may check this to verify the implmentation is correct.

## Test Migration

Since we have TimeRegistry testcase to test precall/postcall work properly, I just migrate these test from TimeRegistry testcase into CallProfiler testcases.

For example, we have a testcase to check behavior with different status in `test_profile.py`.

```py
def test_status(self):

        modmesh.wrapper_profiler_status.disable()
        self.assertFalse(modmesh.wrapper_profiler_status.enabled)

        modmesh.time_registry.clear()
        buf = modmesh.ConcreteBuffer(10)
        self.assertEqual([], modmesh.time_registry.names)

        modmesh.wrapper_profiler_status.enable()
        modmesh.time_registry.clear()
        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
        self.assertEqual(
            ['ConcreteBuffer.__init__'],
            modmesh.time_registry.names)
```

Now we have new `test_status` for CallProfiler in `test_callprofiler.py`.

```py
def test_status(self):
        modmesh.wrapper_profiler_status.disable()
        self.assertFalse(modmesh.wrapper_profiler_status.enabled)

        modmesh.call_profiler.reset()
        buf = modmesh.ConcreteBuffer(10)
        result = modmesh.call_profiler.result()
        self.assertEqual({}, result)

        modmesh.wrapper_profiler_status.enable()
        modmesh.call_profiler.reset()
        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
        result = modmesh.call_profiler.result().get("children")
        names = [data["name"] for data in result]
        self.assertEqual(['ConcreteBuffer.__init__'], names)
```

### DIFF
--- a/contrib/standalone_buffer/Makefile
+++ b/contrib/standalone_buffer/Makefile
@@ -56,6 +56,12 @@ $(SETUPROOT)/simd_support.o: $(SETUPROOT)/modmesh/simd/simd_support.cpp Makefile
 $(SETUPROOT)/modbuf.o: modbuf.cpp
 	c++ $(CFLAGS) -c $< -o $@
 
+$(SETUPROOT)/profile.o: $(SETUPROOT)/modmesh/toggle/profile.cpp Makefile
+	c++ $(CFLAGS) -c $< -o $@
+
+$(SETUPROOT)/radix_tree.o: $(SETUPROOT)/modmesh/toggle/RadixTree.cpp Makefile
+	c++ $(CFLAGS) -c $< -o $@
+
 OBJS = \
 	$(SETUPROOT)/BufferExpander.o \
 	$(SETUPROOT)/SimpleArray.o \
@@ -64,6 +70,8 @@ OBJS = \
 	$(SETUPROOT)/wrap_SimpleArrayPlex.o \
 	$(SETUPROOT)/buffer_pymod.o \
 	$(SETUPROOT)/simd_support.o \
+	$(SETUPROOT)/radix_tree.o \
+	$(SETUPROOT)/profile.o \
 	$(SETUPROOT)/modbuf.o
 
 $(MODPATH): Makefile $(OBJS)
--- a/cpp/modmesh/python/common.hpp
+++ b/cpp/modmesh/python/common.hpp
@@ -124,15 +124,15 @@ struct process_attribute<modmesh::python::mmtag>
     {
         if (modmesh::python::WrapperProfilerStatus::me().enabled())
         {
-            modmesh::TimeRegistry::me().entry(get_name(call)).start();
+            modmesh::CallProfiler::instance().start_caller(get_name(call), nullptr);
         }
     }
 
-    static void postcall(function_call & call, handle &)
+    static void postcall(function_call &, handle &)
     {
         if (modmesh::python::WrapperProfilerStatus::me().enabled())
         {
-            modmesh::TimeRegistry::me().entry(get_name(call)).stop();
+            modmesh::CallProfiler::instance().end_caller();
         }
     }
 
--- a/tests/test_callprofiler.py
+++ b/tests/test_callprofiler.py
@@ -380,4 +380,45 @@ class CallProfilerTC(unittest.TestCase):
         modmesh.call_profiler.reset()
         foo()
 
+    def test_status(self):
+        modmesh.wrapper_profiler_status.disable()
+        self.assertFalse(modmesh.wrapper_profiler_status.enabled)
+
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)
+        self.assertEqual({}, modmesh.call_profiler.result())
+
+        modmesh.wrapper_profiler_status.enable()
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
+        res = modmesh.call_profiler.result().get("children")
+        self.assertEqual(['ConcreteBuffer.__init__'], [d["name"] for d in res])
+
+    def test_names(self):
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)
+        buf2 = buf.clone()  # noqa: F841
+        c = modmesh.call_profiler.result().get("children")
+        self.assertEqual(
+            ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone'],
+            [d["name"] for d in c])
+
+    def test_entry(self):
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)
+        buf2 = buf.clone()  # noqa: F841
+        buf2 = buf.clone()  # noqa: F841
+        c = modmesh.call_profiler.result().get("children")
+        init_result = list(filter(
+            lambda d: d["name"] == "ConcreteBuffer.__init__", c
+        ))[0]
+        clone_result = list(filter(
+            lambda d: d["name"] == "ConcreteBuffer.clone", c
+        ))[0]
+
+        self.assertEqual(1, init_result["count"])
+        self.assertGreater(init_result["total_time"], 0)
+        self.assertEqual(2, clone_result["count"])
+        self.assertGreater(clone_result["total_time"], 0)
+
 # vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/tests/test_profile.py
+++ b/tests/test_profile.py
@@ -98,48 +98,4 @@ class TimeRegistryTC(unittest.TestCase):
         ret = modmesh.time_registry.report()
         self.assertEqual("", ret)
 
-    def test_status(self):
-
-        modmesh.wrapper_profiler_status.disable()
-        self.assertFalse(modmesh.wrapper_profiler_status.enabled)
-
-        modmesh.time_registry.clear()
-        buf = modmesh.ConcreteBuffer(10)
-        self.assertEqual([], modmesh.time_registry.names)
-
-        modmesh.wrapper_profiler_status.enable()
-        modmesh.time_registry.clear()
-        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
-        self.assertEqual(
-            ['ConcreteBuffer.__init__'],
-            modmesh.time_registry.names)
-
-    def test_names(self):
-
-        modmesh.time_registry.clear()
-        buf = modmesh.ConcreteBuffer(10)
-        buf2 = buf.clone()  # noqa: F841
-        self.assertEqual(
-            ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone'],
-            modmesh.time_registry.names)
-
-    def test_entry(self):
-
-        modmesh.time_registry.clear()
-        buf = modmesh.ConcreteBuffer(10)
-        buf2 = buf.clone()  # noqa: F841
-        buf2 = buf.clone()  # noqa: F841
-        self.assertEqual(
-            1,
-            modmesh.time_registry.entry('ConcreteBuffer.__init__').count)
-        self.assertGreater(
-            modmesh.time_registry.entry('ConcreteBuffer.__init__').time,
-            0)
-        self.assertEqual(
-            2,
-            modmesh.time_registry.entry('ConcreteBuffer.clone').count)
-        self.assertGreater(
-            modmesh.time_registry.entry('ConcreteBuffer.clone').time,
-            0)
-
 # vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:


contrib/standalone_buffer/Makefile
```diff
@@ -56,6 +56,12 @@ $(SETUPROOT)/simd_support.o: $(SETUPROOT)/modmesh/simd/simd_support.cpp Makefile
 $(SETUPROOT)/modbuf.o: modbuf.cpp
 	c++ $(CFLAGS) -c $< -o $@
 
+$(SETUPROOT)/profile.o: $(SETUPROOT)/modmesh/toggle/profile.cpp Makefile
```
yungyuc, line 59: The standalone buffer Makefile bloats.  Is it necessary price to pay?  Or is there an (easy, or not so difficult) way to avoid the bloating.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -253,7 +253,8 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 [](wrapped_type const & self)
                 {
                     return self.median();
-                })
+                },
+                modmesh::python::mmtag{})
```
yungyuc: Remove debug code.

cpp/modmesh/python/common.hpp
```diff
@@ -36,6 +36,7 @@
 #include <atomic>
 
 #include <modmesh/toggle/toggle.hpp>
+#include <modmesh/toggle/profile.hpp>
```
yungyuc: Including the module top level `modmesh/toggle/toggle.hpp` should suffice and including `modmesh/toggle/profile.hpp` from outside the toggle module should not be necessary.  Please fix the toggle module level include file `modmesh/toggle/toggle.hpp`.

yungyuc (changes requested):
Thank you, @c1ydehhx .  This is the list of points to address:

- [x] Discuss if we have to make standalone buffer makefile longer.
- [ ] Remove debug code.
- [ ] Fix missing include in the toggle module level include file `modmesh/toggle/toggle.hpp`.
- [ ] Please squash.  A single commit suffices for this PR.

tests/test_callprofiler.py
```diff
@@ -380,4 +380,47 @@ def foo():
         modmesh.call_profiler.reset()
         foo()
 
+    def test_status(self):
```
yungyuc, line 383: Please make inline annotation for what is changed in the test code.  You moved the code and there is no diff.

contrib/standalone_buffer/Makefile
c1ydehhx: I meet the undefined symbol of CallProfiler class, so I added the profile class as dependency to build standalone buffer. I'm not really sure is it possible to avoid the bloating.

For more information, please refer [this](https://github.com/solvcon/modmesh/actions/runs/17699585300/job/50303502724#step:9:573).

tests/test_callprofiler.py
```diff
@@ -380,4 +380,47 @@ def foo():
...
+        modmesh.wrapper_profiler_status.disable()
+        self.assertFalse(modmesh.wrapper_profiler_status.enabled)
+
+        modmesh.call_profiler.reset()
```
c1ydehhx, line 387: The original part is to reset TimeRegistry. To migrate into CallProfiler tests, I reset CallProfiler.

```diff
- modmesh.time_registry.clear()
+ modmesh.call_profiler.reset()
```

tests/test_callprofiler.py
```diff
@@ -380,4 +380,47 @@ def foo():
...
+        buf = modmesh.ConcreteBuffer(10)
+        result = modmesh.call_profiler.result()
+        self.assertEqual({}, result)
```
c1ydehhx: Since TimeRegistry is return a list for result, CallProfiler return a dict for result. So the result should expected as a empty dict.
```diff
- self.assertEqual([], modmesh.time_registry.names)
+ result = modmesh.call_profiler.result()
+ self.assertEqual({}, result)
```

tests/test_callprofiler.py
```diff
@@ -380,4 +380,47 @@ def foo():
...
+
+        modmesh.wrapper_profiler_status.enable()
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
+        result = modmesh.call_profiler.result().get("children")
+        names = [data["name"] for data in result]
+        self.assertEqual(['ConcreteBuffer.__init__'], names)
```
c1ydehhx: Since TimeRegistry have `modmesh.time_registry.names` to extract all names of function, but CallProfiler doesn't. Therefore, I fetch the names in result and compare it.

```diff
-         self.assertEqual(
-             ['ConcreteBuffer.__init__'],
-             modmesh.time_registry.names)
+         result = modmesh.call_profiler.result().get("children")
+         names = [data["name"] for data in result]
+         self.assertEqual(['ConcreteBuffer.__init__'], names)

File: tests/test_callprofiler.py [hunk 8]
```diff
@@ -380,4 +380,47 @@ def foo():
...
+
+    def test_names(self):
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)
+        buf2 = buf.clone()  # noqa: F841
+        result = modmesh.call_profiler.result().get("children")
+        names = [data["name"] for data in result]
+        expected_name = ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone']
+        self.assertEqual(expected_name, names)
```
c1ydehhx: ditto.

```diff
-       self.assertEqual(
-           ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone'],
-           modmesh.time_registry.names)
+       result = modmesh.call_profiler.result().get("children")
+       names = [data["name"] for data in result]
+       expected_name = ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone']
+       self.assertEqual(expected_name, names)

File: tests/test_callprofiler.py [hunk 9]
```diff
@@ -380,4 +380,47 @@ def foo():
...
+            lambda d: d["name"] == "ConcreteBuffer.__init__", result
+        ))[0]
+        clone_result = list(filter(
+            lambda d: d["name"] == "ConcreteBuffer.clone", result)
+        )[0]
+
+        self.assertEqual(1, init_result["count"])
+        self.assertGreater(init_result["total_time"], 0)
+        self.assertEqual(2, clone_result["count"])
+        self.assertGreater(clone_result["total_time"], 0)
```
c1ydehhx: In TimeRegistry, it have a function named `entry` to fetch specific result with specific function name, but CallProfiler doesn't. Therefore, I use filter to fetch specific result.

```diff
-       modmesh.time_registry.entry('ConcreteBuffer.__init__')
+       result = modmesh.call_profiler.result().get("children")
+       init_result = list(filter(
+           lambda d: d["name"] == "ConcreteBuffer.__init__", result
+       ))[0]
```

Same, TimeRegistry can specific attribute to fetch value, but CallProfiler doesn't. Therefore, I use key to fetch value.

```diff
-       modmesh.time_registry.entry('ConcreteBuffer.__init__').count
+       init_result["count"]
```

c1ydehhx (commented):
This review is trying to describe what I change to migrate TimeRegistry precall/postcall tests into CallProfiler precall/postcall tests.

I use diff to describe what original looks like and what I changes for migrate TimeRegistry into CallProfiler. I may excerpt the part of code to explain what I changes, so that may not 100% original code.

tests/test_callprofiler.py
c1ydehhx: No problem. I added a review in below. To prevent comparison that need to check previous code repeatly, I added diff to describe what the original code looks like. Please refer [this](https://github.com/solvcon/modmesh/pull/582#pullrequestreview-3221849192).

c1ydehhx:
In this version:
- I remove `#include <modmesh/toggle/profile.hpp>` since `modmesh/toggle/toggle.hpp` already contains `modmesh/toggle/profile.hpp`. 
- I describe why I need to bloat Makefile due to undefined symbol issue.
- I add a review to describe what I changes when migrate TimeRegistry tests to CallProfiler tests.

contrib/standalone_buffer/Makefile
yungyuc: Thank you.  We will need to reorganize the code for the linkage in the future.  Let's leave it here for now.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Please remove the debug code.

cpp/modmesh/python/common.hpp
```diff
@@ -36,6 +36,7 @@
 #include <atomic>
 
 #include <modmesh/toggle/toggle.hpp>
+#include <modmesh/toggle/RadixTree.hpp>
```
yungyuc: Fix missing include in the toggle module level include file `modmesh/toggle/toggle.hpp`.  `RadixTree.hpp` should be included from `toggle.hpp`.  Do not include it here, as I commented [previously](https://github.com/solvcon/modmesh/pull/582#discussion_r2346991586).

yungyuc (changes requested):
- [x] Remove debug code.
- [x] Fix missing include in the toggle module level include file `modmesh/toggle/toggle.hpp`.
- [x] Please squash.
- [x] Don't change the style of testing code.

tests/test_callprofiler.py
yungyuc: Please revert back to the original one-liner style.  It is not necessary to change it.

tests/test_callprofiler.py
yungyuc: Don't change the style.

yungyuc (changes requested):

c1ydehhx:
In this version:
 - Remove dubug code in `cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp`.
 - Remove redundant import in `cpp/modmesh/python/common.hpp`.
 - Revert the one-liner style in tests.
 - Squash the commit.

Please review it. Thanks.

tests/test_callprofiler.py
```diff
@@ -380,4 +380,44 @@ def foo():
...
+        modmesh.wrapper_profiler_status.enable()
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)  # noqa: F841
+        res = modmesh.call_profiler.result().get("children")
+        self.assertEqual(['ConcreteBuffer.__init__'], [d["name"] for d in res])
+
+    def test_names(self):
+        modmesh.call_profiler.reset()
+        buf = modmesh.ConcreteBuffer(10)
+        buf2 = buf.clone()  # noqa: F841
```
yungyuc, line 400: Don't change the testing code style.  The original style was:
```python
        buf2 = buf.clone()  # noqa: F841
        self.assertEqual(
            ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone'],
            modmesh.time_registry.names)
```

Creating the new variable `name` changes the style unnecessarily.  Keep the style and make it like:
```python
        buf2 = buf.clone()  # noqa: F841
        res = modmesh.call_profiler.result().get("children")
        self.assertEqual(
            ['ConcreteBuffer.__init__', 'ConcreteBuffer.clone'],
            [d['name'] for d in res])
```

yungyuc (changes requested):
- [x] Don't change the style of testing code.

c1ydehhx:
Thanks for the instruction, I revert the test style. 
Please review it, thanks.

yungyuc (approved):
Looks good.

PR #583, opened 2025-09-17
Remove `print_statistics` function

### LINKED ISSUE #560: Replace the old `TimeRegistry` with the new `CallProfiler`

c1ydehhx:
## What's new?

In this PR, I would like to remove print_statistic function.

The reason why:
 - We already have statistic table builder class in this PR (#553), it can construct the same table with `add_column` function.
 - It use TimeRegistry class as dependency to construct table, and we are consider TimeRegistry class should be deprecated.

Therefore, we can remove print_statistic function.

## How to review it?

Check all CI is passed.


yungyuc (changes requested):
Please include the reasons for removing the functions in the commit log (of d83bee7fb2d22dd3bdf30052f13ee813d6b18512).

c1ydehhx:
Hi @yungyuc, I already added the reason in commit log. Please review it. Thanks.

yungyuc (approved):
LGTM

PR #585, opened 2025-09-20
Fix CI error

ThreeMonth03:
The pull request is to solve the following error in CI, which may be casued by OOM.
```console
c++: fatal error: Killed signal terminated program cc1plus\
```


yungyuc:
Thank you for the swift fix, @ThreeMonth03

PR #587, opened 2025-09-21
Add GEMM implementation and tests for SimpleArray

### LINKED ISSUE #502: Plan for native implementation of matrix-matrix operations of `SimpleArray`
[`SimpleArray`](https://github.com/solvcon/modmesh/blob/master/cpp/modmesh/buffer/SimpleArray.hpp) is an array library and it should support matrix-matrix operations.  There should be a list for what matrix-matrix operations should be implemented natively and how.

5000user5000:
Implemented the internal GEMM function in SimpleArray and added corresponding tests.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -830,14 +830,20 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
     const size_t this_ndim = athis->ndim();
     const size_t other_ndim = other.ndim();
 
-    auto format_shape = [](A const * arr) -> std::string {
+    auto format_shape = [](A const * arr) -> std::string
+    {
         Formatter shape_formatter;
-        if (arr->ndim() == 0) {
+        if (arr->ndim() == 0)
+        {
             shape_formatter << "()";
-        } else {
+        }
+        else
+        {
             shape_formatter << "(";
-            for (size_t i = 0; i < arr->ndim(); ++i) {
-                if (i > 0) shape_formatter << ",";
+            for (size_t i = 0; i < arr->ndim(); ++i)
+            {
+                if (i > 0)
+                    shape_formatter << ",";
                 shape_formatter << arr->shape(i);
             }
             shape_formatter << ")";
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -725,6 +725,44 @@ public:
         }
     }
 
+    A matmul(A const & other) const
+    {
+        auto athis = static_cast<A const *>(this);
+
+        if (athis->ndim() != 2 || other.ndim() != 2)
+        {
+            throw std::out_of_range(Formatter() << "SimpleArray::matmul(): ndim mismatch: this=" << athis->ndim()
+                                                << " other=" << other.ndim());
+        }
+
+        const size_t m = athis->shape(0);
+        const size_t k = athis->shape(1);
+        const size_t n = other.shape(1);
+
+        if (k != other.shape(0))
+        {
+            throw std::out_of_range(Formatter() << "SimpleArray::matmul(): shape mismatch: this=(" << m << "," << k
+                                                << ") other=(" << other.shape(0) << "," << n << ")");
+        }
+
+        shape_type result_shape{m, n};
+        A result(result_shape);
+        result.fill(static_cast<value_type>(0));
+
+        for (size_t i = 0; i < m; ++i)
+        {
+            for (size_t j = 0; j < n; ++j)
+            {
+                for (size_t l = 0; l < k; ++l)
+                {
+                    result(i, j) += athis->operator()(i, l) * other(l, j);
+                }
+            }
+        }
+
+        return result;
+    }
+
 private:
     static void find_two_bins(const uint32_t * freq, size_t n, int & bin1, int & bin2);
 }; /* end class SimpleArrayMixinCalculators */
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -728,39 +728,197 @@ public:
     A matmul(A const & other) const
     {
         auto athis = static_cast<A const *>(this);
+        const size_t this_ndim = athis->ndim();
+        const size_t other_ndim = other.ndim();
 
-        if (athis->ndim() != 2 || other.ndim() != 2)
+        if (this_ndim == 1 && other_ndim == 1)
         {
-            throw std::out_of_range(Formatter() << "SimpleArray::matmul(): ndim mismatch: this=" << athis->ndim()
-                                                << " other=" << other.ndim());
+            if (athis->shape(0) != other.shape(0))
+            {
+                throw std::out_of_range(Formatter() << "SimpleArray::matmul(): 1D shape mismatch: this=("
+                                                    << athis->shape(0) << ") other=(" << other.shape(0) << ")");
+            }
+
+            value_type result = static_cast<value_type>(0);
+            for (size_t i = 0; i < athis->shape(0); ++i)
+            {
+                result += athis->operator()(i) * other(i);
+            }
+
+            A scalar_result(shape_type{1});
+            scalar_result(0) = result;
+            return scalar_result;
         }
 
-        const size_t m = athis->shape(0);
-        const size_t k = athis->shape(1);
-        const size_t n = other.shape(1);
-
-        if (k != other.shape(0))
+        if (this_ndim == 1 && other_ndim == 2)
         {
-            throw std::out_of_range(Formatter() << "SimpleArray::matmul(): shape mismatch: this=(" << m << "," << k
-                                                << ") other=(" << other.shape(0) << "," << n << ")");
-        }
+            if (athis->shape(0) != other.shape(0))
+            {
+                throw std::out_of_range(Formatter() << "SimpleArray::matmul(): shape mismatch: this=("
+                                                    << athis->shape(0) << ") other=(" << other.shape(0)
+                                                    << "," << other.shape(1) << ")");
+            }
 
-        shape_type result_shape{m, n};
-        A result(result_shape);
-        result.fill(static_cast<value_type>(0));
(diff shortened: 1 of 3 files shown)


5000user5000:
Hi @yungyuc , I have finished the GEMM in SimpleArray. Please review it.

The following changes from your previous feedback:
  1. **Exception type and messages**: Changed matmul error handling from `std::runtime_error` to `std::out_of_range` with detailed shape information using `modmesh::Formatter`
  2. **Fixed test data**: Replaced random number generation in tests with fixed, hand-calculated test cases for reproducibility
  3. **Enhanced error message validation**: Updated tests to verify specific error message content using `assertRaisesRegex`

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,44 @@ class SimpleArrayMixinCalculators
...
+
+        if (athis->ndim() != 2 || other.ndim() != 2)
+        {
+            throw std::out_of_range(Formatter() << "SimpleArray::matmul(): ndim mismatch: this=" << athis->ndim()
+                                                << " other=" << other.ndim());
+        }
+
+        const size_t m = athis->shape(0);
+        const size_t k = athis->shape(1);
+        const size_t n = other.shape(1);
```
ThreeMonth03: You might consider the case of multi-dimension.

ThreeMonth03 (changes requested):
- [ ] The inner product should support multi dimension, for example:
```python
M1 = array(shape = [1,2 ,3 ,4])
N1 = array(shape = [1,2 ,4 ,7])
ret = M1 * N1 # ret.shape = [1,2,3,7]

## The following function could be implemented optionally.
M2 = array(shape = [1,2 ,3 ,4])
N2 = array(shape = [4 ,7])
ret = M2 * N2  # ret.shape = [1,2,3,7] because of the rule of broadcasting.

```

By the way, would you plan to accelerate this operation?

5000user5000:
Ok, I'll extend this PR to support multi-dimensional case as you suggested.
Regarding acceleration (blocking, BLAS, etc.), I plan to leave that for a later PR so we can focus this one on correctness and API coverage first.

5000user5000:
Hi  @ThreeMonth03 , I’ve implemented the multi-dimensional case as suggested. Please review it.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+        const size_t this_ndim = athis->ndim();
+        const size_t other_ndim = other.ndim();
+
+        if (this_ndim == 1 && other_ndim == 1)
+        {
+            if (athis->shape(0) != other.shape(0))
+            {
+                throw std::out_of_range(Formatter() << "SimpleArray::matmul(): 1D shape mismatch: this=("
+                                                    << athis->shape(0) << ") other=(" << other.shape(0) << ")");
+            }
```
ThreeMonth03: The long function should be refactored, but it could be modified in the following pull request about profiling and accelerating.

ThreeMonth03 (approved):
Thanks, @5000user5000 . This pull request is very helpful for  the implementation of kalman filter.

@yungyuc , could you please review this pull request?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
         }
     }
 
+    A matmul(A const & other) const
```
yungyuc: This new function `SimpleArrayMixinCalculators::matmul` is too long for class declaration.  Please move it outside the class template `SimpleArrayMixinCalculators` like:

```cpp
template <typename A, typename T>
A SimpleArrayMixinClaculators::matmul(A const & other) const
{
    // code
}
```

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Agree.  @5000user5000 or @ThreeMonth03 , please add a note in issue #502 to plan for breaking up the code into smaller helpers.  Perhaps we will end up with a couple of helper classes (templates).

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+    {
+        auto athis = static_cast<A const *>(this);
+        const size_t this_ndim = athis->ndim();
+        const size_t other_ndim = other.ndim();
+
+        if (this_ndim == 1 && other_ndim == 1)
```
yungyuc: This is dot product, not matrix-matrix multiplication.  Please take it to a separate function dedicated to vector dot product.

Alternate treatment is to remove this section of code and use a different issue/PR for it.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+            {
+                result += athis->operator()(i) * other(i);
+            }
+
+            A scalar_result(shape_type{1});
+            scalar_result(0) = result;
+            return scalar_result;
+        }
+
+        if (this_ndim == 1 && other_ndim == 2)
```
yungyuc: This is matrix-vector (row vector and a matrix) multiplication and should use a separate function.  (The same treatment as vector dot product.)

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+                for (size_t l = 0; l < k; ++l)
+                {
+                    result(j) += athis->operator()(l) * other(l, j);
+                }
+            }
+
+            return result;
+        }
+
+        if (this_ndim == 2 && other_ndim == 1)
```
yungyuc: Ditto.  Matrix-vector multiplication for a column vector.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+                for (size_t l = 0; l < k; ++l)
+                {
+                    result(i) += athis->operator()(i, l) * other(l);
+                }
+            }
+
+            return result;
+        }
+
+        if (this_ndim == 2 && other_ndim == 2)
```
yungyuc: This is the matrix-matrix multiplication for issue #502.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -725,6 +725,202 @@ class SimpleArrayMixinCalculators
...
+                    {
+                        result(i, j) += athis->operator()(i, l) * other(l, j);
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        if (this_ndim >= 2 && other_ndim >= 2)
```
yungyuc: This is not matrix-matrix multiplication but tensor operation.  Don't do it now.

Remove the code for the tensor contraction from this PR.

It is great to see code for the tensor contraction, but it is outside the scope of issue #502 and should use more planning.

yungyuc (changes requested):
Great work for issue #502.  Points to address:

- [x] The new function `SimpleArrayMixinCalculators::matmul` is too long for class declaration.  Please move it outside the class template `SimpleArrayMixinCalculators`.
- [x] Use a distinct function for vector dot product and matrix-vector multiplication or do them later with another issue/PR.
- [x] Remove the code for the tensor contraction from this PR.

You may want to keep the removed code in a separate branch for future work.

> Hi @ThreeMonth03 , I’ve implemented the multi-dimensional case as suggested. Please review it.

It is great to know that you are interested in more general tensor contraction.  We can do a lot toward that direction.  Let us stay with two-dimensional arrays for now and move with steady steps.

5000user5000:
Hi @yungyuc , I've finished the modification. Please review it.
- [x] Move `SimpleArrayMixinCalculators::matmul` outside the class template. 
- [x] Only support for 2D matrix-matrix multiplication.
- [x] Remove code for the tensor contraction.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -817,6 +819,52 @@ detail::SimpleArrayMixinCalculators<A, T>::median_freq(small_vector<value_type>
...
+                {
+                    result(i, j) += athis->operator()(i, l) * other(l, j);
+                }
+            }
+        }
+
+        return result;
+    }
+
+    throw std::out_of_range(Formatter() << "SimpleArray::matmul(): unsupported dimensions: this="
```
yungyuc: Include the shape of both `this` and `other` matrix, e.g., `(2, 3, 4)` and `(3, 4, 3)`.

yungyuc: I think we usually expect to see exception to be thrown in the beginning of the function instead of the end, although either logic is correct.

Please throw/return early if you do not have an argument to do it late.

yungyuc (changes requested):
Good progress.  Remaining points to address:

- [x] Include shape in the exception messages.
- [x] Consider to throw/return early.

5000user5000:
Hi @yungyuc, thanks for your suggestions. I’ve updated it. Could you please review it?
- [x] Include shape in the exception messages.
- [x] Consider to throw/return early.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -817,6 +819,73 @@ detail::SimpleArrayMixinCalculators<A, T>::median_freq(small_vector<value_type>
...
+        throw std::out_of_range(Formatter() << "SimpleArray::matmul(): unsupported dimensions: this="
+                                            << format_shape(athis) << " other=" << format_shape(&other)
+                                            << ". Only 2D x 2D matrix multiplication is supported");
+    }
+
+    const size_t m = athis->shape(0);
+    const size_t k = athis->shape(1);
+    const size_t n = other.shape(1);
+
+    if (k != other.shape(0))
```
yungyuc, line 865: Compacting the logic for making the error message can improve readability, but it is not critical.  We can leave it as it for now.

yungyuc (approved):
LGTM.

PR #592, opened 2025-09-28
Refactor: comment for MODMESH_PYSIDE6_FULL macro

### LINKED ISSUE #175: Enable `MODMESH_USE_PYSIDE` in github actions CI for ubuntu and windows
A new cmake parameter `MODMESH_USE_PYSIDE` was introduced in #161 but it was not turned on in the github actions CI for ubuntu or windows.

- [ ] Turn on `MODMESH_USE_PYSIDE` on ubuntu
- [ ] Turn on `MODMESH_USE_PYSIDE` on windows

After it is enabled on the CI, the parameter should be set default ON.

Telomelonia:


yungyuc (changes requested):
Thank you, it is a good implementation.  But building with PySide6 became a requirement since 2023 (see issue #220).  Now it is not making sense to offer a build option to turn it off.

It would make sense to explicitly say that we require PySide6 in the code base.  Maybe a cmakelists.txt could be an option.

CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ endif()
 option(BUILD_QT "build with QT" ON)
 message(STATUS "BUILD_QT: ${BUILD_QT}")
 
+# PySide6 is required since 2023 (see issue #220)
```
yungyuc: Could you please add the full issue URL?

CMakeLists.txt
```diff
@@ -138,6 +141,8 @@ set(MODMESH_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp" CACHE INTERNAL "")
 
 include_directories(${MODMESH_INCLUDE_DIR})
 
+# PySide6 is mandatory - always enabled
```
yungyuc: Since we really have a cpp macro `MODMESH_USE_PYSIDE`, it will be more explicit to add the link to the old PYSIDE-on-by-default issue there.

I thought cmakelists.txt could be a good place for the comment but I was wrong.  My apologies.

yungyuc (changes requested):
The cpp macro `MODMESH_USE_PYSIDE` is a better place for the comment.  Sorry for giving a contradicting comment previously.

yungyuc:
This is where a CPP macro of using PYSIDE is used: https://github.com/solvcon/modmesh/blob/e3cdb4e0f9b78247edcc8ea7254670ee4d37b9b8/cpp/modmesh/pilot/wrap_pilot.cpp#L39

It is not named like `MODMESH_USE_PYSIDE`.  My apologies for misremembering that too.

yungyuc (changes requested):
The comment in `wrap_pilot.cpp` will work the best.  Please revert the change in other files.

Telomelonia:
@yungyuc Ready for review, made some changes

CMakeLists.txt
```diff
@@ -138,6 +138,8 @@ set(MODMESH_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp" CACHE INTERNAL "")
 
 include_directories(${MODMESH_INCLUDE_DIR})
 
+# PySide6 is required since 2023 (see issue https://github.com/solvcon/modmesh/issues/220)
+add_compile_definitions(MODMESH_USE_PYSIDE)
```
yungyuc: This comment duplicates with that in the source code.

Telomelonia: After discussion, I removed the compile definitions for `MODMESH_USE_PYSIDE ` and comment

cpp/modmesh/pilot/wrap_pilot.cpp
```diff
@@ -36,7 +36,7 @@
 #include <QClipboard>
 #include <QMenu>
 
-// Usually MODMESH_PYSIDE6_FULL is not defined unless for debugging.
+// PySide6 is required since 2023 (see issue https://github.com/solvcon/modmesh/issues/220)
```
yungyuc, line 39: The old comment should be kept.  It should read like:

```cpp
// PySide6 is required since 2023 (see issue https://github.com/solvcon/modmesh/issues/220)
// Usually MODMESH_PYSIDE6_FULL is not defined unless for debugging.
```

yungyuc (changes requested):
Keep both old and new comments.

yungyuc:
Close for lack of update

PR #593, opened 2025-09-28
Do not delete `R3DWidget` instance explicitly from unit-testing code

yungyuc:
Deleting `R3DWidget` through `closeAndDestroy()` causes `Segmentation fault: 11` when running Python unit tests.  The error is raised through `~R3Dwidget::R3DWidget()` with a null pointer in the Qt thread cleanup code.

I am not 100% sure about the root cause which requires deeper understanding to how Qt handles Qt3d instances.  But it is certain that `R3DWidget` instances should use a parent when construction.  This PR is to fix the apparent issue.

### DIFF
--- a/cpp/modmesh/pilot/R3DWidget.cpp
+++ b/cpp/modmesh/pilot/R3DWidget.cpp
@@ -113,12 +113,6 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
     cameraController()->setFarPlane(dist + half_height * 2);
 }
 
-void R3DWidget::closeAndDestroy()
-{
-    this->close();
-    this->deleteLater();
-}
-
 void R3DWidget::resizeEvent(QResizeEvent * event)
 {
     QWidget::resizeEvent(event);
--- a/cpp/modmesh/pilot/RManager.cpp
+++ b/cpp/modmesh/pilot/RManager.cpp
@@ -86,7 +86,7 @@ R3DWidget * RManager::add3DWidget()
     R3DWidget * viewer = nullptr;
     if (m_mdiArea)
     {
-        viewer = new R3DWidget();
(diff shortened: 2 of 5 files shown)


cpp/modmesh/pilot/R3DWidget.cpp
```diff
@@ -113,12 +113,6 @@ void R3DWidget::updateWorld(std::shared_ptr<WorldFp64> const & world)
     cameraController()->setFarPlane(dist + half_height * 2);
 }
 
-void R3DWidget::closeAndDestroy()
```
yungyuc, line 116: It is hard to make sense to call `deleteLater()` from within.  Remove the helper for now.

cpp/modmesh/pilot/RManager.cpp
```diff
@@ -86,7 +86,7 @@ R3DWidget * RManager::add3DWidget()
     R3DWidget * viewer = nullptr;
     if (m_mdiArea)
     {
-        viewer = new R3DWidget();
+        viewer = new R3DWidget(/*window*/ nullptr, /*scene*/ nullptr, /*parent*/ m_mdiArea);
```
yungyuc, line 89: Add parent.  Hope this makes lifecycle correct.  It can potentially leak resources and should be reviewed in the future.

yungyuc:
@tychuang1211 Please help review.

tychuang1211:
Hi @yungyuc, thanks for pinging. Is there a unit test log/backtrace I could look at to better understand the crash? Also, how would you suggest we check whether resources are properly released?

yungyuc:
> Hi @yungyuc, thanks for pinging. Is there a unit test log/backtrace I could look at to better understand the crash? Also, how would you suggest we check whether resources are properly released?

Thank you.  It can be reproduced on macos 26 tahoe with qt 6.9.2 (and PySide 6.9.2).  I wonder if windows also show the symptom with  qt and PySide 6.9.2).

Issue #588, opened 2025-09-26
Develop code and create and/or organize unit tests for matrix operations


yungyuc:
This is to follow up issue #514.  A set of matrix operations are listed in https://github.com/solvcon/modmesh/issues/502#issuecomment-2811764852 .  To facilitate the development, supply unit tests for the features that are already available, and add placeholders for those to be implemented.

Existing features or those may be easily extended from existing array operations:

- [x] Matrix-matrix multiplication (naive) is implemented with PR #587 #596.
- [x] Element-wise matrix multiplication.
  - Issue #595.
  - PR #605.
- [x] Matrix addition and subtraction (the same as element-wise addition).
  - Array-array PR #533 
  - Array-scalar PR #605
- [x] Element-wise matrix division.
  - Array-array PR #533
  - Array-scalar PR #716
- [x] Matrix transpose.
  - PRs #509 #516.

The following items in the list will need to be developed:

- [x] Generate identity matrix.
  - Issue #590.
  - PR #591.
- [x] Matrix power.
  - Issue #821.
  - PR #844.
- [x] Matrix trace.
  - Issue #822.
  - PR #842.

Determinant is a complex task that should be tracked separately:

- Matrix determinant.
  - Issue #809 (LU decomposition) is a stepping stone.

The following items in the list will need underneath array implementation first:

- Element-wise equality test.
  - Issue #810.
  - PR #841.
- Scalar-matrix multiplication.

An item that is too big to be included in the scope of this issue:

- Matrix inversion: tracked in issue #719.

ThreeMonth03:
Hi, @5000user5000 , would you plan to implement inverse operation recently?
The algorithm of kalman filter need the inverse operation.

yungyuc:
@ThreeMonth03 thank you for supplementing the dependency information. Could you please also supplement how you like the matrix inversion to be implemented? Additional information helps planning.

I am not certain about the plan of @5000user5000 . We do not need to assume any schedule for now.

ThreeMonth03:
> [@ThreeMonth03](https://github.com/ThreeMonth03) thank you for supplementing the dependency information. Could you please also supplement how you like the matrix inversion to be implemented? Additional information helps planning.
> 
> I am not certain about the plan of [@5000user5000](https://github.com/5000user5000) . We do not need to assume any schedule for now.

The following alogrihtm is kalman filter:
### Predict
$$x_{t|t-1} = F_t x_{t-1|t-1} + B_t u_t,$$
[...]

PR #594, opened 2025-09-29
Feat: Implement LLT solver

### LINKED ISSUE #601: Develop Kalman filter on top of `SimpleArray`
Create an initial implementation of the Kalman filter based on Choleskey decomposition.

Quick notes on the alogrithm for Kalman filter:

* Predict

  $$x_{t|t-1} = F_t x_{t-1|t-1} + B_t u_t,$$
  $$P_{t|t-1} = F_t P_{t-1|t-1} F_t^\top + Q_t.$$

* Innovation

  $$y_t = z_t - H_t x_{t|t-1},$$
  $$S_t = H_t P_{t|t-1} H_t^\top + R_t.$$

* Gain (via solve)

  $$K_t = P_{t|t-1} H_t^\top S_t^{-1}.$$

* Update

  $$x_{t|t} = x_{t|t-1} + K_t y_t,$$
  $$P_{t|t} = (I-K_t H_t) P_{t|t-1} (I-K_t H_t)^\top + K_t R_t K_t^\top.$$

To make the calculation precisely, the calculation of gain should be transformed like: 

$$
S_t\,K_t^\top \;=\; \big(P_{t\mid t-1}\,H_t^\top\big)^\top.
$$

Then we would solve the formula with Cholesky decomposition (LLT):

$$
K_t = \text{solve-with-LLT}\\bigl(S_t,\; (P_{t\mid t-1} H_t^\top)^\top\bigr)^\top.
$$

The issue is created based on the comment: https://github.com/solvcon/modmesh/issues/588#issuecomment-3341283659

ThreeMonth03:
In this pull request, I implement the LLT solver, which is mentioned in the [dependency of kalman filter](https://github.com/solvcon/modmesh/issues/588).

### DIFF
--- a/cpp/modmesh/linalg/factorization.hpp
+++ b/cpp/modmesh/linalg/factorization.hpp
@@ -1,0 +1,177 @@
+#pragma once
+
+/*
+ * Copyright (c) 2025, Chun-Shih Chang <austin20463@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/buffer/SimpleArray.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+
+template <typename T>
+class Llt
+{
+
+private:
+
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+public:
+
+    // Factorization algorithm
+    // Reference: https://www.geeksforgeeks.org/dsa/cholesky-decomposition-matrix-decomposition/
+    static array_type factorize(array_type const & A);
+    // Solver: solve A x = b where A = L L^T (or L L^H for complex)
+    static array_type solve(array_type const & A, array_type const & b);
+
+private:
+
+    // Forward substitution: solve L y = b
+    static array_type forward_substitution(array_type const & L, array_type const & b);
+    // Backward substitution: solve L^T x = y (or L^H x = y for complex)
+    static array_type backward_substitution(array_type const & L, array_type const & y);
+
+}; /* end class Llt */
+
+template <typename T>
+auto Llt<T>::forward_substitution(array_type const & L, array_type const & b) -> array_type
+{
+    const size_t n = L.shape(0);
+    array_type y(n);
+    for (size_t i = 0; i < n; ++i)
+    {
+        T sum = 0;
+        for (size_t j = 0; j < i; ++j)
+        {
+            sum += L(i, j) * y(j);
+        }
+        y(i) = (b(i) - sum) / L(i, i);
+    }
+    return y;
+}
+
+template <typename T>
+auto Llt<T>::backward_substitution(array_type const & L, array_type const & y) -> array_type
+{
+    const size_t n = L.shape(0);
+    array_type x(n);
+    for (int i = n - 1; i >= 0; --i)
+    {
+        T sum = 0;
+        for (size_t j = i + 1; j < n; ++j)
+        {
+            sum += conj_mul(x(j), L(j, i));
+        }
+        x(i) = (y(i) - sum) / L(i, i);
+    }
+    return x;
+}
+
+template <typename T>
+auto Llt<T>::factorize(array_type const & A) -> array_type
+{
+    if (A.ndim() != 2 || A.shape(0) != A.shape(1))
+    {
+        throw std::invalid_argument("Llt factorization requires a square 2D matrix");
+    }
+
+    const size_t n = A.shape(0);
+    small_vector<size_t> shape = {n, n};
+    array_type L(shape, value_type(0));
+    const real_type eps = std::numeric_limits<real_type>::epsilon();
+    for (size_t i = 0; i < n; ++i)
+    {
+        for (size_t j = 0; j <= i; ++j)
+        {
+            value_type sum = 0;
+            for (size_t k = 0; k < j; ++k)
+            {
+                sum += conj_mul(L(i, k), L(j, k));
+            }
+            if (i == j)
+            {
+                real_type dr = real(A(j, j) - sum);
+                real_type tol = std::max<real_type>(1, abs(A(j, j))) * 100 * eps;
+                if (dr <= tol)
+                {
+                    throw std::runtime_error("Cholesky failed: matrix not (numerically) SPD.");
+                }
+                L(j, j) = std::sqrt(dr);
+            }
+            else
+            {
+                L(i, j) = (A(i, j) - sum) / L(j, j);
+            }
+        }
+    }
+    return L;
+}
+
+template <typename T>
+auto Llt<T>::solve(array_type const & A, array_type const & b) -> array_type
+{
+    if (A.ndim() != 2 || A.shape(0) != A.shape(1))
+    {
+        throw std::invalid_argument("Matrix A must be square 2D");
+    }
+    if (b.ndim() != 1)
+    {
+        throw std::invalid_argument("Vector b must be 1D");
+    }
+    if (A.shape(0) != b.shape(0))
+    {
+        throw std::invalid_argument("Matrix A and vector b dimension mismatch");
+    }
+    array_type L = factorize(A);
+    array_type y = forward_substitution(L, b);
+    array_type x = backward_substitution(L, y);
+    return x;
+}
+
+} /* end namespace detail */
+
+template <typename T>
+SimpleArray<T> llt_factorization(SimpleArray<T> const & A)
+{
+    return detail::Llt<T>::factorize(A);
+}
+
+template <typename T>
+SimpleArray<T> llt_solve(SimpleArray<T> const & A, SimpleArray<T> const & b)
+{
+    return detail::Llt<T>::solve(A, b);
+}
+
+} /* end namespace modmesh */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/linalg/pymod/wrap_linalg.cpp
+++ b/cpp/modmesh/linalg/pymod/wrap_linalg.cpp
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2025, Chun-Shih Chang <austin20463@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/linalg/pymod/linalg_pymod.hpp>
+
+namespace modmesh
+{
+
+namespace python
+{
+
+template <typename T>
+void def_llt_factorization(pybind11::module & mod)
+{
+    mod.def(
+        "llt_factorization", [](SimpleArray<T> const & A)
+        { return llt_factorization(A); },
+        pybind11::arg("A"));
+}
+
+template <typename T>
+void def_llt_solve(pybind11::module & mod)
+{
+    mod.def(
+        "llt_solve", [](SimpleArray<T> const & A, SimpleArray<T> const & b)
+        { return llt_solve(A, b); },
+        pybind11::arg("A"),
+        pybind11::arg("b"));
+}
+
+void wrap_linalg(pybind11::module & mod)
+{
+    def_llt_factorization<double>(mod);
+    def_llt_factorization<float>(mod);
+    def_llt_factorization<Complex<double>>(mod);
+    def_llt_factorization<Complex<float>>(mod);
+
+    def_llt_solve<double>(mod);
+    def_llt_solve<float>(mod);
+    def_llt_solve<Complex<double>>(mod);
+    def_llt_solve<Complex<float>>(mod);
+}
+
+} /* end namespace python */
+
+} /* end namespace modmesh */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/math/math.hpp
+++ b/cpp/modmesh/math/math.hpp
@@ -42,6 +42,45 @@ inline constexpr T pow(T base)
 template <typename T>
 constexpr T pi = detail::pi_v<T>;
 
+template <typename T>
+inline constexpr T conj_mul(T const & a, T const & b)
+{
+    if constexpr (is_complex_v<T>)
+    {
+        return a * b.conj();
+    }
+    else
+    {
+        return a * b;
+    }
+}
+
+template <typename T>
+inline auto real(T const & val)
+{
+    if constexpr (is_complex_v<T>)
+    {
+        return val.real();
+    }
+    else
+    {
+        return val;
+    }
+}
+
+template <typename T>
+inline auto abs(T const & val)
+{
+    if constexpr (is_complex_v<T>)
+    {
+        return std::sqrt(val.norm());
+    }
+    else
+    {
+        return std::abs(val);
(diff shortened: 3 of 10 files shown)


cpp/modmesh/math/Complex.hpp
```diff
@@ -129,6 +129,8 @@ struct ComplexImpl
     T imag() const { return imag_v; }
     T norm() const { return real_v * real_v + imag_v * imag_v; }
     ComplexImpl<T> conj() const { return {real_v, -imag_v}; }
+
+    operator T() const { return real_v; }
```
ThreeMonth03: Add implicit conversion from complex to common type.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,98 @@
...
+private:
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+public:
+    // LLT (Cholesky) factorization algorithm
+    // Reference: https://www.geeksforgeeks.org/dsa/cholesky-decomposition-matrix-decomposition/
+    static array_type llt_factorization(array_type const & A);
+};
```
ThreeMonth03: Instead of function template, I implement the function in class template because it is more easier to manage the alias declaration.

tests/test_linalg.py
```diff
@@ -0,0 +1,124 @@
...
+        L = modmesh.Factorization.llt_factorization(A)
+        assert L.shape == (5, 5)
+        L_np = np.array(L)
+        np.testing.assert_allclose(L_np, L_desired, rtol=1e-10)
+
+    def test_llt_factorization_invalid_input(self):
+        A = modmesh.SimpleArrayFloat64([2, 3])
+        with pytest.raises(Exception):
+            modmesh.Factorization.llt_factorization(A)
+
```
ThreeMonth03, line 57: Test llt factorization.

tests/test_linalg.py
```diff
@@ -0,0 +1,124 @@
...
+            [0.0+0.0j, 0.0+0.0j, 0.0+0.0j, 0.0+0.0j, 0.5+0.0j]
+        ])
+        A_np = B_np.conj().T @ B_np + np.eye(5, dtype=complex)
+        b_np = np.array([1.0+0.0j, 2.0+0.0j, 3.0+0.0j, 4.0+0.0j, 5.0+0.0j])
+        A = modmesh.SimpleArrayComplex128(array=A_np)
+        b = modmesh.SimpleArrayComplex128(array=b_np)
+        x = modmesh.LinalgSolver.llt_solve(A, b)
+        assert x.shape == (5,)
+        x_np = np.array(x)
+        A_np = np.array(A)
```
ThreeMonth03, line 115: test llt solver

ThreeMonth03 (commented):
Hi @yungyuc , you can review this pull request when you are available.
Thank you.

cpp/modmesh/linalg/factorization.hpp
yungyuc: Please move `modmesh::FactorizationImpl` to `modmesh::detail::Factorization`.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,98 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <type_traits>
+#include <limits>
+#include <stdexcept>
+#include <modmesh/buffer/SimpleArray.hpp>
```
yungyuc, line 31: Use a space between groups of include files: STL and modmesh.

cpp/modmesh/linalg/linalg_solver.hpp
```diff
@@ -0,0 +1,115 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
```
yungyuc: Remove the duplicated `#pragma once`.

cpp/modmesh/linalg/linalg_solver.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <modmesh/buffer/SimpleArray.hpp>
+#include <modmesh/linalg/factorization.hpp>
+
+namespace modmesh
+{
+
+template <typename T>
+class LinalgSolverImpl
```
yungyuc: Rename `modmesh::LinalgSolverImpl` to `modmesh::detail:LltDecomposition`.  The class template only performs LLT decomposition for now.  Keeping it in `modmesh::detail` provides more freedom to rename in the future.  `LinalgSolver` is too broad a name that works for public API instead of the initial internal implementation detail.

cpp/modmesh/math/Complex.hpp
yungyuc: The implicit conversion from `ComplexImpl` to real number can be dangerous.  If it is not absolutely necessary, please remove it.  Let me know if this needs more discussions.

cpp/modmesh/linalg/pymod/wrap_linalg.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    friend base_type;
+
+    WrapFactorization(pybind11::module & mod, char const * pyname, char const * pydoc)
+        : base_type(mod, pyname, pydoc)
+    {
+        namespace py = pybind11;
+
+        (*this)
+            .def_static(
```
yungyuc: All API exposed currently are Python static functions of the wrapped C++ classes.  It suggests that a better way to provide the API is free frunctions.

Please create free functions in C++ for the factorization and decomposition and wrap them to Python.  In C++, the helper classes are moved to the namespace `modmesh::detail`.  Devs of both C++ and Python should also use the free function API.

yungyuc (changes requested):
It is exciting to have linear-algebraic code in modmesh.  Points to address:

- [x] Create free functions in C++ for the factorization and decomposition and wrap them to Python.
- [x] Move `modmesh::FactorizationImpl` to `modmesh::detail::Factorization`.
- [x] Move `modmesh::LinalgSolverImpl` to `modmesh::detail:LltDecomposition`.
- [x] Try to avoid the dangerous implicit conversion from `ComplexImpl` to real number.

This PR adds a lot of code in include files.  Please check if compilation time increases.

ThreeMonth03:
> It is exciting to have linear-algebraic code in modmesh. Points to address:
> 
> * [ ]  Create free functions in C++ for the factorization and decomposition and wrap them to Python.
> * [ ]  Move `modmesh::FactorizationImpl` to `modmesh::detail::Factorization`.
> * [ ]  Move `modmesh::LinalgSolverImpl` to `modmesh::detail:LltDecomposition`.
> * [ ]  Try to avoid the dangerous implicit conversion from `ComplexImpl` to real number.
> 
> This PR adds a lot of code in include files. Please check if compilation time increases.

The following example is why I use implicit conversion from `ComplexImpl` to real number:

``` cpp
auto Factorization<T>::llt_factorization(array_type const & A) -> array_type
{
    if (A.ndim() != 2 || A.shape(0) != A.shape(1))
    {
        throw std::invalid_argument("LLT factorization requires a square 2D matrix");
    }

    const size_t n = A.shape(0);
    small_vector<size_t> shape = {n, n};
    array_type L(shape, value_type(0));
    const real_type eps = std::numeric_limits<real_type>::epsilon();
    for (size_t i = 0; i < n; ++i)
    {
        for (size_t j = 0; j <= i; ++j)
        {
            value_type sum = 0;
            for (size_t k = 0; k < j; ++k)
            {
                sum += conj_mul(L(i, k), L(j, k));
            }
            if (i == j)
            {
                real_type dr = real_part(A(j, j) - sum); // The critical part
                real_type tol = std::max<real_type>(1, abs_val(A(j, j))) * 100 * eps; // The critical part
                if (dr <= tol)
                {
                    throw std::runtime_error("Cholesky failed: matrix not (numerically) SPD.");
                }
                L(j, j) = std::sqrt(dr);
            }
            else
            {
                L(i, j) = (A(i, j) - sum) / L(j, j);
            }
        }
    }
    return L;
}
``` 

In order to deal with the complex number and real number, it needs additional verbose function:
``` cpp
// Extract real part: for complex, return real component; for real, return as-is
template <typename T>
inline auto real_part(T const & val)
{
    if constexpr (is_complex_v<T>)
    {
        return val.real();
    }
    else
    {
        return val;
    }
}

// Compute absolute value: for complex, return sqrt(norm()); for real, return abs()
template <typename T>
inline auto abs_val(T const & val)
{
    if constexpr (is_complex_v<T>)
    {
        return std::sqrt(val.norm());
    }
    else
    {
        return std::abs(val);
    }
}
```

Of course, the implicit conversion is dangerous, so I will change the operator to explicit conversion.
@yungyuc How do you think?

ThreeMonth03:
> > It is exciting to have linear-algebraic code in modmesh. Points to address:
> > 
> > * [ ]   Create free functions in C++ for the factorization and decomposition and wrap them to Python.
> > * [ ]   Move `modmesh::FactorizationImpl` to `modmesh::detail::Factorization`.
> > * [ ]   Move `modmesh::LinalgSolverImpl` to `modmesh::detail:LltDecomposition`.
> > * [ ]   Try to avoid the dangerous implicit conversion from `ComplexImpl` to real number.
> > 
> > This PR adds a lot of code in include files. Please check if compilation time increases.
> 
> The following example is why I use implicit conversion from `ComplexImpl` to real number:
> 
> ```c++
> auto Factorization<T>::llt_factorization(array_type const & A) -> array_type
> {
>     if (A.ndim() != 2 || A.shape(0) != A.shape(1))
>     {
>         throw std::invalid_argument("LLT factorization requires a square 2D matrix");
>     }
> 
>     const size_t n = A.shape(0);
>     small_vector<size_t> shape = {n, n};
>     array_type L(shape, value_type(0));
>     const real_type eps = std::numeric_limits<real_type>::epsilon();
>     for (size_t i = 0; i < n; ++i)
>     {
>         for (size_t j = 0; j <= i; ++j)
>         {
>             value_type sum = 0;
>             for (size_t k = 0; k < j; ++k)
>             {
>                 sum += conj_mul(L(i, k), L(j, k));
>             }
>             if (i == j)
>             {
>                 real_type dr = real_part(A(j, j) - sum); // The critical part
>                 real_type tol = std::max<real_type>(1, abs_val(A(j, j))) * 100 * eps; // The critical part
>                 if (dr <= tol)
>                 {
>                     throw std::runtime_error("Cholesky failed: matrix not (numerically) SPD.");
>                 }
>                 L(j, j) = std::sqrt(dr);
>             }
>             else
>             {
>                 L(i, j) = (A(i, j) - sum) / L(j, j);
>             }
>         }
>     }
>     return L;
> }
> ```
> 
> In order to deal with the complex number and real number, it needs additional verbose function:
> 
> ```c++
> // Extract real part: for complex, return real component; for real, return as-is
> template <typename T>
> inline auto real_part(T const & val)
> {
>     if constexpr (is_complex_v<T>)
>     {
>         return val.real();
>     }
>     else
>     {
>         return val;
>     }
> }
> 
> // Compute absolute value: for complex, return sqrt(norm()); for real, return abs()
> template <typename T>
> inline auto abs_val(T const & val)
> {
>     if constexpr (is_complex_v<T>)
>     {
>         return std::sqrt(val.norm());
>     }
>     else
>     {
>         return std::abs(val);
>     }
> }
> ```
> 
> Of course, the implicit conversion is dangerous, so I will change the operator to explicit conversion. @yungyuc How do you think?

By the way, the explicit conversion mignt look like:
```c++
template <typename T>
auto Factorization<T>::llt_factorization(array_type const & A) -> array_type
{
    if (A.ndim() != 2 || A.shape(0) != A.shape(1))
    {
        throw std::invalid_argument("LLT factorization requires a square 2D matrix");
    }

    const size_t n = A.shape(0);
    small_vector<size_t> shape = {n, n};
    array_type L(shape, value_type(0));
    const real_type eps = std::numeric_limits<real_type>::epsilon();
    for (size_t i = 0; i < n; ++i)
    {
        for (size_t j = 0; j <= i; ++j)
        {
            value_type sum = 0;
            for (size_t k = 0; k < j; ++k)
            {
                sum += conj_mul(L(i, k), L(j, k));
            }
            if (i == j)
            {
                real_type A_jj = static_cast<real_type>(A(j, j));
                real_type dr = A_jj - static_cast<real_type>(sum);
                real_type tol = std::max<real_type>(1, std::abs(A_jj)) * 100 * eps;
                if (dr <= tol)
                {
                    throw std::runtime_error("Cholesky failed: matrix not (numerically) SPD.");
                }
                L(j, j) = static_cast<value_type>(std::sqrt(dr));
            }
            else
            {
                L(i, j) = (A(i, j) - sum) / L(j, j);
            }
        }
    }
    return L;
}
```

ThreeMonth03 (commented):
Hi, @yungyuc.
I have removed the unecessary header, wrap the free function and remove the dangereous casting.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+private:
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+public:
+    // LLT (Cholesky) factorization algorithm
+    // Reference: https://www.geeksforgeeks.org/dsa/cholesky-decomposition-matrix-decomposition/
+    static array_type llt_factorization(array_type const & A);
```
yungyuc: Why naming it `Factorization::llt_factorization()`?  Is there a reason to duplicate the word factorization?

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,103 @@
...
+};
```
yungyuc: Need an ending comment marker `/* end class Factorization */`.

cpp/modmesh/linalg/linalg_solver.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <modmesh/linalg/factorization.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+
+template <typename T>
+class LltDecomposition
```
yungyuc: Is there a reason that `Factorization` and `LltDecomposition` cannot be merged to a single class template?

cpp/modmesh/math/math.hpp
```diff
@@ -42,6 +42,45 @@ inline constexpr T pow(T base)
...
+        return a * b.conj();
+    }
+    else
+    {
+        return a * b;
+    }
+}
+
+template <typename T>
+inline auto real_part(T const & val)
```
yungyuc: modmesh currently uses its own complex number implementation.  It was intentional to get the most control for the complex number operations.  But we do not need to use a different API than [the stdlib](https://en.cppreference.com/w/cpp/numeric/complex.html).  Please match the API names of `modmesh::Complex<T>` to those of `std::complex<T>`.

cpp/modmesh/linalg/pymod/wrap_linalg.cpp
```diff
@@ -0,0 +1,76 @@
...
+void def_llt_solve(pybind11::module & mod)
+{
+    mod.def(
+        "llt_solve", [](SimpleArray<T> const & A, SimpleArray<T> const & b)
+        { return llt_solve(A, b); },
+        pybind11::arg("A"),
+        pybind11::arg("b"));
+}
+
+void wrap_linalg_solver(pybind11::module & mod)
```
yungyuc: Combine the two functions, `wrap_linalg_solver()` and `wrap_factorization()` to a single one and name it as `wrap_linalg()`.  It is over-engineering to use two wrapping functions when there are only two (sets of) functions to be wrapped.

yungyuc (changes requested):
Points to address:

- [x] Classes need an ending marker.
- [x] Discussion: Can `Factorization` and `LltDecomposition` be merged to a single class template?
- [x] Use the same API names of `std::complex<T>` for those of `modmesh::Complex<T>`.  Only apply to new API added with this PR.  Do not change API already in the code base.
- [x] Combine the two functions, `wrap_linalg_solver()` and `wrap_factorization()` to a single one and name it as `wrap_linalg()`.

The PR makes it clear that we are developing our own complex number system in modmesh.  This is what we want, and we should do it carefully.

cpp/modmesh/linalg/linalg_solver.hpp
ThreeMonth03: They could be merged to a single class template.
I just want to differentiate the process of factorization and solving matrix.

cpp/modmesh/linalg/factorization.hpp
ThreeMonth03: No.

cpp/modmesh/math/math.hpp
```diff
@@ -42,6 +42,45 @@ inline constexpr T pow(T base)
...
+    if constexpr (is_complex_v<T>)
+    {
+        return std::sqrt(val.norm());
+    }
+    else
+    {
+        return std::abs(val);
+    }
+}
+
```
ThreeMonth03, line 83: Align with std function.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+    // Solver: solve A x = b where A = L L^T (or L L^H for complex)
+    static array_type solve(array_type const & A, array_type const & b);
+
+private:
+
+    // Forward substitution: solve L y = b
+    static array_type forward_substitution(array_type const & L, array_type const & b);
+    // Backward substitution: solve L^T x = y (or L^H x = y for complex)
+    static array_type backward_substitution(array_type const & L, array_type const & y);
+
```
ThreeMonth03, line 63: Maybe these two functions could be implemented in seperate namespace/class called `triangle solver`, but they could be moved to the seperate class when more factorization algorithms are implemented.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+public:
+
+    // Factorization algorithm
+    // Reference: https://www.geeksforgeeks.org/dsa/cholesky-decomposition-matrix-decomposition/
+    static array_type factorize(array_type const & A);
+    // Solver: solve A x = b where A = L L^T (or L L^H for complex)
+    static array_type solve(array_type const & A, array_type const & b);
```
ThreeMonth03: Merge factorize and solve in a class.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+private:
+
+    // Forward substitution: solve L y = b
+    static array_type forward_substitution(array_type const & L, array_type const & b);
+    // Backward substitution: solve L^T x = y (or L^H x = y for complex)
+    static array_type backward_substitution(array_type const & L, array_type const & y);
+
+}; /* end class LLT */
```
ThreeMonth03: Add comment after class.

ThreeMonth03 (commented):
@yungyuc You could review this pull request when you are available. Thanks.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <modmesh/buffer/SimpleArray.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+
+template <typename T>
+class LLT
```
yungyuc: Acronyms should be treated like a word, so the class name should be `Llt`, not `LLT`.

cpp/modmesh/linalg/factorization.hpp
yungyuc: So you seem to agree it makes sense to use a single class template for both factorization and solve the system using the factorization.  Let me know if I am wrong.

cpp/modmesh/linalg/factorization.hpp
yungyuc: Agree.  They can stay here as private (static) member functions.

Making them private also makes sense of the use of a class here.

cpp/modmesh/linalg/factorization.hpp
yungyuc: Thank you.

It would be great if there is a linter for this, but I guess there is not one.

yungyuc (changes requested):
Thank you.  It is close to completion.  Points to address:

- [x] Acronyms should be treated like a word, so the class name should be `Llt`, not `LLT`.

I am not aware of a tool to lint for the naming convention.  But it will be good if there is one, or we write a script to do it ourself.  (This is of the topic of this PR.)

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <modmesh/buffer/SimpleArray.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+
+template <typename T>
+class Llt
```
ThreeMonth03, line 40: Fix the naming of acronyms.
I will survey the linter after implementing the algorithm.

cpp/modmesh/linalg/factorization.hpp
```diff
@@ -0,0 +1,177 @@
...
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+public:
+
+    // Factorization algorithm
+    // Reference: https://www.geeksforgeeks.org/dsa/cholesky-decomposition-matrix-decomposition/
+    static array_type factorize(array_type const & A);
+    // Solver: solve A x = b where A = L L^T (or L L^H for complex)
+    static array_type solve(array_type const & A, array_type const & b);
```
ThreeMonth03, line 55: You are right. I put them in a class just because we haven't implemented an abundant of the factorization algorithm and solver temporaily.
When we implement other factorization algorithms, we might consider dividing them into more class.

ThreeMonth03 (commented):

cpp/modmesh/linalg/factorization.hpp
yungyuc: It will be productive to bring the topic of naming linter to discord as well to let more people survey.

cpp/modmesh/linalg/factorization.hpp
yungyuc: This is a good plan.

yungyuc (approved):
LGTM

PR #596, opened 2025-10-10
Add a Python matmul/imatmul operator

### LINKED ISSUE #502: Plan for native implementation of matrix-matrix operations of `SimpleArray`

### LINKED ISSUE #588: Develop code and create and/or organize unit tests for matrix operations

5000user5000:
## Summary

This PR implements the Python matrix multiplication operators (`@` and `@=`) for the `SimpleArray`. 
This allows for a more intuitive and Pythonic syntax (`a @ b`) as a replacement for the method call (`a.matmul(b)`).

This change builds upon the `matmul` logic introduced in #587.

## Key Changes
- Added `__matmul__`, and `__imatmul__` to `SimpleArray`.
- Added corresponding unit tests to ensure operator correctness.

## Related Issues
- #502 
- #588

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -629,6 +629,8 @@ public:
         return *athis;
     }
 
+    A & imatmul(A const & other);
+
     A add_simd(A const & other) const
     {
         A const * athis = static_cast<A const *>(this);
@@ -726,8 +728,6 @@ public:
     }
 
     A matmul(A const & other) const;
-
-    A & imatmul(A const & other);
 
 private:
     static void find_two_bins(const uint32_t * freq, size_t n, int & bin1, int & bin2);
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -727,6 +727,8 @@ public:
 
     A matmul(A const & other) const;
 
+    A & imatmul(A const & other);
+
 private:
     static void find_two_bins(const uint32_t * freq, size_t n, int & bin1, int & bin2);
 }; /* end class SimpleArrayMixinCalculators */
@@ -884,6 +886,21 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
     }
 
     return result;
+}
+
+/**
+ * Perform in-place matrix multiplication for 2D arrays.
+ * This implementation supports only 2D × 2D matrix multiplication.
+ * The result replaces the content of the current array.
+ */
+template <typename A, typename T>
+A & SimpleArrayMixinCalculators<A, T>::imatmul(A const & other)
+{
+    auto athis = static_cast<A *>(this);
+    A result = athis->matmul(other);
+    *athis = std::move(result);
+
+    return *athis;
 }
 
 /**
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -315,6 +315,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
+            .def("__matmul__", &wrapped_type::matmul)
+            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
+                 {
+                     self = self.matmul(other);
+                     return self;
+                 })
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
             .def("iadd", [](wrapped_type & self, wrapped_type const & other)
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -315,10 +315,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
+            .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
+                 { self.imatmul(other); })
             .def("__matmul__", &wrapped_type::matmul)
             .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
                  {
-                     self = self.matmul(other);
+                     self.imatmul(other);
                      return self; })
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -315,15 +315,17 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
-            .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
-                 { self.imatmul(other); })
             .def("__matmul__", &wrapped_type::matmul)
-            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
-                 {
-                     self.imatmul(other);
-                     return self; })
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
+            /*
+             * Regular in-place methods (iadd, imul, etc.) are procedural calls and do
+             * NOT need to return self. However, special __i*__ methods (__iadd__,
+             * __imatmul__, etc.) MUST return self because they implement Python's
+             * augmented assignment operators (+=, @=, etc.), which work via re-assignment
+             * (e.g., a = a.__iadd__(b)).
+             * See: https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types
+             */
             .def("iadd", [](wrapped_type & self, wrapped_type const & other)
                  { self.iadd(other); })
             .def("isub", [](wrapped_type & self, wrapped_type const & other)
@@ -332,6 +334,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                  { self.imul(other); })
             .def("idiv", [](wrapped_type & self, wrapped_type const & other)
                  { self.idiv(other); })
+            .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
+                 { self.imatmul(other); })
+            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
+                 {
+                     self.imatmul(other);
+                     return self; })
             .def("add_simd", &wrapped_type::add_simd)
             .def("sub_simd", &wrapped_type::sub_simd)
             .def("mul_simd", &wrapped_type::mul_simd)
--- a/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
+++ b/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
@@ -319,8 +319,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
                  {
                      self = self.matmul(other);
-                     return self;
-                 })
+                     return self; })
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
             .def("iadd", [](wrapped_type & self, wrapped_type const & other)
--- a/tests/test_gemm.py
+++ b/tests/test_gemm.py
@@ -311,6 +311,25 @@ class GemmTestBase(mm.testing.TestBase):
         self.assertEqual(list(result.shape), [2, 2])
         np.testing.assert_array_almost_equal(result.ndarray, expected)
 
+    def test_imatmul_method(self):
+        """Test imatmul() method for in-place matrix multiplication"""
+        # Create 2x2 matrices
+        a_data = np.array([[1.0, 2.0], [3.0, 4.0]], dtype=self.dtype)
+        b_data = np.array([[5.0, 6.0], [7.0, 8.0]], dtype=self.dtype)
+
+        a = self.SimpleArray(array=a_data)
+        b = self.SimpleArray(array=b_data)
+
+        # Expected result: [[19, 22], [43, 50]]
+        expected = np.array([[19.0, 22.0], [43.0, 50.0]], dtype=self.dtype)
+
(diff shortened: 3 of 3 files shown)


5000user5000:
Hi @yungyuc, please review it when you have time.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,6 +315,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
+            .def("__matmul__", &wrapped_type::matmul)
+            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
```
yungyuc: Need an `imatmul` method as the counter part of `__imatmul__` operator.   Every matrix/array operation has a named method and `imatmul` should not be an exception.

Also need `self.imatmul` like those for other arithmetic operations.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,6 +315,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
+            .def("__matmul__", &wrapped_type::matmul)
```
yungyuc, line 318: The matrix and array operations usually take a lot of runtime and should use a profile node by using `.def_timed` instead of `.def`.  We should fix it later using issue #597.

yungyuc (changes requested):
Thank you, @5000user5000 .  The following points need to be addressed:

- [x] Need an `imatmul` method as the counter part of `__imatmul__` operator.
- [x] C++ `SimpleArray` also needs `imatmul`.

Before this point, we did not add matrix and array operators in Python because it is not obvious how they manifest in unit test code.  This PR starts to add the operators with matrix-matrix multiplication, because `@` and `@=` are very clear.

Having said that, we still should be careful about creating Python operators for matrices and arrays.  A basic guideline is that every special operator function needs to have a counter-part normal method.  The named functions provide two benefits:

1. Users can be sure about the operations they invoke by checking the function names.
2. We may provide additional but optional arguments to the named functions.  We cannot do it with the special functions for Python operators.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
5000user5000: ok,I'll implement `imatmul` method and operator for this PR.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
5000user5000: As for using `.def_timed`, I understand that it will be handled later.
So I won't  modify it in this PR.

yungyuc: Correct.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Thank you.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -886,6 +888,21 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+ */
+template <typename A, typename T>
+A & SimpleArrayMixinCalculators<A, T>::imatmul(A const & other)
+{
+    auto athis = static_cast<A *>(this);
+    A result = athis->matmul(other);
+    *athis = std::move(result);
+
+    return *athis;
+}
```
5000user5000, line 904: Added `imatmul` method.

tests/test_gemm.py
```diff
@@ -293,6 +293,61 @@ def test_unsupported_dimensions_error(self):
...
+
+        # Expected result: [[19, 22], [43, 50]]
+        expected = np.array([[19.0, 22.0], [43.0, 50.0]], dtype=self.dtype)
+
+        # Test imatmul() method
+        a.imatmul(b)
+
+        # Verify the result
+        self.assertEqual(list(a.shape), [2, 2])
+        np.testing.assert_array_almost_equal(a.ndarray, expected)
```
5000user5000, line 331: Add test for `imatmul` method

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,6 +315,13 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def("mul", &wrapped_type::mul)
             .def("div", &wrapped_type::div)
             .def("matmul", &wrapped_type::matmul)
+            .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
+                 { self.imatmul(other); })
```
5000user5000: Defined `imatmul` method as the counter part of `__imatmul__`  operator.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,6 +315,13 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            .def("__matmul__", &wrapped_type::matmul)
+            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
+                 {
+                     self.imatmul(other);
+                     return self; })
```
5000user5000: Use `self.imatmul` in  `__imatmul__` operator.

5000user5000 (commented):
Hi @yungyuc, please review it when you have time.

I've added  `imatmul` method to the C++ SimpleArray and related test.
This method is now used for both the named Python method `a.imatmul(b)` and as the implementation for the `@=` operator.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,6 +315,13 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
             .def("iadd", [](wrapped_type & self, wrapped_type const & other)
```
yungyuc, line 329: Other inplace operations (`i*` functions) did not return self.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Please clarify why do you return self while other inplace functions do not:

```cpp
            .def("iadd", [](wrapped_type & self, wrapped_type const & other)
                 { self.iadd(other); })
            .def("isub", [](wrapped_type & self, wrapped_type const & other)
                 { self.isub(other); })
            .def("imul", [](wrapped_type & self, wrapped_type const & other)
                 { self.imul(other); })
            .def("idiv", [](wrapped_type & self, wrapped_type const & other)
                 { self.idiv(other); })
```

Also, move the wrapper to be close to other inplace functions.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Move `imatmul` to be close to other inplace operation functions.

yungyuc (changes requested):
- [x] Clarify why return self for `imatmul`.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
5000user5000: You've raised a great point about why some in-place operations require a return value while others don't. My reasoning for ensuring that `__imatmul__` **must return self** is based on the fundamental distinction between an in-place assignment operator and a regular method call.

The `__i*__` methods, such as `__iadd__` and `__imatmul__`, are special methods that implement augmented assignment operators (like `+=` and `@=`). The execution model for an operation like `a @=b`  is conceptually equivalent to `a = a.__imatmul__(b)`.

Because of this re-assignment, the return value of the `__imatmul__` method is critical. If the method does not explicitly return self, it will implicitly return None. Consequently, the variable on the left-hand side (a) will be reassigned to None. This is not the intended behavior and would likely cause an AttributeError in subsequent code.

In contrast, a regular method that performs an in-place mutation, like a custom `a.some_method(b)`, is a direct procedural call. It modifies the object's state directly, and since there is no implicit re-assignment of its return value back to a, it is not required to return self for the mutation to be effective.

This behavior is specified in the Python data model documentation:
https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types

image

cpp/modmesh/buffer/SimpleArray.hpp
5000user5000: At the Python wrapper level, that's no problem. I can reorganize the code to put them together.

However, it's more difficult in `SimpleArray.hpp`. The main reason is that the other in-place functions (like `iadd`) are implemented in-line within the class, whereas I chose an **out-of-line** implementation for `imatmul`. This was done to handle its dependency on `matmul` and to follow your advice from PR #587 on keeping class declarations concise.

Because of this, it's difficult to group their implementations. The most I can do is move the declaration of `imatmul` to align with the others.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: Now it's clear.  Please add the summary (the link to the Python 3 document should be included) to the code comment to remind the future maintainers.

In line 326 a TODO comment was added for rectifying the return behavior of the inplace function:

```cpp
// TODO: In-place operation should return reference to self to support function chaining
```

By moving the wrapping code to proper locations, it will be clearer how they should behave by reading the code.  The commend summarizing the differences of `i*` and `__i*__` will be more obvious.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Moving the declaration is good enough.

yungyuc (changes requested):
- [x] Add code comment to summarize the difference between `i*` and `__i*__` functions and move the code to proper places.
- [x] Also move the code in C++ to make it look more organized.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -629,6 +629,8 @@ class SimpleArrayMixinCalculators
         return *athis;
     }
 
+    A & imatmul(A const & other);
+
```
5000user5000, line 633: move `imatmul` declaration near the other in-place methods.

cpp/modmesh/buffer/SimpleArray.hpp
5000user5000: I moved imatmul declaration near the other in-place methods.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -315,8 +315,17 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
             .def_static("eye", &wrapped_type::eye, py::arg("n"), "Create an identity matrix of size n x n")
             // TODO: In-place operation should return reference to self to support function chaining
+            /*
+             * Regular in-place methods (iadd, imul, etc.) are procedural calls and do
+             * NOT need to return self. However, special __i*__ methods (__iadd__,
+             * __imatmul__, etc.) MUST return self because they implement Python's
+             * augmented assignment operators (+=, @=, etc.), which work via re-assignment
+             * (e.g., a = a.__iadd__(b)).
+             * See: https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types
+             */
```
5000user5000, line 328: Comment to summarize the difference between `i*` and `__i*__` functions

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -325,6 +334,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                  { self.imul(other); })
             .def("idiv", [](wrapped_type & self, wrapped_type const & other)
                  { self.idiv(other); })
+            .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
+                 { self.imatmul(other); })
+            .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
+                 {
+                     self.imatmul(other);
+                     return self; })
```
5000user5000, line 342: move `imatmul` and `__imatmul__` near the other in-place mehods.

5000user5000 (commented):
- [x] Add code comment to summarize the difference between `i*` and `__i*__` functions and move the code to proper places.
 - [x]  Move the code in C++ to make it look more organized.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: This looks good.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: There can be another line of code comment to indicate the start of the section of Python operators.  But we can add it later when adding other operator wrappers.

When adding other operators, use pybind11 way: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#operator-overloading 

```cpp
#include <pybind11/operators.h>

PYBIND11_MODULE(example, m, py::mod_gil_not_used()) {
    py::class_<Vector2>(m, "Vector2")
        .def(py::init<float, float>())
        .def(py::self + py::self)
        .def(py::self += py::self)
        .def(py::self *= float())
        .def(float() * py::self)
        .def(py::self * float())
        .def(-py::self)
        .def("__repr__", &Vector2::toString);
}
```

Could you please check if pybind11 also provides its own way to overload the Python matrix-matrix multiplication operator `@`?  I do not recall one but do not trust my memory either.  (This is a separate task following up this PR.)

yungyuc (approved):
LGTM

PR #605, opened 2025-10-22
Implement element-wise addition and subtraction for `SimpleArray`

### LINKED ISSUE #588: Develop code and create and/or organize unit tests for matrix operations

5000user5000:
## Summary

This PR implements scalar-matrix addition and subtraction operations for `SimpleArray`, enabling scalar values to be added or subtracted to all array elements using `add/iadd(scalar)` and `sub/isub(scalar)` methods.

  ## Key Changes
  - Added scalar overload for `add/iadd` and `sub/isub` in `SimpleArray.hpp`
  - Updated Python bindings in `wrap_SimpleArray.cpp` for scalar addition/subtraction support
  - Added unit tests for 1D, 2D, and 3D arrays with multiple data types including boolean.

## Related Issues
- #588

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -513,9 +513,19 @@ public:
         return A(*static_cast<A const *>(this)).iadd(other);
     }
 
+    A add(value_type scalar) const
+    {
+        return A(*static_cast<A const *>(this)).iadd(scalar);
+    }
+
     A sub(A const & other) const
     {
         return A(*static_cast<A const *>(this)).isub(other);
+    }
+
+    A sub(value_type scalar) const
+    {
+        return A(*static_cast<A const *>(this)).isub(scalar);
     }
 
     A mul(A const & other) const
@@ -561,6 +571,30 @@ public:
         return *athis;
     }
 
+    A & iadd(value_type scalar)
+    {
+        auto athis = static_cast<A *>(this);
+        if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+        {
+            const value_type * const end = athis->end();
+            value_type * ptr = athis->begin();
+
+            while (ptr < end)
+            {
+                *ptr += scalar;
+                ++ptr;
+            }
+        }
+        else
+        {
+            if (scalar)
+            {
+                std::fill(athis->begin(), athis->end(), true);
+            }
+        }
+        return *athis;
+    }
+
     A & isub(A const & other)
     {
         auto athis = static_cast<A *>(this);
@@ -575,6 +609,27 @@ public:
                 *ptr -= *other_ptr;
                 ++ptr;
                 ++other_ptr;
+            }
+        }
+        else
+        {
+            throw std::runtime_error(Formatter() << "SimpleArray<bool>::isub(): boolean value doesn't support this operation");
+        }
+        return *athis;
+    }
+
+    A & isub(value_type scalar)
+    {
+        auto athis = static_cast<A *>(this);
+        if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+        {
+            const value_type * const end = athis->end();
+            value_type * ptr = athis->begin();
+
+            while (ptr < end)
+            {
+                *ptr -= scalar;
+                ++ptr;
             }
         }
         else
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -1587,7 +1587,7 @@ class SimpleArrayCalculatorsTC(unittest.TestCase):
 
     def test_add(self):
         # test integer
-        def test_add_type(type):
+        def _check_add_type(type):
             arr1 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
             arr2 = [2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
             res = [3, 6, 9, 12, 15, 18, 21, 24, 27, 30]
@@ -1605,16 +1605,16 @@ class SimpleArrayCalculatorsTC(unittest.TestCase):
                 self.assertEqual(simdres[i], res[i])
                 self.assertEqual(sres[i], nres[i])
 
-        test_add_type('int8')
-        test_add_type('int16')
-        test_add_type('int32')
-        test_add_type('int64')
-        test_add_type('uint8')
-        test_add_type('uint16')
-        test_add_type('uint32')
-        test_add_type('uint64')
-        test_add_type('float32')
-        test_add_type('float64')
+        _check_add_type('int8')
+        _check_add_type('int16')
+        _check_add_type('int32')
+        _check_add_type('int64')
+        _check_add_type('uint8')
+        _check_add_type('uint16')
+        _check_add_type('uint32')
+        _check_add_type('uint64')
+        _check_add_type('float32')
+        _check_add_type('float64')
 
         # test boolean
         arr1 = [True, True, True, False, False, False]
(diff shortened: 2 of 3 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -561,6 +566,34 @@ class SimpleArrayMixinCalculators
...
+                size_t size = athis->size();
+                for (size_t i = 0; i < size; ++i)
+                {
+                    athis->data(i) = true;
+                }
+            }
+        }
+        return *athis;
+    }
+
```
5000user5000, line 597: Extend array + scalar

5000user5000 (commented):
Hi @yungyuc, please review it when you have time.

tests/test_buffer.py
```diff
@@ -1632,6 +1632,148 @@ def test_add_type(type):
...
+            nres = np.add(narr1, narr2)
+            sres = sarr1.add(sarr2)
+            for i in range(len(narr1)):
+                self.assertEqual(sres[i], nres[i])
+
+            sarr1.iadd(sarr2)
+            for i in range(len(narr1)):
+                self.assertEqual(sarr1[i], nres[i])
+
+        test_add_1d_type('int32')
```
yungyuc: [`test_add`](https://github.com/solvcon/modmesh/blob/e2c65fc252f1131940d3547b5e82c16e31eba60f/tests/test_buffer.py#L1608) includes all types.  Please follow.

Each type uses a distinct line.  Stack trace looks nicer when we do it this way.  Do not loop them.

yungyuc (changes requested):
- [x] Test for all data types like other testing code in `test_buffer.py`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -561,6 +566,34 @@ class SimpleArrayMixinCalculators
...
+                *ptr += scalar;
+                ++ptr;
+            }
+        }
+        else
+        {
+            if (scalar)
+            {
+                size_t size = athis->size();
+                for (size_t i = 0; i < size; ++i)
```
KHLee529: What's the reason to loop the boolean array operation indexing helper function and other data type using raw pointer?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -561,6 +566,34 @@ class SimpleArrayMixinCalculators
...
+
+            while (ptr < end)
+            {
+                *ptr += scalar;
+                ++ptr;
+            }
+        }
+        else
+        {
+            if (scalar)
```
KHLee529, line 590: If the feature operation is to set the whole buffer to `true`. Is it better to maybe use `memset(ptr, 1, athis->size())`?

KHLee529 (commented):
I've left some comment on operation to boolean array. Please reply to the comments. 

- [x] Explain the difference between looping with raw pointer and with indexing helper
- [x] Examine if setting the whole buffer using `memset` a better pracatice.

tests/test_buffer.py
5000user5000: I realize I’m still missing several data types including boolean, and I’ll add them.
Regarding your comment — “Each type uses a distinct line. Stack trace looks nicer when we do it this way. Do not loop them.”
Just to confirm, do you mean writing each type explicitly like this:

 ```python
# int8
narr1 = np.array([...], dtype='int8')
narr2 = np.array([...], dtype='int8')
.... (test code)
# int16
narr1 = np.array([...], dtype='int16')
narr2 = np.array([...], dtype='int16')
... (test code)
```

 instead of calling something like:

```python
test_add_1d_type('int32')
test_add_1d_type('int64')
```

cpp/modmesh/buffer/SimpleArray.hpp
5000user5000: You're right — using raw pointers would indeed be more consistent.
The reason I wrote it this way is that I was following the implementation style in `A & imul(value_type scalar)`, which uses a similar looping approach.

cpp/modmesh/buffer/SimpleArray.hpp
5000user5000: I noticed that `memset` isn’t used anywhere else in the codebase, but `std::fill` is.
Would it be better if I used `std::fill(athis->begin(), athis->end(), true)` instead for consistency?

yungyuc: I would expect `std::fill` and `std::fill_n` to use `memset` if the latter is optimal.  In this case, `std::fill` should be more portable.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -584,6 +618,27 @@ class SimpleArrayMixinCalculators
...
+                *ptr -= scalar;
+                ++ptr;
+            }
+        }
+        else
+        {
+            throw std::runtime_error(Formatter() << "SimpleArray<bool>::isub(): boolean value doesn't support this operation");
+        }
+        return *athis;
+    }
```
5000user5000, line 640: Extend array - scalar

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -561,6 +571,30 @@ class SimpleArrayMixinCalculators
...
+                *ptr += scalar;
+                ++ptr;
+            }
+        }
+        else
+        {
+            if (scalar)
+            {
+                std::fill(athis->begin(), athis->end(), true);
+            }
```
5000user5000, line 593: usd `std::fill` instead of loop

5000user5000 (commented):
Hi @yungyuc, please review it when you have time.
- Add scalar subtraction support
- Refactor boolean addition to use `std::fill` instead of manual loop
- Expand tests to explicitly cover all datatypes (`int8/16/32/64,uint8/16/32/64, float32/64, bool`).

tests/test_buffer.py
```diff
@@ -1632,6 +1632,1137 @@ def test_add_type(type):
...
+        sarr2 = modmesh.SimpleArrayBool(array=narr2)
+        nres = np.add(narr1, narr2)
+        sres = sarr1.add(sarr2)
+        for i in range(len(narr1)):
+            self.assertEqual(sres[i], nres[i])
+        sarr1.iadd(sarr2)
+        for i in range(len(narr1)):
+            self.assertEqual(sarr1[i], nres[i])
+
+    def test_add_2d(self):
```
tigercosmos, line 1677: The test case seems to be too long. Maybe seperate into smaller ones: `test_add_2d_int8`, `test_add_2d_int16`, etc. In addition, if only the type changes, perhaps consider using macro-like function to expand the test cases (similar to `SimpleArrayPlexTC`).

yungyuc: I concur.  And the test file bloated to be many thousands of lines.  @5000user5000 please try to reduce the size and consider the points made by @tigercosmos and break the testing code into classes for organization.

Testing code is documents.

tests/test_buffer.py
yungyuc: > ```python
> test_add_1d_type('int32')
> test_add_1d_type('int64')
> ```

This style is preferred.  It is more organized and still keep stack trace.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Nice

yungyuc (changes requested):
@5000user5000 Please leave inline annotations for what you changed.  It is hard to quickly find the change in the PR.  Note that the diff is "Lines changed: 2367 additions & 6 deletions".  Helper your reviewers to save time.

One point to address:

- [x] Reduce the testing code length.

Would it be possible to use 500 lines of diff (give or take some)?  Just wondering.  Please share your thought.

If you used (AI) tools to generate the code, specify the tools and models employed, and leave annotations to explain what they did, and how you changed (withed with it).  The information helps collaborator.

Carefully inspect every lines of code, especially testing code, before requesting for review.

tests/test_buffer.py
```diff
@@ -1632,6 +1632,297 @@ def test_add_type(type):
...
+        test_add_1d_type('int8')
+        test_add_1d_type('int16')
+        test_add_1d_type('int32')
+        test_add_1d_type('int64')
+        test_add_1d_type('uint8')
+        test_add_1d_type('uint16')
+        test_add_1d_type('uint32')
+        test_add_1d_type('uint64')
+        test_add_1d_type('float32')
+        test_add_1d_type('float64')
```
5000user5000: test for all types except for boolean

tests/test_buffer.py
```diff
@@ -1632,6 +1632,297 @@ def test_add_type(type):
...
+                         dtype='bool')
+        sarr1 = modmesh.SimpleArrayBool(array=narr1)
+        sarr2 = modmesh.SimpleArrayBool(array=narr2)
+        nres = np.add(narr1, narr2)
+        sres = sarr1.add(sarr2)
+        for i in range(len(narr1)):
+            self.assertEqual(sres[i], nres[i])
+        sarr1.iadd(sarr2)
+        for i in range(len(narr1)):
+            self.assertEqual(sarr1[i], nres[i])
```
5000user5000, line 1675: boolean type tested separately

tests/test_buffer.py
```diff
@@ -1632,6 +1632,297 @@ def test_add_type(type):
...
+                         dtype='bool')
+        sarr1 = modmesh.SimpleArrayBool(array=narr1)
+        scalar = True
+        nres = np.add(narr1, scalar)
+        sres = sarr1.add(scalar)
+        for i in range(len(narr1)):
+            self.assertEqual(sres[i], nres[i])
+        sarr1.iadd(scalar)
+        for i in range(len(narr1)):
+            self.assertEqual(sarr1[i], nres[i])
```
5000user5000: add scalar test

tests/test_buffer.py
```diff
@@ -1683,6 +1974,315 @@ def test_sub_type(type):
...
+            r"SimpleArray<bool>::isub\(\): "
+            r"boolean value doesn't support this operation"
+        ):
+            sarr1.sub(sarr2)
+        with self.assertRaisesRegex(
+            RuntimeError,
+            r"SimpleArray<bool>::isub\(\): "
+            r"boolean value doesn't support this operation"
+        ):
+            sarr1.isub(sarr2)
```
5000user5000, line 2022: For sub boolean test, it tests error message.

5000user5000 (commented):
Hi @yungyuc , please review it when you have time.

Current refactor reduces the total test diff to around 600 lines.

* Reverted to the original testing style (use helper function, `add/iadd` and `sub/isub` each tested separately for 1–3D cases, plus scalar versions).
* Includes all data types as `test_add`.


Regarding your earlier comment — *“Each type uses a distinct line. Stack trace looks nicer when we do it this way. Do not loop them.”* — I might have misunderstood your intention. My current implementation follows the same helper-function style used in `test_mul`. Both I and the AI initially thought the comment meant expanding all cases explicitly, which led to the longer version before.

---

**Tools Used**

**AI assistance:** *Claude Code (Sonnet 4.5)*

* Identified refactoring patterns from existing code (`test_mul`, `test_add`).
* Generated repetitive helper function conversions.
* Helped interpret reviewer feedback and suggested possible modifications, which I reviewed before applying.
* Ran pre-push CI checks (based on my `Claude.md` setup) for `flake8`, `clang`, and `pytest` on modified functions and fixed minor issues automatically.

**Manual work:**
* Manually reviewed all generated code.

tests/test_buffer.py
```diff
@@ -1632,6 +1632,297 @@ def test_add_type(type):
             self.assertEqual(sarr1[i], res[i])
             self.assertEqual(sres[i], nres[i])
 
+    def test_add_1d(self):
+        def test_add_1d_type(type):
```
yungyuc: Rename the inner functions `test_*_type()` to be `_check_*_type()`.

The prefix `_` signifies the private nature of the (inner) helper.  The prefix `_check` signifies it is a test helper but not the test itself.

Python unit test frameworks treat functions prefixed with `test` as test functions.  Do not confuse peer devs by overloading the name unnecessarily.

yungyuc (changes requested):
> Current refactor reduces the total test diff to around 600 lines.

This is a good number.  My rule of thumb is 300-500 lines of diff in a PR, but giving and taking some is fine.

In this case, I like the current way of writing the testing code.  It's long but clear.  I don't think we can make the style significantly better.  Good work.

> Regarding your earlier comment — _“Each type uses a distinct line. Stack trace looks nicer when we do it this way. Do not loop them.”_ — I might have misunderstood your intention. My current implementation follows the same helper-function style used in `test_mul`. Both I and the AI initially thought the comment meant expanding all cases explicitly, which led to the longer version before.

Trusting tools by too much is counter-productive.

> **Tools Used**

Thanks for the tools summary.  Please add it with every update to PRs.  It helps contributors.

Points to address:

- [x] Rename the inner functions `test_*_type()` to be `_check_*_type()`.
- [x] It's close to merging.  Please squash and write concise and clear commit log.

tests/test_buffer.py
```diff
@@ -1587,7 +1587,7 @@ def type_convertor(self, dtype):
 
     def test_add(self):
         # test integer
-        def test_add_type(type):
+        def _check_add_type(type):
```
5000user5000, line 1590: Rename the inner functions `test_*_type()` to be `_check_*_type()`.

5000user5000 (commented):
Hi @yungyuc , please review it when you have time.
- [x] Rename the inner functions `test_*_type()` to be `_check_*_type()`.
- [x]  Squash and write concise and clear commit log.

yungyuc (approved):
@5000user5000 the commit log should summarize what you do as precise as possible.  Please give it a try instead of use a single line.

The code is good.

5000user5000:
Hi @yungyuc , I have written a more precise git commit log. Please review it when you have time!

yungyuc:
Thank you, @5000user5000 .

Issue #604, opened 2025-10-21
SimpleArrayComplex128[i] = numpy.complex128 fails with cast error


ThreeMonth03:
#### How to reproduce

```python
import modmesh as mm
import numpy as np

nc = np.array([1.0 + 0.1j])
mc = mm.SimpleArrayComplex128([1])
mc[0] = nc[0]          # <- fails

print(mc[0].real, mc[0].imag)
```

#### Actual

```
Traceback (most recent call last):
  ...
  mc[0] = nc[0]
RuntimeError: Unable to cast Python instance of type <class 'numpy.complex128'> to C++ type '?' 
(#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```

#### Expected

`mc[0]` should accept a `numpy.complex128` (and Python `complex`) and store it as `modmesh::Complex<double>` without extra wrapping.

#### Notes

* If we add additional initializer in `WrapComplex.cpp`:
 ```c++
            .def(
                py::init(
                    [](const std::complex<value_type> & c)
                    { return std::make_shared<wrapped_type>(wrapped_type{c.real(), c.imag()}); }),
                py::arg("np_complex"))
```

Constructing `mm.complex128` from a Python/numpy complex could work, but assigning an element to `SimpleArrayComplex128` doesn't work.
```python
  mc = mm.complex128(np.complex128(1+0.1j))  # ok
  print(mc.real, mc.imag) # 1 0.1j
  ma = mm.SimpleArrayComplex128([1]) #ok  
  mc[0] = np.complex128(1+0.1j) # not ok
```

yungyuc:
Agree that `SimpleArray` constructor should take complex-numbered `ndarray`.

@j8xixo12 , please comment.

KHLee529:
> Constructing `mm.complex128` from a Python/numpy complex could work, but assigning an element to `SimpleArrayComplex128` doesn't work.
> 
>   mc = mm.complex128(np.complex128(1+0.1j))  # ok
>   print(mc.real, mc.imag) # 1 0.1j
>   ma = mm.SimpleArrayComplex128([1]) #ok  
>   mc[0] = np.complex128(1+0.1j) # not ok

Could this be fixed by modifying the `__setattr__` helper function?

yungyuc:
> > Constructing `mm.complex128` from a Python/numpy complex could work, but assigning an element to `SimpleArrayComplex128` doesn't work.
[...]

Issue #603, opened 2025-10-21
Bug: F-contigous check in SimpleArray breaks


ThreeMonth03:
## Summary

When constructing a SimpleArray from a NumPy array, the current F-contiguous stride check is overly strict and incorrectly rejects arrays that NumPy itself considers both C and F contiguous — especially those with singleton dimensions such as (1, n) or (n, 1).

## How to reproduce the error
Code:
```python3
import numpy as np
import solvcon as sv

# Test case 1: 1x2 array - fails with Fortran contiguous check
try:
    array1 = np.array([[1.0, 0.1]])
    print(f"Test 1: shape={array1.shape}, strides={array1.strides}")
    print(f"  C_CONTIGUOUS={array1.flags.c_contiguous}, F_CONTIGUOUS={array1.flags.f_contiguous}")
    error1 = sv.SimpleArrayFloat64(array=array1)
    print("  ✓ Success")
except Exception as e:
    print(f"  ✗ Failed: {e}")

# Test case 2: 2x1 array - fails with Fortran contiguous check
try:
    array2 = np.array([[1.0], [0.0]])
    print(f"\nTest 2: shape={array2.shape}, strides={array2.strides}")
    print(f"  C_CONTIGUOUS={array2.flags.c_contiguous}, F_CONTIGUOUS={array2.flags.f_contiguous}")
    error2 = sv.SimpleArrayFloat64(array=array2)
    print("  ✓ Success")
except Exception as e:
    print(f"  ✗ Failed: {e}")

# Working solution: 2 * 2 array
try:
    array_2d = np.array([[1.0, 0.1], [0.0, 1.0]])
    print(f"\nWorking solution: 1D array shape={array_2d.shape}, strides={array_2d.strides}")
    success = sv.SimpleArrayFloat64(array=array_2d)
    print("  ✓ 2D array works")
except Exception as e:
    print(f"  ✗ 2D array failed: {e}")
```

## Root Cause

NumPy defines contiguity more flexibly:
* [Singleton dimensions (shape[i] == 1) do not constrain strides.](https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.flags.html)

The problematic code is located in SimpleArray.hpp, within the `check_f_contiguous()` function:

```c++
void check_f_contiguous(small_vector<size_t> const & shape,
                        small_vector<size_t> const & stride) const
{
    if (stride[0] != 1)
    {
[...]

PR #607, opened 2025-10-23
Implement kalman filter

### LINKED ISSUE #601: Develop Kalman filter on top of `SimpleArray`

ThreeMonth03:
According to issue #601, I implemented the Kalman filter.

By the way, since this PR contains quite a lot of code, I decided to include some temporary workarounds for certain function implementations and bug fixes. These include (but are not limited to):

 - [ ] symmetrize for 2D SimpleArray
 - [ ] hermitian for 2D SimpleArray
 - [ ] Matrix multiplication between 2D and 1D SimpleArray
 - [ ] Fixes related to issues #603 and #604
 - [ ] Unit tests for prediction — i.e., given a reasonable random sequence, the prediction should succeed

### DIFF
--- a/cpp/modmesh/linalg/kalman_filter.hpp
+++ b/cpp/modmesh/linalg/kalman_filter.hpp
@@ -1,0 +1,528 @@
+#pragma once
+
+/*
+ * Copyright (c) 2025, Chun-Shih Chang <austin20463@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/linalg/factorization.hpp>
+#include <modmesh/math/math.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+template <typename T>
+struct select_real_t;
+} /* end namespace detail */
+
+/**
+ * Reference: FilterPy KalmanFilter documentation
+ * https://filterpy.readthedocs.io/en/latest/kalman/KalmanFilter.html
+ */
+template <typename T>
+class KalmanFilter
+{
+
+public:
+
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<T>;
+
+private:
+
+    size_t m_state_size; // state dimension
+    size_t m_measurement_size; // measurement dimension
+    size_t m_control_size; // control dimension
+
+    array_type m_f; // state transition matrix (state_sizexstate_size)
+    array_type m_q; // process noise covariance (state_sizexstate_size)
+    array_type m_h; // measurement matrix (measurement_sizexstate_size)
+    array_type m_r; // measurement noise covariance (measurement_sizexmeasurement_size)
+    array_type m_p; // state covariance (state_sizexstate_size)
+    array_type m_b; // control matrix (state_sizexcontrol_size)
+    array_type m_x; // state vector (state_sizex1)
+    array_type m_i; // identity matrix (state_sizexstate_size)
+
+    real_type m_jitter; // regularization jitter for numerical stability
+
+public:
+
+    /**
+     * @brief Construct a Kalman filter.
+     *
+     * @details
+     * Creates a Kalman filter with the specified system matrices and noise
+     * parameters. The process and measurement noise covariances are created
+     * from scalar variance values (multiplied by identity matrices). The
+     * initial state covariance is set to the identity matrix.
+     *
+     * Control input: an external force or command applied to the system
+     * (e.g., thrust in aircraft, motor command in robot). If the control
+     * matrix B is empty (size 0), control input u is disabled. See
+     * https://kalmanfilter.net/multiExamples.html for examples.
+     *
+     * @param x Initial state vector.
+     * @param f State transition matrix F.
+     * @param b Control matrix B (empty to disable control input u).
+     * @param h Measurement matrix H.
+     * @param process_noise Process noise standard deviation.
+     * @param measurement_noise Measurement noise standard deviation.
+     * @param jitter Numerical stability jitter.
+     */
+    KalmanFilter(
+        array_type const & x,
+        array_type const & f,
+        array_type const & b,
+        array_type const & h,
+        real_type process_noise,
+        real_type measurement_noise,
+        real_type jitter)
+        : m_state_size(x.shape(0))
+        , m_measurement_size(h.shape(0))
+        , m_control_size((b.ndim() == 2) ? b.shape(1) : 0)
+        , m_f(f)
+        , m_q(create_noise_matrix(m_state_size, process_noise * process_noise))
+        , m_h(h)
+        , m_r(create_noise_matrix(m_measurement_size, measurement_noise * measurement_noise))
+        , m_p(array_type::eye(m_state_size))
+        , m_b(b)
+        , m_x(x)
+        , m_i(array_type::eye(m_state_size))
+        , m_jitter(jitter)
+    {
+        check_dimensions();
+    }
+
+    array_type const & state() const { return m_x; }
+    array_type const & covariance() const { return m_p; }
+
+    /**
+     * @brief Predict step without control input u.
+     *
+     * @details
+     * Advances the state using only the transition model and process noise.
+     * Represents a passive system with no external actuation.
+     *
+     * @note Common in finance, tracking, or passive simulation tasks.
+     *
+     * @see predict(const array_type&), update(const array_type&)
+     */
+    void predict()
+    {
+        // x <- F x  (m_x <- m_f @ m_x)
+        predict_state();
+
+        // P <- F P F^H + Q  (m_p <- m_f @ m_p @ m_f^H + m_q)
+        predict_covariance();
+
+        // P <- 0.5(P + P^H)  (m_p <- 0.5(m_p + m_p^H))
+        m_p = symmetrize(m_p);
+    }
+
+    /**
+     * @brief Predict step with control input u.
+     *
+     * @details
+     * Updates the state while applying a known control SimpleArray.
+     * The control term affects only the state mean.
+     *
+     * @note Example: in a vehicle model, the state may contain position and velocity,
+     * while the control input represents acceleration (throttle or brake). The
+     * control term updates how the state evolves between sensor measurements.
+     *
+     * @param u Control input.
+     *
+     * @see predict(), update(const array_type&)
+     */
+    void predict(array_type const & u)
+    {
+        check_control(u);
+
+        // x <- F x + B u  (m_x <- m_f @ m_x + m_b @ u)
+        predict_state(u);
+
+        // P <- F P F^H + Q  (m_p <- m_f @ m_p @ m_f^H + m_q)
+        predict_covariance();
+
+        // P <- 0.5(P + P^H)  (m_p <- 0.5(m_p + m_p^H))
+        m_p = symmetrize(m_p);
+    }
+
+    /**
+     * @brief Update step (measurement correction).
+     *
+     * @details
+     * Incorporates a new measurement to refine the predicted state and covariance
+     * using the Kalman gain.
+     *
+     * @param z Measurement input.
+     *
+     * @see predict(), predict(const array_type&)
+     */
+    void update(array_type const & z)
+    {
+        check_measurement(z);
+
+        // y <- z - H x  (y <- z - m_h @ m_x)
+        array_type y = innovation(z);
+
+        // S <- H P H^H + R + jitter I  (s <- m_h @ m_p @ m_h^H + m_r + m_jitter @ I)
+        array_type s = innovation_covariance();
+
+        // K <- P H^H S^{-1} via LLT solve  (k <- m_p @ m_h^H @ s^{-1})
+        array_type k = kalman_gain(s);
+
+        // x <- x + K y  (m_x <- m_x + k @ y)
+        update_state(k, y);
+
+        // P <- (I-K H) P (I-K H)^H + K R K^H  (m_p <- (I-k@m_h) @ m_p @ (I-k@m_h)^H + k @ m_r @ k^H)
+        update_covariance(k);
+    }
+
+private:
+
+    static array_type create_noise_matrix(size_t size, real_type noise_var)
+    {
+        return array_type::eye(size).mul(static_cast<T>(noise_var));
+    }
+    void check_dimensions();
+    void check_measurement(array_type const & z);
+    void check_control(array_type const & u);
+    static array_type symmetrize(array_type const & p);
+    static array_type hermitian(array_type const & a);
+
+    // Predict
+    void predict_state();
+    void predict_state(array_type const & u);
+    void predict_covariance();
+
+    // Update
+    array_type innovation(array_type const & z);
+    array_type innovation_covariance();
+    array_type kalman_gain(array_type const & s);
+    void update_state(array_type const & k, array_type const & y);
+    void update_covariance(array_type const & k);
+
+}; /* end class KalmanFilter */
+
+template <typename T>
+void KalmanFilter<T>::check_dimensions()
+{
+    if (m_f.ndim() != 2 || m_f.shape(0) != m_state_size || m_f.shape(1) != m_state_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The state transition SimpleArray f must be state_sizexstate_size, but got shape (";
+        for (ssize_t i = 0; i < m_f.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_f.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+    if (m_h.ndim() != 2 || m_h.shape(0) != m_measurement_size || m_h.shape(1) != m_state_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The measurement SimpleArray h must be measurement_sizexstate_size, but got shape (";
+        for (ssize_t i = 0; i < m_h.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_h.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+    if (m_q.ndim() != 2 || m_q.shape(0) != m_state_size || m_q.shape(1) != m_state_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The process noise covariance SimpleArray q must be state_sizexstate_size, but got shape (";
+        for (ssize_t i = 0; i < m_q.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_q.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+    if (m_r.ndim() != 2 || m_r.shape(0) != m_measurement_size || m_r.shape(1) != m_measurement_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The measurement noise covariance SimpleArray r must be measurement_sizexmeasurement_size, but got shape (";
+        for (ssize_t i = 0; i < m_r.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_r.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+    if (m_p.ndim() != 2 || m_p.shape(0) != m_state_size || m_p.shape(1) != m_state_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The state covariance SimpleArray p must be state_sizexstate_size, but got shape (";
+        for (ssize_t i = 0; i < m_p.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_p.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+    if (m_x.ndim() != 1 || m_x.shape(0) != m_state_size)
+    {
+        Formatter oss;
+        oss << "KalmanFilter::check_dimensions: The state SimpleArray x must be 1D of length state_size, but got shape (";
+        for (ssize_t i = 0; i < m_x.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                oss << ", ";
+            }
+            oss << m_x.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+
+    if (m_control_size > 0)
+    {
+        if (m_b.ndim() != 2)
+        {
+            Formatter oss;
+            oss << "KalmanFilter::check_dimensions: The control SimpleArray b must be 2D when control_size > 0, but got " << m_b.ndim() << "D";
+            throw std::invalid_argument(oss.str());
+        }
+        if (m_b.shape(0) != m_state_size || m_b.shape(1) != m_control_size)
+        {
+            Formatter oss;
+            oss << "KalmanFilter::check_dimensions: The control SimpleArray b must be state_sizexcontrol_size, but got shape (";
+            for (ssize_t i = 0; i < m_b.ndim(); ++i)
+            {
+                if (i > 0)
+                {
+                    oss << ", ";
+                }
+                oss << m_b.shape(i);
+            }
+            oss << ")";
+            throw std::invalid_argument(oss.str());
+        }
+    }
+    else
+    {
+        if (m_b.ndim() != 2 || m_b.shape(1) != 0)
+        {
+            Formatter oss;
+            oss << "KalmanFilter::check_dimensions: The control SimpleArray b must be state_sizex0 when control_size = 0, but got shape (";
+            for (ssize_t i = 0; i < m_b.ndim(); ++i)
+            {
+                if (i > 0)
+                {
+                    oss << ", ";
+                }
+                oss << m_b.shape(i);
+            }
+            oss << ")";
+            throw std::invalid_argument(oss.str());
+        }
+    }
+}
+
+template <typename T>
+void KalmanFilter<T>::predict_state()
+{
+    // x <- F x  (m_x <- m_f @ m_x)
+    array_type x_col = m_x.reshape(small_vector<size_t>{m_state_size, 1});
+    x_col = m_f.matmul(x_col);
+    m_x = x_col.reshape(small_vector<size_t>{m_state_size});
+}
+
+template <typename T>
+void KalmanFilter<T>::predict_state(array_type const & u)
+{
+    // x <- F x + B u  (m_x <- m_f @ m_x + m_b @ u)
+    array_type x_col = m_x.reshape(small_vector<size_t>{m_state_size, 1});
+    array_type u_col = u.reshape(small_vector<size_t>{m_control_size, 1});
+    x_col = m_f.matmul(x_col).add(m_b.matmul(u_col));
+    m_x = x_col.reshape(small_vector<size_t>{m_state_size});
+}
+
+template <typename T>
+void KalmanFilter<T>::predict_covariance()
+{
+    // P <- F P F^H + Q  (m_p <- m_f @ m_p @ m_f^H + m_q)
+    array_type f_h = hermitian(m_f);
+    m_p = m_f.matmul(m_p).matmul(f_h).add(m_q);
+}
+
+template <typename T>
+typename KalmanFilter<T>::array_type KalmanFilter<T>::symmetrize(array_type const & p)
+{
+    // P <- 0.5(P + P^H)  (p <- 0.5(p + p^H))
+    array_type p_h = p;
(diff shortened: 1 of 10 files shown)


cpp/modmesh/linalg/pymod/linalg_pymod.hpp
```diff
@@ -42,7 +42,8 @@ namespace python
 {
 
 void initialize_linalg(pybind11::module & mod);
-void wrap_linalg(pybind11::module & mod);
+void wrap_factorization(pybind11::module & mod);
+void wrap_kalman_filter(pybind11::module & mod);
```
ThreeMonth03, line 46: Rename from `linalg` to `factorization`.

cpp/modmesh/math/pymod/wrap_Complex.cpp
```diff
@@ -71,7 +72,12 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapComplex
                 py::init(
                     [](const wrapped_type & other)
                     { return std::make_shared<wrapped_type>(wrapped_type{other.real_v, other.imag_v}); }),
-                py::arg("other"))
+                py::arg("complex"))
+            .def(
+                py::init(
+                    [](const std::complex<value_type> & c)
+                    { return std::make_shared<wrapped_type>(wrapped_type{c.real(), c.imag()}); }),
+                py::arg("np_complex"))
```
ThreeMonth03, line 80: As comment in #604 , this initializer could make it easier to assign the complex from numpy.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,483 @@
...
+    void update(array_type const & z);
+
+private:
+
+    static array_type create_noise_matrix(size_t size, real_type noise_var);
+    void check_dimensions();
+    void check_measurement(array_type const & z);
+    void check_control(array_type const & u);
+    static array_type symmetrize(array_type const & p);
+    static array_type hermitian(array_type const & a);
```
ThreeMonth03: In numpy, [`hermitian` is supported in 2d ndarray](https://numpy.org/doc/stable/reference/generated/numpy.matrix.H.html), and `symmetrize` is not implemented.
We may discuss where to implement these operation in the following issues.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,483 @@
...
+
+template <typename T>
+auto KalmanFilterImpl<T>::innovation(array_type const & z) -> array_type
+{
+    // y ← z - H·x  (y ← z - m_h·m_x)
+    array_type x_col = m_x.reshape(small_vector<size_t>{m_state_size, 1});
+    array_type hx = m_h.matmul(x_col);
+    array_type hx_vec = hx.reshape(small_vector<size_t>{m_measurement_size});
+    return z.sub(hx_vec);
+}
```
ThreeMonth03, line 462: Currently, the function `matmul` doesn't support 2d @ 1d.

tests/test_linalg.py
```diff
@@ -234,19 +234,397 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+        h_wrong = mm.SimpleArrayFloat64([1, 1])
+
+        with self.assertRaisesRegex(
+                ValueError,
+                "KalmanFilterImpl::check_dimensions: The "
+                "measurement SimpleArray h must be "
+                "measurement_size×state_size, but got shape \\(1, 1\\)"):
+            mm.KalmanFilterFp64(
+                x=x0, f=f, h=h_wrong, process_noise=0.1, measurement_noise=1.0)
+
```
ThreeMonth03, line 282: The class tests the checker in initializer.

tests/test_linalg.py
```diff
@@ -234,19 +234,397 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+        x_sa = mm.SimpleArrayFloat64(array=x0)
+        f_sa = mm.SimpleArrayFloat64(array=f)
+        h_sa = mm.SimpleArrayFloat64([1, n])
+        kf = mm.KalmanFilterFp64(
+            x=x_sa, f=f_sa, h=h_sa,
+            process_noise=sigma_w,
+            measurement_noise=1.0,
+        )
+        kf.predict()
```
ThreeMonth03, line 350: The class tests the prediction without control input.

tests/test_linalg.py
```diff
@@ -234,19 +234,397 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+        s = h @ p_pred @ h.conj().T + r_var * np.eye(k, dtype=h.dtype)
+        k_gain = p_pred @ h.conj().T @ np.linalg.inv(s)
+        x_upd = x_pred + k_gain @ y
+        i_n = np.eye(p_pred.shape[0], dtype=p_pred.dtype)
+        a = i_n - k_gain @ h
+        p_upd = (a @ p_pred @ a.conj().T +
+                 k_gain @ (r_var * np.eye(k, dtype=h.dtype)) @
+                 k_gain.conj().T)
+        return x_upd, p_upd
+
```
ThreeMonth03, line 472: The class tests the update.

tests/test_linalg.py
```diff
@@ -234,19 +234,397 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+        h[0, 0] = 1.0
+        sigma_w = 0.05
+        q = (sigma_w**2) * np.eye(n)
+
+        x_pred_np, p_pred_np = self.kf_control_predict_numpy(
+            x0, p0, f, q, b, u
+        )
+
+        x_sa = sa_from_np(x0, mm.SimpleArrayFloat64)
+        f_sa = sa_from_np(f, mm.SimpleArrayFloat64)
```
ThreeMonth03, line 606: The class tests the prediction with control input.

ThreeMonth03 (commented):
@yungyuc You could review this pull request. Thank you.

cpp/modmesh/linalg/pymod/linalg_pymod.hpp
yungyuc: I like the informative names of the wrapping-controlling functions.

cpp/modmesh/linalg/pymod/wrap_factorization.cpp
```diff
@@ -53,7 +53,7 @@ void def_llt_solve(pybind11::module & mod)
         pybind11::arg("b"));
 }
 
-void wrap_linalg(pybind11::module & mod)
+void wrap_factorization(pybind11::module & mod)
```
yungyuc, line 56: Changing name as code goes is a good way to develop.  It is impossible to know the best naming before the code is written.  Back-n-forth changing like this is pragmatic.

cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
```diff
@@ -0,0 +1,113 @@
...
+                "update",
+                &wrapped_type::update,
+                py::arg("z"));
+    }
+
+}; /* end class WrapKalmanFilter */
+
+void wrap_kalman_filter(pybind11::module & mod)
+{
+    WrapKalmanFilter<double>::commit(mod, "KalmanFilterFp64", "Kalman Filter (double)");
```
yungyuc: Please check other code for the order of `double` and `float`.  I think we usually write smaller types and then bigger ones.

tests/test_linalg.py
```diff
@@ -234,19 +234,397 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+            measurement_noise=1.0,
+        )
+        kf.predict()
+
+        x_pred_mm = kf.state.ndarray
+        p_pred_mm = kf.covariance.ndarray
+        np.testing.assert_allclose(x_pred_mm, x_pred_np, atol=1e-12, rtol=0.0)
+        np.testing.assert_allclose(p_pred_mm, p_pred_np, atol=1e-12, rtol=0.0)
+
+# The bug is reported in issue #604
```
yungyuc: It is not clear which section of code is associated with the bug mentioned in this code comment.  Please make it clear.

tests/test_linalg.py
yungyuc: Add code comments in the testing class or function docstrings for what does "update" mean.

tests/test_linalg.py
yungyuc: Ditto.  Explain what is control input in code comments.

yungyuc (changes requested):

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: @5000user5000 please take a look at it.

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: @5000user5000 please take a look at it too.

ThreeMonth03:
> > By the way, since this PR contains quite a lot of code, I decided to include some temporary workarounds for certain function implementations and bug fixes. These include (but are not limited to):
> > 
> > * [ ]   symmetrize for 2D SimpleArray
> > * [ ]   hermitian for 2D SimpleArray
> > * [ ]   Matrix multiplication between 2D and 1D SimpleArray
> > * [ ]   Fixes related to issues [Bug: F-contigous check in SimpleArray breaks #603](https://github.com/solvcon/modmesh/issues/603) and [SimpleArrayComplex128[i] = numpy.complex128 fails with cast error #604](https://github.com/solvcon/modmesh/issues/604)
> > * [ ]   Unit tests for prediction — i.e., given a reasonable random sequence, the prediction should succeed
> 
> Why do you use check boxes? If you plan to use additional PRs for them, you should add them in the associated issue. I am confused and please clarify.
> 
> Points to address in this batch:
> 
> * [ ]  Follow the established convention for ordering wrapping code of types of different sizes.
> * [ ]  Make the testing code clear for how it is associated with the bug (issue #604) mentioned in the code comment.
> * [ ]  Add code comments in the testing class or function docstrings for what does "update" mean for the Kalman filter.
> * [ ]  Add code comments in the testing class or function docstrings for what does "control input" mean for the Kalman filter.
> 
> Testing code is development document. Write proper comments in testing code, which is usually the center of code review.

You’re right — I should’ve just used bullet points.
I mainly added checkboxes because they look visually similar and help organize the list,
but I agree that these potential follow-up items should be handled in separate PRs or issues instead of staying in this one.

cpp/modmesh/linalg/pymod/wrap_factorization.cpp
```diff
@@ -53,17 +53,17 @@ void def_llt_solve(pybind11::module & mod)
...
+    def_llt_factorization<double>(mod);
     def_llt_factorization<Complex<float>>(mod);
+    def_llt_factorization<Complex<double>>(mod);
 
-    def_llt_solve<double>(mod);
     def_llt_solve<float>(mod);
-    def_llt_solve<Complex<double>>(mod);
+    def_llt_solve<double>(mod);
     def_llt_solve<Complex<float>>(mod);
+    def_llt_solve<Complex<double>>(mod);
```
ThreeMonth03, line 66: Change the order from small to large type.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+namespace modmesh
+{
+
+namespace detail
+{
+
+/**
+ * Reference: FilterPy KalmanFilter documentation
+ * https://filterpy.readthedocs.io/en/latest/kalman/KalmanFilter.html
+ */
```
ThreeMonth03, line 46: Add the reference of the api design.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @brief Predict step without control input.
+ *
+ * @details
+ * Advances the state using only the transition model and process noise.
+ * Represents a passive system with no external actuation.
+ *
+ * @note Common in finance, tracking, or passive simulation tasks.
+ *
+ * @see predict(const array_type&), update(const array_type&)
+ */
```
ThreeMonth03: Add doxygen for predict without control.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * The control term affects only the state mean.
+ *
+ * @note Example: in a vehicle model, the state may contain position and velocity,
+ * while the control input represents acceleration (throttle or brake). The
+ * control term updates how the state evolves between sensor measurements.
+ *
+ * @param u Control input SimpleArray.
+ *
+ * @see predict(), update(const array_type&)
+ */
```
ThreeMonth03: Add doxygen for predict with control.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @details
+ * Incorporates a new measurement to refine the predicted state and covariance
+ * using the Kalman gain.
+ *
+ * @param z Measurement SimpleArray.
+ *
+ * @see predict(), predict(const array_type&)
+ */
+template <typename T>
+void KalmanFilterImpl<T>::update(array_type const & z)
```
ThreeMonth03: Add doxygen for update.

tests/test_linalg.py
```diff
@@ -234,19 +234,408 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+# This sa_from_np function provides a workaround for the F-contiguous check
+# issue by manually creating SimpleArray and copying elements instead of using
+# constructor
+def sa_from_np(arr: np.ndarray, cls):
+    arr = np.asarray(arr)
+    if arr.ndim == 1:
+        sa = cls([arr.shape[0]])
+        for i in range(arr.shape[0]):
+            sa[i] = _wrap_complex(arr[i], cls)
+        return sa
```
ThreeMonth03, line 432: Point out the related issue and workaround with comment.

ThreeMonth03 (commented):
@yungyuc Thanks for your review. You could review this pull request when you are available.

tests/test_linalg.py
```diff
@@ -234,19 +234,408 @@ def test_llt_solve_shape_mismatch_errors(self):
...
+        )
+        kf.predict()
+
+        x_pred_mm = kf.state.ndarray
+        p_pred_mm = kf.covariance.ndarray
+        np.testing.assert_allclose(x_pred_mm, x_pred_np, atol=1e-12, rtol=0.0)
+        np.testing.assert_allclose(p_pred_mm, p_pred_np, atol=1e-12, rtol=0.0)
+
+
+# The bug is reported in issue #604:
```
yungyuc, line 383: Turn the two example functions `_wrap_complex()` and `sa_from_np()` as ordinary test functions (or classes) and decorate them as [skip or expected (known) failure](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures).

Keep the comment with the new test code too.

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: This is clear.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+            *ptr = ptr->conj();
+        }
+    }
+    return out;
+}
+
+} /* end namespace detail */
+
+template <typename T>
+using KalmanFilter = detail::KalmanFilterImpl<T>;
```
yungyuc: Move `detail::KalmanFilterImpl` to `KalmanFilter`.  This alias is not needed.  The class does not need to be in the `detail` namespace either.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+    update_covariance(k);
+}
+
+template <typename T>
+auto KalmanFilterImpl<T>::create_noise_matrix(size_t size, real_type noise_var) -> array_type
+{
+    return array_type::eye(size).mul(static_cast<T>(noise_var));
+}
+
+template <typename T>
```
yungyuc, line 234: This function does not involve the `value_type` (correct me if I am wrong) and can be factored out to a helper class (or function) in an implementation (.cpp) file to save build time.

Please consider this code-quality enhancement and track it in a separate issue.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @param u Control input SimpleArray.
+ *
+ * @see predict(), update(const array_type&)
+ */
+template <typename T>
+void KalmanFilterImpl<T>::predict(array_type const & u)
+{
+    check_control(u);
+
+    // x ← F·x + B·u  (m_x ← m_f·m_x + m_b·u)
```
yungyuc: Exclusively use ASCII in source code.  Please check.

yungyuc: We may need a linter to check it automatically.  Issue #608 is created to track it.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/buffer/SimpleArray.hpp>
```
yungyuc: For a peer module, include the leading file: `#include <modmesh/buffer/buffer.hpp>`.

If you have to include individual files from another module, leave a comment above the inclusion line to give the reason.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @details
+ * Advances the state using only the transition model and process noise.
+ * Represents a passive system with no external actuation.
+ *
+ * @note Common in finance, tracking, or passive simulation tasks.
+ *
+ * @see predict(const array_type&), update(const array_type&)
+ */
+template <typename T>
+void KalmanFilterImpl<T>::predict()
```
yungyuc: This function is short and you may consider to move it in class declaration.  Keeping it outside is OK too.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * @note Example: in a vehicle model, the state may contain position and velocity,
+ * while the control input represents acceleration (throttle or brake). The
+ * control term updates how the state evolves between sensor measurements.
+ *
+ * @param u Control input SimpleArray.
+ *
+ * @see predict(), update(const array_type&)
+ */
+template <typename T>
+void KalmanFilterImpl<T>::predict(array_type const & u)
```
yungyuc: Ditto.  This is short.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+    array_type m_i; // identity matrix (state_size×state_size)
+
+    real_type m_jitter; // regularization jitter for numerical stability
+
+public:
+
+    /**
+     * Constructor: KalmanFilter(x, F, B, H, σ_w, σ_v, jitter)
+     * m_q = σ_w² I, m_r = σ_v² I, m_p = I
+     * If B is empty (size 0), control input is not supported
```
yungyuc: I am not seeing the explanation of "control input".  Please leave a code comment for what it is (an external URL is OK, but it needs to point to precisely).

yungyuc (changes requested):
> > Why do you use check boxes? If you plan to use additional PRs for them, you should add them in the associated issue. I am confused and please clarify.
> 
> You’re right — I should’ve just used bullet points. I mainly added checkboxes because they look visually similar and help organize the list, but I agree that these potential follow-up items should be handled in separate PRs or issues instead of staying in this one.

Delete irrelevant comments when replying.  It is OK to use check boxes for better visual but check them when using it that way.

Leave an additional comment to separate what are included in this PR and what will be followed up.  (It is still unclear to me.)  Use separate lists.

Additional points to address:

- [x] Turn the two example functions `_wrap_complex()` and `sa_from_np()` as ordinary test functions (or classes) with [skip or expected (known) failure](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures).
- [x] Move `detail::KalmanFilterImpl` to `KalmanFilter`.
- [x] Exclusively use ASCII in source code.
- [x] For a peer module, include the leading file.
- [x] "Control input" remains unexplained.  Leave a code comment for what it is (an external URL is OK, but it needs to point to precisely).

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+    array_type m_b; // control matrix (state_size×control_size)
+    array_type m_x; // state vector (state_size×1)
+    array_type m_i; // identity matrix (state_size×state_size)
+
+    real_type m_jitter; // regularization jitter for numerical stability
+
+public:
+
+    /**
+     * Constructor: KalmanFilter(x, F, B, H, σ_w, σ_v, jitter)
```
KHLee529: Why the constructor is not documented in doxygen style while all other public functions are? It would be nice if all the public functions keep the same documentation style. Describing the meaning of each parameters may also help users a lot.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,523 @@
...
+        array_type const & x,
+        array_type const & f,
+        array_type const & b,
+        array_type const & h,
+        real_type process_noise,
+        real_type measurement_noise,
+        real_type jitter)
+        : m_state_size(x.shape(0))
+        , m_measurement_size(h.shape(0))
+        , m_control_size((b.ndim() == 2) ? b.shape(1) : 0)
```
KHLee529, line 108: The control size is set to 0 to disable the control input but the logic here is a bit different from the comment `If B is empty (size 0), control input is not supported`. What cause this difference?

cpp/modmesh/linalg/kalman_filter.hpp
KHLee529: I concur this suggestion. I can then know the main process of each public function at the first glance.

KHLee529 (commented):
Some code and comment related to control input confuses me a bit. I've left some comments on them.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: Yes, it may be factored out to a helper class, but I don't find the similar api in other library. Therefore, I keep this function in this class.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: > The control size is set to 0 to disable the control input but the logic here is a bit different from the comment `If B is empty (size 0), control input is not supported`. What cause this difference?

There are not the difference. The shape of B should be 2D array or empty array.
Furthermore, the control input is the array `u` in  `void predict(array_type const & u);`.
If the user don't give the array B, he/she could only use `void predict();` instead of `void predict(array_type const & u);`.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: It should be fixed.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: The contorl input is mentioned in the doxygen of `void predict(array_type & u);`.

yungyuc: It writes:

```cpp
/**
 * @brief Predict step without control input.
 *
 * @details
 * Advances the state using only the transition model and process noise.
 * Represents a passive system with no external actuation.
 *
 * @note Common in finance, tracking, or passive simulation tasks.
 *
 * @see predict(const array_type&), update(const array_type&)
 */
```

"Control input" is not defined there.  Let me know if I should look for it elsewhere.

ThreeMonth03:
``` c++  
  /**
     * @brief Predict step with control input.
     *
     * @details
     * Updates the state while applying a known control SimpleArray.
     * The control term affects only the state mean.
     *
     * @note Example: in a vehicle model, the state may contain position and velocity,
     * while the control input represents acceleration (throttle or brake). The
     * control term updates how the state evolves between sensor measurements.
     *
     * @param u Control input.
     *
     * @see predict(), update(const array_type&)
     */
    void predict(array_type & u);
```

 It is clear that the input `u` is the control input.

ThreeMonth03 (commented):
@yungyuc @KHLee529 . Thanks for your suggestion. You could review this pull request again.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,531 @@
...
+    array_type innovation(array_type const & z);
+    array_type innovation_covariance();
+    array_type kalman_gain(array_type const & s);
+    void update_state(array_type const & k, array_type const & y);
+    void update_covariance(array_type const & k);
+
+}; /* end class KalmanFilter */
+
+template <typename T>
+auto KalmanFilter<T>::create_noise_matrix(size_t size, real_type noise_var) -> array_type
```
yungyuc: Why not putting this one-liner function in declaration?

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,531 @@
...
+template <typename T>
+void KalmanFilter<T>::predict_covariance()
+{
+    // P <- F P F^H + Q  (m_p <- m_f @ m_p @ m_f^H + m_q)
+    array_type f_h = hermitian(m_f);
+    m_p = m_f.matmul(m_p).matmul(f_h).add(m_q);
+}
+
+template <typename T>
+auto KalmanFilter<T>::symmetrize(array_type const & p) -> array_type
```
yungyuc: It would be great if we can use

```cpp
array_type KalmanFilter<T>::symmetrize(array_type const & p)
```

Could you please try if the following style works?  If it does please use it.  `auto` does not look very intuitive in function signature.

```cpp
KalmanFilter<T>::array_type KalmanFilter<T>::symmetrize(array_type const & p)
```

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,531 @@
...
+     *
+     * @details
+     * Updates the state while applying a known control SimpleArray.
+     * The control term affects only the state mean.
+     *
+     * @note Example: in a vehicle model, the state may contain position and velocity,
+     * while the control input represents acceleration (throttle or brake). The
+     * control term updates how the state evolves between sensor measurements.
+     *
+     * @param u Control input.
```
yungyuc, line 159: I got it (control input).  Thank you @ThreeMonth03 .

yungyuc (changes requested):
Points to address:

- [x] Why not putting the one-liner function `create_noise_matrix()` in declaration?
- [x] Check if we can avoid `auto` in function signature without making the code too clumsy.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: Sounds like a good idea. I will fix it.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: The function definition out of the class should work like this, but it looks quite long. 
```c++
template<typename T>
typename KalmanFilter<T>::array_type KalmanFilter<T>::symmetrize(array_type const & p)
```

Therefore, I decide to use this style.
```c++
template <typename T>
auto KalmanFilter<T>::symmetrize(array_type const & p) -> array_type
```

Maybe another good style is to define the function in the class.
```c++
class KalmanFilter {
public:
    array_type symmetrize(array_type const& p) {
        // ...
    }
};
```

By the way, it is illegal for declaring the alias of the return type.
See [this CI fails](https://github.com/solvcon/modmesh/actions/runs/18806175104/job/53660614293#step:12:625)
```c++
using kf_array_type = typename KalmanFilter<T>::array_type;
template<typename T>
kf_array_type KalmanFilter<T>::symmetrize(array_type const & p)
```

```
/home/runner/work/modmesh/modmesh/cpp/modmesh/linalg/kalman_filter.hpp:516:18: error: no declaration matches ‘modmesh::kf_array_type<T> modmesh::KalmanFilter<T>::hermitian(const array_type&)’
  516 | kf_array_type<T> KalmanFilter<T>::hermitian(array_type const & a)
      |                  ^~~~~~~~~~~~~~~
```

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,528 @@
...
+        // P <- (I-K H) P (I-K H)^H + K R K^H  (m_p <- (I-k@m_h) @ m_p @ (I-k@m_h)^H + k @ m_r @ k^H)
+        update_covariance(k);
+    }
+
+private:
+
+    static array_type create_noise_matrix(size_t size, real_type noise_var)
+    {
+        return array_type::eye(size).mul(static_cast<T>(noise_var));
+    }
```
ThreeMonth03, line 213: Move the definition in class.

ThreeMonth03 (commented):
@yungyuc . Please give it a look . Thanks.

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: > The function definition out of the class should work like this, but it looks quite long.
> 
> ```c++
> template<typename T>
> typename KalmanFilter<T>::array_type KalmanFilter<T>::symmetrize(array_type const & p)
> ```

Let us use this one.  It is harder to write but much easier to read.

> By the way, it is illegal for declaring the alias of the return type. See [this CI fails](https://github.com/solvcon/modmesh/actions/runs/18806175104/job/53660614293#step:12:625)
> 
> ```c++
> using kf_array_type = typename KalmanFilter<T>::array_type;
> template<typename T>
> kf_array_type KalmanFilter<T>::symmetrize(array_type const & p)
> ```
> 
> ```
> /home/runner/work/modmesh/modmesh/cpp/modmesh/linalg/kalman_filter.hpp:516:18: error: no declaration matches ‘modmesh::kf_array_type<T> modmesh::KalmanFilter<T>::hermitian(const array_type&)’
>   516 | kf_array_type<T> KalmanFilter<T>::hermitian(array_type const & a)
>       |                  ^~~~~~~~~~~~~~~
> ```

This is illegal.  C++ is one of the most hated programming languages and it is one of the reasons.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -0,0 +1,528 @@
...
+            }
+            oss << u.shape(i);
+        }
+        oss << ")";
+        throw std::invalid_argument(oss.str());
+    }
+}
+
+template <typename T>
+typename KalmanFilter<T>::array_type KalmanFilter<T>::innovation(array_type const & z)
```
yungyuc, line 455: This is good.  Let us follow the style to avoid `auto`.

```cpp
template <typename T>
typename KalmanFilter<T>::array_type KalmanFilter<T>::innovation(array_type const & z)
```

It is harder to write but much easier to read (clearer to parse in mind).

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: This is good.  Thank you.

yungyuc (approved):
LGTM

Issue #505, opened 2025-04-16
Add unit tests for SVG parser and rendering


YenPeiChen07:
Following up on #494, add a unit test case for SVG parser to verify the correctness of SVG parsing logic.

yungyuc:
@YenPeiChen07 this should be done before issue #504 .

yungyuc:
`EPath` is included in the unit test with PR #522: https://github.com/solvcon/modmesh/blob/7a85953f9194f34c5745b7a10e22ef23614c89c0/tests/test_svg.py#L30

But [`PathParser`](https://github.com/solvcon/modmesh/blob/7a85953f9194f34c5745b7a10e22ef23614c89c0/modmesh/plot/svg.py#L440) is not covered by the unit test file.  We should use another PR to test `PathParser` and take the opportunity to refactor the SVG and test code.

yungyuc:
I found an archive for SVG for testing purposes: https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/ .  We can check if it is OK for use in modmesh.

tigercosmos:
Here are the missing test cases for the SVG parser:

- Test cases for individual path commands (current tests lack coverage for some commands and are not isolated by command type):
    - MoveTo: M, m
    - LineTo: L, l, H, h, V, v
    - Cubic Bézier curve: C, c, S, s
    - Quadratic Bézier curve: Q, q, T, t
    - Elliptical arc curve: A, a
    - ClosePath: Z, z
- Test cases for paths containing multiple commands (as parsed by `PathParser`) using [SVG files](https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/) as integration tests for command sequences. A set of around 10 files should suffice.
- Test cases for [basic shapes](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/Basic_shapes). Failures are expected since no implementation exists yet. Likely, we will introduce `svg.ShapeParse`.
- Test cases for SVG syntax, tokenization, and lexing.

It is recommended to shift the core computation logic to C++ while developing the test cases. This would enable the creation of a C++ backend for the SVG parser, making the `svg.PathParse` Python class lightweight. Most parsing and data generation should occur on the C++ side, with Python handling only the returned results. The work can be done within two days.

cc @yungyuc  @YenPeiChen07

yungyuc:
> Here are the missing test cases for the SVG parser:

[...]

PR #610, opened 2025-10-25
Init svg pybind module

tigercosmos:
Part of #505.

Set up the SVG C++ backend with pybind. Will move helper functions of Python SVG parser to C++ later. Eventually move most of the logic from Python to C++.

### DIFF
--- a/cpp/modmesh/python/module.cpp
+++ b/cpp/modmesh/python/module.cpp
@@ -33,6 +33,7 @@
 #include <modmesh/math/pymod/math_pymod.hpp>
 #include <modmesh/transform/pymod/transform_pymod.hpp>
 #include <modmesh/linalg/pymod/linalg_pymod.hpp>
+#include <modmesh/svg/pymod/svg_pymod.hpp>
 
 #ifdef USE_PYTEST_HELPER_BINDING
 #include <modmesh/testhelper/pymod/testbuffer_pymod.hpp>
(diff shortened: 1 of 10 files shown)


cpp/modmesh/svg/pymod/wrap_svg.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace python
+{
+
+void wrap_svg(pybind11::module & mod)
+{
+    namespace py = pybind11;
+
+    mod
+        .def("create_svg_header", svg::create_svg_header, py::arg("width"), py::arg("height"))
```
tigercosmos, line 42: Just a dummy function for testing the setup..

tests/test_svg.py
```diff
@@ -423,4 +424,12 @@ def test_close_path_command(self):
         self.assertEqual(list(sp2d.x1), [20.0, 30.0, 10.0])
         self.assertEqual(list(sp2d.y1), [10.0, 10.0, 10.0])
 
+class SvgBindingTC(ModMeshTB, unittest.TestCase):
```
tigercosmos, line 428: Put the pybind testing also in `test_svg.py`. I suggest rename the file and seperate Python code and C++ pybind code later, such as `test_pysvg.py` for pure Python testing and `test_svg.py` for C++ binding.

tigercosmos:
@yungyuc  Please help me take a look. Thanks.

cpp/modmesh/python/module.cpp
```diff
@@ -58,6 +59,8 @@ void initialize(pybind11::module_ mod)
     initialize_inout(mod);
     initialize_math(mod);
     initialize_linalg(mod);
+    pybind11::module_ svg_mod = mod.def_submodule("svg", "svg");
```
yungyuc, line 62: Do not use a sub-module for SVG.  It's a small feature that should be included in the main module.

cpp/modmesh/svg/pymod/wrap_svg.cpp
yungyuc: If there is not useful function, do not add the module.

yungyuc (commented):
Do not add the boilerplate before having useful C++ code.

PR #615, opened 2025-10-27
SVG File Tests

tigercosmos:
Test the SVG parer using real SVG files, which are picked from [W3C](https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/), which are unique enough for testing paths and shapes.

The primary purpose of these file tests is to verify that the SVG parser can correctly handle real-world SVG files. Ideally, we would use another SVG library as the ground truth; however, that is currently impractical at this stage of development due to missing a lot of features. As long as individual paths and shapes are rendered correctly, we can reasonably assume that the complete image will also render as expected if there are no runtime errors. Therefore, I’m focusing on ensuring that the sample points fall within the spads and cpads, using the current implementation as the ground truth for the upcoming C++ refactoring. The current rendering result are shown below and seems to be correct compared with the original pictures execpt for the `smile.svg` one.

### Failed Case

- `smile.svg` uses `transform`, which is not supported yet.

### modmesh SVG
Screenshot 2025-10-28 at 1 29 41 AM

### Original SVG
Screenshot 2025-10-28 at 1 47 16 AM

### AI Usage

Use Copilot in the action mode to generate `SVG_TEST_DATA`


yungyuc:
Converted to draft for CI is not passing.

tigercosmos:
Fixed the CI. @yungyuc  please have a look.

tests/test_svg.py
```diff
@@ -921,4 +922,339 @@ def test_polygon_shape(self):
...
+
+        for index, expected_length in test_data['spad_lengths']:
+            self.assertEqual(len(spads[index]), expected_length)
+            self.assertEqual(spads[index].ndim, 2)
+
+        for index, expected_length in test_data['cpad_lengths']:
+            self.assertEqual(len(cpads[index]), expected_length)
+            self.assertEqual(cpads[index].ndim, 2)
+
+        for pad_index, segment_index, x0, y0, x1, y1 \
```
yungyuc: Avoid backslash `\` for line breaking.  Consider to use `_`:

```python
_ = test_data['sample_points']['spad']
or pad_index, segment_index, x0, y0, x1, y1 in _:
```

More concise style is welcome.

tests/test_svg.py
```diff
@@ -921,4 +922,339 @@ def test_polygon_shape(self):
...
+
+class SvgFileTC(SvgParserTB):
+    """
+    Test loading SVG files.
+    """
+
+    TESTDIR = os.path.abspath(os.path.dirname(__file__))
+    DATADIR = os.path.join(TESTDIR, "data")
+
+    SVG_TEST_DATA = {
```
yungyuc: Keep test meta data (fixture) close to test functions.  Jumping back-n-forth for the meta data and the test functions is counter-intuitive.

tigercosmos: The entire `SVG_TEST_DATA` was generated by AI, which used Python scripts to compute the numerical values under the assumption that the current implementation is correct. Since it would be difficult to manually verify all these values, they are treated as the ground truth for future refactoring. This allows us to check whether the C++ implementation preserves the same behavior as before.

However, the usefulness of these values for testing the Python implementation itself is debatable, and we may need to find a better approach for validating the integration tests.

yungyuc (changes requested):
Points to address:
- [x] Confirm the license of the SVG files and add them in the testing code and `tests/data/` directory (use a README or similar).
- [x] Multiline style.
- [x] Keep test meta data (fixture) close to test functions.

In the next code review, please add annotation for what (AI) tools do for you and how you react.  Also specify what tools and models are employed.

tests/test_svg.py
```diff
@@ -921,4 +922,339 @@ def test_polygon_shape(self):
...
+        'android.svg': {
+            'spads_count': 3,
+            'cpads_count': 5,
+            'total_segments': 14,
+            'total_curves': 9,
+            'spad_lengths': [(0, 8), (1, 2), (2, 4)],
+            'cpad_lengths': [(2, 1), (3, 4), (4, 4)],
+            'sample_points': {
```
tigercosmos: The sample points are calculated based on the SVG file and represent the key points that should be verified.

tests/test_svg.py
```diff
@@ -921,4 +922,339 @@ def test_polygon_shape(self):
...
+    """
+
+    TESTDIR = os.path.abspath(os.path.dirname(__file__))
+    DATADIR = os.path.join(TESTDIR, "data")
+
+    SVG_TEST_DATA = {
+        'android.svg': {
+            'spads_count': 3,
+            'cpads_count': 5,
+            'total_segments': 14,
```
tigercosmos: The number like `spads_count`, `cpads_count`, `total_segments` are highly related to the implementation, and they should keep the same value while we refactor the SVG parser in C++.

tigercosmos:
@yungyuc  I saw your comments. I will fix them. Meanwhile, I have added some inline annotations. Could you share your opion about the current design of file testing? I wonder if you have better ideas or other requirements to add into the design.

tests/test_svg.py
tigercosmos: Fixed according to the suggestion.

tests/data/svg/mars.svg
```diff
@@ -0,0 +1,21 @@
+<!--
```
tigercosmos, line 1: Added the licsense in the SVG file.

tests/data/svg/shapes.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" standalone="no"?>
+<!--
```
tigercosmos, line 2: It's weird that I must put the comment after `<?xml>` to let the XML parser work. The issue doesn't appear in other SVG files.

tests/data/svg/COPYING
```diff
@@ -0,0 +1,13 @@
+The SVG files (android.svg, beacon.svg, bozo.svg, caution.svg, debian.svg,
```
tigercosmos, line 1: The license file for SVG files.

COPYING
```diff
@@ -24,3 +24,8 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+---
+
+This project includes certain files licensed under the
```
tigercosmos: Also mention the license of SVGs in the root `COPYING`.

tests/data/svg/shapes.svg
tychuang1211: According to https://www.w3.org/TR/xml/#sec-prolog-dtd, 
[Definition: XML documents SHOULD begin with an XML declaration which specifies the version of XML being used.] 
A well-formed XML should begin with `<?xml>`,  and if the XML is not well-formed, the [SVG is not conforming](https://www.w3.org/TR/SVG/conform.html#DocumentConformanceClasses).

tests/test_svg.py
```diff
@@ -921,4 +922,335 @@ def test_polygon_shape(self):
...
+
+class SvgFileTC(SvgParserTB):
+    """
+    Test loading SVG files.
+    """
+
+    TESTDIR = os.path.abspath(os.path.dirname(__file__))
+    DATADIR = os.path.join(TESTDIR, "data")
+
+    def _test_svg_file(self, filename, test_data):
```
yungyuc: Rename `_test_svg_file()` to `_check_svg_file()`.  `test*` will be treat like a test function, so `_test*` is confusing.

Take each key in `test_data` as an argument because you are treating each key as an argument in this function.  The calling site will be like:

```python
self._test_svg_file(
    filename='android.svg',
    spads_count=3,
    cpads_count=5,
    total_segments=14,
    ...
)
```

In this way, the function signature will work like the schema (document) of the dictionary and make future addition and modification easier.

COPYING
yungyuc: Replace

```
This project ...
```

with just the simple description:

```
See also tests/data/COPYING for addition license information.
```

tests/data/svg/COPYING
yungyuc: An open question: Would it be good for maintenance to add a subdirectory `tests/data/svg/` for the SVG sample files?  10 files are many.

yungyuc (changes requested):
This is good!

Points to address:

- [x] Rename `_test_svg_file()` to `_check_svg_file()`.
- [x] Take each key in `test_data` as an argument to `_test_svg_file()`.
- [x] Simply mention the different license in a directory.
- [x] An open question: Would it be good for maintenance to add a subdirectory `tests/data/svg/` for the SVG sample files?
- [x] Add original URL to each of the SVG files in `tests/data/README.rst` or `tests/data/svg/README.rst`.

I think we need a readme file to keep source information of the svg files.

tests/data/svg/smile.svg
tychuang1211, line 1: Please move `<?xml>` to the beginning of the file to meet XML well-formedness constraint.

tychuang1211 (changes requested):

tests/data/svg/shapes.svg
tigercosmos: I see. Thanks for the information.

tests/data/svg/COPYING
tigercosmos: I am okay to do this.

COPYING
tigercosmos: Sure.

tests/data/svg/smile.svg
tigercosmos: Sure, thanks for the suggestion.

tests/data/svg/shapes.svg
tychuang2385: Thanks for your comment too! I wouldn’t have learned this without your question.

tigercosmos:
@yungyuc  All issues are addressed. Please have a look.

tests/data/svg/README.rst
```diff
@@ -0,0 +1,24 @@
+==============
+SVG Test Files
+==============
+
+This directory contains SVG files used for testing purposes in modmesh.
+
+Test Files
+==========
+
+- android.svg (original source: https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/android.svg)
```
yungyuc, line 10: Beautiful

tests/test_svg.py
```diff
@@ -28,6 +28,7 @@
 
 from modmesh.testing import TestBase as ModMeshTB
 import modmesh.plot.svg as svg
+import os
```
yungyuc, line 31: stdlib import should go before modmesh.  But this is minor.  Fix it by piggybacking a future PR.

yungyuc (approved):
LGTM

yungyuc:
@tigercosmos your commit log should include a summary of that you did.  The repository should contain most history if not all.  Missing a good summary in the commit log forces maintainers to rely on github.

tigercosmos:
@yungyuc  Copied that. I will be careful next time.

PR #619, opened 2025-10-29
feat: add object detector feature in modmesh

rockleona:
I started to have a research on computer vision, as the first step, this PR introduce [Ultralytics YOLO](https://github.com/ultralytics/ultralytics) as the object detection tool, and import yolov11 model as the base model.

Usage is really easy, just load an image, enable the detection, then you will see the result on the Console Widget. You can check the screenshot as below: 
截圖 2025-10-29 23 16 42


cpp/modmesh/pilot/RVisionDockWidget.cpp
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (c) 2022, Yung-Yu Chen <yyc@solvcon.net>
```
tigercosmos: Change to your name if you are the first author.

cpp/modmesh/pilot/RVisionDockWidget.hpp
```diff
@@ -0,0 +1,103 @@
...
+    void executeCommand();
+    void navigateCommand(int offset);
+
+private:
+    class Impl {
+    public:
+        QPushButton *toggleBtn = nullptr; 
+        bool enableDetection = false;
+    };
+    std::unique_ptr<Impl> m_impl;
```
tigercosmos: I don't think we need to use pimpl idiom.

cpp/modmesh/pilot/wrap_pilot.cpp
```diff
@@ -267,6 +267,23 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRPythonConsoleDockWidget
...
 }; /* end class WrapRPythonConsoleDockWidget */
 
+
+class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRVisionDockWidget
+    : public WrapBase<WrapRVisionDockWidget, RVisionDockWidget>
+{
+
+    friend root_base_type;
+
+    WrapRVisionDockWidget(pybind11::module & mod, char const * pyname, char const * pydoc)
```
tigercosmos: If there is nothing, maybe we don't need this?

modmesh/pilot/_vision.py
```diff
@@ -0,0 +1,99 @@
...
+
+        except Exception as e:
+            # If unable to get console widget, fall back to original method
+            log_stream = io.StringIO()
+            handler = logging.StreamHandler(log_stream)
+            logger.addHandler(handler)
+            logger.setLevel(logging.DEBUG)
+
+    if 'model' not in globals():
+        model = YOLO('./modmesh/pilot/yolo11n.pt')  # Please check the model path
```
tigercosmos: How do you get the model? Maybe have a runtime download logic?

cpp/modmesh/pilot/RVisionDockWidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+            boxes.push_back({rect, label, score});
+        }
+    }
+    m_image->setDetections(boxes);
+}
+
+RVisionImage::RVisionImage(QWidget *parent)
+    : QWidget(parent)
+{
+    m_image.load(":/default.jpg");
```
tigercosmos: You may want to put the image as an asset file.

modmesh/pilot/_vision.py
```diff
@@ -0,0 +1,99 @@
+"""
```
tigercosmos: Is it possible to write tests to validate the implementation?

modmesh/pilot/_vision.py
rockleona: Yes, it will check the path if the model is exist during runtime, nor it will download it directly to the specified path.

modmesh/pilot/_vision.py
rockleona: How would you recommend to place the tests, put them in the `tests/test_pilot.py` perhaps?

modmesh/pilot/_vision.py
rockleona: I just found there is a directory called `thirdparty`, maybe I should specify the path overthere instead?

cpp/modmesh/pilot/RVisionDockWidget.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+
+struct BoundingBox {
+    QRect bbox;
+    std::string label;
+    float score;
+};
+
+class RVisionDockWidget;
+class Impl;
```
yungyuc: The forward declaration `Impl` is no-op.

cpp/modmesh/pilot/RVisionDockWidget.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <QImage>
+#include <QPainter>
+#include <QPen>
+#include <QPushButton>
+#include <vector>
+
+namespace modmesh
+{
+
+struct BoundingBox {
```
yungyuc: Do not name it as `modmesh::BoundingBox`, which will be used for other purposes.

Do not create additional namespace under `modmesh`.  It will be over-engineering.

You may make it a nested class in `RVisionDockWidget`.  A shorter name would make the code easier.

cpp/modmesh/pilot/RVisionDockWidget.hpp
```diff
@@ -0,0 +1,103 @@
...
+    void setDetections(const std::vector<BoundingBox>& boxes);
+    void clearDetections();
+    QImage currentImage() const;
+    void loadImage(const QString& path);
+protected:
+    void paintEvent(QPaintEvent* event) override;
+private:
+    QImage m_image;
+    std::vector<BoundingBox> m_boxes;
+};
```
yungyuc: Always keep an end marker:

```cpp
...
}; /* end class RVisionImage */
```

cpp/modmesh/pilot/RVisionDockWidget.cpp
yungyuc: If you add an image in the repository, make sure it is small and uses a compatible license.

Alternately, load an image from online.

cpp/modmesh/pilot/RVisionDockWidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+    namespace py = pybind11;
+    py::gil_scoped_acquire gil;
+    int width = img.width();
+    int height = img.height();
+    int channels = 3;
+    QImage rgbImg = img.convertToFormat(QImage::Format_RGB888);
+    const uchar *data = rgbImg.bits();
+    py::array_t<uint8_t> np_img({height, width, channels}, data);
+    py::object vision_mod = py::module_::import("modmesh.pilot._vision");
+    py::object yolo_func = vision_mod.attr("yolo_detect");
```
yungyuc: If you call YOLO from Python, why not use PySide?

yungyuc (changes requested):
Please make sure CI passes before requesting for review.

- [ ] Correct copyright headers.
- [ ] Remove unnecessary code like `WrapRVisionDockWidget`, which looks like a placeholder.
- [ ] The pimpl class `RVisionDockWidget::Impl` is not necessary.  Do not use pimpl.
- [ ] Do not create a symbol named `modmesh::BoundingBox`.
- [ ] Always add an end marker to classes and namespaces.

I see you are using `pybind11` to call back into Python to use YOLO.  Why don't you just write PySide6 to do it?

modmesh/pilot/_vision.py
tigercosmos: `thirdparty` is for the 3rd libraries. In this case, I think you can put the model file at the same directory of pilot runtime. Btw, it seems that the download logic is not implemented yet, right?

modmesh/pilot/_vision.py
tigercosmos: I think you can put it at `tests/test_vision.py`?

rockleona:
> Please make sure CI passes before requesting for review.
> 
> * [ ]  Correct copyright headers.
> * [ ]  Remove unnecessary code like `WrapRVisionDockWidget`, which looks like a placeholder.
> * [ ]  The pimpl class `RVisionDockWidget::Impl` is not necessary.  Do not use pimpl.
> * [ ]  Do not create a symbol named `modmesh::BoundingBox`.
> * [ ]  Always add an end marker to classes and namespaces.
> 
> I see you are using `pybind11` to call back into Python to use YOLO. Why don't you just write PySide6 to do it?

I thought it was a must to write all the GUI component with `qt`, I will change it to PySide6 since these functions were executed only from Python

modmesh/pilot/_vision.py
rockleona: `ultralytics` library had already done the download logic, no need to do it agin, perhaps they will find the model name is in their file server or not, then download it when trigger class `YOLO` initialization.

yungyuc:
@rockleona The code base has changed a lot.  Please rebase to refresh the CI status.

rockleona:
I've made a lots of changes, please find the items below:

- Change all GUI components with PySide6
- Write unit test cases for `_yolo_detector`

The latest GUI will be look like this, I didn't change the layout but slightly different on the detail like model status and logging message in `pycon` widget
 
截圖 2025-11-24 22 04 07

rockleona:
> Please make sure CI passes before requesting for review.

I cannot find a button to run the CI process, maybe it should be executed by a repo maintainer?

yungyuc:
> > Please make sure CI passes before requesting for review.
> 
> I cannot find a button to run the CI process, maybe it should be executed by a repo maintainer?

You should use your own fork to test for CI.

> The latest GUI will be look like this, I didn't change the layout but slightly different on the detail like model status and logging message in `pycon` widget
> 
> 截圖 2025-11-24 22 04 07

Can you move the image preview away from the widget window (lower left) to the central sub-window, like other windows for 2D and 3D plots?

tests/data/jpg/COPYING
```diff
@@ -0,0 +1,12 @@
+The JPG file (cat.jpg) is licensed under
+the Unsplash License (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://unsplash.com/license
```
yungyuc, line 6: Make a copy of the license text here in this file.

tests/data/jpg/README.rst
```diff
@@ -0,0 +1,15 @@
+==============
+JPG Test Files
+==============
+
+This directory contains JPG files used for testing purposes in modmesh.
+
+Test Files
+==========
+
+- cat.jpg (original source: https://unsplash.com/photos/orange-and-white-cat-on-yellow-surface-sR0cTmQHPug?utm_source=unsplash&utm_medium=referral&utm_content=creditShareLink)
```
yungyuc, line 10: Clean up link.

tests/test_pilot_vision.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+import io
+import os
+import logging
+import unittest
+
+import numpy as np
+import requests
+from PIL import Image
```
yungyuc, line 35: If possible, I do not want to have PIL.  modmesh is already using Qt which should include all image handling features that PIL provides.  Please evaluate if you can simply use Qt/PySide for processing images.

tests/test_pilot_vision.py
```diff
@@ -0,0 +1,89 @@
...
+from ultralytics import YOLO
```
yungyuc, line 36: I feel including a huge thirdparty like ultralytics defeats the principle of "doing it ourselves" in modmesh.  @rockleona please elaborate why you include ultralytics.

yungyuc (changes requested):
There will be some discussions before we can merge any code about YOLO.

- [ ] Include upsplash license text.
- [ ] Clean up image source link.
- [ ] Evaluate to use Qt instead of PIL.
- [ ] Discuss why including ultralytics.

PR #622, opened 2025-11-02
Add memory alignment support for ConcreteBuffer and SimpleArray

### LINKED ISSUE #620: Memory Alignment

TODO list:

tigercosmos:
Part of #620.

Following the same approach as `BufferExpander` and `SimpleCollector`, and the memory alignment for `ConcreteBuffer` and `SimpleArray`.

## AI Usage

Env: GitHub Copilot + Claude 4.5

1. Asked AI to follow the same memory alignment approach from `BufferExpander` and `SimpleCollector` for `ConcreteBuffer` and `SimpleArray`.
2. Asked AI also add tests, which follows the privious PR as well.

Most of the code are correct, I added some comments and check over the test cases.

### DIFF
--- a/cpp/modmesh/buffer/ConcreteBuffer.hpp
+++ b/cpp/modmesh/buffer/ConcreteBuffer.hpp
@@ -61,11 +61,26 @@ struct ConcreteBufferRemover
     ConcreteBufferRemover & operator=(ConcreteBufferRemover &&) = default;
     virtual ~ConcreteBufferRemover() = default;
 
-    // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays,readability-non-const-parameter)
-    virtual void operator()(int8_t * p) const
+    static void deallocate_memory(int8_t * p, size_t alignment)
     {
-        // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-        delete[] p;
+        if (alignment > 0)
+        {
+#ifdef _WIN32
+            _aligned_free(p);
+#else
+            std::free(p);
+#endif
+        }
+        else
+        {
+            std::free(p);
+        }
+    }
+
+    // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays,readability-non-const-parameter)
+    virtual void operator()(int8_t * p, size_t alignment) const
+    {
+        deallocate_memory(p, alignment);
     }
 
 }; /* end struct ConcreteBufferRemover */
@@ -74,7 +89,7 @@ struct ConcreteBufferNoRemove : public ConcreteBufferRemover
 {
 
     // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays,readability-non-const-parameter)
-    void operator()(int8_t *) const override {}
+    void operator()(int8_t *, size_t) const override {}
 
 }; /* end struct ConcreteBufferNoRemove */
 
@@ -90,8 +105,9 @@ struct ConcreteBufferDataDeleter
     ConcreteBufferDataDeleter(ConcreteBufferDataDeleter &&) = default;
     ConcreteBufferDataDeleter & operator=(ConcreteBufferDataDeleter &&) = default;
     ~ConcreteBufferDataDeleter() = default;
-    explicit ConcreteBufferDataDeleter(std::unique_ptr<remover_type> && remover_in)
+    explicit ConcreteBufferDataDeleter(std::unique_ptr<remover_type> && remover_in, size_t alignment_in = 0)
         : remover(std::move(remover_in))
+        , alignment(alignment_in)
     {
     }
 
@@ -100,16 +116,16 @@ struct ConcreteBufferDataDeleter
     {
         if (!remover)
         {
-            // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-            delete[] p;
+            remover_type::deallocate_memory(p, alignment);
         }
         else
         {
-            (*remover)(p);
+            (*remover)(p, alignment);
         }
     }
 
     std::unique_ptr<remover_type> remover{nullptr};
+    size_t alignment = 0; // Alignment of the data buffer in bytes. 0 means no alignment.
 
 }; /* end struct ConcreteBufferDataDeleter */
 
@@ -134,10 +150,34 @@ private:
 public:
 
     using remover_type = detail::ConcreteBufferRemover;
+    using size_type = std::size_t;
 
-    static std::shared_ptr<ConcreteBuffer> construct(size_t nbytes)
+    /// Ensure the alignment value is valid.
+    static size_type validate_alignment(size_type alignment)
     {
-        return std::make_shared<ConcreteBuffer>(nbytes, ctor_passkey());
+        if (alignment != 0 && alignment != 16 && alignment != 32 && alignment != 64)
+        {
+            throw std::invalid_argument(
+                Formatter()
+                << "ConcreteBuffer: alignment must be 0, 16, 32, or 64, but got " << alignment);
+        }
+        return alignment;
+    }
+
+    /// Ensure the size value is valid with respect to alignment.
+    static void validate_size_alignment(size_type size, size_type alignment)
+    {
+        if (alignment > 0 && size % alignment != 0)
+        {
+            throw std::invalid_argument(
+                Formatter()
+                << "ConcreteBuffer: size " << size << " must be a multiple of alignment " << alignment);
+        }
+    }
+
+    static std::shared_ptr<ConcreteBuffer> construct(size_t nbytes, size_t alignment = 0)
+    {
+        return std::make_shared<ConcreteBuffer>(nbytes, alignment, ctor_passkey());
     }
 
     /*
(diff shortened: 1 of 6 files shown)


cpp/modmesh/buffer/pymod/wrap_ConcreteBuffer.cpp
```diff
@@ -53,19 +53,22 @@ WrapConcreteBuffer::WrapConcreteBuffer(pybind11::module & mod, char const * pyna
     (*this)
         .def_timed(
             py::init(
-                [](size_t nbytes)
-                { return wrapped_type::construct(nbytes); }),
-            py::arg("nbytes"))
+                [](size_t nbytes, size_t alignment)
+                { return wrapped_type::construct(nbytes, alignment); }),
+            py::arg("nbytes"),
+            py::arg("alignment") = 0)
```
tigercosmos, line 59: Python side also addes the `alignment` argument.

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -62,10 +62,20 @@ struct ConcreteBufferRemover
...
+        {
+#ifdef _WIN32
+            _aligned_free(p);
+#else
+            std::free(p);
+#endif
+        }
+        else
+        {
+            std::free(p);
```
tigercosmos, line 76: Same logic as previous PR.

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -134,10 +156,34 @@ class ConcreteBuffer
...
 
     using remover_type = detail::ConcreteBufferRemover;
+    using size_type = std::size_t;
 
-    static std::shared_ptr<ConcreteBuffer> construct(size_t nbytes)
+    /// Ensure the alignment value is valid.
+    static size_type validate_alignment(size_type alignment)
     {
-        return std::make_shared<ConcreteBuffer>(nbytes, ctor_passkey());
+        if (alignment != 0 && alignment != 16 && alignment != 32 && alignment != 64)
```
tigercosmos, line 158: Maybe we can use enum class for them, but maybe next time?

cpp/modmesh/buffer/ConcreteBuffer.hpp
```diff
@@ -241,23 +297,44 @@ class ConcreteBuffer
...
         {
-            ret = unique_ptr_type(new int8_t[nbytes], data_deleter_type());
+            void * ptr = nullptr;
+            if (alignment > 0)
+            {
+                validate_size_alignment(nbytes, alignment);
+#ifdef _WIN32
+                ptr = _aligned_malloc(nbytes, alignment);
+#else
+                ptr = std::aligned_alloc(alignment, nbytes);
```
tigercosmos, line 314: This is the key.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1162,13 +1172,30 @@ class SimpleArray
...
         {
-            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE);
+            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE, 0);
             m_body = m_buffer->template data<T>();
         }
     }
 
     // NOLINTNEXTLINE(modernize-pass-by-value)
-    explicit SimpleArray(small_vector<size_t> const & shape, value_type const & value)
+    SimpleArray(small_vector<size_t> const & shape, with_alignment_t, size_t alignment)
```
tigercosmos: Use  a `with_alignment_t` to distinguish the ctors.

tests/test_buffer.py
```diff
@@ -86,6 +86,107 @@ def test_ConcreteBuffer_from_ndarray(self):
         self.assertTrue((ndarr == 0).all())
 
 
+class ConcreteBufferAlignmentTC(unittest.TestCase):
```
tigercosmos, line 89: Test cases are similar to expaner and collector and they have been verified the correctness.

cpp/modmesh/buffer/ConcreteBuffer.hpp
yungyuc: These blocks (aligned allocation and deallocation) of code appear in multiple places.  What do you think to use a (free) helper function for them?

cpp/modmesh/buffer/ConcreteBuffer.hpp
yungyuc: A bitmask along with proper comments probably makes more sense.  But for the 4 possible value it does not matter how we check for them.  The current style is good enough.

For very wide vector registers we may need more value, but I don't think we need to check for more than 10, which is still OK for an explicit list.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1111,6 +1112,12 @@ class SimpleArrayMixinSearch
 
 } /* end namespace detail */
 
+// Tag type for explicit alignment constructor
+struct with_alignment_t
+{
+};
+inline constexpr with_alignment_t with_alignment{};
```
yungyuc: I don't know what does the `inline` here mean.  Could you please give me a hint?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Tagging argument should be at the end.

yungyuc (commented):
- [x] Discuss: should we use a free helper function for the aligned allocation and deallocation?
- [x] Make tagging argument at the end.

yungyuc (changes requested):
My previous review used a wrong type.  This is to request for changes.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: `with_alignment_t ` is a variable here in the header. Without `inline`, including SimpleArray.hpp in multiple `.cpp` files would result in multiple definitions of with_alignment, causing a linker error.

tigercosmos:

cpp/modmesh/buffer/ConcreteBuffer.hpp
tigercosmos: I think a helper function doesn't affact much, but I don't mind to add it. Just added it.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: I see.  There should not be a global instance of `with_alignment_t`.  Instead, construct `with_alignment_t` when calling the function that requires it.  It's a single-byte tagging class and compiler should optimize its construction away.  (It is how the passkeys work in the codebase.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -60,12 +60,25 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     [](py::object const & shape)
                     { return wrapped_type(make_shape(shape)); }),
                 py::arg("shape"))
+            .def_timed(
+                py::init(
+                    [](py::object const & shape, size_t alignment)
+                    { return wrapped_type(make_shape(shape), alignment, with_alignment); }),
```
yungyuc: Write the following instead to use the tagging class:

```cpp
wrapped_type(make_shape(shape), alignment, with_alignment_t{})
```

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1162,13 +1172,30 @@ class SimpleArray
...
         {
-            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE);
+            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE, 0);
             m_body = m_buffer->template data<T>();
         }
     }
 
     // NOLINTNEXTLINE(modernize-pass-by-value)
-    explicit SimpleArray(small_vector<size_t> const & shape, value_type const & value)
+    SimpleArray(small_vector<size_t> const & shape, size_t alignment, with_alignment_t)
```
yungyuc: Usually I use write the signature for tagging using `const &`:

```cpp
func(type real_argument, with_alignment_t const & /* unnamed argument for tagging */)
```

Reviewing the assembly will confirm if the compiler optimizes the unnamed argument away through inlining.

I think it works the same for an unnamed argument passed by value.

----

Even if the extra argument is really passed to the constructor, for a big object like `SimpleArray`, it's not a serious overhead.

yungyuc (changes requested):
Thanks.  There is follow-up work:

- [x] Do not create a global instance of `with_alignment_t`.
- [x] Double check that compiler optimizes an anonymous argument away when passing by value.

yungyuc:
I saw PR #625 follows up this one.  Note it here.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Yes, you are right. Asking AI to do some experiment on this, and the result shows that it's equivalent to pass by reference or pass by value here.

tigercosmos:
@yungyuc  the issue is addressed. Please check.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1162,13 +1171,30 @@ class SimpleArray
...
         {
-            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE);
+            m_buffer = buffer_type::construct(m_shape[0] * m_stride[0] * ITEMSIZE, 0);
             m_body = m_buffer->template data<T>();
         }
     }
 
     // NOLINTNEXTLINE(modernize-pass-by-value)
-    explicit SimpleArray(small_vector<size_t> const & shape, value_type const & value)
+    SimpleArray(small_vector<size_t> const & shape, size_t alignment, with_alignment_t const & /* unnamed argument for tagging */)
```
yungyuc, line 1180: The comment `/* unnamed argument for tagging */` is redundant and should be removed.  Unnamed argument for tagging is a well-established convention that explains itself, just like SFINAE.

yungyuc (approved):
LGTM

PR #626, opened 2025-11-11
add memory alignment for PointPad

### LINKED ISSUE #620: Memory Alignment

tigercosmos:
Part of #620. Add the memory alignment for `PointPad` with test cases.

## AI Usage

GitHub Copilot + Claude 4.5

1. Ask AI to add memory alignment for `PointPad`
2. Ask AI to add test cases (with more prompt for minor adjustment)
3. Manual check all implementation and test cases

### DIFF
--- a/cpp/modmesh/buffer/BufferBase.hpp
+++ b/cpp/modmesh/buffer/BufferBase.hpp
@@ -33,6 +33,31 @@
 namespace modmesh
 {
 
+/// Validate that alignment is one of the supported values (0, 16, 32, 64).
+inline std::size_t validate_alignment(std::size_t alignment, const char * prefix = nullptr)
+{
+    if (alignment != 0 && alignment != 16 && alignment != 32 && alignment != 64)
+    {
+        throw std::invalid_argument(
+            Formatter()
+            << (prefix ? prefix : "") << (prefix ? ": " : "")
+            << "alignment must be 0, 16, 32, or 64, but got " << alignment);
+    }
+    return alignment;
+}
+
+/// Validate that size is a multiple of alignment (when alignment > 0).
+inline void validate_size_alignment(std::size_t size, std::size_t alignment, const char * prefix = nullptr)
+{
+    if (alignment > 0 && size % alignment != 0)
+    {
+        throw std::invalid_argument(
+            Formatter()
+            << (prefix ? prefix : "") << (prefix ? ": " : "")
+            << "size " << size << " must be a multiple of alignment " << alignment);
+    }
+}
+
 /// Base class for buffer-like objects.
 template <typename Derived>
 class BufferBase
--- a/cpp/modmesh/universe/bezier.hpp
+++ b/cpp/modmesh/universe/bezier.hpp
@@ -294,6 +294,7 @@ public:
     using real_type = T;
     using value_type = T;
     using point_type = Point3d<T>;
+    using size_type = std::size_t;
 
     template <typename... Args>
     static std::shared_ptr<PointPad<T>> construct(Args &&... args)
@@ -303,6 +304,7 @@ public:
 
     PointPad(uint8_t ndim, ctor_passkey const &)
         : m_ndim(ndim)
+        , m_alignment(0)
     {
         if (ndim > 3)
         {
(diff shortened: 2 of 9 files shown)


cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -33,6 +33,29 @@
 namespace modmesh
 {
 
+/// Validate that alignment is one of the supported values (0, 16, 32, 64).
+inline std::size_t validate_alignment(std::size_t alignment)
```
tigercosmos: Extract the function to be a common one.

cpp/modmesh/buffer/BufferBase.hpp
```diff
@@ -33,6 +33,29 @@
...
+    {
+        throw std::invalid_argument(
+            Formatter()
+            << "alignment must be 0, 16, 32, or 64, but got " << alignment);
+    }
+    return alignment;
+}
+
+/// Validate that size is a multiple of alignment (when alignment > 0).
+inline void validate_size_alignment(std::size_t size, std::size_t alignment)
```
tigercosmos: Extract the function to be a common one.

cpp/modmesh/universe/bezier.hpp
```diff
@@ -322,9 +324,37 @@ class PointPad
...
+        else if (ndim < 2)
+        {
+            throw std::invalid_argument(
+                Formatter()
+                << "PointPad::PointPad: "
+                << "ndim = " << int(ndim) << " < 2");
+        }
+    }
+
+    PointPad(uint8_t ndim, size_t nelem, size_type alignment, ctor_passkey const &)
```
tigercosmos, line 352: add `alignment`, also for other ctors

tests/test_buffer.py
```diff
@@ -118,38 +118,38 @@ def test_alignment_validation_valid_values(self):
     def test_alignment_validation_invalid_values(self):
         with self.assertRaisesRegex(
                 ValueError,
-                "ConcreteBuffer: alignment must be 0, 16, 32, or 64, but got 17"  # noqa: E501
+                "alignment must be 0, 16, 32, or 64, but got 17"
```
tigercosmos: Change due to the refactoring.

tests/test_universe.py
```diff
@@ -946,6 +946,206 @@ def test_mirror_3d(self):
                 ValueError, "PointPad::mirror: axis must be 'x', 'y', or 'z'"):
             pp.mirror('w')
 
+    def test_alignment(self):
```
tigercosmos, line 949: The test cases explain themselves.

tigercosmos:
CI failed due to macos-13. Blocked by #627.

cpp/modmesh/buffer/BufferBase.hpp
yungyuc: This is good.  We probably do not need the `inline` prefix, but it's OK.

tests/test_buffer.py
yungyuc: We need the prefix name for class/function in the error messages.  Please add them back.

There are two ways to address it:

1. Make the validation helper returns an error state and create the exception object from the caller.
2. Send the prefix string to the common validation helper.  To save runtime overhead, the string may be C-style string rather than `std::string`.

The first approach may be more performant but use more duplicate code.  I am fine with duplicate code, which may be handled by using a macro.

tests/test_universe.py
```diff
@@ -946,6 +946,206 @@ def test_mirror_3d(self):
...
+        pp3d_aligned = self.PointPad(ndim=3, nelem=16, alignment=64)
+        self.assertEqual(pp3d_aligned.alignment, 64)
+
+        # Test invalid alignment values
+        invalid_alignments = [
+            1, 2, 4, 8, 128, 256, 1024, 15, 17, 31, 33, 63, 65
+        ]
+        for alignment in invalid_alignments:
+            with self.assertRaisesRegex(
+                    ValueError, "alignment must be 0, 16, 32, or 64"):
```
yungyuc: Need prefix string in the error message.

yungyuc (changes requested):
> CI failed due to macos-13. Blocked by #627.

It's scheduled burnout: https://github.com/actions/runner-images/issues/13046 .  A new upload after the scheduled time range should get it running again.  It can be waived for now.

One point to address:

- [ ] Add back the prefix name for class/function in the error messages.

tigercosmos:
@yungyuc  The issue is addressed. Please have a look. BTW, it seems that macos-13 still works. lol

yungyuc (approved):
LGTM.

> BTW, it seems that macos-13 still works. lol

Right.  It was a scheduled burnout.

PR #627, opened 2025-11-11
remove macos-13 and use macos-15 in the CI

### LINKED ISSUE #624: Upgrade mac runners from macos-13
Github Actions are deprecating `macos-13` runners.  We should upgrade mac builds to a proper version.  See https://github.com/actions/runner-images/issues/13046

tigercosmos:
Solution for #624. As discussed in Discord, we only need to keep one macOS CI, so update to use macos-15.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -201,7 +201,7 @@ jobs:
         # linter run on macos-15 and revisit in the future.
         # (20251112) Re-enabled macos-15 as macos-13 is being deprecated.
         os: [macos-15]
-        cmake_build_type: [Debug]
+        cmake_build_type: [Release]
 
       fail-fast: false
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -195,13 +195,10 @@ jobs:
 
     strategy:
       matrix:
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md
-        # 20240811: clang-tidy (llvm@16) runs extremely slowly on macos-15
-        # (i.e., macos-15-arm64), but not macos-13. Temporarily turn off the
-        # linter run on macos-15 and revisit in the future.
-        # (20251112) Re-enabled macos-15 as macos-13 is being deprecated.
-        os: [macos-15]
-        cmake_build_type: [Release]
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
+        # use macos-15-intel to have bigger 14 GB RAM; otherwise, macos-15 only has 7GB RAM
+        os: [macos-15-intel]
+        cmake_build_type: [Debug]
 
       fail-fast: false
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -195,13 +195,12 @@ jobs:
 
     strategy:
       matrix:
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
-        # 20240811: clang-tidy (llvm@16) runs extremely slowly on macos-14
-        # (i.e., macos-14-arm64), but not macos-13. Temporarily turn off the
-        # linter run on macos-14 and revisit in the future.
-        #os: [macos-13, macos-14]
-        os: [macos-13]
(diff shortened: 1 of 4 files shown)


yungyuc:
Thank you @tigercosmos .  But we shall not turn off CI runs when waiving the failures is an option for a specific PR.

I leave this PR open before @tychuang1211 gets a first attempt for upgrading to macos-15.

tigercosmos:
It seems that the CI gets stuck:
image

tigercosmos:
@yungyuc  could you take a look?

.github/workflows/lint.yml
```diff
@@ -195,13 +195,12 @@ jobs:
...
       matrix:
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
-        # 20240811: clang-tidy (llvm@16) runs extremely slowly on macos-14
-        # (i.e., macos-14-arm64), but not macos-13. Temporarily turn off the
-        # linter run on macos-14 and revisit in the future.
-        #os: [macos-13, macos-14]
-        os: [macos-13]
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md
+        # 20240811: clang-tidy (llvm@16) runs extremely slowly on macos-15
```
yungyuc: It seems still an issue:

Image

Could you please try if it works better with macos-15-intel and/or `cmake_build_type: [Release]`?

If not, we may need to turn the clang-tidy check off with the macos run, and create an issue to find a solution and turn it back on.

.github/workflows/nouse_install.yml
```diff
@@ -124,21 +124,22 @@ jobs:
...
+        run: |
+          brew install pybind11
+
       - name: dependency by pip
         run: |
           echo "which python3: $(which python3)"
           ls -al $(which python3)
-          python3 -m pip -v install --upgrade setuptools
-          python3 -m pip -v install --upgrade wheel # required by "--no-use-pep517"
+          python3 -m pip -v install --break-system-packages --upgrade setuptools
```
yungyuc, line 135: In the future, we may consider to switch to uv.  (We do not need to do anything in this PR.)

@tychuang1211 @terrychan999 what do you think?

yungyuc (changes requested):
- [ ] Check if we can make macos-15 clang-tidy build reasonably fast, e.g., about 10 minutes.

.github/workflows/lint.yml
tigercosmos: I guess it's due to the memory issue, there is only 7GB for ARM (M1)
image

tigercosmos: macos-15-intel takes about 20min, which is similar to other images. I think we can use it.

tigercosmos:
@yungyuc  please take a look.

.github/workflows/lint.yml
ExplorerRay: I just ran the CI on my [fork](https://github.com/ExplorerRay/modmesh/actions/runs/19383584849/job/55467629860).

About 16 minutes with 0% ccache hit and less than 10 minutes (which is faster than macos-13 on my fork) with 92.86% ccache hit at the most time-consuming step, so I think this change is good.

.github/workflows/lint.yml
```diff
@@ -195,13 +195,9 @@ jobs:
...
       matrix:
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
-        # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
-        # 20240811: clang-tidy (llvm@16) runs extremely slowly on macos-14
-        # (i.e., macos-14-arm64), but not macos-13. Temporarily turn off the
-        # linter run on macos-14 and revisit in the future.
-        #os: [macos-13, macos-14]
-        os: [macos-13]
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
+        # use macos-15-intel to have bigger 14 GB RAM; otherwise, macos-15 only has 7GB RAM
```
yungyuc, line 199: Great!

yungyuc (approved):
This is great!

.github/workflows/nouse_install.yml
terrychan999: Moving to uv in the future sounds great to me 🚀

PR #628, opened 2025-11-12
add the memory alignment check for NEON SIMD

### LINKED ISSUE #620: Memory Alignment

tigercosmos:
Part of #620. Add the memory alignment check for NEON SIMD to ensure the performance.

## AI Usage

GitHub Copilot + Claude 4.5

1. Ask AI to add checks in the SIMD code
2. Ask AI to finish the `test_alignment_with_simd_operations` tests
3. Manually adjust the implementation and tests.

### DIFF
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -59,6 +59,11 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     using cmpvec_t = type::vector_t<uint64_t>;
     constexpr size_t N_lane = type::vector_lane<T>;
 
+#ifndef NDEBUG
+    constexpr size_t alignment = get_recommended_alignment();
+    detail::check_alignment(start, alignment, "check_between start");
+#endif
+
     vec_t max_vec = vdupq(max_val);
     vec_t min_vec = vdupq(min_val);
     vec_t data_vec = {};
@@ -117,6 +122,14 @@ void add(T * dest, T const * dest_end, T const * src1, T const * src2)
     {
         using vec_t = type::vector_t<T>;
         constexpr size_t N_lane = type::vector_lane<T>;
+
+#ifndef NDEBUG
+        constexpr size_t alignment = get_recommended_alignment();
+        detail::check_alignment(dest, alignment, "add dest");
+        detail::check_alignment(src1, alignment, "add src1");
+        detail::check_alignment(src2, alignment, "add src2");
+#endif
+
         vec_t src1_vec;
         vec_t src2_vec;
         vec_t res_vec;
@@ -146,6 +159,14 @@ void sub(T * dest, T const * dest_end, T const * src1, T const * src2)
     {
         using vec_t = type::vector_t<T>;
         constexpr size_t N_lane = type::vector_lane<T>;
+
+#ifndef NDEBUG
+        constexpr size_t alignment = get_recommended_alignment();
+        detail::check_alignment(dest, alignment, "sub dest");
+        detail::check_alignment(src1, alignment, "sub src1");
+        detail::check_alignment(src2, alignment, "sub src2");
+#endif
+
         vec_t src1_vec;
         vec_t src2_vec;
         vec_t res_vec;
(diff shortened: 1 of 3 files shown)


cpp/modmesh/simd/simd.hpp
```diff
@@ -39,6 +39,49 @@ namespace modmesh
...
+    return (reinterpret_cast<std::uintptr_t>(pointer) % alignment) == 0;
+}
+
+template <typename T>
+inline void check_alignment(T const * pointer, size_t required_alignment, const char * name)
+{
+    if (!is_aligned(pointer, required_alignment))
+    {
+#ifndef NDEBUG
+        std::fprintf(stderr,
```
tigercosmos, line 56: only print the warning because nonalignment will not cause errors.

tests/test_buffer.py
```diff
@@ -1220,6 +1220,45 @@ def test_alignment_with_different_shapes(self):
         self.assertEqual(64, array3d.alignment)
         self.assertEqual((2, 4, 4), array3d.shape)
 
+    def test_alignment_with_simd_operations(self):
```
tigercosmos, line 1223: Better to put under `SimpleArrayAlignmentTC`.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -117,6 +122,14 @@ void add(T * dest, T const * dest_end, T const * src1, T const * src2)
     {
         using vec_t = type::vector_t<T>;
         constexpr size_t N_lane = type::vector_lane<T>;
+
+#ifndef NDEBUG
+        constexpr size_t alignment = get_recommended_alignment();
+        detail::check_alignment(dest, alignment, "add dest");
```
tigercosmos, line 128: Give warning if it's not aligned.

tests/test_buffer.py
yungyuc: Absolutely.

tests/test_buffer.py
```diff
@@ -1220,6 +1220,45 @@ def test_alignment_with_different_shapes(self):
...
+        size = 16
+        alignments = [16, 32, 64]
+
+        for alignment in alignments:
+            arr1 = modmesh.SimpleArrayFloat64((size,), alignment)
```
yungyuc, line 1228: What will happen with multi-dimensional arrays?

cpp/modmesh/simd/simd.hpp
```diff
@@ -39,6 +39,48 @@ namespace modmesh
 namespace simd
 {
 
+namespace detail
+{
+#ifndef NDEBUG
+template <typename T>
+inline bool is_aligned(T const * pointer, size_t alignment)
```
yungyuc: `inline` for function template doesn't make sense.

cpp/modmesh/simd/simd.hpp
yungyuc: Agree for NEON/AppleSilicon.  On x86 misalignment crashes.  But this implementation is correct regardless.

A side note.  In the future we will need a common helper function for managing warning messages.

yungyuc (changes requested):
One point to clarify:

- [x] Clarify what will happen with multi-dimensional arrays.

@KHLee529 please help review.

tests/test_buffer.py
tigercosmos: The current SIMD operations process the entire contiguous buffer linearly, so row alignment is irrelevant. Only the starting address and total size matter, both of which are validated during array creation. However, to prevent any issues by internal change of the SIMD implementation, I added more test cases.

tigercosmos:
@yungyuc  please take a look.

cpp/modmesh/simd/simd.hpp
KHLee529: I'm wondering why it doesn't make sense?

tests/test_buffer.py
yungyuc: I see.  Thanks for the clear analysis.

yungyuc (approved):
LGTM.

cpp/modmesh/simd/simd.hpp
yungyuc: Filed issue #629

cpp/modmesh/simd/simd.hpp
```diff
@@ -39,6 +39,48 @@ namespace modmesh
...
+                     "SIMD performance may be degraded.\n",
+                     name,
+                     static_cast<const void *>(pointer),
+                     required_alignment);
+    }
+}
+#endif
+
+// Get the recommended memory alignment for SIMD operations based on the detected SIMD instruction set.
+inline constexpr size_t get_recommended_alignment()
```
KHLee529, line 67: Where does these numbers come from?

KHLee529:
I'm wondering where the recommended alignment comes from?

cpp/modmesh/simd/simd.hpp
tigercosmos: @KHLee529 the numbers are based on the instruction set, which can be detected by the macros.

cpp/modmesh/simd/simd.hpp
tigercosmos: Compiler will expand the template in cpp code anyway, so `inline` does nothing.

PR #630, opened 2025-11-16
Move NEON SIMD helpers from `simd.hpp` to `neon.hpp`

yungyuc:
Without the patch, I got build error on my local macos 26:

```
cd /Users/yungyuc/work/modmesh/build/clion/restdebug/cpp/modmesh && /usr/bin/c++ cpp/modmesh/CMakeFiles/modmesh_primary.dir/toggle/pymod/toggle_pymod.cpp.o -MF CMakeFiles/modmesh_primary.dir/toggle/pymod/toggle_pymod.cpp.o.d -o CMakeFiles/modmesh_primary.dir/toggle/pymod/toggle_pymod.cpp.o -c /Users/yungyuc/work/modmesh/cpp/modmesh/toggle/pymod/toggle_pymod.cpp
In file included from /Users/yungyuc/work/modmesh/cpp/modmesh/buffer/SimpleArray.cpp:29:
In file included from /Users/yungyuc/work/modmesh/cpp/modmesh/buffer/SimpleArray.hpp:33:
In file included from /Users/yungyuc/work/modmesh/cpp/modmesh/simd/simd.hpp:34:
/Users/yungyuc/work/modmesh/cpp/modmesh/simd/neon/neon.hpp:63:34: error: use of undeclared identifier 'get_recommended_alignment'
   63 |     constexpr size_t alignment = get_recommended_alignment();
      |                                  ^
/Users/yungyuc/work/modmesh/cpp/modmesh/simd/neon/neon.hpp:64:13: error: no member named 'check_alignment' in namespace 'modmesh::simd::detail'
   64 |     detail::check_alignment(start, alignment, "check_between start");
      |     ~~~~~~~~^
/Users/yungyuc/work/modmesh/cpp/modmesh/simd/neon/neon.hpp:127:38: error: use of undeclared identifier 'get_recommended_alignment'
  127 |         constexpr size_t alignment = get_recommended_alignment();
      |                                      ^
/Users/yungyuc/work/modmesh/cpp/modmesh/simd/neon/neon.hpp:128:17: error: no member named 'check_alignment' in namespace 'modmesh::simd::detail'
  128 |         detail::check_alignment(dest, alignment, "add dest");
      |         ~~~~~~~~^
```

It seems to come from PR #628 , but I do not know why CI did not catch it.

@KHLee529 is `simd_generic.hpp` a good location for the helpers?  I am not sure about it.

### DIFF
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -75,6 +75,7 @@ inline constexpr size_t get_recommended_alignment()
 {
 #if defined(__aarch64__) || defined(__arm__)
     return 16;
+// TODO: The non-NEON conditional should be factored out elsewhere in the future.
 #elif defined(__AVX512F__)
     return 64;
 #elif defined(__AVX__) || defined(__AVX2__)
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -45,6 +45,49 @@ namespace simd
 namespace neon
 {
 
+namespace detail
+{
+
+#ifndef NDEBUG
+template <typename T>
+bool is_aligned(T const * pointer, size_t alignment)
+{
+    return (reinterpret_cast<std::uintptr_t>(pointer) % alignment) == 0;
+}
+
+template <typename T>
+void check_alignment(T const * pointer, size_t required_alignment, const char * name)
+{
+    if (!is_aligned(pointer, required_alignment))
+    {
+        std::fprintf(stderr,
+                     "Warning: %s pointer %p is not aligned to %zu bytes. "
+                     "SIMD performance may be degraded.\n",
+                     name,
+                     static_cast<const void *>(pointer),
+                     required_alignment);
+    }
+}
+#endif // NDEBUG
+
+// Get the recommended memory alignment for SIMD operations based on the detected SIMD instruction set.
+inline constexpr size_t get_recommended_alignment()
+{
+#if defined(__aarch64__) || defined(__arm__)
+    return 16;
+#elif defined(__AVX512F__)
+    return 64;
+#elif defined(__AVX__) || defined(__AVX2__)
+    return 32;
+#elif defined(__SSE__) || defined(__SSE2__) || defined(__SSE3__) || defined(__SSSE3__) || defined(__SSE4_1__) || defined(__SSE4_2__)
+    return 16;
+#else
+    return 0;
+#endif
+}
+
+} /* end namespace detail */
+
 #if defined(__aarch64__)
 template <typename T, typename std::enable_if_t<!type::has_vectype<T>> * = nullptr>
 const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
@@ -60,7 +103,7 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     constexpr size_t N_lane = type::vector_lane<T>;
 
 #ifndef NDEBUG
-    constexpr size_t alignment = get_recommended_alignment();
+    constexpr size_t alignment = detail::get_recommended_alignment();
     detail::check_alignment(start, alignment, "check_between start");
 #endif
 
@@ -91,7 +134,7 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
 
     if (ptr != end)
     {
-        ret = generic::check_between<T>(ptr, end, min_val, max_val);
+        ret = check_between<T>(ptr, end, min_val, max_val);
     }
 
     return ret;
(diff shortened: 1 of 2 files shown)


yungyuc:
@tigercosmos @KHLee529 this fix is still work in progress and does not yet pass CI, but I am curious why we got the build error in the code base (or is it in my environment)?

KHLee529:
As what I found from simply look into code, `get_recommended_alignment` and `check_alignment` is defined in file `simd.hpp` which is not included by `neon.hpp`. This may possibly be the reason of this compile error. 
The reason why CI doesn't reproduce this problem might be that `NDEBUG` is not defined in CI compile option though I haven't confirm this.

KHLee529:
> @KHLee529 is simd_generic.hpp a good location for the helpers? I am not sure about it.

As for this question, in short, I prefer putting it in `simd_support.hpp` or even each platform.

When I designed the SIMD library, I treated `simd.hpp` as the only entry point of SIMD library and the library function interfaces in it are implemented based on it's functionality, that's why there're functions like `check_between`. In contrast, most of the SIMD library seems to wrap the different hardwares into same operations. For example, they might provide `vec_load`, `vec_save` functions to load/save data from memory to/back from vector registers, and `vec_add`, `vec_subtract` functions to simply add two vector registers together. Maybe the `neon_alias.hpp` can be a analogy to them. 

`simd_generic.hpp` is designed to prevent malfunction when no SIMD feature is supported on the target platform so we can find also `check_between` function without SIMD operations in it. The flow chart may look like below. 

```mermaid
flowchart TD
    A[SIMD::check_between] --> C{detect SIMD support}
    C --> |support neon| N[neon::check_between]
    C --> |support AVX512| X[avx512::check_between]
    C --> |not support any SIMD feature| G[generic::check_between]
```

So `simd_support.hpp` that currently provides only the platform detection function plays the role as a utility library for all SIMD features, that's why I prefer putting the helper functions there. Maybe it can be renamed to `simd_util.hpp`

In another way of thinking, if the `get_recommended_alignement` functions are platform-specific features, they may also be properly located in the directories of each platform like set it to 16 in `neon.hpp` and 0 in `simd_generic.hpp` and 64 for `avx512.hpp` in the future. 

@yungyuc @tigercosmos This is my thought to this SIMD library. Please comment or suggest your opinion.

tigercosmos:
I've tested the latest master locally on Ubuntu and macOS, and I haven't encountered any compilation errors. However, I agree with @KHLee529's solution.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -45,6 +45,49 @@ namespace simd
...
+                     "SIMD performance may be degraded.\n",
+                     name,
+                     static_cast<const void *>(pointer),
+                     required_alignment);
+    }
+}
+#endif // NDEBUG
+
+// Get the recommended memory alignment for SIMD operations based on the detected SIMD instruction set.
+inline constexpr size_t get_recommended_alignment()
```
yungyuc, line 74: I think `get_recommended_alignment` should be an implementation detail in this early stage of development.  If we find it is a generally useful helper, then take it outside `detail` in the future.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -45,6 +45,49 @@ namespace simd
...
+{
+#if defined(__aarch64__) || defined(__arm__)
+    return 16;
+#elif defined(__AVX512F__)
```
yungyuc, line 79: The helper contains non-NEON conditionals that should be refactored in the future.

cpp/modmesh/simd/simd.hpp
```diff
@@ -39,56 +39,13 @@ namespace modmesh
...
-
-} // namespace detail
-
 // Check if each element from start to end (excluded end) is within the range [min_val, max_val)
 template <typename T>
 const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
 {
-    using namespace detail;
-    switch (detect_simd())
+    switch (detail::detect_simd())
```
yungyuc, line 46: We should not `using namespace detail`, or defeat part of the purpose of the namespace `detail`.  I created issue #631 to get rid of the misusing.

yungyuc:
Thank you, @tigercosmos @KHLee529 .  I am following @KHLee529 recommendation to move the NEON-specific code to `neon.hpp`.  More refactoring may be done as we go.

PR is ready for review.  The code passed CI in my fork.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -45,6 +45,49 @@ namespace simd
...
+    return 64;
+#elif defined(__AVX__) || defined(__AVX2__)
```
tigercosmos, line 81: if we put this function in NEON, then we don't need these lines.

yungyuc: It's OK.  The non-NEON conditional code is OK as a hint for a need for refactoring.

To make it clear, I added a comment for TODO 3 lines above.

yungyuc:
`macos-15` devbuild has run for more than 3 hours.  Merge without waiting for it.

Issue #631, opened 2025-11-16
Do not `using namespace` even in function local


yungyuc:
We have [`using namespace detail`](https://github.com/solvcon/modmesh/blob/47adbc94a853a350a5b816fcec674377db91fb82/cpp/modmesh/simd/simd.hpp#L103) in function local.  It should not be allowed because it prevents a maintainer to quickly know which functions are from the lower namespace `detail`.

```cpp
template <typename T>
void add(T * dest, T const * dest_end, T const * src1, T const * src2)
{
    using namespace detail;
    //...
}
```

c1ydehhx:
@yungyuc 
I think I can fix this issue. Please assign this to me.

yungyuc:
> @yungyuc 
> I think I can fix this issue. Please assign this to me.

Thank you @c1ydehhx.  Assigned.

rockleona:
Hi @yungyuc, I'm interested in this issue and would like to solve it.

yungyuc:
> Hi [@yungyuc](https://github.com/yungyuc), I'm interested in this issue and would like to solve it.

Thank you @rockleona .  Reassigned.

yungyuc:
Resolved by PR #1360

PR #632, opened 2025-11-16
Refactor(simd): Improve clarity and consistency in dispatcher

hakkiai:
Removed the `using namespace detail` statements and replaced them with explicit qualifiers to make the code easier to follow. Also removed redundant break statements and added NEON dispatch for the max function to be consistent with other SIMD operations. @yungyuc

### DIFF
--- a/cpp/modmesh/simd/simd.hpp
+++ b/cpp/modmesh/simd/simd.hpp
@@ -47,7 +47,6 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     {
     case detail::SIMD_NEON:
         return neon::check_between<T>(start, end, min_val, max_val);
-        break;
 
     default:
         return generic::check_between<T>(start, end, min_val, max_val);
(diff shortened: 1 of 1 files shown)


cpp/modmesh/simd/simd.hpp
```diff
@@ -57,12 +56,10 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
...
 void add(T * dest, T const * dest_end, T const * src1, T const * src2)
 {
-    using namespace detail;
-    switch (detect_simd())
+    switch (detail::detect_simd())
     {
-    case SIMD_NEON:
+    case detail::SIMD_NEON:
         return neon::add<T>(dest, dest_end, src1, src2);
-        break;
```
yungyuc, line 65: Removing `break` is not conventional.  Don't do it.

yungyuc (changes requested):
Always adding `break` after case in case a maintainer forgets about it after changing the code before it.

- [ ] Don't remove `break` even if it is no-op.

yungyuc:
Make sure CI passes before requesting for a review.  You can run it in your local fork.

yungyuc:
This PR is for issue #631 .

cpp/modmesh/simd/simd.hpp
hakkiai: I initially removed the break statements because they were technically unreachable after a return, so I deleted the break statement to simplify the switch block code. But I understand your point now.

hakkiai:
Thanks for the review & feedback.

PR #636, opened 2025-11-21
support memory alignment for arrayplex

### LINKED ISSUE #620: Memory Alignment

tigercosmos:
Part of #620. Since ArrayPlex should have the same APIs as SimpleArray. The corresponding changes are made in this PR.

Using Copilot + Claude 4.5 to finish most of the code in this PR.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -219,6 +219,37 @@ SimpleArrayPlex::SimpleArrayPlex(const shape_type & shape, const std::shared_ptr
 
 #undef DECL_MM_CREATE_SIMPLE_ARRAY
 
+SimpleArrayPlex::SimpleArrayPlex(const shape_type & shape, const DataType data_type, size_t alignment)
+    : m_data_type(data_type)
+    , m_has_instance_ownership(true)
+{
+#define DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType, ArrayType)                            \
+    case DataType:                                                                                 \
+        m_instance_ptr = static_cast<void *>(new ArrayType(shape, alignment, with_alignment_t{})); \
+        break;
+
+    switch (data_type)
+    {
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Bool, SimpleArrayBool)
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Int8, SimpleArrayInt8)
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Int16, SimpleArrayInt16)
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Int32, SimpleArrayInt32)
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Int64, SimpleArrayInt64)
+        DECL_MM_CREATE_SIMPLE_ARRAY_WITH_ALIGNMENT(DataType::Uint8, SimpleArrayUint8)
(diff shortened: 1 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1989,6 +1989,7 @@ class SimpleArrayPlex
 
     explicit SimpleArrayPlex(const shape_type & shape, const DataType data_type);
     explicit SimpleArrayPlex(const shape_type & shape, const std::shared_ptr<ConcreteBuffer> & buffer, const DataType data_type);
+    explicit SimpleArrayPlex(const shape_type & shape, const DataType data_type, size_t alignment);
```
tigercosmos, line 1992: The meat.

tests/test_buffer.py
```diff
@@ -3216,6 +3216,163 @@ def test_SimpleArrayPlex_set_item_ellipse(self):
                     self.assertEqual(stride_arr[i, j, k], sarr[i, j, k])
 
 
+class SimpleArrayPlexAlignmentTC(unittest.TestCase):
```
tigercosmos, line 3219: We don't need much test cases as the underlying is SimpleArray, which is well tested.

tigercosmos:

tests/test_buffer.py
yungyuc: Agree.

tests/test_buffer.py
```diff
@@ -3216,6 +3216,163 @@ def test_SimpleArrayPlex_set_item_ellipse(self):
...
+
+    def test_alignment_validation_valid_values(self):
+        modmesh.SimpleArray((4, 4), dtype='float64', alignment=16)
+        modmesh.SimpleArray((8, 8), dtype='float64', alignment=32)
+        modmesh.SimpleArray((16, 8), dtype='float64', alignment=64)
+
+    def test_alignment_validation_invalid_values(self):
+        with self.assertRaisesRegex(
+                ValueError,
+                "ConcreteBuffer::ConcreteBuffer: alignment must be 0, 16, 32, or 64, but got 17"
```
yungyuc: Fix the long line linter errors like in this way:

```python
        with self.assertRaisesRegex(
                ValueError,
                "ConcreteBuffer::ConcreteBuffer: "
                "alignment must be 0, 16, 32, or 64, but got 17"
        ):
```

So it does not use explicit line continuation `\`, and still looks like a line.

yungyuc (changes requested):
- [ ] Make linter happy.

tigercosmos:
@yungyuc  please have a look

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -478,6 +509,46 @@ SimpleArrayPlex & SimpleArrayPlex::operator=(SimpleArrayPlex && other) noexcept
...
+size_t SimpleArrayPlex::alignment() const
+{
+    if (m_instance_ptr == nullptr)
+    {
+        return 0;
+    }
+
+    switch (m_data_type)
+    {
+    case DataType::Bool:
```
yungyuc, line 521: A question just occurred to me: when we want to add struct array support, this switch case will be a problem.

But we worry about it when needing to support struct array.  We do not worry about it now.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Tasty

yungyuc (approved):
LGTM

cpp/modmesh/buffer/SimpleArray.cpp
tigercosmos: Sorry, not quite sure about your point. Could you describe more about the issue you see? @yungyuc

yungyuc: It's not critical atm.  Discuss on dc if you want: https://discord.com/channels/730297880140578906/730393342818648204/1442850883401416705

PR #640, opened 2025-11-24
 Add Polygon3d and PolygonPad support

### LINKED ISSUE #633: Prototype an R-tree for `SegmentPad`
An [R-tree](https://dl.acm.org/doi/10.1145/602259.602266) is a spatial index allowing fast search for geometry entities.  We will need the spatial indices for points, line segments, Bezier curves, and more geometry entities to be included in modmesh.

This issue is to track the first, prototype-ish implementation of R-tree for the line segments in `SegmentPad`.  The spatial index will be a foundation for high-level features.

The R-tree implementation should allow both 2 and 3 dimensional entities, but the tests can focus on 2D at the moment.

tigercosmos:
## Summary
- add `Polygon3d` handle and `PolygonPad` container to the universe module, including polygon construction from nodes, segments, and curves
- compute polygon properties (node/edge access, signed area, bound boxes) and index edges with an R-tree for spatial search
- expose polygon APIs to Python bindings and export types in `modmesh.core`
- add `Polygon3d` unit tests covering creation, orientation, equality/identity, and R-tree search

## AI Usage

GitHub Copilot + Cluade 4.5
1. Provided the class and API draft and asked AI to finish the implementation
2. Asked AI to generate tests
3. Asked AI to generate PR description

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -78,14 +78,15 @@ jobs:
 
     - name: dependency by apt
       run: |
-        VERSION_ID=$(bash -c 'source /etc/os-release ; echo $VERSION_ID')
-        if [ "20.04" == "$VERSION_ID" ] ; then CLANG_TIDY_VERSION=10 ; else CLANG_TIDY_VERSION=18 ; fi
+        # Install clang-tidy-21 from LLVM repository for full C++23 support
+        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+        echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-21 main" | sudo tee /etc/apt/sources.list.d/llvm.list
         sudo apt-get -qqy update
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ silversearcher-ag \
-            clang-tidy-${CLANG_TIDY_VERSION} \
+            clang-tidy-21 \
             python3 python3-dev python3-venv
-        sudo ln -fs "$(which clang-tidy-${CLANG_TIDY_VERSION})" "$(dirname $(which clang-tidy-${CLANG_TIDY_VERSION}))/clang-tidy"
+        sudo ln -fs "$(which clang-tidy-21)" "/usr/local/bin/clang-tidy"
         # Install qt6 only with ubuntu-24.04
         # This page explains why we need libgl1-mesa-dev
         # https://doc-snapshots.qt.io/qt6-dev/linux.html
--- a/cpp/modmesh/universe/polygon.hpp
+++ b/cpp/modmesh/universe/polygon.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
 /*
- * Copyright (c) 2023, Yung-Yu Chen <yyc@solvcon.net>
+ * Copyright (c) 2025, An-Chi Liu <phy.tiger@gmail.com>
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
@@ -33,8 +33,16 @@
  */
 
 #include <modmesh/base.hpp>
+#include <modmesh/buffer/SimpleCollector.hpp>
 #include <modmesh/buffer/buffer.hpp>
+#include <modmesh/universe/bezier.hpp>
 #include <modmesh/universe/coord.hpp>
+#include <modmesh/universe/rtree.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <vector>
 
 namespace modmesh
 {
@@ -652,28 +660,28 @@ public:
             z2(i) = p2.z();
         }
     }
-    void set(size_t i, value_type x0, value_type y0, value_type x1, value_type y1, value_type x2, value_type y2)
+    void set(size_t i, value_type x0_value, value_type y0_value, value_type x1_value, value_type y1_value, value_type x2_value, value_type y2_value)
     {
-        x0(i) = x0;
-        y0(i) = y0;
-        x1(i) = x1;
-        y1(i) = y1;
-        x2(i) = x2;
-        y2(i) = y2;
+        x0(i) = x0_value;
+        y0(i) = y0_value;
+        x1(i) = x1_value;
+        y1(i) = y1_value;
+        x2(i) = x2_value;
+        y2(i) = y2_value;
     }
-    void set(size_t i, value_type x0, value_type y0, value_type z0, value_type x1, value_type y1, value_type z1, value_type x2, value_type y2, value_type z2)
+    void set(size_t i, value_type x0_value, value_type y0_value, value_type z0_value, value_type x1_value, value_type y1_value, value_type z1_value, value_type x2_value, value_type y2_value, value_type z2_value)
     {
-        x0(i) = x0;
-        y0(i) = y0;
-        x1(i) = x1;
-        y1(i) = y1;
-        x2(i) = x2;
-        y2(i) = y2;
+        x0(i) = x0_value;
+        y0(i) = y0_value;
+        x1(i) = x1_value;
+        y1(i) = y1_value;
+        x2(i) = x2_value;
+        y2(i) = y2_value;
         if (ndim() == 3)
         {
-            z0(i) = z0;
-            z1(i) = z1;
-            z2(i) = z2;
+            z0(i) = z0_value;
+            z1(i) = z1_value;
+            z2(i) = z2_value;
         }
     }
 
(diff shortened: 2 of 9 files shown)


cpp/modmesh/universe/polygon.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 /*
- * Copyright (c) 2023, Yung-Yu Chen <yyc@solvcon.net>
+ * Copyright (c) 2025, An-Chi Liu <phy.tiger@gmail.com>
```
tigercosmos, line 4: Fix the copyright. The code was written by me originally (`TrianglePad`)

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,215 @@ class TrianglePad
 using TrianglePadFp32 = TrianglePad<float>;
 using TrianglePadFp64 = TrianglePad<double>;
 
+template <typename T>
+class Polygon3d
```
tigercosmos, line 811: This class will be used later in SVG and polygen operation. In my current design, each line or shape will be a polygen object, and we can operate on multiple polygen with boolean operations.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,215 @@ class TrianglePad
...
+    using curve_pad_type = CurvePad<T>;
+    using rtree_type = RTree<segment_type, BoundBox3d<T>, RTreeValueOps<segment_type, BoundBox3d<T>>>;
+
+    template <typename... Args>
+    static std::shared_ptr<Polygon3d<T>> construct(Args &&... args)
+    {
+        return std::make_shared<Polygon3d<T>>(std::forward<Args>(args)..., ctor_passkey());
+    }
+
+    Polygon3d(std::shared_ptr<segment_pad_type> segments, ctor_passkey const &)
```
tigercosmos: We can already parse the path or shape into segments or curves, just put them as the argument to create a polygon object.

tests/test_polygon3d.py
```diff
@@ -0,0 +1,334 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import unittest
+import modmesh as mm
+from modmesh.testing import TestBase as ModMeshTB
+
+
+class Polygon3dTB(ModMeshTB):
```
tigercosmos, line 32: Some basic operations. Tests should explain themselves.

tigercosmos:
@yungyuc  Please help take a look, thanks!

cpp/modmesh/universe/polygon.hpp
yungyuc: Thank you.  Good catch!

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,215 @@ class TrianglePad
...
+    void build_rtree()
+    {
+        for (size_t i = 0; i < m_segments->size(); ++i)
+        {
+            m_rtree->insert(m_segments->get(i));
+        }
+    }
+
+    std::shared_ptr<segment_pad_type> m_segments; ///< Segments composing the polygon
+    std::unique_ptr<rtree_type> m_rtree; ///< RTree for segments in the polygon
```
yungyuc: No, the tree (spatial index) should be built for all geometric entities that you want to manipulate, but not on a single polygon.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,215 @@ class TrianglePad
...
+
+    void build_rtree()
+    {
+        for (size_t i = 0; i < m_segments->size(); ++i)
+        {
+            m_rtree->insert(m_segments->get(i));
+        }
+    }
+
+    std::shared_ptr<segment_pad_type> m_segments; ///< Segments composing the polygon
```
yungyuc: Storing line segments duplicates memory use.  Store only points (aka node lists).

cpp/modmesh/universe/polygon.hpp
yungyuc: Do not create a class for polygon, because itself is a container.  When dealing with polygons, we always do it with many.

Create `PolygonPad<T>` instead.  Keep polygon information in the pad.

yungyuc (changes requested):
- [x] Do not create a class for a single polygon.  Create `PolygonPad<T>` instead.  Keep polygon information in the pad.
- [x] Store points, not line segments, for polygons.
- [x] Index all geometric entities but not one polygon.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+
+private:
+
+    size_t offset() const { return m_offset; }
+    size_t count() const { return m_count; }
+
+    polygon_pad_type * m_pad = nullptr;
+    size_t m_polygon_id = 0;
+    size_t m_offset = 0;
+    size_t m_count = 0;
```
tigercosmos: `Polygon` is a handle for `PolygonPad`.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+
+        size_t const polygon_id = m_polygon_offsets.size();
+        m_polygon_offsets.push_back(offset);
+        m_polygon_counts.push_back(count);
+
+        polygon_type polygon(this, polygon_id, offset, count);
+        rebuild_polygon_rtree(polygon);
+
+        return polygon;
+    }
```
tigercosmos: When we add a new polygon, add all nodes to the `PointPad`, and record `offset` and `count` to create a `Polygon` handle.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+        {
+            segment_type const edge = get_edge(polygon_id, i);
+            m_rtree->insert(edge);
+        }
+    }
+
+    std::shared_ptr<point_pad_type> m_nodes;
+    std::vector<size_t> m_polygon_offsets;
+    std::vector<size_t> m_polygon_counts;
+    std::unique_ptr<rtree_type> m_rtree;
```
tigercosmos: How we store data in PolygonPad.

tigercosmos (commented):
@yungyuc  The new code is just a draft. Could you help check whether `PolygonPad` and `Polygon` implementation matches your expectation?

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+    {
+        return !(*this == other);
+    }
+
+private:
+
+    size_t offset() const { return m_offset; }
+    size_t count() const { return m_count; }
+
+    polygon_pad_type * m_pad = nullptr;
```
yungyuc: `Polygon` needs to own `PolygonPad` so use shared pointer here.

We may need a non-owning polygon handle helper class for faster processing in the future (like `pybind11::handle` vs `pybind11.object`, but the code should start with an owning handle for memory safety at the moment.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+ * This is a non-owning handle that references a polygon stored in a PolygonPad
+ * container. Polygons are defined by an ordered list of nodes following the
+ * right-hand rule: counter-clockwise for positive area, clockwise for negative.
+ *
+ * The handle uses polygon_id as public API, with internal offset/count for efficient access.
+ *
+ * @tparam T floating-point type
+ */
+template <typename T>
+class Polygon
```
yungyuc: `Polygon` or `Polygon3d`?  I am not sure but incline for `Polygon3d` for (1) consistency with other geometry element classes and (2) leave room for `Polygon2d` in the future.  We need to discuss.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+
+        for (size_t i = 0; i < count; ++i)
+        {
+            segment_type const edge = get_edge(polygon_id, i);
+            m_rtree->insert(edge);
+        }
+    }
+
+    std::shared_ptr<point_pad_type> m_nodes;
+    std::vector<size_t> m_polygon_offsets;
```
yungyuc: Use `SimpleCollector<ssize_type>` instead of `std::vector<size_t>` for lookup tables.

Define `ssize_type` as `int32_t` at the moment.  Two key points: (1) use 32-bits for indexing for now to save memory and (2) use signed integer for indexing so that we can use negative integers for sentinels and other conveniences.

We do not respect STL convention for indexing integers in modmesh.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+    std::vector<size_t> m_polygon_counts;
```
yungyuc: Replace `m_polygon_offsets` and `m_polygon_counts` with `m_begins` (index at the first point in the polygon) and `m_ends` (index after the last point in the polygon).

It has no disadvantage but 3 benefits: (1) the lookup tables are symmetric, (2) the lookup tables are consistent to C++ iterator convention, and (3) getting count is instant like `m_ends[i] - m_begin[i]` (with a functioning cache it takes no time).

cpp/modmesh/universe/polygon.hpp
yungyuc: By using `SimpleCorrector`, our life will be happy when syncing the data to GPU.

You can't manage the memory buffer under `std::vector` which defines no ownership against its memory buffer (`std::vector::data()`).

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+        size_t const count = polygon.count();
+
+        for (size_t i = 0; i < count; ++i)
+        {
+            segment_type const edge = get_edge(polygon_id, i);
+            m_rtree->insert(edge);
+        }
+    }
+
+    std::shared_ptr<point_pad_type> m_nodes;
```
yungyuc: Rename `m_nodes` to `m_points` for consistency with the type names of `Point3d` and `PointPad`.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+    Polygon(Polygon const &) = default;
+    Polygon(Polygon &&) = default;
+    Polygon & operator=(Polygon const &) = default;
+    Polygon & operator=(Polygon &&) = default;
+    ~Polygon() = default;
+
+    size_t polygon_id() const { return m_polygon_id; }
+    polygon_pad_type * pad() const { return m_pad; }
+
+    size_t num_nodes() const;
```
yungyuc: Rename `num_nodes()` to `nnode()`.  Check the existing code in [`StaticMesh.hpp`](https://github.com/solvcon/modmesh/blob/04680baa30f2a823bef320fc0a03caec2ad8664a/cpp/modmesh/mesh/StaticMesh.hpp).

Of course, the code in `StaticMesh.hpp` is quite old and need revamping, but it has a convention already.  Let's develop code based on it.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+    uint8_t ndim() const;
+    point_type node(size_t index) const;
+    segment_type edge(size_t index) const;
+    value_type compute_signed_area() const;
+    bool is_counter_clockwise() const { return compute_signed_area() > 0; }
```
yungyuc: Factor `is_counter_clockwise()` into two functions: (1) `calc_counter_clockwise()` and (2) `is_counter_clockwise()`, like:

```cpp
// You can use your judgement for naming.
bool calc_counter_clockwise() const
{
    // BE VERY CAREFUL ABOUT !!! ZERO !!!
    m_is_counter_clockwise = compute_signed_area() >= 0;
    return m_is_counter_clockwise;
}
bool is_counter_clockwise() const
{
    if (!m_calced_counter_clockwise)
    {
        calc_counter_clockwise();
        m_calced_counter_clockwise = true;
    }
    return m_is_counter_clockwise;
}
```

Recalculating the counter-clockwiseness and using the cached results are both commonplace.

Also note that zero-area polygons are tricky.  They usually consist of two line segments as an Ouroboros and the line segments are also usually zero length.  This is just a heads-up.  You can only know how to deal with them when seeing them in real cases.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+        return !(*this == other);
+    }
+
+private:
+
+    size_t offset() const { return m_offset; }
+    size_t count() const { return m_count; }
+
+    polygon_pad_type * m_pad = nullptr;
+    size_t m_polygon_id = 0;
```
yungyuc: Rename `Polygon::m_polygon_id` as `id`.  Of course it is polygon id and you don't need to repeat.

cpp/modmesh/universe/polygon.hpp
yungyuc: Beautiful.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +787,577 @@ class TrianglePad
...
+
+    size_t num_nodes() const;
+    uint8_t ndim() const;
+    point_type node(size_t index) const;
+    segment_type edge(size_t index) const;
+    value_type compute_signed_area() const;
+    bool is_counter_clockwise() const { return compute_signed_area() > 0; }
+    BoundBox3d<T> calc_bound_box() const;
+
+    bool operator==(Polygon const & other) const
```
yungyuc: Make `operator==()` and `operator!=()` compare the coordinate, and define `Polygon3d::is()` and `is_not()` for identity test.

We need coordinate test, and the equality operator should be used for it.

If you don't need the equality operator at the moment, just rename them to `is()` and `is_not()`, and leave code comments to remind for future implementation.  It will be even better if the comment includes an issue to track it.

yungyuc (changes requested):
> @yungyuc The new code is just a draft. Could you help check whether `PolygonPad` and `Polygon` implementation matches your expectation?

The quality of the code so far is beyond a draft, or at least a very good draft.

There are many points to address because we are moving right in the direction:

- [x] Discuss whether to name it `Polygon` or `Polygon3d`.
- [x] `Polygon` needs to hold a `shared_ptr` to `PolygonPad`.
- [x] Rename `num_nodes()` to `nnode()`.
- [x] Factor `is_counter_clockwise()` into two functions: (1) `calc_counter_clockwise()` and (2) `is_counter_clockwise()`.
- [x] Rename `Polygon3d::m_polygon_id` as `id`.
- [x] Make `operator==()` and `operator!=()` compare the coordinate, and define `Polygon3d::is()` for identity test.
- [x] Use `SimpleCollector<ssize_type>` instead of `std::vector<size_t>` for lookup tables.
- [x] Replace `m_polygon_offsets` and `m_polygon_counts` with `m_begins` (index at the first point in the polygon) and `m_ends` (index after the last point in the polygon).

cpp/modmesh/universe/polygon.hpp
tigercosmos: Got it.

cpp/modmesh/universe/polygon.hpp
tigercosmos: Got it.

cpp/modmesh/universe/polygon.hpp
tigercosmos: Added `is` and `is_not` and modified `==` and `=!`

cpp/modmesh/universe/polygon.hpp
tigercosmos: chnaged to `m_id`

cpp/modmesh/universe/polygon.hpp
tigercosmos: Let's use Polygon3d

cpp/modmesh/universe/polygon.hpp
tigercosmos: ok

cpp/modmesh/universe/polygon.hpp
tigercosmos: Got it. Smart!

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+ * This is a non-owning handle that references a polygon stored in a PolygonPad
+ * container. Polygons are defined by an ordered list of nodes following the
+ * right-hand rule: counter-clockwise for positive area, clockwise for negative.
+ *
+ * The handle uses polygon_id as public API, with internal offset/count for efficient access.
+ *
+ * @tparam T floating-point type
+ */
+template <typename T>
+class Polygon3d
```
tigercosmos, line 811: Named to `Polygon3d`

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+private:
+    bool calc_counter_clockwise() const
+    {
+        // Be careful for the zero area case
+        m_is_counter_clockwise = compute_signed_area() >= 0;
+        return m_is_counter_clockwise;
+    }
+
+private:
+    std::shared_ptr<polygon_pad_type> m_pad = nullptr;
```
tigercosmos: Held shared pointer of `PolygonPad`

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+            min_z = std::min(min_z, node.z());
+            max_x = std::max(max_x, node.x());
+            max_y = std::max(max_y, node.y());
+            max_z = std::max(max_z, node.z());
+        }
+
+        return BoundBox3d<T>(min_x, min_y, min_z, max_x, max_y, max_z);
+    }
+
+    bool operator==(Polygon3d const & other) const
```
tigercosmos, line 884: Compared to nodes now.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+        }
+        return true;
+    }
+
+    bool operator!=(Polygon3d const & other) const
+    {
+        return !(*this == other);
+    }
+
+    bool is(Polygon3d const & other) const
```
tigercosmos: New added `is` and `is_not`

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+    Polygon3d(Polygon3d const &) = default;
+    Polygon3d(Polygon3d &&) = default;
+    Polygon3d & operator=(Polygon3d const &) = default;
+    Polygon3d & operator=(Polygon3d &&) = default;
+    ~Polygon3d() = default;
+
+    size_t polygon_id() const { return m_id; }
+    std::shared_ptr<polygon_pad_type> pad() const { return m_pad; }
+
+    size_t nnode() const
```
tigercosmos, line 851: Renamed to `nnode`

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+    bool calc_counter_clockwise() const
+    {
+        // Be careful for the zero area case
+        m_is_counter_clockwise = compute_signed_area() >= 0;
+        return m_is_counter_clockwise;
+    }
+
+private:
+    std::shared_ptr<polygon_pad_type> m_pad = nullptr;
+    size_t m_id = 0;
```
tigercosmos, line 922: Renamed to `m_id`

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+        return m_pad == other.m_pad && m_id == other.m_id;
+    }
+
+    bool is_not(Polygon3d const & other) const
+    {
+        return !is(other);
+    }
+
+private:
+    bool calc_counter_clockwise() const
```
tigercosmos, line 913: New added for lazy evalution of counter clockwise.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+        for (size_t i = 0; i < count; ++i)
+        {
+            segment_type const edge = polygon.edge(i);
+            m_rtree->insert(edge);
+        }
+    }
+
+    std::unordered_map<size_t, polygon_type> m_polygons;
+    std::shared_ptr<point_pad_type> m_points;
+    SimpleCollector<ssize_type> m_begins;
```
tigercosmos, line 1139: Used `SimpleCollector` for lookup table.

tigercosmos: Used `m_begins` instead of `m_offset`.

tigercosmos (commented):
@yungyuc  all issues above are fixed. Please take a look.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    bool operator!=(Polygon3d const & other) const
+    {
+        return !(*this == other);
```
yungyuc: For speed, we may need to spell out the not-equal-to in the future.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+        size_t const count = polygon.nnode();
+
+        for (size_t i = 0; i < count; ++i)
+        {
+            segment_type const edge = polygon.edge(i);
+            m_rtree->insert(edge);
+        }
+    }
+
+    std::unordered_map<size_t, polygon_type> m_polygons;
```
yungyuc: Do not store `Polygon3d`.  Always create the handle class `Polygon3d` on the fly.

Hash table is a killer of runtime and memory.  Do not use it unless (1) you have profiling results, (2) convincing prior experience that it is faster than LUT, or (3) you do not care about performance.  Here we put performance first.

`m_begins` and `m_ends` alone defines all polygons in this pad.  We will need other LUT in the future, but the two suffice for now.

cpp/modmesh/universe/polygon.hpp
yungyuc: Good.  And `Polygon3d` should keep `ssize_t m_begin, m_end` for the node list.

`Polygon3d` is a handle object, and it should not use an indirect other than LUT in `PolygonPad`.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+    {
+        if (nnode() != other.nnode())
+        {
+            return false;
+        }
+        if (ndim() != other.ndim())
+        {
+            return false;
+        }
+        size_t const node_count = nnode();
```
yungyuc: The local variable `node_count` may be kept in the beginning of the function as a cache.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+    using segment_type = Segment3d<T>;
+    using polygon_pad_type = PolygonPad<T>;
+
+    Polygon3d(std::shared_ptr<polygon_pad_type> pad, size_t polygon_id)
+        : m_pad(std::move(pad))
+        , m_id(polygon_id)
+    {
+    }
+
+    Polygon3d() = default;
```
yungyuc: Default constructor of `Polygon3d` does not make sense.  It should be deleted.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+
+    using point_type = Point3d<T>;
+    using value_type = T;
+    using segment_type = Segment3d<T>;
+    using polygon_pad_type = PolygonPad<T>;
+
+    Polygon3d(std::shared_ptr<polygon_pad_type> pad, size_t polygon_id)
+        : m_pad(std::move(pad))
+        , m_id(polygon_id)
+    {
```
yungyuc, line 831: Check for `m_id` range.  Throw `std::out_of_range` when it is beyond the size of `pad`.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+{
+
+public:
+
+    using point_type = Point3d<T>;
+    using value_type = T;
+    using segment_type = Segment3d<T>;
+    using polygon_pad_type = PolygonPad<T>;
+
+    Polygon3d(std::shared_ptr<polygon_pad_type> pad, size_t polygon_id)
```
yungyuc: Add passkey to `Polygon3d` constructor so that only `PolygonPad` can construct it.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+        typename polygon_pad_type::ssize_type const end_index = m_pad->m_ends[m_id];
+        return static_cast<size_t>(end_index - begin_index);
+    }
+
+    uint8_t ndim() const
+    {
+        return m_pad->ndim();
+    }
+
+    point_type node(size_t index) const
```
yungyuc: Most code in `Polygon3d::node()` should be moved to `PolygonPad()` and the former just calls the latter.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+
+    size_t polygon_id() const { return m_id; }
+    std::shared_ptr<polygon_pad_type> pad() const { return m_pad; }
+
+    size_t nnode() const
+    {
+        if (m_id >= m_pad->m_begins.size())
+        {
+            throw std::out_of_range(
+                std::format("Polygon3d::nnode: polygon_id {} >= num_polygons {}",
```
yungyuc: Good to have `std::format`.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+                            m_id,
+                            m_pad->m_begins.size()));
+        }
+        typename polygon_pad_type::ssize_type const begin_index = m_pad->m_begins[m_id];
+        typename polygon_pad_type::ssize_type const end_index = m_pad->m_ends[m_id];
+        return static_cast<size_t>(end_index - begin_index);
+    }
+
+    uint8_t ndim() const
```
yungyuc: If

```cpp
uint8_t ndim() const { return m_pad->ndim(); }
```

works, use the one-liner format.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +788,617 @@ class TrianglePad
...
+                std::format("Polygon3d::edge: index {} >= count {}",
+                            index,
+                            count));
+        }
+        point_type const p0 = m_pad->m_points->get(static_cast<size_t>(begin_index + static_cast<typename polygon_pad_type::ssize_type>(index)));
+        point_type const p1 = m_pad->m_points->get(static_cast<size_t>(begin_index + static_cast<typename polygon_pad_type::ssize_type>((index + 1) % count)));
+        return segment_type(p0, p1);
+    }
+
+    value_type compute_signed_area() const
```
yungyuc: `Polygon3d::compute_signed_area()` is too long and has a loop.  It should be moved outside the class declaration.

modmesh/core.py
```diff
@@ -159,6 +161,12 @@
     'CurvePadFp64',
     'WorldFp32',
     'WorldFp64',
+    'PolygonFp32',
```
yungyuc: Why `PolygonFp??` are still exposed from C++?

yungyuc (changes requested):
Good progress.  Points to address:

- [x] Delete the default constructor of `Polygon3d` because it does not make sense.
- [x] Check for `m_id` range in `Polygon3d` constructor.
- [x] Add passkey to `Polygon3d` constructor so that only `PolygonPad` can construct it.
- [x] Most code in `Polygon3d::nnode()` should be moved to `PolygonPad::nnode()` and the former just calls the latter.  So is that of `Polygon3d::node()` and `Polygon3d::edge()`.
- [x] `Polygon3d::compute_signed_area()` is too long and has a loop.  It should be moved outside the class declaration.  Do the same to `Polygon3d::calc_bound_box()` and other functions like that.
- [x] Do not store `Polygon3d` and do not use hash table.  Create the handle object `Polygon3d` on the fly and keep the nodes of begin and end in it directly.
- [x] Why `PolygonFp??` are still exposed from C++?

cpp/modmesh/universe/polygon.hpp
tigercosmos: One concern here is that we already calculated `m_is_counter_clockwise` in `Polygon3d`. If we always create the polygon on the fly, then we may need to recalculate the value every time. However, maybe we rarely need to do the calculation, and it may not be a problem.

cpp/modmesh/universe/polygon.hpp
tigercosmos: Thanks for catching this!

cpp/modmesh/universe/polygon.hpp
tigercosmos: Deleted

cpp/modmesh/universe/polygon.hpp
tigercosmos: It should never happen though, but we can add it just in case.

cpp/modmesh/universe/polygon.hpp
tigercosmos: done

cpp/modmesh/universe/polygon.hpp
tigercosmos: moved.

cpp/modmesh/universe/polygon.hpp
tigercosmos: moved.

cpp/modmesh/universe/polygon.hpp
tigercosmos: Sure

modmesh/core.py
tigercosmos: We can get the `Polygon3d` object from calling `PolyganPad::get_polygon`.

cpp/modmesh/universe/polygon.hpp
tigercosmos: Removed the map.

tigercosmos:
@yungyuc All issues memtioned are addressed.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +756,425 @@ class TrianglePad
...
+    ~Polygon3d() = default;
+
+    size_t polygon_id() const { return m_id; }
+    std::shared_ptr<polygon_pad_type> pad() const { return m_pad; }
+
+    size_t nnode() const
+    {
+        if (m_id >= m_pad->m_begins.size())
+        {
+            throw std::out_of_range(std::format("Polygon3d::nnode: polygon_id {} >= num_polygons {}",
```
yungyuc, line 855: `nnode()` should not throw an exception, but we may have `nnode(bool check_for_range)` to check for range.  Python wrapper should call the latter.

We don't need to do it in this PR.  I leave this comment only for future reference.

modmesh/core.py
yungyuc: Shouldn't the return from `PolygonPad::get_polygon()` be `Polygon3dFp??`?  I still do not see why `PolygonFp??` should be in Python.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +756,425 @@ class TrianglePad
...
+    //  - Trapezoid decomposition structure
+    //  - Point location query structure
+    //  - Integration with node lists for synchronization
+
+}; /* end class PolygonPad */
+
+using PolygonPadFp32 = PolygonPad<float>;
+using PolygonPadFp64 = PolygonPad<double>;
+
+using PolygonFp32 = Polygon3d<float>;
```
yungyuc: I don't think we need `PolygonFp??`.  We do not have `BezierFp??`.

cpp/modmesh/universe/polygon.hpp
yungyuc: > One concern here is that we already calculated `m_is_counter_clockwise` in `Polygon3d`. If we always create the polygon on the fly, then we may need to recalculate the value every time. However, maybe we rarely need to do the calculation, and it may not be a problem.

We should always recalculate `m_is_counter_clockwise` because the polygon vertices may **move**.

For performance we may want to freeze `PolygonPad` and the polygons.  That will require a well-thought mechanism of freezing.  Not what we should do at the moment.

But keep in mind that even when we do freezing and caching, we use LUT, not hash table.

yungyuc (changes requested):
I don't see why we need `PolygonFp??` in C++ and Python.  Is there a reason?

cpp/modmesh/universe/polygon.hpp
tigercosmos: I see your point. Fixed the naming.

tigercosmos:
@yungyuc  Fixed the naming issue.

modmesh/core.py
```diff
@@ -157,6 +159,12 @@
     'CurvePadFp64',
     'WorldFp32',
     'WorldFp64',
+    'Polygon3dFp32',
+    'Polygon3dFp64',
+    'PolygonPadFp32',
+    'PolygonPadFp64',
+    'Polygon3dFp32',
```
yungyuc, line 164: Duplicate with lines 162 and 163.

cpp/modmesh/universe/pymod/wrap_polygon.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+void wrap_polygon(pybind11::module & mod)
+{
+    WrapPolygon<float>::commit(mod, "Polygon3dFp32", "Polygon3dFp32");
+    WrapPolygon<double>::commit(mod, "Polygon3dFp64", "Polygon3dFp64");
+
+    WrapPolygonPad<float>::commit(mod, "PolygonPadFp32", "PolygonPadFp32");
+    WrapPolygonPad<double>::commit(mod, "PolygonPadFp64", "PolygonPadFp64");
+
+    mod.attr("Polygon3dFp32") = mod.attr("PolygonPadFp32");
```
yungyuc: This is plainly wrong.  What kind of stupid AI did this?

yungyuc (changes requested):

modmesh/core.py
tigercosmos: Fixed.

cpp/modmesh/universe/pymod/wrap_polygon.cpp
tigercosmos: Fixed.

.github/workflows/lint.yml
```diff
@@ -78,14 +78,15 @@ jobs:
 
     - name: dependency by apt
       run: |
-        VERSION_ID=$(bash -c 'source /etc/os-release ; echo $VERSION_ID')
-        if [ "20.04" == "$VERSION_ID" ] ; then CLANG_TIDY_VERSION=10 ; else CLANG_TIDY_VERSION=18 ; fi
+        # Install clang-tidy-21 from LLVM repository for full C++23 support
```
tigercosmos, line 81: Upgrade to clang-tidy 21 to resolve the build issue on Ubuntu. The version aligns with the one that macOS uses.

cpp/modmesh/universe/polygon.cpp
```diff
@@ -0,0 +1,361 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/universe/polygon.hpp>
+
+namespace modmesh
+{
+
+template <typename T>
+SimpleArray<T> TrianglePad<T>::pack_array() const
```
tigercosmos: Move this funciton in cpp file becuase it's big.

cpp/modmesh/universe/polygon.cpp
```diff
@@ -0,0 +1,361 @@
...
+    {
+        throw std::invalid_argument("PolygonPad::add_polygon_from_segments: empty segment pad");
+    }
+    if (segments->ndim() != ndim())
+    {
+        throw std::invalid_argument(
+            std::format("PolygonPad::add_polygon_from_segments: segments.ndim() {} != pad.ndim() {}",
+                        int(segments->ndim()),
+                        int(ndim())));
+    }
```
tigercosmos: More checks.

cpp/modmesh/universe/polygon.cpp
```diff
@@ -0,0 +1,361 @@
...
+    {
+        throw std::invalid_argument("PolygonPad::add_polygon_from_segments_and_curves: curves is null");
+    }
+    if (!(sample_length > 0))
+    {
+        throw std::invalid_argument(
+            std::format("PolygonPad::add_polygon_from_segments_and_curves: sample_length {} must be > 0",
+                        sample_length));
+    }
+    if (segments->ndim() != ndim() || curves->ndim() != ndim())
```
tigercosmos: More checks.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -652,28 +628,28 @@ class TrianglePad
...
-        y2(i) = y2;
+        x0(i) = x0_value;
+        y0(i) = y0_value;
+        x1(i) = x1_value;
+        y1(i) = y1_value;
+        x2(i) = x2_value;
+        y2(i) = y2_value;
     }
-    void set(size_t i, value_type x0, value_type y0, value_type z0, value_type x1, value_type y1, value_type z1, value_type x2, value_type y2, value_type z2)
+    void set(size_t i, value_type x0_value, value_type y0_value, value_type z0_value, value_type x1_value, value_type y1_value, value_type z1_value, value_type x2_value, value_type y2_value, value_type z2_value)
```
tigercosmos, line 672: Error that we didn't catch before. Since we move code into cpp, the error occurs at compilation.

cpp/modmesh/universe/polygon.hpp
```diff
@@ -781,6 +757,402 @@ class TrianglePad
...
+
+    using ssize_type = int32_t;
+    using point_type = Point3d<T>;
+    using value_type = T;
+    using segment_type = Segment3d<T>;
+    using point_pad_type = PointPad<T>;
+    using polygon_type = Polygon3d<T>;
+    using segment_pad_type = SegmentPad<T>;
+    using curve_pad_type = CurvePad<T>;
+    using rtree_type = RTree<segment_type, BoundBox3d<T>, RTreeValueOps<segment_type, BoundBox3d<T>>>;
```
tigercosmos, line 970: Use case of Rtree.

modmesh/core.py
```diff
@@ -157,6 +159,10 @@
     'CurvePadFp64',
     'WorldFp32',
     'WorldFp64',
+    'PolygonPadFp32',
+    'PolygonPadFp64',
+    'Polygon3dFp32',
+    'Polygon3dFp64',
```
tigercosmos, line 165: Export correct types now.

tests/test_polygon3d.py
tigercosmos: I have checked the test cases again. The AI generated cases should be good.

tigercosmos (commented):
@yungyuc  The PR is ready for review.

.github/workflows/lint.yml
yungyuc: Upgrading clang-tidy deserves a distinct PR, but it's OK to piggyback for convenience at the moment.

cpp/modmesh/universe/polygon.cpp
yungyuc: You can't move templates outside include files unless you only allow selected specializations.  In our case, we want all possible type input to the template argument.

For now, `polygon.hpp` is not very long.  All polygon and rtree code can live there.  In the future, when code grows, rtree code can be factored out to distinct files.

Another trick to play (but not now) it to split declaration and implementation in different header files, like `polygon_decl.hpp` and `polygon_impl.hpp`.

yungyuc (changes requested):
- [x] Templates needs to be placed in header/include file.

cpp/modmesh/universe/polygon.cpp
tigercosmos: @yungyuc Sorry about this. I have moved the fucntion implementation back to the header file, and removed the cpp file.

tigercosmos:
@yungyuc I resolved the issue above. I hope this time it's good enough...

yungyuc (approved):
LGTM

yungyuc:
@tigercosmos could you please rebase to squash "add new Polygon3d for later support of Polygon operation" 65359035d12a8ce05331cc0079014451825ecb43 with "delete polygon.cpp and move all template function implementations into headers" e7553c78ac0ab971c0b60352007479d165c7c17a ?  "fix clang-tidy" b58ab80993a5ea25e8da24d24542c426dc90a0df can remain as a distinct changeset.

yungyuc:
@tigercosmos I did the squashing for you at https://github.com/yungyuc/modmesh/tree/polygon1125 but did not update your branch, because you can beef up the commit log.  Let me know what do you think.

tigercosmos:
@yungyuc I have rebased the commits by myself and modified the commit messages, but still thanks for the help anyway.

PR #634, opened 2025-11-19
Temporarily turns off profiling run on macos-15

yungyuc:
`make pyprof` takes forever and make `build_macos-15_Release` fails after running for 6 hours:

```
Run make pyprof
  make pyprof
  shell: /bin/bash -e {0}
  env:
    QT_DEBUG_PLUGINS: 1
    PIP_BREAK_SYSTEM_PACKAGES: 1
    AQT_CONFIG: thirdparty/aqt_settings.ini
    PKG_CONFIG_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/lib/pkgconfig
    QT_ROOT_DIR: /Users/runner/work/modmesh/Qt/6.8.1/macos
    QT_PLUGIN_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/plugins
    QML2_IMPORT_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/qml
cmake --build build/dev314 --target _modmesh_py VERBOSE= -j
[ 96%] Built target modmesh_primary
[100%] Built target _modmesh
[100%] Built target _modmesh_py
/opt/homebrew/bin/python3 profiling/profile_arithmetic_simd.py > profiling/results/profile_arithmetic_simd.output
/Users/runner/work/modmesh/modmesh/profiling/profile_arithmetic_simd.py:111: RuntimeWarning: divide by zero encountered in divide
  return np.divide(src1, src2)
/opt/homebrew/bin/python3 profiling/profile_arithmetic.py > profiling/results/profile_arithmetic.output
/opt/homebrew/bin/python3 profiling/profile_median.py > profiling/results/profile_median.output
/opt/homebrew/bin/python3 profiling/profile_statistic_ops.py > profiling/results/profile_statistic_ops.output
Matplotlib is building the font cache; this may take a moment.
/opt/homebrew/bin/python3 profiling/profile_take_along_axis.py > profiling/results/profile_take_along_axis.output
Error: The operation was canceled.
```

A CI run history can be found at: https://github.com/solvcon/modmesh/actions/runs/19475614444/job/55734494567

PR #641, opened 2025-11-25
fix `make pyprof` hang issue

### LINKED ISSUE #635: `make pyprof` runs super slow on `macos-15` GA runner
PR #634 temporarily turns off `make pyprof` step in `build_macos-15_Release` because one or more profiling scripts takes super long time, like hours:

```
Run make pyprof
  make pyprof
  shell: /bin/bash -e {0}
  env:
    QT_DEBUG_PLUGINS: 1
    PIP_BREAK_SYSTEM_PACKAGES: 1
    AQT_CONFIG: thirdparty/aqt_settings.ini
    PKG_CONFIG_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/lib/pkgconfig
    QT_ROOT_DIR: /Users/runner/work/modmesh/Qt/6.8.1/macos
    QT_PLUGIN_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/plugins
    QML2_IMPORT_PATH: /Users/runner/work/modmesh/Qt/6.8.1/macos/qml
cmake --build build/dev314 --target _modmesh_py VERBOSE= -j
[ 96%] Built target modmesh_primary
[100%] Built target _modmesh
[100%] Built target _modmesh_py
/opt/homebrew/bin/python3 profiling/profile_arithmetic_simd.py > profiling/results/profile_arithmetic_simd.output
/Users/runner/work/modmesh/modmesh/profiling/profile_arithmetic_simd.py:111: RuntimeWarning: divide by zero encountered in divide
  return np.divide(src1, src2)
/opt/homebrew/bin/python3 profiling/profile_arithmetic.py > profiling/results/profile_arithmetic.output
/opt/homebrew/bin/python3 profiling/profile_median.py > profiling/results/profile_median.output
/opt/homebrew/bin/python3 profiling/profile_statistic_ops.py > profiling/results/profile_statistic_ops.output
Matplotlib is building the font cache; this may take a moment.
/opt/homebrew/bin/python3 profiling/profile_take_along_axis.py > profiling/results/profile_take_along_axis.output
Error: The operation was canceled.
```

@KHLee529 [suggested](https://github.com/solvcon/modmesh/pull/634#issuecomment-3549993190):

> From the error message it seems like that `profile_take_along_axis.py` is the script that cause infinite run. Could you please try to first remove this script by renaming it to any name without prefix `profile_`?

We need to fix the bad profiling script (or SIMD code).

nccvector:
## Relates to & fixes Issue
https://github.com/solvcon/modmesh/issues/635

## Overview
This PR contains safe guards for:
- passing empty list [] to SimpleArray
- array large enough for simd in `check_between` function in `neon.hpp` <- (array size can be incompatible with neon simd lane size causing a segfault)

# How I discovered a solution
Out of curiosity I just created the gtest to test the `take_along_axis_simd` function directly and found segfaults. I just fixed the segfaults and ran `make pyprof` again just to see what happens and I observed that `make pyprof` no longer hanged.

# I tried to reproduce the hang
Then I got curious and tried to reproduce the bug so I:
- removed the segfault safeguards (reverted my changes)
- Then I created a new python script to reproduce the hanging issue and called the simd function, but it still did not hang (gave the same segfault as gtest does)
- Then I added the `modmesh.call_profiler` similar to the actual profiling script and viola it hanged.

So to reproduce you need:
- a segfault
- python script with profiler

# More info
It seems passing the array smaller than N_lane size causes infinite recursion (because the first for loop is skipped but there is a recursive call right below that loop which calls the function again with the same array). That infinite recursion leads to stack corruption which causes the segfault. I still have not confirmed why running under profiler makes any difference.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -1874,6 +1874,11 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis_simd(SimpleArray<I> const 
                                              << " but the array is " << athis->ndim() << " dimension");
     }
 
+    if (indices.size() == 0)
+    {
+        return A(indices.shape());
+    }
+
     size_t max_idx = athis->shape()[0];
 
     I const * oor_ptr = check_index_range(indices, max_idx);
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -115,6 +115,13 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     T const * ret = NULL;
 
     T const * ptr = start;
+
+    // Check if array is large enough for SIMD processing
+    if (end - start < N_lane)
+    {
+        return generic::check_between<T>(start, end, min_val, max_val);
+    }
+
     for (; ptr <= end - N_lane; ptr += N_lane)
     {
         data_vec = vld1q(ptr);
--- a/gtests/CMakeLists.txt
+++ b/gtests/CMakeLists.txt
@@ -32,6 +32,7 @@ add_executable(
     ${MODMESH_BUFFER_SOURCES}
     ${MODMESH_SERIALIZATION_SOURCES}
     ${MODMESH_TRANSFORM_SOURCES}
+    ${MODMESH_SIMD_SOURCES}
 )
 
 target_link_libraries(
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -404,11 +404,9 @@ jobs:
           fi
           make pytest ${JOB_MAKE_ARGS}
 
-      # FIXME: Temporarily turns off profiling run on macos-15
-      # because it takes 6 hours to fail
-      #- name: make pyprof
-      #  run: |
-      #    make pyprof
+      - name: make pyprof
+        run: |
+          make pyprof
 
       - name: make buildext BUILD_QT=ON USE_PYTEST_HELPER_BINDING=OFF
         run: |
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -167,4 +167,194 @@ TEST(small_vector, select_kth_random)
         EXPECT_EQ(it, i);
     }
 }
+
+TEST(TakeAlongAxisSimd, basic_int32)
+{
+    using namespace modmesh;
+
+    // Create a simple array with values [10, 20, 30, 40, 50]
+    SimpleArray<int32_t> data(small_vector<size_t>{5});
+    data[0] = 10;
+    data[1] = 20;
+    data[2] = 30;
+    data[3] = 40;
+    data[4] = 50;
+
+    // Create indices [2, 0, 4, 1]
+    SimpleArray<uint64_t> indices(small_vector<size_t>{4});
+    indices[0] = 2;
+    indices[1] = 0;
+    indices[2] = 4;
+    indices[3] = 1;
+
+    // Call take_along_axis_simd
+    SimpleArray<int32_t> result = data.take_along_axis_simd(indices);
+
+    // Verify the result
+    EXPECT_EQ(result.size(), 4);
+    EXPECT_EQ(result[0], 30);
+    EXPECT_EQ(result[1], 10);
+    EXPECT_EQ(result[2], 50);
+    EXPECT_EQ(result[3], 20);
+}
+
+TEST(TakeAlongAxisSimd, basic_float64)
+{
+    using namespace modmesh;
+
+    // Create a simple array with float values
+    SimpleArray<double> data(small_vector<size_t>{6});
+    data[0] = 1.5;
+    data[1] = 2.5;
+    data[2] = 3.5;
+    data[3] = 4.5;
+    data[4] = 5.5;
+    data[5] = 6.5;
+
+    // Create indices [5, 2, 0, 3]
+    SimpleArray<uint64_t> indices(small_vector<size_t>{4});
+    indices[0] = 5;
+    indices[1] = 2;
+    indices[2] = 0;
+    indices[3] = 3;
+
+    // Call take_along_axis_simd
+    SimpleArray<double> result = data.take_along_axis_simd(indices);
+
+    // Verify the result
+    EXPECT_EQ(result.size(), 4);
+    EXPECT_DOUBLE_EQ(result[0], 6.5);
+    EXPECT_DOUBLE_EQ(result[1], 3.5);
+    EXPECT_DOUBLE_EQ(result[2], 1.5);
+    EXPECT_DOUBLE_EQ(result[3], 4.5);
+}
+
+TEST(TakeAlongAxisSimd, large_array)
+{
+    using namespace modmesh;
+
+    // Create a larger array
+    const size_t data_size = 1000;
+    SimpleArray<int64_t> data(small_vector<size_t>{data_size});
+    for (size_t i = 0; i < data_size; ++i)
+    {
+        data[i] = static_cast<int64_t>(i * 10);
+    }
+
+    // Create indices that sample from the array
+    const size_t indices_size = 100;
+    SimpleArray<uint64_t> indices(small_vector<size_t>{indices_size});
+    for (size_t i = 0; i < indices_size; ++i)
+    {
+        indices[i] = i * 10; // Sample every 10th element
+    }
+
+    // Call take_along_axis_simd
+    SimpleArray<int64_t> result = data.take_along_axis_simd(indices);
+
+    // Verify the result
+    EXPECT_EQ(result.size(), indices_size);
+    for (size_t i = 0; i < indices_size; ++i)
+    {
+        EXPECT_EQ(result[i], static_cast<int64_t>(i * 10 * 10));
+    }
+}
+
+TEST(TakeAlongAxisSimd, out_of_range)
+{
+    using namespace modmesh;
+
+    // Create a simple array
+    SimpleArray<int32_t> data(small_vector<size_t>{5});
+    data[0] = 10;
+    data[1] = 20;
+    data[2] = 30;
+    data[3] = 40;
+    data[4] = 50;
+
+    // Create indices with out-of-range value
+    SimpleArray<uint64_t> indices(small_vector<size_t>{3});
+    indices[0] = 2;
+    indices[1] = 10; // Out of range
+    indices[2] = 1;
+
+    // Should throw an exception
+    EXPECT_THROW(data.take_along_axis_simd(indices), std::out_of_range);
+}
+
+TEST(TakeAlongAxisSimd, empty_indices)
+{
+    using namespace modmesh;
+
+    // Create a simple array
+    SimpleArray<int32_t> data(small_vector<size_t>{5});
+    data[0] = 10;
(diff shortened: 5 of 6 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1874,6 +1874,11 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis_simd(SimpleArray<I> const
                                              << " but the array is " << athis->ndim() << " dimension");
     }
 
+    if (indices.size() == 0)
+    {
+        return A(indices.shape());
+    }
+
```
nccvector, line 1881: Unrelated: Fixes segfault if we pass empty list [] (unrelated edge case, not the meat of this PR)

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -115,6 +115,13 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     T const * ret = NULL;
 
     T const * ptr = start;
+
+    // Check if array is large enough for SIMD processing
+    if (end - start < N_lane)
+    {
+        return generic::check_between<T>(start, end, min_val, max_val);
+    }
+
```
nccvector, line 124: Meat of this PR: if the array is smaller than lane size then the loop below goes on forever

gtests/CMakeLists.txt
```diff
@@ -46,3 +46,19 @@ gtest_discover_tests(test_nopython)
...
+    test_take_along_axis_simd_native.cpp
+    ${MODMESH_TOGGLE_SOURCES}
+    ${MODMESH_BUFFER_SOURCES}
+    ${MODMESH_SIMD_SOURCES}
+)
+
+target_link_libraries(
+    test_take_along_axis_simd_native
+    GTest::gtest_main
+)
```
nccvector: I just thought it might be a good idea to add gtest to test the code directly without bindings first

yungyuc:
Thank you, @nccvector , when you are ready for review, please also turn the profiling run back on (reverting change in PR #634).

KHLee529:
Cool, so the reason of infinite run is actually segmentation fault? I'll wait for the description for more details.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: It's nice to just but I wonder if we should raise directly?

cpp/modmesh/simd/neon/neon.hpp
tigercosmos: why not return error directly? Also I wonder why this can ever happen because we have checked the alignment using `check_alignment`?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: It makes sense to pass this corner case silently because empty input should be equivalent to no-op.  No-op should not throw exception.

cpp/modmesh/simd/neon/neon.hpp
yungyuc: A single element in an array and having less elements in a vector register (or a cache line) are valid input, especially for very wide cache line or stream processor.  We may choose to process it with SIMD or not, but it must work.

nccvector:
> Cool, so the reason of infinite run is actually segmentation fault? I'll wait for the description for more details.

Segfault + Profiler

gtests/CMakeLists.txt
yungyuc: I concur.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -115,6 +115,13 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
     T const * ret = NULL;
 
     T const * ptr = start;
+
+    // Check if array is large enough for SIMD processing
```
yungyuc, line 119: This prevents wrong counting when the number of elements is less than a SIMD vector width for `check_between()` (vector mode).

I am curious whether this issue is also in other vector-mode calculating functions?  @KHLee529 what do you think?

gtests/CMakeLists.txt
```diff
@@ -46,3 +46,19 @@ gtest_discover_tests(test_nopython)
 add_custom_target(run_gtest
     COMMAND $<TARGET_FILE:test_nopython>
     DEPENDS test_nopython)
+
+# New binary target for testing take_along_axis_simd directly from C++
```
yungyuc: Why do you need a new binary instead of using [`test_nopython` at L48](https://github.com/solvcon/modmesh/pull/641/files#diff-d7873e4ae849e77b02084c569b8d18ff4037158430e93f47869159c273a97b64R48) above?

yungyuc (changes requested):
This is a good fix.  Thank you, @nccvector.

One point to discuss:
- [x] Why create a new binary instead of using the existing `test_nopython`?

Review protocol reminder: please leave a global comment to request for a next review, and use the inline comments (annotations) to discuss each of the points.

tychuang1211:
https://github.com/nccvector/modmesh/blob/f485c5292821d784334031f0cce01402adce4442/tests/test_buffer.py#L1020
Would you like fix the typo here as well?

yungyuc:
> https://github.com/nccvector/modmesh/blob/f485c5292821d784334031f0cce01402adce4442/tests/test_buffer.py#L1020 Would you like fix the typo here as well?

Good catch, @tychuang1211 .  Feel free to fix the existing code in a separate PR.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -115,6 +115,13 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
...
+    if (end - start < N_lane)
+    {
+        return generic::check_between<T>(start, end, min_val, max_val);
+    }
+
     for (; ptr <= end - N_lane; ptr += N_lane)
```
KHLee529, line 125: @nccvector I'm wondering, why this for loop is not skipped if `end - start < N_lane`.

KHLee529: > I am curious whether this issue is also in other vector-mode calculating functions? @KHLee529 what do you think?

@yungyuc To answer your question in another comment, I found this point a bit strange. This loop should be skip if the amount of array elements between `ptr`, which is equal to `start`, and `end` is less than `N_lane`. I did this in every other vector computation. Therefore if `check_between` may cause segmentation fault, others may do so, too.

KHLee529:
> * Then I added the modmesh.call_profiler similar to the actual profiling script and viola it hanged.

@nccvector did you find the reason why the application wont just crash when call_profiler is used?

nccvector:
> > * Then I added the modmesh.call_profiler similar to the actual profiling script and viola it hanged.
> 
> @nccvector did you find the reason why the application wont just crash when call_profiler is used?

Sorry I have been away for a while, will check today or over the weekend

cpp/modmesh/simd/neon/neon.hpp
nccvector: @KHLee529 the loop gets skipped but below this loop there is a recursive call under `if(ptr != end)` condition which calls this function again with the same array. If we pass an array of size 1 (which is N_lane=2) we run into an infinite recursion. Now I am trying to figure out why this results in a segfault consistently (possible stack corruption?)

gtests/CMakeLists.txt
nccvector: Sorry I missed that. Making the change

cpp/modmesh/simd/neon/neon.hpp
nccvector: The use of recursion also seems unnecessary here, we can refactor a bit and avoid recursion altogether

gtests/CMakeLists.txt
nccvector: Done

cpp/modmesh/simd/neon/neon.hpp
yungyuc: Good catch!  My bad that I did not notice the recursion.  I concur.  Recursion should not be used here.  High-performance code should exclusively use loop so that it will be more straight-forward to check the assembly generated.

> The use of recursion also seems unnecessary here, we can refactor a bit and avoid recursion altogether

Agree.  Let's make it a separate task.  I can find time to create an issue to track it, but anyone is welcome to create one.

gtests/CMakeLists.txt
yungyuc: Cool

yungyuc (approved):
LGTM

@nccvector this is a ~~small~~ straight-forward change and is suitable to be squashed into a single commit.  Do you want to do it before I merge?

yungyuc:
> > > * Then I added the modmesh.call_profiler similar to the actual profiling script and viola it hanged.
> > 
> > 
> > @nccvector did you find the reason why the application wont just crash when call_profiler is used?
> 
> Sorry I have been away for a while, will check today or over the weekend

This is an interesting point.  If github is not the best place for brainstorming, we can also discuss on discord: https://discord.com/channels/730297880140578906/730393342818648204/1446289303864086601

cpp/modmesh/simd/neon/neon.hpp
tychuang1211: Why can't we just replace line 138 with `return generic::check_between<T>(start, end, min_val, max_val);`?

nccvector:
> LGTM
> 
> @nccvector this is a ~small~ straight-forward change and is suitable to be squashed into a single commit. Do you want to do it before I merge?

Squashed, now its one commit 👍

cpp/modmesh/simd/neon/neon.hpp
KHLee529: Oh my god, this is my fault. What I wanted to do was `generic::check_between` instead of `check_between`. The recursion is not intended. You can refer to the arithmetic operations below, they are correct.

In this case, I would suggest fix this issue by simply replace `check_between` by `generic::check_between` at L145. That would make more sense to me.

cpp/modmesh/simd/neon/neon.hpp
KHLee529: > Why can't we just replace line 138 with `return generic::check_between<T>(start, end, min_val, max_val);`?

This is exactly the way I suggest to fix this issue.

KHLee529 (changes requested):
- [ ] Fix this issue by change the `check_between` to `generic::check_between` in L138 (L145 in current context)

yungyuc:
> > LGTM
> > @nccvector this is a ~small~ straight-forward change and is suitable to be squashed into a single commit. Do you want to do it before I merge?
> 
> Squashed, now its one commit 👍

Wonderful.  Merged.

yungyuc:
> * [ ]  Fix this issue by change the `check_between` to `generic::check_between` in L138 (L145 in current context)

Good catch.  @KHLee529 , please create an issue to track the fix.

tigercosmos:
@nccvector  Good job!

KHLee529:
@nccvector nice job!

KHLee529:
> > * [ ]   Fix this issue by change the `check_between` to `generic::check_between` in L138 (L145 in current context)
> 
> Good catch. @KHLee529 , please create an issue to track the fix.

Issue #646 is created to track this fix.

PR #644, opened 2025-12-03
ci: send email to mailing list when CI fails on master branch

ExplorerRay:
This PR sends an email to the SOLVCON Google Group per workflow when CI fails on the `master` branch.

I only check the `master` branch to prevent unnecessary email noise in the Google Group from PR CI failures.

The CI I ran on my local fork to trigger the following mail sending:
https://github.com/ExplorerRay/modmesh/actions/runs/19896588063

The email sent out will be something looks like this:
image

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -641,3 +641,33 @@ jobs:
         with:
           name: modmesh-pilot-win64
           path: modmesh-pilot-win64/
+
+  send_email_on_failure:
+    name: send_emails_when_fails
+    needs: [standalone_buffer, build_ubuntu, build_macos, build_windows]
+    runs-on: ubuntu-latest
+    # Run if any of the dependencies failed in master branch
+    if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' }}
+    steps:
+      - name: Send mail to mailing list
+        uses: dawidd6/action-send-mail@v6
+        with:
+          server_address: smtp.gmail.com
+          server_port: 465
+          username: ${{ secrets.EMAIL_USERNAME }}
+          password: ${{ secrets.EMAIL_PASSWORD }}
+          # Subject: Uses the Repository name
+          subject: CI Failure in ${{ github.repository }} ${{ github.workflow }} workflow
+          # Body: Lists the result of every job
+          body: |
+            The workflow ${{ github.workflow }} has failed.
+            Job Status Report:
+            ------------------
+            - standalone_buffer: ${{ needs.standalone_buffer.result }}
+            - build_ubuntu: ${{ needs.build_ubuntu.result }}
+            - build_macos: ${{ needs.build_macos.result }}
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -138,7 +138,7 @@ jobs:
         #       libgl1-mesa-dev
         # fi
 
-    - name: Add 8G swap (Ubuntu) 
+    - name: Add 8G swap (Ubuntu)
```
yungyuc: Good catch!  But make the cosmetic change in a separate PR.  We should also check for the TWS in a script.

The TWS removal work is tracked in issue #645.

.github/workflows/devbuild.yml
```diff
@@ -642,4 +642,34 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: modmesh-pilot-win64
-          path: modmesh-pilot-win64/
\ No newline at end of file
+          path: modmesh-pilot-win64/
+
+  send_email_on_failure:
+    name: Send_emails_when_CI_fails
```
yungyuc: Use all-lower-case for the name.

yungyuc (changes requested):
- [x] Revert changes about TWS.
- [x] Use all-lower-case for the name: send_emails_when_fails.  ("ci" is redundant, but you may still add if you want.)

.github/workflows/devbuild.yml
ExplorerRay: Thanks for the remind, this is auto fixed by my text editor. I already exclude them for this PR.

yungyuc:
Please rebase before moving forward.

yungyuc:
@ExplorerRay do not resolve comments.  Resolving comments makes them hard to be tracked.

Do not use AI tools to add comments.  The way you use it adds more comments than non-trivial code diff and causes confusion.

The PR now looks unreadable.  I am closing it.  Please recreate one and start over.

yungyuc:
Tools (AI or not) must help productivity.  **Clarity** is a critical ingredient to productivity.  In this regard, it usually is a bad idea to directly copy what tools give you.  We must absorb and conclude it ourselves before showing it to collaborators.

.github/workflows/devbuild.yml
yungyuc: Do not make text editor fix formatting automatically.

ExplorerRay:
Sorry for that, I just tested some functions of copilot and just enable every feature without further checking it. I will be more careful when using tools.

By the way, I just recreated another PR [here](https://github.com/solvcon/modmesh/pull/647).

yungyuc:
> Sorry for that, I just tested some functions of copilot and just enable every feature without further checking it. I will be more careful when using tools.
> 
> By the way, I just recreated another PR [here](https://github.com/solvcon/modmesh/pull/647).

No problem.  I can see that it is an experiement.

There is a trick for experimenting new features on github: create PR or issue in your own fork.  In this way you do not mess up the discussions.

PR #647, opened 2025-12-06
ci: send email to mailing list when CI fails on master branch

ExplorerRay:
This PR sends an email to the SOLVCON Google Group per workflow when CI fails on the `master` branch.

I only check the `master` branch to prevent unnecessary email noise in the Google Group from PR CI failures.

The CI I ran on my local fork to trigger the following mail sending:
https://github.com/ExplorerRay/modmesh/actions/runs/19896588063

The email content looks like below:
image

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -641,3 +641,34 @@ jobs:
         with:
           name: modmesh-pilot-win64
           path: modmesh-pilot-win64/
+
+  send_email_on_failure:
+    name: send_emails_when_fails
+    needs: [standalone_buffer, build_ubuntu, build_macos, build_windows]
+    runs-on: ubuntu-latest
+    # Run if any of the dependencies failed in master branch
+    if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' }}
+    steps:
+      - name: Send mail to mailing list
+        uses: dawidd6/action-send-mail@v6
+        with:
+          server_address: smtp.gmail.com
+          server_port: 465
+          username: ${{ secrets.EMAIL_USERNAME }}
+          password: ${{ secrets.EMAIL_PASSWORD }}
+          # Subject: Uses the Repository name
+          subject: CI Failure in ${{ github.repository }} ${{ github.workflow }} workflow
+          # Body: Lists the result of every job
+          body: |
+            The workflow ${{ github.workflow }} has failed.
+            Job Status Report:
+            ------------------
+            - standalone_buffer: ${{ needs.standalone_buffer.result }}
+            - build_ubuntu: ${{ needs.build_ubuntu.result }}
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -641,3 +641,34 @@ jobs:
         with:
           name: modmesh-pilot-win64
           path: modmesh-pilot-win64/
+
+  send_email_on_failure:
```
tigercosmos, line 645: Is it possible to have some retry machenism while CI fails on master, so it will not be a spam? The idea is we want to receive the mail when the CI is really broken.

yungyuc: Good point.  Or send the email after one or two retrials.

.github/workflows/devbuild.yml
```diff
@@ -641,3 +641,34 @@ jobs:
...
+            The workflow ${{ github.workflow }} has failed.
+            Job Status Report:
+            ------------------
+            - standalone_buffer: ${{ needs.standalone_buffer.result }}
+            - build_ubuntu: ${{ needs.build_ubuntu.result }}
+            - build_macos: ${{ needs.build_macos.result }}
+            - build_windows: ${{ needs.build_windows.result }}
+
+            Check the details at ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          to: solvcon@googlegroups.com
```
yungyuc, line 672: Do we need to put the email address in a secret too?

yungyuc (changes requested):
- [x] Discussion: use secret for email address or not.

.github/workflows/devbuild.yml
ExplorerRay: I think the pros of putting it as secret is that avoiding other malicious people get this email with this github action yaml and join the google group, then send spam. 
The cons will be that only those who manage secret can directly change the email destination without public PR check. Those who want to get notified also don't know how to do.

About the pros, google group also has some anti-spam setting and I think google group admin can kick those who send spams which may make the pros more trivial. Therefore, I think it's better leave it as plaintext.

.github/workflows/devbuild.yml
ExplorerRay: Thanks for the idea. I will survey how to retry previous failed jobs in Github actions.

.github/workflows/devbuild.yml
yungyuc: OK.  We can keep it this way for a while.  Please survey how other projects treat it and bring it to the discord.

.github/workflows/devbuild.yml
yungyuc: Sounds good.  Please create an issue to track the research.  We keep the current implementation for now.

yungyuc (approved):
LGTM

PR #650, opened 2025-12-07
Refactor buffer code using `concept`

tigercosmos:
Use C++20 `concept` for the code in `buffer`.

Using Copilot + Claude 4.5.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -160,7 +160,7 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 SpaceBeforeSquareBrackets: false
 BitFieldColonSpacing: Both
-Standard:        c++14
+Standard: c++20
 StatementAttributeLikeMacros:
   - Q_EMIT
 StatementMacros:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,9 +29,9 @@ jobs:
           echo "github.event_name: ${{ github.event_name }}"
 
       - name: Run clang-format style check for C/C++/Protobuf programs.
-        uses: jidicula/clang-format-action@v4.11.0
+        uses: jidicula/clang-format-action@v4.16.0
         with:
-          clang-format-version: '16'
+          clang-format-version: '20'
           check-path: ${{ matrix.path }}
           fallback-style: 'LLVM' # optional
 
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -38,6 +38,7 @@
 #include <modmesh/toggle/RadixTree.hpp>
 
 #include <algorithm>
+#include <concepts>
 #include <functional>
 #include <limits>
 #include <numeric>
@@ -53,6 +54,28 @@ namespace modmesh
 
 template <typename T>
 class SimpleArray; // forward declaration
+
+template <typename T>
+concept IntegralType = std::is_integral_v<T>;
+
+template <typename T>
+concept ArithmeticType = std::is_arithmetic_v<T>;
+
+template <typename T>
+concept SimpleArrayType = requires(T t) {
+    { t.data() } -> std::convertible_to<typename T::value_type *>;
+    { t.size() } -> std::convertible_to<size_t>;
+};
+
+template <typename T, typename U>
+concept IsSameRemoveConstType = std::is_same_v<std::remove_const_t<T>, U>;
+
+template <typename It>
+concept InputIterator = requires(It it) {
+    { *it };
+    { ++it } -> std::same_as<It &>;
+    { it++ } -> std::same_as<It>;
+};
 
 namespace detail
 {
@@ -1046,9 +1069,9 @@ public:
 
     void sort(void);
     SimpleArray<uint64_t> argsort(void);
-    template <typename I>
+    template <IntegralType I>
     A take_along_axis(SimpleArray<I> const & indices);
-    template <typename I>
+    template <IntegralType I>
     A take_along_axis_simd(SimpleArray<I> const & indices);
 
 }; /* end class SimpleArrayMixinSort */
@@ -1067,10 +1090,10 @@ void SimpleArrayMixinSort<A, T>::sort(void)
     std::sort(athis->begin(), athis->end());
 }
 
-template <typename T, typename I>
+template <typename T, IntegralType I>
 void indexed_copy(T * dest, T const * data, I const * begin, I const * const end);
 
-template <typename T>
+template <IntegralType T>
 T const * check_index_range(SimpleArray<T> const & indices, size_t max_idx);
 
 template <typename A, typename T>
@@ -1165,7 +1188,7 @@ public:
     {
     }
 
-    template <class InputIt>
+    template <InputIterator InputIt>
     SimpleArray(InputIt first, InputIt last, size_t alignment = 0)
         : SimpleArray(last - first, alignment)
     {
@@ -1824,10 +1847,9 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSort<A, T>::argsort(void)
 }
 
 template <typename A, typename T>
-template <typename I>
+template <IntegralType I>
 A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & indices)
 {
-    static_assert(std::is_integral_v<I>, "I must be integral type");
     auto athis = static_cast<A *>(this);
     if (athis->ndim() != 1)
     {
@@ -1879,7 +1901,7 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
-template <typename T>
+template <IntegralType T>
 T const * detail::check_index_range(SimpleArray<T> const & indices, size_t max_idx)
 {
     constexpr T DataTypeMax = std::numeric_limits<T>::max();
@@ -1892,7 +1914,7 @@ T const * detail::check_index_range(SimpleArray<T> const & indices, size_t max_i
     return simd::check_between<T>(indices.begin(), indices.end(), 0, max_idx);
 }
 
-template <typename T, typename I>
+template <typename T, IntegralType I>
 void detail::indexed_copy(T * dest, T const * data, I const * index0, I const * const index1)
 {
     T * dst = dest;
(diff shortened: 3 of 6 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -57,6 +58,33 @@ class SimpleArray; // forward declaration
...
+concept ArithmeticType = std::is_arithmetic_v<T>;
+
+template <typename T>
+concept SimpleArrayType = requires(T t)
+{
+    typename T::value_type;
+    // clang-format off
+    {t.data()}->std::convertible_to<typename T::value_type *>;
+    {t.size()}->std::convertible_to<size_t>;
+    // clang-format on
```
tigercosmos: Not use clang-format to be compact.

tigercosmos:
@yungyuc  Please help take a look. Thanks!

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -43,9 +43,10 @@ namespace python
 /// @param arrayplex the plex array, which is the wrapper of the typed array
 /// @param callback the callback function, which has the mutable typed array as the argument
 /// @return the return type of the callback function
-template <typename A, typename C, typename = std::enable_if_t<std::is_same_v<std::remove_const_t<A>, SimpleArrayPlex>>>
-// NOLINTNEXTLINE(misc-use-anonymous-namespace,cppcoreguidelines-missing-std-forward)
-static auto execute_callback_with_typed_array(A & arrayplex, C && callback)
+template <typename A, typename C>
+requires modmesh::detail::SameAs<A, SimpleArrayPlex>
+    // NOLINTNEXTLINE(misc-use-anonymous-namespace,cppcoreguidelines-missing-std-forward)
```
yungyuc: Misleading indentation.  If it is enforced by clang-format, please try to fix the clang-format configuration.  Let me know if it is a limitation of the linter.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -57,6 +58,35 @@ class SimpleArray; // forward declaration
...
+};
+
+template <typename T, typename U>
+concept SameAs = std::is_same_v<std::remove_const_t<T>, U>;
```
yungyuc: I don't like the name `SameAs`.  It's too generic and similar to `std::is_same`.  Please use a name more specific to what it does.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -57,6 +58,35 @@ class SimpleArray; // forward declaration
...
+template <typename T>
+concept IntegralType = std::is_integral_v<T>;
+
+template <typename T>
+concept ArithmeticType = std::is_arithmetic_v<T>;
+
+template <typename T>
+concept SimpleArrayType = requires(T t)
+{
+    typename T::value_type;
```
yungyuc: Do you need this line?


```cpp
typename T::value_type;
```

Or did you intend to write:

```cpp
using value_type = typename T::value_type;
{t.data()}->std::convertible_to<value_type *>;
{t.size()}->std::convertible_to<size_t>;
```

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -28,10 +28,10 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <stdexcept>
+#include <algorithm>
```
yungyuc: Why reordering the includes?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -57,6 +58,35 @@ class SimpleArray; // forward declaration
 namespace detail
 {
 
+template <typename T>
+concept IntegralType = std::is_integral_v<T>;
```
yungyuc, line 59: Concepts should be made in `modmesh` namespace rather than `modmesh::detail`.  Thus we should be careful creating and naming them.

yungyuc (changes requested):
I like concepts.  Points to address:

- [x] Concepts should be made in `modmesh` namespace rather than `modmesh::detail`.
- [x] Find a better name than `SameAs`.
- [x] Clarify why reorder inclusions in `small_vector.hpp`.  If it is not necessary, revert the change.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: It's correct indentation. L49(` static auto execute_callback_with_typed_array`) has indent because it's under `requires`, and this comment is for L49, which should follow the same indent.

cpp/modmesh/buffer/small_vector.hpp
tigercosmos: My formattor did it and I thought okay to sort the headers. Reverted.

cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp
tigercosmos: It seems that we can update the C++ standard for clang-format, and this should be fixed.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Thanks for catching this. Should be a mistake.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Moved all concepts outside `detail`

.github/workflows/lint.yml
```diff
@@ -29,9 +29,9 @@ jobs:
           echo "github.event_name: ${{ github.event_name }}"
 
       - name: Run clang-format style check for C/C++/Protobuf programs.
-        uses: jidicula/clang-format-action@v4.11.0
+        uses: jidicula/clang-format-action@v4.16.0
         with:
-          clang-format-version: '16'
+          clang-format-version: '20'
```
tigercosmos, line 34: Use new clang-format to support C++23.

cpp/modmesh/buffer/small_vector.hpp
```diff
@@ -78,8 +78,8 @@ class small_vector
         std::copy_n(vector.begin(), m_size, begin());
     }
 
-    template <class InputIt>
-    small_vector(InputIt first, InputIt last)
+    template <std::input_iterator It>
```
tigercosmos, line 81: I found that we can just use std here.

cpp/modmesh/spacetime/pymod/wrap_spacetime.hpp
```diff
@@ -361,7 +361,7 @@ class WrapSolverBase
                 py::arg("odd_plane") = false)
             .def("get_so0p", &wrapped_type::get_so0p, py::arg("iv"), py::arg("odd_plane") = false);
 
-// clang-format off
+        // clang-format off
```
tigercosmos, line 364: `clang-format` catches this now. Let's just accept it.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Renamed.

.clang-format
```diff
@@ -160,7 +160,7 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 SpaceBeforeSquareBrackets: false
 BitFieldColonSpacing: Both
-Standard:        c++14
+Standard: c++20
```
tigercosmos, line 163: Latest is c++20, which supports c++23.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#standard

tigercosmos:
@yungyuc  all issues addressed. Please take a look.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -54,6 +55,28 @@ namespace modmesh
...
 class SimpleArray; // forward declaration
 
+template <typename T>
+concept IntegralType = std::is_integral_v<T>;
+
+template <typename T>
+concept ArithmeticType = std::is_arithmetic_v<T>;
+
+template <typename T>
+concept SimpleArrayType = requires(T t) {
```
yungyuc, line 65: Does clang-format force the opening curly brace to be in the same line?  It does not look nice consistent with other opening braces.  Please help clarify if it can be made consistent.

cpp/modmesh/buffer/small_vector.hpp
yungyuc: Good catch.

cpp/modmesh/spacetime/pymod/wrap_spacetime.hpp
yungyuc: It's nice.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -54,6 +55,28 @@ namespace modmesh
...
+    { t.data() } -> std::convertible_to<typename T::value_type *>;
+    { t.size() } -> std::convertible_to<size_t>;
+};
+
+template <typename T, typename U>
+concept SameAsWithoutConst = std::is_same_v<std::remove_const_t<T>, U>;
```
yungyuc: `SameAsWithoutConst` only removes the ambiguity of the name `SameAs`, but is not really good.

It's OK for now.  But what about `IsSameRemoveConstType`?  To match `is_same_v` and `remove_const_t`.

yungyuc (changes requested):
One point to address:

- [x] Clarify how to make opening brace for `requires` use its own line.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: Sounds better. Let me change.

cpp/modmesh/buffer/SimpleArray.hpp
tigercosmos: It's impossible for now unless turnning off clang-format.
https://github.com/llvm/llvm-project/issues/86778

tigercosmos:
@yungyuc  Updated. Please take a look.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: OK.  Some day we may try to make it happen in clang-format.  Or they can make it happen before us.

yungyuc (approved):
LGTM

PR #657, opened 2025-12-19
ci: add date to email subject when CI fails

### LINKED ISSUE #656: Add date to email subject when CI fails
The current email subject only includes the workflow name, which causes Google Groups to aggregate CI failure emails from different dates into a single thread.
To improve clarity, we would like to include the date (month and day) in the email subject.

ExplorerRay:
- Prepend date (format is like `12/19`) to email subject when CI fails
- Solves https://github.com/solvcon/modmesh/issues/656

The following test shows that setting command output to env is available.
https://github.com/ExplorerRay/modmesh/actions/runs/20363563177/job/58513478643


yungyuc (approved):
Sweet

PR #660, opened 2026-01-19
Implement batch API for Kalman filter

ThreeMonth03:
As metioned in #601 , if there are many time series data-points, we currently get states information by iterating functions
`predict()` and `update()`.  In other words, the program cannot compute a batch of data with calling them zero times or once. To use Kalman Filter handy, I implement new member function `KalmanFilter<T>::batch_filter(...)`, which can iteratively predict and update a batch of data points. 
For example, in `tests/test_linalg.py`, if we pass the arguments `observations` and `contorls`, the member function `kf.batch_filter(zs_sa, us_sa)` would return an object whose type is `KalmanStateInfo`, and the struct includes the prediction of `prior_states`, `prior_states_covariance`,  `posterior_states`, and `posterior_states_covariance` .

```python3
def test_batchfilter_with_control(self):
        # ...
        kf = mm.KalmanFilterFp64(
            x=x_sa, f=f_sa, b=b_sa, h=h_sa,
            process_noise=sigma_w,
            measurement_noise=1.0,
        )
        bps = kf.batch_filter(zs_sa, us_sa)
        xs_pred = bps.prior_states
        ps_pred = bps.prior_states_covariance
        xs_upd = bps.posterior_states
        ps_upd = bps.posterior_states_covariance
```

### DIFF
--- a/cpp/modmesh/linalg/kalman_filter.hpp
+++ b/cpp/modmesh/linalg/kalman_filter.hpp
@@ -41,6 +41,42 @@ struct select_real_t;
 } /* end namespace detail */
 
 /**
+ * @brief `KalmanStateInfo` includes prior and posterior states and their covariances,
+ *  and it is the return type of `KalmanFilter<T>::batch_filter(...)`.
+ *
+ * @details
+ * Due to the iteration of the predict and update steps in `KalmanFilter<T>::batch_filter(...)`, the
+ * intermediate results are stored for each time step, including
+ *  -   prior_states
+ *  -   prior_states_covariance
+ *  -   posterior_states
+ *  -   posterior_states_covariance
+ *
+ * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs)
+ * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs, array_type const & us)
+ */
+template <typename T>
+struct KalmanStateInfo
+{
+    using tuple_type = std::tuple<SimpleArray<T>, SimpleArray<T>, SimpleArray<T>, SimpleArray<T>>;
+
+    KalmanStateInfo(size_t observation_size, size_t state_size)
+    {
+        small_vector<size_t> xs_shape{observation_size, state_size};
+        small_vector<size_t> ps_shape{observation_size, state_size, state_size};
+        prior_states = SimpleArray<T>(xs_shape);
+        prior_states_covariance = SimpleArray<T>(ps_shape);
+        posterior_states = SimpleArray<T>(xs_shape);
+        posterior_states_covariance = SimpleArray<T>(ps_shape);
+    }
+
+    SimpleArray<T> prior_states;
+    SimpleArray<T> prior_states_covariance;
+    SimpleArray<T> posterior_states;
+    SimpleArray<T> posterior_states_covariance;
+}; /* end struct KalmanStateInfo */
+
+/**
  * Reference: FilterPy KalmanFilter documentation
  * https://filterpy.readthedocs.io/en/latest/kalman/KalmanFilter.html
  */
@@ -205,6 +241,41 @@ public:
         update_covariance(k);
     }
 
+    /**
+     * @brief Predict and update in batch mode without a batch of control input `us`.
+     *
+     * @details
+     * Given a lot of observation inputs `zs` in different timestamps,
+     * `batch_filter(array_type const & zs)` predicts and updates iteratively,
+     * and return prior and posterior states and covariancein in different timestamps.
+     * This api design is refered from [filterpy]
+     * (https://filterpy.readthedocs.io/en/latest/_modules/filterpy/kalman/kalman_filter.html#KalmanFilter.batch_filter).
+     *
+     * @param zs A batch of measurement inputs.
+     *
+     * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs, array_type const & us)
+     * @see struct KalmanStateInfo<T>;
+     */
+    KalmanStateInfo<T> batch_filter(array_type const & zs);
+
+    /**
+     * @brief Predict and update in batch mode with a batch of control input `us`.
+     *
+     * @details
+     * Given a lot of observation inputs `zs` and control inputs `us` in different timestamps,
+     * `batch_filter(array_type const & zs)` predicts and updates iteratively,
+     * and return prior and posterior states and covariancein in different timestamps.
+     * This api design is refered from [filterpy]
+     * (https://filterpy.readthedocs.io/en/latest/_modules/filterpy/kalman/kalman_filter.html#KalmanFilter.batch_filter).
+     *
+     * @param zs A batch of measurement inputs.
+     * @param us A batch of control inputs.
+     *
+     * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs)
+     * @see struct KalmanStateInfo<T>;
+     */
+    KalmanStateInfo<T> batch_filter(array_type const & zs, array_type const & us);
+
 private:
 
     void check_dimensions();
@@ -222,6 +293,10 @@ private:
     array_type kalman_gain(array_type const & s);
     void update_state(array_type const & k, array_type const & y);
     void update_covariance(array_type const & k);
+
+    // Batch filter
+    void predict_and_update(array_type const & z, KalmanStateInfo<T> & bfs, size_t iter);
+    void predict_and_update(array_type const & z, array_type const & u, KalmanStateInfo<T> & bfs, size_t iter);
 
 }; /* end class KalmanFilter */
 
@@ -480,6 +555,117 @@ void KalmanFilter<T>::update_covariance(array_type const & k)
 
     // P <- 0.5(P + P^H)  (m_p <- 0.5(m_p + m_p^H))
     m_p = m_p.symmetrize();
+}
+
+template <typename T>
+KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs)
+{
+    size_t z_m = zs.shape(0);
+    size_t z_n = zs.shape(1);
+    array_type z(small_vector<size_t>{z_n});
+    KalmanStateInfo<T> bfs(z_m, m_state_size);
+
+    for (size_t iter = 0; iter < z_m; ++iter)
+    {
+        for (size_t j = 0; j < z_n; ++j)
+        {
+            z(j) = zs(iter, j);
+        }
+        predict_and_update(z, bfs, iter);
+    }
+    return bfs;
+}
+
+template <typename T>
+void KalmanFilter<T>::predict_and_update(array_type const & z, KalmanStateInfo<T> & bfs, size_t iter)
+{
+    SimpleArray<T> & prior_state = bfs.prior_states;
+    SimpleArray<T> & prior_state_covariance = bfs.prior_states_covariance;
+    SimpleArray<T> & posterior_state = bfs.posterior_states;
+    SimpleArray<T> & posterior_state_covariance = bfs.posterior_states_covariance;
+
+    predict();
+    for (size_t j = 0; j < m_state_size; ++j)
+    {
+        prior_state(iter, j) = m_x(j);
+        for (size_t k = 0; k < m_state_size; ++k)
+        {
+            prior_state_covariance(iter, j, k) = m_p(j, k);
+        }
+    }
+
+    update(z);
+    for (size_t j = 0; j < m_state_size; ++j)
+    {
+        posterior_state(iter, j) = m_x(j);
+        for (size_t k = 0; k < m_state_size; ++k)
+        {
+            posterior_state_covariance(iter, j, k) = m_p(j, k);
+        }
+    }
+}
+
+template <typename T>
+KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs, array_type const & us)
+{
+    size_t z_m = zs.shape(0);
+    size_t z_n = zs.shape(1);
+    array_type z(small_vector<size_t>{z_n});
+    KalmanStateInfo<T> bfs(z_m, m_state_size);
+
+    array_type u;
+    size_t u_n = 0;
+
+    size_t u_m = us.shape(0);
+    if (u_m != z_m)
+    {
+        throw std::invalid_argument("KalmanFilter::batch_filter: The number of control inputs must match the number of measurements.");
+    }
+    u_n = us.shape(1);
+    u = array_type(small_vector<size_t>{u_n});
+
+    for (size_t iter = 0; iter < z_m; ++iter)
+    {
+        for (size_t j = 0; j < z_n; ++j)
+        {
+            z(j) = zs(iter, j);
+        }
+        for (size_t j = 0; j < u_n; ++j)
+        {
+            u(j) = us(iter, j);
+        }
+        predict_and_update(z, u, bfs, iter);
+    }
+    return bfs;
+}
+
+template <typename T>
+void KalmanFilter<T>::predict_and_update(array_type const & z, array_type const & u, KalmanStateInfo<T> & bfs, size_t iter)
+{
+    SimpleArray<T> & prior_state = bfs.prior_states;
+    SimpleArray<T> & prior_state_covariance = bfs.prior_states_covariance;
+    SimpleArray<T> & posterior_state = bfs.posterior_states;
+    SimpleArray<T> & posterior_state_covariance = bfs.posterior_states_covariance;
+
+    predict(u);
+    for (size_t j = 0; j < m_state_size; ++j)
+    {
+        prior_state(iter, j) = m_x(j);
+        for (size_t k = 0; k < m_state_size; ++k)
+        {
+            prior_state_covariance(iter, j, k) = m_p(j, k);
+        }
+    }
+
+    update(z);
+    for (size_t j = 0; j < m_state_size; ++j)
+    {
+        posterior_state(iter, j) = m_x(j);
+        for (size_t k = 0; k < m_state_size; ++k)
+        {
+            posterior_state_covariance(iter, j, k) = m_p(j, k);
+        }
+    }
 }
 
 } /* end namespace modmesh */
--- a/cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
+++ b/cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
@@ -37,6 +37,42 @@ namespace python
 {
 
 template <typename T>
+class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapKalmanStateInfo
+    : public WrapBase<WrapKalmanStateInfo<T>, KalmanStateInfo<T>>
+{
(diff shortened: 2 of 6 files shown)


cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -493,6 +498,80 @@ void KalmanFilter<T>::update_covariance(array_type const & k)
...
+                u(j) = us(iter, j);
+            }
+        }
+        else if (iter == u_m)
+        {
+            u = array_type(small_vector<size_t>{0});
+        }
+
+        predict_and_update(z, u, bfs, iter);
+    }
```
ThreeMonth03, line 668: Iterate the data points according to the first dimension.

ThreeMonth03 (commented):

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -53,6 +53,7 @@ class KalmanFilter
     using value_type = T;
     using real_type = typename detail::select_real_t<value_type>::type;
     using array_type = SimpleArray<T>;
+    using bf_type = std::tuple<array_type, array_type, array_type, array_type>;
```
yungyuc: This `bf_type` needs a helper class. `std::tuple` should not be used because it is too difficult to document for each of the field.

The alias should probably be avoided once the helper class is made and named carefully.

yungyuc (changes requested):
We can do this for now.  But to really test the implementation, we need an application for the Kalman filter.

- [ ] `KalmanFilter::bf_type` needs a helper class.

We are about to have a hammer.  Let's find a nail.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -493,6 +498,80 @@ void KalmanFilter<T>::update_covariance(array_type const & k)
...
+{
+    auto & [prior_xs, prior_ps, posterior_xs, posterior_ps] = bfs;
+    if (u.shape(0) > 0)
+    {
+        predict(u);
+    }
+    else
+    {
+        predict();
+    }
```
j8xixo12: Is possible to separate with / without control input into 2 functions to avoid this branch ?

ThreeMonth03: Yes, I thought that the row of control input may be missing value, but it can be deal with at a different module.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,49 @@ template <typename T>
...
+    {
+        return std::make_tuple(m_prior_xs, m_prior_ps, m_posterior_xs, m_posterior_ps);
+    }
+
+private:
+    SimpleArray<T> m_prior_xs;
+    SimpleArray<T> m_prior_ps;
+    SimpleArray<T> m_posterior_xs;
+    SimpleArray<T> m_posterior_ps;
+}; /* end class BatchFilterHelper */
```
ThreeMonth03: Create the batch filter helper.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -493,6 +547,74 @@ void KalmanFilter<T>::update_covariance(array_type const & k)
...
+        {
+            for (size_t j = 0; j < u_n; ++j)
+            {
+                (*u_ptr)(j) = (*us_ptr)(iter, j);
+            }
+        }
+        predict_and_update(z, u_ptr, bfs, iter);
+    }
+    return bfs;
+}
```
ThreeMonth03, line 640: I'm not sure whether it is a good idea to deal with the similar pattern by `std::unique_ptr` or `std::optional`.

cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
```diff
@@ -93,7 +93,22 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapKalmanFilter
...
+                },
+                py::arg("zs"),
+                py::arg("us"))
+            .def(
+                "batch_filter",
+                [](wrapped_type & self, array_type const & zs)
+                {
+                    return self.batch_filter(zs).to_tuple();
+                },
+                py::arg("zs"));
```
ThreeMonth03, line 147: To pass a few return variables, the helper class should be casted to std::tuple.

ThreeMonth03 (commented):
> We can do this for now. But to really test the implementation, we need an application for the Kalman filter.
> 
> * [x]  `KalmanFilter::bf_type` needs a helper class.
> 
> We are about to have a hammer. Let's find a nail.


@yungyuc. I fix it. Please review this pull request. Thanks.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,49 @@ template <typename T>
 struct select_real_t;
 } /* end namespace detail */
 
+template <typename T>
+class BatchFilterHelper
```
yungyuc: Provide a description block of comment for the class.  I don't know what does "filter" mean and please explain that at least.

Rename it to be meaningful (and as concise as `BatchFilterHelper`).

`BatchFilterHelper` does not carry much information.  It does not help readers.

Also consider to make it a `struct` with all member data public and remove the accessors.  All the member data now are arrays.  In this scenario, accessors are not very useful.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,49 @@ template <typename T>
...
+    {
+        return m_posterior_ps;
+    }
+    tuple_type to_tuple() const
+    {
+        return std::make_tuple(m_prior_xs, m_prior_ps, m_posterior_xs, m_posterior_ps);
+    }
+
+private:
+    SimpleArray<T> m_prior_xs;
```
yungyuc: What do the member data array mean?  It's hard to tell from the variable names.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -162,16 +198,7 @@ class KalmanFilter
...
-
-        // x <- F x + B u  (m_x <- m_f @ m_x + m_b @ u)
-        predict_state(u);
-
-        // P <- F P F^H + Q  (m_p <- m_f @ m_p @ m_f^H + m_q)
-        predict_covariance();
-
-        // P <- 0.5(P + P^H)  (m_p <- 0.5(m_p + m_p^H))
-        m_p = m_p.symmetrize();
+        predict(std::make_unique<array_type>(u));
```
yungyuc: `make_unique` constructs a new array and copy the data.  It will be slow.  Are you sure you want to force data copy for calling `predict`?

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: I don't see the necessary why you need to use `unique_ptr`.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -493,6 +547,74 @@ void KalmanFilter<T>::update_covariance(array_type const & k)
     m_p = m_p.symmetrize();
 }
 
+template <typename T>
+BatchFilterHelper<T> KalmanFilter<T>::batch_filter(array_type const & zs, std::unique_ptr<array_type> const & us_ptr)
```
yungyuc: It's a bad idea to be needing to include `_ptr` in the variable name to signify it's a pointer.

One of the argument is `array_type` and the other is `unique_ptr<array_type>`.  I do not see why you need different argument types for the two arrays.  It smells bad.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -205,15 +232,40 @@ class KalmanFilter
...
+    }
+
+    BatchFilterHelper<T> batch_filter(array_type const & zs, array_type const & us)
+    {
+        return batch_filter(zs, std::make_unique<array_type>(us));
+    }
+
 private:
 
+    void predict(std::unique_ptr<array_type> const & u_ptr)
```
yungyuc: I do not see why you need to use `unique_ptr` here.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -205,15 +232,40 @@ class KalmanFilter
...
+
     void check_dimensions();
     void check_measurement(array_type const & z);
     void check_control(array_type const & u);
 
     // Predict
-    void predict_state();
-    void predict_state(array_type const & u);
+    // void predict_state();
+    void predict_state(std::unique_ptr<array_type> const & u_ptr);
```
yungyuc: This is the reason you change many places to use `unique_ptr` and it does not make sense to me.

yungyuc (changes requested):
The use of `unique_ptr` is not necessary and should be removed from the PR.

More points to address:

- [x] Add a comment block to describe `BatchFilterHelper` and give the class a better name.
- [x] Evaluate if `BatchFilterHelper` should have accessors removed and changed to be `struct`.
- [x] ALERT: Are you sure you want to force data copy for calling `predict`?

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: > I don't see the necessary why you need to use `unique_ptr`.

The argument of control input `u` is optional, I'm not sure which design is the best choice:
    1. Overloading: It the contol input is 0 dimensional SimpleArray, it should be seen as the null input.
    2. std::optional
    3. std:unique_ptr
    4. Coding the similar logic twice: It looks quite long.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,41 @@ template <typename T>
...
+    tuple_type to_tuple() const
+    {
+        return std::make_tuple(prior_state, prior_state_covariance, posterior_state, posterior_state_covariance);
+    }
+
+    SimpleArray<T> prior_state;
+    SimpleArray<T> prior_state_covariance;
+    SimpleArray<T> posterior_state;
+    SimpleArray<T> posterior_state_covariance;
+}; /* end struct BFType */
```
ThreeMonth03: Change to POD design, and name it as `BFtype`.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -205,6 +240,31 @@ class KalmanFilter
...
+    /**
+     * @brief Predict and update in batch mode without a batch of control input us.
+     *
+     * @ref https://filterpy.readthedocs.io/en/latest/_modules/filterpy/kalman/kalman_filter.html#KalmanFilter.batch_filter
+     *
+     * @param zs A batch of measurement inputs.
+     *
+     * @see BFType<T> KalmanFilter<T>::batch_filter(array_type const & zs, array_type const & us)
+     * @see struct BFType<T>;
+     */
```
ThreeMonth03, line 258: The api design is refered from filter.py, and it is the origin of the api naming.

ThreeMonth03 (commented):
> * [x]  Add a comment block to describe `BatchFilterHelper` and give the class a better name.
> * [x]  Evaluate if `BatchFilterHelper` should have accessors removed and changed to be `struct`.
> * [x]  The use of `unique_ptr` is not necessary and should be removed from the PR.

@yungyuc Please review this pull request again. Thanks.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,41 @@ template <typename T>
 struct select_real_t;
 } /* end namespace detail */
 
+/**
+ * @brief Return type for `KalmanFilter<T>::batch_filter(...)`.
```
yungyuc: It is not informative:

> Return type for `KalmanFilter<T>::batch_filter(...)`

Define the type clearly in the comment block.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,41 @@ template <typename T>
...
+    BFType(size_t observation_size, size_t state_size)
+    {
+        small_vector<size_t> xs_shape{observation_size, state_size};
+        small_vector<size_t> ps_shape{observation_size, state_size, state_size};
+        prior_state = SimpleArray<T>(xs_shape);
+        prior_state_covariance = SimpleArray<T>(ps_shape);
+        posterior_state = SimpleArray<T>(xs_shape);
+        posterior_state_covariance = SimpleArray<T>(ps_shape);
+    }
+    tuple_type to_tuple() const
```
yungyuc: This `BFType::to_tuple` is only for Python wrapper.  It should not be put in the C++ core.  The code should be moved to the Python wrapper (pybind11).

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: Clarify what is filter.py.

cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
yungyuc: Tuple is hard to maintain.  Python invented named tuple to mitigate.

But we are using C++ already.  Why don't you just wrap the array container (`BFType`) to Python and return it?  It will work just like a named tuple in Python.

Doing it will need to turn `BFType` to a class with accessors, though.

yungyuc (changes requested):
> @yungyuc Please review this pull request again. Thanks.

I did not see your response to all my comments.  In the future, when changing the code, please also leave comments in the review to make it smoother.

Points to address:
- [x] Clearly define `BFType` in the header comment block.  Currently it only says it's the return type and it is not informative enough.
- [x] Move the logic in `BFType::to_tuple` to the Python wrapper (pybind11).
- [x] Clarify what is "filter.py".
- [x] Evaluate if it reasonable to just wrap the array container (`BFType`) to Python and return it from C++ to Python.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: It's not filter.py, that is FilterPy, [a kalman filter library](https://filterpy.readthedocs.io/en/latest/) implemented by python.

yungyuc: Make it clear in the code comment block.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,42 @@ template <typename T>
...
+        prior_states_covariance = SimpleArray<T>(ps_shape);
+        posterior_states = SimpleArray<T>(xs_shape);
+        posterior_states_covariance = SimpleArray<T>(ps_shape);
+    }
+
+    SimpleArray<T> prior_states;
+    SimpleArray<T> prior_states_covariance;
+    SimpleArray<T> posterior_states;
+    SimpleArray<T> posterior_states_covariance;
+}; /* end struct StatesInfo */
```
ThreeMonth03: Rename `BFType` to `StatesInfo`, and write a detail comment above.

cpp/modmesh/linalg/kalman_filter.hpp
ThreeMonth03: Done.

cpp/modmesh/linalg/pymod/wrap_kalman_filter.cpp
```diff
@@ -36,6 +36,42 @@ namespace modmesh
...
+            .def_readwrite(
+                "prior_states_covariance", &wrapped_type::prior_states_covariance)
+            .def_readwrite(
+                "posterior_states", &wrapped_type::posterior_states)
+            .def_readwrite(
+                "posterior_states_covariance", &wrapped_type::posterior_states_covariance);
+    }
+
+}; /* end class StatesInfo */
+
```
ThreeMonth03, line 74: Wrap `StatesInfo` to python.

ThreeMonth03 (commented):
> > @yungyuc Please review this pull request again. Thanks.
> 
> I did not see your response to all my comments. In the future, when changing the code, please also leave comments in the review to make it smoother.
> 
> Points to address:
> 
> * [x]  Clearly define `BFType` in the header comment block.  Currently it only says it's the return type and it is not informative enough.
> * [x]  Move the logic in `BFType::to_tuple` to the Python wrapper (pybind11).
> * [x]  Clarify what is "filter.py".
> * [x]  Evaluate if it reasonable to just wrap the array container (`BFType`) to Python and return it from C++ to Python.
@yungyuc You could review this pull request when you are available. Thanks.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,42 @@ template <typename T>
 struct select_real_t;
 } /* end namespace detail */
 
+/**
+ * @brief `StatsInfo` includes prior and posterior states and their covariances,
```
yungyuc: Fix the typo `StatsInfo` here in the brief (missing "e").

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: This is clear.  Rename to `KalmanStateInfo`.  `StateInfo` is too generic.  Prefixing `Kalman` works like a namespace (but we do not want to use a real namespace).

cpp/modmesh/linalg/kalman_filter.hpp
yungyuc: It's clear.

yungyuc (changes requested):
One more point to address:

- [x] Rename `StatesInfo` to `KalmanStateInfo`.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,42 @@ template <typename T>
...
+ *  -   prior_states_covariance
+ *  -   posterior_states
+ *  -   posterior_states_covariance
+ *
+ * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs)
+ * @see KalmanStateInfo<T> KalmanFilter<T>::batch_filter(array_type const & zs, array_type const & us)
+ */
+template <typename T>
+struct KalmanStateInfo
+{
```
ThreeMonth03, line 60: Rename `StatesInfo` to `KalmanStateInfo`.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -40,6 +40,42 @@ template <typename T>
 struct select_real_t;
 } /* end namespace detail */
 
+/**
+ * @brief `KalmanStateInfo` includes prior and posterior states and their covariances,
```
ThreeMonth03, line 44: Fix typo.

ThreeMonth03 (commented):
> One more point to address:
> 
> * [x]  Rename `StatesInfo` to `KalmanStateInfo`.


yungyuc (changes requested):
The code looks good.  Please update the commit log and the PR leading comment with a summary of what you did.

ThreeMonth03:
> The code looks good. Please update the commit log and the PR leading comment with a summary of what you did.

Hi @yungyuc . I've modified the commit message and the leading comment of this pull request. Please review it.

yungyuc (approved):
LGTM

PR #661, opened 2026-01-19
Feat: Modify __matmul__ to support SimpleArray of 1 dimension and 2 dimension

ThreeMonth03:
## Problem
There are some `__matmul__` operations in Kalman Filter. However, the current design only supports the arguments of 2 dimensions SimpleArray. If we want to matrix multiply from 2 dimensions SimpleArray to 1 dimension SimpleArray, we should reshape the SimpleArray twice. The detail of the discussion could be found in #601 .

```python3
void KalmanFilter<T>::predict_state()
{
    // x <- F x  (m_x <- m_f @ m_x)
    array_type x_col = m_x.reshape(small_vector<size_t>{m_state_size, 1});
    x_col = m_f.matmul(x_col);
    m_x = x_col.reshape(small_vector<size_t>{m_state_size});
    m_x = m_f.matmul(m_x);
}
```

## Solution

In this pull request, I modify `__matmul__` operation, so that it can support the following number of dimension of SimpleArray:
 * 1D x 1D
 * 1D x 2D
 * 2D x 1D
 * 2D x 2D

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -937,8 +937,8 @@ detail::SimpleArrayMixinCalculators<A, T>::median_freq(small_vector<value_type> 
 }
 
 /**
- * Perform matrix multiplication for 2D arrays.
- * This implementation supports only 2D x 2D matrix multiplication.
+ * Perform matrix multiplication for SimpleArrays.
+ * This implementation supports 1D x 1D, 1D x 2D, 2D x 1D, and 2D x 2D matrix multiplication.
  */
 template <typename A, typename T>
 A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
@@ -969,43 +969,112 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
         }
     };
 
-    if (this_ndim != 2 || other_ndim != 2)
+    auto check_product_shape = [&](A const * athis, A const * other, ssize_t athis_idx, ssize_t other_idx) -> void
+    {
+        if (athis->shape(athis_idx) != other->shape(other_idx))
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray::matmul(): shape mismatch: this={} other={}",
+                            format_shape(athis),
+                            format_shape(other)));
+        }
+    };
+
+    if ((this_ndim != 2 && this_ndim != 1) || (other_ndim != 2 && other_ndim != 1))
     {
         const std::string err = std::format("SimpleArray::matmul(): unsupported dimensions: "
-                                            "this={} other={}. Only 2D x 2D matrix multiplication is supported",
+                                            "this={} other={}. SimpleArray must be 1D or 2D.",
                                             format_shape(athis),
                                             format_shape(&other));
         throw std::out_of_range(err);
     }
 
-    const size_t m = athis->shape(0);
-    const size_t k = athis->shape(1);
-    const size_t n = other.shape(1);
+    bool this_is_1d = (this_ndim == 1);
+    bool other_is_1d = (other_ndim == 1);
 
-    if (k != other.shape(0))
+    // 1D x 1D
+    if (this_is_1d && other_is_1d)
     {
-        throw std::out_of_range(
-            std::format("SimpleArray::matmul(): shape mismatch: this={} other={}",
-                        format_shape(athis),
-                        format_shape(&other)));
+        check_product_shape(athis, &other, 0, 0);
+        A result(1);
+        value_type v = 0;
+        for (size_t i = 0; i < athis->shape(0); ++i)
+        {
+            v += (*athis)(i)*other.data(i);
+        }
+        result.data(0) = v;
+        return result;
     }
-
-    typename detail::SimpleArrayInternalTypes<T>::shape_type result_shape{m, n};
-    A result(result_shape);
-    result.fill(static_cast<value_type>(0));
-
-    for (size_t i = 0; i < m; ++i)
+    // 1D x 2D
+    else if (this_is_1d)
     {
+        const size_t k = athis->shape(0);
+        const size_t n = other.shape(1);
+        check_product_shape(athis, &other, 0, 0);
+        A result(n);
+
         for (size_t j = 0; j < n; ++j)
         {
+            value_type v = 0;
             for (size_t l = 0; l < k; ++l)
             {
-                result(i, j) += athis->operator()(i, l) * other(l, j);
+                v += (*athis)(l)*other(l, j);
+            }
+            result.data(j) = v;
+        }
+        return result;
+    }
+    // 2D x 1D
+    else if (other_is_1d)
+    {
+        const size_t m = athis->shape(0);
+        const size_t k = athis->shape(1);
+
+        check_product_shape(athis, &other, 1, 0);
+        A result(m);
+
+        for (size_t i = 0; i < m; ++i)
+        {
+            value_type v = 0;
+            for (size_t l = 0; l < k; ++l)
+            {
+                v += (*athis)(i, l) * other(l);
+            }
+            result.data(i) = v;
+        }
+        return result;
+    }
+    // 2D x 2D
+    else
+    {
+        const size_t m = athis->shape(0);
+        const size_t k = athis->shape(1);
+        const size_t n = other.shape(1);
+        check_product_shape(athis, &other, 1, 0);
+
+        shape_type result_shape{m, n};
+        A result(result_shape);
+
+        for (size_t i = 0; i < m; ++i)
+        {
+            for (size_t j = 0; j < n; ++j)
+            {
+                value_type v = 0;
+                for (size_t l = 0; l < k; ++l)
+                {
+                    v += (*athis)(i, l) * other(l, j);
+                }
+                result(i, j) = v;
             }
         }
+        return result;
     }
 
-    return result;
+    throw std::out_of_range(
+        std::format("SimpleArray::matmul(): this={} other={}"
+                    " cannot perform matrix multiplication.",
+                    format_shape(athis),
+                    format_shape(&other)));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,103 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
     }
 
-    const size_t m = athis->shape(0);
-    const size_t k = athis->shape(1);
-    const size_t n = other.shape(1);
+    bool this_is_1d = (this_ndim == 1);
+    bool other_is_1d = (other_ndim == 1);
 
-    if (k != other.shape(0))
+    if (this_is_1d && other_is_1d)
```
ThreeMonth03, line 996: The condition looks verbose. However, the input argument should be constant value, and if we try to reshape the input argument to another SimpleArry, the assignment operator would clone the buffer.

cpp/modmesh/linalg/kalman_filter.hpp
```diff
@@ -367,19 +367,14 @@ template <typename T>
...
 
 template <typename T>
 void KalmanFilter<T>::predict_state(array_type const & u)
 {
     // x <- F x + B u  (m_x <- m_f @ m_x + m_b @ u)
-    array_type x_col = m_x.reshape(small_vector<size_t>{m_state_size, 1});
-    array_type u_col = u.reshape(small_vector<size_t>{m_control_size, 1});
-    x_col = m_f.matmul(x_col).add(m_b.matmul(u_col));
-    m_x = x_col.reshape(small_vector<size_t>{m_state_size});
+    m_x = m_f.matmul(m_x).add(m_b.matmul(u));
```
ThreeMonth03, line 377: Simplify the logic.

ThreeMonth03 (commented):

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,103 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+        }
+        return result;
     }
-
-    typename detail::SimpleArrayInternalTypes<T>::shape_type result_shape{m, n};
-    A result(result_shape);
-    result.fill(static_cast<value_type>(0));
-
-    for (size_t i = 0; i < m; ++i)
+    else if (this_is_1d)
```
tigercosmos, line 1009: I will add a comment for "1D x 2D"

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,103 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
         {
             for (size_t l = 0; l < k; ++l)
             {
-                result(i, j) += athis->operator()(i, l) * other(l, j);
+                result.data(j) += (*athis)(l)*other(l, j);
+            }
+        }
+        return result;
+    }
+    else if (other_is_1d)
```
tigercosmos, line 1028: I would leave a comment "2D x 1D".

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,103 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
-
-    for (size_t i = 0; i < m; ++i)
+    else if (this_is_1d)
     {
+        const size_t k = athis->shape(0);
+        const size_t n = other.shape(1);
+        check_product_shape(athis, &other, 0, 0);
+        shape_type result_shape{n};
+        A result(result_shape);
+        result.fill(static_cast<value_type>(0));
```
tigercosmos: No need? Any element will not be modified later?

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,103 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+                {
+                    result(i, j) += (*athis)(i, l) * other(l, j);
+                }
             }
         }
+        return result;
     }
 
-    return result;
+    return {};
```
tigercosmos: I will let it throw here as it should never reach here.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fixed.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fixed.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: `SimpleArray` is constructed by `std::malloc`, which doesn't initialize the buffer. Furthermore, the elements are multiplied and added in the following for loop, so it is necessary to filled 0 in SimpleArray.
https://github.com/solvcon/modmesh/blob/2a316f46afde29e490319b38f13edfa9d6d00e91/cpp/modmesh/buffer/ConcreteBuffer.hpp#L294-L297

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fixed.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -937,8 +937,8 @@ detail::SimpleArrayMixinCalculators<A, T>::median_freq(small_vector<value_type>
 }
 
 /**
- * Perform matrix multiplication for 2D arrays.
- * This implementation supports only 2D x 2D matrix multiplication.
+ * Perform matrix multiplication for SimpleArrays.
+ * This implementation supports 2D x 2D, 2D x 1D, 1D x 2D, and 1D x 1D matrix multiplication.
```
yungyuc: List the dimension from low to high: 1D x 1D, 1D x 2D, 2D x 1D, 2D x 2D.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,108 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+                            format_shape(athis),
+                            format_shape(other)));
+        }
+    };
+
+    if ((this_ndim != 2 && this_ndim != 1) || (other_ndim != 2 && other_ndim != 1))
     {
         const std::string err = std::format("SimpleArray::matmul(): unsupported dimensions: "
-                                            "this={} other={}. Only 2D x 2D matrix multiplication is supported",
+                                            "this={} other={}. SimpleArray must be 2D or 1D.",
```
yungyuc: Order dimension in the error message from low to high: 1D or 2D.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: This zero-initialization loop should be removed by using a temporary scalar for the inner loop below.

```cpp
        for (size_t j = 0; j < n; ++j)
        {
            T v = 0.0;
            for (size_t l = 0; l < k; ++l)
            {
                v += (*athis)(l)*other(l, j);
            }
            result.data(j);
        }
```

yungyuc (changes requested):
Good enhancement.  Points to address:

- [x] Update code comments for dimension order.
- [x] Update error messages for dimension order.
- [x] Remove the zero-initialization loop by using a temporary scalar for the inner loop below.
- [x] Write informative summary in the commit log.

ThreeMonth03 (commented):
> Good enhancement. Points to address:
> 
> * [x]  Update code comments for dimension order.
> * [x]  Update error messages for dimension order.
> * [x]  Remove the zero-initialization loop by using a temporary scalar for the inner loop below.
> * [x]  Write informative summary in the commit log.

@yungyuc Please review again. Thanks.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fixed.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Fixed.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,112 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+    else if (this_is_1d)
     {
+        const size_t k = athis->shape(0);
+        const size_t n = other.shape(1);
+        check_product_shape(athis, &other, 0, 0);
+        A result(n);
+
         for (size_t j = 0; j < n; ++j)
         {
+            value_type v = static_cast<value_type>(0);
```
yungyuc: Why do you need to write `value_type v = static_cast<value_type>(0);` instead of just:

```cpp
value_type v = 0;
```

yungyuc (changes requested):
- [x] Clarify why `static_cast` is needed for value initialization instead of just `value_type v = 0`.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Yes, we don't need to explicitly cast here, and I will fix it.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -969,43 +969,112 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
...
+    else if (this_is_1d)
     {
+        const size_t k = athis->shape(0);
+        const size_t n = other.shape(1);
+        check_product_shape(athis, &other, 0, 0);
+        A result(n);
+
         for (size_t j = 0; j < n; ++j)
         {
+            value_type v = 0;
```
ThreeMonth03, line 1018: Fixed.

yungyuc (approved):
LGTM

PR #662, opened 2026-01-20
Fix: Support to assign the complex number to SimpleArray

### LINKED ISSUE #604: SimpleArrayComplex128[i] = numpy.complex128 fails with cast error

ThreeMonth03:
## Problem
In issue #604 , when we assign values from `np.complex128` or `Complex` to `SimpleArrayComplex128`, the cast of python instance fails.

```python
import modmesh as mm
import numpy as np

nc = np.array([1.0 + 0.1j])
mc = mm.SimpleArrayComplex128([1])
mc[0] = nc[0]          # <- fails

print(mc[0].real, mc[0].imag)
```

```shell
  ...
  mc[0] = nc[0]
RuntimeError: Unable to cast Python instance of type <class 'numpy.complex128'> to C++ type '?' 
(#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```

## Solution
The function `__setitem__` in `wrap_SimpleArray.cpp` `calls &property_helper::setitem_parser`, and the fuunction `calls &property_helper::setitem_parser` checks if the value is number based on the following logic:

```c++
const bool is_number = py::isinstance<py::bool_>(py_value) || py::isinstance<py::int_>(py_value) || py::isinstance<py::float_>(py_value) || is_complex_v<T>;
```

This condition ignores whether `py_value` is `np.complex128` or `Complex`, it just check whether `py_value` is `modmesh.complex128`.

### DIFF
--- a/cpp/modmesh/buffer/pymod/array_common.hpp
+++ b/cpp/modmesh/buffer/pymod/array_common.hpp
@@ -174,6 +174,21 @@ inline modmesh::detail::shape_type make_shape(pybind11::object const & shape_in)
     return shape;
 }
 
+template <typename T>
+struct convert_to_std
+{
+    using type = T;
+};
+
+template <typename T>
+struct convert_to_std<Complex<T>>
+{
+    using type = std::complex<T>;
+};
+
+template <typename T>
+using convert_to_std_t = typename convert_to_std<T>::type;
+
 /// Helper class for array property in Python.
 template <typename T>
 class ArrayPropertyHelper
@@ -211,22 +226,38 @@ public:
             const py::object & py_key = args[0];
             const py::object & py_value = args[1];
 
-            const bool is_number = py::isinstance<py::bool_>(py_value) || py::isinstance<py::int_>(py_value) || py::isinstance<py::float_>(py_value) || is_complex_v<T>;
+            // Determine if py_value is a numpy complexx
+            // Reference: https://stackoverflow.com/a/79673025
+            py::object builtins = py::module_::import("builtins");
+            py::object complex_class = builtins.attr("complex");
+            const bool is_number = py::isinstance<py::bool_>(py_value) || py::isinstance<py::int_>(py_value) || py::isinstance<py::float_>(py_value) || is_complex_v<T> || py::isinstance(py_value, complex_class);
 
             // sarr[K] = V
             if (py::isinstance<py::int_>(py_key) && is_number)
             {
                 const auto key = py_key.cast<ssize_t>();
-
-                arr_out.at(key) = py_value.cast<T>();
+                if (is_complex_v<T> && py::isinstance(py_value, complex_class))
+                {
+                    arr_out.at(key) = py_value.cast<convert_to_std_t<T>>();
+                }
+                else
+                {
+                    arr_out.at(key) = py_value.cast<T>();
+                }
                 return;
             }
             // sarr[K1, K2, K3] = V
             if (py::isinstance<py::tuple>(py_key) && is_number)
             {
                 const auto key = py_key.cast<std::vector<ssize_t>>();
-
-                arr_out.at(key) = py_value.cast<T>();
+                if (is_complex_v<T> && py::isinstance(py_value, complex_class))
+                {
+                    arr_out.at(key) = py_value.cast<convert_to_std_t<T>>();
+                }
+                else
+                {
+                    arr_out.at(key) = py_value.cast<T>();
+                }
                 return;
             }
 
--- a/cpp/modmesh/math/Complex.hpp
+++ b/cpp/modmesh/math/Complex.hpp
@@ -29,6 +29,7 @@
 #include <type_traits>
 #include <cmath>
 #include <stdexcept>
+#include <complex>
 
 namespace modmesh
 {
@@ -55,6 +56,12 @@ struct ComplexImpl
     ComplexImpl(T t)
         : real_v(t)
         , imag_v(0.0)
+    {
+    }
+
+    ComplexImpl(std::complex<T> const & c)
+        : real_v(c.real())
+        , imag_v(c.imag())
     {
     }
 
@@ -125,6 +132,7 @@ struct ComplexImpl
         return *this;
     }
 
+    std::complex<T> to_std_complex() const { return std::complex<T>(real_v, imag_v); }
     T real() const { return real_v; }
     T imag() const { return imag_v; }
     T norm() const { return real_v * real_v + imag_v * imag_v; }
--- a/cpp/modmesh/math/pymod/wrap_Complex.cpp
+++ b/cpp/modmesh/math/pymod/wrap_Complex.cpp
@@ -71,7 +71,7 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapComplex
             .def(
                 py::init(
                     [](const wrapped_type & other)
-                    { return std::make_shared<wrapped_type>(wrapped_type{other.real_v, other.imag_v}); }),
+                    { return std::make_shared<wrapped_type>(other); }),
                 py::arg("complex"))
             .def(
                 py::init(
@@ -106,6 +106,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapComplex
             .def_readonly("imag", &wrapped_type::imag_v)
             .def("norm", &wrapped_type::norm)
             .def("conj", &wrapped_type::conj)
+            .def("__complex__",
+                 [](const wrapped_type & self)
+                 {
+                     return self.to_std_complex();
+                 })
             .def("dtype",
                  []()
                  { return complex_dtype; });
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -620,6 +620,23 @@ class SimpleArrayBasicTC(unittest.TestCase):
         sarr.ndarray.fill(100)
         self.assertTrue((ndarr == 100).all())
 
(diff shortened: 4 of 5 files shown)


cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -211,21 +211,39 @@ class ArrayPropertyHelper
...
                 const auto key = py_key.cast<ssize_t>();
-
+                if constexpr (is_complex_v<T>)
+                {
+                    if (py::isinstance(py_value, complex_class))
+                    {
+                        arr_out.at(key) = py_value.cast<std_complex_t<T>>();
+                        return;
+                    }
+                }
```
ThreeMonth03: When `py_value` is `Complex` or `np.complex`, casting to std::complex at first.

cpp/modmesh/math/Complex.hpp
```diff
@@ -261,6 +269,23 @@ struct is_complex<Complex<T>> : std::true_type {};
...
+{
+    using type = T;
+};
+
+template <typename T>
+using complex_element_type_t = typename complex_element_type<T>::type;
+
+template <typename T>
+using std_complex_t = std::complex<complex_element_type_t<T>>;
+
```
ThreeMonth03: Given the type of `complex<T>`, return `std::complex<T>`

cpp/modmesh/math/pymod/wrap_Complex.cpp
```diff
@@ -106,6 +106,11 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapComplex
             .def_readonly("imag", &wrapped_type::imag_v)
             .def("norm", &wrapped_type::norm)
             .def("conj", &wrapped_type::conj)
+            .def("__complex__",
+                 [](const wrapped_type & self)
+                 {
+                     return self.to_std_complex();
+                 })
```
ThreeMonth03, line 113: Add an interface to support casting.

ThreeMonth03 (commented):
@yungyuc @tigercosmos Please review this pull request. Thanks.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: The if/else code is unbalanced and hard to read:

```cpp
if constexpr (is_complex_v<T>)
{
    if (py::isinstance(py_value, complex_class))
    {
        arr_out.at(key) = py_value.cast<std_complex_t<T>>();
        return;
    }
}
arr_out.at(key) = py_value.cast<T>();
return;
```

The following code is more concise and maintainable:

```cpp
if (is_complex_v<T> && py::isinstance(py_value, complex_class))
{
    arr_out.at(key) = py_value.cast<std_complex_t<T>>();
}
else
{
    arr_out.at(key) = py_value.cast<T>();
}
```

`constexpr` is unnecessary because `py::isinstance(py_value, complex_class)` needs to happen during runtime rather than compile time.

yungyuc (changes requested):
Good job.  Points to address:

- [x] Balance the if/else to improve readability for the value assignment.
- [x] Write an informative summary in the commit log.

cpp/modmesh/buffer/pymod/array_common.hpp
ThreeMonth03: However, the program will return compile error if we don't use `if constexpr`.
```c++
template <typename T>
struct complex_element_type
{
};

template <typename T>
struct complex_element_type<Complex<T>>
{
    using type = T;
};

template <typename T>
using complex_element_type_t = typename complex_element_type<T>::type;

template <typename T>
using std_complex_t = std::complex<complex_element_type_t<T>>;
```

The following message is compile error.
```shell
/home/threemonth03/Downloads/modmesh/cpp/modmesh/buffer/pymod/wrap_SimpleArrayPlex.cpp:325:55:   required from here
/home/threemonth03/Downloads/modmesh/cpp/modmesh/math/Complex.hpp:287:7: error: no type named ‘type’ in ‘struct modmesh::complex_element_type<double>’
In file included from /home/threemonth03/Downloads/modmesh/cpp/modmesh/math/math.hpp:8,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/buffer/SimpleArray.hpp:32,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/buffer/buffer.hpp:38,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/toggle/toggle.hpp:32,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/python/common.hpp:38,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/buffer/pymod/buffer_pymod.hpp:34,
                 from /home/threemonth03/Downloads/modmesh/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp:29:
```

yungyuc: Interesting.  Please figure out why.

ThreeMonth03: > Interesting. Please figure out why.

The error message is simple, it caused from the missing alias in generic template.
```c++
template <typename T>
struct complex_element_type
{
};
```

However, if we add the alias `using type = T;` in generic template, it would deduce to a wrong type, so I think I should redesign the template , such that it might look like `using convert_to_std_t = something...;`.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -174,6 +174,21 @@ inline modmesh::detail::shape_type make_shape(pybind11::object const & shape_in)
...
+
+template <typename T>
+struct convert_to_std<Complex<T>>
+{
+    using type = std::complex<T>;
+};
+
+template <typename T>
+using convert_to_std_t = typename convert_to_std<T>::type;
+
```
ThreeMonth03, line 191: The template could identify not only `modmesh::complex<T>` but also standard type.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -211,21 +226,33 @@ class ArrayPropertyHelper
...
             // sarr[K] = V
             if (py::isinstance<py::int_>(py_key) && is_number)
             {
                 const auto key = py_key.cast<ssize_t>();
-
+                if (is_complex_v<T> && py::isinstance(py_value, complex_class))
+                {
+                    arr_out.at(key) = py_value.cast<convert_to_std_t<T>>();
+                    return;
+                }
```
ThreeMonth03, line 242: Remove `if constexpr`.

ThreeMonth03 (commented):
> * [x]  Balance the if/else to improve readability for the value assignment.
> * [x]  Write an informative summary in the commit log.

@yungyuc Please review this pull request again. Thanks.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: Please balance if/else as I mentioned in the previous review:

```cpp
if (/* ... */)
{
    // ...
}
else
{
    // ...
}
```

The early `return` should be factored outside the branch.

yungyuc (changes requested):
The commit log looks good, but you did not balance the branch.

cpp/modmesh/buffer/pymod/array_common.hpp
ThreeMonth03: I notice that if we replace the branch with the following code, it also could work.
```c++
arr_out.at(key) = py_value.cast<convert_to_std_t<T>>();
```
However, if the input is `modmesh::complex<T>`, it casts to `std::complex<T>` at first, then it casts to  `modmesh::complex<T>`. I'm not sure which design is better.

ThreeMonth03 (commented):
@yungyuc Please review this pr. Thanks.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: Interesting.  Right, @j8xixo12 ?

Please study it if you have time.  For now the branching is fine because the cycle cost should be too low to appear in benchmark.

yungyuc (approved):
LGTM

Issue #648, opened 2025-12-07
Upgrade cmake to be higher than version 4


yungyuc:
The cmake version we use is pretty old.  We should upgrade to version 4 or newer.

```bash
$ ag cmake_minimum_required
CMakeLists.txt:4:cmake_minimum_required(VERSION 3.27)
cmake/FindNumPy.cmake:6:cmake_minimum_required(VERSION 3.9)
cpp/binary/pilot/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/binary/pymod_modmesh/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/mesh/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/multidim/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/linalg/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/inout/CMakeLists.txt:2:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/python/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/serialization/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/onedim/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/simd/CMakeLists.txt:27:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/math/CMakeLists.txt:1:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/toggle/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/testhelper/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/pilot/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/device/metal/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/universe/CMakeLists.txt:4:cmake_minimum_required(VERSION 3.16)
cpp/modmesh/transform/CMakeLists.txt:1:cmake_minimum_required(VERSION 3.16)
[...]

PR #664, opened 2026-01-25
Upgrade cmake and update minimum version in cmake files

ThreeMonth03:
## Problem
According to the discussion in issue #648, we decide to upgrade the CMake to version>=4.

## Solution
The latest version of cmake is `4.2.3`. Due to the [script in devenv](https://github.com/solvcon/devenv/blob/7b8eca4859510dbedc67bcc7402fe740db3882f5/scripts/build.d/cmake#L11), this pull request upgrades the minimum version of CMakeList.txt to 4.0.1, and upgrades cmake>=4.0.1 in GA runner.

The following information is the current cmake version of different operation systems in GA runner:
 - ubuntu-24.04: 4.2.3
 - windows-2022: 4.2.3
 - macos-15: 4.2.2


.github/workflows/devbuild.yml
```diff
@@ -37,9 +37,15 @@ jobs:
...
-              sudo curl git build-essential make cmake libc6-dev gcc g++ \
+              sudo curl git build-essential make libc6-dev gcc g++ \
               python3 python3-dev python3-venv
 
+      # The latest CMake version from apt-get is 3.28.3, so we install CMake>=4.0.1 from github marketplace.
+      - name: install CMake=4.0.1
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 4.0.1
+
```
ThreeMonth03, line 59: Upgrade CMake from another github actions.

.github/workflows/nouse_install.yml
```diff
@@ -84,9 +90,11 @@ jobs:
         echo "flake8 path: $(which flake8)"
         echo "flake8 version: $(flake8 --version)"
 
+    # Pass local environment variables to sudo
+    # Reference: https://askubuntu.com/a/713135
     - name: setup.py install build_ext
       run: |
-        sudo python3 setup.py install build_ext \
+        sudo env "PATH=$PATH" python3 setup.py install build_ext \
```
ThreeMonth03: When running the command with sudo, the command would use the original cmake and raise error.
In this [discussion](https://github.com/lukka/get-cmake/issues/223), the author suggests to pass the enviroment variables to bypass the problem.

ThreeMonth03 (commented):
@yungyuc Please review this issue. Thanks.

cmake/FindNumPy.cmake
```diff
@@ -3,7 +3,7 @@
 # NUMPY_FOUND
 # will be set by this script
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 4.0.1)
```
tigercosmos, line 6: I wonder if it is possible to have a global config so we don't need to modify all CMake files' values every time when we upgrade CMake. Not sure if this makes sense or not, though.

.github/workflows/devbuild.yml
yungyuc: Do not use github actions to install dependencies unless there is no other ways, e.g., for Windows or for very complex binaries like Qt.

Installing cmake using apt provides a clear list of dependencies for new contributors.  The clarity matters more.

Check the `get-cmake` action code and learn how to integrate it in this action YML.  That is, copy and paste.

.github/workflows/devbuild.yml
```diff
@@ -523,6 +535,13 @@ jobs:
 
       - uses: ilammy/msvc-dev-cmd@v1
 
+      # CMake version in windows-2022 runner-image is 3.31.6, so we install CMake>=4.0.1 from github marketplace.
```
yungyuc, line 565: I can tolerate using a github action for Windows, because there is not a clearer way to do it.

.github/workflows/nouse_install.yml
yungyuc: This is another reason to install cmake to the runner system directory.

cmake/FindNumPy.cmake
yungyuc: I don't know if there is a global way, but cmake version requirement should be repeated in every cmake file.  It also serves as the purpose of documentation.  Making it too "programming" defeats the purpose.

We won't update the minimum requirement for all files every time we upgrade cmake in runner.  This kind of update will be extremely infrequent, maybe once in a couple of years.  Manual work is more productive in this case.

yungyuc (changes requested):
I don't like semantic/conventional commit messages, because they oversimplifies development work, and give maintainers a wrong impression that the changes are organized.  Never use them in this project.  Write a summary in good English.

Commits are seldom organized.  Don't fool ourselves.

Point to address:

- [x] Do not use `get-cmake` (or any) github action for installing cmake.

.github/workflows/devbuild.yml
ThreeMonth03: OK, I decide not to modify this script.

.github/workflows/nouse_install.yml
```diff
@@ -37,10 +37,27 @@ jobs:
...
+        test -f /usr/share/doc/kitware-archive-keyring/copyright || \
+            sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
+        sudo apt-get -qy install kitware-archive-keyring
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ \
             python3 python3-dev python3-venv
+        # The path of apt-get cmake is `/usr/bin/cmake`,
+        # but the path of built-in cmake of runner image is `/usr/local/bin/cmake`.
+        # We remove the built-in cmake.
+        sudo rm -rf /usr/local/bin/cmake
```
ThreeMonth03, line 60: In order to build the latest cmake, we add the new apt repository named `kitware`.

ThreeMonth03 (commented):
> I don't like semantic/conventional commit messages, because they oversimplifies development work, and give maintainers a wrong impression that the changes are organized. Never use them in this project. Write a summary in good English.
> 
> Commits are seldom organized. Don't fool ourselves.
> 
> Point to address:
> 
> * [x]  Do not use `get-cmake` (or any) github action for installing cmake.

yungyuc (approved):
LGTM

Issue #461, opened 2025-01-30
Improve Python-domain docstrings in `modmesh.pilot._euler1d`


yungyuc:
Docstrings in [`modmesh.pilot._euler1d` (e.g., `L70`)](https://github.com/solvcon/modmesh/blob/59ca28f167b13e485b4b3c1d844278547658b5b4/modmesh/pilot/_euler1d.py#L70) should use [Sphinx Python domain markups](https://www.sphinx-doc.org/en/master/usage/domains/python.html) instead of the custom format.  For example, the docstrings

```
Class attributes:
    - `ana` (matplotlib.lines.Line2D): Line2D for analytical data.
    - `num` (matplotlib.lines.Line2D): Line2D for numerical data.
    ...
```

should be changed to

```
:ivar ana: Drawn line for analytical solution
:type ana: matplotlib.lines.Line2D
:ivar num: Drawn line for numerical solution
:type num: matplotlib.lines.Line2D
...
```

yungyuc:
@j8xixo12 , could you please take a look and let me know what do you think?

ThreeMonth03:
Hi, @yungyuc. I'm interested in solving this issue. Do you recommend to build the document simply when solving this issue?

yungyuc:
> Hi, [@yungyuc](https://github.com/yungyuc). I'm interested in solving this issue. Do you recommend to build the document simply when solving this issue?

Thanks.  Feel free to update the docstrings.  It is not needed to build documents.

The issue is assigned to you.

ThreeMonth03:
@yungyuc By the way, I just built a simple documentby Sphinx, and I notice that currently Sphinx supports [napoleon extension](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/), which enables Sphinx to parse both [NumPy](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html) and [Google](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) style docstrings. How do you think?

yungyuc:
[...]

PR #669, opened 2026-02-01
Rewrite docstring with Sphinx-style in `modmesh/**.py`

ThreeMonth03:
According to the discussion in pr #461, I decide to rewrite docstring with Sphinx format. With changing the format, we could establish a better document website in the future.

For example, if we write python code with Sphinx-style docstring:
```python3
class Foo:
    """
    Example class.

    :cvar a: Discription of a.
    :vartype a: int
    :ivar b: Discription of b.
    :vartype b: int

    :Methods:
        - :py:meth:`do_something`: Return a list of Bar class.
    """

    def __init__(self, b):
        self.b = b

    def do_something(self, a: int, b: int):
        """
        :param a: Do nothing
        :type a: int
        :param b: Do nothing
        :type b: int
        :return: Some complex type
        :rtype: list[:class:`Bar`, 
            :class:`Bar`]
        """
        m = Bar()
        n = Bar()
        return [m, n]


class Bar:
    pass
```

The document would be rendered like:
Screenshot_2026-02-01_21-15-24


modmesh/pilot/_base_app.py
```diff
@@ -56,20 +56,26 @@ class QuantityLine(object):
     in a plot. It is designed to be used in conjunction with Matplotlib for
     visualizing analytical and numerical data.
 
-    Class attributes:
-        - `ana` (matplotlib.lines.Line2D): Line2D for analytical data.
-        - `num` (matplotlib.lines.Line2D): Line2D for numerical data.
```
ThreeMonth03, line 61: We don't need to add the line about ` Class attributes:`, because wnen we define the `cvar` or `ivar`, the sphinx would automatically add the prefix `variables`.

modmesh/pilot/_base_app.py
```diff
@@ -56,20 +56,26 @@ class QuantityLine(object):
...
+    :cvar color: Color of line or symbol.
+    :vartype color: str
+    :cvar y_upper_lim: y axis upper limit.
+    :vartype y_upper_lim: float
+    :cvar y_bottom_lim: y axis bottom limit.
+    :vartype y_bottom_lim: float
+
+    :Methods:
+        - :py:meth:`update_ana`: Update the line data and redraw the plot.
+        - :py:meth:`update_num`: Update the line data and redraw the plot.
```
ThreeMonth03, line 78: Unfortunately, when using `:py:meth: some_method`, the prefix of method is not rendered. Therefore, I decide to add the prefix `:Methods:` here.

modmesh/pilot/_base_app.py
```diff
@@ -155,24 +162,23 @@ class GUIConfig(object):
...
-            table columns.
+    :ivar: _tbl_content (list[list]): The content of the configuration
+        table.
+    :ivar: _col_header (list): The header for the configuration
+        table columns.
 
-    Methods:
-        :meth:`data(row, col)`: Get the value at a specific row and column
+    :Methods:
+        - :py:meth:`data`: Get the value at a specific row and column
```
ThreeMonth03, line 171: With the format :py:meth: member: , it can link the class to the corresponding member function. However, I don't find any method to render parameters here properly.

modmesh/pilot/_base_app.py
```diff
@@ -266,11 +272,6 @@ class SolverConfig(GUIConfig):
 
     This class provides a configuration interface for the solver, allowing
     users to set and retrieve parameters related to the simulation.
-
-    Attributes:
-        - `_tbl_content` (list[list]): The content of the configuration table.
-        - `_col_header` (list): The header for the configuration
-          table columns.
```
ThreeMonth03, line 273: I don't find `_tbl_content` and `_col_header` here, so I delete them.

modmesh/pilot/_base_app.py
```diff
@@ -283,11 +284,6 @@ class PlotConfig(GUIConfig):
 
     This class provides a configuration interface for the plot, allowing
     users to set and retrieve parameters related to the plotting arae.
-
-    Attributes:
-        - `_tbl_content` (list[list]): The content of the configuration table.
-        - `_col_header` (list): The header for the configuration
-          table columns.
     """
```
ThreeMonth03, line 287: I don't find `_tbl_content` and `_col_header` here, so I delete them.

ThreeMonth03 (commented):

modmesh/pilot/_base_app.py
yungyuc: It's OK.  We treat the docstrings as comments and the consistent style is for aligning among contributors.  We do not render documents from docstrings.

modmesh/pilot/_base_app.py
yungyuc: OK.

yungyuc (approved):
LGTM

PR #674, opened 2026-02-04
Implement OASIS file writer

c1ydehhx:
In this PR, I added OASIS device. This device support writer only currently. This writer implement manhattan delta format by OASIS spec.

For example, if we expected +340 for N/S direction, it will be encoded with 0xD0 0x0A0xA8 0x05, and -340 will be 0xD2 0x0A0xA9 0x05.

I added some test to make sure it work properly, and I add some edge case to make sure it will not buggy when met edge case.

---

Updated (2026/02/19):

After the discussion with @yungyuc, I added 3 classes `modmesh::OasisWriter`, `modmesh::OasisRecordRect`, `modmesh::OasisRecordPoly` to output OASIS-format bytes.

The two record classes convert geometry information into OASIS-format bytes:
 - `OasisRecordPoly` class accepts coordinate pairs (e.g., `[[410, 720], [410, 920], [70, 920], ...]`) and converts them into OASIS-format bytes.
 - `OasisRecordRect` class accepts a lower-left coordinate pair (e.g., `[70, 800]`), width, height and converts them into OASIS-format bytes.

With these record classes, `OasisWriter` accepts `OasisRecordPoly` and `OasisRecordRect` objects and integrates these records with the minimal required headers (e.g., magic-bytes, START, END, CELL, and CELLNAME records) and outputs OASIS-format bytes. The output bytes should be loadable by third-party software.

These classes above are wrapped as Python classes via pybind11. In this PR, I added a few unit tests to verify the classes work properly:

 - `OasisRecordPoly` and `OasisRecordRect`: verify that the produced OASIS-format bytes are correct.
 - `OasisWriter`: verify the output bytes contain the required headers with integrated polygon and rectangle records.
 
In addition, I tested the output bytes manually with [KLayout software](https://www.klayout.de/). The output bytes from `OasisWriter` is accepted by KLayout, as shown below.

image

### DIFF
--- a/cpp/modmesh/oasis/oasis_device.cpp
+++ b/cpp/modmesh/oasis/oasis_device.cpp
@@ -1,0 +1,302 @@
+/*
+ * Copyright (c) 2026, Han-Xuan Huang <c1ydehhx@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/oasis/oasis_device.hpp>
+#include <cstdint>
+#include <utility>
+
+namespace modmesh
+{
+
+/*
+ * Convert value to OASIS unsigned integer bytes and append to segment array.
+ * Encodes payload as a base-128 variable-length sequence, emitting 7 data
+ * bits per byte and using the MSB as a continuation flag.
+ * Refer OASIS Draft 7.2.1 for more information.
+ */
+void OasisDevice::append_unsigned_integer(std::vector<uint8_t> & segment, int value)
+{
+    int payload = value;
+
+    if (payload == 0)
+    {
+        segment.push_back(0x00);
+    }
+
+    while (payload >= 1)
+    {
+        int first_bit = payload >= 128 ? 1 : 0;
+        segment.push_back((first_bit << 7) | (payload % 128));
+        payload /= 128;
+    }
+}
+
+/*
+ * Convert value to OASIS signed integer bytes and append to segment array.
+ * 1. DIR_BIT will be 1 if value is negative, otherwise, DIRBIT will be 0.
+ * 2. The payload will left-shift 1 bit on value and do OR with DIR_BIT.
+ * 3. Encodes payload as a base-128 variable-length sequence, emitting 7 data
+ *    bits per byte and using the MSB as a continuation flag. This operation is
+ *    as same as OASIS unsigned interger bytes.
+ * Refer OASIS Draft 7.2.2 for more information.
+ */
+void OasisDevice::append_signed_integer(std::vector<uint8_t> & segment, int value)
+{
+    const int DIR_BIT = value < 0 ? 1 : 0;
+    int delta_codec = abs(value) << 1 | DIR_BIT;
+
+    int payload = delta_codec;
+
+    if (payload == 0)
+    {
+        segment.push_back(0x02);
+    }
+
+    while (payload >= 1)
+    {
+        int first_bit = payload >= 128 ? 1 : 0;
+        segment.push_back((first_bit << 7) | (payload % 128));
+        payload /= 128;
+    }
+}
+
+std::vector<uint8_t> OasisRecordRect::to_bytes() const
+{
+    // RECTANGLE record should be
+    // '20' info-bytes [layer] [datatype] [w] [h] [x] [y] [repetition]
+    // Please refer OASIS Draft section 25.
+    std::vector<uint8_t> segment;
+
+    segment.push_back(0x14);
+
+    const int S = (m_w == m_h); // Is square? (1 if yes, 0 if no)
+    const int W = 1; // Have width? (1 if yes, 0 if no)
+    const int H = (m_w != m_h); // Have height? (1 if yes, 0 if no, must be 0 if S = 1)
+    const int X = 1;
+    const int Y = 1;
+    const int R = 0;
+    const int D = 1; // Have Datatype? (1 if yes, 0 if no)
+    const int L = 1; // Have Layer? (1 if yes, 0 if no)
+
+    const int INFO = (S << 7) | (W << 6) | (H << 5) |
+                     (X << 4) | (Y << 3) | (R << 2) | (D << 1) | L;
+    segment.push_back(INFO);
+
+    // Layer-num and datatype-num (0 in default).
+    segment.push_back(0x00);
+    segment.push_back(0x00);
+
+    OasisDevice::append_unsigned_integer(segment, m_w);
+    OasisDevice::append_unsigned_integer(segment, m_h);
+    OasisDevice::append_signed_integer(segment, m_left);
+    OasisDevice::append_signed_integer(segment, m_lower);
+
+    return segment;
+}
+
+std::vector<uint8_t> OasisRecordPoly::to_bytes() const
+{
+    // Polygon record should be
+    // '21' 00PXYRDL [layer-num] [datatype-num] [point-list] [x] [y] [rep]
+    // Please refer OASIS Draft section 26.
+    std::vector<uint8_t> segment;
+
+    segment.push_back(0x15);
+
+    // Info bytes:
+    // - Have point-list, X and Y.
+    // - Have datatype and layer (Use 0)
+    const int P = 1; // Have Polygon-List? (1 if yes, 0 if no)
+    const int X = 1;
+    const int Y = 1;
+    const int R = 0;
+    const int D = 1; // Have Datatype? (1 if yes, 0 if no)
+    const int L = 1; // Have Layer? (1 if yes, 0 if no)
+    const int INFO = (P << 5) | (X << 4) | (Y << 3) | (R << 2) | (D << 1) | L;
+    segment.push_back(INFO);
+
+    // Layer-num and datatype-num (0 in default).
+    segment.push_back(0x00);
+    segment.push_back(0x00);
+
+    // The 1-Delta have two different type:
+    //  - Type 0: Start with horizontal
+    //  - Type 1: Start with vertical
+    if (m_vertices[0].second == m_vertices[1].second)
+    {
+        segment.push_back(0x00);
+    }
+    else if (m_vertices[0].first == m_vertices[0].first)
+    {
+        segment.push_back(0x01);
+    }
+
+    // The vertex count shoud be (vertex - 1)
+    segment.push_back(m_vertices.size() - 1);
+
+    // In this implementation, point-list only support 1-delta format.
+    // Please refer Point-list in OASIS draft 7.7.
+    std::vector<uint8_t> point_list;
+
+    for (int i = 0; i < m_vertices.size() - 1; i++)
+    {
+        std::pair<int, int> curr_v = m_vertices[i];
+        std::pair<int, int> next_v = m_vertices[i + 1];
+
+        // Convert delta value to OASIS signed interegr bytes.
+        OasisDevice::append_signed_integer(
+            point_list,
+            (next_v.first - curr_v.first) + (next_v.second - curr_v.second));
+    }
+
+    segment.insert(segment.end(), point_list.begin(), point_list.end());
+
+    // X value
+    OasisDevice::append_signed_integer(segment, m_vertices[0].first);
+
+    // Y value
+    OasisDevice::append_signed_integer(segment, m_vertices[0].second);
+
+    return segment;
+}
+
+std::vector<uint8_t> OasisDevice::to_bytes()
+{
+    // The simple OASIS byte format should be:
+    // <magic-byte> <START>
+    // <CELLNAME> <CELL>
+    // <GEOMETRY_RECORD_1> <GEOMETRY_RECORD_2> ... <GEOMETRY_RECORD_N>
+    // <END>
+    std::vector<uint8_t> result;
+
+    append_magic_bytes(result);
+    append_start_record_bytes(result);
+    append_cell_and_cell_name_record_byte(result);
+
+    for (const OasisRecordRect & record : m_rect_records)
+    {
+        append_record_bytes(result, record);
+    }
+
+    for (const OasisRecordPoly & record : m_polygon_records)
+    {
+        append_record_bytes(result, record);
+    }
+
+    append_end_record_byte(result);
+
+    return result;
+}
+
+void OasisDevice::add_poly_record(const OasisRecordPoly & record)
+{
+    m_polygon_records.push_back(record);
+}
+
+void OasisDevice::add_rect_record(const OasisRecordRect & record)
+{
+    m_rect_records.push_back(record);
+}
+
+void OasisDevice::append_magic_bytes(std::vector<uint8_t> & segment)
+{
+    // Magic byte should be %SEMI-OASIS<CR><NL>, which <CR><NL> is 0x0D 0x0A.
+    // Please refer OASIS Draft section 6.4.
+
+    std::string magic_byte = "%SEMI-OASIS\x0D\x0A";
+
+    segment.insert(segment.end(), magic_byte.begin(), magic_byte.end());
+}
+
+void OasisDevice::append_start_record_bytes(std::vector<uint8_t> & segment)
+{
+    // START record should be
+    // '1' version-string unit offset-flag [ table-offsets ]
+    // Please refer OASIS Draft section 13.
+
+    // The first byte should 1.
+    segment.push_back(0x01);
+
+    // The version string will be [LENGTH][STRING-ASCII], we
+    // fixed it as "1.0".
+    segment.push_back(0x03);
+
+    std::string version = "1.0";
+    segment.insert(segment.end(), version.begin(), version.end());
+
+    // The unit we using 0.001 as default.
+    std::vector<uint8_t> unit = {0x00, 0xE8, 0x07};
+    segment.insert(segment.end(), unit.begin(), unit.end());
+
+    // The offset-flag we using 0x00 as default.
+    // The table-offset will store in END record if offset-flag is 0x01.
+    // We put 6 pairs of 0x00 to describe that we don't have any tables.
+    // Therefore, it insert 13 0x00 in total.
+    segment.insert(segment.end(), 13, 0x00);
+}
+
+void OasisDevice::append_cell_and_cell_name_record_byte(std::vector<uint8_t> & segment)
+{
+    // Append CELLNAME record, the default CELLNAME is TOP.
+    // TODO: This record is fixed. It can be modify by user in the future.
+    std::vector<uint8_t> cellname = {0x03, 03, 0x54, 0x4F, 0x50};
+    segment.insert(segment.end(), cellname.begin(), cellname.end());
+
+    // Append CELL record.
+    // TODO: This record is fixed. It can be modify by user in the future.
+    std::vector<uint8_t> cell = {0x0D, 0x00};
+    segment.insert(segment.end(), cell.begin(), cell.end());
+}
+
+void OasisDevice::append_end_record_byte(std::vector<uint8_t> & segment)
+{
+    // END record should be
+    // '2' [table-offsets] padding validation-scheme [validation-signature]
+    // Please refer OASIS Draft section 14.
+
+    // The first byte should 2.
+    segment.push_back(0x02);
+
+    // Write padding. The padding should be 256 bytes.
+    int padding_length = 254;
+    segment.insert(segment.end(), padding_length, 0x00);
+
+    // Validation-scheme: No validation.
+    segment.push_back(0x00);
+}
+
+template <typename T>
+void OasisDevice::append_record_bytes(std::vector<uint8_t> & bytes, const T & rec)
+{
+    std::vector<uint8_t> rec_bytes = rec.to_bytes();
+    bytes.insert(bytes.end(), rec_bytes.begin(), rec_bytes.end());
+}
+
+} // namespace modmesh
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/modmesh/oasis/oasis_device.hpp
+++ b/cpp/modmesh/oasis/oasis_device.hpp
@@ -1,0 +1,136 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, Han-Xuan Huang <c1ydehhx@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/base.hpp>
+
+#include <cstdint>
+#include <utility>
+#include <vector>
+
+namespace modmesh
+{
+
+class OasisRecordPoly;
+class OasisRecordRect;
+
+/**
+ * \class OasisDevice
+ * \brief OASIS device converts coordinates information to OASIS format.
+ *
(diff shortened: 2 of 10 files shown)


modmesh/plot/plane_layer.py
```diff
@@ -25,6 +25,35 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 
+class ManhattanDeltaCodec:
```
yungyuc: Write I/O code in C++.  Python is too slow to be taken seriously for I/O.

Name it `modmesh::OasisDevice` for now.

tests/test_plane_layer.py
```diff
@@ -29,6 +29,49 @@
 from modmesh.plot import plane_layer
 
 
+class ManhattanDeltaCodecTC(unittest.TestCase):
```
yungyuc: Unit tests for the OASIS I/O should be in Python.

yungyuc (changes requested):
Points to address:

- [x] Write OASIS I/O code in C++.  Name it `modmesh::OasisDevice`.  Wrap it to Python for testing.
- [x] Manually verify the written file by loading it using a third-party viewer and paste the screenshot.

c1ydehhx:
Hi @yungyuc,
So far away, I already migrate Manhatten 1-delta writer (Named it OasisWriter) from Python to C++, and use pybind11 to wrap it as a module.

>  Write OASIS I/O code in C++. Name it modmesh::OasisDevice. Wrap it to Python for testing.

Done.

> Manually verify the written file by loading it using a third-party viewer and paste the screenshot.

I'm concern that it will have too much diff in this PR if we want to output OASIS file. Althought we have geometry record now, we don't have some necessary OASIS header like magic bytes, START and END.

I would like to confirm that this PR should support OASIS file output or we can just test Manhatten 1-delta work properly with unit tests?

yungyuc:
Linter did not pass.  Please fix code format and then request for review again.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/base.hpp>
+#include <pybind11/stl.h>
+
+#include <cstdint>
+#include <vector>
+
+using bytes = std::vector<uint8_t>;
```
yungyuc: Do not make an intrusive alias like this.  It is horrible to maintain.

There is nothing wrong with using `std::vector<uint8)_t>`?  Just spell out the type.

It is horrible to add an alias in the global name space.  Unacceptable almost always.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace oasis
+{
+
+class OasisDevice
+{
+private:
+    static void append_1_delta(bytes & segment, int value);
+
+public:
+    OasisDevice() = default;
```
yungyuc: Rule of 5.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,59 @@
...
+    static bytes writer(std::vector<int32_t> & values);
```
yungyuc: Name it `write`, not `writer`, which is not a verb.

yungyuc (changes requested):
Some very basic points to address:

- [x] Do not make an intrusive alias in global namespace and for STL things like `std::vector<uint8)_t>`.
- [x] Rule of 5 for `OasisDevice`.
- [x] Name as `OasisDevice::write`, not `writer`, which is not a verb.

yungyuc:
@tigercosmos @KHLee529 please help review

cpp/modmesh/oasis/pymod/wrap_oasis.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <modmesh/oasis/oasis_device.hpp>
+
+namespace modmesh
+{
+
+namespace python
+{
+
+using namespace modmesh::oasis; // NOLINT(google-build-using-namespace)
```
tigercosmos: I recommend remove this.  Namespace is important for reading the code.

cpp/modmesh/oasis/oasis_device.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace modmesh
+{
+
+namespace oasis
+{
+
+void OasisDevice::append_1_delta(std::vector<uint8_t> & segment, int value)
+{
+    int dir_bit = value < 0 ? 1 : 0;
```
tigercosmos: Please add `const` for all the places if applicable.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <cstdint>
+#include <vector>
+
+namespace modmesh
+{
+
+namespace oasis
+{
+
+class OasisDevice
```
tigercosmos, line 53: Please add document comments for classes and functions.

cpp/modmesh/oasis/oasis_device.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    while (payload >= 1)
+    {
+        int first_bit = payload >= 128 ? 1 : 0;
+        segment.push_back((first_bit << 7) | (payload % 128));
+        payload /= 128;
+    }
+}
+
+std::vector<uint8_t> OasisDevice::write(std::vector<int32_t> & values)
```
tigercosmos: `const std::vector<int32_t> & values`

cpp/modmesh/oasis/oasis_device.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <modmesh/oasis/oasis_device.hpp>
+#include <cstdint>
+
+namespace modmesh
+{
+
+namespace oasis
+{
+
+void OasisDevice::append_1_delta(std::vector<uint8_t> & segment, int value)
```
tigercosmos: This can be a local function in cpp only.

cpp/modmesh/oasis/oasis_device.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    if (payload == 0)
+    {
+        segment.push_back(0x02);
+    }
+
+    while (payload >= 1)
+    {
+        int first_bit = payload >= 128 ? 1 : 0;
+        segment.push_back((first_bit << 7) | (payload % 128));
```
tigercosmos, line 83: Please add comments to explain the algorithm. Including source doucment is even better.

modmesh/plot/plane_layer.py
tigercosmos: Please remove it.

tests/test_oasis_device.py
```diff
@@ -0,0 +1,75 @@
...
+        device = modmesh.OasisDevice()
+        seg = device.writer([-340])
+
+        self.assertEqual(seg, [0xD2, 0x0A])
+
+    def test_with_integers(self):
+        device = modmesh.OasisDevice()
+        seg = device.writer([+340, +200, -340, -40, +300, -120, -300])
+
+        self.assertEqual(seg, [0xD0, 0x0A, 0xA0, 0x06, 0xD2,
```
tigercosmos: It will be nice if you can explain what you try to test in the comments.

cpp/modmesh/oasis/pymod/wrap_oasis.cpp
yungyuc: Good catch.  `modmesh::oasis` namespace is not necessary and should be removed.

And, never using namespace unless it is in a very limited scope and can resolve extreme pain.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <modmesh/base.hpp>
+#include <pybind11/stl.h>
+
+#include <cstdint>
+#include <vector>
+
+namespace modmesh
+{
+
+namespace oasis
```
yungyuc: Do not use namespace `modmesh::oasis`.

tests/test_oasis_device.py
yungyuc: Use code comments to explain.

yungyuc (changes requested):
Points to address:

- [x] Do not use namespace `modmesh::oasis`.
- [x] Add very concise code comments for the key points in unit tests.

c1ydehhx:
Hi @yungyuc and @tigercosmos.

Thanks for the review. 

I complete OASIS device to store polygon and rectangle and write as OASIS format. The OASIS format will need magic-byte, START record, END record, CELLNAME record and CELL record to work. For each record, I add [comment to describe its format and how to find the record spec in OASIS spec draft](https://github.com/solvcon/modmesh/pull/674/changes#diff-80d53e0d8adc09a351b6d83e630454c2d6414fe7a17a38bfed1a150ff10f95d8). I also add some comment to describe how unsigned and signed value work in OASIS. Please refer the comment in `oasis_device.cpp`. For the comment of tests. I leave a simple commant to let viewer refer the comment in `oasis_device.cpp`, which is more complete, and not leave too much words in test comment.

Also, the `write()` function in OASIS will return a list of bytes that represent the bytes in OASIS file. I write a simple script to check it works. The test script is not include in this PR. The script looks like:

```py
import modmesh

device = modmesh.OasisDevice()

rec_poly = modmesh.OasisPolyRecord([
    [70, 720], [410, 720], [410, 920], [70, 920],
    [70, 880], [370, 880], [370, 760], [70, 760]])
rec_rect = modmesh.OasisRectRecord([70, 800], 180, 40)

device.add_poly_record(rec_poly)
device.add_rect_record(rec_rect)

rec_poly = modmesh.OasisPolyRecord([
    [70, 980], [410, 980], [410, 1180], [70, 1180],
    [70, 1140], [370, 1140], [370, 1020], [70, 1020]])
rec_rect = modmesh.OasisRectRecord([70, 1060], 180, 40)

device.add_poly_record(rec_poly)
device.add_rect_record(rec_rect)

with open("test.oas", "wb") as f:
    f.write(bytearray(device.write()))
```

It will generate `test.oas` file. The file in KLayout will looks like:

image

Please review it. Thanks.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+void append_record_bytes(std::vector<uint8_t> bytes, T record);
+
+/**
+ * \class PolyRecord
+ * \brief Convert Rect information to OASIS polygon record bytes.
+ */
+class PolyRecord
+{
+private:
+    std::vector<std::pair<int, int>> vertexes;
```
yungyuc: All non-POD class member data need to prefix `m_`.

Put member data after member functions like other modmesh classes do.  Follow convention unless you have a wonderful reason not to.

Use "vertices" as the plural form of "vertex".  Do not use vertexes.

I dislike `std::vector` member data.  But it can be here for a while.  They should be changed to a pad soon.

cpp/modmesh/oasis/oasis_device.hpp
yungyuc: @c1ydehhx You forgot to fix it.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    OasisDevice(OasisDevice const &) = default;
+    OasisDevice(OasisDevice &&) = default;
+    OasisDevice & operator=(OasisDevice const &) = default;
+    OasisDevice & operator=(OasisDevice &&) = default;
+    ~OasisDevice() = default;
+
+    void add_poly_record(const PolyRecord & record);
+    void add_rect_record(const RectRecord & record);
+
+    std::vector<uint8_t> write();
```
yungyuc: If returning a byte container, rename `write` to `to_bytes`.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+static void append_magic_bytes(std::vector<uint8_t> & segment);
+
+template <typename T>
+void append_record_bytes(std::vector<uint8_t> bytes, T record);
+
+/**
+ * \class PolyRecord
+ * \brief Convert Rect information to OASIS polygon record bytes.
+ */
+class PolyRecord
```
yungyuc: Rename `PolyRecord` to `OasisRecordPoly`.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    ~PolyRecord() = default;
+
+    std::vector<uint8_t> to_bytes() const;
+};
+
+/**
+ * \class RectRecord
+ * \brief Convert Rect information to OASIS rectangle record bytes.
+ */
+class RectRecord
```
yungyuc: Rename `RectRecord` to `OasisRecordRect`.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <utility>
+#include <vector>
+
+namespace modmesh
+{
+
+namespace oasis
+{
+
+static void append_signed_integer(std::vector<uint8_t> & segment, int value);
```
yungyuc: After removing the namespace `modmesh::oasis`, move these free functions to `OasisDevice` as static functions..

tests/test_oasis_device.py
```diff
@@ -0,0 +1,123 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+import unittest
+import modmesh
+
+
+class OasisRectRecordTC(unittest.TestCase):
```
yungyuc: Good one.  Need to be renamed with the C++ class.

tests/test_oasis_device.py
```diff
@@ -0,0 +1,123 @@
...
+    def test_to_byte(self):
+        rec = modmesh.OasisRectRecord([70, 800], 180, 40)
+
+        expected_record_bytes = '\x14\x7B\x00\x00\xB4\x01\x28\x8C\x01\xC0\x0C'
+        record_bytes = rec.to_bytes()
+
+        self.assertEqual(record_bytes, list(map(ord, expected_record_bytes)))
+
+
+class OasisPolyRecordTC(unittest.TestCase):
```
yungyuc: Good one.  Need to be renamed with the C++ class.

tests/test_oasis_device.py
```diff
@@ -0,0 +1,123 @@
...
+        expected_record_bytes = '\x15\x3B\x00\x00\x00\x07' \
+                                '\xA8\x05\x90\x03\xA9\x05' \
+                                '\x51\xD8\x04\xF1\x01\xD9' \
+                                '\x04\x8C\x01\xA0\x0B'
+        record_bytes = rec.to_bytes()
+
+        self.assertEqual(record_bytes, list(map(ord, expected_record_bytes)))
+
+
+# For the format of OASIS, please refer OasisDevice comment in oasis_device.cpp
```
yungyuc: Make it clear with the namespace `modmesh::OasisDevice`.

yungyuc (changes requested):

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <modmesh/base.hpp>
+#include <pybind11/stl.h>
```
yungyuc: Do not import any Python-related (pybind11 included) headers outside a `pymod/` directory.

yungyuc (changes requested):
- [x] Do not include any Python-related (pybind11 included) headers outside a `pymod/` directory.

c1ydehhx:
Hi @yungyuc. Thanks for the review.

I already fix all issues in comment. I adjust the function name, class name, member name, and comment that mention the class name,. Moreover, I remove pybind header that include in wrong place and remove the `oasis` namespace.

For the comment that move free function to `OasisDevice` class, since all classes depends `append_signed_integer`, and `append_unsigned_integer` function, I can't move them into `OasisDevice` class. However, `append_magic_bytes`, `append_start_record_bytes`, `append_cell_and_cell_name_record_byte` and `append_end_record_byte` only used in `OasisDevice`, I move them to `OasisDevice` class.

Please review it. Thanks.

yungyuc:
> For the comment that move free function to `OasisDevice` class, since all classes depends `append_signed_integer`, and `append_unsigned_integer` function, I can't move them into `OasisDevice` class.

You can.  Move `OasisDevice` to be before the declaration of `OasisRecord*`, and forward declare `OasisRecord*` before `OasisDevice`.  See:

```cpp

// Forward declaration (but do not add comment in your PR because it's obvious.
OasisRecordPoly;
OasisRecordRect;

class OasisDevice
{

public:

  // all static functions for OasisRecord*.

private:

  std::vector<OasisRecordPoly> //...;
  std::vector<OasisRecordRect> //...;

};

class OasisRecordPoly
{
  // content.s
};

class OasisRecordRect
{
  // content.s
};
```

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    std::vector<uint8_t> to_bytes();
+
+private:
+    static void append_magic_bytes(std::vector<uint8_t> & segment);
+    static void append_start_record_bytes(std::vector<uint8_t> & segment);
+    static void append_cell_and_cell_name_record_byte(std::vector<uint8_t> & segment);
+    static void append_end_record_byte(std::vector<uint8_t> & segment);
+
+    template <typename T>
+    static void append_record_bytes(
```
yungyuc: It's too short to use two lines.  Use a single line for the signature.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    OasisRecordRect(OasisRecordRect const &) = default;
+    OasisRecordRect(OasisRecordRect &&) = default;
+    OasisRecordRect & operator=(OasisRecordRect const &) = default;
+    OasisRecordRect & operator=(OasisRecordRect &&) = default;
+    ~OasisRecordRect() = default;
+
+    std::vector<uint8_t> to_bytes() const;
+
+private:
+    std::pair<int, int> m_bottom_left;
```
yungyuc: Split the field `m_bottom_left` to be `m_lower` and `m_left`.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    int w;
```
yungyuc: `m_` prefix.  Please review carefully.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+    int h;
+};
```
yungyuc: Class end marker.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * \class PolyRecord
+ * \brief Convert Rect information to OASIS polygon record bytes.
+ */
+class OasisRecordPoly
+{
+
+public:
+    explicit OasisRecordPoly(std::vector<std::pair<int, int>> vertices)
+        : m_vertices(std::move(vertices)) {};
+    OasisRecordPoly(OasisRecordPoly const &) = default;
```
yungyuc, line 95: You forgot to delete default constructor.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,133 @@
...
+class OasisRecordRect
+{
+public:
+    OasisRecordRect(std::pair<int, int> bottom_left, int w, int h)
+        : m_bottom_left(bottom_left)
+        , w(w)
+        , h(h)
+    {
+    }
+    OasisRecordRect(OasisRecordRect const &) = default;
```
yungyuc, line 122: Ditto.

yungyuc (changes requested):
Points to address:

- [x] Split the field `OasisRecordRect::m_bottom_left` to be `m_lower` and `m_left`.
- [x] Add end marker to all classes.  You forgot some.
- [x] Missing rule of five.

c1ydehhx:
Hi @yungyuc. Thanks for the review.

I fix all issue mentioned in comment. I delete copy assignment on `OasisDevice` and record classes, add all end class marker and adjust signature to one-line, and split `m_bottom_left` to `m_lower` and `m_left`. Also, I check the member of non-POD class should have `m_` prefix again. Sorry for the misunderstand of your instruction.

Please review it. Thanks.

yungyuc (changes requested):
You have previous points unaddressed:

- [x] You forgot to move free functions.
- [x] Missing rule of five.

cpp/modmesh/oasis/oasis_device.hpp
yungyuc: You forgot to move the free functions.

c1ydehhx:
Hi @yungyuc. Thanks for the review.

I just move the free function into `OasisDevice` class, and call these function in `OasisRecord*` class with static way. It looks like:

```cpp
std::vector<uint8_t> OasisRecordRect::to_bytes(){
    // Implementation ...
    OasisDevice::append_unsigned_integer(segment, m_w);
    OasisDevice::append_unsigned_integer(segment, m_h);
    OasisDevice::append_signed_integer(segment, m_left);
    OasisDevice::append_signed_integer(segment, m_lower);
}
```

Also, I delete default constructor in `OasisRecord*`. I previous concern that delete default constructor will happen compilation error since the vector with `OasisRecord*` can not be construct. 

```cpp
    std::vector<OasisRecordPoly> m_polygon_records;
    std::vector<OasisRecordRect> m_rect_records;
```

After think a while, since I don't pre-allocate the space to let the vector using default constructor to create element, it still work properly.

I already fixed these issue. Thanks.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,139 @@
...
+    OasisDevice(OasisDevice const &) = default;
+    OasisDevice(OasisDevice &&) = default;
+    OasisDevice & operator=(OasisDevice const &) = default;
+    OasisDevice & operator=(OasisDevice &&) = default;
+    ~OasisDevice() = default;
+
+    void add_poly_record(const OasisRecordPoly & record);
+    void add_rect_record(const OasisRecordRect & record);
+
+    static void append_signed_integer(std::vector<uint8_t> & segment, int value);
```
yungyuc: I usually move the static functions to the beginning of the class because they do not depend on other code in it.

It also makes sense to move the static member functions to be before other member functions, because the latter (or the code in the latter) may call the former.

My order is usually: (1) aliases, constants, and enum, (2) static member functions, (3) member functions, (4) member data.  It's also the convention used in modmesh.

Please follow the convention.

yungyuc (changes requested):
> Also, I delete default constructor in `OasisRecord*`. I previous concern that delete default constructor will happen compilation error since the vector with `OasisRecord*` can not be construct.
> 
> ```c++
>     std::vector<OasisRecordPoly> m_polygon_records;
>     std::vector<OasisRecordRect> m_rect_records;
> ```
> 
> After think a while, since I don't pre-allocate the space to let the vector using default constructor to create element, it still work properly.
> 
> I already fixed these issue. Thanks.

If you want to keep the default constructor, make it explicit like:

```cpp
OasisRecordPoly() = default;
```

Otherwise it's deleted implicitly: https://stackoverflow.com/a/38995677.

The point is to make your intention explicit.  You did not seem to know that not supplying the default constructor actually removed it.  It will confuse you and maintainers in the future.

- [x] Review if you want to have the default constructor of `OasisRecord*` classes or not.  If yes, change the code.  If not, leave a comment and stay with the deletion.

Other points to address:

- [x] Move the static functions to the beginning of the class because they do not depend on other code in it.

cpp/modmesh/oasis/oasis_device.hpp
```diff
@@ -0,0 +1,139 @@
...
+    OasisRecordRect(OasisRecordRect const &) = default;
+    OasisRecordRect(OasisRecordRect &&) = default;
+    OasisRecordRect & operator=(OasisRecordRect const &) = default;
+    OasisRecordRect & operator=(OasisRecordRect &&) = default;
+    ~OasisRecordRect() = default;
+
+    std::vector<uint8_t> to_bytes() const;
+
+private:
+    int m_left;
```
yungyuc: You may use 4 lines for the 4 fields or 1 line for them.

c1ydehhx:
Hi @yungyuc. Thanks for the review.

I already fix the function order of class. If my realize correct, the member function may include constructor, so I put static function before constructor. I would like to check is it fit the convention. Thanks. 

For the private static function, I'm not sure the convention that also need to move to the beginning of the class. If so, it will look like this:

```cpp
class OasisDevice
{
public:
    static void append_signed_integer(std::vector<uint8_t> & segment, int value);
    static void append_unsigned_integer(std::vector<uint8_t> & segment, int value);

private:
    static void append_magic_bytes(std::vector<uint8_t> & segment);
    static void append_start_record_bytes(std::vector<uint8_t> & segment);
    static void append_cell_and_cell_name_record_byte(std::vector<uint8_t> & segment);
    static void append_end_record_byte(std::vector<uint8_t> & segment);

public:
    // Public member function.

private:
    // Private member function.
    // Private member.
} /* end class OasisDevice */
```

Please let me know that I just keep the current version or adjust to the version above.

Moreover, I condense rect member to 1 line. 

> Review if you want to have the default constructor of OasisRecord* classes or not. If yes, change the code. If not, leave a comment and stay with the deletion.

After think of the while, I think I'll keep default constructor deleted. Since the empty record of poly or rect is nonsense afaik, I'll keep the default constructor deleted. Thanks for the instruction about the behavior of constructor.

*These changes already co-work with Codex to check the point you address is met.*

yungyuc (approved):
LGTM

> Please let me know that I just keep the current version or adjust to the version above.

For private static functions, either works.

> After think of the while, I think I'll keep default constructor deleted. Since the empty record of poly or rect is nonsense afaik, I'll keep the default constructor deleted. Thanks for the instruction about the behavior of constructor.

Not necessarily.  We usually reserve in containers.  But maybe reserving can use placement new.  It would be interesting to check what the standard says about it.

yungyuc (changes requested):
Please update commit logs and the PR leading comment with a concise summary of what you do.

You do not have to squash although you may.  Request for the next review once it's done.

c1ydehhx:
Hi @yungyuc. I alreday update PR description and squash the commit and add properly message.

Please merge it after CI passed. Thanks.

yungyuc:
> Hi @yungyuc. I alreday update PR description and squash the commit and add properly message.
> 
> Please squash it after CI passed. Thanks.

You already squashed.  But the commit log does not have a clear summary.  This is what I see:

> Add OASIS device that store rect/poly and convert to OASIS format bytes

Please try again.

c1ydehhx:
Hi @yungyuc. I already rewrite the commit message that try to let it more clear. Please review it. Thanks.

yungyuc:
> Hi @yungyuc. I already rewrite the commit message that try to let it more clear. Please review it. Thanks.

You wrote a short commit log but it's not clear.  Try this:

> Add a helper class to return a byte sequence in the OASIS file format.
> 
> * A new class, `modmesh::OasisDevice`, is added.  The member function `to_bytes()` returns the byte sequence.  It currently supports two OASIS record types: polygon and rectangle.
>   * The returned byte sequence can be written to a file.
> * 2 new classes `modmesh::OasisRecordPoly` and `modmesh::OasisRecordRect` correspond to the supported OASIS record types.
> * The 3 new classes are wrapped to Python for testing.

Please supplement missing information and rewrite to make it more clear and concise.

cpp/modmesh/oasis/oasis_device.cpp
```diff
@@ -0,0 +1,315 @@
...
+{
+    // Magic byte should be %SEMI-OASIS<CR><NL>, which <CR><NL> is 0x0D 0x0A.
+    // Please refer OASIS Draft section 6.4.
+
+    std::string magic_byte_plain_text = "%SEMI-OASIS\x0D\x0A";
+
+    for (char c : magic_byte_plain_text)
+    {
+        segment.push_back(c);
+    }
```
tigercosmos: why not copy the whole buffer directly?

tigercosmos: same for all other similar operations.

c1ydehhx: I think it will be better if I copy whole buffer directly. Thanks for the suggestion.

yungyuc:
@c1ydehhx Are you ready for updating the commit log?

yungyuc:
@Han-Xuan ***@***.***>,

Do you have a plan for finishing the commit log?

Regards,
Yung-Yu

From: Yung-Yu Chen ***@***.***>
Date: Wednesday, February 18, 2026 at 14:31
To: solvcon/modmesh ***@***.***>
Cc: Yung-Yu Chen ***@***.***>, Your activity ***@***.***>
Subject: Re: [solvcon/modmesh] Implement OASIS file writer (PR #674)

[https://avatars.githubusercontent.com/u/399122?s=20&v=4]yungyuc left a comment (solvcon/modmesh#674)<https://github.com/solvcon/modmesh/pull/674#issuecomment-3918992777>

@c1ydehhx<https://github.com/c1ydehhx> Are you ready for updating the commit log?

cpp/modmesh/oasis/oasis_device.cpp
c1ydehhx: Hi @tigercosmos. I changes few byte append operation with for-loop with `std::vector` insert operation instead, which makes the code clear. For the other byte append operation with `std::vector::push_back` function, I would like to append bytes sepreately and with comments to clarify how the OASIS format is organized.

c1ydehhx:
Hi @yungyuc. Thanks for the review. I've been updated the commit message with short description summarizing what this PR does.

yungyuc:
> Hi @yungyuc. Thanks for the review. I've been updated the commit message with short description summarizing what this PR does.

Thank you, @c1ydehhx .  It's still unclear and please improve.

c1ydehhx:
Hi @yungyuc. Thanks for the review. I've updated the lead comment and commit message. Please review it. Thanks.

yungyuc (approved):
LGTM

Issue #510, opened 2025-04-18
Add a toggle to switch between light/dark theme


yungyuc:
It could be a feature Qt (or PySide6) provides. It seems that the color will adapt to the setting of the computer since my computer is in dark mode.

_Originally posted in https://github.com/solvcon/modmesh/issues/507#issuecomment-2814589780_

Sean20405:
I can take this issue!

yungyuc:
Thank you, @Sean20405 .  I think we should start with some study.  Could you please use 30-60 minutes to survey how easy or hard this feature is, and make an estimate for how much time it may take?

yungyuc:
@Sean20405 Any comment?

yungyuc:
Remove assignee for lack of activities.

JunliXiao:
Hi @yungyuc , I can help this one
I did a bit searches, it should be done in PySide6 code only, with libs like qt-matieral.

yungyuc:
> Hi [@yungyuc](https://github.com/yungyuc) , I can help this one

Thank you, @JunliXiao .  Assigning to you.

> I did a bit searches, it should be done in PySide6 code only, with libs like qt-matieral.

Sounds like a plan.

tigercosmos:
Can't wait to see the result @JunliXiao

JunliXiao:
> Can't wait to see the result @JunliXiao 

Hi @tigercosmos ,

Thanks for your support!
It will take some time for me to refine this feature, please wait 😀

JunliXiao:
Hi @yungyuc ,

Please take a look at this [https://youtu.be/1ZzNGnlexEY](https://youtu.be/1ZzNGnlexEY).

This time I build a simple toggle to change light/dark theme, while keeping the native look.
Before toggling, pilot always reacts to OS-level theme changing.
After toggling, which takes effect after clicking OK, it changes theme to the selected button only, ignoring OS-level changes.
As shown in the end, for accent color/強調顏色 pilot changes as usual, if set differently from the system.

Last thing, for those icon buttons at the top of the plot area, which are Matplotlib things, they too have light/dark styles. 
But since I haven't touched them, they don't change when theme is toggled.

yungyuc:
It looks wonderful!  I think there can be the third toggle of using the system theme, in addition to be explicit light and dark.

The matplotlib icons can be a future task.  @j8xixo12 you must see the cool movie.

tigercosmos:
done in the latest them system

[...]

PR #678, opened 2026-02-07
Demo of pilot's new look, first try

JunliXiao:
Ref #510 

I got a new look for pilot gui using the Python library [qt-material](https://pypi.org/project/qt-material/), as below:
light_red_theme

Compared to my macOS dark mode native look:
macos_default_theme

As you can see, the qt-material default (light-red) themes looks cool, but its font and buttons etc, not so much
And it lacks a simple white-ish, light mode theme, without light COLORS

Now my primary goal is to let users select light mode or dark mode
Either one of them is platform dependent (Windows/Mac), but you can choose freely
Then I can extend to provide more fancy-looking themes...

Next step I'm going to dive into Qt code and see how to customize

### DIFF
--- a/.github/actions/setup_linux/action.yml
+++ b/.github/actions/setup_linux/action.yml
@@ -89,16 +89,17 @@ runs:
         sudo ln -fs "$(which clang-tidy-22)" "/usr/local/bin/clang-tidy"
         echo "::endgroup::"
 
-    - name: Install and configure gcc-14
+    - name: Install and configure gcc-16
       shell: bash
       run: |
-        echo "::group::Install and configure gcc-14"
+        echo "::group::Install and configure gcc-16"
+        sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
         sudo apt-get -qqy update
-        sudo apt-get -qy install gcc-14 g++-14
-        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
-        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
-        sudo update-alternatives --install /usr/bin/cc cc /usr/bin/gcc-14 100
-        sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++-14 100
+        sudo apt-get -qy install gcc-16 g++-16
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-16 100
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -93,6 +93,8 @@ CMAKE_CMD = cmake $(MODMESH_ROOT) \
 	-DLINT_AS_ERRORS=ON \
 	-DMODMESH_PROFILE=$(MODMESH_PROFILE) \
 	-DQT3D_USE_RHI=$(QT3D_USE_RHI) \
+	-Dpybind11_DIR=`pybind11-config --cmakedir` \
+  -DCMAKE_TOOLCHAIN_FILE=${DEVENVPREFIX}/lib/cmake/Qt6/qt.toolchain.cmake \
```
c1ydehhx: Wrong indent.

```suggestion
	-Dpybind11_DIR=`pybind11-config --cmakedir` \
    -DCMAKE_TOOLCHAIN_FILE=${DEVENVPREFIX}/lib/cmake/Qt6/qt.toolchain.cmake \
```

.gitignore
```diff
@@ -21,3 +21,5 @@ setup.mk
 *.pdb
 *.env
 profiling/results/
+
+.DS_Store
```
c1ydehhx: This change is not related by this PR. Please revert it.

yungyuc (changes requested):
The PR is work in progress and I will change it to be so.

> Fix #xxx

Do not add issue operating commands in PR like above.  I'll edit your comment to remove it.

> I got a new look for pilot gui using the Python library [qt-material](https://pypi.org/project/qt-material/)

This is a good start.  Mac does not use material design.  Perhaps we need to find a configuration for the mac light theme.  Could you please help look into it?

----

Next time when requesting for a review, please (1) run CI in your fork to make sure it passes linters and (2) leave a global comment to request for the review explicitly (no need to press the button, which we do not pay attention to).

Additional points to address:

- [ ] Formatting and history issues.

JunliXiao:
> Perhaps we need to find a configuration for the mac light theme. Could you please help look into it?

Yes, I'm looking into it now.

Since I want the light mode to look like the same as platform native, or very close to, I try to locate the default style sheets(qss). Interestingly, Qt does not set the default look from stylesheets, but it at runtime directly creates and applies the platform-specific subclass of QStyle. (QStyle works at level lower than qss) For Windows and Linux there exist subclass files; for macOS thing are tricky: it goes to find some .dylib at some path and load it for use, which I still can't find that on my mac🤔

Other than that, I will try using qss, to be able to style custom colors.

Issue #677, opened 2026-02-07
Reduce duplication by sharing common steps between GitHub Actions


chestercheng:
The `.github/workflows/*.yml` workflows contain several repeated steps (e.g. dependency installation).
This makes maintenance harder and increases the risk of workflows drifting over time.

It would be better to extract the common parts into shared steps, and keep each workflow focused on its specific logic.

ExplorerRay:
I think this can be solved by [composite actions](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action).

terrychan999:
I reviewed the workflows and broke down the dependencies used in modmesh by level and usage.
This could help clarify what should be moved into the dependency installation composite action.

## OS level
- C++ compiler (clang/gcc/msvc)
- cmake
- python
- Qt

### for test
- gtest

### for lint
- clang-tidy
- clang-format
- silversearcher-ag (installed in lint.yml but not currently used)

## Python level
- setuptools
- pybind11[global]
- numpy
- PySide6 (should align with the same version of Qt)
- matplotlib

### for test
- pytest
- jsonschema

### for lint
- flake8

ExplorerRay:
I have done Linux and MacOS composite actions in https://github.com/solvcon/modmesh/pull/691.

I think there are some todos in this issue.
- [ ] Windows composite action for `devbuild.yml`
- [x] Linux composite action for `nouse_install.yml`
- [x] MacOS composite action for `nouse_install.yml`
- [x] Try removing suspiciously unnecessary packages
  - For example, `silversearcher-ag` mentioned [here](https://github.com/solvcon/modmesh/issues/677#issuecomment-4015981000)

yungyuc:
@ExplorerRay @chestercheng I believe this issue is done.  Let me know if you do not think so.

ExplorerRay:
If Windows part can be ignored, I think this issue can be closed.

yungyuc:
Let's leave Windows as future work.

PR #680, opened 2026-02-20
ci: use reuse workflow for send email on fail

ExplorerRay:
To remove duplicate steps in CI which is mentioned in https://github.com/solvcon/modmesh/issues/677, I use [reuse workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) provided by Github to combine sending email on failure ones.

With this PR, we only need to maintain one workflow about email sending and other workflows only need to reuse it.

I have tested this on my local fork and ensure it works ([ref](https://github.com/ExplorerRay/modmesh/actions/runs/22229357166)). And the email body looks like below: 
image

### DIFF
--- a/.github/workflows/send_email_on_fail.yml
+++ b/.github/workflows/send_email_on_fail.yml
@@ -1,0 +1,43 @@
+name: Send Email on Failure
+
+on:
+  workflow_call:
+    inputs:
+      job_results:
+        description: 'String of job results'
+        required: true
+        type: string
+    secrets:
+      EMAIL_USERNAME:
+        required: true
+      EMAIL_PASSWORD:
+        required: true
+
+jobs:
+  send_email_on_failure:
+    name: send_emails_when_fails
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get current date and set to env
+        run: |
+          echo "DATE=$(date +%m/%d)" >> $GITHUB_ENV
+
+      - name: Send mail to mailing list
+        uses: dawidd6/action-send-mail@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/send_email_on_fail.yml
yungyuc, line 1: I wonder if the action should use a version.

yungyuc (approved):
This looks good.  I like the modularization.

@chestercheng, please take a look.

yungyuc:
Merged.  Discussion can continue.

.github/workflows/send_email_on_fail.yml
ExplorerRay: Based on the [reference](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow), it states:

> - {owner}/{repo}/.github/workflows/{filename}@{ref} for reusable workflows in public and private repositories.
> - ./.github/workflows/{filename} for reusable workflows in the same repository.

Since we are using a workflow within the same repository, we do not need to specify `@{ref}` (such as a specific commit SHA, release tag, or branch name).

Regarding the reusable workflow YAML file itself, the official documentation doesn't mention specifying a version inside the file. I think it simply follows the repository’s version control.

yungyuc: Thanks for the clarification.  In our case, which uses the action within the same repository, it makes more sense to follow the latest checkin (repository version control).

We _should not_ use a version.  I like this conclusion.

PR #691, opened 2026-03-12
ci: composite actions for setting up dependencies for macOS and Linux runners

### LINKED ISSUE #677: Reduce duplication by sharing common steps between GitHub Actions

ExplorerRay:
I just add `setup_macOS` and `setup_Linux` composite actions for dependency and use them in `devbuild` and `lint` workflows to solve part of https://github.com/solvcon/modmesh/issues/677.

These two composite actions have one input called `workflow` to decide whether some packages should be installed or not.
For example, `clang-tidy` is only needed when doing lint.

### DIFF
--- a/.github/actions/setup_Linux/action.yml
+++ b/.github/actions/setup_Linux/action.yml
@@ -1,0 +1,133 @@
+name: Setup modmesh CI dependencies on Linux
+description: Sets up the necessary dependencies for running CI tests on Github action Linux runners
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 1
+
+    - name: Add 8G swap
+      # Prevent hitting runner's resource limits when running clang-tidy
+      shell: bash
+      run: |
+        # Remove /swapfile first to avoid "fallocate: Text file busy" error
+        sudo swapoff -a
+        sudo rm -f /swapfile
+        sudo fallocate -l 8G /swapfile
+        sudo chmod 600 /swapfile
+        sudo mkswap /swapfile
+        sudo swapon /swapfile
+        free -h
+
+    - name: Cache Qt download
+      uses: actions/cache@v3
+      with:
+        path: ${{ github.workspace }}/Qt
+        key: ${{ runner.os }}-qt-6.8.1-linux_gcc_64
+        # fallback to any previous Qt cache on this OS
+        restore-keys: |
+          ${{ runner.os }}-qt-6.8.1-
+
+    - name: dependency by apt
+      shell: bash
+      run: |
+        # To update the cmake version > 4, install the latest cmake by kitware apt repository.
+        # Reference: https://apt.kitware.com/
+        sudo apt-get -y update
+        sudo apt-get -qy install ca-certificates gpg wget
+        test -f /usr/share/doc/kitware-archive-keyring/copyright || \
+            wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | \
+            gpg --dearmor - | \
+            sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
+        echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ noble main' | \
+            sudo tee /etc/apt/sources.list.d/kitware.list >/dev/null
+        sudo apt-get -qqy update
+        test -f /usr/share/doc/kitware-archive-keyring/copyright || \
+            sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
+        sudo apt-get -qy install kitware-archive-keyring
+        # The path of apt-get cmake is `/usr/bin/cmake`,
+        # but the path of built-in cmake of runner image is `/usr/local/bin/cmake`.
+        # We remove the built-in cmake.
+        sudo rm -rf /usr/local/bin/cmake
+        # Install clang-tidy-21 from LLVM repository for full C++23 support
+        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+        echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-21 main" | sudo tee /etc/apt/sources.list.d/llvm.list
+        sudo apt-get -qqy update
+        sudo apt-get -qy install \
+            sudo curl git build-essential make cmake libc6-dev gcc g++ silversearcher-ag \
+            clang-tidy-21 \
+            python3 python3-dev python3-venv
+        sudo ln -fs "$(which clang-tidy-21)" "/usr/local/bin/clang-tidy"
+        # Install qt6 only with ubuntu-24.04
+        # This page explains why we need libgl1-mesa-dev
+        # https://doc-snapshots.qt.io/qt6-dev/linux.html
+        #
+        # In short, OpenGL libraries and headers are required. Without
+        # installing this package, cmake won't find the correct lib path.
+        # This has been replaced by the 'install qt' section below to manage
+        # qt6 versioning independently from the OS.
+
+        # if [ "${{ matrix.os }}" == "ubuntu-24.04" ] ; then \
+        #   sudo apt-get -qy install \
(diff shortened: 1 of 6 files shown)


.github/workflows/devbuild.yml
```diff
@@ -334,202 +188,6 @@ jobs:
...
-      - name: make pytest BUILD_QT=ON USE_PYTEST_HELPER_BINDING=ON
-        run: |
-          # PySide6 installed by pip will bundle with a prebuilt Qt,
-          # this will cause duplicated symbol.
-          # Solve this issue by removed PySide6 prebuilt Qt library
-          PYSIDE6_PATH=$(python3 -c "import sys, os, PySide6; sys.stdout.write(os.path.dirname(PySide6.__file__))")
-          echo "pyside6 path: ${PYSIDE6_PATH}"
-          rm -rf ${PYSIDE6_PATH}/Qt/lib/*.framework
-          # maunally add homebrew's Qt rpath to PySide6
-          install_name_tool -add_rpath $(qtpaths --install-prefix)/lib ${PYSIDE6_PATH}/QtWidgets.abi3.so
```
ExplorerRay, line 498: I want to mention that I remove some lines like these (L498 - 500), I originally want to fix it if the CI fails.
However, it works without any error, so I think it's okay to remove these things.

.github/actions/setup_Linux/action.yml
```diff
@@ -0,0 +1,151 @@
+name: Setup modmesh CI dependencies on Linux
```
yungyuc: "Setup" is not a verb (at least was not one).  Use "Set up" like what you wrote in the next line.

.github/actions/setup_Linux/action.yml
```diff
@@ -0,0 +1,151 @@
...
+description: Sets up the necessary dependencies for running CI tests on Github action Linux runners
```
yungyuc: Use correct tense.  It's "Set up", not "Sets up".

.github/workflows/devbuild.yml
yungyuc: Please keep the lines.  They are hard-learnt know-how.  It will be hard to recover the knowledge 5 years after we drop them from the file.

But you can keep them commented out in the file if not wanting to have them active.  If you can find the PR originally added the lines, it will be great to add it to the comment.

yungyuc (changes requested):
- [x] Use all-lower-case for directory names: `setup_linux` and `setup_macos`.  It is ambiguous to mix lower and upper cases.
- [x] Keep the lines for mac `install_name_tool`.  They are hard-learnt know-how.

yungyuc:
@terrychan999 @chestercheng please help review the changes.  It looks good to me but it's too long for me to be sure that everything is right.  More eye balls help.

ExplorerRay:
> * [ ]  Use all-lower-case for directory names: `setup_linux` and `setup_macos`.  It is ambiguous to mix lower and upper cases.

About the directory name, I use these cases because Linux and macOS are defined in `runner.os` ([ref](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context)).

With this, we can further improve the following code into another one refers to [this discussion](https://github.com/orgs/community/discussions/25246).

```
- name: setup dependencies for Linux
    if: runner.os == 'Linux'
    uses: ./.github/actions/setup_Linux
    with:
      workflow: 'lint'

  - name: setup dependencies for macOS
    if: runner.os == 'macOS'
    uses: ./.github/actions/setup_macOS
    with:
      workflow: 'lint'
```

```
- name: setup dependencies
    uses: jenseng/dynamic-uses@v1
    with:
      uses: ./.github/actions/setup_${{ runner.os }}
      with: 
        workflow: 'lint'
```

yungyuc:
> > * [ ]   Use all-lower-case for directory names: `setup_linux` and `setup_macos`.  It is ambiguous to mix lower and upper cases.
> 
> About the directory name, I use these cases because Linux and macOS are defined in `runner.os` ([ref](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context)).
> 
> With this, we can further improve the following code into another one refers to [this discussion](https://github.com/orgs/community/discussions/25246).
> 
> ```
> - name: setup dependencies for Linux
>     if: runner.os == 'Linux'
>     uses: ./.github/actions/setup_Linux
>     with:
>       workflow: 'lint'
> 
>   - name: setup dependencies for macOS
>     if: runner.os == 'macOS'
>     uses: ./.github/actions/setup_macOS
>     with:
>       workflow: 'lint'
> ```
> 
> ```
> - name: setup dependencies
>     uses: jenseng/dynamic-uses@v1
>     with:
>       uses: ./.github/actions/setup_${{ runner.os }}
>       with: 
>         workflow: 'lint'
> ```

Directory names of mixed cases are not an improvement.  The directory names do not need to be identical to variable contents.

Let's use all-lower-case directory.

ExplorerRay:
> * [ ]  Use all-lower-case for directory names: `setup_linux` and `setup_macos`.  It is ambiguous to mix lower and upper cases.
> * [ ]  Keep the lines for mac `install_name_tool`.  They are hard-learnt know-how.

I just fixed them. Please review again. Thanks.

yungyuc (approved):
LGTM.  @terrychan999 @chestercheng please help review.

.github/actions/setup_linux/action.yml
```diff
@@ -0,0 +1,151 @@
...
+
+inputs:
+  workflow:
+    description: The name of the workflow that will use this action
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v6
```
chestercheng: The workflow already checkout the repo before. This step may be redundant.

.github/actions/setup_macos/action.yml
```diff
@@ -0,0 +1,90 @@
...
+
+inputs:
+  workflow:
+    description: The name of the workflow that will use this action
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v6
```
chestercheng: Same as above

chestercheng (changes requested):
@ExplorerRay thanks for the refactor.
Moving the environment setup into composite actions helps reduce duplication and makes the workflows easier to maintain.

.github/actions/setup_macos/action.yml
ExplorerRay: I just fixed them and CI worked well.

yungyuc:
@chestercheng make a global comment again if you are good with the change.

chestercheng (approved):
LGTM

terrychan999:
I noticed that `check ascii` and `check include` were removed in this change.
It seems they may have been left out unintentionally, so it might be worth adding them back.
Aside from that, everything looks great — nice work!

.github/workflows/lint.yml
```diff
@@ -237,157 +142,15 @@ jobs:
...
-          echo "flake8 version: $(flake8 --version)"
-
-      - name: ccache
-        uses: hendrikmuhs/ccache-action@v1.2
-        with:
-          key: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
-          restore-keys: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
-          create-symlink: true
-
-      - name: make cinclude (check_include)
```
terrychan999, line 342: `check include` was removed here.

.github/workflows/lint.yml
```diff
@@ -237,157 +142,15 @@ jobs:
...
-        run: make cinclude
-
-      - name: make checkascii (check ASCII-only characters)
```
terrychan999, line 345: `check ascii` was removed here

ExplorerRay: I think they are still there in L106 and L109 which Github requires you to click expand to show all the content. 
I originally think it was missing by checking the diff provided by Github. Thanks for the feedback.

terrychan999: You're right — I misread the diff. Thanks for pointing that out.

terrychan999 (approved):
LGTM

PR #695, opened 2026-03-16
ci: composite action for nouse_install workflow support

### LINKED ISSUE #677: Reduce duplication by sharing common steps between GitHub Actions

ExplorerRay:
I expand composite action support (https://github.com/solvcon/modmesh/issues/677) for `nouse_install`. And some packages like Qt is not necessary for this workflow. 
Therefore, I use some conditions check to avoid those steps being run in `nouse_install`.

### DIFF
--- a/.github/workflows/nouse_install.yml
+++ b/.github/workflows/nouse_install.yml
@@ -8,135 +8,7 @@ on:
 
 jobs:
 
-  nouse_install_ubuntu:
-
-    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
-
-    name: nouse_install_${{ matrix.os }}_Release
-
-    runs-on: ${{ matrix.os }}
-
-    env:
-      JOB_CMAKE_ARGS: -DBUILD_QT=OFF -DUSE_CLANG_TIDY=OFF -DCMAKE_BUILD_TYPE=Release
-
-    strategy:
-        matrix:
-          os: [ubuntu-24.04]
-
-        fail-fast: false
-
-    steps:
-
-    - uses: actions/checkout@v6
-      with:
-        fetch-depth: 1
-
-    - name: event name
-      run: |
-        echo "github.event_name: ${{ github.event_name }}"
-
-    - name: Add 8G swap (Ubuntu)
-      # Prevent hitting runner's resource limits
-      if: runner.os == 'Linux'
-      run: |
-        # Remove /swapfile first to avoid "fallocate: Text file busy" error
-        sudo swapoff -a
-        sudo rm -f /swapfile
-        sudo fallocate -l 8G /swapfile
-        sudo chmod 600 /swapfile
-        sudo mkswap /swapfile
-        sudo swapon /swapfile
-        free -h
-
-    - name: dependency by apt
-      run: |
-        # To update the cmake version > 4, install the latest cmake by kitware apt repository.
-        # Reference: https://apt.kitware.com/
-        sudo apt-get -y update
-        sudo apt-get -qy install ca-certificates gpg wget
-        test -f /usr/share/doc/kitware-archive-keyring/copyright || \
-            wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | \
-            gpg --dearmor - | \
-            sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
-        echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ noble main' | \
-            sudo tee /etc/apt/sources.list.d/kitware.list >/dev/null
-        sudo apt-get -qqy update
-        test -f /usr/share/doc/kitware-archive-keyring/copyright || \
-            sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
-        sudo apt-get -qy install kitware-archive-keyring
-        sudo apt-get -qy install \
-            sudo curl git build-essential make cmake libc6-dev gcc g++ \
-            python3 python3-dev python3-venv
-        # The path of apt-get cmake is `/usr/bin/cmake`,
-        # but the path of built-in cmake of runner image is `/usr/local/bin/cmake`.
-        # We remove the built-in cmake.
-        sudo rm -rf /usr/local/bin/cmake
-
-    - name: install and configure gcc-14
-      run: |
-        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-        sudo apt-get -qqy update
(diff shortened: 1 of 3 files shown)


.github/workflows/nouse_install.yml
```diff
@@ -164,50 +36,21 @@ jobs:
...
+        if: runner.os == 'MacOS'
+        uses: ./.github/actions/setup_macos
+        with:
+          workflow: 'nouse_install'
 
       - name: setup.py install build_ext
         run: |
-          # Using legacy setup.py install for build_ext with cmake args
-          python3 setup.py install build_ext \
+          sudo python3 setup.py install build_ext \
```
yungyuc: Why sudo for `python3 setup.py install`?  It seems to run sudo on both mac and linux while the former does not need sudo.

.github/workflows/nouse_install.yml
```diff
@@ -8,135 +8,7 @@ on:
 
 jobs:
 
-  nouse_install_ubuntu:
```
yungyuc, line 11: Deletion of the long block of code looks pretty!

yungyuc (changes requested):
Point to discuss and/or address:

- [x] Why sudo for `python3 setup.py install`?

yungyuc:
@terrychan999 @chestercheng please help review.

.github/workflows/nouse_install.yml
chestercheng: Same question as @yungyuc. I'm not comfortable introducing `sudo` here unless there's a strong reason.

Could we do this at the user level instead?

ExplorerRay: After trying, I found this `sudo` is due to Ubuntu uses system-wide python and tries to install modmesh package under `/usr/local/lib/python3.12/dist-packages/` which requires root.
While MacOS uses homebrew python, so it's okay to install modmesh package without root.

How about using one if condition to check `runner.os` and run it with `sudo` for Linux, without `sudo` for MacOS?
I think this is much simpler. If we use `venv` for Linux, there might be many Python-related dependencies need to be changed.

ExplorerRay: I already revised the code into sudo for Linux and no sudo for MacOS.

yungyuc: LGTM

.github/workflows/nouse_install.yml
```diff
@@ -164,50 +36,25 @@ jobs:
...
+        uses: ./.github/actions/setup_macos
+        with:
+          workflow: 'nouse_install'
 
       - name: setup.py install build_ext
         run: |
-          # Using legacy setup.py install for build_ext with cmake args
-          python3 setup.py install build_ext \
+          SUDO_CMD=""
+          if [ ${{ matrix.os }} == "ubuntu-24.04" ] ; then
```
yungyuc, line 54: Limiting sudo to ubuntu which needs it looks good.  The implementation speaks for itself.

yungyuc (approved):
LGTM

yungyuc:
[Run clang-tidy and flake8 on macos-15-intel](https://github.com/solvcon/modmesh/actions/runs/23272983566/job/67671230372#logs) failed after merging.  I am not sure about the correlation.  @ExplorerRay please help take a look.

```
Run export HOMEBREW_NO_AUTO_UPDATE=1
  export HOMEBREW_NO_AUTO_UPDATE=1
  export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
  # This has been replaced by the 'install qt' section below to manage
  # qt6 versioning independently from the OS.
  # brew install llvm@21 qt6
  brew install llvm@21
  ln -s "$(brew --prefix llvm@21)/bin/clang-format" "/usr/local/bin/clang-format"
  ln -s "$(brew --prefix llvm@21)/bin/clang-tidy" "/usr/local/bin/clang-tidy"
  shell: /bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    JOB_MAKE_ARGS: VERBOSE=1 BUILD_QT=ON USE_CLANG_TIDY=ON LINT_AS_ERRORS=ON
    QT_DEBUG_PLUGINS: 1
    QT_QPA_PLATFORM: offscreen
    PIP_BREAK_SYSTEM_PACKAGES: 1
    AQT_CONFIG: thirdparty/aqt_settings.ini
Warning: `$HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` is set: not checking for outdated
dependents or dependents with broken linkage!

==> Fetching downloads for: llvm@21
✔︎ Bottle Manifest llvm@21 (21.1.8)
✔︎ Bottle llvm@21 (21.1.8)
==> Pouring llvm@21--21.1.8.sonoma.bottle.tar.gz
Warning: llvm@21 was installed but not linked because llvm@18 is already linked.
To link this version, run:
  brew link llvm@21

==> Caveats
CLANG_CONFIG_FILE_SYSTEM_DIR: /usr/local/etc/clang
CLANG_CONFIG_FILE_USER_DIR:   ~/.config/clang

LLD is now provided in a separate formula:
  brew install lld@21

Using `clang`, `clang++`, etc., requires a CLT installation at `/Library/Developer/CommandLineTools`.
If you don't want to install the CLT, you can write appropriate configuration files pointing to your
SDK at ~/.config/clang.

To use the bundled libunwind please use the following LDFLAGS:
  LDFLAGS="-L/usr/local/opt/llvm@21/lib/unwind -lunwind"

To use the bundled libc++ please use the following LDFLAGS:
  LDFLAGS="-L/usr/local/opt/llvm@21/lib/c++ -L/usr/local/opt/llvm@21/lib/unwind -lunwind"
Features newer than system libc++ will require the following define to enable:
  CPPFLAGS="-D_LIBCPP_DISABLE_AVAILABILITY"

NOTE: You probably want to use the libunwind and libc++ provided by macOS unless you know what you're doing.

llvm@21 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have llvm@21 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/llvm@21/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find llvm@21 you may need to set:
  export LDFLAGS="-L/usr/local/opt/llvm@21/lib"
  export CPPFLAGS="-I/usr/local/opt/llvm@21/include"

For cmake to find llvm@21 you may need to set:
  export CMAKE_PREFIX_PATH="/usr/local/opt/llvm@21"
==> Summary
🍺  /usr/local/Cellar/llvm@21/21.1.8: 8,670 files, 1.6GB
ln: /usr/local/bin/clang-format: File exists
```

PR #709, opened 2026-04-04
Fix console output area to scroll independently

### LINKED ISSUE #700: Python console does not automatically scroll to the bottom
When the text box contains too many lines, the Python console does not automatically scroll to the last line.

Example: The following commands are entered in sequence.

```python
print("a")
print("b")
print("c")
print("d")
print("e")
print("f")
print("g")
```

The input box disappears after `print("e")`:

Image

The following commands are entered in sequence, and the input box disappears.

To solve this issue, only the upper text box for the execution results should allow to scroll.

tigercosmos:
## Summary
- Replace `QScrollArea` wrapper with a `QVBoxLayout` so the history output area has a fixed height within the dock widget and scrolls its own content internally
- Pin the command input at the bottom with `stretch=0`, while the history area fills remaining space with `stretch=1`
- Enable vertical scrollbar on the history edit (`ScrollBarAsNeeded`) and auto-scroll to bottom on new output

## Before & After

**Before**

https://github.com/user-attachments/assets/2784713b-7454-49f6-b530-837f7f96928f

**After**

https://github.com/user-attachments/assets/18048168-0ab3-4836-8329-9126ae54b56c

### DIFF
--- a/cpp/modmesh/pilot/RPythonConsoleDockWidget.cpp
+++ b/cpp/modmesh/pilot/RPythonConsoleDockWidget.cpp
@@ -30,7 +30,6 @@
 #include <QVBoxLayout>
 #include <QKeyEvent>
 #include <QScrollBar>
-#include <QTimer>
 
 namespace modmesh
 {
@@ -80,17 +79,16 @@ void RPythonCommandTextEdit::keyPressEvent(QKeyEvent * event)
 
 RPythonConsoleDockWidget::RPythonConsoleDockWidget(const QString & title, QWidget * parent, Qt::WindowFlags flags)
     : QDockWidget(title, parent, flags)
-    , m_scroll_area(new QScrollArea)
-    , m_container(new QWidget)
     , m_history_edit(new RPythonHistoryTextEdit)
     , m_command_edit(new RPythonCommandTextEdit)
     , m_python_redirect(Toggle::instance().fixed().get_python_redirect())
 {
-    m_scroll_area->setWidgetResizable(true);
-    setWidget(m_scroll_area);
-
-    m_container->setLayout(new QVBoxLayout);
-    m_scroll_area->setWidget(m_container);
+    auto * container = new QWidget;
+    auto * layout = new QVBoxLayout;
+    layout->setContentsMargins(0, 0, 0, 0);
+    layout->setSpacing(0);
+    container->setLayout(layout);
+    setWidget(container);
 
     QPalette palette = QPalette();
     palette.setColor(QPalette::Base, Qt::white);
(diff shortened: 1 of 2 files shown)


yungyuc:
Is this work in progress (for issue #700)?  The output area still scrolled wrongly:
image

tigercosmos:
@yungyuc interesting... it works fine on my side, though
image

yungyuc:
Your snapshot does not look fine.  The text area does not scroll to the bottom like a normal console.  And you ran only one command.

tigercosmos:
@yungyuc Let me record the screen, it will show more clearly.

yungyuc:
An online meeting may work more efficiently for GUI discussion.  But we should wait until the PR takes shape.

tigercosmos:
@yungyuc  Could you take a look at the PR description? I uploaded the screen records.

tigercosmos:
Also, I just checked #700, and I think it's the same bug.

yungyuc:
You are right.  It's fixed.  I probably made a mistake building your branch.

image

cpp/modmesh/pilot/RPythonConsoleDockWidget.cpp
```diff
@@ -80,29 +79,28 @@ void RPythonCommandTextEdit::keyPressEvent(QKeyEvent * event)
...
     palette.setColor(QPalette::PlaceholderText, Qt::darkGray);
 
     m_history_edit->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
-    m_history_edit->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    m_history_edit->setVerticalScrollBarPolicy(Qt::ScrollBarAsNeeded);
     m_history_edit->setFont(QFont("Courier New"));
     m_history_edit->setReadOnly(true);
     m_history_edit->setPalette(palette);
-    m_container->layout()->addWidget(m_history_edit);
+    layout->addWidget(m_history_edit, /*stretch=*/1);
```
yungyuc, line 103: I prefer to have a space between the argument comment and the argument, like, `/*stretch*/ 1`, but it is ok.  Even having mixed style in the code could be ok.

I am not sure if clang-format can enforce a style like this.  It'd be good if someone can take a look.

cpp/modmesh/pilot/RPythonConsoleDockWidget.cpp
```diff
@@ -111,16 +109,17 @@ RPythonConsoleDockWidget::RPythonConsoleDockWidget(const QString & title, QWidge
...
 
     connect(
         m_history_edit->document(),
         &QTextDocument::contentsChanged,
         this,
         [this]()
         {
-            const int newHeight = calcHeightToFitContents(m_history_edit);
-            m_history_edit->setMinimumHeight(newHeight);
+            // Keep the latest output visible to the user
```
yungyuc, line 120: Great.

yungyuc (approved):
LGTM.

There is a minor style issue that is not a reason to block merging.  I'll merge later tonight if I no one stops me.

@Tucchhaa feel free to comment.  This is for issue #700.

cpp/modmesh/pilot/RPythonConsoleDockWidget.cpp
tigercosmos: Formatter doesn't allow me to put a space. I think we can let it as it is.

PR #711, opened 2026-04-05
Add equality and inequality operators for Bezier3d class 

### LINKED ISSUE #568: Develop equality operator for `Bezier3d` class template and its Python wrapper
Because of the lack of the equality (and inequality) operator for `Bezier3d` class template, the following fails:

```python
from modmesh import Point3dFp64, Bezier3dFp64

# Create one 3rd-order Bezier curve object
b1 = Bezier3dFp64(Point3dFp64(607.7, -64.2, 0), Point3dFp64(504.8, -64.2, 0),
                  Point3dFp64(421.4, -147.6, 0), Point3dFp64(421.4, -250.5, 0))
# Exactly the same as b1
b2 = Bezier3dFp64(Point3dFp64(607.7, -64.2, 0), Point3dFp64(504.8, -64.2, 0),
                  Point3dFp64(421.4, -147.6, 0), Point3dFp64(421.4, -250.5, 0))
# This fails because of lack of equality comparison operator.
assert b1 == b2
```

Need to develop:
- [x] `Bezier3d::operator==` and `Bezier3d::operator!=`
- [x] Wrapper of the equality and inequality operator to Python
- [x] Associated unit tests

YenPeiChen07:
For #568. This PR adds equality (`==`) and inequality (`!=`) operators for `Bezier3d` class. It also includes Python bindings for these two operators and add corresponding unit tests for them.

### DIFF
--- a/cpp/modmesh/universe/bezier.hpp
+++ b/cpp/modmesh/universe/bezier.hpp
@@ -90,6 +90,38 @@ public:
     Bezier3d & operator=(Bezier3d &&) = default;
     ~Bezier3d() = default;
 
+    bool operator==(Bezier3d const & other) const
+    {
+        return m_data.v[0] == other.m_data.v[0] &&
+               m_data.v[1] == other.m_data.v[1] &&
+               m_data.v[2] == other.m_data.v[2] &&
+               m_data.v[3] == other.m_data.v[3] &&
+               m_data.v[4] == other.m_data.v[4] &&
+               m_data.v[5] == other.m_data.v[5] &&
+               m_data.v[6] == other.m_data.v[6] &&
+               m_data.v[7] == other.m_data.v[7] &&
+               m_data.v[8] == other.m_data.v[8] &&
+               m_data.v[9] == other.m_data.v[9] &&
+               m_data.v[10] == other.m_data.v[10] &&
+               m_data.v[11] == other.m_data.v[11];
+    }
+
+    bool operator!=(Bezier3d const & other) const
+    {
+        return m_data.v[0] != other.m_data.v[0] ||
+               m_data.v[1] != other.m_data.v[1] ||
+               m_data.v[2] != other.m_data.v[2] ||
+               m_data.v[3] != other.m_data.v[3] ||
+               m_data.v[4] != other.m_data.v[4] ||
+               m_data.v[5] != other.m_data.v[5] ||
+               m_data.v[6] != other.m_data.v[6] ||
+               m_data.v[7] != other.m_data.v[7] ||
+               m_data.v[8] != other.m_data.v[8] ||
+               m_data.v[9] != other.m_data.v[9] ||
+               m_data.v[10] != other.m_data.v[10] ||
+               m_data.v[11] != other.m_data.v[11];
+    }
+
 #define DECL_VALUE_ACCESSOR(C, I)                     \
     value_type C##I() const { return m_data.f.C##I; } \
     value_type & C##I() { return m_data.f.C##I; }
--- a/cpp/modmesh/universe/pymod/wrap_shape1d.cpp
+++ b/cpp/modmesh/universe/pymod/wrap_shape1d.cpp
@@ -528,6 +528,7 @@ protected:
     WrapBezier3d(pybind11::module & mod, char const * pyname, char const * pydoc);
 
     WrapBezier3d & wrap_management();
+    WrapBezier3d & wrap_operator();
     WrapBezier3d & wrap_accessor();
     WrapBezier3d & wrap_geometry();
 }; /* end class WrapBezier3d */
@@ -540,6 +541,7 @@ WrapBezier3d<T>::WrapBezier3d(pybind11::module & mod, const char * pyname, const
 
     (*this)
         .wrap_management()
+        .wrap_operator()
         .wrap_accessor()
         .wrap_geometry()
         //
@@ -575,6 +577,20 @@ WrapBezier3d<T> & WrapBezier3d<T>::wrap_management()
                                    self.p3().value_string());
             })
         .def_alias("__repr__", "__str__")
+        //
+        ;
+
+    return *this;
+}
+
+template <typename T>
+WrapBezier3d<T> & WrapBezier3d<T>::wrap_operator()
+{
+    namespace py = pybind11;
+
+    (*this)
+        .def(py::self == py::self) // NOLINT(misc-redundant-expression)
+        .def(py::self != py::self) // NOLINT(misc-redundant-expression)
         //
         ;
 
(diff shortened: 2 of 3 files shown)


cpp/modmesh/universe/pymod/wrap_shape1d.cpp
```diff
@@ -581,6 +583,20 @@ WrapBezier3d<T> & WrapBezier3d<T>::wrap_management()
...
+template <typename T>
+WrapBezier3d<T> & WrapBezier3d<T>::wrap_operator()
+{
+    namespace py = pybind11;
+
+    (*this)
+        .def(py::self == py::self) // NOLINT(misc-redundant-expression)
+        .def(py::self != py::self) // NOLINT(misc-redundant-expression)
+        //
+        ;
```
tigercosmos, line 595: it's interesting that these line are repeating many places. Maybe there is a good way to handle this.

cpp/modmesh/universe/bezier.hpp
```diff
@@ -90,6 +90,38 @@ class Bezier3d
     Bezier3d & operator=(Bezier3d &&) = default;
     ~Bezier3d() = default;
 
+    bool operator==(Bezier3d const & other) const
+    {
+        return m_data.v[0] == other.m_data.v[0] &&
```
tigercosmos, line 95: I wonder if we can just do memory comparison here?

cpp/modmesh/universe/bezier.hpp
```diff
@@ -90,6 +90,38 @@ class Bezier3d
     Bezier3d & operator=(Bezier3d &&) = default;
     ~Bezier3d() = default;
 
+    bool operator==(Bezier3d const & other) const
```
YenPeiChen07, line 93: Add equality operator for `Bezier3d` by checking all control points between two curves are equal.

cpp/modmesh/universe/bezier.hpp
```diff
@@ -90,6 +90,38 @@ class Bezier3d
...
+               m_data.v[5] == other.m_data.v[5] &&
+               m_data.v[6] == other.m_data.v[6] &&
+               m_data.v[7] == other.m_data.v[7] &&
+               m_data.v[8] == other.m_data.v[8] &&
+               m_data.v[9] == other.m_data.v[9] &&
+               m_data.v[10] == other.m_data.v[10] &&
+               m_data.v[11] == other.m_data.v[11];
+    }
+
+    bool operator!=(Bezier3d const & other) const
```
YenPeiChen07, line 109: Add inequality operator for `Bezier3d`. It is not equal when one of control point coordinates is not matched.

tests/test_universe_shape1d.py
```diff
@@ -263,12 +272,7 @@ def test_repr_str(self):
         self.assertEqual(str(b), golden)
         # Evaluate the string and test the result
         e = eval(golden, vars(modmesh))
-        # FIXME: Bezier3d does not have equality operator
-        # Tracked in https://github.com/solvcon/modmesh/issues/568
-        self.assertEqual(b[0], e[0])
-        self.assertEqual(b[1], e[1])
-        self.assertEqual(b[2], e[2])
-        self.assertEqual(b[3], e[3])
+        self.assertEqual(b, e)
```
YenPeiChen07, line 275: Replace point-wise checking with curve equality checking in testing `Bezier3d` representation string.

tests/test_universe_shape1d.py
```diff
@@ -297,12 +301,7 @@ def test_repr_str(self):
...
```
YenPeiChen07, line 304: Replace point-wise checking with curve equality checking.

tests/test_universe_shape1d.py
```diff
@@ -150,6 +150,15 @@ def test_control_points(self):
         self.assertEqual(list(bzr[2]), [3, 1, 0])
         self.assertEqual(list(bzr[3]), [4, 0, 0])
 
+        # Test equality and inequality comparison operators
+        bzr_copy = Bezier(p0=Point(0, 0, 0), p1=Point(1, 1, 0),
+                          p2=Point(3, 1, 0), p3=Point(4, 0, 0))
+        self.assertTrue(bzr_copy == bzr)
```
YenPeiChen07, line 156: Add test for `Bezier3d` equality operator (`==`).

tests/test_universe_shape1d.py
```diff
@@ -150,6 +150,15 @@ def test_control_points(self):
...
+
+        bzr1 = Bezier(p0=Point(0, 0, 0), p1=Point(1, 1, 0), p2=Point(3, 1, 0),
+                      p3=Point(4, 4, 4))
+        self.assertTrue(bzr1 != bzr)
```
YenPeiChen07, line 160: Add test for `Bezier3d` inequality operator (`!=`).

cpp/modmesh/universe/bezier.hpp
YenPeiChen07: > I wonder if we can just do memory comparison here?

I think using value comparison may be more accurate for numerical comparison. But I could explore these two ways for more checking.

YenPeiChen07:
@yungyuc, this PR is ready for review. Please take a look. Thank you.

cpp/modmesh/universe/bezier.hpp
tigercosmos: Since we are not considering numerial errors, I think the result will be the same by using `std::memcmp`.

yungyuc: We should not compare memory address for `Bezier3d::operator==` because the operators of other geometry entities compare the coordiates.

cpp/modmesh/universe/pymod/wrap_shape1d.cpp
yungyuc: It is usually more readable to keep the wrappers for the Python operator overload in the original way.

There might be a more systematic way when we have a lot of operators to be exposed from C++ to Python for a class.

yungyuc (approved):
I changed the PR description to be `For #568`.  I will resolve issues manually.

cpp/modmesh/universe/bezier.hpp
tigercosmos: @yungyuc Sorry I didn't get it. What would be the drawback of comparing `m_data.v` memory directly? For other geometry entities, since they are also using continuous arrays as the coordiates, I think it is also possible to compare the memory for them.

yungyuc: The semantic of `operator==()` is to compare the content elements.  Comparing memory address can be done by `(&a) == (&b)`, without needing an operator overload.

Implementing `operator==()` with `m_data.v == other.m_data.v` is the same as implementing that with `this == (&other)`, and it will be ambiguous how to provide element-wise comparison.

tigercosmos: I see. Thanks for the explanation.

PR #705, opened 2026-03-28
Reorganize Python unit tests for `modmesh::universe`

yungyuc:
Python unit test cases for C++ `modmesh::universe` scattered in `test_universe.py`, `test_polygon3d.py`, and `test_universe_trap.py`, and were not organized well.  This PR reorganize the tests to `test_universe_shape0d.py`, `test_universe_shape1d.py`, and `test_universe_shape2d.py`.

> [!NOTE]
> There are not unit tests for 3-D shapes yet.  Then file `test_universe_shape3d.py` is **not** created.

Factor out the tests in `test_universe.py` to distinct files:

* `test_universe_shape0d.py`:
  * `BersteinTB`, `BersteinPolynomialTC`, `BernsteinInterpolationTC`
  * `Point3dTB`, `Point3dFp32TC`, `Point3dFp64TC`
* `test_universe_shape1d.py`:
  * `Segment3dTB`, `Segment3dFp32TC`, `Segment3dFp64TC`
  * `Bezier3dTB`, `Bezier3dFp32TC`, `Segment3dFp64TC`
  * `SegmentPadTB`, `SegmentPadFp32TC`, `SegmentPadFp64TC`
  * `CurvePadTB`, `CurvePadFp32TC`, `CurvePadFp64TC`
* `test_universe_shape2d.py`
  * `Triangle3dTB`, `Triangle3dFp32TC`, `Triangle3dFp64TC`
  * `TrianglePadTB`, `TrianglePadFp32TC`, `TrianglePadFp64TC`

Merge 2-D shape test cases to `test_universe_shape2d.py`:

* From `test_polygon3d.py`:
  * `Polygon3dTB`, `Polygon3dFp32TC`, `Polygon3dFp64TC`
* Form `test_universe_trap.py`:
  * `TrapezoidalDecomposerTB`, `TrapezoidalDecomposerFp32TC`, `TrapezoidalDecomposerFp64TC`

The tests that remains in `test_universe.py`:

* `WorldTB`, `WorldFp32TC`, `WorldFp64TC`

I have manually diff'ed to confirm the moved code is identical.

---

> [!NOTE]
> The prompts used through clion claude plugin are:
> - Split the tests in `test_universe.py` into `test_universe_shape0d.py`,
>   `test_universe_shape1d.py`, `test_universe_shape2d.py`,
>   `test_universe_shape3d.py` based on the wrapped C++ classes that are wrapped
[...]

PR #704, opened 2026-03-28
refactor: break the Python wrapping code in `universe/` to different files by shape dimension

yungyuc:
> [!NOTE]
> Moving code across files results in the massive size in diff line count in this PR.
> * I used visual diff to compare the code before and after moving (in the different files) and confirmed that it is identical.

Reorganize the Python wrapping code in `universe/pymod/` into 5 files:

1. `wrap_shape0d.cpp`: Code that are not for shape.
   - Helpers for the Bernstein polynomials.
   - `WrapPoint3d`, `WrapPointPad`.
2. `wrap_shape1d.cpp`: One-dimensional geometry entities.
   - `WrapSegment3d`, `WrapSegmentPad`.
   - `WrapBezier3d`, `WrapCurvePad`.
3. `wrap_shape2d.cpp`: Two-dimensional geometry entities.
   - `WrapTriangle3d`, `WrapTrianglePad`.
   - `WrapPolygon`, `WrapPolygonPad`.
   - `WrapTrapezoidPad`.
   - `WrapTrapezoidalDecomposer`.
4. `wrap_shape3d.cpp`: Three-dimensional geometry entities.
   - `WrapBoundBox3d`.
5. `wrap_World.cpp`: The container `World`.
   - `WrapWorld`.

> [!NOTE]
> Do not mistaken the spatial dimension of shapes with the dimension of coordinate system (the postfix in `Point3d`, `Segment3d`, etc.)

Comments are provided to guide the review.

---

> [!TIP]
> The prompts used through clion claude plugin, chronologically, are:
> * Extract `WrapPoint3d` and `WrapPointPad` from `wrap_World.cpp` to a new file `wrap_shape0d.cpp`
> * Extract `WrapSegment3d` and `WrapSegmentPad` from `wrap_World.cpp` to a new file `wrap_shape1d.cpp`
> * Extract `WrapBezier3d` and `WrapCurvePad` from the file `wrap_World.cpp` to `wrap_shape1d.cpp`
> * Extract `WrapTriangle3d` and `WrapTrianglePad` from `wrap_World.cpp` to `wrap_polygon.cpp`
> * Extract `WrapBoundBox3d` from `wrap_World.cpp` to a new file `wrap_shape3d.cpp`
> * In `wrap_polygon.cpp`, refactor `WrapPolygonPad` like `WrapTrianglePad`

PR #712, opened 2026-04-05
Add drawing shape interface to World

### LINKED ISSUE #682: Make a prototype for drawing 2D graphics
Create a prototype "canvas" for drawing 2D vector graphics.  The existing "3D viewer" can be used as a starting point.  "Sample: NACA 4-digit" may be a good example because it draws cubic Bezier curves.

The first PR should only add the most basic features.  Try to stay within 500 lines of diff.

The prototype scope only needs to create line segments and/or rectangles.  It may also create polygons and curves, but do not make one PR too long.

The 2D drawing should use [modmesh::World in universe/World.hpp](https://github.com/solvcon/modmesh/blob/abb54d5f903bbe4abf562f25d09ae2133aa89fb4/cpp/modmesh/universe/World.hpp#L44) as the data object.  The prototype strictly focuses on 2D.  3D drawing will be extended from prototype canvas in the future, probably after the 2D drawing reaches certain maturity and is no longer a prototype.

Menu items for the (2D) drawing feature should use the "Canvas" menu.

File I/O, e.g., SVG and/or OASIS, is not included in the prototype and will be tracked in follow-up issues (to be filed after this one is done).

---

Refactoring work to clarify the API to be developed:
- PR #703
- PR #704
- PR #705

tigercosmos:
## Changes

**World.hpp**: Add `ShapeType` enum, `ShapeRecord` struct, and `ShapeEntry` R-tree entry type. World gains a shape registry (`std::vector<ShapeRecord>`) and R-tree (`RTree<ShapeEntry<T>, BoundBox2d<T>>`) as new members. New methods:
- `add_triangle()` — appends 3 segments to `SegmentPad`, records offset/count, inserts bounding box into R-tree
- `translate_shape()` — modifies segments in-place, updates R-tree
- `remove_shape()` — marks `ShapeRecord` as DEAD, removes from R-tree
- `query_visible()` — R-tree viewport query returning shape IDs
- `collect_live_segments()` — returns a new `SegmentPad` excluding segments owned by DEAD shapes
- `clear()` — resets all pads, registry, and R-tree

**wrap_World.cpp**: Add `wrap_triangle()` exposing `add_triangle`. Added `translate_shape`, `remove_shape`, `nshape`, `shape_type_of`, `query_visible`, and `clear` to `wrap_management`.

**RWorld.cpp**: `RLines` constructor uses `collect_live_segments()` instead of `segments()->clone()` so removed shapes are not rendered.

**test_universe.py**: 16 tests covering shape CRUD, DEAD shape error handling, and R-tree viewport queries.

## Example

```
import modmesh as mm
from _modmesh import pilot

world = mm.WorldFp64()

# Add some triangles
t0 = world.add_triangle(0, 0, 10, 0, 5, 8)
t1 = world.add_triangle(15, 0, 25, 0, 20, 8)
t2 = world.add_triangle(0, 15, 10, 15, 5, 23)

# Translate one
world.translate_shape(t1, 0, 15)

# Remove one
world.remove_shape(t2)

# Render
widget = pilot.mgr.add3DWidget()
widget.updateWorld(world)
widget.showMark()
```

## Screenshot
Screenshot 2026-04-11 at 4 44 49 AM

### DIFF
--- a/cpp/modmesh/pilot/RWorld.cpp
+++ b/cpp/modmesh/pilot/RWorld.cpp
@@ -146,8 +146,8 @@ RLines::RLines(std::shared_ptr<WorldFp64> const & world, Qt3DCore::QNode * paren
     , m_renderer(new Qt3DRender::QGeometryRenderer())
     , m_material(new Qt3DExtras::QDiffuseSpecularMaterial())
 {
-    // Create segment pad
-    std::shared_ptr<SegmentPadFp64> segments = world->segments()->clone();
+    // Collect all segments except those from removed shapes.
+    std::shared_ptr<SegmentPadFp64> segments = world->collect_live_segments();
     // Create sampled segments in a pad from the curves
     std::shared_ptr<SegmentPadFp64> csegs = world->curves()->sample(/*length*/ 0.1);
     // Extend the overall segment pad with the sampled segments
--- a/cpp/modmesh/universe/World.hpp
+++ b/cpp/modmesh/universe/World.hpp
@@ -29,13 +29,49 @@
  */
 
 #include <modmesh/base.hpp>
+#include <modmesh/buffer/SimpleCollector.hpp>
 #include <modmesh/universe/bernstein.hpp>
 #include <modmesh/universe/bezier.hpp>
+#include <modmesh/universe/rtree.hpp>
 
 #include <deque>
+#include <vector>
 
 namespace modmesh
 {
+
+enum class ShapeType : uint8_t
+{
+    DEAD = 0, ///< deleted / unused slot
+    TRIANGLE = 1,
+}; /* end of enum class ShapeType */
+
+/**
+ * Lightweight record mapping a shape ID to its segment range in the pad.
+ */
+struct ShapeRecord
+{
+    ShapeType type;
+    size_t segment_offset; ///< first index in SegmentPad
+    size_t segment_count; ///< number of segments this shape occupies
+}; /* end of struct ShapeRecord */
+
+/**
+ * Entry stored in the R-tree: shape ID + bounding box.
+ */
+template <typename T>
+struct ShapeEntry
+{
+    int32_t shape_id;
+    BoundBox2d<T> bbox;
+    bool operator==(ShapeEntry const & other) const { return shape_id == other.shape_id; }
+}; /* end of struct ShapeEntry */
+
+template <typename T>
+struct RTreeValueOps<ShapeEntry<T>, BoundBox2d<T>>
+{
+    static BoundBox2d<T> calc_bound_box(ShapeEntry<T> const & entry) { return entry.bbox; }
+}; /* end of struct RTreeValueOps */
 
 /**
  * Manage all geometry entities.
@@ -60,10 +96,11 @@ public:
     using point_type = Point3d<T>;
     using segment_type = Segment3d<T>;
     using bezier_type = Bezier3d<T>;
-
     using point_pad_type = PointPad<T>;
     using segment_pad_type = SegmentPad<T>;
     using curve_pad_type = CurvePad<T>;
+    using bbox_type = BoundBox2d<T>;
+    using rtree_type = RTree<ShapeEntry<T>, bbox_type>;
 
     template <typename... Args>
     static std::shared_ptr<World<T>> construct(Args &&... args)
@@ -75,6 +112,7 @@ public:
         : m_points(point_pad_type::construct(/* ndim */ 3))
         , m_segments(segment_pad_type::construct(/* ndim */ 3))
         , m_curves(curve_pad_type::construct(/* ndim */ 3))
+        , m_rtree(std::make_unique<rtree_type>())
     {
     }
 
@@ -105,6 +143,7 @@ public:
     void add_segment(segment_type const & segment)
     {
         m_segments->append(segment);
+        m_bare_segment_indices.push_back(m_segments->size() - 1);
     }
     void add_segment(point_type const & p0, point_type const & p1)
     {
@@ -136,6 +175,44 @@ public:
     }
     std::shared_ptr<curve_pad_type> const & curves() { return m_curves; }
 
+    /**
+     * Add a triangle by decomposing it into 3 segments in the pad.
+     * Returns the shape ID for later reference.
+     */
+    int32_t add_triangle(T x0, T y0, T x1, T y1, T x2, T y2);
+
+    /**
+     * Translate all segments belonging to a shape by (dx, dy).
+     */
+    void translate_shape(int32_t shape_id, value_type dx, value_type dy);
+
+    /**
+     * Remove a shape from the R-tree and registry.
+     * Segments remain in the pad as dead data; use clear() to reclaim.
+     */
+    void remove_shape(int32_t shape_id);
+
+    ShapeType shape_type_of(int32_t shape_id) const { return find_shape_or_throw(shape_id).type; }
+
+    size_t nshape() const { return m_nshape; }
+
+    /**
+     * Query the R-tree for shapes whose bounding box overlaps the viewport.
+     */
+    std::vector<int32_t> query_visible(T min_x, T min_y, T max_x, T max_y) const;
+
+    /**
+     * Collect all segments except those belonging to DEAD shapes.
+     * Includes bare segments (added via add_segment) and live shape segments.
+     */
+    std::shared_ptr<segment_pad_type> collect_live_segments() const;
+
+    /**
+     * Remove all geometry entities (points, segments, curves, shapes)
+     * from the world. Rebuilds pads from scratch to reclaim memory.
+     */
+    void clear();
+
 private:
 
     void check_size(size_t i, size_t s, char const * msg) const
@@ -146,11 +223,171 @@ private:
         }
     }
 
+    bbox_type compute_shape_bbox(ShapeRecord const & rec) const;
+
+    /// Check if shape_id is valid and not DEAD.
+    /// @throw std::out_of_range if shape_id is out of bounds or shape is DEAD.
+    void check_shape_id(int32_t shape_id) const;
+
+    ShapeRecord const & find_shape_or_throw(int32_t shape_id) const
+    {
+        check_shape_id(shape_id);
+        return m_shape_registry[shape_id];
+    }
+
+    ShapeRecord & find_shape_or_throw(int32_t shape_id)
+    {
+        check_shape_id(shape_id);
+        return m_shape_registry[shape_id];
+    }
+
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    SimpleCollector<size_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
+
     std::shared_ptr<curve_pad_type> m_curves;
 
+    // TODO: Replace std::vector with a custom SoA container and BoundBoxPad
+    // auxiliary class. Consider moving the registry into the R-tree.
+    std::vector<ShapeRecord> m_shape_registry;
+
+    size_t m_nshape = 0; ///< count of live (non-DEAD) shapes
+    std::unique_ptr<rtree_type> m_rtree; ///< spatial index for shapes for viewport query
+
 }; /* end class World */
+
+template <typename T>
+int32_t World<T>::add_triangle(T x0, T y0, T x1, T y1, T x2, T y2)
+{
+    size_t offset = m_segments->size();
+    m_segments->append(point_type(x0, y0, 0), point_type(x1, y1, 0));
+    m_segments->append(point_type(x1, y1, 0), point_type(x2, y2, 0));
+    m_segments->append(point_type(x2, y2, 0), point_type(x0, y0, 0));
+
+    int32_t shape_id = static_cast<int32_t>(m_shape_registry.size());
+    m_shape_registry.push_back(ShapeRecord{ShapeType::TRIANGLE, offset, 3});
+    ++m_nshape;
+    m_rtree->insert(ShapeEntry<T>{shape_id, compute_shape_bbox(m_shape_registry[shape_id])});
+    return shape_id;
+}
+
+template <typename T>
+void World<T>::translate_shape(int32_t shape_id, value_type dx, value_type dy)
+{
+    ShapeRecord const & rec = find_shape_or_throw(shape_id);
+    // Remove old entry from R-tree before modifying segments.
+    m_rtree->remove(ShapeEntry<T>{shape_id, compute_shape_bbox(rec)});
+    for (uint32_t i = 0; i < rec.segment_count; ++i)
+    {
+        size_t idx = rec.segment_offset + i;
+        m_segments->x0(idx) += dx;
+        m_segments->y0(idx) += dy;
+        m_segments->x1(idx) += dx;
+        m_segments->y1(idx) += dy;
+    }
+    // Reinsert with updated bounding box.
+    m_rtree->insert(ShapeEntry<T>{shape_id, compute_shape_bbox(rec)});
+}
+
+template <typename T>
+void World<T>::remove_shape(int32_t shape_id)
+{
+    ShapeRecord & rec = find_shape_or_throw(shape_id);
+    m_rtree->remove(ShapeEntry<T>{shape_id, compute_shape_bbox(rec)});
+    rec.type = ShapeType::DEAD;
+    --m_nshape;
+}
+
+template <typename T>
+std::vector<int32_t> World<T>::query_visible(T min_x, T min_y, T max_x, T max_y) const
+{
+    bbox_type viewport(min_x, min_y, max_x, max_y);
+    std::vector<ShapeEntry<T>> hits;
+    m_rtree->search(viewport, hits);
+    std::vector<int32_t> ids;
+    ids.reserve(hits.size());
+    for (auto const & entry : hits)
+    {
+        ids.push_back(entry.shape_id);
+    }
+    return ids;
+}
+
+template <typename T>
+std::shared_ptr<typename World<T>::segment_pad_type> World<T>::collect_live_segments() const
+{
+    // Mark segment indices owned by DEAD shapes.
+    small_vector<bool> dead(m_segments->size(), false);
+    for (auto const & rec : m_shape_registry)
+    {
+        if (rec.type != ShapeType::DEAD)
+        {
+            continue;
+        }
+        for (uint32_t i = 0; i < rec.segment_count; ++i)
+        {
+            dead[rec.segment_offset + i] = true;
+        }
+    }
+    auto result = segment_pad_type::construct(/* ndim */ 3);
+    for (size_t i = 0; i < m_segments->size(); ++i)
+    {
+        if (!dead[i])
+        {
+            // includes both bare segments and segments of live shapes
+            result->append(m_segments->get(i));
+        }
+    }
+    return result;
+}
+
+template <typename T>
+void World<T>::clear()
+{
+    m_points = point_pad_type::construct(/* ndim */ 3);
+    m_segments = segment_pad_type::construct(/* ndim */ 3);
+    m_curves = curve_pad_type::construct(/* ndim */ 3);
+    // TODO: SimpleCollector should have a `clear()` method.
+    m_bare_segment_indices = SimpleCollector<size_t>();
+    m_shape_registry.clear();
+    m_nshape = 0;
+    m_rtree = std::make_unique<rtree_type>();
+}
+
+template <typename T>
+typename World<T>::bbox_type World<T>::compute_shape_bbox(ShapeRecord const & rec) const
+{
+    T mn_x = std::numeric_limits<T>::max();
+    T mn_y = std::numeric_limits<T>::max();
+    T mx_x = std::numeric_limits<T>::lowest();
+    T mx_y = std::numeric_limits<T>::lowest();
+    for (uint32_t i = 0; i < rec.segment_count; ++i)
+    {
+        size_t idx = rec.segment_offset + i;
+        mn_x = std::min({mn_x, m_segments->x0(idx), m_segments->x1(idx)});
+        mn_y = std::min({mn_y, m_segments->y0(idx), m_segments->y1(idx)});
+        mx_x = std::max({mx_x, m_segments->x0(idx), m_segments->x1(idx)});
+        mx_y = std::max({mx_y, m_segments->y0(idx), m_segments->y1(idx)});
+    }
+    return bbox_type(mn_x, mn_y, mx_x, mx_y);
+}
+
+template <typename T>
+void World<T>::check_shape_id(int32_t shape_id) const
+{
+    if (shape_id < 0 ||
+        static_cast<size_t>(shape_id) >= m_shape_registry.size())
+    {
+        throw std::out_of_range(
+            std::format("World: shape_id {} not found", shape_id));
+    }
+    if (m_shape_registry[shape_id].type == ShapeType::DEAD)
+    {
+        throw std::invalid_argument(
+            std::format("World: shape_id {} is DEAD", shape_id));
+    }
+}
 
 using WorldFp32 = World<float>;
 using WorldFp64 = World<double>;
--- a/cpp/modmesh/universe/pymod/wrap_World.cpp
+++ b/cpp/modmesh/universe/pymod/wrap_World.cpp
@@ -62,6 +62,7 @@ protected:
     WrapWorld & wrap_point();
     WrapWorld & wrap_segment();
     WrapWorld & wrap_bezier();
+    WrapWorld & wrap_triangle();
 };
 /* end class WrapWorld */
 
@@ -76,6 +77,7 @@ WrapWorld<T>::WrapWorld(pybind11::module & mod, const char * pyname, const char 
         .wrap_point()
         .wrap_segment()
         .wrap_bezier()
+        .wrap_triangle()
         //
         ;
 }
@@ -90,6 +92,41 @@ WrapWorld<T> & WrapWorld<T>::wrap_management()
             py::init(
                 []()
                 { return wrapped_type::construct(); }))
+        //
+        ;
+
+    (*this)
+        .def_property_readonly("nshape", &wrapped_type::nshape)
+        .def(
+            "remove_shape",
+            &wrapped_type::remove_shape,
+            py::arg("shape_id"))
+        .def(
+            "shape_type_of",
+            [](wrapped_type const & self, int32_t shape_id)
+            {
+                ShapeType st = self.shape_type_of(shape_id);
+                switch (st)
+                {
+                case ShapeType::TRIANGLE: return std::string("triangle");
+                default: return std::string("unknown");
+                }
+            },
+            py::arg("shape_id"))
+        .def("clear", &wrapped_type::clear)
+        .def(
+            "translate_shape",
+            &wrapped_type::translate_shape,
+            py::arg("shape_id"),
+            py::arg("dx"),
+            py::arg("dy"))
+        .def(
+            "query_visible",
+            &wrapped_type::query_visible,
+            py::arg("min_x"),
+            py::arg("min_y"),
+            py::arg("max_x"),
+            py::arg("max_y"))
         //
         ;
 
@@ -217,6 +254,30 @@ WrapWorld<T> & WrapWorld<T>::wrap_bezier()
                 return self.bezier_at(i);
             })
         .def_property_readonly("curves", &wrapped_type::curves)
+        //
+        ;
+
+    return *this;
+}
+
+template <typename T>
+WrapWorld<T> & WrapWorld<T>::wrap_triangle()
+{
+    namespace py = pybind11;
+
+    (*this)
+        .def(
+            "add_triangle",
+            [](wrapped_type & self, value_type x0, value_type y0, value_type x1, value_type y1, value_type x2, value_type y2)
+            {
+                return self.add_triangle(x0, y0, x1, y1, x2, y2);
+            },
+            py::arg("x0"),
+            py::arg("y0"),
+            py::arg("x1"),
+            py::arg("y1"),
+            py::arg("x2"),
+            py::arg("y2"))
         //
         ;
 
--- a/cpp/modmesh/universe/rtree.hpp
+++ b/cpp/modmesh/universe/rtree.hpp
@@ -411,7 +411,9 @@ private:
             return node;
         }
 
-        node_type * best_child = nullptr;
+        // Initialize to first child to avoid undefined behavior when all
+        // enlargements equal std::numeric_limits<value_type>::max().
+        node_type * best_child = &node->nodes.front();
         value_type min_enlargement = std::numeric_limits<value_type>::max();
 
         for (auto & child : node->nodes)
--- a/tests/test_universe.py
+++ b/tests/test_universe.py
@@ -272,4 +272,119 @@ class WorldFp64TC(WorldTB, unittest.TestCase):
     def test_type(self):
         self.assertIs(modmesh.WorldFp64, self.World)
 
+
+class WorldShapeTC(unittest.TestCase):
+    """Shape registry: add, translate, remove, clear."""
+
+    def setUp(self):
+        self.w = modmesh.WorldFp64()
+
+    def test_add_triangle(self):
+        sid = self.w.add_triangle(0, 0, 1, 0, 0, 1)
+        self.assertEqual(self.w.nshape, 1)
+        self.assertEqual(self.w.nsegment, 3)
+        self.assertEqual(self.w.shape_type_of(sid), "triangle")
+
+    def test_add_multiple(self):
+        self.w.add_triangle(0, 0, 1, 0, 0, 1)
+        self.w.add_triangle(2, 2, 3, 2, 2, 3)
+        self.assertEqual(self.w.nshape, 2)
+        self.assertEqual(self.w.nsegment, 6)
+
+    def test_ids_are_unique(self):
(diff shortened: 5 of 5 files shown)


tigercosmos (commented):
Key design points highlighted inline.

tigercosmos:
@yungyuc  Please take a look. Thank you.

yungyuc (changes requested):
This does not look right.  `World` is the canvas.  It is not necessary to have another class inside `World` for creating geometry objects.

And it is called `World` so that the geometry objects do not have to be in a surface, but can be in a space.  The `World` is part of the `universe` module.  While we do not have a `Universe` container yet, it may be a future extension to hold multiple `World`.

And it is not correct to postfix the names to the geometry entities with `2D` (although `d` should be lower-cased per the naming convention, it is not the point).  Polygons, rectangles, triangles, circles, polylines, ellipses are all 2D entities.

The `3d` postfix we have for `Point`, `Segment`, etc., meant the entities are in 3D space.

I am moving this PR to draft for you to follow up.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+ * Open polyline (not closed) defined by an ordered vertex list (>= 2).
+ */
+template <typename T>
+class Polyline
+    : public Shape<T>
+{
+
+public:
+
+    using typename Shape<T>::value_type;
```
tigercosmos: **Shape management on World**: Shapes are managed directly by World — no separate Canvas2D. `add_*` methods create shapes, `get_shape(id)` retrieves by ID, `remove_shape(id)` deletes, `clear_shapes()` removes all. Uses `unordered_map<int32_t, shape_ptr>` with sequential IDs for O(1) average lookup.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+// ---------------------------------------------------------------------------
+// Shape base class
+// ---------------------------------------------------------------------------
+
+/**
+ * Abstract base for every shape managed by World.
+ *
+ * All concrete shapes must implement the pure-virtual methods below.  Shapes
+ * live on the z = 0 plane; Point3d is used for compatibility with the
+ * existing rendering pipeline.
```
tigercosmos: **Shape base class**: Abstract base for all shapes managed by World. Concrete shapes implement `draw(World&)` to emit segments, plus `translate/scale/rotate` for the **Update** part of CRUD (transforms are called on the shape object, not through World).

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +835,91 @@ class World
     }
     std::shared_ptr<curve_pad_type> const & curves() { return m_curves; }
 
+    // -- shapes -------------------------------------------------------------
+
+    /// Add a pre-built shape. Returns the assigned ID.
+    int32_t add_shape(shape_ptr shape)
```
tigercosmos: **World shape CRUD**: **Create** — `add_shape()` and typed helpers (`add_rectangle`, `add_circle`, etc.) return an `int32_t` ID. **Read** — `get_shape(id)`, `nshape`. **Update** — via shape methods (translate/scale/rotate). **Delete** — `remove_shape(id)`, `clear_shapes()`. Storage is `unordered_map<int32_t, shape_ptr>` with sequential IDs for O(1) average lookup.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -28,13 +28,278 @@
...
+            .def("rotate", &wrapped_type::rotate, py::arg("angle_deg"), py::arg("cx") = 0, py::arg("cy") = 0)
+            .def("type_name", &wrapped_type::type_name)
+            //
+            ;
+    }
+};
+
+template <typename T>
+class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapPolygon
+    : public WrapBase<WrapPolygon<T>, Polygon<T>, std::shared_ptr<Polygon<T>>, Shape<T>>
```
tigercosmos: **pybind11 inheritance**: Concrete shape wrappers pass `Shape<T>` as the 4th `WrapBase` template arg so pybind11 registers the C++ inheritance hierarchy. This enables polymorphic returns from `get_shape()` and correct `isinstance()` checks in Python.

modmesh/core.py
```diff
@@ -163,6 +163,22 @@
     'CurvePadFp64',
     'WorldFp32',
     'WorldFp64',
+    'ShapeFp32',
+    'ShapeFp64',
```
tigercosmos: Shape types exported here so they are accessible as `modmesh.ShapeFp64`, `modmesh.RectangleFp64`, etc.

tigercosmos:
@yungyuc  I have update the interface design based on the review. Please take a look, thanks!

cpp/modmesh/universe/World.hpp
yungyuc: It is not a correct design to use a shared pointer and `enable_shared_from_this` for a polymorphic class.  The shared pointer provides runtime resource management, and polymorphism incurs additional runtime function dispatching.  Their interaction creates complexity unnecessary to our geometry operation system (while they can be used together technically).

We do not have a lot of shapes to develop.  We probably only need polygon/polyline, triangle, circle/ellipse, (parabola), (hyperbola).  It is not a number large enough to justify the use of polymorphism, which may be a must when there are more than 30 types to manage.

In conclusion, do not implement the shape creation code with polymorphism.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+    {
+        // TODO(drawing API): implement
+        throw std::runtime_error("Polygon::clone: not yet implemented");
+    }
+
+    std::string type_name() const override { return "Polygon"; }
+
+private:
+
+    std::vector<point_type> m_vertices;
```
yungyuc: Do not use STL for coordinate member data.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+
+}; /* end class Polygon */
+
+/**
+ * Axis-aligned rectangle (before transforms) defined by lower-left corner
+ * and dimensions.
+ */
+template <typename T>
+class Rectangle
```
yungyuc: We already have axis-aligned rectangles called `Box`.  Rectangles should not be axis-aligned.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+    T m_width;
+    T m_height;
+
+}; /* end class Rectangle */
+
+/**
+ * Triangle defined by three vertices.
+ */
+template <typename T>
+class Triangle
```
yungyuc: We already have `Triangle` (`Triangle3d`).  Do not duplicate.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+    point_type m_p1;
+    point_type m_p2;
+
+}; /* end class Triangle */
+
+/**
+ * Circle approximated as a regular *nseg*-gon.
+ */
+template <typename T>
+class Circle
```
yungyuc: Circle and ellipse can use a single class.  A circle is a special case ellipse.

cpp/modmesh/universe/World.hpp
```diff
@@ -32,11 +32,702 @@
...
+
+// ---------------------------------------------------------------------------
+// Concrete shape classes
+// ---------------------------------------------------------------------------
+
+/**
+ * Closed polygon defined by an arbitrary vertex list (>= 3 vertices).
+ */
+template <typename T>
+class Polygon
```
yungyuc: There is already `Polygon3d`.  Do not duplicate.

Polygon and polyline are the same thing except the former is closed loop while the latter is open.  They should be the same class.  Just use `Polygon3d` with an additional open/close bit for polyline.

We can discuss whether or not to rename `Polygon3d` to `Polyline3d`.

cpp/modmesh/universe/World.hpp
yungyuc: A general, add-any-shape helper is not needed.  It should not be provided.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +835,103 @@ class World
...
+    // -- shapes -------------------------------------------------------------
+
+    /// Add a pre-built shape. Returns the assigned ID.
+    int32_t add_shape(shape_ptr shape)
+    {
+        // TODO(drawing API): implement
+        throw std::runtime_error("World::add_shape: not yet implemented");
+    }
+
+    int32_t add_polygon(std::vector<point_type> const & vertices)
```
yungyuc: You should implement the shape CRUD helper functions by using existing ones: `add_point()`, `add_segment()`, etc.

yungyuc (changes requested):
To add drawing shape interface to the `World`, helper functions should be created.  It does not look right to create container objects.

You should implement the shape CRUD helper functions by using existing ones: `add_point()`, `add_segment()`, etc.  The PR should use less than 300-500 lines if you do it this way.

Points needed to be addressed:

- [x] Do not use polymorphism to create shapes.
- [x] Do not make rectangles to be axis-aligned.  We already have axis-aligned rectangles called `Box`.
- [x] We already have `Triangle3d`.  Do not duplicate.
- [x] Use ellipsis for circle.
- [x] We already have `Polygon3d`.  Do not duplicate.

A general reminder that do not need to be addressed:

- Do not use STL for coordinate member data.

cpp/modmesh/universe/World.hpp
tigercosmos: @yungyuc I wonder if using CRTP makes sense in this case, and what would be mininum number of shapes to use CRTP?

yungyuc: We will eventually have CRTP somewhere for the shapes.  I feel using CRTP now can be an over design, but am not 100% sure.

A simple guidance is to do it if you can make it within 300 lines of diff.  Otherwise, don't.

cpp/modmesh/pilot/RWorld.cpp
```diff
@@ -146,8 +146,8 @@ RLines::RLines(std::shared_ptr<WorldFp64> const & world, Qt3DCore::QNode * paren
     , m_renderer(new Qt3DRender::QGeometryRenderer())
     , m_material(new Qt3DExtras::QDiffuseSpecularMaterial())
 {
-    // Create segment pad
-    std::shared_ptr<SegmentPadFp64> segments = world->segments()->clone();
+    // Collect all segments except those from removed shapes.
+    std::shared_ptr<SegmentPadFp64> segments = world->collect_live_segments();
```
tigercosmos, line 150: Since triangle only affact segments, so we only take care segments here. Points and curves are TODO.

cpp/modmesh/universe/World.hpp
```diff
@@ -31,12 +31,46 @@
...
+enum class ShapeType : uint8_t
+{
+    DEAD = 0, ///< deleted / unused slot
+    TRIANGLE = 1,
+};
+
+/**
+ * Lightweight record mapping a shape ID to its segment range in the pad.
+ */
+struct ShapeRecord
```
tigercosmos, line 52: Lightweight record for tracking.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
     }
     std::shared_ptr<curve_pad_type> const & curves() { return m_curves; }
 
+    // -- shapes -------------------------------------------------------------
+
+    /**
+     * Add a triangle by decomposing it into 3 segments in the pad.
+     * Returns the shape ID for later reference.
+     */
+    int32_t add_triangle(T x0, T y0, T x1, T y1, T x2, T y2)
```
tigercosmos: Use adding triangles as the GUI API example.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
...
+    {
+        size_t offset = m_segments->size();
+        m_segments->append(point_type(x0, y0, 0), point_type(x1, y1, 0));
+        m_segments->append(point_type(x1, y1, 0), point_type(x2, y2, 0));
+        m_segments->append(point_type(x2, y2, 0), point_type(x0, y0, 0));
+
+        int32_t shape_id = static_cast<int32_t>(m_shape_registry.size());
+        m_shape_registry.emplace_back(ShapeType::TRIANGLE, offset, 3);
```
tigercosmos: Register the shape.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
...
+        ++m_nshape;
+        m_rtree->insert(ShapeEntry<T>{shape_id, compute_shape_bbox(m_shape_registry.back())});
+        return shape_id;
+    }
+
+    /**
+     * Translate all segments belonging to a shape by (dx, dy).
+     */
+    void translate_shape(int32_t shape_id, value_type dx, value_type dy)
```
tigercosmos: Only add translation as the "update" API example.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
...
+    }
+
+    size_t nshape() const { return m_nshape; }
+
+    // -- viewport query / renderer ------------------------------------------
+
+    /**
+     * Query the R-tree for shapes whose bounding box overlaps the viewport.
+     */
+    std::vector<int32_t> query_visible(T min_x, T min_y, T max_x, T max_y) const
```
tigercosmos: example of querying by RTree for later support of the viewport.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
...
+            ids.push_back(entry.shape_id);
+        }
+        return ids;
+    }
+
+    /**
+     * Collect all segments except those belonging to DEAD shapes.
+     * Includes bare segments (added via add_segment) and live shape segments.
+     */
+    std::shared_ptr<segment_pad_type> collect_live_segments() const
```
tigercosmos: Find the live segments to print.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +179,131 @@ class World
...
+                dead[rec.segment_offset + i] = true;
+            }
+        }
+        auto result = segment_pad_type::construct(/* ndim */ 3);
+        for (size_t i = 0; i < m_segments->size(); ++i)
+        {
+            if (!dead[i])
+            {
+                // includes both bare segments and segments of live shapes
+                result->append(m_segments->get(i));
```
tigercosmos: Includes both bare segments and shape segments. Note that curves and points are TODO.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,10 +314,64 @@ class World
...
+    ShapeRecord & find_shape_or_throw(int32_t shape_id)
+    {
+        check_shape_id(shape_id);
+        return m_shape_registry[shape_id];
+    }
+
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    std::vector<int32_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
```
tigercosmos: Fore bare segments. Points and curvers are TODO.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,10 +314,64 @@ class World
...
+
     std::shared_ptr<curve_pad_type> m_curves;
 
+    //// Use contiguous vector for shape registry for the larger number of shapes and better cache locality.
+    // Deleted shapes are marked as DEAD but not removed to preserve segment offsets.
+    std::vector<ShapeRecord> m_shape_registry;
```
tigercosmos: Use std::vector to support a large number of shapes in the future.

tests/test_pilot_drawing.py
```diff
@@ -0,0 +1,154 @@
...
+    def test_clear(self):
+        self.w.add_triangle(0, 0, 1, 0, 0, 1)
+        self.w.add_triangle(2, 2, 3, 2, 2, 3)
+        self.w.add_segment(mm.Point3dFp64(0, 0), mm.Point3dFp64(1, 1))
+        self.w.clear()
+        self.assertEqual(self.w.nshape, 0)
+        self.assertEqual(self.w.nsegment, 0)
+
+
+class WorldViewportTC(unittest.TestCase):
```
tigercosmos: Viewport test.

tests/test_pilot_drawing.py
```diff
@@ -0,0 +1,154 @@
...
+    world.translate_shape(sid, dx=10, dy=20)
+    widget.updateWorld(world)
+"""
+
+import unittest
+
+import modmesh as mm
+
+
+class WorldShapeTC(unittest.TestCase):
```
tigercosmos: CRUD test.

tigercosmos (commented):
@yungyuc  Please review the new design. For now, the API example includes only triangles. The design has already considered supporting a large number of shapes. The shape registry uses a vector for contiguous memory storage, and an R-tree is planned to enable efficient viewport-based queries going forward. Please also check the PR description for the screenshot.

cpp/modmesh/universe/rtree.hpp
```diff
@@ -411,7 +411,7 @@ class RTree
             return node;
         }
 
-        node_type * best_child = nullptr;
+        node_type * best_child = &node->nodes.front();
```
tigercosmos, line 416: Fix the bug.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -223,6 +225,61 @@ WrapWorld<T> & WrapWorld<T>::wrap_bezier()
     return *this;
 }
 
+template <typename T>
+WrapWorld<T> & WrapWorld<T>::wrap_shape()
```
yungyuc: Segments and Bezier curves are also (kind of) shapes.  In my definition, there are 0, 1, 2, and 3 dimensional shapes.  That's why the test files `test_universe_shape0d.py`, `test_universe_shape1d.py`, and `test_universe_shape2d.py` are there.

The name `wrap_shape()` here is ambiguous.  At the time being, it's better to name it as `wrap_triangle()`.

In the future, when there are more shapes to wrap, we can reorganize the wrapping code.

cpp/modmesh/universe/rtree.hpp
yungyuc: What problem it caused?

cpp/modmesh/universe/World.hpp
```diff
@@ -31,12 +31,46 @@
...
+#include <modmesh/universe/rtree.hpp>
 
 #include <deque>
 
 namespace modmesh
 {
 
+enum class ShapeType : uint8_t
+{
+    DEAD = 0, ///< deleted / unused slot
```
yungyuc, line 45: I like the name.  `DEAD` is vivid.  Much better than `INVALID` or `NULLITY`.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -223,6 +225,61 @@ WrapWorld<T> & WrapWorld<T>::wrap_bezier()
...
+            {
+                return self.add_triangle(x0, y0, x1, y1, x2, y2);
+            },
+            py::arg("x0"),
+            py::arg("y0"),
+            py::arg("x1"),
+            py::arg("y1"),
+            py::arg("x2"),
+            py::arg("y2"))
+        .def(
```
yungyuc: The wrappers for the generic concept of "your shape" should be factored to the second `(*this) ... ;` section, like:

```cpp
    (*this)
        .def(
            "add_triangle",
        ...
        ;

    // Separate section.
    (*this)
        .def(
            "translate_shape",
        ...
        ;
```

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -223,6 +225,61 @@ WrapWorld<T> & WrapWorld<T>::wrap_bezier()
...
+            {
+                ShapeType st = self.shape_type_of(shape_id);
+                switch (st)
+                {
+                case ShapeType::TRIANGLE: return std::string("triangle");
+                default: return std::string("unknown");
+                }
+            },
+            py::arg("shape_id"))
+        .def("clear", &wrapped_type::clear)
```
yungyuc: `nshape`, `remove_shape()`, `shape_type_of()`, and `clear()` should go to `wrap_management()`

cpp/modmesh/universe/World.hpp
```diff
@@ -85,6 +121,8 @@ class World
     World & operator=(World &&) = delete;
     ~World() = default;
 
+    // -- points -------------------------------------------------------------
```
yungyuc: Do not add the "structural" separator `---` comments in the file.  They are your coding notes.

The ethereal notes will soon get outdated and misleads contributors.  They look innocent right now, but are malign.

cpp/modmesh/universe/World.hpp
yungyuc: This is exactly what we need.

cpp/modmesh/universe/World.hpp
yungyuc: Replace `std::vector` with `SimpleCollector` when `value_type` is a fundamental type.

Do not use STL for member data unless it is just in a prototype phase (then you need a TODO comment and a follow-up PR/issue).

For local variables, sometimes STL is OK, but discouraged.

cpp/modmesh/universe/World.hpp
```diff
@@ -31,12 +31,46 @@
...
+}; /* end of struct ShapeRecord */
+
+/**
+ * Entry stored in the R-tree: shape ID + bounding box.
+ */
+template <typename T>
+struct ShapeEntry
+{
+    int32_t shape_id;
+    BoundBox2d<T> bbox;
```
yungyuc, line 66: I don't think we need `BoundBox2d`.  `BoundBox3d` can be used for 2D.  It should be fixed and tracked in a separate issue.

A nice thing for the semantic of "box" is the implication for both 2 and 3 dimensional space.

cpp/modmesh/universe/World.hpp
yungyuc: Leave a TODO to remind to replace it with a custom SoA container for the record.  It will also create `BoundBoxPad` as an auxiliary class.

Maybe the registry can be moved to r tree.  We can think about it later.

tests/test_pilot_drawing.py
yungyuc, line 1: Do not create a new test file for the code.  Reuse `test_universe.py`.

yungyuc (changes requested):
Also take a look at `test_universe_shape0d.py`, `test_universe_shape1d.py`, and `test_universe_shape2d.py`, and PR #704 and #705.

Points to address:
- [x] Rename `wrap_shape()` to `wrap_triangle()`.  Organize the wrapped helpers.
- [x] Remove the "structural" separator `---` comments.
- [x] Do not use STL for member data unless it is just in a prototype phase.
  - [x] `m_bare_segment_indices`
  - [x] `m_shape_registry`
- [x] Do not create the new test file `test_pilot_drawing.py`.  Reuse `test_universe.py`.

cpp/modmesh/universe/World.hpp
tigercosmos: Since the size of `m_bare_segment_indices` is unknown, I suppose using `small_vector` would be better in this case.

cpp/modmesh/universe/World.hpp
tigercosmos: Created a issue: https://github.com/solvcon/modmesh/issues/721

cpp/modmesh/universe/rtree.hpp
tigercosmos: Initialize to first child to avoid undefined behavior when all enlargements equal `std::numeric_limits<value_type>::max()`.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -76,6 +77,7 @@ WrapWorld<T>::WrapWorld(pybind11::module & mod, const char * pyname, const char
         .wrap_point()
         .wrap_segment()
         .wrap_bezier()
+        .wrap_triangle()
```
tigercosmos, line 80: Renamed.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -90,6 +92,37 @@ WrapWorld<T> & WrapWorld<T>::wrap_management()
             py::init(
                 []()
                 { return wrapped_type::construct(); }))
+        .def_property_readonly("nshape", &wrapped_type::nshape)
```
tigercosmos, line 99: The following are under `wrap_management` now.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,10 +304,64 @@ class World
...
+    }
+
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    small_vector<size_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
+
     std::shared_ptr<curve_pad_type> m_curves;
 
+    // TODO: Replace with a custom SoA container and BoundBoxPad auxiliary
```
tigercosmos: Added a TODO.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,10 +304,64 @@ class World
...
+    ShapeRecord & find_shape_or_throw(int32_t shape_id)
+    {
+        check_shape_id(shape_id);
+        return m_shape_registry[shape_id];
+    }
+
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    small_vector<size_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
```
tigercosmos: Changed to use `small_vector`.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,10 +304,64 @@ class World
...
+
     std::shared_ptr<curve_pad_type> m_curves;
 
+    // TODO: Replace with a custom SoA container and BoundBoxPad auxiliary
+    // class.  Consider moving the registry into the R-tree.
+    small_vector<ShapeRecord> m_shape_registry;
```
tigercosmos: Changed to use `small_vector`.

tests/test_universe.py
```diff
@@ -272,4 +272,119 @@ def assert_allclose(self, *args, **kw):
     def test_type(self):
         self.assertIs(modmesh.WorldFp64, self.World)
 
+
+class WorldShapeTC(unittest.TestCase):
```
tigercosmos, line 276: Put the test cases under `test_universe.py`.

tigercosmos (commented):
@yungyuc  Please take look. Thanks!

cpp/modmesh/universe/World.hpp
tigercosmos: Removed.

cpp/modmesh/universe/World.hpp
yungyuc: Thank you for creating issue #721 to track it.

cpp/modmesh/universe/pymod/wrap_World.cpp
yungyuc: It's better to separate constructors and others in two `(*this) /*...*/ ;` sections.  But it can also be done in a future PR.

cpp/modmesh/universe/World.hpp
yungyuc: As usual, move non-accessor function bodies to be outside the class declaration.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +174,126 @@ class World
...
+        }
+        // Reinsert with updated bounding box.
+        m_rtree->insert(ShapeEntry<T>{shape_id, compute_shape_bbox(rec)});
+    }
+
+    /**
+     * Remove a shape from the R-tree and registry.
+     * Segments remain in the pad as dead data; use clear() to reclaim.
+     */
+    void remove_shape(int32_t shape_id)
```
yungyuc: This function `remove_shape()` is short enough to have a place in the class declaration, but when it grows longer, it needs to move out.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +174,126 @@ class World
...
+    {
+        ShapeRecord & rec = find_shape_or_throw(shape_id);
+        m_rtree->remove(ShapeEntry<T>{shape_id, compute_shape_bbox(rec)});
+        rec.type = ShapeType::DEAD;
+        --m_nshape;
+    }
+
+    ShapeType shape_type_of(int32_t shape_id) const
```
yungyuc: It is OK to make this function `shape_type_of()` a one-liner like `nshape()`.

cpp/modmesh/universe/World.hpp
yungyuc: Don't use `small_vector` for `m_shape_registry`.  `std::vector` is better than `small_vector` here because it does not make sense to add only a couple of shapes in a `World`.

The name "registry" is misleading here, but it should be handled in the future because the additional shapes (types) will force us to change the architecture anyway.

cpp/modmesh/universe/World.hpp
yungyuc: Don't use `small_vector` for `m_bare_segment_indices`.  `std::vector` is better than `small_vector` here because it does not make sense to add only a couple of segments in a `World`.

Use a `SimpleCollector` instead.

yungyuc (changes requested):
Container types strongly influence the architecture.  Points to address:

- [x] Move non-accessor function bodies to be outside the class declaration when the code is not 2-3 times longer than an accessor.
- [x] Don't use `small_vector` for `m_bare_segment_indices`.  Use `SimpleCollector` instead.
- [x] Don't use `small_vector` for `m_shape_registry`.

cpp/modmesh/universe/World.hpp
tigercosmos: Copied.

cpp/modmesh/universe/pymod/wrap_World.cpp
```diff
@@ -93,6 +95,41 @@ WrapWorld<T> & WrapWorld<T>::wrap_management()
         //
         ;
 
+    (*this)
```
tigercosmos, line 98: Separated.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,12 +223,172 @@ class World
...
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    SimpleCollector<size_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
+
     std::shared_ptr<curve_pad_type> m_curves;
 
+    // TODO: Replace std::vector with a custom SoA container and BoundBoxPad
+    // auxiliary class. Consider moving the registry into the R-tree.
+    std::vector<ShapeRecord> m_shape_registry;
```
tigercosmos, line 253: Use `std::vector`. Also modified the comment.

cpp/modmesh/universe/World.hpp
```diff
@@ -136,6 +175,44 @@ class World
     }
     std::shared_ptr<curve_pad_type> const & curves() { return m_curves; }
 
+    /**
+     * Add a triangle by decomposing it into 3 segments in the pad.
+     * Returns the shape ID for later reference.
+     */
+    int32_t add_triangle(T x0, T y0, T x1, T y1, T x2, T y2);
```
tigercosmos, line 182: Move the implementation after the class defination.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,12 +223,172 @@ class World
...
+    ShapeRecord & find_shape_or_throw(int32_t shape_id)
+    {
+        check_shape_id(shape_id);
+        return m_shape_registry[shape_id];
+    }
+
     std::shared_ptr<point_pad_type> m_points;
+
     std::shared_ptr<segment_pad_type> m_segments;
+    SimpleCollector<size_t> m_bare_segment_indices; ///< indices of segments not owned by any shape
```
tigercosmos, line 247: Use SimpleCollector.

cpp/modmesh/universe/World.hpp
```diff
@@ -146,12 +223,172 @@ class World
...
+}
+
+template <typename T>
+void World<T>::clear()
+{
+    m_points = point_pad_type::construct(/* ndim */ 3);
+    m_segments = segment_pad_type::construct(/* ndim */ 3);
+    m_curves = curve_pad_type::construct(/* ndim */ 3);
+    // TODO: SimpleCollector should have a `clear()` method.
+    m_bare_segment_indices = SimpleCollector<size_t>();
```
tigercosmos, line 352: SimpleCollector doesn't have the `clear()` method.

tigercosmos (commented):
@yungyuc  The issues are addressed. Please take a look. Thanks!

cpp/modmesh/universe/World.hpp
tigercosmos: Created an issue: https://github.com/solvcon/modmesh/issues/725

cpp/modmesh/universe/World.hpp
yungyuc: "Owned by" is not accurate.  It's more accurate to write:

```cpp
///< indices of segments not belonging to any shape
```

yungyuc (approved):
LGTM

yungyuc:
This PR is for issue #682

PR #714, opened 2026-04-05
Add pre-commit and replace redundant CI lint steps

tigercosmos:
## Summary
- Add `.pre-commit-config.yaml` with hooks matching CI lint checks:
  - **clang-format** (v20.1.0) — C++ formatting for `cpp/` and `gtests/`
  - **flake8** (7.3.0) — Python linting (excludes `thirdparty/tmp/_deps`)
  - **cinclude** — Ensure C++ includes use angle brackets, not quotes
  - **checkascii** — Ensure source files contain only ASCII characters
  - **checktws** — Ensure no trailing whitespace in source files
- Replace redundant CI lint steps with a single `pre_commit` job using [`pre-commit/action`](https://github.com/pre-commit/action)
- Remove the `clang_format_check` job and individual `make cinclude`, `make checkascii`, `make checktws`, `make flake8` steps from CI
- Rename `tidy_flake8` job to `clang_tidy` (now only runs clang-tidy + pilot builds)

## Test plan
- [x] `pre-commit run --all-files` passes all 5 hooks locally
- [x] CI `pre_commit` job passes
- [x] CI `clang_tidy` job passes (unchanged functionality)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,17 +8,11 @@ on:
 
 jobs:
 
-  clang_format_check:
+  pre_commit:
 
     runs-on: ubuntu-latest
 
     if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
-
-    strategy:
-      matrix:
-        path:
-          - 'cpp'
-          - 'gtests'
 
     steps:
 
@@ -28,29 +22,15 @@ jobs:
         run: |
           echo "github.event_name: ${{ github.event_name }}"
 
-      - name: Add 8G swap (Ubuntu)
-        # Prevent hitting runner's resource limits
-        if: runner.os == 'Linux'
-        run: |
-          # Remove /swapfile first to avoid "fallocate: Text file busy" error
-          sudo swapoff -a
-          sudo rm -f /swapfile
-          sudo fallocate -l 8G /swapfile
-          sudo chmod 600 /swapfile
-          sudo mkswap /swapfile
-          sudo swapon /swapfile
-          free -h
-
-      - name: Run clang-format style check for C/C++/Protobuf programs.
-        uses: jidicula/clang-format-action@v4.16.0
+      - uses: actions/setup-python@v5
         with:
-          clang-format-version: '20'
-          check-path: ${{ matrix.path }}
-          fallback-style: 'LLVM' # optional
+          python-version: '3.x'
 
-  tidy_flake8:
+      - uses: pre-commit/action@v3.0.1
 
-    name: Run clang-tidy and flake8 on ${{ matrix.os }}
+  clang_tidy:
+
+    name: Run clang-tidy on ${{ matrix.os }}
 
     if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
 
@@ -103,15 +83,6 @@ jobs:
         restore-keys: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
         create-symlink: true
 
-    - name: make cinclude (check_include)
-      run: make cinclude
-
-    - name: make checkascii (check ASCII-only characters)
-      run: make checkascii
-
-    - name: make checktws (check trailing whitespace)
-      run: make checktws
-
     - name: make pilot USE_PYTEST_HELPER_BINDING=OFF
       run: |
         make pilot \
@@ -135,22 +106,15 @@ jobs:
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_PYTEST_HELPER_BINDING=ON"
 
-    - name: make flake8
-      run: |
-        make flake8 \
-          ${JOB_MAKE_ARGS} \
-          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
-          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
-
   send_email_on_failure:
-    needs: [clang_format_check, tidy_flake8]
+    needs: [pre_commit, clang_tidy]
     # Run if any of the dependencies failed in master branch
     if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' && github.event.repository.fork == false }}
     uses: ./.github/workflows/send_email_on_fail.yml
     with:
       job_results: |
-        - clang_format_check: ${{ needs.clang_format_check.result }}
-        - tidy_flake8: ${{ needs.tidy_flake8.result }}
+        - pre_commit: ${{ needs.pre_commit.result }}
+        - clang_tidy: ${{ needs.clang_tidy.result }}
     secrets:
       EMAIL_USERNAME: ${{ secrets.EMAIL_USERNAME }}
       EMAIL_PASSWORD: ${{ secrets.EMAIL_PASSWORD }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,50 @@
+# Copyright (c) 2026, An-Chi Liu <phy.tiger@gmail.com>
+# BSD-style license; see COPYING
+#
+# See https://pre-commit.com for more information
+# Run `pre-commit install` to set up the git hooks.
+# Run `pre-commit run --all-files` to check the entire project.
+# Run `pre-commit autoupdate` to update hook versions.
+repos:
+  # C++ formatting
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v20.1.0
+    hooks:
+      - id: clang-format
+        files: ^(cpp|gtests)/.*\.[ch]pp$
+
+  # Python linting
+  - repo: https://github.com/PyCQA/flake8
+    rev: 7.3.0
+    hooks:
+      - id: flake8
+        exclude: ^(thirdparty|tmp|_deps)/
+
+  # Project-specific checks
+  - repo: local
+    hooks:
+      # Ensure C++ files use angle brackets for includes, not quotes
+      - id: cinclude
+        name: check-include-quotes
+        description: Ensure C++ includes use angle brackets, not quotes
+        entry: '#include\s*"'
+        language: pygrep
+        files: ^cpp/
+
+      # Ensure source files contain only ASCII characters
+      - id: checkascii
+        name: check-ascii
+        description: Ensure source files contain only ASCII characters
+        entry: python3 contrib/lint/check_ascii.py
+        language: system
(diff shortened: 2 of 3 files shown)


contrib/prompt/CLAUDE.md
```diff
@@ -65,6 +65,20 @@ make flake8       # Python style check
 make checkascii   # ASCII character check
 ```
 
+### Pre-commit
```
tigercosmos, line 68: Let AI use pre-commit before it pushes the code.

tigercosmos:

.github/workflows/lint.yml
```diff
@@ -8,18 +8,12 @@ on:
 
 jobs:
 
-  clang_format_check:
+  pre_commit:
```
tigercosmos, line 11: Use pre-commit for CI to shorten the CI file.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2026, An-Chi Liu <phy.tiger@gmail.com>
```
yungyuc, line 1: CI files do not need copyright.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,50 @@
...
+# BSD-style license; see COPYING
+#
+# See https://pre-commit.com for more information
```
yungyuc, line 4: I am not sure an additional dependency for running the check is good.

yungyuc (changes requested):
I am not familiar with https://pre-commit.com/ .  After skimming through the website, I do not see its benefits to modmesh.

For local hacking, I preferred to have all checks done manually.  It is difficult to align the local development environments.

One who wants to run checks before committing can write a script, which looks like what `$ pre-commit run --all-files` does.

I am converting the PR to draft.  We can discuss here.

ExplorerRay:
As far as I know, `pre-commit` is a tool for developers to check some basic tests in local environments. After setting `pre-commit` up, it will check before people do `git commit`.

Although this tool adds some dependencies, these things are optional. The lint and code style we want to check still remain in the CI, so everything would be checked before merging.

Therefore, I think it's good because it can make developers do some basic prechecks in local side easier by packaging them under `pre-commit`.

[This](https://github.com/sustainable-computing-io/kepler/blob/main/docs/developer/pre-commit.md) is the pre-commit docs I have read before from an open-source project called `kepler`. I think we may need a docs like this, otherwise if no one knows how to set up `pre-commit`, it might be meaningless to replace original step with `pre-commit`.

tigercosmos:
`pre-commit` is useful for doing linting or styling checks. Of course we can manually run each check one by one, but it is more convinient to run by a single command (even better than a hand-crafted all-in-one script). As `pre-commit` is widely used by open source projects, it should be easy for developers to catch up and it can also benifit AI usage.

yungyuc:
Feeling easy or not depends on familiarity.  For someone not familiar with a tool, it will not be easy.  From my personal perspective, the existing [`make lint`](https://github.com/solvcon/modmesh/blob/b926ff48c6670b2f7581eb59deabb09632229df5/Makefile#L195) is the easiest and the only dependency is make.

Could you please provide more objective analysis of why pre-commit is good (not necessarily better)?  And where it is better than the existing make targets?

ExplorerRay:
> the existing [make lint](https://github.com/solvcon/modmesh/blob/b926ff48c6670b2f7581eb59deabb09632229df5/Makefile#L195) is the easiest and the only dependency is make.

I think current `make lint` also needs clang-format and flake8 as its dependencies. 

Using `pre-commit` can make all those dependencies installed by `pre-commit install` and the versions of these dependencies can be controlled by `.pre-commit-config.yaml`.

tigercosmos:
@ExplorerRay pointed out a good feature.
Other than that, one of my favoriate feature of pre-commit is that it can import actions (like GitHub Action) just like how a GitHub worflow import other pre-defined actions. With pre-commit, we can also introduce usuful predefined actions without our implementation, which can save a lot of time and gain benifits.

yungyuc:
> Using `pre-commit` can make all those dependencies installed by `pre-commit install` and the versions of these dependencies can be controlled by `.pre-commit-config.yaml`.

It is a plus to have an automatic way to manage the dependencies required by the actions.  What are the limitations of the actions and action dependencies?

> Other than that, one of my favoriate feature of pre-commit is that it can import actions (like GitHub Action) just like how a GitHub worflow import other pre-defined actions. With pre-commit, we can also introduce usuful predefined actions without our implementation, which can save a lot of time and gain benifits.

It sounds good to have additional modularity for building/CI.  It probably worth of clarification for how it would strengthen or interfere the modularization that we want for the Github Actions.  @tigercosmos @chestercheng @ExplorerRay I'd like to have your input on this.

> > the existing [make lint](https://github.com/solvcon/modmesh/blob/b926ff48c6670b2f7581eb59deabb09632229df5/Makefile#L195) is the easiest and the only dependency is make.
> 
> I think current `make lint` also needs clang-format and flake8 as its dependencies.

Good point.  What I meant by "the only dependency" was for the tool itself, not the jobs it runs.  Pre-commit has few dependencies but still more than make has.

tigercosmos:
> It is a plus to have an automatic way to manage the dependencies required by the actions. What are the limitations of the actions and action dependencies?

AFAIK, it has no limitation for most of the usage. It works as same as using `make` with scripts.

> It sounds good to have additional modularity for building/CI. It probably worth of clarification for how it would strengthen or interfere the modularization that we want for the Github Actions. @tigercosmos @chestercheng @ExplorerRay I'd like to have your input on this.

Currently, the modularity exists in GitHub action level (workflow yaml), and now the modularity exists in pre-commit level (config yaml). I would say there is no intererence. Whatever GitHub action can do, pre-commit can also do, e.g. instaling actions, executing actions.

> Good point. What I meant by "the only dependency" was for the tool itself, not the jobs it runs. Pre-commit has few dependencies but still more than make has.

Pre-commit also doesn't have any dependecy. If you meant that now we also need to install one more `pre-commit` tool other than just `make`, then yes, unfortunetly. However, considering the aforementioned benifits, I think it is worth it.

chestercheng:
I‘m not sure whether pre-commit solves a real problem. It still requires most of the same dependencies as make lint, so it feels more like a workflow change than an improvement.

From my perspective, if we introduce a new workflow, I would prefer it to remain compatible with the existing one, as most contributors are already familiar with it. Otherwise, it should convince developers to adopt it and adjust their workflows.

yungyuc:
> > It is a plus to have an automatic way to manage the dependencies required by the actions. What are the limitations of the actions and action dependencies?
> 
> AFAIK, it has no limitation for most of the usage. It works as same as using `make` with scripts.

It seems to require Python for managing the actions and action dependencies.  Please correct me and elaborate if I am wrong.

> > It sounds good to have additional modularity for building/CI. It probably worth of clarification for how it would strengthen or interfere the modularization that we want for the Github Actions. @tigercosmos @chestercheng @ExplorerRay I'd like to have your input on this.
> 
> Currently, the modularity exists in GitHub action level (workflow yaml), and now the modularity exists in pre-commit level (config yaml). I would say there is no intererence. Whatever GitHub action can do, pre-commit can also do, e.g. instaling actions, executing actions.

It does not look accurate to say what GitHub Actions can do, pre-commit can also do.  The way to install action dependencies differs between GitHub Actions and pre-commit.

> > Good point. What I meant by "the only dependency" was for the tool itself, not the jobs it runs. Pre-commit has few dependencies but still more than make has.
> 
> Pre-commit also doesn't have any dependecy. If you meant that now we also need to install one more `pre-commit` tool other than just `make`, then yes, unfortunetly. However, considering the aforementioned benifits, I think it is worth it.

I see.  Thank you for correcting.  I saw https://github.com/pre-commit/pre-commit/blob/main/requirements-dev.txt but it is probably only for developing pre-commit, not using it.

> I‘m not sure whether pre-commit solves a real problem. It still requires most of the same dependencies as make lint, so it feels more like a workflow change than an improvement.

It's true.  In terms of workflow, it also seems to me an alternate way for execution.  The modularity is similar.  It modularizes differently from what make does.

There are "1.5" additional capabilities.  One is to install actions and action dependencies by using the mechanism provided by pre-commit.  Make can also manage the actions (targets) and their dependencies, but requires writing custom recipes/commands.

Another half is to unify the modularity between Github Actions and pre-commit using YAML.  I say this is only half, because it's of similar complexity to call make from GitHub Actions for a similar level of integration.

> From my perspective, if we introduce a new workflow, I would prefer it to remain compatible with the existing one, as most contributors are already familiar with it. Otherwise, it should convince developers to adopt it and adjust their workflows.

This PR indeeds is an introduction of a new workflow and should be discussed among a broader audience.  @tigercosmos @ExplorerRay @chestercheng , could you please help take it to the [solvcon channel on sciwork discord](https://discord.gg/P9U7PFv)?

I am making this PR ready for review for we have a good understanding for what it does.

tigercosmos:
> It seems to require Python for managing the actions and action dependencies. Please correct me and elaborate if I am wrong.

True, it requires `pip` to install `pre-commi`, and the pre-commit workflow might be Python as well.

> It does not look accurate to say what GitHub Actions can do, pre-commit can also do. The way to install action dependencies differs between GitHub Actions and pre-commit.

I guess you are talking about manually installing dependency from scratch. For this part, GitHub CI might be more easilier.

> Another half is to unify the modularity between Github Actions and pre-commit using YAML. I say this is only half, because it's of similar complexity to call make from GitHub Actions for a similar level of integration.

I think it takes more effort if we use `make` to archieve the same result. But funtionality-wide, I agree it's almost the same.

> This PR indeeds is an introduction of a new workflow and should be discussed among a broader audience. @tigercosmos @ExplorerRay @chestercheng , could you please help take it to the [solvcon channel on sciwork discord](https://discord.gg/P9U7PFv)?

Already mentioned on Discord. Let's see.

yungyuc:
@chestercheng Please gather to discuss in the sprint today and settle it.

ExplorerRay:
We had a discussion in today’s sprint and asked @j8xixo12 and @c1ydehhx for their opinions.

1. Users tend to run `make lint` instead of `pre-commit`. If we plan to change the CI, `pre-commit` should align with the current `make lint` behavior so that existing users' local workflows remain compatible.
2. Users often commit partially staged changes, and the `pre-commit` hook may block this.

Therefore, I think it's better to only add this as an option by maintaining `.pre-commit-config.yaml` instead of combining it into CI.

yungyuc:
I did not realize that it is suggested to install pre-commit as a git hook for every commit, including local ones.  In that case, it strongly interferes with a productive workflow that makes partially-working commits.  Thank you for clarifying that, @ExplorerRay @j8xixo12 @c1ydehhx .  (Especially to @ExplorerRay for leading the discussions.)  It is a no-go to me.

Also, if not combining pre-commit to CI, it's just an alternate way to lint requiring additional maintenance.  It does not look good.

@tigercosmos I suggest to close this PR.  I don't think we are ready for it.

Alternately, if people want to use pre-commit for themselves, we may add an ignore entry for the config file.

tigercosmos:
I see. I am sorry to hear this. Let me prepare a PR for `make lint`.

yungyuc:
This is a good PR and discussion.  There are merits of pre-commit.  The project is not in a shape ready for the tool.  We can revisit in the future.

Issue #718, opened 2026-04-12
Separate and stabilize profiling runs


yungyuc:
The `pyprof` runs failed with [`build_macOS-15_Release`](https://github.com/solvcon/modmesh/actions/runs/24288224310/job/70921055164#step:13:14):
```
/Users/runner/work/modmesh/modmesh/profiling/profile_arithmetic_simd.py:111: RuntimeWarning: divide by zero encountered in divide
  return np.divide(src1, src2)
/bin/sh: line 1: 61862 Abort trap: 6           env PYTHONPATH=/Users/runner/work/modmesh/modmesh /opt/homebrew/bin/python3 ${fn} > ${outfn}
make: *** [pyprof] Error 1
```

At the same time, a building phase failed with `build_ubuntu-24.04_Release`.  It is uncertain whether or not they are related.

But having profiling runs for every checkin is distracting as the development of the project becoming more active.  We should move the profiling runs to a distinct workflow.

The profiling workflow should be run nightly instead of every commit or PR.  But it should use a (GitHub) repository secret to allow running it along with a commit or PR, so that a contributor can test it using only their fork, without needing to modify the workflow file or set up a special running environment.

ExplorerRay:
Is there any other steps than `make pyprof` running for profiling?

If not, I think I can quickly split this step into another workflow file called `profiling.yml`.

yungyuc:
Thank you.  That sounds like a plan.  I took a look and did not find other profiling runs.  I think it's good to go ahead.

@KHLee529 might want to double check.

KHLee529:
@ExplorerRay As far as I know, the profiling job runs only `make pyprof` and I agree with your plan. That's good.

yungyuc:
Fixed by PR #727

PR #727, opened 2026-04-13
ci: split profiling into another workflow file

ExplorerRay:
This PR splits profiling step original in devbuild workflow into another workflow for issues #718.

Because the original profiling step only runs on macos and ubuntu runners, I only run this on them.

In order to speed up compiling time, I reuse the ccache saved from devbuild. Because I saw [the current usage of Github cache](https://github.com/solvcon/modmesh/actions/caches) is always approaching limit, I don't save extra ccache for this workflow.

### DIFF
--- a/.github/workflows/profiling.yml
+++ b/.github/workflows/profiling.yml
@@ -1,0 +1,84 @@
+name: profiling
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '34 17 * * *'
+
+jobs:
+  profile:
+
+    # Run profiling only on schedule or when MMGH_FORCE_PROFILE is set to 'enable'
+    # You can refer to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables to set MMGH_FORCE_PROFILE in the fork repository settings for testing purposes.
+    if: ${{ github.event_name == 'schedule' || vars.MMGH_FORCE_PROFILE == 'enable' }}
+
+    name: profile_${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+
+    env:
+      QT_DEBUG_PLUGINS: 1
+      QT_QPA_PLATFORM: offscreen # for Ubuntu runner
+      PIP_BREAK_SYSTEM_PACKAGES: 1 # disabling PEP668 (for MacOS runner)
+      # Fix issue: https://github.com/solvcon/modmesh/issues/366
+      # Use custom config for jurplel/install-qt-action@v4
+      AQT_CONFIG: "thirdparty/aqt_settings.ini"
+
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, macos-15]
+      fail-fast: false
+
+    steps:
+
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: event name
+        run: |
+          echo "Event name: ${{ github.event_name }}"
+
+      - name: setup dependencies for Linux
+        if: runner.os == 'Linux'
+        uses: ./.github/actions/setup_linux
+        with:
+          workflow: 'build'
+
+      - name: setup dependencies for macOS
+        if: runner.os == 'macOS'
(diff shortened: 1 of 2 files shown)


.github/workflows/profiling.yml
```diff
@@ -0,0 +1,82 @@
...
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '34 17 * * *'
+
+jobs:
+  profile:
+
+    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
```
yungyuc: In issue #718, we said that *the profiling workflow should be run nightly instead of every commit or PR. But it should use a (GitHub) repository secret to allow running it along with a commit or PR*.

It seems to me when the workflow is run by schedule, `github.event_name == 'schedule'`.  If that is truth, I think the condition should be:

```
if: ${{ github.event_name == 'schedule' || (github.event_name != 'schedule' && vars.MMGH_FORCE_PROFILE == 'enable') }}
```

The variable name `MMGH_FORCE_PROFILE` looks more like what we meant for the action, rather than `MMGH_NIGHTLY`.

Please correct me if I am wrong.

Please also add code comment to document how to use the secret variable `MMGH_FORCE_PROFILE`.

yungyuc (changes requested):
- [x] Please clarify the running condition.
- [x] Add comment.

.github/workflows/profiling.yml
ExplorerRay: > In issue https://github.com/solvcon/modmesh/issues/718, we said that the profiling workflow should be run nightly instead of every commit or PR. But it should use a (GitHub) repository secret to allow running it along with a commit or PR.

Thanks for the remind, I think I missed this part.

> `if: ${{ github.event_name == 'schedule' || (github.event_name != 'schedule' && vars.MMGH_FORCE_PROFILE == 'enable') }}`

If we want to use `MMGH_FORCE_PROFILE` variable for forcing profiling, I think the condition can be simplified to 
this one because I think we only care whether this event is scheduled(nightly) or someone sets this var as enable.
```
if: ${{ github.event_name == 'schedule' || vars.MMGH_FORCE_PROFILE == 'enable' }}
```

ExplorerRay: I have updated.

About the comment, the length of reference docs link is really long, I am not sure whether it's good or not.

.github/workflows/profiling.yml
```diff
@@ -0,0 +1,84 @@
...
+  push:
+  pull_request:
+  schedule:
+    - cron: '34 17 * * *'
+
+jobs:
+  profile:
+
+    # Run profiling only on schedule or when MMGH_FORCE_PROFILE is set to 'enable'
+    # You can refer to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#defining-configuration-variables-for-multiple-workflows to set MMGH_FORCE_PROFILE in the fork repository settings for testing purposes.
```
yungyuc: Please keep only https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables for the link to the document.

.github/workflows/profiling.yml
```diff
@@ -0,0 +1,84 @@
...
+    if: ${{ github.event_name == 'schedule' || vars.MMGH_FORCE_PROFILE == 'enable' }}
```
yungyuc, line 14: It's good to use a simpler condition.

yungyuc (changes requested):
- [x] Use short URL.

Please leave a global comment to notify when you have done the change.  Inline annotation is harder to track.  An inline annotation is good for focused discussion, but should be used along with global comment to drive the review workflow.

ExplorerRay:
> Please keep only https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables for the link to the document.

Okay, I have updated it.

yungyuc (approved):
LGTM.  I'll merge once CI finishes.

PR #716, opened 2026-04-10
Implement array to scalar division to `SimpleArray`

KHLee529:
This PR implements the division operation between matrices and scalar number for `SimpleArray`, mentioned in issue #588 
 
The following formula is an example of such operation
```math
\underbrace{\begin{bmatrix}
1 & 2 \\
3 & 4 \\ 
\end{bmatrix}}_\text{Matrix} \div \underbrace{2}_\text{scalar} = 
\begin{bmatrix}
0.5 & 1 \\
1.5 & 2 \\ 
\end{bmatrix}
```

PR #730, opened 2026-04-14
Increase swap of ubuntu runner to avoid CI failure

KHLee529:
The CI action [failed](https://github.com/solvcon/modmesh/actions/runs/24280038147/job/70906378483) after merging PR #716. The failure due to job time-out is assumed to be caused by lack of memory in ubuntu runner.

This PR, therefore, add 4GB swap to the ubuntu runner (from 8GB to 12GB) and the effectiveness has been [verified in my forked repo](https://github.com/KHLee529/modmesh/actions/runs/24307280526) where this job always failed.

### DIFF
--- a/.github/actions/setup_linux/action.yml
+++ b/.github/actions/setup_linux/action.yml
@@ -9,14 +9,14 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Add 8G swap
+    - name: Add 12G swap
       # Prevent hitting runner's resource limits when running clang-tidy
       shell: bash
       run: |
         # Remove /swapfile first to avoid "fallocate: Text file busy" error
         sudo swapoff -a
         sudo rm -f /swapfile
-        sudo fallocate -l 8G /swapfile
+        sudo fallocate -l 12G /swapfile
         sudo chmod 600 /swapfile
         sudo mkswap /swapfile
         sudo swapon /swapfile


KHLee529:
@yungyuc @ExplorerRay Please review this PR.

yungyuc:
> @yungyuc @ExplorerRay Please review this PR.

@ExplorerRay please help make a decision what is proper swap size for the Ubuntu runner.

ExplorerRay:
LGTM. But I am not sure the remaining 2GB space for storing dependencies and modmesh code is enough or not in the future.

KHLee529:
> LGTM. But I am not sure the remaining 2GB space for storing dependencies and modmesh code is enough or not in the future.

@ExplorerRay Do you think adding 2GB is better maybe?

yungyuc:
> > LGTM. But I am not sure the remaining 2GB space for storing dependencies and modmesh code is enough or not in the future.
> 
> @ExplorerRay Do you think adding 2GB is better maybe?

Just give it a try and let us know what you find out.

ExplorerRay:
For me, I may use some command like `df -h` in Github runner to check remaining space. If current dependency only costs less than 1GB, I think it is enough.

chestercheng:
If we run into disk space issues in the future, we could consider free up space by removing preinstalled toolchain that we don’t use.
Ref: https://github.com/actions/runner-images/discussions/9329

So, I’m okay with this PR.

chestercheng (approved):

yungyuc:
Merged.  If the value needs update, a new PR can be used.

PR #732, opened 2026-04-16
ci: upgrade MacOS runners to Tahoe (macos-26)

### LINKED ISSUE #729: Upgrade macos runners to Tahoe `macos-26`
Github Actions introduced `macos-26-arm64` last year.  Most of our contributors are using the latest macos.  We should upgrade the runner version.

The workflows use both amr64 and intel.  When upgrading, we should also make it clear between `macos-26-arm64` or `macos-26-intel`.  The tag `macos-26` means arm64 but the name is ambiguous and should be avoided.

ExplorerRay:
This PR is for https://github.com/solvcon/modmesh/issues/729.

Besides revising matrix in original workflow, I upgrade Qt version for MacOS so that missing AGL framework failure can be fixed. Another is adding macos-26 support in `thirdparty/metal-cpp.sh`.

### DIFF
--- a/thirdparty/metal-cpp.sh
+++ b/thirdparty/metal-cpp.sh
@@ -45,6 +45,10 @@ if [ $(uname) == Darwin ] ; then
     filename=metal-cpp_macOS15.2_iOS18.2.zip
     url=https://developer.apple.com/metal/cpp/files/metal-cpp_macOS15.2_iOS18.2.zip
     checksum=3437e4abfbd3d45217f34772ef3502f3
+  elif [ ${ver:0:2} == "26" ] ; then
+    filename=metal-cpp_26.4.zip
+    url=https://developer.apple.com/metal/cpp/files/metal-cpp_26.4.zip
+    checksum=9065f9930f2e9f1bb48eb0094da46e73
   else
     echo "Unsupported macOS version $ver"
     exit 1
--- a/.github/actions/setup_macos/action.yml
+++ b/.github/actions/setup_macos/action.yml
@@ -25,12 +25,12 @@ runs:
         # qt6 versioning independently from the OS.
         # brew install llvm@21 qt6
         brew install llvm@21
-        # The macos-15 runner ships with Homebrew llvm@18 linked into
+        # The macos-26 runner ships with Homebrew llvm@20 linked into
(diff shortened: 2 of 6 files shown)


thirdparty/metal-cpp.sh
```diff
@@ -45,6 +45,10 @@ if [ $(uname) == Darwin ] ; then
     filename=metal-cpp_macOS15.2_iOS18.2.zip
     url=https://developer.apple.com/metal/cpp/files/metal-cpp_macOS15.2_iOS18.2.zip
     checksum=3437e4abfbd3d45217f34772ef3502f3
+  elif [ ${ver:0:2} == "26" ] ; then
```
yungyuc, line 48: Thanks!  It's not been updated for a while.

yungyuc (approved):
LGTM.

In the future, do not write like "fix issue number" in PR comment or checkin logs.  See https://github.com/solvcon/modmesh/blob/2ef304dbe54f1491ca3d06715a8397a19f4ce7e1/contrib/prompt/CLAUDE.md?plain=1#L237

We manage the issues manually without letting github automation to do it.

I'll merge once CI finishes.  @chestercheng please take a look.

ExplorerRay:
> In the future, do not write like "fix issue number" in PR comment or checkin logs.

If I don't mention the issue in this PR, how do people know that this PR is related to one specific issue quickly?

ExplorerRay:
BTW, I just run the profile CI in local fork to check macos-26 also works well in profiling workflow. [ref](https://github.com/ExplorerRay/modmesh/actions/runs/24495828076/job/71630402014)

yungyuc:
> If I don't mention the issue in this PR, how do people know that this PR is related to one specific issue quickly?

We should mention the issue number, but do not say "fix" or "resolve".  You can write like what I modified for you: "for #729".  Alternately you can write "related to issue #729" or alike.

ExplorerRay:
> > If I don't mention the issue in this PR, how do people know that this PR is related to one specific issue quickly?
> 
> We should mention the issue number, but do not say "fix" or "resolve". You can write like what I modified for you: "for #729". Alternately you can write "related to issue #729" or alike.

Oh, I got it. Thanks for the remind.

PR #734, opened 2026-04-18
Rework make lint and unify CI to use make targets

tigercosmos:
## Summary

This PR makes contributors and CI run the exact same lint recipes through `make`. `make lint` now drives `cformat`, `cinclude`, `flake8`, `checkascii`, and `checktws` straight from the Makefile, with no CMake configure step and no `silversearcher-ag` dependency.

Two behavior changes are worth calling out. First, `cformat` now defaults to `clang-format --dry-run -Werror`, so formatting drift fails the build; previously it printed warnings and exited 0. Second, `cformat` warns when the local `clang-format` major version differs from the CI pin (`CLANG_FORMAT_CI_VERSION`, currently `20`). If `clang-format`, `flake8`, or `black` is missing, the target now prints an explicit `pip install` hint instead of a raw "command not found".

Two new convenience targets are added: `make format` auto-fixes both C++ and Python, and `make pyformat` auto-fixes Python only (using `black`). `black` is deliberately kept out of `make lint`, because enforcing it today would require a large unrelated reformat. Wiring `black --check` into lint can happen in a follow-up after that reformat.

On the CI side, `jidicula/clang-format-action` is replaced with `make cformat`. The separate `clang_format_check` and `tidy_flake8` jobs collapse into one `lint` job. `clang-format==20.*` is installed via pip on both Linux and macOS runners. The `silversearcher-ag` apt install and the macOS `llvm@21`→`clang-format` symlink are dropped.

This PR grew out of #714, which proposed adopting `pre-commit`. That PR was closed because `pre-commit`'s hook model blocks the partial-working commits the project's workflow relies on. The useful parts of that proposal — version pinning, tool-presence feedback, one entry point for local and CI lint — are preserved here inside the existing `make lint` flow, so contributors do not need to adopt a new tool.

## Installing `clang-format`

The recommended install on every platform is `pip install 'clang-format==20.*'`. This matches the CI pin and avoids the version drift that comes with `apt` or `brew`. `make cformat` prints this exact command when `clang-format` is missing, and warns when the installed major version differs from `CLANG_FORMAT_CI_VERSION`. `apt` and `brew` do work, but they are discouraged, because the packaged version tracks the distribution, and the formatting output can change between machines.

## Test plan

- [x] `make lint` passes on the current tree.
- [x] `make cformat` fails on an injected misformatted `.cpp` (confirms the `-Werror` default).
- [x] `make format` rewrites a misformatted file in place.
- [x] `make cinclude` flags an injected `#include "..."`.
- [x] Missing-tool errors fire for `clang-format`, `flake8`, and `black`.
- [x] Version-mismatch warning fires when local `clang-format` major differs from the CI pin.
- [x] CI lint job passes on Linux (`ubuntu-24.04`) and macOS (`macos-26-intel`).

Related to #714.


.github/workflows/lint.yml
```diff
@@ -8,49 +8,9 @@ on:
...
-        if: runner.os == 'Linux'
-        run: |
-          # Remove /swapfile first to avoid "fallocate: Text file busy" error
-          sudo swapoff -a
-          sudo rm -f /swapfile
-          sudo fallocate -l 8G /swapfile
-          sudo chmod 600 /swapfile
-          sudo mkswap /swapfile
-          sudo swapon /swapfile
-          free -h
```
tigercosmos, line 42: Duplicated in the `setup_linux/action.yml`

.github/workflows/lint.yml
```diff
@@ -8,49 +8,9 @@ on:
 
 jobs:
 
-  clang_format_check:
+  lint:
```
tigercosmos, line 11: Rename to single `lint`

.github/workflows/lint.yml
```diff
@@ -112,6 +75,9 @@ jobs:
     - name: make checktws (check trailing whitespace)
       run: make checktws
 
+    - name: make flake8
+      run: make flake8
```
tigercosmos, line 79: Make it simple.

Makefile
```diff
@@ -177,11 +201,19 @@ cformat: $(CFFILES)
 
 .PHONY: cinclude
 cinclude: $(CFFILES)
-	if [ "$(shell ag \#include\ \*\" cpp/)" != "" ] ; then exit 1 ; fi
```
tigercosmos, line 180: Replace `ag` dependency by just using `grep`.

Makefile
```diff
@@ -194,6 +218,19 @@ checktws:
 .PHONY: lint
 lint: cformat cinclude flake8 checkascii checktws
 
+.PHONY: pyformat
```
tigercosmos, line 221: New added.

tigercosmos:

.github/workflows/lint.yml
```diff
@@ -8,49 +8,9 @@ on:
...
-          sudo mkswap /swapfile
-          sudo swapon /swapfile
-          free -h
-
-      - name: Run clang-format style check for C/C++/Protobuf programs.
-        uses: jidicula/clang-format-action@v4.16.0
-        with:
-          clang-format-version: '20'
-          check-path: ${{ matrix.path }}
-          fallback-style: 'LLVM' # optional
```
tigercosmos, line 49: Use `make cformat` now.

.github/actions/setup_linux/action.yml
```diff
@@ -119,6 +119,16 @@ runs:
...
         sudo pip3 install pyside6==$(qmake6 -query QT_VERSION)
 
+    - name: Install clang-format for lint
+      if: ${{ inputs.workflow == 'lint' }}
+      shell: bash
+      run: |
+        # Pinned to the clang-format major version that the Makefile's
+        # CLANG_FORMAT_CI_VERSION expects. Installed via pip so the same
+        # package works on both Linux and macOS runners.
+        sudo pip3 install 'clang-format==20.*'
```
yungyuc, line 129: Why do we install clang-format using pip but not apt for Linux?

.github/actions/setup_macos/action.yml
```diff
@@ -66,6 +68,16 @@ runs:
...
         python3 -m pip -v install --upgrade pyside6==$(qmake -query QT_VERSION) --ignore-requires-python
 
+    - name: Install clang-format for lint
+      if: ${{ inputs.workflow == 'lint' }}
+      shell: bash
+      run: |
+        # Pinned to the clang-format major version that the Makefile's
+        # CLANG_FORMAT_CI_VERSION expects. Installed via pip so the same
+        # package works on both Linux and macOS runners.
+        python3 -m pip install --break-system-packages --upgrade 'clang-format==20.*'
```
yungyuc, line 78: Why do we install clang-format using pip but not brew for mac?

.github/workflows/lint.yml
yungyuc: @ExplorerRay please help double check.

.github/workflows/lint.yml
```diff
@@ -103,6 +63,9 @@ jobs:
         restore-keys: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
         create-symlink: true
 
+    - name: make cformat (clang-format check)
+      run: make cformat
```
yungyuc, line 67: It's good to align local and CI flow.

Makefile
yungyuc: I am fine with the target name "pyformat".  But "black" make be more straight-forward and intuitive (at least to me).  Either way works for me.

yungyuc (changes requested):
The bullets in summary are not very clear.  It's fine for this PR.  But in future PR, please use a concise passage for summary, and bullets only when necessary.

It is hard to get the essence for what the PR does from a long list of bullets.  In this PR, there are 6 + 4 = 10 bullets.

Points to address:
- [x] Clarify what is the best way to install clang-format.

@ExplorerRay @chestercheng please help review.

Makefile
```diff
@@ -177,11 +193,19 @@ cformat: $(CFFILES)
...
 
 .PHONY: flake8
 flake8:
-	cmake --build $(BUILD_PATH) --target $@
+	@command -v $(FLAKE8) >/dev/null 2>&1 || { \
+		echo "Error: '$(FLAKE8)' not found in PATH."; \
+		echo "  Install: pip install flake8"; \
+		exit 1; \
+	}
+	$(FLAKE8) . --exclude thirdparty,tmp,_deps
```
ExplorerRay, line 208: `make flake8` in this PR doesn't rely on cmake anymore, maybe we can also remove `cmake/Flake8.cmake` and its related code.

ExplorerRay (changes requested):
It's good to see lint tools like `flake8` no longer need cmake configure so that this step can be put before other `clang-tidy`-related steps. 

Because this PR removes the usage of flake8 cmake related code, I think we can also remove them in this PR.

- [x] Remove old flake8 cmake related code (e.g. `cmake/Flake8.cmake`)

.github/actions/setup_linux/action.yml
tigercosmos: Using `pip` is easier, cross-plotform and easy to maintain. Technically speaking `apt` also works, but it is more troublesome when you need to specify the version that is not the default one. Previously we relied on the other third-party action, so we didn't encounter issues.

.github/actions/setup_macos/action.yml
tigercosmos: I think both `brew` and `pip` work in this case. Choosing `pip` for consistancy between Linux and macOS.

Makefile
tigercosmos: Sure, then I would like to keep `pyformat` because maybe we will change the formattor in the future.

tigercosmos:
@yungyuc  @ExplorerRay  Thanks for review. I left some comments and fix the issue. Please take a look again.

CMakeLists.txt
```diff
@@ -193,9 +191,6 @@ if(BUILD_QT)
 add_subdirectory(cpp/binary/pilot)
 endif()
 
-add_custom_target(flake8)
-flake8("flake8")
```
tigercosmos, line 197: Removed.

.github/actions/setup_linux/action.yml
yungyuc: I prefer OS native installation of dependency.  While some deviation is good for practicality, this does not look so.  OS native installation should be more stable.

Having said that, I am fine with the current approach.  If there are issues, it's not something difficult to fix.

CMakeLists.txt
yungyuc: Looks about right.

yungyuc (approved):
LGTM

contrib/prompt/CLAUDE.md
```diff
@@ -55,16 +55,26 @@ make run_pilot_pytest
...
-make flake8       # Python style check
-make checkascii   # ASCII character check
+# Individual checks
+make cinclude    # C++ #include uses <...>, not "..."
+make flake8      # Python style
+make checkascii  # ASCII-only source files
+make checktws    # no trailing whitespace
+
+# Auto-fix formatting
+make cformat     # C++ formatting (clang-format --dry-run -Werror)
```
chestercheng, line 68: `make cformat` doesn't auto-fix. Might be worth updating to avoid confusion, though since it is for AI, it may not be critical.

.github/workflows/lint.yml
chestercheng: Nice catch. I double checked and the same logic is indeed present in `setup_linux/action.yml`

Issue #719, opened 2026-04-12
Implement a linear solver for `SimpleArray`


yungyuc:
This issue is to follow up issue #588.

`SimpleArray` needs a linear solver.  It should solve for $$\vec{x}$$ with $$\mathrm{A}\vec{x} = \vec{b}$$, and also enable the evaluation of $$\mathrm{A}^{-1}$$ (matrix inversion).

This is the first implementation of a linear solver in the code base.  Any method is OK.  More methods can be added in the future.

tigercosmos:
I will take care of this issue.

yungyuc:
> I will take care of this issue.

Thank you.  Assigned.

yungyuc:
The issue is done with PR #735.

PR #735, opened 2026-04-18
Add LU decomposition with partial pivoting

tigercosmos:
## Summary
- Implement LU factorization with partial pivoting (PA=LU) for general (non-symmetric) matrices in `cpp/modmesh/linalg/lu_factorization.hpp`
- Add `lu_factorization()`, `lu_solve()`, and `lu_inv()` free functions with Python bindings for float32, float64, complex64, and complex128 types
- Add `.solve(b)` and `.inv()` convenience methods on `SimpleArray` for floating-point and complex types (integer arrays excluded)

## Test plan
- [x] `test_lu_solve_and_inv_dtypes`: parametric test across all 4 supported dtypes (float64, float32, complex128, complex64) verifying solve and inverse
- [x] `test_lu_solve_2d_rhs`: multi-RHS solve with 5×5 matrix and 5×4 RHS (n,m both > 3)
- [x] `test_lu_singular`: singular matrix raises `RuntimeError`
- [x] `test_lu_invalid_inputs`: non-square, 1D, dimension mismatch, wrong rank all raise `ValueError`
- [x] `test_simplearray_int_no_solve_inv`: integer SimpleArray does not expose solve/inv
- [x] All 35 tests pass (`pytest tests/test_linalg.py`)

## Document

Please also check the [Markdown file](https://gist.github.com/tigercosmos/4ff723582f5707e213ccc4d2cfe31a31) (gist) for detailed explanation of the math and code.

The PR is part of https://github.com/solvcon/modmesh/issues/719.

### DIFF
--- a/cpp/modmesh/linalg/lu_factorization.hpp
+++ b/cpp/modmesh/linalg/lu_factorization.hpp
@@ -1,0 +1,456 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, Anchi Liu <phy.tiger@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ * - Neither the name of the copyright holder nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <algorithm>
+#include <format>
+#include <limits>
+#include <numeric>
+#include <stdexcept>
+#include <utility>
+#include <vector>
+
+#include <modmesh/buffer/buffer.hpp>
+
+namespace modmesh
+{
+
+namespace detail
+{
+
+/**
+ * LU decomposition with partial pivoting for general (non-symmetric) matrices.
+ *
+ * Given an n-by-n matrix A, this class computes the factorization PA = LU,
+ * where:
+ *   - P is a permutation matrix (represented as a pivot index vector),
+ *   - L is a unit lower triangular matrix (diagonal elements are implicitly 1),
+ *   - U is an upper triangular matrix.
+ *
+ * L and U are stored compactly in a single n-by-n array: the strictly lower
+ * triangle holds L (excluding the unit diagonal), and the upper triangle
+ * (including the diagonal) holds U.
+ *
+ * The pivot vector piv[k] records that row k was swapped with row piv[k]
+ * during the k-th elimination step.  Swaps are applied sequentially from
+ * k = 0 to k = n-1.
+ *
+ * Supported element types: float, double, Complex<float>, Complex<double>.
+ */
+template <typename T>
+class Lu
+{
+
+    static_assert(
+        std::is_floating_point_v<T> || is_complex_v<T>,
+        "Lu<T> requires T to be a floating-point or complex type");
+
+private:
+
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+    // Helper: format a SimpleArray shape as "(d0, d1, ...)" for error messages.
+    static std::string format_shape(array_type const & arr)
+    {
+        std::string result = "(";
+        for (size_t i = 0; i < arr.ndim(); ++i)
+        {
+            if (i > 0)
+            {
+                result += ", ";
+            }
+            result += std::to_string(arr.shape(i));
+        }
+        result += ")";
+        return result;
+    }
+
+public:
+
+    /**
+     * Compute the LU factorization with partial pivoting: PA = LU.
+     *
+     * @param a  Square 2D SimpleArray (the matrix to factorize).
+     * @return   A pair of (lu_matrix, pivot_vector):
+     *           - lu_matrix: n-by-n array with L in the strictly lower triangle
+     *             and U in the upper triangle (including diagonal).
+     *           - pivot_vector: length-n vector where piv[k] is the row index
+     *             that was swapped with row k at step k.
+     * @throws std::invalid_argument if a is not a square 2D array.
+     * @throws std::runtime_error    if the matrix is singular or near-singular.
+     */
+    static std::pair<array_type, std::vector<ssize_t>> factorize(array_type const & a);
+
+    /**
+     * Solve the linear system Ax = b using LU factorization.
+     *
+     * @param a  Square 2D SimpleArray (the coefficient matrix).
+     * @param b  1D or 2D SimpleArray (the right-hand side).
+     *           If 1D, it is treated as a single column vector.
+     *           If 2D with shape (n, m), each column is a separate RHS.
+     * @return   The solution x with the same shape as b.
+     * @throws std::invalid_argument if a is not square, b has wrong rank, or
+     *         dimensions are incompatible.
+     * @throws std::runtime_error    if a is singular.
+     */
+    static array_type solve(array_type const & a, array_type const & b);
+
+    /**
+     * Compute the matrix inverse A^(-1).
+     *
+     * Internally solves AX = I where I is the identity matrix.
+     *
+     * @param a  Square 2D SimpleArray.
+     * @return   The inverse matrix A^(-1) with the same shape as a.
+     * @throws std::invalid_argument if a is not a square 2D array.
+     * @throws std::runtime_error    if a is singular.
+     */
+    static array_type inv(array_type const & a);
+
+private:
+
+    /**
+     * Forward substitution: solve Ly = Pb.
+     *
+     * Applies the row permutation recorded in piv to b, then solves the
+     * unit lower triangular system.  L is stored in the strictly lower
+     * triangle of the combined lu matrix (diagonal of L is implicitly 1).
+     *
+     * @param lu   Combined LU matrix from factorize().
+     * @param piv  Pivot vector from factorize().
+     * @param b    2D RHS matrix (already reshaped if originally 1D).
+     * @return     The intermediate result y.
+     */
+    static array_type forward_substitution(array_type const & lu, std::vector<ssize_t> const & piv, array_type const & b);
+
+    /**
+     * Backward substitution: solve Ux = y.
+     *
+     * U is stored in the upper triangle (including diagonal) of the
+     * combined lu matrix.
+     *
+     * @param lu  Combined LU matrix from factorize().
+     * @param y   The result of forward_substitution().
+     * @return    The final solution x.
+     */
+    static array_type backward_substitution(array_type const & lu, array_type const & y);
+
+}; /* end class Lu */
+
+template <typename T>
+auto Lu<T>::factorize(array_type const & a) -> std::pair<array_type, std::vector<ssize_t>>
+{
+    if (a.ndim() != 2 || a.shape(0) != a.shape(1))
+    {
+        throw std::invalid_argument(std::format(
+            "Lu::factorize: a must be a square 2D SimpleArray, but got shape {}",
+            format_shape(a)));
+    }
+
+    const auto n = static_cast<ssize_t>(a.shape(0));
+
+    // Create a working copy of the input matrix.  The algorithm modifies
+    // this array in-place, overwriting it with the combined L and U factors.
+    auto lu = array_type(a.shape(), value_type{0});
+    std::copy(a.begin(), a.end(), lu.begin());
+
+    // piv[k] records which row was swapped with row k at step k.
+    // Initialize with std::iota so piv[k] = k (identity permutation).
+    std::vector<ssize_t> piv(n);
+    std::iota(piv.begin(), piv.end(), ssize_t{0});
+
+    const auto eps = std::numeric_limits<real_type>::epsilon();
+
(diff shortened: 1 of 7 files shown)


cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+ * L and U are stored compactly in a single n-by-n array: the strictly lower
+ * triangle holds L (excluding the unit diagonal), and the upper triangle
+ * (including the diagonal) holds U.
+ *
+ * The pivot vector piv[k] records that row k was swapped with row piv[k]
+ * during the k-th elimination step.  Swaps are applied sequentially from
+ * k = 0 to k = n-1.
+ *
+ * Supported element types: float, double, Complex<float>, Complex<double>.
+ */
```
tigercosmos, line 65: I tried to make the comment clear. The PR description also includes a HTML explanation file. Almost all code are generated by AI, but I have checked all logic.

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+     * @param a  Square 2D SimpleArray (the matrix to factorize).
+     * @return   A pair of (lu_matrix, pivot_vector):
+     *           - lu_matrix: n-by-n array with L in the strictly lower triangle
+     *             and U in the upper triangle (including diagonal).
+     *           - pivot_vector: length-n vector where piv[k] is the row index
+     *             that was swapped with row k at step k.
+     * @throws std::invalid_argument if a is not a square 2D array.
+     * @throws std::runtime_error    if the matrix is singular or near-singular.
+     */
+    static std::pair<array_type, std::vector<ssize_t>> factorize(array_type const & a);
```
tigercosmos, line 110: Meat.

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+     * @param a  Square 2D SimpleArray (the coefficient matrix).
+     * @param b  1D or 2D SimpleArray (the right-hand side).
+     *           If 1D, it is treated as a single column vector.
+     *           If 2D with shape (n, m), each column is a separate RHS.
+     * @return   The solution x with the same shape as b.
+     * @throws std::invalid_argument if a is not square, b has wrong rank, or
+     *         dimensions are incompatible.
+     * @throws std::runtime_error    if a is singular.
+     */
+    static array_type solve(array_type const & a, array_type const & b);
```
tigercosmos, line 124: Solver's entry.

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+ *
+ * @param a  Square 2D SimpleArray (coefficient matrix).
+ * @param b  1D or 2D SimpleArray (right-hand side).
+ * @return   The solution x with the same shape as b.
+ */
+template <typename T>
+SimpleArray<T> lu_solve(SimpleArray<T> const & a, SimpleArray<T> const & b)
+{
+    return detail::Lu<T>::solve(a, b);
+}
```
tigercosmos, line 440: Export the free function.

tests/test_linalg.py
```diff
@@ -758,4 +758,110 @@ def test_batchfilter_with_control(self):
...
+class TestLuSolver(unittest.TestCase):
+
+    def test_lu_solve_and_inv_dtypes(self):
+        cases = [
+            # (np_dtype, SimpleArray class, matrix size, rtol, atol)
+            (np.float64, mm.SimpleArrayFloat64, 4, 1e-10, 1e-10),
+            (np.float32, mm.SimpleArrayFloat32, 3, 1e-5, 1e-5),
+            (np.complex128, mm.SimpleArrayComplex128, 3, 1e-10, 1e-10),
+            (np.complex64, mm.SimpleArrayComplex64, 2, 1e-5, 1e-5),
+        ]
```
tigercosmos, line 771: Try to cover different dtypes and matrix sizes.

tests/test_linalg.py
```diff
@@ -758,4 +758,110 @@ def test_batchfilter_with_control(self):
         np.testing.assert_allclose(xs_upd, xs_upd_np, atol=1e-12, rtol=0.0)
         np.testing.assert_allclose(ps_upd, ps_upd_np, atol=1e-12, rtol=0.0)
 
+
+class TestLuSolver(unittest.TestCase):
```
tigercosmos, line 762: There could be more test cases. I intended keep the most important ones to keep the PR small.

tigercosmos:
@yungyuc  Please take a look. The PR slightly exceeds 500 LOC, but some of the code are just document comments.

tests/test_linalg.py
yungyuc: In this case, just add more tests without worrying about the diff length.  The tests are necessary.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -29,6 +29,10 @@
 #include <modmesh/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
 
 #include <modmesh/buffer/pymod/array_common.hpp>
+#if __has_include(<modmesh/linalg/lu_factorization.hpp>)
```
yungyuc: Why do you need to test for the existence of the include file?  In what scenario it would not exist?

yungyuc (changes requested):
Points to address:
- [x] It's OK to add more testing code even though there are many lines already.  The testing code is the key to the work.
- [x] Use explicit test fixture.
- [x] Clarify why testing for the include file existence.

yungyuc:
> ## Document
> Please also check the [HTML file](https://gist.github.com/tigercosmos/4ff723582f5707e213ccc4d2cfe31a31) (gist) for detailed explanation of the math and code.

It's not easy to read the raw HTML.  Gist does not seem to allow rendering it.

image

Using markdown with mathjax (e.g., $$\mathrm{A} x = b$$) would be easier to read, but it may take additional time to write.  We can discuss what is a better approach.

tigercosmos:
> Using markdown with mathjax would be easier to read, but it may take additional time to write. We can discuss what is a better approach.

I think gist will only show the raw file anyway? Maybe I am wrong.
I choose HTML files because it can have the most flexibility on visualization, such as charts, colors or even animations.

In this case, I would suggest you download the html file, and use a browser to open it.

yungyuc:
Gist render mathjax in markdown.  See https://gist.github.com/yungyuc/3f5fd600c91a536d881c581a2045eb06 for an example.

tigercosmos:
@yungyuc Thanks, I will try using markdown next time.

yungyuc:
No problem.  Could you rewrite it in markdown this time?  It is very inconvenient to download and review HTML files.  It is also insecure.

tigercosmos:
> No problem. Could you rewrite it in markdown this time? It is very inconvenient to download and review HTML files. It is also insecure.

Sure, I have update the gist to Markdown.

tests/test_linalg.py
tigercosmos: Sure, added more test cases.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
tigercosmos: Indeed. Remove it.

tigercosmos:
@yungyuc  I have added more test cases in fixtures. Please take a look, thanks!

tests/test_linalg.py
```diff
@@ -758,4 +758,424 @@ def test_batchfilter_with_control(self):
...
+class TestLuFactorization(unittest.TestCase):
+    """Verify lu_factorization() produces a decomposition with PA = LU."""
+
+    def setUp(self):
+        # 3x3 invertible matrix with no zero pivots.
+        self.A_3x3 = np.array([
+            [2.0, 1.0, 1.0],
+            [4.0, -6.0, 0.0],
+            [-2.0, 7.0, 2.0],
+        ])
```
yungyuc, line 795: Always add `dtype='string'` when creating `ndarray`.  The explicit `dtype` specification makes maintenance and grepping easier.

tests/test_linalg.py
```diff
@@ -758,4 +758,424 @@ def test_batchfilter_with_control(self):
...
+            [13.0, 15.0, 16.0, 17.0],
+        ])
+        # 3x3 whose A[0][0] is negligibly small; a correct implementation
+        # must pivot away from it.
+        self.A_tiny_pivot = np.array([
+            [1e-15, 1.0, 0.0],
+            [1.0, 1.0, 1.0],
+            [0.0, 1.0, 2.0],
+        ])
+        # 3x3 complex matrix for complex-path coverage.
```
yungyuc, line 811: What does "complex-path" mean?  Please provide a reference in the PR comment.

tests/test_linalg.py
```diff
@@ -758,4 +758,424 @@ def test_batchfilter_with_control(self):
...
+            mm.lu_factorization(A_rect)
+
+    def test_factorize_rejects_1d_input(self):
+        # 1D input must raise the same shape error (not silently reshape).
+        A_1d = mm.SimpleArrayFloat64(array=np.array([1.0, 2.0, 3.0]))
+        with self.assertRaisesRegex(
+                ValueError, r"must be a square 2D SimpleArray"):
+            mm.lu_factorization(A_1d)
+
+    def test_factorize_rejects_singular_duplicate_row(self):
```
yungyuc, line 1104: It's important to test for singular system.  Duplicated rows or columns make the system singular, but very small eigenvalues also make the system almost singular.

Please add a test for tiny eigenvalue.

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
yungyuc: I guess you meant that it is not necessary.

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+        "Lu<T> requires T to be a floating-point or complex type");
+
+private:
+
+    using value_type = T;
+    using real_type = typename detail::select_real_t<value_type>::type;
+    using array_type = SimpleArray<value_type>;
+
+    // Helper: format a SimpleArray shape as "(d0, d1, ...)" for error messages.
+    static std::string format_shape(array_type const & arr)
```
yungyuc, line 81: `Lu::format_shape()` is long enough to be put outside the class declaration of `Lu`.  But if it is intentional to put it here, it's short enough to be OK.

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,456 @@
...
+ * k = 0 to k = n-1.
+ *
+ * Supported element types: float, double, Complex<float>, Complex<double>.
+ */
+template <typename T>
+class Lu
+{
+
+    static_assert(
+        std::is_floating_point_v<T> || is_complex_v<T>,
```
yungyuc, line 71: In the future we can consider to implement `is_real_v` by making it an alias to `std::is_floating_point_v`.  Then the error message can simply write "... a real or complex number type".

yungyuc (changes requested):
A point to address in the engine:
- [ ] `Lu::format_shape()` is long enough to be put outside the class declaration of `Lu`.

Points to address for testing:
- [ ] Always add `dtype` when creating `ndarray`.
- [ ] Clarify what does "complex-path" mean.
- [ ] Make sure the factorization error out correctly when the system has a very small eigenvalue (nearly singular).

cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
tigercosmos: Yes, thank you.

tests/test_linalg.py
tigercosmos: Fixed, add `dtype` for all arrays.

tests/test_linalg.py
tigercosmos: Modified the comment. It should meant testing `Lu<Complex<double>>`.

tests/test_linalg.py
tigercosmos: Added a new `test_factorize_rejects_near_singular_tiny_eigenvalue`.

cpp/modmesh/linalg/lu_factorization.hpp
tigercosmos: Modified.

cpp/modmesh/math/Complex.hpp
```diff
@@ -264,11 +264,17 @@ struct is_complex : std::false_type {};
...
+
+template <typename T>
+struct is_real : std::is_floating_point<T> {};
 // clang-format on
 
 template <typename T>
 constexpr bool is_complex_v = is_complex<T>::value;
 
+template <typename T>
+constexpr bool is_real_v = is_real<T>::value;
```
tigercosmos, line 276: Added `is_real_v`.

tests/test_linalg.py
```diff
@@ -758,4 +758,451 @@ def test_batchfilter_with_control(self):
...
+            [13.0, 15.0, 16.0, 17.0],
+        ], dtype=np.float64)
+        # 3x3 whose A[0][0] is negligibly small; a correct implementation
+        # must pivot away from it.
+        self.A_tiny_pivot = np.array([
+            [1e-15, 1.0, 0.0],
+            [1.0, 1.0, 1.0],
+            [0.0, 1.0, 2.0],
+        ], dtype=np.float64)
+        # 3x3 complex matrix to exercise Lu<T> instantiated for
```
tigercosmos, line 811: Fixed the comment.

tests/test_linalg.py
```diff
@@ -758,4 +758,451 @@ def test_batchfilter_with_control(self):
...
+            [1.0, 2.0, 3.0],
+            [1.0, 2.0, 3.0],
+            [4.0, 5.0, 6.0],
+        ], dtype=np.float64)
+        A = mm.SimpleArrayFloat64(array=A_sing)
+        with self.assertRaisesRegex(
+                RuntimeError, r"singular or near-singular"):
+            mm.lu_factorization(A)
+
+    def test_factorize_rejects_near_singular_tiny_eigenvalue(self):
```
tigercosmos, line 1123: New added test case for the small eigenvalue.

tests/test_linalg.py
```diff
@@ -758,4 +758,451 @@ def test_batchfilter_with_control(self):
...
+            [-2.0, 7.0, 2.0],
+        ], dtype=np.float64)
+        # 4x4 whose leading element is not the column-max, so partial
+        # pivoting must trigger.
+        self.A_4x4 = np.array([
+            [1.0, 2.0, 3.0, 4.0],
+            [5.0, 6.0, 8.0, 8.0],
+            [9.0, 10.0, 11.0, 16.0],
+            [13.0, 15.0, 16.0, 17.0],
+        ], dtype=np.float64)
```
tigercosmos: Always added `dtype`.

tigercosmos (commented):

cpp/modmesh/linalg/lu_factorization.hpp
```diff
@@ -0,0 +1,460 @@
...
+
+        // Step 3: Reject both exact singularity (pivot = 0, e.g. duplicate
+        // rows) and near-singularity (pivot at noise level).
+        real_type const pivot = abs(lu(k, k));
+        const auto eps = std::numeric_limits<real_type>::epsilon();
+
+        // Absolute threshold (~100 * machine eps); works for well-scaled inputs.
+        // TODO: make it relative to matrix/column magnitude for better robustness.
+        real_type const singular_tol = real_type(100) * eps;
+        if (pivot <= singular_tol)
```
tigercosmos, line 234: One more changed compared with the previous review.

tests/test_linalg.py
yungyuc: Good addition.

tests/test_linalg.py
yungyuc: It will help reader to always know what data type the array uses.

Please use the string version `dtype='float64'` (double quotes is OK if you like: `dtype="float64"`) for fundamental types.

In terms of functionality it does not matter to use string or object for `dtype`.  It's just for style consistency.

yungyuc (changes requested):
Only a minor formatting issue remains.  If you want to fix it in the future, I can merge now:
- [x] Use the string version for `dtype` of fundamental types, e.g., `dtype='float64'` (double quotes is OK if you like: `dtype="float64"`) for fundamental types.

tigercosmos:
@yungyuc  I have fixed the formating issue.

yungyuc (approved):
LGTM

PR #737, opened 2026-04-19
ci: update showdep.sh and call it in CI

ExplorerRay:
Original CI maintains `show dependency` step in two places for different OS. 

I think it's better to follow the comment which is directly using `contrib/dependency/showdep.sh` and only maintain this script. Therefore, I just updated current CI step commands into the script and let both `show dependency` steps run it.


contrib/dependency/showdep.sh
```diff
@@ -8,11 +8,19 @@ echo "cmake path: $(which cmake)"
...
 echo "pytest path: $(which pytest)"
 echo "pytest version: $(pytest --version)"
 echo "clang-tidy path: $(which clang-tidy)"
 echo "clang-tidy version: $(clang-tidy -version)"
+echo "clang-format path: $(which clang-format 2>/dev/null || echo not-installed)"
+echo "clang-format version: $(clang-format --version 2>/dev/null || echo not-installed)"
 echo "flake8 path: $(which flake8)"
-echo "flake8 version: $(flake8 --version)"
\ No newline at end of file
+echo "flake8 version: $(flake8 --version)"
```
tigercosmos, line 26: In latest master, we introduce `blank`. Maybe also add it.

yungyuc (approved):
LGTM

Issue #721, opened 2026-04-12
Unify BoundBox2d and BoundBox3d into a single BoundBox3d


tigercosmos:
## Context

From PR review discussion: https://github.com/solvcon/modmesh/pull/712#discussion_r3067216248

`BoundBox2d` and `BoundBox3d` in `cpp/modmesh/universe/rtree.hpp` are separate classes, but `BoundBox3d` can serve both 2D and 3D use cases. The semantic of "box" already implies applicability to both 2 and 3 dimensional space.

## Proposed Change

- Remove `BoundBox2d` and use `BoundBox3d` everywhere (with `z` defaulting to 0 for 2D cases).
- Update all usages in `World.hpp`, `rtree.hpp`, and related code that currently reference `BoundBox2d`.

## Files Affected

- `cpp/modmesh/universe/rtree.hpp` — `BoundBox2d` definition and `RTreeValueOps` specialization
- `cpp/modmesh/universe/World.hpp` — `bbox_type`, `ShapeEntry`, `compute_shape_bbox`, etc.

yungyuc:
Perpahs the issue was automated fixed by the branch name.  Please try not to name a branch as "fix" or something like that.

tigercosmos:
Interesting ... I don't know that branch name can also trigger auto-closing. lol

yungyuc:
> Interesting ... I don't know that branch name can also trigger auto-closing. lol

I did not know either.  Just a theory.  Try to avoid "fix" in branch name.

PR #738, opened 2026-04-19
Unify BoundBox2d and BoundBox3d into a single BoundBox3d

tigercosmos:
## Summary

`BoundBox2d` is redundant because a `BoundBox3d` with zero z-extent already represents a 2D box. This PR deletes it. Code that needed a 2D box now builds a `BoundBox3d` with `z = 0`.

One behavior change is that `calc_area()` now returns `dx*dy` for planar boxes and `dx*dy*dz` for volumetric boxes. The R-tree split heuristic requires this value to increase as boxes grow. That assumption still holds because each `RTree` in the codebase stores only one box type: `World` uses planar boxes, and `PolygonPad` uses volumetric boxes.

- Remove `BoundBox2d`; use `BoundBox3d` for both 2D (`z = 0`) and 3D.
- `calc_area()` → rectangle area for planar boxes, volume for volumetric boxes.
- Update `World.hpp` (`ShapeEntry`, `RTreeValueOps`, `bbox_type`, `compute_shape_bbox`, `query_visible`) and the RTree gtests to the unified API.

Related to #721.

## Test plan

- [x] `make gtest` — all 126 tests pass, including the 18 RTree cases.
- [x] `make buildext` — Python extension builds cleanly.

### DIFF
--- a/cpp/modmesh/universe/rtree.hpp
+++ b/cpp/modmesh/universe/rtree.hpp
@@ -42,74 +42,8 @@
 namespace modmesh
 {
 
-/// Bounding box for 2D objects
-/// @tparam T floating-point type
-template <typename T>
-class BoundBox2d : public NumberBase<int32_t, T>
-{
-
-public:
-    using value_type = T;
-
-private:
-    value_type m_min_x;
-    value_type m_min_y;
-    value_type m_max_x;
-    value_type m_max_y;
-
-public:
-
-    /// construct a bounding box with given min and max coordinates
-    BoundBox2d(value_type min_x_in, value_type min_y_in, value_type max_x_in, value_type max_y_in)
-        : m_min_x(min_x_in)
-        , m_min_y(min_y_in)
-        , m_max_x(max_x_in)
-        , m_max_y(max_y_in)
-    {
-    }
-
-    /// construct a bounding box that encloses two bounding boxes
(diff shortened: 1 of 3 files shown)


cpp/modmesh/universe/rtree.hpp
```diff
@@ -42,74 +42,8 @@
 namespace modmesh
 {
 
-/// Bounding box for 2D objects
-/// @tparam T floating-point type
-template <typename T>
-class BoundBox2d : public NumberBase<int32_t, T>
```
tigercosmos, line 48: No need `BoundBox2d` now.

yungyuc:
Please update the opening comment and add a short summary passage.  Bullets alone are not clear.

tigercosmos:
> Please update the opening comment and add a short summary passage. Bullets alone are not clear.

Sure, and I need to put it into the Claude instruction as well.

tigercosmos:
@yungyuc  It's ready for review. Please take a look. Thanks.

gtests/test_nopython_rtree.cpp
yungyuc, line 1: In the future, we will expose rtree to Python.  Then we can use and test it from Python.

There do not seem to be features that should be only available and tested in C++.

yungyuc (approved):
LGTM

Issue #646, opened 2025-12-05
Refine SIMD implementation


KHLee529:
# Abstract
According to the discussion in PR #641, there is an unintended recursive program flow in current SIMD implementation. Recursion is not proper in low level basic container especially for performance-critical functions.
This issue tracks the refinement of SIMD implementation and reconsider the overall structure of our SIMD library. 

## Tasks

- [x] Remove unintended recursion
- [ ] Discuss on SIMD implementation

## Problem on current SIMD implementation

Current SIMD implementation is based on functionality module. For example, `add`, `subtract`, `multiply`, `divide` are implemented in every different SIMD platform. In this form, the logic of each functionality are actually implemented multiple times in different part of the library. Even platform without SIMD support needs its own "generic" version of implementation. Comparing with other SIMD library, take [xsimd](https://github.com/xtensor-stack/xsimd) for instance, are modularized by the SIMD basic instructions such as load value to vector registry, store vector register to memory, arithmetic operations on two vector registers. In this way, the program logic of different SIMD platform can be unify and implement only once, which should be more maintainable and understandable.

KHLee529:
@yungyuc @nccvector @tigercosmos @tychuang1211 this issue is opened for the follow-up discussion of PR #641. Please share your opinion.

tigercosmos:
I agree with (1) and (2). However, for the SIMD API desgin, it's somehow hard to design now because we don't have actual use case yet.

KHLee529:
> I agree with (1) and (2). However, for the SIMD API desgin, it's somehow hard to design now because we don't have actual use case yet.

I don't really get it, what are the (1) and (2) you mentioned?

tigercosmos:
Ops, sorry, it means the task 1 and task 2 in the list.

yungyuc:
> * [ ]  Remove unintended recursion
> * [ ]  Discuss on SIMD implementation

Did you mean task 1 and task 2 above?

> Ops, sorry, it means the task 1 and task 2 in the list.

It will be easier to describe by quoting the parts.

yungyuc:
@KHLee529 Please update.

PR #12, opened 2020-11-14
Fix CI failures

yungyuc:
Changes included:

* Enable `PYTHON_EXECUTABLE` in `CMakeLists.txt`.
* Add suppression for false alarms with `clang-tidy` 11.
* Update pybind11 to v2.6.1.

This supplements the missing parts in #11

PR #740, opened 2026-04-19
Unify SIMD arithmetic under a shared transform_binary template

tigercosmos:
## Summary

Refs #646 (Task 2). The generic and NEON backends each had four near-identical loops for `add` / `sub` / `mul` / `div`. This PR collapses them into a single `transform_binary` per backend that takes the operation as an injected functor.

## What changed

In the generic backend, the four ops now pass `std::plus` / `std::minus` / `std::multiplies` / `std::divides` into `transform_binary`. In the NEON backend they pass `vec_add` / `vec_sub` / `vec_mul` / `vec_div` wrappers around `neon_alias`, and `std::invocable<VecOp, vec_t, vec_t>` routes types without a matching vector overload (e.g. `int64` for `vmulq`) to the scalar path at compile time. This replaces the ad-hoc `vector_lane > 2` and `is_floating_point_v` guards.

## Bugs fixed along the way

- **Sub-lane UB in NEON.** `ptr <= dest_end - N_lane` formed a pointer before the buffer when the input was shorter than one SIMD lane. The vector loop now runs a counted trip count (`blocks = (dest_end - dest) / N_lane`), which is UB-safe on sub-lane inputs and lowers to a `subs/b.ne` macro-op-fused back-edge on AArch64. The scalar remainder is inline instead of a recursive call into `generic::`. The same rewrite is applied to `check_between`.
- **`check_between` diagnostic.** The SIMD body checked the `>= max` mask first and only looked at `< min` if the first was empty, so a later too-large lane could hide an earlier too-small one. Both bounds are now inspected before picking the returned pointer.
- **`has_vectype` typing.** Declared `size_t`; retyped to `bool` to match its predicate role.

## Profiling

Verified on Apple M3 Pro (clang 17, `-O3 -DNDEBUG -mcpu=apple-m1`):

- `transform_binary` inlines fully — no functor calls, no extra moves, no spills. Hot loop is 5 instr/iter with one fused control-chain macro-op (`subs/b.ne`), matching the hand-written master baseline byte-for-byte in shape.
- Throughput on cache-resident buffers (n = 256 / 16384) is within ±2% of master across `add`/`sub`/`mul`/`div` × `float`/`int32`/`double`, well inside run-to-run noise. `mul<int64>` (scalar fallback) is byte-identical.
- The same counted-trip form on `check_between` measures ~1.8× faster than master in an inlined hot scan, because the new loop shape avoids a register-shuffle the post-incrementing form was triggering under inlining pressure.

## Tests

`tests/test_buffer_simd.py` pins `_simd_feature() == "NEON"` on aarch64 so a silent fallback to the scalar path cannot pass unnoticed. It then covers the int32 shape matrix (n = 1, 3, 4, 5, 8, 17) for `transform_binary`, the int64-mul SFINAE fallback, and float sub/mul/div with one block + tail. A new private `modmesh._modmesh._simd_feature()` binding exposes the runtime-detected backend.

## Follow-up

`simd::check_between` has inconsistent bound semantics across paths: the NEON SIMD body treats `value == max_val` as out-of-range, while the scalar fallback accepts it. Out of scope here; left for a separate change.

## Test plan

- [x] `make gtest`
- [x] `tests/test_buffer_simd.py` on aarch64
- [x] CI on Linux / macOS / aarch64

### DIFF
--- a/cpp/modmesh/buffer/pymod/buffer_pymod.cpp
+++ b/cpp/modmesh/buffer/pymod/buffer_pymod.cpp
@@ -28,11 +28,34 @@
 
 #include <modmesh/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
 
+#include <modmesh/simd/simd_support.hpp>
+
 namespace modmesh
 {
 
 namespace python
 {
+
+static char const * simd_feature_name()
+{
+    namespace simd_detail = modmesh::simd::detail;
+    switch (simd_detail::detect_simd())
+    {
+    case simd_detail::SIMD_NONE: return "NONE";
+    case simd_detail::SIMD_NEON: return "NEON";
+    case simd_detail::SIMD_SSE: return "SSE";
+    case simd_detail::SIMD_SSE2: return "SSE2";
+    case simd_detail::SIMD_SSE3: return "SSE3";
+    case simd_detail::SIMD_SSSE3: return "SSSE3";
+    case simd_detail::SIMD_SSE41: return "SSE41";
+    case simd_detail::SIMD_SSE42: return "SSE42";
+    case simd_detail::SIMD_AVX: return "AVX";
+    case simd_detail::SIMD_AVX2: return "AVX2";
+    case simd_detail::SIMD_AVX512: return "AVX512";
+    case simd_detail::SIMD_UNKNOWN: return "UNKNOWN";
+    }
+    return "UNKNOWN";
+}
 
 struct buffer_pymod_tag;
 
@@ -52,6 +75,14 @@ void initialize_buffer(pybind11::module & mod)
         wrap_ConcreteBuffer(mod);
         wrap_SimpleArray(mod);
         wrap_SimpleArrayPlex(mod);
+
+        // Reports the runtime-detected SIMD feature so pytest can verify that
+        // NEON dispatch is active on aarch64. Without this guard, a regression
+        // that silently routes everything to the scalar path would still pass
+        // every correctness check. Kept under an underscore-prefixed name
+        // because detect_simd() only meaningfully reflects the dispatched
+        // backend on aarch64 today; on other targets it would mislead users.
+        mod.def("_simd_feature", &simd_feature_name);
     };
 
     OneTimeInitializer<buffer_pymod_tag>::me()(mod, initialize_impl);
--- a/cpp/modmesh/simd/neon/neon.hpp
+++ b/cpp/modmesh/simd/neon/neon.hpp
@@ -28,6 +28,9 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <functional>
+#include <type_traits>
+
 #include <modmesh/simd/simd_generic.hpp>
 #include <modmesh/simd/neon/neon_type.hpp>
 #include <modmesh/simd/neon/neon_alias.hpp>
@@ -90,6 +93,114 @@ inline constexpr size_t get_recommended_alignment()
 } /* end namespace detail */
 
 #if defined(__aarch64__)
+
+// Each vec_op's `operator()` is expression-SFINAE'd on `decltype(v*q(a, b))`
+// so that `std::is_invocable_v<VecOp, V, V>` reports false for element types
+// that lack a NEON overload (e.g. vmulq has no int64 form, vdivq is float
+// only), routing those types to the scalar generic path at compile time.
+struct vec_add
+{
+    template <typename V>
+    auto operator()(V a, V b) const -> decltype(vaddq(a, b)) { return vaddq(a, b); }
+};
+struct vec_sub
+{
+    template <typename V>
+    auto operator()(V a, V b) const -> decltype(vsubq(a, b)) { return vsubq(a, b); }
+};
+struct vec_mul
+{
+    template <typename V>
+    auto operator()(V a, V b) const -> decltype(vmulq(a, b)) { return vmulq(a, b); }
+};
+struct vec_div
+{
+    template <typename V>
+    auto operator()(V a, V b) const -> decltype(vdivq(a, b)) { return vdivq(a, b); }
+};
+
+// The scalar remainder is an inline loop rather than a recursive call back
+// into `generic::transform_binary`: the old recursive form corrupted the
+// stack for inputs shorter than one SIMD lane (issue #635 / PR #641).
+template <typename T, typename ScalarOp, typename VecOp>
+void transform_binary(T * dest, T const * dest_end, T const * src1, T const * src2, ScalarOp scalar_op, VecOp vec_op)
+{
+    // The two checks are nested rather than combined with `||` because
+    // `if constexpr` does not suppress template substitution of its
+    // subexpressions; naming `type::vector_t<T>` in a single combined
+    // condition would instantiate it for element types without a NEON
+    // vector binding (e.g. bool, Complex<T>), breaking compilation.
+    if constexpr (!type::has_vectype<T>)
+    {
+        generic::transform_binary<T>(dest, dest_end, src1, src2, scalar_op);
+    }
+    else
+    {
+        using vec_t = type::vector_t<T>;
+        if constexpr (!std::is_invocable_v<VecOp, vec_t, vec_t>)
+        {
+            generic::transform_binary<T>(dest, dest_end, src1, src2, scalar_op);
+        }
+        else
+        {
+            constexpr size_t N_lane = type::vector_lane<T>;
+
+#ifndef NDEBUG
+            constexpr size_t alignment = detail::get_recommended_alignment();
+            detail::check_alignment(dest, alignment, "transform_binary dest");
+            detail::check_alignment(src1, alignment, "transform_binary src1");
+            detail::check_alignment(src2, alignment, "transform_binary src2");
+#endif
+
+            // Remaining-length compare avoids forming `dest_end - N_lane`,
+            // which is UB (pointer before the buffer) when the input is
+            // shorter than one SIMD lane. Both pointers lie inside the
+            // buffer so `dest_end - ptr` is well-defined.
+            T * ptr = dest;
+            while (static_cast<size_t>(dest_end - ptr) >= N_lane)
+            {
+                vec_t v1 = vld1q(src1);
+                vec_t v2 = vld1q(src2);
+                vst1q(ptr, vec_op(v1, v2));
+                ptr += N_lane;
+                src1 += N_lane;
+                src2 += N_lane;
+            }
+            while (ptr < dest_end)
+            {
+                *ptr = scalar_op(*src1, *src2);
+                ++ptr;
+                ++src1;
+                ++src2;
+            }
+        }
+    }
+}
+
+template <typename T>
+inline void add(T * dest, T const * dest_end, T const * src1, T const * src2)
+{
+    transform_binary<T>(dest, dest_end, src1, src2, std::plus<T>{}, vec_add{});
+}
+
+template <typename T>
+inline void sub(T * dest, T const * dest_end, T const * src1, T const * src2)
+{
+    transform_binary<T>(dest, dest_end, src1, src2, std::minus<T>{}, vec_sub{});
+}
+
+template <typename T>
+inline void mul(T * dest, T const * dest_end, T const * src1, T const * src2)
+{
+    transform_binary<T>(dest, dest_end, src1, src2, std::multiplies<T>{}, vec_mul{});
+}
+
+template <typename T>
+inline void div(T * dest, T const * dest_end, T const * src1, T const * src2)
+{
+    transform_binary<T>(dest, dest_end, src1, src2, std::divides<T>{}, vec_div{});
+}
+
 template <typename T, typename std::enable_if_t<!type::has_vectype<T>> * = nullptr>
 const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
 {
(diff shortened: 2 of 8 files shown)


cpp/modmesh/buffer/pymod/buffer_pymod.cpp
```diff
@@ -52,6 +80,14 @@ void initialize_buffer(pybind11::module & mod)
...
         wrap_SimpleArray(mod);
         wrap_SimpleArrayPlex(mod);
+
+        // Reports the runtime-detected SIMD feature so pytest can verify that
+        // NEON dispatch is active on aarch64. Without this guard, a regression
+        // that silently routes everything to the scalar path would still pass
+        // every correctness check. Kept under an underscore-prefixed name
+        // because detect_simd() only meaningfully reflects the dispatched
+        // backend on aarch64 today; on other targets it would mislead users.
+        mod.def("_simd_feature", &simd_feature_name);
```
tigercosmos, line 85: For checking if simd is working.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
+struct vec_add
 {
-    return generic::check_between<T>(start, end, min_val, max_val);
-}
-
-template <typename T, typename std::enable_if_t<type::has_vectype<T>> * = nullptr>
-const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
+    template <typename V>
+    static auto operator()(V a, V b) -> decltype(vaddq(a, b)) { return vaddq(a, b); }
+};
```
tigercosmos, line 102: Key design in this PR.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
-        return generic::add<T>(dest, dest_end, src1, src2);
+        generic::transform_binary<T>(dest, dest_end, src1, src2, scalar_op);
     }
     else
     {
         using vec_t = type::vector_t<T>;
-        constexpr size_t N_lane = type::vector_lane<T>;
+        if constexpr (!std::invocable<VecOp, vec_t, vec_t>)
+        {
+            generic::transform_binary<T>(dest, dest_end, src1, src2, scalar_op);
```
tigercosmos, line 131: T does have a vector type, but the specific VecOp functor can't be called with it. For example, `vdivq` doesn't exist for integer vector types in NEON, so `vec_div{}` isn't invocable with `int32x4_t`.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
-        {
-            generic::add<T>(ptr, dest_end, src1, src2);
+            // Compare on remaining length, not `dest_end - N_lane`: the latter
+            // forms a pointer before the buffer (UB) on sub-lane inputs.
+            T * ptr = dest;
+            while (static_cast<size_t>(dest_end - ptr) >= N_lane)
+            {
+                vec_t v1 = vld1q(src1);
+                vec_t v2 = vld1q(src2);
+                vst1q(ptr, vec_op(v1, v2));
```
tigercosmos, line 155: `vec_op` is called here.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
-template <typename T>
-void add(T * dest, T const * dest_end, T const * src1, T const * src2)
+template <typename T, std::invocable<T, T> ScalarOp, typename VecOp>
+void transform_binary(T * dest, T const * dest_end, T const * src1, T const * src2, ScalarOp scalar_op, VecOp vec_op)
 {
-    if constexpr (!(type::has_vectype<T>))
+    if constexpr (!type::has_vectype<T>)
     {
-        return generic::add<T>(dest, dest_end, src1, src2);
+        generic::transform_binary<T>(dest, dest_end, src1, src2, scalar_op);
```
tigercosmos, line 124: The scalar type T itself has no corresponding NEON vector type (e.g., bool, int64_t). There's no vector register representation at all, so SIMD is impossible.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
-        vec_t src1_vec;
-        vec_t src2_vec;
-        vec_t res_vec;
-        T * ptr = dest;
-        for (; ptr <= dest_end - N_lane; ptr += N_lane, src1 += N_lane, src2 += N_lane)
+        vec_t max_vec = vdupq(max_val);
+        vec_t min_vec = vdupq(min_val);
+
+        // Vector loop runs while a full lane still fits. The remaining-count
+        // form keeps the condition valid for buffers shorter than one lane.
```
tigercosmos: Fixed the previous UB issue.

cpp/modmesh/simd/neon/neon_type.hpp
```diff
@@ -30,8 +30,8 @@
 
 #if defined(__aarch64__)
 
-#include <cstddef>
 #include <arm_neon.h>
+#include <cstddef>
```
tigercosmos, line 34: Formattor fixes the order, I think it should be fine. Let me know if I should revert it.

cpp/modmesh/simd/neon/neon_type.hpp
```diff
@@ -133,7 +133,7 @@ template <typename T>
 inline constexpr size_t vector_lane = detail::vector<T>::N_lane;
 
 template <typename T>
-inline constexpr size_t has_vectype = detail::vector<T>::N_lane > 0;
+inline constexpr bool has_vectype = detail::vector<T>::N_lane > 0;
```
tigercosmos, line 136: Fixed the boolean type.

cpp/modmesh/simd/simd_generic.hpp
```diff
@@ -40,69 +43,51 @@ namespace generic
...
-    T * ptr = dest;
-    while (ptr < dest_end)
-    {
-        *ptr = *src1 - *src2;
-        ++ptr;
-        ++src1;
-        ++src2;
-    }
+    transform_binary<T>(dest, dest_end, src1, src2, std::plus<T>{});
 }
```
tigercosmos, line 73: Main design of this PR.

tests/test_buffer_simd.py
```diff
@@ -0,0 +1,99 @@
...
+    # Without this guard, missing NEON dispatch on aarch64 would silently route
+    # every SIMD operation to the scalar path -- correctness tests would still
+    # pass and the regression would be invisible.
+    def test_neon_active_on_aarch64(self):
+        # _simd_feature is intentionally private: the underlying detector only
+        # reflects the dispatched backend on aarch64 today, so it is reached
+        # through the C++ module rather than the public modmesh namespace.
+        feature = modmesh._modmesh._simd_feature()
+        if platform.machine() in ("arm64", "aarch64"):
+            self.assertEqual(feature, "NEON")
```
tigercosmos, line 46: Check if NEON is working that we didn't test it before.

tests/test_buffer_simd.py
```diff
@@ -0,0 +1,99 @@
...
+        else:
+            self.skipTest("_simd_feature() = " + feature)
+
+
+class SimdTransformBinaryTC(unittest.TestCase):
```
tigercosmos, line 51: Some cases for checking `transform_binary` functionality.

tigercosmos (commented):
@yungyuc  The PR is ready for review. Thanks!

yungyuc:
@KHLee529 Could you please take a look?

KHLee529:
The unified backend look nice in my first glance. I'll dive into details later.

cpp/modmesh/buffer/pymod/buffer_pymod.cpp
yungyuc: `cpp/modmesh/toggle/` may be a more on-topic module for the SIMD check, but it's fine to have it here in `buffer`.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
+        T const * ptr = start;
+        while (static_cast<size_t>(end - ptr) >= N_lane)
```
yungyuc: `while` reads clearer than `for`.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
+                }
+                return ptr;
+            }
+
+            ptr += N_lane;
         }
-        if (ptr != dest_end)
+
+        // Tail scalar loop for remaining elements
+        for (; ptr < end; ++ptr)
```
yungyuc: Why not using `while` too?

cpp/modmesh/simd/neon/neon_type.hpp
yungyuc: It's fine.

cpp/modmesh/simd/neon/neon_type.hpp
yungyuc: Good catch.

cpp/modmesh/simd/simd_generic.hpp
yungyuc: I am not sure if the additional abstraction still generates good SIMD binaries.  Please profile to check.  If you have time, also check the built assembly.

tests/test_buffer_simd.py
yungyuc, line 1: Since most tests are against `SimpleArray`, I suggest to name the new test file as `test_buffer_simd.py`?

yungyuc (changes requested):
- [x] Clarify if some `for` loops can also be replaced with `while`.
- [x] Run performance test to compare the runtime before and after the change. List the results to show that the change does not degrade runtime performance.
- [x] Rename `test_simd.py` to `test_buffer_simd.py`.  We can discuss which name is better.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
+// SFINAE helpers for vectorized operations.
+struct vec_add
 {
-    return generic::check_between<T>(start, end, min_val, max_val);
-}
-
-template <typename T, typename std::enable_if_t<type::has_vectype<T>> * = nullptr>
-const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
+    template <typename V>
+    static auto operator()(V a, V b) -> decltype(vaddq(a, b)) { return vaddq(a, b); }
```
KHLee529, line 101: Can these operator helper functions be also inlined? Based on my experience profiling the speed of `SimpleArray` SIMD operations, whether the vector operations are inlined impact a lot on the performance

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -90,216 +93,163 @@ inline constexpr size_t get_recommended_alignment()
...
+                vec_t v1 = vld1q(src1);
+                vec_t v2 = vld1q(src2);
+                vst1q(ptr, vec_op(v1, v2));
+                ptr += N_lane;
+                src1 += N_lane;
+                src2 += N_lane;
+            }
+            while (ptr < dest_end)
+            {
+                *ptr = scalar_op(*src1, *src2);
```
KHLee529, line 162: Nice way to remove dependency to `generic` functions.

cpp/modmesh/simd/simd_generic.hpp
```diff
@@ -40,69 +43,51 @@ namespace generic
 template <typename T>
 const T * check_between(T const * start, T const * end, T const & min_val, T const & max_val)
 {
-    T const * ptr = start;
-    while (ptr < end)
+    for (T const * ptr = start; ptr < end; ++ptr)
     {
-        T idx = *ptr;
-        if (idx < min_val || idx > max_val)
+        if (*ptr < min_val || *ptr > max_val)
```
KHLee529, line 48: Is this refinement potentially slower due to one more dereference execution?

tests/test_buffer_simd.py
KHLee529: Why do you isolate this unit test out from `test_buffer.py`?

KHLee529 (commented):
No change requested. Only some comments and questions listed.

tests/test_buffer_simd.py
tigercosmos: The whole SIMD implementation is also outside `buffer` directory. I think it worths a new file.

tigercosmos:
# PR #740 profiling — `transform_binary` is performance-neutral

Addressing the codegen/throughput question raised in the
[code review](https://github.com/solvcon/modmesh/pull/740#discussion_r3160569153).

**TL;DR.** The unified `transform_binary` template inlines cleanly,
the hot loop is a 5-instruction body with a fused
compare-and-branch, and microbenchmark throughput matches master to
within run-to-run noise.

## Hot-loop assembly (`simd_add_f32`, `clang -O3 -DNDEBUG -mcpu=apple-m1`)

**BEFORE** (`3d73a5d`, hand-written loop):
```
LBB0_1:
    ldr     q0, [x2], #16
    ldr     q1, [x3], #16
    fadd.4s v0, v0, v1
    str     q0, [x0], #16
    cmp     x0, x8        ; cmp + b.ls macro-op-fuses
    b.ls    LBB0_1
```

**AFTER_PR** (`8ab5f78`, `transform_binary` with counted trip count):
```
LBB0_2:
    ldr     q0, [x2], #16
    ldr     q1, [x3], #16
    fadd.4s v0, v0, v1
    str     q0, [x0], #16
    subs    x8, x8, #1     ; subs + b.ne macro-op-fuses
    b.ne    LBB0_2
```

5 instr/iter on both sides, one fused control-chain macro-op on both.
Same body for `sub`/`mul`/`div`/`add<int32>`/`mul<int32>`/`add<double>`.
`mul<int64>` (scalar fallback) is byte-identical.

## Throughput (Apple M3 Pro, Gelem/s, median of 3 runs)

`n = 16384` (L2-resident, the most stable measurement):

| Op | BEFORE | AFTER_PR | Δ |
|---|---:|---:|---:|
| `add<float>`  | 14.91 | 14.78 | −0.9% |
| `sub<float>`  | 14.93 | 14.75 | −1.2% |
| `mul<float>`  | 14.82 | 14.77 | −0.4% |
| `div<float>`  | 14.78 | 14.76 | −0.2% |
| `add<int32>`  | 14.87 | 14.71 | −1.0% |
| `mul<int32>`  | 14.88 | 14.71 | −1.2% |
| `mul<int64>`  |  3.19 |  3.18 | −0.3% |
| `add<double>` |  5.63 |  5.69 | +1.2% |

`n = 256` (L1-resident) is similarly flat (every op within ±2%).
`n = 262144` (DRAM-bound) is noisy in both directions on a
non-isolated machine.

## Note on the review history

An intermediate revision used `while (dest_end - ptr >= N_lane)` for
the bound check. That form is UB-safe but forces clang to emit a
non-flag-setting `sub` + `cmp #12` pair, which breaks macro-op fusion
on AArch64 and showed a real ~20–25% regression on cache-resident
loops. The current head replaces it with a counted trip count
(`blocks = (dest_end - dest) / N_lane`), which restores fusion. That
is what the numbers above measure.

cpp/modmesh/simd/simd_generic.hpp
tigercosmos: No — clang CSEs the two textual *ptr reads into a single load per iteration. Same one ldr in both versions:
```
  ; BEFORE                              ; AFTER                
  ldr  w10, [x0], #4                    ldr  w10, [x0]         
  cmp  w10, w8                          cmp  w10, w8           
  ccmp w10, w9, #0, ge                  ccmp w10, w9, #0, ge 
  b.le ...                              b.gt ...               
                                        add  x0, x0, #4
                                        cmp  x0, x1            
                                        b.lo ...             
```

No extra dereference. In fact, when this function inlines into a hot caller the new form is measurably faster (~1.8× on a tight scan loop, M3 Pro) — the for (...; ++ptr) shape decouples the load operand from the pointer-bump, which gives the register allocator more freedom and avoids a redundant  register-shuffle that the post incrementing `ldr [x0], #4` triggers under inlining pressure.

cpp/modmesh/simd/simd_generic.hpp
tigercosmos: > I am not sure if the additional abstraction still generates good SIMD binaries. Please profile to check. If you have time, also check the built assembly.

Profiled on Apple M3 Pro (clang 17, `-O3 -DNDEBUG -mcpu=apple-m1`). Both the assembly and the throughput look clean.

**Assembly.** The `transform_binary<T, ScalarOp, VecOp>` template inlines fully — no functor calls, no extra moves, no spills. Hot-loop body for `simd_add_f32`:

```
; BEFORE (master)                     ; AFTER (this PR)
ldr     q0, [x2], #16                  ldr     q0, [x2], #16
ldr     q1, [x3], #16                  ldr     q1, [x3], #16
fadd.4s v0, v0, v1                     fadd.4s v0, v0, v1
str     q0, [x0], #16                  str     q0, [x0], #16
cmp     x0, x8                         subs    x8, x8, #1
b.ls    LBB0_1                         b.ne    LBB0_2
```

5 instr/iter on both sides; one fused control-chain macro-op (`cmp/b.ls` ≡ `subs/b.ne` for fusion purposes). Same pattern for `sub`/`mul`/`div`/`add<int32>`/`mul<int32>`/`add<double>`. `mul<int64>` (scalar fallback via `std::invocable<VecOp, vec_t, vec_t>`) is byte-identical to master.

**Throughput** (Gelem/s, `n=16384`, L2-resident, median of 3 runs):

| Op | BEFORE | AFTER | Δ |
|---|---:|---:|---:|
| `add<float>`  | 14.91 | 14.78 | −0.9% |
| `mul<float>`  | 14.82 | 14.77 | −0.4% |
| `div<float>`  | 14.78 | 14.76 | −0.2% |
| `add<int32>`  | 14.87 | 14.71 | −1.0% |
| `add<double>` |  5.63 |  5.69 | +1.2% |

All ops within ±2% of master across L1/L2-resident sizes — well inside run-to-run noise. Full write-up + reproducer in `profiling/simd_pr740/`.

cpp/modmesh/simd/neon/neon.hpp
```diff
@@ -91,216 +94,170 @@ inline constexpr size_t get_recommended_alignment()
...
-        for (; ptr <= dest_end - N_lane; ptr += N_lane, src1 += N_lane, src2 += N_lane)
+        vec_t max_vec = vdupq(max_val);
+        vec_t min_vec = vdupq(min_val);
+
+        // Vector loop runs while a full lane still fits. Counted trip form
+        // for the same reason as transform_binary above: avoids UB on
+        // sub-lane inputs and the per-iter `sub` overhead.
+        size_t const blocks = static_cast<size_t>(end - start) / N_lane;
+        T const * ptr = start;
+        for (size_t block = 0; block < blocks; ++block)
```
tigercosmos, line 221: An intermediate revision used `while (dest_end - ptr >= N_lane)` for the bound check. That form is UB-safe but forces clang to emit a non-flag-setting `sub` + `cmp #12` pair, which breaks macro-op fusion on AArch64 and showed a real ~20–25% regression on cache-resident loops. The current head replaces it with a counted trip count (`blocks = (dest_end - dest) / N_lane`), which restores fusion. That is what the numbers above measure.

cpp/modmesh/simd/neon/neon.hpp
tigercosmos: They are implicit inlined, and confirmed by the profiling.

tigercosmos:
@yungyuc  Please take a look. All comments are addressed. Please take a look, thanks!

tests/test_buffer_simd.py
tigercosmos: done.

cpp/modmesh/simd/neon/neon.hpp
tigercosmos: done

cpp/modmesh/simd/neon/neon.hpp
yungyuc: I did not know that.  Good to learn.

cpp/modmesh/simd/simd_generic.hpp
yungyuc: > Profiled on Apple M3 Pro (clang 17, `-O3 -DNDEBUG -mcpu=apple-m1`). Both the assembly and the throughput look clean.

The profiling results and assembly look good.  But clang 17 looks old.  The latest version provided by xcode is version 21.0.0 (clang-2100.0.123.102).  Old compiler is OK since both before and after use the same version.

yungyuc (approved):
LGTM

PR #744, opened 2026-05-01
ci: remove ppa setting for ubuntu 24.04 runner

ExplorerRay:
According to [Ubuntu available GCC versions](https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/gcc/), Ubuntu 24.04 supports gcc-14 natively without running `sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y`.

Therefore, I remove this line to avoid potential timeout failure when adding PPA.

### DIFF
--- a/.github/actions/setup_linux/action.yml
+++ b/.github/actions/setup_linux/action.yml
@@ -86,7 +86,6 @@ runs:
     - name: Install and configure gcc-14
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


yungyuc (approved):
Good catch and it's great to be able to fix the setting at one place for all Ubuntu jobs.

Merging.

PR #751, opened 2026-05-04
Prevent duplicate CI runs when pushing with a pull request

yungyuc:
Restrict push trigger to run only on branches listed in the repository variable `MMGH_PUSH_RUN_BRANCH`.  If the variable is not set, do not trigger CI with the push events.  Pull request and schedule events are unaffected.

Add a concurrency group per workflow and PR number (or ref) so that a new push automatically cancels any still-running jobs for the same PR.

This is to save resources.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -6,11 +6,19 @@ on:
   schedule:
     - cron: '34 17 * * *'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   standalone_buffer:
 
-    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
+    if: |
+      github.event_name == 'pull_request' ||
+      (github.event_name == 'push' && vars.MMGH_PUSH_RUN_BRANCH != '' &&
+       contains(vars.MMGH_PUSH_RUN_BRANCH, github.ref_name)) ||
+      (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable')
 
     name: standalone_buffer_${{ matrix.os }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -6,11 +6,19 @@ on:
   schedule:
     - cron: '34 17 * * *'
 
+concurrency:
```
yungyuc, line 9: A new push automatically cancels any still-running jobs for the same PR.

.github/workflows/devbuild.yml
```diff
@@ -6,11 +6,19 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
 
   standalone_buffer:
 
-    if: ${{ github.event_name != 'schedule' || (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') }}
+    if: |
+      github.event_name == 'pull_request' ||
+      (github.event_name == 'push' && vars.MMGH_PUSH_RUN_BRANCH != '' &&
```
yungyuc, line 19: Restrict push trigger to run only on branches listed in the repository variable `MMGH_PUSH_RUN_BRANCH`.  If the variable is not set, do not trigger CI with the push events.  Pull request and schedule events are unaffected.

Other events do not trigger the CI.

yungyuc:
@ExplorerRay Please take a look.

ExplorerRay:
I think there should be a comment showing that how to set `MMGH_PUSH_RUN_BRANCH` to trigger CI when push.
After reading the code for a while, I found that this variable should set a string like `"master,ci/dedup"` to run CI on these two branches. I think it's not that intuitive instead of just a true of false.

yungyuc:
> I think there should be a comment showing that how to set `MMGH_PUSH_RUN_BRANCH` to trigger CI when push. After reading the code for a while, I found that this variable should set a string like `"master,ci/dedup"` to run CI on these two branches. I think it's not that intuitive instead of just a true of false.

Agree.  Could you please create a PR to add the comment.

I merged this one quickly so that subsequent PRs won't have the duplicated runs.

ExplorerRay:
BTW, with this setting, every developer who forks this repository will be required to set MMGH_PUSH_RUN_BRANCH. Otherwise, they won’t be able to run any CI on their forks before opening a PR.

In my case, I usually rely on CI in my fork to verify that my code works before creating a PR. With this setup, I would need to update the MMGH_PUSH_RUN_BRANCH variable every time I create a new branch.

I’m curious about the scenario that led you to disable all CI runs by default unless a PR is opened. Perhaps we can find an alternative condition that satisfies both your needs and mine.

yungyuc:
> BTW, with this setting, every developer who forks this repository will be required to set MMGH_PUSH_RUN_BRANCH. Otherwise, they won’t be able to run any CI on their forks before opening a PR.

Yes but not entirely true, assuming "opening a PR" means "opening a PR to the upstream".  Developers can create a PR to their own fork to kick off a CI run.

> In my case, I usually rely on CI in my fork to verify that my code works before creating a PR. With this setup, I would need to update the MMGH_PUSH_RUN_BRANCH variable every time I create a new branch.

We can add wildcard (`*`) support in `MMGH_PUSH_RUN_BRANCH`.

> I’m curious about the scenario that led you to disable all CI runs by default unless a PR is opened. Perhaps we can find an alternative condition that satisfies both your needs and mine.

I create PR to my fork as part an automation flow and the duplicated runs elongate the development turnaround time.  The duplicated runs also waste CI resources.

PR #752, opened 2026-05-04
Move menu management from C++ to Python

yungyuc:
Work includes:
* Wrap Qt menu management in `RMenu` (new class) for Python access
* Populate the menu in Python instead.

`RMenu` extends Qt's `QMenu` and accepts Python callables directly. It eliminate the need for Python code to create QAction objects via PySide6.

`RManager` is changed to return `RMenu *` instead of `QMenu *`.

On the Python side, `_gui_common._add_menu_item()` calls `RMenu.add_action()`, and no longer imports `QtGui`. Associated helpers are rewritten. All of the 7 menus are changed to be managed in Python: File, View, One, Mesh, Canvas, Profiling, Window.


cpp/modmesh/pilot/RManager.hpp
```diff
@@ -90,24 +100,21 @@ class RManager
...
+    QMenuBar * m_menuBar = nullptr;
 
-    QMenu * m_fileMenu = nullptr;
-    QMenu * m_viewMenu = nullptr;
-    QMenu * m_oneMenu = nullptr;
-    QMenu * m_meshMenu = nullptr;
-    QMenu * m_canvasMenu = nullptr;
-    QMenu * m_profilingMenu = nullptr;
-    QMenu * m_windowMenu = nullptr;
+    std::unordered_map<std::string, RMenu *> m_menus;
```
yungyuc, line 117: Use a container (`std::unordered_map`) for all menus.

cpp/modmesh/pilot/RMenu.hpp
```diff
@@ -0,0 +1,72 @@
...
+ */
+class RMenu
+    : public QMenu
+{
+    Q_OBJECT
+
+public:
+
+    using QMenu::addAction; // Keep all QMenu overloads visible
+    using QMenu::QMenu; // Inherit constructors
```
yungyuc, line 59: `clang-format` ask `using QMenu:QMenu` to be after `using QMenu::addAction` but I don't know why.  It does not hurt to follow the tool though.

cpp/modmesh/pilot/wrap_pilot.cpp
```diff
@@ -267,6 +267,55 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRPythonConsoleDockWidget
...
+                "add_action",
+                [](wrapped_type & self,
+                   std::string const & text,
+                   std::string const & tip,
+                   py::object func,
+                   bool checkable,
+                   bool checked)
+                {
+                    auto func_ = [func]()
+                    { func(); };
```
yungyuc, line 293: Formatter forces to add a line here.

cpp/modmesh/pilot/wrap_pilot.cpp
```diff
@@ -339,19 +388,61 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapRManager
         namespace py = pybind11;
 
         (*this)
+            .def(
```
yungyuc, line 391: The function `wrap_mainWindow()` becomes too long and the menu wrappers should use a standalone `wrap_menu()`.

At the same time, these long functions should be moved outside the class declaration.  Many lines of code need to be moved and should use a distinct PR, not this one.

modmesh/pilot/_gui_common.py
```diff
@@ -26,7 +26,7 @@
 
 
 from . import _pilot_core as _pcore
-from PySide6 import QtCore, QtGui
+from PySide6 import QtCore
```
yungyuc, line 29: This should be before the above line and use a blank line to separate.  But to keep the patch clean (this line of diff simply dropped `QtGui`), the reformatting should use another PR.

yungyuc:
@c1ydehhx @tigercosmos Please take a look

Issue #753, opened 2026-05-05
Support wildcard in `MMGH_PUSH_RUN_BRANCH`


yungyuc:
PR #751 introduced `MMGH_PUSH_RUN_BRANCH` to control the branch to allow CI runs on push.  It currently supports only exact match of the branch name.

It should be enhanced to support the wildcard character `*`.  When `MMGH_PUSH_RUN_BRANCH` has `*` and only has `*` in, CI runs are triggered on pushing to any branch.

Discussion: https://github.com/solvcon/modmesh/pull/751#issuecomment-4372083525

yungyuc:
Fixed with PR #757

PR #757, opened 2026-05-05
Allow `MMGH_PUSH_RUN_BRANCH=*` to enable CI on every push

yungyuc:
Add a wildcard short-circuit when the variable `MMGH_PUSH_RUN_BRANCH` is set to exactly `*`, the push event matches regardless of branch name.  Any other value keeps the existing substring-contains behaviour.

This is for issue #753

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -17,7 +17,8 @@ jobs:
     if: |
       github.event_name == 'pull_request' ||
       (github.event_name == 'push' && vars.MMGH_PUSH_RUN_BRANCH != '' &&
-       contains(vars.MMGH_PUSH_RUN_BRANCH, github.ref_name)) ||
+       (vars.MMGH_PUSH_RUN_BRANCH == '*' ||
+        contains(vars.MMGH_PUSH_RUN_BRANCH, github.ref_name))) ||
       (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable')
 
     name: standalone_buffer_${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -17,7 +17,8 @@ jobs:
     if: |
       github.event_name == 'pull_request' ||
       (github.event_name == 'push' && vars.MMGH_PUSH_RUN_BRANCH != '' &&
-       contains(vars.MMGH_PUSH_RUN_BRANCH, github.ref_name)) ||
+       (vars.MMGH_PUSH_RUN_BRANCH == '*' ||
```
yungyuc, line 20: If it's (exactly) the wildcard short-circuiting, trigger the CI.

yungyuc:
@ExplorerRay please take a look.

.github/workflows/devbuild.yml
```diff
@@ -17,7 +17,8 @@ jobs:
     if: |
       github.event_name == 'pull_request' ||
       (github.event_name == 'push' && vars.MMGH_PUSH_RUN_BRANCH != '' &&
```
ExplorerRay: Does `vars.MMGH_PUSH_RUN_BRANCH != ''` need to be here when we check the conditions below?

yungyuc: Yeah.  If `vars.MMGH_PUSH_RUN_BRANCH` is not set, i.e., empty `''`, CI should not trigger.

ExplorerRay: If `vars.MMGH_PUSH_RUN_BRANCH` is not set, the remaining condition checks (=* or contains ref name) can help handle this and let it not trigger CI.

yungyuc: You are right.  Thanks.  The condition `vars.MMGH_PUSH_RUN_BRANCH != ''` is redundant.  Removed.

ExplorerRay (approved):
LGTM.

PR #758, opened 2026-05-05
Add a script to build development environment in macos VM

yungyuc:
Automate setup of a modmesh development environment on a fresh macOS VM.  The automation speeds up provisioning of an UTM VM on mac.

Usage:
```
bash mac26_vmsetup.sh [step]
```

Without arguments, runs all steps in order. Read the code to learn the steps available.

The current implementation is experimental.  Enhancements are expected to come in the future.

### DIFF
--- a/contrib/vm/macos/mac26_vmsetup.sh
+++ b/contrib/vm/macos/mac26_vmsetup.sh
@@ -1,0 +1,161 @@
+#!/usr/bin/env bash
+#
+# Automate setup of a modmesh development environment on a fresh macOS VM.
+#
+# Usage:
+#   bash mac26_vmsetup.sh [step]
+#
+# Without arguments, runs all steps in order. Read the code to learn the steps
+# available.
+#
+# Convenient tools to install manually before running the script:
+# * https://macvim.org/
+# * https://iterm2.com/
+#
+# Before or after running the script, create the ssh private key:
+# ```
+# ssh-keygen
+# ```
+#
+# When running the script, the two steps, xcode-select install dialog and chsh
+# password, are interactive. They pop up windows and prompt for password.
+#
+# After all packages and tools are installed, agent tools will need
+# authentication before use:
+# * `gh auth login`
+# * `claude`
+#
+# In a VM, automatic detection of available cores may not be accurate. Use a
+# setup.mk file in the modmesh repository root with the explicit process
+# number:
+# ```
+# MAKE_PARALLEL ?= -j4
+# ```
+#
+# Also make a symbolic link for Claude Code: `ln -s contrib/prompt/CLAUDE.md`
+
+set -euo pipefail
+
+WORKSPACE_REPO="https://github.com/yungyuc/workspace"
+MODMESH_REPO="git@github.com:yaagent/modmesh.git"
+MWORK_DIR="${HOME}/mwork"
+MODMESH_DIR="${MWORK_DIR}/modmesh"
+
+log() { printf '\n==> %s\n' "$*"; }
+warn() { printf '\n!!  %s\n' "$*" >&2; }
+
+require_macos() {
+    if [[ "$(uname -s)" != "Darwin" ]]; then
+        warn "This script is intended for macOS only."
+        exit 1
+    fi
+}
+
+step_keyboard() {
+    log "Setting key repeat faster (require re-login)"
+    # normal minimum is 15 (225 ms)
+    defaults write -g InitialKeyRepeat -float 10.0
+    # normal minimum is 2 (30 ms)
+    defaults write -g KeyRepeat -float 1.0
+}
+
+step_shell() {
+    log "Setting login shell to /bin/bash"
+    if [[ "${SHELL:-}" == "/bin/bash" ]]; then
+        echo "Login shell is already /bin/bash"
+    else
+        chsh -s /bin/bash
+    fi
+}
+
+step_workspace() {
+    log "Making the modmesh working directory"
+    mkdir -p "${MWORK_DIR}"
+    mkdir -p "${HOME}/tmp"
+}
+
+step_cli_tools() {
+    log "Installing Apple Command Line Developer Tools"
+    if xcode-select -p >/dev/null 2>&1; then
+        echo "Command Line Tools already installed at: $(xcode-select -p)"
+    else
+        xcode-select --install || true
+        echo "A GUI dialog should appear. Complete the installation, then re-run this script."
+        echo "Waiting for installation to finish..."
+        until xcode-select -p >/dev/null 2>&1; do
+            sleep 10
+        done
+    fi
+}
+
+step_homebrew() {
+    log "Installing Homebrew"
+    if command -v brew >/dev/null 2>&1; then
+        echo "brew already installed at: $(command -v brew)"
+    else
+        /bin/bash -c \
+            "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+    fi
+
+    # Make brew available in this shell (Apple Silicon vs Intel paths).
+    if [[ -x /opt/homebrew/bin/brew ]]; then
+        eval "$(/opt/homebrew/bin/brew shellenv)"
+    elif [[ -x /usr/local/bin/brew ]]; then
+        eval "$(/usr/local/bin/brew shellenv)"
+    fi
+}
+
+step_dependency() {
+    log "Installing modmesh build/runtime dependencies via Homebrew"
+    brew install \
+        tree gh cmake clang-format qt pyside \
+        python numpy python-matplotlib pybind11 \
+        pytest flake8 black
+
+    log "Installing extra Python packages"
+    pip3 install jsonschema pyyaml --break-system-packages
+}
+
+step_claude() {
+    log "Installing Claude CLI"
+    if command -v claude >/dev/null 2>&1; then
+        echo "claude already installed at: $(command -v claude)"
+    else
+        curl -fsSL https://claude.ai/install.sh | bash
+    fi
+}
+
+run_all() {
+    step_keyboard
+    step_shell
+    step_workspace
+    step_cli_tools
+    step_homebrew
+    step_dependency
+    step_claude
+    log "All steps complete."
+}
+
+main() {
+    require_macos
+    local target="${1:-all}"
+    case "${target}" in
+        all) run_all ;;
+        keyboard) step_keyboard ;;
+        shell) step_shell ;;
+        workspace) step_workspace ;;
+        cli-tools) step_cli_tools ;;
+        homebrew) step_homebrew ;;
+        dependency) step_dependency ;;
+        claude) step_claude ;;
+        *)
+            warn "Unknown step: ${target}"
+            echo "Valid: all, keyboard, shell, workspace, cli-tools, homebrew, dependency, claude"
+            exit 2
+            ;;
+    esac
+}
+
+main "$@"
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/contrib/vm/macos/mac26_vmsetup.sh
+++ b/contrib/vm/macos/mac26_vmsetup.sh
@@ -33,13 +33,16 @@
 # ```
 #
 # Also make a symbolic link for Claude Code: `ln -s contrib/prompt/CLAUDE.md`
+#
+# Additional information.  The following commands speed up key repeat:
+# ```
+# defaults write -g InitialKeyRepeat -float 10.0 # normal minimum is 15 (225 ms)
+# defaults write -g KeyRepeat -float 1.0 # normal minimum is 2 (30 ms)
+# ```
 
 set -euo pipefail
 
-WORKSPACE_REPO="https://github.com/yungyuc/workspace"
-MODMESH_REPO="git@github.com:yaagent/modmesh.git"
 MWORK_DIR="${HOME}/mwork"
-MODMESH_DIR="${MWORK_DIR}/modmesh"
 
 log() { printf '\n==> %s\n' "$*"; }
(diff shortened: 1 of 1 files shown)


contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+# Without arguments, runs all steps in order. Read the code to learn the steps
+# available.
+#
+# Convenient tools to install manually before running the script:
+# * https://macvim.org/
+# * https://iterm2.com/
+#
+# Before or after running the script, create the ssh private key:
+# ```
+# ssh-keygen
```
yungyuc, line 17: SSH is needed for github automation.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+# ```
+#
+# When running the script, the two steps, xcode-select install dialog and chsh
+# password, are interactive. They pop up windows and prompt for password.
+#
+# After all packages and tools are installed, agent tools will need
+# authentication before use:
+# * `gh auth login`
```
yungyuc, line 25: Agent automation with GitHub support.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+# * `claude`
+#
+# In a VM, automatic detection of available cores may not be accurate. Use a
+# setup.mk file in the modmesh repository root with the explicit process
+# number:
+# ```
+# MAKE_PARALLEL ?= -j4
```
yungyuc, line 32: Number of process for building should be added manually and vary by VM setup.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+    if [[ "${SHELL:-}" == "/bin/bash" ]]; then
+        echo "Login shell is already /bin/bash"
+    else
+        chsh -s /bin/bash
+    fi
+}
+
+step_workspace() {
+    log "Making the modmesh working directory"
+    mkdir -p "${MWORK_DIR}"
```
yungyuc, line 68: Create directories for later use.  Do not clone in root.  An additional level of directory makes it easier to do other things with the VM.

yungyuc:
@chestercheng @xingularity please help take a look

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+# ```
+#
+# Also make a symbolic link for Claude Code: `ln -s contrib/prompt/CLAUDE.md`
+
+set -euo pipefail
+
+WORKSPACE_REPO="https://github.com/yungyuc/workspace"
+MODMESH_REPO="git@github.com:yaagent/modmesh.git"
+MWORK_DIR="${HOME}/mwork"
+MODMESH_DIR="${MWORK_DIR}/modmesh"
```
tigercosmos: These three constants (`WORKSPACE_REPO`, `MODMESH_REPO`, `MODMESH_DIR`) are declared but never referenced anywhere in the script. Either wire them up — e.g., `git clone "${MODMESH_REPO}" "${MODMESH_DIR}"` inside `step_workspace` — or drop them so they don't bit-rot.

Also note `MODMESH_REPO` points to `git@github.com:yaagent/modmesh.git` (a fork), not `solvcon/modmesh`. If cloning is later enabled, double-check that's intentional.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+warn() { printf '\n!!  %s\n' "$*" >&2; }
+
+require_macos() {
+    if [[ "$(uname -s)" != "Darwin" ]]; then
+        warn "This script is intended for macOS only."
+        exit 1
+    fi
+}
+
+step_keyboard() {
```
tigercosmos: macOS ships system bash 3.2 (Apple won't ship newer due to GPLv3), and the default shell on modern macOS is zsh. Switching to `/bin/bash` is a strong opinion — worth confirming it's intentional rather than personal preference.

Minor: `chsh -s /bin/bash` will fail if `/bin/bash` isn't listed in `/etc/shells`. A `grep -q '^/bin/bash$' /etc/shells` guard would give a friendlier error.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+    log "Making the modmesh working directory"
+    mkdir -p "${MWORK_DIR}"
+    mkdir -p "${HOME}/tmp"
+}
+
+step_cli_tools() {
+    log "Installing Apple Command Line Developer Tools"
+    if xcode-select -p >/dev/null 2>&1; then
+        echo "Command Line Tools already installed at: $(xcode-select -p)"
+    else
```
tigercosmos, line 76: This `until … sleep 10` loop has no timeout. If the user dismisses the GUI dialog, loses network, or the install otherwise stalls, the script hangs forever with no signal to the operator.

Suggest a max-wait (e.g., 30 minutes) with a clear bailout message:

```bash
local waited=0
until xcode-select -p >/dev/null 2>&1; do
    sleep 10
    waited=$((waited + 10))
    if (( waited >= 1800 )); then
        warn "Timed out waiting for Command Line Tools install."
        exit 1
    fi
done
```

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+            sleep 10
+        done
+    fi
+}
+
+step_homebrew() {
+    log "Installing Homebrew"
+    if command -v brew >/dev/null 2>&1; then
+        echo "brew already installed at: $(command -v brew)"
+    else
```
tigercosmos, line 96: Good handling of Apple Silicon vs Intel paths. One gap: if the install ran but neither path exists (rare, but possible if the installer was interrupted), the next `brew install` fails with a confusing "command not found."

A final guard would surface the real failure:

```bash
command -v brew >/dev/null || { warn "brew not on PATH after install"; exit 1; }
```

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+        /bin/bash -c \
+            "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+    fi
+
+    # Make brew available in this shell (Apple Silicon vs Intel paths).
+    if [[ -x /opt/homebrew/bin/brew ]]; then
+        eval "$(/opt/homebrew/bin/brew shellenv)"
+    elif [[ -x /usr/local/bin/brew ]]; then
+        eval "$(/usr/local/bin/brew shellenv)"
```
tigercosmos, line 105: Two small concerns:

1. With `set -e`, a single transient brew failure (network, tap update) kills the whole script and forces a manual re-run. Splitting into groups or tolerating individual failures with a summary at the end would make partial failures recoverable.
2. `pyside` is unpinned — Homebrew's `pyside` formula currently ships PySide6, but a future major version bump will silently follow. Worth either pinning (`pyside@6` if/when it exists) or adding a comment noting the assumption.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+    fi
+}
+
+step_dependency() {
```
tigercosmos, line 110: `--break-system-packages` bypasses PEP 668 protections. This aligns with CLAUDE.md's "no venv" stance for modmesh, so it's appropriate here — but a one-line comment explaining the rationale would help future readers (and avoid a well-meaning PR removing it).

tigercosmos (commented):
Nice provisioning script — well-documented header, idempotent steps, and clean dispatcher. Left a few inline comments on dead code, an unbounded wait loop, and a couple of robustness nits. Nothing blocking.

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+
+log() { printf '\n==> %s\n' "$*"; }
+warn() { printf '\n!!  %s\n' "$*" >&2; }
+
+require_macos() {
+    if [[ "$(uname -s)" != "Darwin" ]]; then
+        warn "This script is intended for macOS only."
+        exit 1
+    fi
+}
```
tigercosmos, line 55: Add a `require_vm` guard alongside `require_macos`. The env var `MAC26_VMSETUP_FORCE=1` is the explicit escape hatch for users who really do want to run it on a host (e.g., dedicated build machine).

```suggestion
require_macos() {
    if [[ "$(uname -s)" != "Darwin" ]]; then
        warn "This script is intended for macOS only."
        exit 1
    fi
}

require_vm() {
    # Refuse to run on a host Mac. kern.hv_vmm_present=1 when the kernel
    # itself runs under a hypervisor (UTM/Virtualization.framework,
    # QEMU+HVF, etc.). Bare-metal reports 0 even if VMs are running on top.
    local vmm
    vmm="$(sysctl -n kern.hv_vmm_present 2>/dev/null || echo 0)"
    if [[ "${vmm}" != "1" ]]; then
        if [[ "${MAC26_VMSETUP_FORCE:-0}" != "1" ]]; then
            warn "This script is intended for a fresh macOS VM."
            warn "It mutates keyboard defaults, login shell, and system brew packages."
            warn "If you really want to run it on this host, re-run with:"
            warn "    MAC26_VMSETUP_FORCE=1 bash $0 $*"
            exit 1
        fi
        warn "MAC26_VMSETUP_FORCE=1 set — proceeding on host Mac at user's risk."
    fi
}
```

contrib/vm/macos/mac26_vmsetup.sh
```diff
@@ -0,0 +1,161 @@
...
+    step_workspace
+    step_cli_tools
+    step_homebrew
+    step_dependency
+    step_claude
+    log "All steps complete."
+}
+
+main() {
+    require_macos
```
tigercosmos, line 144: Wire the new guard in. Passing `"$@"` lets the warning echo back the original step argument so the user can copy-paste the suggested override command verbatim.

```suggestion
    require_macos
    require_vm "$@"
```

tigercosmos (commented):
Follow-up suggestion: add a host-vs-VM guard so accidentally running this on a working Mac doesn't trash its dev environment. `kern.hv_vmm_present` reliably distinguishes them — `1` inside UTM/Virtualization.framework, `0` on bare metal (verified on an Apple Silicon host).

tigercosmos:
Maybe we can also have a `contrib/vm/macos/host_bootstrap_mac_vm.sh` to set up a new macOS vm using UTM from scratch.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: Good catch.  I forgot to remove the unused variables.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: The script can also be used for a non-VM workstation.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: This review comment is misplaced.  It's intentional to use bash.  Bash is my personally preference, but opinions are welcome.  As long as the automation does not remove bash, it works for me.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: Good idea.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: Good catch.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: > Two small concerns:
> 
> 1. With `set -e`, a single transient brew failure (network, tap update) kills the whole script and forces a manual re-run. Splitting into groups or tolerating individual failures with a summary at the end would make partial failures recoverable.

It can be a future enhancement.

> 2. `pyside` is unpinned — Homebrew's `pyside` formula currently ships PySide6, but a future major version bump will silently follow. Worth either pinning (`pyside@6` if/when it exists) or adding a comment noting the assumption.

All packages should be upgrade per the default setting, unless we find an issue in the future.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: Point taken.

contrib/vm/macos/mac26_vmsetup.sh
yungyuc: It's intentional to not `require_vm`.

yungyuc (commented):
Points are taken. @tigercosmos please take a look.

> Follow-up suggestion: add a host-vs-VM guard so accidentally running this on a working Mac doesn't trash its dev environment. `kern.hv_vmm_present` reliably distinguishes them — `1` inside UTM/Virtualization.framework, `0` on bare metal (verified on an Apple Silicon host).

I think it will be good to warn rather than guard.  It can be an enhancement in the future.

> Maybe we can also have a `contrib/vm/macos/host_bootstrap_mac_vm.sh` to set up a new macOS vm using UTM from scratch.

It's possible, but not very useful IMHO.  Setting up macos VM needs some manual clicking.  I think they can be automated, but users will need to tweak the settings anyway.

And setting up the VM also needs to download the ipfw from online.  UTM does that automatically.  It takes much less time to set up a template VM and clone it for additional VM.  The template VM takes 25 GB of disk space, but it worths.

Issue #756, opened 2026-05-05
`test_pilot_canvas.py` needs a better name


yungyuc:
`test_pilot_canvas.py` tests code in `universe`, and should be renamed as a new `test_universe_*.py`, or integrated into one of the existing `test_universe_*.py`.

ThreeMonth03:
Could I give it a try?

yungyuc:
Thank you.  Assigned.

yungyuc:
Completed by PR #759

Issue #689, opened 2026-03-07
Copyright notice in file header


tigercosmos:
Per https://github.com/solvcon/modmesh/pull/683#discussion_r2896503272, we currently only allow one name in the copyright notice, which does not accurately reflect the actual contribution of the file. I propose that we can have multiple authors in a file.

yungyuc:
Thank you for bringing it up, @tigercosmos .  The purpose of the copyright notice in the file header is to remind of the license, which requires to have a copyright holder.  The convention was to have only one copyright holder in the header file.

I did not think it was important to accurately list all contributors in the file header, since the git history should suffice.  But is does not hurt to be rigorous.

If we want to be rigorous, a clear rule of listing authors should be written first.

tigercosmos:
We have https://github.com/solvcon/solvcon/blob/master/CONTRIBUTORS.md now.

PR #759, opened 2026-05-06
Refactor pilot/canvas.py and rename tests/test_pilot_canvas.py

ThreeMonth03:
This pull request is to make `test_pilot_canvas.py` naming better, see issue #756 . 

@yungyuc suggests that it could be renamed as `test_universe_*.py`, because the unittest checks whether `BezierSampler` in `Canvas` helper works as usual, which is more related to `modmesh.world` than `pilot`.

From my perspective, this file tests `Canvas` helper class, so I rename this file as  `test_universe_canvas.py`. I considerd the name `test_universe_pilot_helpers.py`, but this pull request doesn't deal with the other similar test file like `tests/test_pilot_airfoil.py `. Therefore, I decide to explicitly show the `Canvas` in filename.

### DIFF
--- a/modmesh/pilot/__init__.py
+++ b/modmesh/pilot/__init__.py
@@ -48,7 +48,7 @@ if enable:
         launch,
     )
     from . import airfoil  # noqa: F401
-    from . import _canvas  # noqa: F401
+    from . import _canvas_gui  # noqa: F401
 
 # NOTE: intentionally omit __all__ for now
 
--- a/modmesh/pilot/_canvas_gui.py
+++ b/modmesh/pilot/_canvas_gui.py
@@ -1,0 +1,182 @@
+# Copyright (c) 2026, Han-Xuan Huang <c1ydehhx@gmail.com>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of the copyright holder nor the names of its contributors
+#   may be used to endorse or promote products derived from this software
+#   without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""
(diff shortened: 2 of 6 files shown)


ThreeMonth03 (commented):
@yungyuc Please take a look, thanks. I don't know why I cannot leave the comment for file change.

yungyuc (changes requested):
The issue #759 description (see quote below) implied that it is not simply renaming, but also also includes the work to refactor the [`_canvas`](https://github.com/solvcon/modmesh/blob/master/modmesh/pilot/_canvas.py) code.  That is why it may be renamed or integrated, depending on the refactoring implementation.

> `test_pilot_canvas.py` tests code in `universe`, and should be renamed as a new `test_universe_*.py`, or integrated into one of the existing `test_universe_*.py`.

Most code in the `_canvas` module involves the class `World` exposed to Python (from the C++ geometry module `universe`), e.g., [`CurveSampler.draw_cbc()`](https://github.com/solvcon/modmesh/blob/3142aec7ec37f1024bd51a09dae9f193366ead46/modmesh/pilot/_canvas.py#L215-L255):

```python
    def draw_cbc(self, spacing=0.01):
        """
        Draw sampled points as a cubic Bezier chain.

        spacing is the target chord-length step used to choose per-segment
        Bezier sampling density. Smaller spacing produces denser rendering.
        """
        if self.points is None:
            raise RuntimeError(
                "populate_points() must be called before draw_cbc()"
            )
        if spacing <= 0:
            raise ValueError("spacing must be positive")
        if len(self.points) < 2:
            return

        point_type = core.Point3dFp64
        point_x = self.points.x.ndarray
        point_y = self.points.y.ndarray
        segment_length = np.hypot(
            point_x[:-1] - point_x[1:],
            point_y[:-1] - point_y[1:],
        )
        # Minimum of 2 so very short segments are still visible.
        nsample = np.maximum((segment_length // spacing).astype(int) - 1, 2)

        for index in range(len(self.points) - 1):
            p0 = np.array(self.points[index])
            p3 = np.array(self.points[index + 1])
            delta = p3 - p0
            # Place interior cubic control points at 1/3 and 2/3 so each
            # cubic segment represents a straight line between p0 and p3.
            p1 = p0 + (1.0 / 3.0) * delta
            p2 = p0 + (2.0 / 3.0) * delta
            bezier = self.world.add_bezier(
                p0=point_type(p0[0], p0[1], 0.0),
                p1=point_type(p1[0], p1[1], 0.0),
                p2=point_type(p2[0], p2[1], 0.0),
                p3=point_type(p3[0], p3[1], 0.0),
            )
            bezier.sample(int(nsample[index]))
```

The Python module `_canvas` is under the GUI sub-package `modmesh.pilot`.  Non-GUI, pure geometry code in Python has a separate sub-package `modmesh.plot`.

---

> @yungyuc Please take a look, thanks. I don't know why I cannot leave the comment for file change.

Probably because it's only renaming without diff.

modmesh/pilot/__init__.py
```diff
@@ -48,7 +48,7 @@
         launch,
     )
     from . import airfoil  # noqa: F401
-    from . import _canvas  # noqa: F401
+    from . import _canvas_gui  # noqa: F401
```
ThreeMonth03, line 51: I rename `pilot/_canvas.py` to `pilot/_canvas_gui.py`, because the name is aligned to `_svg_gui.py`.

modmesh/plot/curve.py
ThreeMonth03, line 1: Move non-gui code from `pilot/_canvas.py` to `plot/curve.py`, like sampler. They are not related to pilot.

tests/test_universe_curve.py
ThreeMonth03, line 1: Rename ‎`tests/test_pilot_canvas.py‎` to `‎tests/test_universe_curve.py`.

ThreeMonth03 (commented):
I refactor the source code, and I make filename more precise. @yungyuc please take a look. Thanks.

modmesh/pilot/__init__.py
yungyuc: Good name.

tests/test_universe_curve.py
```diff
@@ -28,26 +28,28 @@
 import unittest
 
 import modmesh as mm
-import pytest
 
-pytest.importorskip("PySide6")
```
yungyuc, line 33: Good to have the unnecessary import removed.  It will be easier to understand the modularization.

yungyuc (approved):
LGTM.  @tigercosmos @c1ydehhx please take a look.

modmesh/pilot/_canvas_gui.py
```diff
@@ -0,0 +1,182 @@
+# Copyright (c) 2026, Han-Xuan Huang <c1ydehhx@gmail.com>
```
tigercosmos, line 1: This reminds me about the copyright issue again https://github.com/solvcon/modmesh/issues/689.

@yungyuc  Maybe we can just change to use `Copyright (c) 2000, modmesh contributors`

tigercosmos: The reason is that the code may be modified, refactored, deleted, copied, or split over time. As a result, the original author can become disconnected from the latest version of the code. Therefore, attributing copyright to “modmesh contributors” is more reasonable.

PR #766, opened 2026-05-10
Prototype class `EigenSystem` for solving eigenvalue problems

yungyuc:
Use Apple Accelerate/veclib DGEEV (LAPACK interface) to prototype an eigenvalue solver for a general system.

Provide Python binding using pybind11 and create basic tests in Python.


yungyuc:
Sorry, this PR was sent to a wrong repository.

Issue #772, opened 2026-05-13
Timed-out CI job should be marked as failed


yungyuc:
A timed-out CI job is silently skipped rather than issuing a failure.  Example happened last night (timeout is 30 minutes): https://github.com/solvcon/modmesh/actions/runs/25756633699/job/75646672499

<img width="1474" height="613" alt="Image" src="https://github.com/user-attachments/assets/038584a4-c4e5-4eed-a799-71d2386ab5e2" />

A time-out should be reconfigured to fail loudly.

ThreeMonth03:
@yungyuc I wanna give it a hand. Please assign this task for me.

ExplorerRay:
The current PR, such as https://github.com/solvcon/modmesh/pull/780, does not handle one situation.

If a job is cancelled for other reasons, such as when another commit is pushed while the previous CI run is still in progress, the post step will also be triggered and report a failure.

However, this does not affect the CI result of the final commit. Therefore, I think this can be left as future work.

ExplorerRay:
I also tried to solve this in [my fork](https://github.com/ExplorerRay/modmesh/tree/ci-timeout-post). 
But cancellation in github action seems to override the failure I set in post step. You can refer to [this workflow](https://github.com/ExplorerRay/modmesh/actions/runs/25957410057/job/76306472288).

ThreeMonth03:
> I also tried to solve this in [my fork](https://github.com/ExplorerRay/modmesh/tree/ci-timeout-post). But cancellation in github action seems to override the failure I set in post step. You can refer to [this workflow](https://github.com/ExplorerRay/modmesh/actions/runs/25957410057/job/76306472288).

I don't find a method that check the job status without creating files. If you have any progress, please let us know.

ThreeMonth03:
@rockleona Are you interested in solving this issue?

rockleona:
> [@rockleona](https://github.com/rockleona) Are you interested in solving this issue?

Sure, I like to investigate this issue, will let you know if I have any new finding.

rockleona:
I found some reference jobs as the table below:

| Date (UTC) | Run # | Job | Duration | Timeout limit | Link |
|---|---|---|---|---|---|
| 2026-05-13 | 3401 | build_macos-26_Release | 34m02s | 30 min | [job 75861534232](https://github.com/solvcon/modmesh/actions/runs/25820890798/job/75861534232) |
[...]

PR #780, opened 2026-05-14
Create new job to make cancelled jobs as failure jobs in CI runner

ThreeMonth03:
In issue #772, we notice that when a job times out, it would be cancelled instead of fail.
Therefore, this pull request is to create new job, that could monitor cancelled jobs, and make cancelled jobs as failure jobs in CI runner.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -40,6 +40,8 @@ jobs:
         with:
           fetch-depth: 1
 
+      - uses: ./.github/actions/job_guard
+
       - name: event name
         run: |
           echo "github.event_name: ${{ github.event_name }}"
@@ -60,6 +62,8 @@ jobs:
         run: |
           make standalone_buffer_setup
           make standalone_buffer
+
+      - uses: ./.github/actions/job_done
 
   build:
 
@@ -97,6 +101,8 @@ jobs:
     - uses: actions/checkout@v6
       with:
         fetch-depth: 1
+
+    - uses: ./.github/actions/job_guard
 
     - name: event name
       run: |
@@ -202,6 +208,8 @@ jobs:
         make buildext VERBOSE=1
         make pytest VERBOSE=1
 
+    - uses: ./.github/actions/job_done
+
   build_windows:
 
     if: |
@@ -237,6 +245,8 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 1
+
+      - uses: ./.github/actions/job_guard
 
       - name: event name
         run: |
@@ -373,15 +383,18 @@ jobs:
           name: modmesh-pilot-win64
           path: modmesh-pilot-win64/
 
+      - uses: ./.github/actions/job_done
+
   send_email_on_failure:
     needs: [standalone_buffer, build, build_windows]
-    # Run if any of the dependencies failed in master branch
+    # Run if any of the dependencies failed or timed out in master branch
     if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' && github.event.repository.fork == false }}
     uses: ./.github/workflows/send_email_on_fail.yml
     with:
       job_results: |
         - standalone_buffer: ${{ needs.standalone_buffer.result }}
         - build: ${{ needs.build.result }}
+        - build_windows: ${{ needs.build_windows.result }}
     secrets:
       EMAIL_USERNAME: ${{ secrets.EMAIL_USERNAME }}
       EMAIL_PASSWORD: ${{ secrets.EMAIL_PASSWORD }}
--- a/.github/workflows/nouse_install.yml
+++ b/.github/workflows/nouse_install.yml
@@ -43,6 +43,8 @@ jobs:
         with:
           fetch-depth: 1
 
+      - uses: ./.github/actions/job_guard
+
       - name: event name
         run: |
           echo "github.event_name: ${{ github.event_name }}"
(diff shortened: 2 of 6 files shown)


.github/workflows/nouse_install.yml
```diff
@@ -85,14 +85,28 @@ jobs:
...
+
+  send_email_on_failure:
+    needs: [nouse_install, fail_on_cancelled_jobs]
+    # Run if any of the dependencies failed or timed out in master branch
     if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' && github.event.repository.fork == false }}
     uses: ./.github/workflows/send_email_on_fail.yml
     with:
       job_results: |
         - nouse_install: ${{ needs.nouse_install.result }}
+        - fail_on_cancelled_jobs: ${{ needs.fail_on_cancelled_jobs.result }}
```
ThreeMonth03: When a job is cancelled, a notified email would be sent now.

.github/workflows/nouse_install.yml
```diff
@@ -85,14 +85,28 @@ jobs:
...
+    if: ${{ always() }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Fail on cancelled jobs
+        run: |
+          result="${{ needs.nouse_install.result }}"
+          echo "nouse_install: ${result}"
+          if [ "${result}" = "cancelled" ]; then
+            exit 1
+          fi
```
ThreeMonth03: We cannot make a time-out job as a failure job easily. There are two workaround method. 
The first opinion is to create a new job to monitor the cancelled jobs, and it would raise failure when there is a cancelled job.
The second opinion is to time the process by a timer. If the process times out, the timer would return 1.

I prefer the first opinion, so I create a new job to monitor the cancelled jobs.

ThreeMonth03 (commented):
@yungyuc Please review this pull request when you are available. Thanks.

.github/workflows/devbuild.yml
```diff
@@ -373,15 +373,39 @@ jobs:
           name: modmesh-pilot-win64
           path: modmesh-pilot-win64/
 
-  send_email_on_failure:
+  fail_on_cancelled_jobs:
     needs: [standalone_buffer, build, build_windows]
-    # Run if any of the dependencies failed in master branch
+    if: ${{ always() }}
```
yungyuc: Is it necessary to write the `if`, if it `always()` happens?

.github/workflows/nouse_install.yml
yungyuc: This would work, but is not as elegant as I hoped for.

If there is not a better way, I would take it.  But I hope there is.

yungyuc (commented):
@ExplorerRay @chestercheng Could you please help brainstorm?

.github/workflows/devbuild.yml
ThreeMonth03: Yes, it is. It's an interesting technique. See [this reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds).


> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue. If a run contains a series of jobs that need each other, a failure or skip applies to all jobs in the dependency chain from the point of failure or skip onwards. If you would like a job to run even if a job it is dependent on did not succeed, use the always() conditional expression in jobs.<job_id>.if.

In other words, if we remove `if: ${{ always() }}`, this job would be skipped when the jobs in needs are cancelled or fail. However, we don't want to skip this job at any time, so we need to add `if: ${{ always() }}`.

yungyuc: This is cool.  Could you please add a code comment above the line for the link to the reference?  It look so innocent and future maintainers will thank for the point.

ThreeMonth03: No problem, I've added the code comment to make the statement clear.

ExplorerRay:
I think this PR can use https://github.com/marketplace/actions/run-and-post-run-action to solve this so that we don't need to handle a new job in CI.

I am considering adding a step like the following for each job that might timeout. And we may differentiate return code for failure and timeout (I write 2 in the following example).

```
- name: Failure when job timeout cancelled
  uses: gacts/run-and-post-run@v1
  with:
    post: |
      if [ "${{ job.status }}" == "cancelled" ]; then
        echo "::error::CI job timed out! Forcing failure."
        exit 2
      fi
```

.github/workflows/devbuild.yml
```diff
@@ -375,13 +403,14 @@ jobs:
...
     needs: [standalone_buffer, build, build_windows]
-    # Run if any of the dependencies failed in master branch
+    # Run if any of the dependencies failed or timed out in master branch
     if: ${{ always() && contains(needs.*.result, 'failure') && github.ref_name == 'master' && github.event.repository.fork == false }}
     uses: ./.github/workflows/send_email_on_fail.yml
     with:
       job_results: |
         - standalone_buffer: ${{ needs.standalone_buffer.result }}
         - build: ${{ needs.build.result }}
+        - build_windows: ${{ needs.build_windows.result }}
```
ExplorerRay, line 397: Nice catch. I forget this part.

.github/workflows/devbuild.yml
```diff
@@ -36,6 +36,15 @@ jobs:
...
 
+      - name: Fail if job was cancelled
+        uses: gacts/run-and-post-run@v1
+        with:
+          post: |
+            if [ "${{ job.status }}" = "cancelled" ]; then
+              echo "::error::CI job timed out or was cancelled; forcing failure."
+              exit 2
+            fi
+
```
ThreeMonth03: Register run and post job.

.github/workflows/devbuild.yml
```diff
@@ -36,6 +36,15 @@ jobs:
 
     steps:
 
+      - name: Fail if job was cancelled
```
ExplorerRay: I think these lines can be moved into the beginning of setup dependency composite action so that we can set this when running that action.

ExplorerRay (changes requested):
I think it's better to adding post check step into setup dependency composite action.

If there isn't a composite action to set up dependency like Windows, add an extra step like what you original did.

ThreeMonth03:
Close because of inactivity.

PR #787, opened 2026-05-17
Allow passing `pathlib.PosixPath` to the function `read_from_text_file()`

### LINKED ISSUE #786: pathlib.PosixPath is passed instead of string
https://github.com/solvcon/modmesh/blob/bf36b16f70c147fca2235e5e8e89db6838be870e/modmesh/track/dataframe.py#L70-L77

These code assumes that fname is either a string or a file handler. However, in Linux environment, it is pathlib.PosixPath and this scenario is not properly handled here.

xingularity:
Modified if statement in modmesh/track/dataframe. This is to handle file name as a PosixPath instance.

This is for issue #786.

### DIFF
--- a/tests/test_timeseries_dataframe.py
+++ b/tests/test_timeseries_dataframe.py
@@ -25,6 +25,9 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import io
+import os
+import pathlib
+import tempfile
 import unittest
 
 import numpy as np
@@ -163,3 +166,39 @@ class TimeSeriesDataFrameTC(unittest.TestCase):
         col_data = reordered_tsdf['DELTA_VEL[1]']
         nd_arr = np.genfromtxt(io.StringIO(self.dlc_data), delimiter=',')[1:]
         self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+    def test_read_from_text_file_accepts_str_path(self):
+        tsdf = dataframe.DataFrame()
+        with tempfile.NamedTemporaryFile(
+            mode='w', suffix='.csv', delete=False,
+        ) as fh:
+            fh.write(self.dlc_data)
+            path = fh.name
+        try:
+            tsdf.read_from_text_file(path)
+            self.assertEqual(tsdf._columns, self.col_sol)
+            self.assertEqual(tsdf._index_name, 'EPOCH')
+        finally:
+            os.unlink(path)
+
+    def test_read_from_text_file_accepts_pathlib_path(self):
+        tsdf = dataframe.DataFrame()
+        with tempfile.NamedTemporaryFile(
+            mode='w', suffix='.csv', delete=False,
+        ) as fh:
+            fh.write(self.dlc_data)
+            path = pathlib.Path(fh.name)
+        try:
+            tsdf.read_from_text_file(path)
+            self.assertEqual(tsdf._columns, self.col_sol)
+            self.assertEqual(tsdf._index_name, 'EPOCH')
+        finally:
+            path.unlink()
+
+    def test_read_from_text_file_missing_raises_filenotfound(self):
+        tsdf = dataframe.DataFrame()
+        missing = pathlib.Path(tempfile.gettempdir()) / 'no_such_file.csv'
+        expected_msg = "Text file '{}' does not exist".format(missing)
+        with self.assertRaises(FileNotFoundError) as ctx:
+            tsdf.read_from_text_file(missing)
+        self.assertEqual(str(ctx.exception), expected_msg)
--- a/tests/test_timeseries_dataframe.py
+++ b/tests/test_timeseries_dataframe.py
@@ -27,6 +27,7 @@
 import io
 import os
 import pathlib
+import re
 import tempfile
 import unittest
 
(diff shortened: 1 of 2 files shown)


tigercosmos:
Sorry maybe I didn't fully understand the test cases. I wonder are the test cases be able to catch the errors that happens in #786?

xingularity:
> Sorry maybe I didn't fully understand the test cases. I wonder are the test cases be able to catch the errors that happens in #786?

Yes, test case `test_read_from_text_file_accepts_pathlib_path` is for it. The `pathlib.Path` function call is going to generate a `pathlib.PosixPath` instance under linux.

yungyuc:
I changed the subject to be more informative.

yungyuc (changes requested):
Please leave inline annotations in the PR.  The fix itself is simple, but the unit tests are not trivial.  Also leave a global comment to request for review.

See the PR protocol in issue #811 .

xingularity:
> Please leave inline annotations in the PR. The fix itself is simple, but the unit tests are not trivial. Also leave a global comment to request for review.
> 
> See the PR protocol in issue #811 .

@yungyuc  and @tigercosmos The comment has been added, you can review now.

tests/test_timeseries_dataframe.py
```diff
@@ -163,3 +166,45 @@ def test_dataframe_sort(self):
...
+            fh.write(self.dlc_data)
+            path = fh.name
+        try:
+            tsdf.read_from_text_file(path)
+            self.assertEqual(tsdf._columns, self.col_sol)
+            self.assertEqual(tsdf._index_name, 'EPOCH')
+        finally:
+            os.unlink(path)
+
+# Theis test is for a bug reported in issue #786:
```
yungyuc: It seems that you misunderstood what we meant by inline annotations.  In https://github.com/solvcon/modmesh/pull/787#pullrequestreview-4350887671, "inline annotations in the PR" is like the inline comment I am leaving here.  It is not code remarks/comments.

Please remove the unnecessary code comments and turn it into a PR inline annotation.

tests/test_timeseries_dataframe.py
```diff
@@ -163,3 +166,45 @@ def test_dataframe_sort(self):
         col_data = reordered_tsdf['DELTA_VEL[1]']
         nd_arr = np.genfromtxt(io.StringIO(self.dlc_data), delimiter=',')[1:]
         self.assertEqual(list(col_data), list(nd_arr[:, 1]))
+
+    def test_read_from_text_file_accepts_str_path(self):
```
yungyuc, line 171: This does not look like related to the fix.  Please share with us why do you want to include it.

tests/test_timeseries_dataframe.py
```diff
@@ -163,3 +166,45 @@ def test_dataframe_sort(self):
...
+            tsdf.read_from_text_file(path)
+            self.assertEqual(tsdf._columns, self.col_sol)
+            self.assertEqual(tsdf._index_name, 'EPOCH')
+        finally:
+            path.unlink()
+
+    def test_read_from_text_file_missing_raises_filenotfound(self):
+        tsdf = dataframe.DataFrame()
+        missing = pathlib.Path(tempfile.gettempdir()) / 'no_such_file.csv'
+        with self.assertRaises(FileNotFoundError):
```
yungyuc: Also check for error messages, not just the exception type.

yungyuc (changes requested):
- [x] Use inline annotation for discussions in the PR, not code comments.
- [x] Also check for error messages, not just the exception type.

tests/test_timeseries_dataframe.py
tigercosmos: Nitpick: there is a typo `Theis`.

xingularity: This test is for a bug reported in issue #786: https://github.com/solvcon/modmesh/issues/786

If user uses time series DataFrame under Linux, it is likely to be given a `pathlib.PosixPath` instance to be the path to text file. This test is to make sure DataFrame support a path (frame) which is from `pathlib`.

tests/test_timeseries_dataframe.py
xingularity: The file path can be either a string or a `os.PathLike` instance. This test is added to explicitly test if the DataFrame can accept `str` type text file path.

tests/test_timeseries_dataframe.py
xingularity: Added in the commit f13f83de2fe3d000004984a2788785ee55ce6d72

xingularity:
Hi @yungyuc  and @tigercosmos The CI/CD has finally passed, this PR is ready for review again.

tests/test_timeseries_dataframe.py
```diff
@@ -163,3 +166,39 @@ def test_dataframe_sort(self):
...
+            self.assertEqual(tsdf._columns, self.col_sol)
+            self.assertEqual(tsdf._index_name, 'EPOCH')
+        finally:
+            path.unlink()
+
+    def test_read_from_text_file_missing_raises_filenotfound(self):
+        tsdf = dataframe.DataFrame()
+        missing = pathlib.Path(tempfile.gettempdir()) / 'no_such_file.csv'
+        expected_msg = "Text file '{}' does not exist".format(missing)
+        with self.assertRaises(FileNotFoundError) as ctx:
```
yungyuc: Use [`assertRaisesRegex()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) to test for both exception type and messages.

yungyuc (changes requested):
The point has not been addressed:
- [x] Also check for error messages, not just the exception type.

tests/test_timeseries_dataframe.py
xingularity: @yungyuc , it has been modified in c0cd272 commit.

yungyuc (approved):
LGTM

PR #794, opened 2026-05-20
Add `SimpleArray::transpose_copy()` and majoring helpers

### LINKED ISSUE #792: `SimpleArray::transpose()` should support a deep-copy variant
It is a `TODO` item added in [`cpp/modmesh/linalg/EigenSystem.hpp:124`](https://github.com/solvcon/modmesh/blob/a9c804f6bdfc9e91f88601cfdca53ef5a353c72c/cpp/modmesh/linalg/EigenSystem.hpp#L124) (constructor body):

```cpp
    // TODO: Enhance SimpleArray::transpose() to copy the contents when
    // transposing.  The current SimpleArray::transpose() is a *view-only*
    // operation: it reverses the shape and stride vectors but the underlying
    // buffer is untouched.  So we have to manually copy element-wisely in the
    // nested loops below.
```

This pushes manual `for (i)(j) m_colmajor(i,j) = matrix(i,j)` loops into every LAPACK caller that needs column-major data. We should provide a first-class operation in `SimpleArray`.

Add `SimpleArray::transpose_copy()`, `to_column_major()`, and `to_row_major()` that returns a freshly allocated buffer with physically transposed contents.  `transpose_copy()` should detects the current majoring and call the other two functions as helpers.

`SimpleArray::transpose()` should add a default argument `copy=false`.  It preserves the old view behavior, and allow to use the same function to perform deep-copy transpose.

Refactor `EigenSystem` constructor to use `transpose(copy=true)` to remove the hand-written loops and the `TODO` comment.

Add Python unit tests on `SimpleArray` for non-square and contiguous/non-contiguous inputs of 1D, 2D, 3D, and 4D shapes.

Related to #768.

yungyuc:
The existing function `transpose(copy=false)` is augmented with the new optional argument `copy`.  Default `false` keeps the view-only flipping behavior; `copy=true` does an in-place deep-copy transpose into a fresh row-major (C-contiguous) buffer.

Add new member functions:
- `transpose_copy()` returns that deep-copy transpose without mutating `*this`.
- `to_row_major()` / `to_column_major()` materialise the same logical data in the requested contiguity, with a clone fast-path when the source already matches.
- Private `copy_logical_into()` is the heavy-lifting element-copy helper.  It currently uses naive copy.  Cache optimization may be added in the future.

The new methods are exposed with pybind11 and tested in Python.  A new test function `test_SimpleArray_transpose_copy()` is added under `SimpleArrayBasicTC` in `test_buffer.py` to cover 1D-4D, square and non-square, row- and column-major inputs.

`EigenSystem` now stages its LAPACK workspace by initializing `m_colmajor` directly from `matrix.to_column_major()` in the member initializer list.  The former TODO for removing the manual element-wise loop is addressed.

Related to issue #792.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -2176,36 +2176,15 @@ public:
         }
     }
 
-    void transpose()
-    {
-        std::reverse(m_shape.begin(), m_shape.end());
-        std::reverse(m_stride.begin(), m_stride.end());
-    }
+    void transpose(bool copy = false);
 
-    void transpose(shape_type const & axis)
-    {
-        if (axis.size() != m_shape.size())
-        {
-            throw std::runtime_error("SimpleArray: axis size mismatch");
-        }
-        shape_type new_shape(m_shape.size(), -1);
-        shape_type new_stride(m_stride.size());
-        for (size_t it = 0; it < m_shape.size(); ++it)
-        {
-            if (axis[it] >= m_shape.size() || axis[it] < 0)
-            {
-                throw std::runtime_error("SimpleArray: axis out of range");
-            }
-            if (new_shape[it] != -1)
-            {
-                throw std::runtime_error("SimpleArray: axis already set");
-            }
-            new_shape[it] = m_shape[axis[it]];
-            new_stride[it] = m_stride[axis[it]];
-        }
-        m_shape = new_shape;
-        m_stride = new_stride;
-    }
+    void transpose(shape_type const & axis, bool copy = false);
+
+    SimpleArray transpose_copy() const;
+
+    SimpleArray to_row_major() const;
+
+    SimpleArray to_column_major() const;
 
     template <typename... Args>
     value_type const & operator()(Args... args) const { return *vptr(args...); }
@@ -2284,6 +2263,8 @@ public:
     bool is_f_contiguous() const { return is_f_contiguous(m_shape, m_stride); }
 
 private:
+    void copy_logical_into(SimpleArray & out) const;
+
     static bool is_c_contiguous(small_vector<size_t> const & shape,
                                 small_vector<size_t> const & stride)
     {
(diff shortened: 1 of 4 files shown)


cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -203,21 +203,42 @@ class MODMESH_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 { return self.reshape(make_shape(shape)); })
             .def(
                 "transpose",
-                [](wrapped_type & self, py::object const & axis, bool const & inplace)
+                [](wrapped_type & self, py::object const & axis, bool inplace, bool copy)
```
yungyuc, line 206: `bool const &` does not help anything.  `bool` to pass by value is efficient.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2176,36 +2176,15 @@ class SimpleArray
         }
     }
 
-    void transpose()
-    {
-        std::reverse(m_shape.begin(), m_shape.end());
-        std::reverse(m_stride.begin(), m_stride.end());
-    }
+    void transpose(bool copy = false);
```
yungyuc, line 2179: New functions have code put outside the class declaration.  The code of the old `transpose()` is moved too.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+template <typename T>
+void SimpleArray<T>::transpose(bool copy)
+{
+    if (!copy)
+    {
+        std::reverse(m_shape.begin(), m_shape.end());
+        std::reverse(m_stride.begin(), m_stride.end());
+        return;
+    }
+    SimpleArray tc = transpose_copy();
```
yungyuc, line 2451: Call `transpose_copy()` to copy.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+        }
+        new_shape[it] = m_shape[axis[it]];
+        new_stride[it] = m_stride[axis[it]];
+    }
+    if (!copy)
+    {
+        // View-only flip: install the permuted metadata in place.
+        m_shape = new_shape;
+        m_stride = new_stride;
+        return;
```
yungyuc, line 2495: Early return if not copy.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+    }
+    // Deep-copy path: 1. stage a strided view with the permuted shape/stride.
+    SimpleArray const view(new_shape, new_stride, m_buffer);
+    SimpleArray out(new_shape);
```
yungyuc, line 2499: Construct and allocate.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+    // 2. Use the same helper to keep the iteration logic the same as the other
+    // deep-copy variants.
+    view.copy_logical_into(out);
```
yungyuc, line 2502: Copy elements.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+ * @return Freshly allocated C-contiguous SimpleArray.
+ */
+template <typename T>
+SimpleArray<T> SimpleArray<T>::to_row_major() const
+{
+    if (is_c_contiguous())
+    {
+        return SimpleArray(*this);
+    }
+    SimpleArray out(m_shape);
```
yungyuc, line 2549: Allocate and copy to make the return as row major.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+    {
+        return SimpleArray(*this);
+    }
+    if (is_f_contiguous())
+    {
+        return SimpleArray(*this);
+    }
+    // Compute column-major strides: the fastest-varying axis is the leading
+    // one (stride[0] == 1).
+    shape_type fstride(m_shape.size());
```
yungyuc, line 2580: Same shape, reversed stride.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+ * @brief Copy `*this` into `out` element-wise.
+ *
+ * @details
+ *      Heavy-lifting helper to copy data. `out` must share the same shape but
+ *      may carry a different stride.
+ *
+ * @param out Destination array; its shape must match the receiver's.
+ */
+template <typename T>
+void SimpleArray<T>::copy_logical_into(SimpleArray & out) const
```
yungyuc, line 2609: `copy_logical_into()` is the heavy lifting of the element copy.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2450,6 +2431,219 @@ class SimpleArray
...
+    if (total == 0)
+    {
+        return;
+    }
+    // Walk a multi-dimensional index in row-major order.  out's stride is only
+    // used to compute the destination offset to bridge row-major,
+    // column-major, and arbitrary strided layouts.
+    size_t const ndim = m_shape.size();
+    shape_type idx(ndim, 0);
+    for (size_t step = 0; step < total; ++step)
```
yungyuc, line 2629: Loop to copy.

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -108,7 +108,8 @@ class EigenSystem
 
 inline EigenSystem::EigenSystem(array_type const & matrix)
     : m_matrix(matrix)
-    , m_colmajor(matrix.shape())
+    // Stage matrix into a column-major workspace for LAPACK.
+    , m_colmajor(matrix.to_column_major())
```
yungyuc, line 112: Just call `to_column_major()` to get an array using column majoring.

tests/test_buffer.py
```diff
@@ -371,6 +371,100 @@ def check_equal(sarr, ndarr):
...
+        # 3D, 4D; square / non-square; C-contiguous / F-contiguous input.
+
+        def make_sarr(ndarr):
+            return modmesh.SimpleArrayFloat64(array=ndarr.copy())
+
+        def values_equal(sarr, ref):
+            np.testing.assert_array_equal(np.array(sarr, copy=False), ref)
+
+        shapes = [
+            (7,),
```
yungyuc, line 385: 1D-4D fixtures.

tests/test_buffer.py
```diff
@@ -371,6 +371,100 @@ def check_equal(sarr, ndarr):
...
+            (3, 5),
+            (4, 4),
+            (2, 3, 4),
+            (2, 3, 4, 5),
+        ]
+
+        for shape in shapes:
+            ndarr = np.arange(
```
yungyuc, line 393: Use `arange` to populate non-trivial data.

tigercosmos (commented):
LGTM

PR #795, opened 2026-05-20
Add do_vl/do_vr opt-out flags to `EigenSystem`

### LINKED ISSUE #791: Add API/arguments to opt out of computing left and/or right eigenvectors in `EigenSystem`
In PR #768, `EigenSystem::run()` hardcodes `jobvl = jobvr = 'V'`, always computing both eigenvector matrices and the workspace they need.  Many use sites only need one (or just eigenvalues). DGEEV supports `'N'` to skip.

Add constructor and `run()` arguments `do_vl`, `do_vr` (default both true to preserve the current behavior).  Skip allocation of `m_vl` / `m_vr` when the corresponding flag is false; pass `'N'` to DGEEV.  Reject reads of `vl`/`vr` when they not computed.  The default behavior is to throw an exception, but add an optional function argument to suppress the exception and return empty.

Add Python wrappers and tests.

yungyuc:
Allow to construct the class `EigenSystem` with `do_vl=false` and/or `do_vr=false` to skip the corresponding eigenvector computation.

With the options to be off, the corresponding matrix stays empty.  Calling the getters `vl()` / `vr()` throw when the matrix was not computed unless called with `suppress_exception=true`, which returns the empty placeholder.

Python keeps `vl` / `vr` as read-only properties and adds `get_vl()` / `get_vr()` helpers exposing the flag.

Related to issue #791.

### DIFF
--- a/cpp/modmesh/linalg/EigenSystem.hpp
+++ b/cpp/modmesh/linalg/EigenSystem.hpp
@@ -73,7 +73,7 @@ public:
     using value_type = double;
     using array_type = SimpleArray<double>;
 
-    explicit EigenSystem(array_type const & matrix);
+    explicit EigenSystem(array_type const & matrix, bool do_vl = true, bool do_vr = true);
 
     EigenSystem() = delete;
     EigenSystem(EigenSystem const &) = delete;
@@ -82,14 +82,15 @@ public:
     EigenSystem & operator=(EigenSystem &&) = delete;
     ~EigenSystem() = default;
 
-    /// Run DGEEV on the prepared workspace.
     void run();
 
     array_type const & matrix() const { return m_matrix; }
     array_type const & wr() const { return m_wr; }
     array_type const & wi() const { return m_wi; }
-    array_type const & vl() const { return m_vl; }
-    array_type const & vr() const { return m_vr; }
+    array_type const & vl(bool suppress_exception = false) const;
+    array_type const & vr(bool suppress_exception = false) const;
+    bool do_vl() const { return m_do_vl; }
+    bool do_vr() const { return m_do_vr; }
     bool done() const { return m_done; }
 
 private:
@@ -102,18 +103,22 @@ private:
     SimpleArray<value_type> m_wi;
     SimpleArray<value_type> m_vl;
     SimpleArray<value_type> m_vr;
+    bool const m_do_vl;
+    bool const m_do_vr;
     bool m_done = false;
 
 }; /* end class EigenSystem */
 
-inline EigenSystem::EigenSystem(array_type const & matrix)
+inline EigenSystem::EigenSystem(array_type const & matrix, bool do_vl, bool do_vr)
     : m_matrix(matrix)
     // Stage matrix into a column-major workspace for LAPACK.
     , m_colmajor(matrix.to_column_major())
     , m_wr(matrix.shape(0))
     , m_wi(matrix.shape(0))
-    , m_vl(matrix.shape())
-    , m_vr(matrix.shape())
+    , m_vl(do_vl ? array_type(matrix.shape()) : array_type())
+    , m_vr(do_vr ? array_type(matrix.shape()) : array_type())
+    , m_do_vl(do_vl)
+    , m_do_vr(do_vr)
 {
     if (matrix.ndim() != 2 || matrix.shape(0) != matrix.shape(1))
     {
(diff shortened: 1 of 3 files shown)


cpp/modmesh/linalg/pymod/wrap_EigenSystem.cpp
```diff
@@ -78,8 +80,32 @@ WrapEigenSystem::WrapEigenSystem(pybind11::module & mod, char const * pyname, ch
...
+        .def_property_readonly(
+            "vr",
+            [](wrapped_type const & self) -> array_type const &
+            {
+                return self.vr();
+            })
+        .def(
+            "get_vl",
+            &wrapped_type::vl,
+            py::arg("suppress_exception") = false,
```
yungyuc, line 98: New getters to allow throwing an exception and suppress it.

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -82,14 +82,15 @@ class EigenSystem
...
 
-    /// Run DGEEV on the prepared workspace.
     void run();
 
     array_type const & matrix() const { return m_matrix; }
     array_type const & wr() const { return m_wr; }
     array_type const & wi() const { return m_wi; }
-    array_type const & vl() const { return m_vl; }
-    array_type const & vr() const { return m_vr; }
+    array_type const & vl(bool suppress_exception = false) const;
```
yungyuc, line 90: Add suppression argument.

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -82,14 +82,15 @@ class EigenSystem
...
+    array_type const & vr(bool suppress_exception = false) const;
+    bool do_vl() const { return m_do_vl; }
```
yungyuc, line 92: Add getter for `do_vl/vr()`

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -155,12 +169,19 @@ inline void EigenSystem::run()
      *   https://www.netlib.org/lapack/explore-html/d4/d68/group__geev_ga7d8afe93d23c5862e238626905ee145e.html
      *   https://www.netlib.org/lapack/explore-html/d9/d28/dgeev_8f_source.html
      */
-    char const jobvl = 'V';
-    char const jobvr = 'V';
+    char const jobvl = m_do_vl ? 'V' : 'N';
```
yungyuc, line 159: Use the calculation selector.

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -222,6 +243,50 @@ inline void EigenSystem::run()
...
+ *      Return the empty placeholder instead of throwing when the matrix was
+ *      not computed.
+ * @return
+ *      Column-major n-by-n matrix; empty when do_vl=false and
+ *      suppress_exception=true.
+ * @throws std::runtime_error When do_vl=false and suppress_exception=false.
+ */
+inline EigenSystem::array_type const & EigenSystem::vl(bool suppress_exception) const
+{
+    if (!m_do_vl && !suppress_exception)
```
yungyuc, line 246: Throw exception if `vl` is not requested.

cpp/modmesh/linalg/EigenSystem.hpp
```diff
@@ -222,6 +243,50 @@ inline void EigenSystem::run()
...
+ *      Return the empty placeholder instead of throwing when the matrix was
+ *      not computed.
+ * @return
+ *      Column-major n-by-n matrix; empty when do_vr=false and
+ *      suppress_exception=true.
+ * @throws std::runtime_error When do_vr=false and suppress_exception=false.
+ */
+inline EigenSystem::array_type const & EigenSystem::vr(bool suppress_exception) const
+{
+    if (!m_do_vr && !suppress_exception)
```
yungyuc, line 268: Throw exception if `vr` is not requested.

tests/test_linalg.py
```diff
@@ -1422,4 +1367,111 @@ def test_matrix_property_survives_input_gc(self):
         del A
         np.testing.assert_array_equal(solver.matrix.ndarray, A_np)
 
+    def test_default_constructor_flags_both_true(self):
```
yungyuc, line 1425: New test functions for the optional `vl` / `vr` calculation.

cpp/modmesh/linalg/EigenSystem.hpp
tigercosmos: Could you explain why there is a "suppress_exception" design? Why not always throwing?

tigercosmos (approved):
LGTM. Raised [one question](https://github.com/solvcon/modmesh/pull/795/changes#r3276019164).

cpp/modmesh/linalg/EigenSystem.hpp
yungyuc: Good point.  It just happened many times that the caller sites want to unconditionally run through and leave exceptions behind.  Instead of adding a lot of error handling code, I chose to collect them at one place.

PR #762, opened 2026-05-09
Add a prototype macOS app bundle script

yungyuc:
Add a prototype script contrib/bundle/bundle-with-homebrew.sh to build pilot, vendor every Homebrew Qt/Python/transitive dependency into the .app, ad-hoc codesign, packages a DMG, and verify hermeticity (static Mach-O scan plus a sandboxed smoke launch under `DYLD_PRINT_LIBRARIES=1`).

Use a specific build path `build/relbundle`.

Each step reports wall time; total is ~5 minutes end-to-end; ~3 minutes for incremental build.

Flags: --skip-build, --skip-check, --output DIR.

It only works with a system that uses homebrew to manage the macOS dependency. Homebrew is not a dependable way to manage dependency for an official build. The current approach in the prototype is just a starting point. We will plan for improvement.

It can only be tested manually and locally now.  CI should be set up to run it nightly as a separate PR later.

Claude Opus 4.7 proudly generated the code with tokens and taught me something about making a macOS bundle.

PR #798, opened 2026-05-21
Make pilot bundle script tolerant of host-specific state

tigercosmos:
The bundle file can only work in the VM environment. This PR adds the support for the host. Now the bundle file can work in both the host or the VM.

The change has been verified in both the pure VM macOS 26 environment and the host macOS 26 environment.

Related to #762.

### DIFF
--- a/contrib/bundle/bundle-with-homebrew.sh
+++ b/contrib/bundle/bundle-with-homebrew.sh
@@ -188,6 +188,41 @@ for d in site.getsitepackages():
 print('\n'.join(out))
 ")
 
+# Resolve the closure of Python packages pilot actually imports, so Step 4
+# copies only those, not the developer's full site-packages (which can
+# contain unrelated packages: vtkmodules, openvino, PyQt6, ...). Override
+# seeds via BUNDLE_SEED_MODULES; force-add via BUNDLE_EXTRA_PACKAGES (for
+# lazy / runtime imports the probe cannot observe, e.g. matplotlib backends
+# loaded on first render). Failed seed imports go to stderr so a partial
+# closure is visible instead of silently shipping a broken bundle.
+WANTED_PATHS=()
+while IFS= read -r line; do
+    [[ -n "$line" ]] && WANTED_PATHS+=("$line")
+done < <(python3 -c "
+import importlib, os, sys
+seeds = os.environ.get('BUNDLE_SEED_MODULES',
+    'PySide6.QtCore,PySide6.QtWidgets,PySide6.QtGui,numpy,matplotlib').split(',')
+extras = os.environ.get('BUNDLE_EXTRA_PACKAGES', '').split(',')
+for m in seeds + extras:
+    m = m.strip()
+    if not m: continue
+    try: importlib.import_module(m)
+    except Exception as e:
+        sys.stderr.write('    WARN: seed import %s failed: %s\n' % (m, e))
+out = set()
+for name, mod in list(sys.modules.items()):
+    if mod is None or '.' in name: continue
+    p = getattr(mod, '__path__', None)
+    f = getattr(mod, '__file__', None) or ''
+    if p:
+        for d in p:
+            if 'site-packages' in d:
+                out.add(os.path.realpath(d)); break
+    elif 'site-packages' in f:
+        out.add(os.path.realpath(f))
+print('\n'.join(sorted(out)))
+")
+
 echo "==> Build path : $BUILD_PATH"
 echo "==> App bundle : $APP"
 echo "==> Python fw  : $PY_FW"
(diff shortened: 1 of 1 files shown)


contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -188,6 +188,41 @@ for d in site.getsitepackages():
...
 ")
 
+# Resolve the closure of Python packages pilot actually imports, so Step 4
+# copies only those, not the developer's full site-packages (which can
+# contain unrelated packages: vtkmodules, openvino, PyQt6, ...). Override
+# seeds via BUNDLE_SEED_MODULES; force-add via BUNDLE_EXTRA_PACKAGES (for
+# lazy / runtime imports the probe cannot observe, e.g. matplotlib backends
+# loaded on first render). Failed seed imports go to stderr so a partial
+# closure is visible instead of silently shipping a broken bundle.
+WANTED_PATHS=()
```
tigercosmos, line 198: **Fix 1/4: import-closure probe.** Resolves the site-packages dirs pilot actually imports, so Step 4 stops copying unrelated dev packages (vtkmodules, openvino, PyQt6, ...) into the bundle. Override via `BUNDLE_SEED_MODULES` / `BUNDLE_EXTRA_PACKAGES`.

contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -258,14 +293,26 @@ echo "    [Step 3 (Bundle Python framework): $((SECONDS - T_STEP))s]"
...
-# and the directories they point at are themselves enumerated as SITE_DIRS
-# above.
-for SITE in "${SITE_DIRS[@]}"; do
-    rsync -aL --exclude '__pycache__' --exclude '*.pth' \
-        "$SITE/" "$BUNDLED_SITE/"
-done
+# Copy the closure resolved above (dir => rsync tree, file => cp). -L
+# dereferences Cellar symlinks. Falls back to every SITE_DIR if the probe
+# failed, so the script still runs when import discovery breaks.
+if [[ ${#WANTED_PATHS[@]} -gt 0 ]]; then
```
tigercosmos, line 299: Step 4 now copies only the resolved closure. Falls back to the old full-`SITE_DIRS` rsync if the probe returned nothing, so the script still runs when import discovery breaks.

contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -475,13 +545,112 @@ strip_homebrew_rpaths() {
     done < <(list_macho "$APP")
 }
 
+# Drop any Mach-O still referencing /opt/homebrew after vendor + rewrite have
+# run -- typically Qt plugins (e.g. PySide6's sqldrivers) pre-linked against
+# Homebrew kegs not installed on this host. dyld would fail to load them
+# anyway. Skips framework MAIN binaries since removing those breaks the
+# framework structure; that case is better surfaced by the static check.
+prune_unfixable_machos() {
```
tigercosmos, line 553: **Fix 2/4: `prune_unfixable_machos`.** After vendor + rewrite, any Mach-O still pointing at `/opt/homebrew` (typically Qt plugins linked against kegs not installed on this host) gets deleted with a WARN. Prevents a silently-broken bundle when a Homebrew source is missing; pairs with the new `WARN: missing source` lines in `vendor_homebrew_deps`.

contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -475,13 +545,112 @@ strip_homebrew_rpaths() {
...
+}
+
+# PySide6 wheels ship a private copy of Qt under PySide6/Qt/lib. Its .abi3.so
+# modules load that copy via @rpath; pilot's C++ binary loads Qt from
+# Contents/Frameworks/. Two Qt instances in one process => duplicate Obj-C
+# classes, separate global state, "QWidget: Must construct a QApplication"
+# abort. Replace each duplicated nested framework with a relative symlink to
+# the main bundle copy. Frameworks unique to PySide6 (QtCharts, ...) are left
+# in place; their @rpath/QtCore... refs resolve through these symlinks too.
+dedupe_pyside_qt() {
```
tigercosmos, line 591: **Fix 3/4: `dedupe_pyside_qt`.** PySide6 ships a private Qt under `PySide6/Qt/lib`; pilot's C++ binary loads Qt from `Contents/Frameworks/`. Two Qt instances in one process => duplicate Obj-C classes and the `QWidget: Must construct a QApplication` abort. This replaces each duplicated nested framework with a relative symlink into the main bundle Qt.

contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -475,13 +545,112 @@ strip_homebrew_rpaths() {
...
+    if [[ $count -gt 0 ]]; then
+        echo "    deduped $count PySide6/Qt/lib framework(s) into main Qt"
+    fi
+}
+
+# Set PYTHONNOUSERSITE=1 in LSEnvironment so a Finder/open launch of pilot
+# never adds ~/Library/Python/<X.Y>/lib/python/site-packages to sys.path,
+# which on dev machines may carry a second PySide6 (and its Qt). Same
+# dual-instance crash as dedupe_pyside_qt fixes for the in-bundle case.
+inject_pythonnousersite() {
```
tigercosmos, line 615: **Fix 4/4: `inject_pythonnousersite`.** Sets `LSEnvironment.PYTHONNOUSERSITE=1` in `Info.plist` so a Finder/`open` launch never picks up `~/Library/Python/<X.Y>/.../site-packages` — which on dev machines can carry a second PySide6 and trigger the same dual-Qt crash `dedupe_pyside_qt` fixes for the in-bundle case.

tigercosmos:

contrib/bundle/bundle-with-homebrew.sh
yungyuc: Great enhancement.  This should reduce bundle size.  The bundle size I had was like 700MB.  How large is the bundle built using this version?

contrib/bundle/bundle-with-homebrew.sh
```diff
@@ -258,14 +293,26 @@ echo "    [Step 3 (Bundle Python framework): $((SECONDS - T_STEP))s]"
...
-# rsync every site-packages dir into the bundle. -L dereferences Cellar
-# symlinks; .pth files are skipped because they encode absolute Cellar paths,
-# and the directories they point at are themselves enumerated as SITE_DIRS
-# above.
-for SITE in "${SITE_DIRS[@]}"; do
-    rsync -aL --exclude '__pycache__' --exclude '*.pth' \
-        "$SITE/" "$BUNDLED_SITE/"
-done
+# Copy the closure resolved above (dir => rsync tree, file => cp). -L
+# dereferences Cellar symlinks. Falls back to every SITE_DIR if the probe
```
yungyuc: I guess we do not need to say "Celler".  Saying symlinks should suffice.

yungyuc (changes requested):
> The change has been verified in both the pure VM macOS 26 environment and the host macOS 26 environment.

Wonderful!

Just a minor issue in comment:
- [x] Do not mention "Celler" in comments when it is not required.

Let me know if you want this to check in now and leave the comment issue in a later PR, or fixing it in a new commit.  Either way works for me.

@iefiru, please take a look.

yungyuc (approved):

tigercosmos:
@yungyuc The comment issues has been fixed. The new bundle is about 250MB.

yungyuc:
> @yungyuc The comment issues has been fixed. The new bundle is about 250MB.

Much less than the 700MB I had.  While it is still larger than ideal, we are not in a hurry to reduce the size, because the primary features are not fully ready.  It is substantial improvement already.

PR #802, opened 2026-05-23
Make `SimpleArray::transpose()` cache-friendly

### LINKED ISSUE #797: Cache-friendly block transpose for `SimpleArray`
Follow-up to issue #792 and its PR #794.  PR #794 added a private `SimpleArray::copy_logical_into()` for element-wise copy, without considering cache friendliness.  This issue tracks the development of cache-friendly transpose.

yungyuc:
Replace the single-element walker in SimpleArray::copy_logical_into with a family of physical-copy kernels grouped into a helper class `SimpleArrayCopier`.

`SimpleArrayCopier` takes a pair of `ConcreteBuffer` and holds the pointers to the internal memory, with the array metadata shape, strides, and itemsize.  It provides 4 copy kernels:

- `memcpy()`: single-buffer fast path for matching-stride contiguous layouts.
- `tiled_2d()`: 32x32 tile, stride-hoisted; the headline case for LAPACK staging.
- `tiled_nd()`: 32x32 tile on the two innermost axes plus a carry-propagating walk over the outer axes.
- `naive()`: original single-element walker, retained as a reference implementation for benchmarking and validation; not called by the dispatcher.

`SimpleArray<T>::copy_logical_into()` stays in the header as a template dispatcher to `SimpleArrayCopier`.  No change of public API.

Microbenchmarks on `SimpleArray<double>::transpose_copy()`, single-threaded, best of 3 x 5-iteration runs (not included in benchmark/profiling code; not yet needed):

| Shape | naive (ms) | tiled (ms) | Speedup |
|---|---:|---:|---:|
| `(256, 256)` | 0.345 | 0.125 | 2.8x |
| `(1024, 1024)` | 11.575 | 2.782 | 4.2x |
| `(2048, 2048)` | 70.817 | 16.336 | 4.3x |
| `(4, 256, 256)` | 4.402 | 0.587 | 7.5x |
| `(128, 128, 128)` | 46.956 | 9.696 | 4.8x |
| `(64, 64, 64, 64)` | 538.971 | 279.808 | 1.9x |

For issue #797.  Follow up on PR #794.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -29,10 +29,320 @@
 #include <modmesh/buffer/SimpleArray.hpp>
 #include <modmesh/math/math.hpp>
 
+#include <algorithm>
+#include <cstring>
 #include <unordered_map>
 
 namespace modmesh
 {
+
+/**
+ * Typed element-copy helper used by the per-itemsize specializations of the
+ * SimpleArrayCopier kernels.
+ */
+template <size_t N>
+static inline void copy_one(int8_t * dst, int8_t const * src)
+{
+    // Reduces every element move in the inner loops to a std::memcpy with a
+    // compile-time-constant size, which the compiler inlines into a single
+    // load/store on common dtypes.
+    std::memcpy(dst, src, N);
+}
+
+template <size_t N>
+static void tiled_2d_impl(
+    int8_t * const dst_body, int8_t const * const src_body, size_t const n0, size_t const n1, size_t const ss0, size_t const ss1, size_t const os0, size_t const os1)
+{
+    constexpr size_t BLOCK = 32;
+    for (size_t i0 = 0; i0 < n0; i0 += BLOCK)
+    {
+        size_t const i_end = std::min(i0 + BLOCK, n0);
+        for (size_t j0 = 0; j0 < n1; j0 += BLOCK)
+        {
+            size_t const j_end = std::min(j0 + BLOCK, n1);
+            for (size_t i = i0; i < i_end; ++i)
+            {
+                int8_t const * src_row = src_body + i * ss0;
+                int8_t * dst_row = dst_body + i * os0;
+                for (size_t j = j0; j < j_end; ++j)
+                {
+                    copy_one<N>(dst_row + j * os1, src_row + j * ss1);
+                }
+            }
+        }
+    }
+}
+
+/**
+ * Generic per-itemsize 2-D kernel that falls back to memcpy.  Used only for
+ * itemsizes that are not in the specialized {1, 2, 4, 8, 16} set.
+ */
+static inline void tiled_2d_generic(
+    int8_t * const dst_body, int8_t const * const src_body, size_t const n0, size_t const n1, size_t const ss0, size_t const ss1, size_t const os0, size_t const os1, size_t const itemsize)
+{
+    constexpr size_t BLOCK = 32;
+    for (size_t i0 = 0; i0 < n0; i0 += BLOCK)
+    {
+        size_t const i_end = std::min(i0 + BLOCK, n0);
+        for (size_t j0 = 0; j0 < n1; j0 += BLOCK)
+        {
+            size_t const j_end = std::min(j0 + BLOCK, n1);
+            for (size_t i = i0; i < i_end; ++i)
+            {
+                int8_t const * src_row = src_body + i * ss0;
+                int8_t * dst_row = dst_body + i * os0;
+                for (size_t j = j0; j < j_end; ++j)
+                {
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2621,27 +2674,33 @@ void SimpleArray<T>::copy_logical_into(SimpleArray & out) const
...
-    // used to compute the destination offset to bridge row-major,
-    // column-major, and arbitrary strided layouts.
-    size_t const ndim = m_shape.size();
-    shape_type idx(ndim, 0);
-    for (size_t step = 0; step < total; ++step)
+    // Subtract on T* so no reinterpret_cast is needed; ITEMSIZE then maps
+    // the element offset to the byte offset the helper expects.
+    auto const src_body_offset = static_cast<size_t>(m_body - m_buffer->template data<value_type>()) * ITEMSIZE;
+    auto const dst_body_offset = static_cast<size_t>(out.m_body - out.m_buffer->template data<value_type>()) * ITEMSIZE;
+    SimpleArrayCopier const copier(
```
yungyuc, line 2681: Replace the naive carry-walk to `SimpleArrayCopier` which abstract the optimized element copy.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2431,6 +2431,58 @@ class SimpleArray
...
+
+    SimpleArrayCopier(
+        buffer_type const & src_buffer,
+        size_t src_body_offset,
+        shape_type const & src_stride,
+        buffer_type & dst_buffer,
+        size_t dst_body_offset,
+        shape_type const & dst_stride,
+        shape_type const & shape,
+        size_t itemsize);
```
yungyuc, line 2461: `SimpleArrayCopier` is made a type erasure by using `itemsize` explicitly.  The implementation can be taken outside an include file to save compile time.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -29,11 +29,321 @@
...
+            outer_idx[i] = 0;
+        }
+    }
+}
+
+/**
+ * Naive single-element walker, kept as a reference implementation.  Not called
+ * by the dispatcher.
+ */
+void SimpleArrayCopier::naive() const
```
yungyuc, line 304: Reference implementation without being wired to an execution path.  It can be manually turned on when debugging is needed.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -29,11 +29,321 @@
...
+    {
+        total *= s;
+    }
+    std::memcpy(m_dst, m_src, total * m_itemsize);
+}
+
+/**
+ * 2-D 32x32 tile kernel.  Valid only for ndim == 2.
+ */
+void SimpleArrayCopier::tiled_2d() const
```
yungyuc, line 216: 2D transpose is the most common use case.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -29,11 +29,321 @@
...
+static inline void copy_one(int8_t * dst, int8_t const * src)
+{
+    // Reduces every element move in the inner loops to a std::memcpy with a
+    // compile-time-constant size, which the compiler inlines into a single
+    // load/store on common dtypes.
+    std::memcpy(dst, src, N);
+}
+
+template <size_t N>
+static void tiled_2d_impl(
```
yungyuc, line 53: Most useful helper for 2D transpose element copy.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -29,11 +29,321 @@
...
+            }
+        }
+    }
+}
+
+/**
+ * Generic per-itemsize 2-D kernel that falls back to memcpy.  Used only for
+ * itemsizes that are not in the specialized {1, 2, 4, 8, 16} set.
+ */
+static inline void tiled_2d_generic(
```
yungyuc, line 80: Having this just in case for elements of larger sizes.  I don't think we have them now, but if someone adds the larger elements, this will readily work for make the elements.  Optimization can be planned in the future when it is really needed.

cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -29,11 +29,321 @@
...
+                {
+                    std::memcpy(dst_row + j * os1, src_row + j * ss1, itemsize);
+                }
+            }
+        }
+    }
+}
+
+template <size_t N>
+static void tiled_nd_inner(
```
yungyuc, line 104: Transpose copy for N-D.  I expect it to be less useful than 2D transpose.  If more optimization is needed we can do it in the future.

PR #812, opened 2026-05-23
Add create-pr skill encoding the project PR protocol

### LINKED ISSUE #811: Create PR templates
The PR protocol for this project is:
* Use a concise and informative subject.
* Use a clear global description.
* When the PR is not ready for review, it should be made as a draft.
* When the PR is ready for review, the author needs to leave a *global comment* explicitly requesting for review.  Pushing the "ready for review" button means nothing.  (The push-button record cannot be quoted in follow-up comments.)
* The author is responsible for adding inline annotations to guide the review, unless the diff is one-liner-ish.
* All comments are written by humans.  Tool-assisting is OK.  It does not matter how the patches and comments are created, but the authors need to know what they write.
* When opening a pull request, reference the related issue (e.g., "related to #xxx", "for #xxx") instead of using closing keywords like "close #xxx", "closes #xxx", or "fixes #xxx".  We do not let PR and commit log comments to mandate the management.

tigercosmos:
This adds a Claude Code skill at `.claude/skills/create-pr/SKILL.md` that helps the author open a pull request following the modmesh PR protocol from issue #811. The skill walks through confirming scope, verifying branch state, drafting the subject and body for approval, checking title and body for banned closing keywords ("close", "fix", "resolve" forms followed by `#`), and calling `gh pr create` with text loaded from quoted heredocs so a subject containing backticks, `$`, `"`, or apostrophes does not need escaping.

PRs default to draft. Opening as ready-for-review triggers a mandatory global review-request comment step, because the "ready for review" button alone is not quotable in follow-ups. After creation the skill reminds the author to add inline annotations when the diff is more than one-liner-ish, and refuses to append attribution trailers since this project treats PR text as human-authored.

The skill also requires that PR description paragraphs are written as single unbroken lines separated by blank lines, rather than hard-wrapped at 79/80 columns. The source-code line-length limit does not belong in PR text; GitHub reflows to the viewer's width, and mid-sentence breaks render as ragged prose. This PR description itself follows the rule.

This PR is opened via the new `create-pr` skill.

For issue #811.

### DIFF
--- a/.claude/skills/create-pr/SKILL.md
+++ b/.claude/skills/create-pr/SKILL.md
@@ -5,9 +5,9 @@ description: Open a modmesh pull request that follows the project PR protocol (c
 
 # Create Pull Request (modmesh)
 
-Authoritative reference is issue solvcon/modmesh#811 and the "Pull Request
-Guidelines" section of `CLAUDE.md`. The rules below are a working summary;
-if those extend them, follow them and flag the drift.
+This file is the authoritative reference for the modmesh PR protocol.
+The "Pull Request Guidelines" section of `CLAUDE.md` is the project-wide
+cross-reference; flag any drift between the two.
 
 ## Protocol the PR must satisfy
 
--- a/.claude/skills/create-pr/SKILL.md
+++ b/.claude/skills/create-pr/SKILL.md
@@ -1,0 +1,147 @@
+---
+name: create-pr
+description: Open a modmesh pull request that follows the project PR protocol (concise subject, clear description, "related to #xxx" wording, draft by default, ready-for-review via global comment). Use when the user asks to create, open, or draft a pull request.
+---
+
+# Create Pull Request (modmesh)
+
+Authoritative reference is issue solvcon/modmesh#811 and the "Pull Request
+Guidelines" section of `CLAUDE.md`. The rules below are a working summary;
+if those extend them, follow them and flag the drift.
+
+## Protocol the PR must satisfy
+
+1. **Subject** -- concise and informative.
+2. **Description** -- clear global description. State *what* changed and
+   *why* in prose; use bullets for enumerated changes; include benchmark
+   tables or verification notes when relevant.
+3. **Issue reference** -- end with "Related to #xxx" or "For issue #xxx".
+   **Never** use "close #xxx", "closes #xxx", "fixes #xxx", or any closing
+   keyword. We do not let PR/commit text drive issue management.
+4. **Draft by default** -- open as draft unless the user explicitly says it
+   is ready for review. Pushing the "ready for review" button alone is
+   *not* a review request in this project.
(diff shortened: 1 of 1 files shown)


.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,171 @@
+---
+name: create-pr
+description: Open a modmesh pull request that follows the project PR protocol (concise subject, clear description, "related to #xxx" wording, draft by default, ready-for-review via global comment). Use when the user asks to create, open, or draft a pull request.
+---
+
+# Create Pull Request (modmesh)
+
+Authoritative reference is issue solvcon/modmesh#811 and the "Pull Request
```
tigercosmos: We can also remove this statement. I chose to keep it for now.

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+   button push is not quotable in follow-ups, so the comment is required.
+6. **Inline annotations** -- the author is expected to add inline review
+   annotations to guide the reviewer, unless the diff is one-liner-ish.
+   The skill should remind the user; it does not write the annotations.
+7. **Human authorship** -- all PR comments are written by humans. Tool
+   assistance is OK, but the user must know what the text says. When the
+   skill drafts text, present it for the user's review and edits before
+   posting.
+
+## Workflow
```
tigercosmos, line 42: Workflow confirmed while opening this PR.

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+   ```
+
+   Drop `--draft` only when the user has explicitly said the PR is
+   ready for review. Do not append a second issue-reference footer here;
+   the body from step 3 already has one. The `trap` ensures the temp
+   file is removed on any exit path. If the approved body would itself
+   contain the literal token `BODY` on its own line, swap the delimiter
+   to something unique (e.g. `BODY_EOF_811`).
+
+5. **After creation.** Report the PR URL back to the user. Then:
```
tigercosmos, line 128: My favor part.

tigercosmos (commented):
@yungyuc  Please take a look. I would suggest merging it first and revise it later based on real usage.

.claude/skills/create-pr/SKILL.md
yungyuc: Please remove this reference.  After merging this PR, this file becomes the authoritative reference.

yungyuc (changes requested):
- [x] Remove the reference to issue #811.

tigercosmos:
@yungyuc  please take a look. Thanks!

tigercosmos:
I like the skill:
Screenshot 2026-05-24 at 8 53 19 AM

yungyuc (approved):
LGTM.  I'll merge once CI finishes.

PR #834, opened 2026-05-26
Harden the `WrapBase` pybind11 binding architecture

yungyuc:
Fix a latent lifetime bug in expose_SimpleArray and `expose_SimpleArrayAsNdarray()`: the accessor was captured by reference into callbacks that pybind11 stores for the class lifetime.   Capture it by value (getter copies, setter takes ownership via `std::forward`) and drop the FIXME/NOLINT.  Behavior is unchanged for the current stateless call sites.

Make `WrapBase::commit()` variadic so extra pybind11 arguments (e.g. `buffer_protocol()`) can be forwarded through it.

### DIFF
--- a/cpp/modmesh/python/common.hpp
+++ b/cpp/modmesh/python/common.hpp
@@ -273,9 +273,14 @@ public:
     >;
     // clang-format on
 
-    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc)
+    template <class... Extra>
+    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc, Extra &&... extra)
     {
-        static wrapper_type derived(mod, pyname, pydoc);
+        // The static local is constructed exactly once, on the first call, so
+        // only the first call's pyname/pydoc/extra take effect; later calls
+        // return the already-registered singleton. Existing three-argument call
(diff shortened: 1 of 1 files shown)


cpp/modmesh/python/common.hpp
```diff
@@ -273,9 +273,14 @@ WrapBase
     >;
     // clang-format on
 
-    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc)
+    template <class... Extra>
+    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc, Extra &&... extra)
```
yungyuc, line 277: Use variadic template to make `commit()` extensible.

cpp/modmesh/python/common.hpp
```diff
@@ -332,22 +337,26 @@ WrapBase
     }
 
     template <typename Func>
-    // FIXME: NOLINTNEXTLINE(cppcoreguidelines-missing-std-forward)
     wrapper_type & expose_SimpleArray(char const * name, Func && f)
     {
         namespace py = pybind11;
 
-        using array_reference = typename std::invoke_result_t<Func, wrapped_type &>;
+        using array_reference = typename std::invoke_result_t<Func &, wrapped_type &>;
```
yungyuc, line 344: Fix the reference and ownership.

PR #839, opened 2026-05-28
ci: send mail when any profiling CI fails

### LINKED ISSUE #838: Send email on failure doesn't trigger when profiling CI fails
I saw that this profiling action fails but not sending any emails out. 
https://github.com/solvcon/modmesh/actions/runs/26534174355

I think it should be `contains(needs.*.result, 'failure')` not `(needs.*.result == 'failure')` in [the profiling action file](https://github.com/solvcon/modmesh/actions/runs/26534174355/workflow#L79), otherwise all dependencies need to fail before sending the mail.

ExplorerRay:
Related to https://github.com/solvcon/modmesh/issues/838.

Original condition only triggers email sending when both linux and macos fail. To fix this, I follow the way that other workflows used (`contains(needs.*.result, 'failure')`) for it.

### DIFF
--- a/.github/workflows/profiling.yml
+++ b/.github/workflows/profiling.yml
@@ -76,7 +76,7 @@ jobs:
 
   send_email_on_failure:
     needs: [profile]
-    if: ${{ always() && (needs.*.result == 'failure') && github.ref_name == 'master' && github.event.repository.fork == false }}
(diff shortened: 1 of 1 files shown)


yungyuc (approved):
LGTM

yungyuc:
All ubuntu jobs failed at updating cmake from kitware apt repository now.  We may need to take a look if it persists.

PR #841, opened 2026-05-29
Add element-wise eq() to SimpleArray

### LINKED ISSUE #810: Implement element-wise equality (`eq`) for `SimpleArray`
# Implement element-wise equality (`eq`) for `SimpleArray`

Follow-up to solvcon/modmesh#588 (matrix operations roadmap).

Element-wise equality is one remaining element-wise op that has not landed, while the add/sub/mul/div family has both array-array and array-scalar overloads in
`SimpleArrayMixinCalculators` (see [`cpp/modmesh/buffer/SimpleArray.hpp`](https://github.com/solvcon/modmesh/blob/2cbc729a26cff2ad22c95f9e289bfdc0532f0658/cpp/modmesh/buffer/SimpleArray.hpp#L975-L1013).

Unlike the other element-wise ops, `eq` returns a boolean array regardless
of the source dtype.

The element-wise function should be wrapped to Python but should not use the `__eq__()` special method name right now.  The binding to Python operators is future work.  New tests should be added adjacent to where other element-wise operations.

yungyuc:
Add `SimpleArray::eq()` with array-array and array-scalar overloads, each returning a `SimpleArray<bool>` regardless of the source dtype, completing the element-wise op family alongside add/sub/mul/div.  It is exposed to Python with the same name and without using `__eq__()`, so Python `==` remains to be object identity check for now.

Tests assert the resulting boolean arrays explicitly and cross-check against numpy, covering 1D/2D/3D, the scalar overload, bool sources, and that `==` is not element-wise.

For issue #810.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.cpp
+++ b/cpp/modmesh/buffer/SimpleArray.cpp
@@ -31,6 +31,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <string>
 #include <unordered_map>
 
 namespace modmesh
@@ -953,6 +954,27 @@ SimpleArrayPlex::~SimpleArrayPlex()
     m_instance_ptr = nullptr;
     m_has_instance_ownership = false;
 }
+
+namespace detail
+{
+
+// Format a shape such as (3, 4) for diagnostic messages.
+std::string format_shape(shape_type const & shape)
+{
+    std::string ret = "(";
+    for (size_t it = 0; it < shape.size(); ++it)
+    {
+        if (it != 0)
+        {
+            ret += ", ";
+        }
+        ret += std::to_string(shape[it]);
+    }
+    ret += ")";
+    return ret;
+}
+
+} /* end namespace detail */
 
 } /* end namespace modmesh */
 
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -48,6 +48,7 @@
 #include <numeric>
 #include <span>
 #include <stdexcept>
+#include <string>
 
 #ifdef _MSC_VER
 #include <BaseTsd.h>
@@ -125,6 +126,8 @@ namespace detail
 using shape_type = small_vector<size_t>;
 using sshape_type = small_vector<ssize_t>;
 using slice_type = small_vector<ssize_t>;
+
+std::string format_shape(shape_type const & shape);
 
 template <typename T>
 struct SimpleArrayInternalTypes
@@ -655,6 +658,9 @@ public:
     {
         return A(*static_cast<A const *>(this)).idiv(scalar);
     }
+
+    SimpleArray<bool> eq(A const & other) const;
+    SimpleArray<bool> eq(value_type scalar) const;
 
     A & iadd(A const & other)
     {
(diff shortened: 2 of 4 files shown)


cpp/modmesh/buffer/SimpleArray.cpp
```diff
@@ -954,6 +955,26 @@ SimpleArrayPlex::~SimpleArrayPlex()
     m_has_instance_ownership = false;
 }
 
+namespace detail
+{
+
+std::string format_shape(shape_type const & shape)
```
yungyuc, line 962: Add the free function helper in the cpp file.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2102,6 +2111,49 @@ class SimpleArray
     value_type * m_body = nullptr;
 }; /* end class SimpleArray */
 
+template <typename A, typename T>
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(A const & other) const
```
yungyuc, line 2112: Array-array equality test.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2102,6 +2111,49 @@ class SimpleArray
...
+        *ret_ptr = (*ptr == *other_ptr);
+        ++ptr;
+        ++other_ptr;
+        ++ret_ptr;
+    }
+    return ret;
+}
+
+template <typename A, typename T>
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(value_type scalar) const
```
yungyuc, line 2138: Array-scalar equality test.

tests/test_buffer.py
```diff
@@ -3304,6 +3304,199 @@ def _check_div_scalar_3d_type(type):
         _check_div_scalar_3d_type('float32')
         _check_div_scalar_3d_type('float64')
 
+    def test_eq_1d(self):
+        """Test 1D array element-wise equality"""
+        def _check_eq_1d_type(type):
+            arr1 = [1, 2, 3, 4, 5, 6, 7, 8]
+            arr2 = [1, 0, 3, 0, 5, 0, 7, 0]
+            res = [True, False, True, False, True, False, True, False]
```
yungyuc, line 3312: The expected result.

tests/test_buffer.py
```diff
@@ -3304,6 +3304,199 @@ def _check_div_scalar_3d_type(type):
         _check_div_scalar_3d_type('float32')
         _check_div_scalar_3d_type('float64')
 
+    def test_eq_1d(self):
```
yungyuc, line 3307: Test for 1d, 2d and array-array, array-scalar.

tests/test_buffer.py
```diff
@@ -3304,6 +3304,199 @@ def _check_div_scalar_3d_type(type):
...
+        _check_eq_scalar_2d_type('int32')
+        _check_eq_scalar_2d_type('int64')
+        _check_eq_scalar_2d_type('uint8')
+        _check_eq_scalar_2d_type('uint16')
+        _check_eq_scalar_2d_type('uint32')
+        _check_eq_scalar_2d_type('uint64')
+        _check_eq_scalar_2d_type('float32')
+        _check_eq_scalar_2d_type('float64')
+
+    def test_eq_3d(self):
```
yungyuc, line 3432: 3D array test.

tests/test_buffer.py
```diff
@@ -3304,6 +3304,199 @@ def _check_div_scalar_3d_type(type):
...
+
+        res_scalar = [True, True, False, False]
+        nres_scalar = np.equal(narr1, True)
+        sres_scalar = sarr1.eq(True)
+        self.assertEqual(sres_scalar.ndarray.dtype, np.bool_)
+        for i in range(len(res_scalar)):
+            self.assertEqual(sres_scalar[i], res_scalar[i])
+            self.assertEqual(sres_scalar[i], nres_scalar[i])
+
+    def test_eq_does_not_bind_operator(self):
```
yungyuc, line 3484: Make sure `eq()` does not bind to `__eq__()`.

tests/test_buffer.py
```diff
@@ -3304,6 +3304,199 @@ def _check_div_scalar_3d_type(type):
...
+        _check_eq_2d_type('int8')
+        _check_eq_2d_type('int16')
+        _check_eq_2d_type('int32')
+        _check_eq_2d_type('int64')
+        _check_eq_2d_type('uint8')
+        _check_eq_2d_type('uint16')
+        _check_eq_2d_type('uint32')
+        _check_eq_2d_type('uint64')
+        _check_eq_2d_type('float32')
+        _check_eq_2d_type('float64')
```
tigercosmos, line 3375: @yungyuc I think looping through a list of types can make the code shorter. The list can also be resued.

yungyuc: Testing code should avoid loops.  When it fails, the line should precisely point to the call path.

PR #843, opened 2026-05-30
ci: wrap workflow run blocks in log groups

### LINKED ISSUE #782: Wrap more CI steps in a workflow log groups
Follow the approach in PR #781 (for issue #778), use [the pairs of `::group::` and `::endgroup::` markers around the run-block body](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines) to beautify step output that is too long to quickly skim by human eyes.

rockleona:
Related to #782

## Summary
 
 Wrap the remaining multi-line GitHub Actions `run:` blocks with
 `::group::` / `::endgroup::` so long CI logs are folded by default.
 
 ## Changes
 
 - Add log grouping to the remaining workflow steps in:
   - `.github/workflows/devbuild.yml`
   - `.github/workflows/lint.yml`
   - `.github/workflows/nouse_install.yml`
   - `.github/workflows/profiling.yml`
   - `.github/workflows/send_email_on_fail.yml`
 - Keep the workflow behavior unchanged; only improve log readability

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -53,9 +53,7 @@ jobs:
 
       - name: event name
         run: |
-          echo "::group::event name"
           echo "github.event_name: ${{ github.event_name }}"
-          echo "::endgroup::"
 
       - name: setup dependencies
         uses: ./.github/actions/setup_linux
@@ -118,9 +116,7 @@ jobs:
 
     - name: event name
       run: |
-        echo "::group::event name"
         echo "github.event_name: ${{ github.event_name }}"
-        echo "::endgroup::"
 
     - name: setup dependencies for Linux
       if: runner.os == 'Linux'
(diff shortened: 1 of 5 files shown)


yungyuc (changes requested):
@rockleona Did you add the groups for everything, or have some contents missing?  It's OK to include partial contents.  Please leave inline annotations in the PR diff for what you include and what you do not include.

.github/workflows/devbuild.yml
rockleona, line 56: But if there is only a single `echo` line, I won't place `::group::` annotation. Since I think it is annoying if there's only a single line after open the collapse.

rockleona (commented):
I've scan all `.yml` files, all stages with `run` will have `::group::` annotation.

yungyuc (approved):
LGTM

@ExplorerRay @chestercheng could you please help take a look?

ExplorerRay (approved):
LGTM

PR #845, opened 2026-05-30
Fix JSON deserialization of empty containers and string-embedded delimiters

tigercosmos:
The hand-rolled JSON deserializer in SerializableItem.cpp could not parse several valid inputs and silently accepted some invalid ones. Empty arrays `[]` and empty objects `{}` were rejected outright, and any primitive string value embedding a structural delimiter (`,`, `]`, `}`) was truncated at the first delimiter because the value scanner did not track quoted strings. Both parse_array and parse_object now scan leaf values with quoted-string and backslash-escape awareness, and accept empty containers.

The same string-blindness affected nested containers: the loops that extracted a nested `{...}` or `[...]` counted raw braces and brackets, so a string value such as `"a]b"` inside a nested array still split the container at the wrong place. Those three loops are replaced by a single shared scan_balanced_expression helper that ignores string contents (and their escapes) when matching the closing delimiter, so nested structures like `{"k":["a]b"]}` parse correctly.

Trailing commas are now rejected rather than silently tolerated: an empty container stays valid, but a closing delimiter immediately following a comma raises a parse error, keeping the parser aligned with strict JSON.

The change adds twenty gtest cases under the Json suite covering empty and whitespace-only containers, leaf and nested string values that embed each delimiter, escaped quotes and backslashes through both the leaf and container scanners, trailing/leading/double-comma rejection, unterminated nested containers, and an end-to-end round trip. A few pre-existing weaknesses are intentionally deferred: the End state still accepts a single trailing character after the closing bracket, malformed end-of-input can leave an empty value expression before classification, and number classification remains a character-class check.

### DIFF
--- a/cpp/modmesh/serialization/SerializableItem.cpp
+++ b/cpp/modmesh/serialization/SerializableItem.cpp
@@ -105,6 +105,66 @@ std::string trim_string(const std::string & str)
         column += 1;                     \
     }
 
+// Scan a balanced {...} or [...] starting at json[index], which must be the
+// opening character. Appends the whole balanced expression to out and leaves
+// index on the matching closing character. Characters inside quoted strings
+// (with backslash escapes) are ignored when matching, so brackets or braces
+// embedded in string values do not unbalance the scan. Throws when the
+// container is not closed.
+inline void scan_balanced_expression(const std::string & json, size_t & index, char open_char, char close_char, int & line, int & column, std::string & out)
+{
+    int depth = 0;
+    bool in_string = false;
+    bool escape = false;
+
+    while (index < json.size())
+    {
+        const char c_curr = json[index];
+        out.push_back(c_curr);
+
+        if (escape)
+        {
+            escape = false;
+        }
+        else if (in_string)
+        {
+            if (c_curr == '\\')
+            {
+                escape = true;
+            }
+            else if (c_curr == '"')
+            {
+                in_string = false;
+            }
+        }
+        else if (c_curr == '"')
+        {
+            in_string = true;
+        }
+        else if (c_curr == open_char)
+        {
+            depth += 1;
+        }
+        else if (c_curr == close_char)
+        {
+            depth -= 1;
+        }
+
+        if (!in_string && depth == 0)
+        {
+            break;
+        }
+
+        MM_DECL_CACULATE_LINE_COLUMN(json[index])
+        index += 1;
+    }
+
+    if (depth != 0)
+    {
+        throw_serialization_error("Invalid JSON format: missing closing bracket.", line, column);
+    }
+}
+
 // NOLINTBEGIN(readability-function-cognitive-complexity)
 // NOLINTNEXTLINE(misc-no-recursion)
 JsonArray JsonNode::parse_array(const std::string & json)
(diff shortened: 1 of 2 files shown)


cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -105,15 +105,74 @@ std::string trim_string(const std::string & str)
         column += 1;                     \
     }
 
+// Scan a balanced {...} or [...] starting at json[index], which must be the
+// opening character. Appends the whole balanced expression to out and leaves
+// index on the matching closing character. Characters inside quoted strings
+// (with backslash escapes) are ignored when matching, so brackets or braces
+// embedded in string values do not unbalance the scan. Throws when the
+// container is not closed.
+inline void scan_balanced_expression(const std::string & json, size_t & index, char open_char, char close_char, int & line, int & column, std::string & out)
```
tigercosmos, line 114: Extract the reused logic.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -196,12 +244,36 @@ JsonArray JsonNode::parse_array(const std::string & json)
 
                 // we assume the value is a string, number, boolean, or null, and the expression is correct
                 // if the expression is not correct, the exception will be thrown when parsing the value later
+                // Track quoted-string state so that ',' or ']' inside a
+                // string value does not terminate the scan early.
```
tigercosmos, line 248: Fix the issue.

cpp/modmesh/serialization/SerializableItem.cpp
```diff
@@ -445,12 +483,36 @@ JsonMap JsonNode::parse_object(const std::string & json)
 
                 // we assume the value is a string, number, boolean, or null, and the expression is correct
                 // if the expression is not correct, the exception will be thrown when parsing the value later
+                // Track quoted-string state so that ',' or '}' inside a
+                // string value does not terminate the scan early.
```
tigercosmos, line 487: Fix the issue.

gtests/test_nopython_serializable.cpp
```diff
@@ -322,6 +322,225 @@ TEST(Json, deserialize_with_unordered_map)
     EXPECT_EQ(item.pet_map["cat"].is_cat, true);
 }
 
+// The tests below cover empty containers, lenient trailing commas, and
+// string values that embed JSON delimiters (',', ']', '}') or escape
+// sequences. They exercise the parser directly through detail::JsonNode so
+// that the produced structure can be inspected precisely.
+
+TEST(Json, parse_empty_array)
```
tigercosmos, line 330: More test cases to increae the coverage.

tigercosmos (commented):

gtests/test_nopython_serializable.cpp
yungyuc: I'd like the tests to be moved to Python, which requires the Python wrapper for the serialization code.  Please help create an issue to track the work (it's future work that does not need to be included in this PR).

yungyuc (approved):
The fix looks good to me.

A derived issue: The JSON processing (serialization) code should live in C++ for performance and flexibility, but the tests should be in Python for ease of maintenance and natural documentation.  It should be fixed in the future.

gtests/test_nopython_serializable.cpp
tigercosmos: @yungyuc  Sure, let me move all serialization test to python in the next PR.

PR #820, opened 2026-05-24
build: skip CI for pull requests with skip-ci label or inline `[skip-ci]` control string

yungyuc:
Add a reusable check_skip_ci workflow that gates the devbuild, lint, and nouse_install jobs.  CI is skipped when:
- a pull request carries the "skip-ci" label, or
- when its description or a repository member's comment has "`[skip-ci]`" alone on a line

I tested the two scenarios in a separate repository.  This PR intends to run the CI.

Toggling the label re-evaluates via the labeled/unlabeled triggers; push and schedule events keep their existing gating.  The "skip-ci" label must exist in the repository.

For issue #814.

PR #850, opened 2026-06-02
Move macOS CI from intel to arm64 macos-26

### LINKED ISSUE #846: CI lint job keeps timing out on `macos-26-intel`
The [CI linting (including tidying) job](https://github.com/solvcon/modmesh/blob/32e5cd9e523f0f00f17845e78b27751cb18b2427/.github/workflows/lint.yml#L23) has been failing on the macos platform with the 45-min timeout for a while.  It has not been successfully run through for maybe more than a week.  The latest failure log is https://github.com/solvcon/modmesh/actions/runs/26753256104/job/78846413292 .

It only fails on mac (`macos-26-intel`).  The linting job on ubuntu 24.04 does not time out.

The job runs on `macos-26-intel` and the VM provides 14 GB of memory, according to https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories .

tigercosmos:
GitHub is retiring the intel macOS runners, so this moves the macOS jobs in both the lint (lint.yml) and devbuild (devbuild.yml) workflows from macos-26-intel to the arm64 macos-26 image. The arm64 image has 7 GB of RAM rather than the intel image's 14 GB, and the default unbounded `make -j` oversubscribes it: in the lint job clang-tidy runs per translation unit alongside the compiler and the build runs out of memory, while in the devbuild job the build spawns roughly two dozen parallel clang processes that swap until the job trips its 45-minute timeout. Both workflows now cap macOS build parallelism to -j 2 (Linux keeps the unbounded default); a full build at -j 2 was verified to complete with zero swap.

Switching to arm64 also changes which code clang-tidy sees: it now analyzes the NEON SIMD translation unit that the intel runner never compiled. That surfaced 27 diagnostics which are all false positives on inspection -- methods flagged as convertible to static or const that in fact read or write the members in question, constructors flagged for fields they do initialize, if-constexpr branches reported as clones although they differ by element type, and decltype trailing return types that are required for SFINAE. Each is suppressed with a targeted NOLINT and a short rationale, matching the NOLINT convention already used in the tree. Separately, the pilot binary target is given the same pybind11 and apple accelerate/veclib -Wno suppressions that the modmesh library already carries, because pilot.cpp is linted under -warnings-as-errors=* and pulls in the same SDK headers.

Related to #846.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -94,6 +94,12 @@ jobs:
     timeout-minutes: ${{ fromJSON(vars.MMGH_TIMEOUT_BUILD || '45') }}
 
     env:
+      # Cap build parallelism. The github-hosted runners have only 3 (macOS)
+      # / 4 (ubuntu) virtual cores shared with co-tenant VMs, so the default
+      # unbounded `make -j` oversubscribes memory and CPU; on the 7 GB arm64
+      # macOS runner that swaps until the job hits its timeout. -j2 is the
+      # safe bet on both.
+      MAKE_PARALLEL: -j2
       QT_DEBUG_PLUGINS: 1
       QT_QPA_PLATFORM: offscreen # for Ubuntu runner
       PIP_BREAK_SYSTEM_PACKAGES: 1 # disabling PEP668 (for MacOS runner)
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,6 +40,11 @@ jobs:
 
     env:
       JOB_MAKE_ARGS: VERBOSE=1 BUILD_QT=ON USE_CLANG_TIDY=ON LINT_AS_ERRORS=ON
+      # Cap build parallelism. The github-hosted runners have only 3 (macOS)
+      # / 4 (ubuntu) virtual cores shared with co-tenant VMs, so the default
+      # unbounded `make -j` oversubscribes memory and CPU; on the 7 GB arm64
+      # macOS runner clang-tidy then OOMs. -j2 is the safe bet on both.
+      MAKE_PARALLEL: -j2
       QT_DEBUG_PLUGINS: 1
       QT_QPA_PLATFORM: offscreen # for Ubuntu runner
       PIP_BREAK_SYSTEM_PACKAGES: 1 # disabling PEP668 (for MacOS runner)
@@ -50,8 +55,9 @@ jobs:
     strategy:
       matrix:
         # https://github.com/actions/runner-images/blob/main/images/macos/macos-26-Readme.md
-        # use macos-26-intel to have bigger 14 GB RAM; otherwise, macos-26-arm64 only has 7GB RAM
-        os: [ubuntu-24.04, macos-26-intel]
+        # macos-26 is arm64 with 7 GB RAM; build parallelism is capped via
+        # MAKE_PARALLEL in the env block above so clang-tidy does not OOM.
+        os: [ubuntu-24.04, macos-26]
         cmake_build_type: [Debug]
 
       fail-fast: false
--- a/cpp/binary/pilot/CMakeLists.txt
+++ b/cpp/binary/pilot/CMakeLists.txt
@@ -75,6 +75,19 @@ target_compile_options(
     ${COMMON_COMPILER_OPTIONS}
 )
 
+if(APPLE)
+    # pilot.cpp pulls in apple's accelerate/veclib headers, whose CF_ENUM
+    # macros trip clang-diagnostic-elaborated-enum-base, and the lint job
+    # runs clang-tidy with -warnings-as-errors=*. modmesh_primary already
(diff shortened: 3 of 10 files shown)


.github/workflows/devbuild.yml
```diff
@@ -137,6 +137,15 @@ jobs:
...
         create-symlink: true
 
+    # The arm64 macos-26 runner has only ~7 GB RAM and few cores. The
+    # default unbounded `make -j` launches ~24 parallel clang processes,
+    # which oversubscribes memory, swaps, and trips the 45-minute job
+    # timeout. Cap macOS to -j 2 for every make below (the Makefile reads
+    # MAKE_PARALLEL from the environment); Linux keeps the unbounded -j.
+    - name: cap build parallelism on macOS
+      if: runner.os == 'macOS'
+      run: echo "MAKE_PARALLEL=-j 2" >> "$GITHUB_ENV"
```
tigercosmos: meat

.github/workflows/lint.yml
```diff
@@ -118,8 +119,11 @@ jobs:
     - name: make pilot
       run: |
         echo "::group::make pilot"
+        MAKE_PARALLEL="-j"
+        if [ "${{ runner.os }}" == "macOS" ]; then MAKE_PARALLEL="-j 2"; fi
```
tigercosmos: meat

tigercosmos (commented):
@yungyuc  Please take a look.

.github/workflows/devbuild.yml
yungyuc, line 96: Environment variable section.

.github/workflows/devbuild.yml
yungyuc: Github Actions runner machines have 3/4 virtual cores on mac and ubuntu, respectively: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories .  Oversubscription may also happen on ubuntu, so we can set the environment variable `MAKE_PARALLEL` the same on both platform [at line 96](https://github.com/solvcon/modmesh/pull/850/changes#diff-e965449d73f20540cbc710aa8e44b689c636143f0986b9ed679eae7e588e99fbR96):

```yaml
    env:
      QT_DEBUG_PLUGINS: 1
      QT_QPA_PLATFORM: offscreen # for Ubuntu runner
      PIP_BREAK_SYSTEM_PACKAGES: 1 # disabling PEP668 (for MacOS runner)
      # Fix issue: https://github.com/solvcon/modmesh/issues/366
      # Use custom config for jurplel/install-qt-action@v4
      AQT_CONFIG: "thirdparty/aqt_settings.ini"
```

`-j2` is a safe bet for the virtual cores are shared with other VM/containers for sure.  `-j4` oversubscribe a little bit.

yungyuc (changes requested):
- [x] Set the environment variable `MAKE_PARALLEL` homogeneously in the dedicated section.

yungyuc:
@rockleona Please take a look.

yungyuc:
It seems that the CI is now triggered by commenting and it should be fixed too.  It can be done with this or a separate PR/issue.

tigercosmos:
> It seems that the CI is now triggered by commenting and it should be fixed too. It can be done with this or a separate PR/issue.

It seems to be related to https://github.com/solvcon/modmesh/pull/820

I suggest handling this in the next PR.

tigercosmos:
@yungyuc Please take a look.

.github/workflows/lint.yml
```diff
@@ -40,6 +40,11 @@ jobs:
 
     env:
       JOB_MAKE_ARGS: VERBOSE=1 BUILD_QT=ON USE_CLANG_TIDY=ON LINT_AS_ERRORS=ON
+      # Cap build parallelism. The github-hosted runners have only 3 (macOS)
+      # / 4 (ubuntu) virtual cores shared with co-tenant VMs, so the default
+      # unbounded `make -j` oversubscribes memory and CPU; on the 7 GB arm64
+      # macOS runner clang-tidy then OOMs. -j2 is the safe bet on both.
+      MAKE_PARALLEL: -j2
```
rockleona, line 47: Quite curious, why apply `-j2` to both of the runner?`ubuntu` should be fine in both `lint` and `devbuild` actions.

yungyuc: Ubuntu just happened to be fine.  It's still in a VM/container with virtual CPU.  That's why I asked to unify the parallel setting: https://github.com/solvcon/modmesh/pull/850#discussion_r3345390423

yungyuc (approved):
LGTM.

rockleona (approved):
LGTM, I also ran @tigercosmos commit on my repo, both action `devbuild` and `lint` are passed.

devbuild: https://github.com/rockleona/modmesh/actions/runs/26859515160
lint: https://github.com/rockleona/modmesh/actions/runs/26859515131

PR #855, opened 2026-06-03
Fix SimpleArrayComplex slice assignment from ndarray

### LINKED ISSUE #854: SimpleArrayComplex misclassifies ndarray-to-slice assignment as scalar assignment
The following code is the origin of bug.
https://github.com/solvcon/modmesh/blob/6fca1a59f459cdf6c8a26b8de573196253975db8/cpp/modmesh/buffer/pymod/array_common.hpp#L237
When complex numpy array is assigned to `modmesh.SimpleArrayComplex` by `__setitem__`, the program treats the array as a scalar. The scalar check uses `is_complex_v<T>`, which only reflects the destination SimpleArray element type. For `Complex<T>` arrays, this makes any RHS look scalar-like, including NumPy ndarrays, instead of checking whether `py_value` is actually a scalar.

## How to reproduce the bug

```shell
$ python3 - <<'PY'
import numpy as np
import modmesh

sarr = modmesh.SimpleArrayComplex128((2, 3))

rhs = np.arange(2 * 3, dtype='float64').reshape((2, 3))
rhs = (rhs + 3j).astype('complex128')

sarr[:, :] = rhs
np.testing.assert_array_equal(rhs, sarr.ndarray)
PY
  File "<stdin>", line 9, in <module>
RuntimeError: Unable to cast Python instance of type <class 'tuple'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```

ThreeMonth03:
The following code line cannot distinguish whether `py_value` is actually a scalar for a complex array.
https://github.com/solvcon/modmesh/blob/6fca1a59f459cdf6c8a26b8de573196253975db8/cpp/modmesh/buffer/pymod/array_common.hpp#L237

Thus, this pull request correct the logic of `is_number` by checking whether `py_value` is a sequence at first.

Related to #854.

### DIFF
--- a/cpp/modmesh/buffer/pymod/array_common.hpp
+++ b/cpp/modmesh/buffer/pymod/array_common.hpp
@@ -177,21 +177,6 @@ inline modmesh::detail::shape_type make_shape(pybind11::object const & shape_in)
     return shape;
 }
 
-template <typename T>
-struct convert_to_std
-{
-    using type = T;
-};
-
-template <typename T>
-struct convert_to_std<Complex<T>>
-{
-    using type = std::complex<T>;
-};
-
-template <typename T>
-using convert_to_std_t = typename convert_to_std<T>::type;
-
 /// Helper class for array property in Python.
 template <typename T>
 class ArrayPropertyHelper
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -177,21 +177,6 @@ inline modmesh::detail::shape_type make_shape(pybind11::object const & shape_in)
...
-
-template <typename T>
-struct convert_to_std<Complex<T>>
-{
-    using type = std::complex<T>;
-};
-
-template <typename T>
-using convert_to_std_t = typename convert_to_std<T>::type;
-
```
ThreeMonth03, line 194: Remove the snifae since we only check whether the type of a variable is `Complex<double>` and `Complex<float>`. We can enumerate these two cases.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -230,46 +215,27 @@ class ArrayPropertyHelper
             const py::object & py_key = args[0];
             const py::object & py_value = args[1];
 
-            // Determine if py_value is a numpy complexx
-            // Reference: https://stackoverflow.com/a/79673025
-            py::object builtins = py::module_::import("builtins"); // NOLINT(misc-const-correctness)
-            py::object complex_class = builtins.attr("complex"); // NOLINT(misc-const-correctness)
-            const bool is_number = py::isinstance<py::bool_>(py_value) || py::isinstance<py::int_>(py_value) || py::isinstance<py::float_>(py_value) || is_complex_v<T> || py::isinstance(py_value, complex_class);
+            const bool is_sequence_value = is_sequence(py_value);
+            const bool is_scalar_value = is_scalar(py_value);
```
ThreeMonth03, line 219: Move logic to private static functions.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -343,6 +309,57 @@ class ArrayPropertyHelper
...
+               pybind11::isinstance<pybind11::array>(py_value) ||
+               pybind11::isinstance<pybind11::tuple>(py_value);
+    }
+
+    static bool is_scalar(pybind11::object const & py_value)
+    {
+        if (is_sequence(py_value))
+        {
+            return false;
+        }
```
ThreeMonth03, line 324: If it is a sequence, it is not a scalar.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -343,6 +309,57 @@ class ArrayPropertyHelper
...
+
+        if constexpr (std::is_same_v<T, Complex<float>> || std::is_same_v<T, Complex<double>>)
+        {
+            return is_number || pybind11::isinstance<T>(py_value);
+        }
+        else
+        {
+            return is_number;
+        }
+    }
```
ThreeMonth03, line 336: If LHS is `modmesh.SimpleArrayComplex`, `py_value` should be either a python number or `modmesh::complex`.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -343,6 +309,57 @@ class ArrayPropertyHelper
...
+        }
+        else if constexpr (std::is_same_v<T, Complex<double>>)
+        {
+            return cast_complex_scalar<double>(py_value);
+        }
+        else
+        {
+            return py_value.cast<T>();
+        }
+    }
```
ThreeMonth03, line 361: Cast complex number without Snifae.

ThreeMonth03 (commented):

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -343,6 +309,57 @@ class ArrayPropertyHelper
...
+
+    template <typename U>
+    static Complex<U> cast_complex_scalar(
+        pybind11::object const & py_value)
+    {
+        pybind11::object const complex_class =
+            pybind11::module_::import("builtins").attr("complex");
+        return complex_class(py_value).cast<std::complex<U>>();
```
tigercosmos, line 344: It loades everytime. Consider to cache it.

cpp/modmesh/buffer/pymod/array_common.hpp
```diff
@@ -343,6 +309,57 @@ class ArrayPropertyHelper
...
+    }
+
+    static bool is_scalar(pybind11::object const & py_value)
+    {
+        if (is_sequence(py_value))
+        {
+            return false;
+        }
+
+        bool const is_number = PyNumber_Check(py_value.ptr());
```
tigercosmos, line 326: `PyNumber_Check` seems better than the old way.

yungyuc: Agree.  Since we are dealing with a Python object, using the CPython API at runtime makes perfect sense.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: Caching it may need to hash and store a hash table.  The overhead does not seem to be worthy.

cpp/modmesh/buffer/pymod/array_common.hpp
yungyuc: The new implmentation looks much less cryptic.

yungyuc (approved):
LGTM

PR #867, opened 2026-06-06
Expose raw span and strided mdspan for SimpleArray

### LINKED ISSUE #784: Support non-contiguous buffers in `SimpleArray::as_span/as_mdspan()`
In issue #771 and PR #777, `as_span()` and `as_mdspan`() are added to `SimpleArray` support only C-contiguous memory buffers.

This issue is created to support non-contiguous memory buffer in the `as_span/as_mdspan()` interface, as mentioned in https://github.com/solvcon/modmesh/pull/777#pullrequestreview-4295013976.

ThreeMonth03:
API `as_span()` and `as_mdspan()` only support for non-contiguous SimpleArray now, so this pull request expose raw span and strided mdspan for SimpleArray.

Related to #784.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -49,6 +49,7 @@
 #include <span>
 #include <stdexcept>
 #include <string>
+#include <utility>
 
 #ifdef _MSC_VER
 #include <BaseTsd.h>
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1942,53 +1943,37 @@ class SimpleArray
...
     }
     std::span<value_type const> as_span() const
     {
-        if (!is_c_contiguous())
-        {
-            throw std::runtime_error("SimpleArray::as_span: array is not C-contiguous");
-        }
-        return std::span<value_type const>(data(), size());
+        return std::span<value_type const>(data(), buffer().nbytes() / ITEMSIZE);
     }
```
ThreeMonth03, line 1951: `as_span()` cannot return elements based on strides, so I modify return content. It returns whole buffer now.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1942,53 +1943,37 @@ class SimpleArray
...
-        {
-            throw std::runtime_error("SimpleArray::as_mdspan: array is not C-contiguous");
-        }
-        std::array<size_t, N> exts;
-        for (size_t i = 0; i < N; ++i) { exts[i] = shape(i); }
-        return std::mdspan<value_type, std::dextents<size_t, N>>(data(), exts);
+        return std::mdspan<value_type, std::dextents<size_t, N>, std::layout_stride>(
+            data(),
+            make_mdspan_mapping<N>());
     }
```
ThreeMonth03, line 1963: `as_mdspan()` will return elements based on `layout_stride::mapping`, which includes strides and shapes.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Is it a requirement for `span` by the C++ standard?

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: If you think

```cpp
            data(), make_mdspan_mapping<N>());
```

is more readable than the current

```cpp
            data(),
            make_mdspan_mapping<N>());
```

use the shorter form.  So to the similar code in the rest of this PR.

yungyuc (approved):
There is only a formatting point:
- [ ] Discussion: one or more lines for passing arguments.

I am OK either way.  If you do not want to change it, let me know and I'll directly merge.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Yes, it is. According to the [implementation of gcc-15.1.0](https://gcc.gnu.org/onlinedocs/gcc-15.1.0/libstdc%2B%2B/api/a00206_source.html), the type of extent is `size_t`, which implies that it doesn't take strides into consider. You can find the following code at line171.
```c++
      // member constants
      static constexpr size_t extent = _Extent;
 
      // constructors, copy and assignment
 
      constexpr
      span() noexcept
      requires (_Extent == dynamic_extent || _Extent == 0)
      : _M_ptr(nullptr), _M_extent(__v<0>)
      { }
```

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Thanks for reminding. I didn't notice the slight difference.
There are only two arguments here, so it is needless to emunerate two arguments in multiple lines, and I rearrange them to one line.

ThreeMonth03 (commented):
- [x]  Discussion: one or more lines for passing arguments.

@yungyuc Thanks for reminding. I adjust the format now.

yungyuc (approved):

PR #868, opened 2026-06-06
Add diff-based pilot clang-tidy check for PRs

### LINKED ISSUE #676: Reduce PR lint time by checking changed files
Currently, the [lint](https://github.com/solvcon/modmesh/blob/master/.github/workflows/lint.yml) workflow in PR takes a long time.
This is because it runs lint checks on the entire codebase, even when only a few files are changed.

The goal of this issue is to reduce the average turnaround time of PR CI runs to under 5 minutes, so that PR checks can finish faster and developer iteration can be improved.

As a first step, we can start with the `make pilot` step, since `clang-tidy` is one of the more expensive checks in PR CI.

Instead of running `clang-tidy` over the entire codebase for every PR, PR linting could be made incremental by checking only the files or lines changed by the PR, for example via `clang-tidy-diff.py`.

### LINKED ISSUE #679: Improve cmakelists to clang-tidy all and selected files
Clarify how `make pilot` works with `USE_CLANG_TIDY=ON`, and find a way to make it work in two modes:
1. Tidy all files
2. Tidy a selected list of files provided from command line

Ref to the comment from @ExplorerRay: https://github.com/solvcon/modmesh/issues/676#issuecomment-3864128962

chestercheng:
Related to #676.

This PR introduces a diff-based `pilot_clang_tidy_diff` target for pull
request linting. The target runs clang-tidy-diff.py against pilot files
changed from `MODMESH_DIFF_BASE`.

PR lint jobs now use this target, while non-PR lint jobs continue running the
full `make pilot` and `make run_pilot_pytest`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -60,7 +60,7 @@ jobs:
 
     - uses: actions/checkout@v6
       with:
-        fetch-depth: 1
+        fetch-depth: ${{ github.event_name == 'pull_request' && 100 || 1 }}
 
     - name: event name
       run: |
@@ -101,6 +101,7 @@ jobs:
       run: make flake8
 
     - name: make pilot
+      if: github.event_name != 'pull_request'
       run: |
         make pilot \
           ${JOB_MAKE_ARGS} \
@@ -108,12 +109,32 @@ jobs:
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
 
     - name: make run_pilot_pytest
+      if: github.event_name != 'pull_request'
       run: |
         export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
         make run_pilot_pytest \
           ${JOB_MAKE_ARGS} \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+
+    - name: make pilot_clang_tidy_diff
+      if: github.event_name == 'pull_request'
+      run: |
+        make pilot_clang_tidy_diff \
+          ${JOB_MAKE_ARGS} \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DCMAKE_EXPORT_COMPILE_COMMANDS=ON" \
+          MODMESH_DIFF_BASE=${{ github.event.pull_request.base.sha }}
+
+    - name: make pilot_clang_tidy_diff with Python library path
+      if: github.event_name == 'pull_request'
+      run: |
+        export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
+        make pilot_clang_tidy_diff \
+          ${JOB_MAKE_ARGS} \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DCMAKE_EXPORT_COMPILE_COMMANDS=ON" \
+          MODMESH_DIFF_BASE=${{ github.event.pull_request.base.sha }}
 
   send_email_on_failure:
     needs: [lint]
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,15 @@ pyprof: buildext $(PROFFILES)
 pilot: cmake
 	cmake --build $(BUILD_PATH) --target $@ VERBOSE=$(VERBOSE) $(MAKE_PARALLEL)
 
+.PHONY: pilot_clang_tidy_diff
+pilot_clang_tidy_diff: cmake
+	@test -n "$(MODMESH_DIFF_BASE)" || { \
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -160,6 +156,15 @@ pyprof: buildext $(PROFFILES)
 pilot: cmake
 	cmake --build $(BUILD_PATH) --target $@ VERBOSE=$(VERBOSE) $(MAKE_PARALLEL)
 
+.PHONY: pilot_clang_tidy_diff
+pilot_clang_tidy_diff: cmake
+	@test -n "$(MODMESH_DIFF_BASE)" || { \
+		echo "Error: MODMESH_DIFF_BASE is required."; \
+		exit 1; \
+	}
+	env MODMESH_DIFF_BASE="$(MODMESH_DIFF_BASE)" \
```
chestercheng, line 165: `MODMESH_DIFF_BASE` is passed as env because it changes per PR run and should not be stored in the cmake cache

.github/workflows/lint.yml
```diff
@@ -66,7 +60,7 @@ jobs:
 
     - uses: actions/checkout@v6
       with:
-        fetch-depth: 1
+        fetch-depth: ${{ github.event_name == 'pull_request' && 100 || 1 }}
```
chestercheng, line 63: PR runs need enough history to diff against the base sha, non-PR runs keep the shallow checkout

chestercheng:
The CI job fails during cmake configuration because pybind11 cannot be found, which seems unrelated to the changes in this PR.

I'll check it tomorrow.

yungyuc:
Could also be instability at Github.  I did not know what triggered the rerun, but it's just triggered.

.github/workflows/lint.yml
```diff
@@ -92,53 +86,55 @@ jobs:
...
       run: |
-        echo "::group::make run_pilot_pytest"
         export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
         make run_pilot_pytest \
           ${JOB_MAKE_ARGS} \
           CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
-        echo "::endgroup::"
+
+    - name: make pilot_clang_tidy_diff
```
ExplorerRay, line 120: I noticed that `make pilot_clang_tidy_diff` is run twice here: once with the Python library path, and once without it. Could you help me understand the test scope of `clang-tidy-diff` and why it is split this way?

cpp/binary/pilot/CMakeLists.txt
```diff
@@ -123,8 +110,41 @@ endif()
...
+    endif()
+endif()
+
+add_custom_target(
+    pilot_clang_tidy_diff
+    COMMAND
+        bash
+        -e
+        -o pipefail
+        -c "test -n \"$MODMESH_DIFF_BASE\"; git diff --no-color -U0 \"$MODMESH_DIFF_BASE\" HEAD -- $<JOIN:${MODMESH_PILOT_FILES}, > | \"${CLANG_TIDY_DIFF_EXE}\" -p1 -path \"${CMAKE_BINARY_DIR}\" -config-file \"${PROJECT_ROOT_DIR}/.clang-tidy\""
```
ExplorerRay: I am wondering whether it's necessary to put `test -n $MODMESH_DIFF_BASE` in both here and `pilot_clang_tidy_diff` target in Makefile. I think current setting will make this variable be checked two times.

yungyuc: This line is way too long: 244 chars!  It is too hard to check and edit by naked eyes.  How about making it a script in `contrib/` (maybe `contrib/lint/`)?

yungyuc (changes requested):
- [ ] A single line of 244 chars in CMakeLists.txt is too long for maintenance.

.github/workflows/lint.yml
chestercheng: The second run uses the Python shiboken library path, which may affect `compile_commands.json` and `clang-tidy` results.
This is intentional to align with the original pilot flow.

cpp/binary/pilot/CMakeLists.txt
chestercheng: > I am wondering whether it's necessary to put `test -n $MODMESH_DIFF_BASE` in both here and `pilot_clang_tidy_diff` target in Makefile. I think current setting will make this variable be checked two times.

I kept the cmake side check because `pilot_clang_tidy_diff` can be run directly, not only through the Makefile.

cpp/binary/pilot/CMakeLists.txt
```diff
@@ -123,8 +110,51 @@ endif()
...
+    PILOT_CLANG_TIDY_DIFF_COMMAND
+    "test -n \"$MODMESH_DIFF_BASE\""
+    " && "
+    "git diff --no-color -U0 \"$MODMESH_DIFF_BASE\" HEAD"
+    " -- $<JOIN:${MODMESH_PILOT_FILES}, >"
+    " | "
+    "\"${CLANG_TIDY_DIFF_EXE}\" -p1"
+    " -path \"${CMAKE_BINARY_DIR}\""
+    " -config-file \"${PROJECT_ROOT_DIR}/.clang-tidy\""
+)
```
chestercheng, line 146: Since this command is composed from cmake variables (e.g. `CLANG_TIDY_DIFF_EXE` and `MODMESH_PILOT_FILES`).
I would prefer keeping it in CMake and splitting it with a variable.

yungyuc (approved):
LGTM

PR #872, opened 2026-06-07
Document Python-first testing and Numpy dtype conventions

yungyuc:
Python is part of the user interface.  Testing in Python will exercise the longest code path through the architecture.  If there is not a good reason, Python tests should be preferred over C++.

Some code can only be tested in C++, then we test it in C++ using the google-test framework.

STYLE.md gains a "Testing" section, and CLAUDE.md/AGENTS.md updates "Testing Structure" for the preference of Python tests over C++ tests.

Also document how to create NumPy arrays:
- STYLE.md gains a "NumPy Array Creation" section requiring every array-creating call (np.empty, np.zeros, np.ones, np.full, np.array, np.arange, ...) to pass an explicit dtype spelled as a single-quoted string, e.g. np.empty(10, dtype='float64').
- Update python-style-review skill accordingly.

[skip-ci]

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -349,6 +349,42 @@ Never do implicit relative import:
 import onedim
 ```
 
+## NumPy Array Creation
+
+Always specify the `dtype` when creating a NumPy `ndarray`, and spell the
+dtype as a string. The default floating-point and integer types depend on the
+platform and the NumPy version, so an explicit dtype keeps the array layout
+deterministic and mirrors the fixed-width integer rule below.
+
+```python
+# GOOD: explicit dtype given as a string.
+a = np.empty(10, dtype='float64')
+b = np.zeros((3, 4), dtype='int32')
+c = np.array([1, 2, 3], dtype='uint8')
+
+# BAD: no dtype; the type is implicit and platform-dependent.
+a = np.empty(10)
+b = np.zeros((3, 4))
+
+# BAD: dtype given as a type object instead of a string.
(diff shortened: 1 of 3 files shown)


STYLE.md
```diff
@@ -349,6 +349,42 @@ Never do implicit relative import:
...
+# BAD: dtype given as a type object instead of a string.
+a = np.empty(10, dtype=np.float64)
+```
+
+This applies to every array-creating call (`np.empty`, `np.zeros`, `np.ones`,
+`np.full`, `np.array`, `np.arange`, etc.).
+
+## Testing
+
+Python tests are the default. Write tests in Python (`tests/`, pytest, files
```
yungyuc, line 378: Prefer Python tests.  Python is part of the user interface, and tests should exercise the longest code path.

STYLE.md
```diff
@@ -349,6 +349,42 @@ Never do implicit relative import:
...
+named `test_*.py`) whenever the code can be exercised through the Python
+bindings. Use C++ gtest (`gtests/`, files named `test_nopython_*.cpp`) only
+when the code cannot or should not be tested from Python -- for example,
+internals with no Python binding, or behavior that must be verified at the C++
```
yungyuc, line 382: Sometimes the code can only be tested in C++, then do it in C++ with google-test.

yungyuc:
@tigercosmos @ThreeMonth03 @ExplorerRay @KHLee529 please take a look.

PR #875, opened 2026-06-08
Fix SimpleArray ghost negative indexing

### LINKED ISSUE #856: SimpleArray should handle negative NumPy strides and Python slice semantics consistently

Affected APIs include:

## Tasks

## Related Issues
* issue #580
* issue #603

## How to reproduce the error
```shell
$ python3 - <<'PY'
import numpy as np
import modmesh

expected = ndarr.copy()

next_value = 1000.0
                    next_value += 1

sarr[key] = value
expected[key] = value
PY
```

```shell
$ python3 - <<'PY'
import numpy as np
import modmesh

next_value = 1000.0
                    next_value += 1

sarr[key] = value
expected[key] = value
PY
```

```shell
$ python3 -u - <<'PY'
import numpy as np
import solvcon

print(
    "view strides:",
)

clone = sarr.clone()
print("clone created")

PY
source sample: 2530933.0
clone created
```

ThreeMonth03:
To support negative indexing in [`SimpleArray.__getitem__`](https://github.com/solvcon/modmesh/blob/ee8174b7b559ebef870ab63d0faf91a5c28584a8/cpp/modmesh/buffer/pymod/wrap_SimpleArray.cpp#L191-L194), this pull request normalize index based on the shape size and ghost value.

To be more specific, `SimpleArray.__getitem__` shifts the first index according to `nghost` at first, then check whether all indices after shifted are in valid range(-n , n). If they are in valid range, normalize indices to non-negative indices.

Related to #856.

### DIFF
--- a/cpp/modmesh/buffer/SimpleArray.hpp
+++ b/cpp/modmesh/buffer/SimpleArray.hpp
@@ -1788,6 +1788,9 @@ public:
     value_type const & operator[](size_t it) const noexcept { return data(it); }
     value_type & operator[](size_t it) noexcept { return data(it); }
 
+    // FIXME: This overload treats it as a raw non-negative offset, unlike
+    // at(ssize_t), which applies Python/ghost logical indexing. Remove this
+    // overload in a future cleanup PR after raw-offset users are explicit.
     value_type const & at(size_t it) const
     {
         validate_range(it);
@@ -1801,15 +1804,13 @@ public:
 
     value_type const & at(ssize_t it) const
     {
-        validate_range(it);
-        it += m_nghost;
-        return data(it);
+        shape_type const idx{normalize_index(it)};
+        return data(buffer_offset(m_stride, idx));
     }
     value_type & at(ssize_t it)
     {
-        validate_range(it);
-        it += m_nghost;
-        return data(it);
+        shape_type const idx{normalize_index(it)};
+        return data(buffer_offset(m_stride, idx));
     }
 
     value_type const & at(std::vector<size_t> const & idx) const { return at(shape_type(idx)); }
@@ -1831,19 +1832,15 @@ public:
     value_type const & at(std::vector<ssize_t> const & idx) const { return at(sshape_type(idx)); }
     value_type & at(std::vector<ssize_t> const & idx) { return at(sshape_type(idx)); }
 
-    value_type const & at(sshape_type sidx) const
+    value_type const & at(sshape_type const & sidx) const
     {
-        validate_shape(sidx);
-        sidx[0] += m_nghost;
-        shape_type const idx(sidx.begin(), sidx.end());
+        shape_type const idx = normalize_index(sidx);
         const size_t offset = buffer_offset(m_stride, idx);
         return data(offset);
     }
-    value_type & at(sshape_type sidx)
+    value_type & at(sshape_type const & sidx)
     {
-        validate_shape(sidx);
-        sidx[0] += m_nghost;
-        shape_type const idx(sidx.begin(), sidx.end());
+        shape_type const idx = normalize_index(sidx);
         const size_t offset = buffer_offset(m_stride, idx);
         return data(offset);
     }
@@ -2059,7 +2056,14 @@ private:
         }
     }
 
-    void validate_range(ssize_t it) const
+    /**
+     * Validate a raw, non-negative buffer offset.
+     *
+     * This helper is currently used by unsigned-index paths. Signed logical
+     * indices, including ghost indices, are normalized by normalize_index().
+     * FIXME: Remove this function after raw-offset users are made explicit.
+     */
+    void validate_range(size_t it) const
     {
         if (m_nghost != 0 && ndim() != 1)
         {
@@ -2069,25 +2073,57 @@ private:
                             ndim(),
                             m_nghost));
         }
-        if (it < -static_cast<ssize_t>(m_nghost))
-        {
-            throw std::out_of_range(
-                std::format("SimpleArray: index {} < -nghost: {}",
-                            it,
-                            -static_cast<ssize_t>(m_nghost)));
-        }
-        if (it >= static_cast<ssize_t>((buffer().nbytes() / ITEMSIZE) - m_nghost))
+        size_t const last = (buffer().nbytes() / ITEMSIZE) - m_nghost;
+        if (it >= last)
         {
             throw std::out_of_range(
                 std::format("SimpleArray: index {} >= {} (buffer size: {} - nghost: {})",
                             it,
-                            (buffer().nbytes() / ITEMSIZE) - m_nghost,
-                            (buffer().nbytes() / ITEMSIZE),
+                            last,
+                            buffer().nbytes() / ITEMSIZE,
                             m_nghost));
         }
     }
 
-    void validate_shape(small_vector<ssize_t> const & idx) const
+    size_t normalize_index(ssize_t it) const
+    {
+        if (ndim() != 1)
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray::normalize_index(): "
+                            "cannot use scalar index for {}-dimensional array",
+                            ndim()));
+        }
+        ssize_t const dim_length = static_cast<ssize_t>(shape(0));
+        ssize_t const ghost_offset = static_cast<ssize_t>(m_nghost);
+        ssize_t const shifted_index = it + ghost_offset;
+
+        // This overload handles scalar indexing, not slicing. Python sequence
+        // indexing substitutes "len(s) + i" for a negative i and raises
+        // "IndexError" when i is outside the sequence range. Therefore the
+        // normalized index must satisfy 0 <= len(s) + i < len(s), so the valid
+        // scalar interval is [-len(s), len(s)).
+        // https://docs.python.org/3/library/stdtypes.html#common-sequence-operations
+        if (shifted_index < -dim_length)
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray: index {} < -nghost - shape[0]: {}",
+                            it,
+                            -dim_length - ghost_offset));
+        }
+        if (shifted_index >= dim_length)
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray: index {} >= {} (shape[0]: {} - nghost: {})",
+                            it,
+                            dim_length - ghost_offset,
+                            shape(0),
+                            m_nghost));
+        }
+        return to_nonnegative_index(shifted_index, dim_length);
+    }
+
+    shape_type normalize_index(small_vector<ssize_t> const & idx) const
     {
         auto index2string = [&idx]() -> std::string
         {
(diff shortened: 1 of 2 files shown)


cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1788,6 +1788,9 @@ class SimpleArray
     value_type const & operator[](size_t it) const noexcept { return data(it); }
     value_type & operator[](size_t it) noexcept { return data(it); }
 
+    // FIXME: This overload treats it as a raw non-negative offset, unlike
+    // at(ssize_t), which applies Python/ghost logical indexing. Remove this
+    // overload in a future cleanup PR after raw-offset users are explicit.
     value_type const & at(size_t it) const
     {
         validate_range(it);
```
ThreeMonth03, line 1796: The behavior of `value_type const & at(size_t it) ` is not aligned with `value_type const & at(ssize_t it)`, so it may be removed in another pr.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -1831,19 +1832,15 @@ class SimpleArray
...
-    value_type & at(sshape_type sidx)
+    value_type & at(sshape_type const & sidx)
     {
-        validate_shape(sidx);
-        sidx[0] += m_nghost;
-        shape_type const idx(sidx.begin(), sidx.end());
+        shape_type const idx = normalize_index(sidx);
         const size_t offset = buffer_offset(m_stride, idx);
         return data(offset);
     }
```
ThreeMonth03, line 1846: Normalize index before taking data.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2069,25 +2066,51 @@ class SimpleArray
...
-                std::format("SimpleArray: index {} < -nghost: {}",
+                std::format("SimpleArray: index {} >= {} (buffer size: {} - nghost: {})",
                             it,
-                            -static_cast<ssize_t>(m_nghost)));
+                            last,
+                            buffer().nbytes() / ITEMSIZE,
+                            m_nghost));
         }
-        if (it >= static_cast<ssize_t>((buffer().nbytes() / ITEMSIZE) - m_nghost))
+    }
```
ThreeMonth03, line 2086: This function only used in the function with unsigned parameter now.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2069,25 +2066,51 @@ class SimpleArray
...
+    size_t normalize_index(ssize_t it) const
+    {
+        if (ndim() != 1)
         {
             throw std::out_of_range(
-                std::format("SimpleArray: index {} >= {} (buffer size: {} - nghost: {})",
+                std::format("SimpleArray::normalize_index(): "
+                            "cannot use scalar index for {}-dimensional array",
+                            ndim()));
+        }
```
ThreeMonth03, line 2096: This api was supported both 1D array with ghost and nD array without ghost, but I think it violate the single responsibility principle. 
Thus, I remove the support of nD array, and it only supports for 1D array now.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2069,25 +2066,51 @@ class SimpleArray
...
+        if (shifted_index >= dim_length)
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray: index {} >= {} (shape[0]: {} - nghost: {})",
+                            it,
+                            dim_length - ghost_offset,
+                            shape(0),
                             m_nghost));
         }
```
ThreeMonth03, line 2122: `shifted_index` should be between `-dim_length` and  `dim_length` after considering `nghost`.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2121,43 +2144,59 @@ class SimpleArray
...
+        for (size_t dim = 0; dim < m_shape.size(); ++dim)
         {
-            if (idx[it] < 0)
+            ssize_t const dim_length = static_cast<ssize_t>(m_shape[dim]);
+            ssize_t ghost_offset = 0;
+            if (dim == 0)
+            {
+                ghost_offset = static_cast<ssize_t>(m_nghost);
+            }
+            ssize_t const shifted_index = idx[dim] + ghost_offset;
```
ThreeMonth03, line 2169: When normalizing indices, API only shifts index0 based on nghost.

ThreeMonth03 (commented):
@yungyuc It seems that the archive of pybind11 cannot be downloaded now, so CI fails. You might review this pull request later.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Right.  With ghost index, we may need to avoid unsigned index `size_t`.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Please add this to the function comment block to highlight.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Which API/function you are talking about?  `normalize_index`?  This overload only consumes a scalar and cannot support multi-dimensional arrays.  Please clarify.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: Does Python has the limitation?  If yes, leave a code comment with the URL pointing to the Python document.

yungyuc (changes requested):
Good enhancement.  Points to clarify:
- [x] Clarify what do you mean by the 1D&ND support for `normalize_index()`
- [x] Clarify if Python negative index needs to be in the interval of [-dim_length, dim_length).

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: I don't refer the code block properly. The function `size_t normalize_index(ssize_t it)` is used in `value_type const & at(ssize_t it) const` now. The original design of  `value_type const & at(ssize_t it) const` supports both 1D array with nghost and nD array without nghost, which violate the single responsibility principle.

tigercosmos (commented):
Maybe it's better to separate the ghost related test into different test cases.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: Yes, it is. According to the following 2 rules in [python docs](https://docs.python.org/3/library/stdtypes.html#common-sequence-operations), they infer that a valid index is in the interval of [-dim_length, dim_length).

* If i or j is negative, the index is relative to the end of sequence s: `len(s) + i` or `len(s) + j` is substituted. But note that `-0` is still `0`.
* An [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError) is raised if i is outside the sequence range.

ThreeMonth03:
> Maybe it's better to separate the ghost related test into different test cases.

Sounds great. I will divide them into different testcases.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2059,7 +2056,14 @@ class SimpleArray
...
 
-    void validate_range(ssize_t it) const
+    /**
+     * Validate a raw, non-negative buffer offset.
+     *
+     * This helper is currently used by unsigned-index paths. Signed logical
+     * indices, including ghost indices, are normalized by normalize_index().
+     * FIXME: Remove this function after raw-offset users are made explicit.
+     */
+    void validate_range(size_t it) const
```
ThreeMonth03, line 2066: Add comment block for labeling code as deprecated function.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: `size_t normalize_index(ssize_t it) const` is called here. It replaces the deprecated function `void validate_range(size_t it) const`, which checked whether it is 1D SimpleArray or `nghost` is zero.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2069,25 +2073,57 @@ class SimpleArray
...
         {
             throw std::out_of_range(
-                std::format("SimpleArray: index {} < -nghost: {}",
+                std::format("SimpleArray: index {} >= {} (buffer size: {} - nghost: {})",
                             it,
-                            -static_cast<ssize_t>(m_nghost)));
+                            last,
+                            buffer().nbytes() / ITEMSIZE,
+                            m_nghost));
         }
```
ThreeMonth03, line 2085: `void validate_range(size_t it) const` checked whether it is 1D SimpleArray or `nghost` is zero.

cpp/modmesh/buffer/SimpleArray.hpp
ThreeMonth03: In other words, the api `void validate_range(size_t it) const` was supported for both multi-dimensional SimpleArray without `nghost` and one-dimensional SimpleArray with `nghost`, which violated single responsiblity principle.
Thus, I replace `void validate_range(size_t it) const`  with `size_t normalize_index(ssize_t it) const`, and `size_t normalize_index(ssize_t it) const` only supports for one-dimensional SimpleArray with `nghost` now.

tests/test_buffer.py
ThreeMonth03, line 1: Divide testcases into more functions.

cpp/modmesh/buffer/SimpleArray.hpp
```diff
@@ -2069,25 +2073,57 @@ class SimpleArray
...
+        }
+        if (shifted_index >= dim_length)
+        {
+            throw std::out_of_range(
+                std::format("SimpleArray: index {} >= {} (shape[0]: {} - nghost: {})",
+                            it,
+                            dim_length - ghost_offset,
+                            shape(0),
                             m_nghost));
         }
```
ThreeMonth03, line 2122: Add code block for explaining why a valid index is in the interval of [-dim_length, dim_length).

ThreeMonth03 (commented):
* [x]  Clarify what do you mean by the 1D&ND support for `normalize_index()`
* [x]  Clarify if Python negative index needs to be in the interval of [-dim_length, dim_length).

@yungyuc @tigercosmos Thanks for reviewing this pull request. I modify and clarify functions afterwards. Please take a look.

cpp/modmesh/buffer/SimpleArray.hpp
yungyuc: I see.

But I wouldn't say to support both 1D and ND arrays as violating single-responsibility principle, which was for a class, not a function.  1D and ND arrays are the same thing, not two concepts.  FYI.

yungyuc (approved):
LGTM

PR #876, opened 2026-06-08
Enable ccache for nouse_install

chestercheng:
I noticed that the `nouse_install` workflow builds the same Release buildext
path exercised by `devbuild`, so it can reuse the existing per OS ccache.


yungyuc (approved):
LGTM

Issue #814, opened 2026-05-23
Add a GitHub label to skip CI


tigercosmos:
We shall skip CI for changes of documents or AI prompts.

yungyuc:
Also scan for specific control string in the PR comments for skipping CI.

yungyuc:
@chestercheng I'd like to have your input on this.

yungyuc:
Completed with PR #820.

PR #877, opened 2026-06-09
Ensure skip-ci label workable and do not trigger unexpected CI runs

### LINKED ISSUE #858: Changing label should not trigger CI run
Changing labels triggers Github Actions CI runs, but it should not.  It is especially annoying when adding comment right before merging.

The desired behavior is:
1. When `skip-ci` label exists on the PR, subsequent push does not trigger CI run.
2. When `skip-ci` label does not exist on the PR, subsequent push triggers CI run.

This relates to issue #814.

rockleona:
Simplify profiling job conditions and ensure the skip-ci label is honored for push-triggered updates so CI runs are not noisy or duplicated.

This PR removes push/pull_request triggers from the profiling workflow and updates check_skip_ci to detect and respect skip-ci on push-triggered runs so profiling jobs run only when intended.

Related to #858.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -3,7 +3,7 @@ name: devbuild
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
   schedule:
     - cron: '34 17 * * *'
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
   schedule:
     - cron: '34 17 * * *'
 
--- a/.github/workflows/nouse_install.yml
+++ b/.github/workflows/nouse_install.yml
@@ -3,7 +3,7 @@ name: nouse_install
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
   schedule:
     - cron: '34 17 * * *'
 


.github/workflows/check_skip_ci.yml
```diff
@@ -66,38 +65,76 @@ jobs:
...
+                  per_page: 100,
+                });
+              const hit = comments.find(
+                comment => trusted.includes(comment.author_association) && hasControlLine(comment.body));
+              if (hit) {
+                core.notice(`CI skipped: control string in a comment on PR #${pr.number}.`);
+                return true;
+              }
+
+              return false;
```
rockleona: This helper function will check the content in specific PR, and return boolean of `true` if it needs to skip.

.github/workflows/check_skip_ci.yml
```diff
@@ -66,38 +65,76 @@ jobs:
...
+                const num = p.number;
+                if (await checkPrNumber(num)) {
+                  core.setOutput('skip', 'true');
+                  return;
+                }
+              }
+
+              core.setOutput('skip', 'false');
               return;
             }
```
rockleona: `check_skip_ci` will call by other action (workflow), it will check what triggered the action and check if the push commit is related to any PR, then call the helper function `checkPrNumber`.

.github/workflows/profiling.yml
```diff
@@ -1,8 +1,6 @@
 name: profiling
 
 on:
-  push:
-  pull_request:
```
rockleona, line 5: Removed these 2 triggers

.github/workflows/profiling.yml
```diff
@@ -11,7 +9,7 @@ jobs:
 
     # Run profiling only on schedule or when MMGH_FORCE_PROFILE is set to 'enable'
     # You can refer to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables to set MMGH_FORCE_PROFILE in the fork repository settings for testing purposes.
-    if: ${{ (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') || vars.MMGH_FORCE_PROFILE == 'enable' }}
```
rockleona, line 14: So no need to check `event_name` since the trigger is clean.

yungyuc: If removing `push` and `pull_request` from `on`, `MMGH_FORCE_PROFILE` will not work when `on` is not `schedule`.  We need to allow to use `MMGH_FORCE_PROFILE` to force profiling on `push` and `pull_request`.

.github/workflows/check_skip_ci.yml
```diff
@@ -66,38 +65,76 @@ jobs:
...
               (text || '').split(/\r?\n/)
                 .some(line => control.test(line.trim()));
-            const pr = context.payload.pull_request;
 
-            // 1. The skip-ci label, set by a repository member.
-            if (pr.labels.some(label => label.name === 'skip-ci')) {
-              core.notice('CI skipped: the "skip-ci" label is present.');
-              core.setOutput('skip', 'true');
-              return;
+            async function checkPrNumber(prNumber) {
```
yungyuc: The patch looks overly complex.  I am not sure why a simple conditional check need to use async.  If it is necessary, explain in PR annotation.  If it is not, use minimal diff for the fix.

yungyuc (changes requested):
- [x] Need to allow to use `MMGH_FORCE_PROFILE` to force profiling on `push` and `pull_request`.
- [x] Reduce patch complexity if possible.

yungyuc:
@ExplorerRay @chestercheng please help take a look.

chestercheng:
The requirement in #858 seems to be:
* If the PR has the `skip-ci` label, a later push should not run CI.
* If the PR does not have the `skip-ci` label, a later push should run CI.

So, I wonder if we can simplify this design.

Maybe we can make the `skip-ci` label the single source of truth,
so that `check_skip_ci.yml` only manages the label and all other workflows
only check the label.

Then labels become the only CI gate:
* `skip-ci` controls whether CI should be skipped.
* `profiling` controls whether the profiling workflow should run.
* Removing `skip-ci` allows CI to run again.

This may make the behavior easier to understand and maintain.

yungyuc:
> The requirement in #858 seems to be:
> 
> * If the PR has the `skip-ci` label, a later push should not run CI.
> * If the PR does not have the `skip-ci` label, a later push should run CI.

Issue #858 is just about the `skip-ci` label, but the full picture includes the `[skip-ci]` control line in PR comments: https://github.com/solvcon/modmesh/issues/814#issuecomment-4526848678 .  See also PR #820 .

> Maybe we can make the `skip-ci` label the single source of truth, so that `check_skip_ci.yml` only manages the label and all other workflows only check the label.

It sounds like a good simplification for `check_skip_ci.yml` to be a single source of truth.

.github/workflows/check_skip_ci.yml
```diff
@@ -54,50 +54,56 @@ jobs:
...
+                repo: context.repo.repo,
+                pull_number: prNumber,
+              });
+
+              if (pr.labels && pr.labels.some(label => label.name === 'skip-ci')) {
+                core.notice(`CI skipped: the "skip-ci" label is present on PR #${pr.number}.`);
+                return true;
+              }
+              return false;
             }
```
rockleona: The helper function is simplified, and focus on `skip-ci` label is existed

.github/workflows/profiling.yml
```diff
@@ -1,17 +1,19 @@
 name: profiling
 
 on:
-  push:
-  pull_request:
   schedule:
     - cron: '34 17 * * *'
+  workflow_dispatch:
+  push:
+  pull_request:
```
rockleona: Bring `push` and `pull_request` back and add `workflow_dispatch` in case we have to trigger manually via GitHub Action page without 3 actions above.

rockleona (commented):
Thank you @yungyuc and @chestercheng, I've committed the change from your feedback, here are the changes:

.github/workflows/profiling.yml
yungyuc: Is it a requirement to move them after `schedule`?  If not, keep them the original location to avoid unnecessary diff.

.github/workflows/profiling.yml
```diff
@@ -1,17 +1,19 @@
...
 
 jobs:
   profile:
 
-    # Run profiling only on schedule or when MMGH_FORCE_PROFILE is set to 'enable'
-    # You can refer to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables to set MMGH_FORCE_PROFILE in the fork repository settings for testing purposes.
-    if: ${{ (github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable') || vars.MMGH_FORCE_PROFILE == 'enable' }}
+    # Run profiling on schedule or when MMGH_FORCE_PROFILE is set to 'enable'.
```
yungyuc: The new URL redirects to the original URL.  It is not clear to me why the comments need to be updated.  What is the reason?

If it does not need to, revert the change.

.github/workflows/profiling.yml
```diff
@@ -1,17 +1,19 @@
...
+    # MMGH_FORCE_PROFILE can force profiling for push and pull_request events.
+    # See: https://docs.github.com/en/actions/learn-github-actions/variables#organization-and-repository-variables
+    if: ${{ vars.MMGH_NIGHTLY == 'enable' || vars.MMGH_FORCE_PROFILE == 'enable' }}
```
yungyuc: It is not clear to me why dropping `github.event_name == 'schedule'` works.  Could you please elaborate?

.github/workflows/check_skip_ci.yml
yungyuc: `async` makes it more complex, not simpler.  Do you really need async?

yungyuc (changes requested):
- [x] Avoid unnecessary diff.

Please also explain why do you use async.

.github/workflows/check_skip_ci.yml
```diff
@@ -54,50 +54,56 @@ jobs:
...
 
-            // 2. The control string in the pull request description.
```
yungyuc, line 78: I cannot find the checking for the control string.  Please point out where it is in the new code.

.github/workflows/check_skip_ci.yml
rockleona: Yes, `actions/github-script@v7` use JavaScript as the script language, we have to use `async` and `await` to make sure the value had been fetched.

yungyuc: The JavaScript code before this PR did not use async.  Why do we need it now?

.github/workflows/profiling.yml
rockleona: Fixed just now.

.github/workflows/profiling.yml
rockleona: Take back since it needs to be catch if force running.

Mentioned in [this comment](https://github.com/solvcon/solvcon/pull/877/changes#r3381144872).

.github/workflows/profiling.yml
rockleona: Not really need the update, so I reverted this section.

.github/workflows/check_skip_ci.yml
```diff
@@ -101,4 +151,4 @@ jobs:
               return;
             }
```
rockleona: Control string was back in this commit, I misunderstood the requirement of your request from [the comment](https://github.com/solvcon/solvcon/pull/877#issuecomment-4675961215).

.github/workflows/check_skip_ci.yml
rockleona: I agree it was complex, since I found the usage of `github.rest` in [the document](https://github.com/solvcon/solvcon/pull/877#issuecomment-4675961215), I think it will be fine even if we use without `async`.

rockleona (commented):
I've got control line back in these new commits, so back to the original request, catch **control line** in PR was fine, but `skip-ci` label wasn't really catch on every events.

So I split checking `skip-ci` label into independent step, in case it was not triggered by `pull_request`

.github/workflows/profiling.yml
rockleona: That was my mistake, I put it back now.

.github/workflows/check_skip_ci.yml
```diff
@@ -48,12 +48,69 @@ jobs:
...
 
     outputs:
-      skip: ${{ steps.check.outputs.skip }}
+      skip: ${{ steps.check-label.outputs.skip || steps.check-description.outputs.skip || 'false' }}
 
     steps:
 
-      - name: scan label, description, and comments
-        id: check
+      - name: scan label
```
yungyuc: Label exists on PR.  Please clarify in what use cases the flow should check for label without a PR.

yungyuc (commented):
> I've got control line back in these new commits, so back to the original request, catch **control line** in PR was fine, but `skip-ci` label wasn't really catch on every events.

Label exists only on issues and PRs, and an issue does not trigger CI.  Please clarify how checking for label outside a PR works.

And it is not clear how the current patch address the requirement written in issue #858:

> The desired behavior is:
> 1. When `skip-ci` label exists on the PR, subsequent push does not trigger CI run.
> 2. When `skip-ci` label does not exist on the PR, subsequent push triggers CI run.

Please elaborate.

.github/workflows/devbuild.yml
```diff
@@ -3,7 +3,7 @@ name: devbuild
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
```
rockleona, line 6: There are only three actions will trigger by label / unlabel action on PR.

This is the 1st one.

.github/workflows/lint.yml
```diff
@@ -3,7 +3,7 @@ name: lint
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
```
rockleona, line 6: 2nd one.

.github/workflows/nouse_install.yml
```diff
@@ -3,7 +3,7 @@ name: nouse_install
 on:
   push:
   pull_request:
-    types: [opened, reopened, synchronize, labeled, unlabeled]
+    types: [opened, reopened, synchronize]
```
rockleona, line 6: 3rd one.

rockleona (commented):
> Label exists only on issues and PRs, and an issue does not trigger CI. Please clarify how checking for label outside a PR works.

I was lost on investigating this issue, to trigger `check_skip_ci`, which is to check if PR has `skip-ci` label and control line, means this behavior only need to focus on Pull Request related actions.

> And it is not clear how the current patch address the requirement written in issue https://github.com/solvcon/solvcon/issues/858:
> 
> The desired behavior is:
> 
> When skip-ci label exists on the PR, subsequent push does not trigger CI run.
> When skip-ci label does not exist on the PR, subsequent push triggers CI run.

So from the root cause mentioned in #858, it should be remove label / unlabel action on the PR, this will prevent the unexpected run.

.github/workflows/devbuild.yml
yungyuc: This looks like right, and very clean.

yungyuc (approved):
LGTM.

PR #899, opened 2026-06-15
pilot: Extract world render logic into RWorldRenderer2d (issue #754 PR1)

### LINKED ISSUE #754: Make a truly 2D canvas
The current 2D drawing widget is actually a 3D renderer using Qt3D.  Although all geometry entities (points, lines, polygons, etc.) have 3D for future use, the 3D rendering overheads are not necessary for 2D geometry.

A truly 2D drawing widget (canvas) can simply ignore the third dimension (z) and draw in the 2D space.

Make a Qt widget that can draw `World` in strictly 2D space.  The widget should support basic viewing actions including panning and zooming.  The control should be done in Python.

All geometry manipulation code should be testable without Qt.

Keep each PR to be less than 500 lines of change.  If a PR can be longer than that, break it into stages.

tigercosmos:
Move `R2DWidget::paintWorld` into a new `RWorldRenderer2d` class in `RWorldRenderer2d.{hpp,cpp}`. `R2DWidget::paintEvent` now constructs `RWorldRenderer2d{world, view}` and calls `paint(painter)`, so the on-screen widget and the future offscreen renderer share one code path.

`World::points()` is made `const` to allow the renderer to hold a const world reference. It already returned a const reference to the shared pointer, so this only widens where it can be called and breaks no existing caller.

Pure refactor; rendered pixels unchanged.

Built the pilot, ran the pilot Python tests (`test_pilot.py` and `test_pilot_2d.py` both green), and ran the C++ lint set (`cformat`, `cinclude`, `checkascii`, `checktws`) plus `flake8` — all clean.

Related to #754.


cpp/modmesh/pilot/RPainter2d.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+QColor const GEOMETRY(120, 180, 240);
+
+// Cosmetic (zoom-independent) screen widths for world geometry.
+constexpr double GEOMETRY_LINE_WIDTH_PX = 1.5;
+constexpr int GEOMETRY_POINT_WIDTH_PX = 5;
+
+} // unnamed namespace
+
+void paint_world_2d(QPainter & painter, WorldFp64 const & world, ViewTransform2dFp64 const & view)
```
tigercosmos: Refactor the function to here. `QPainter` can take the argument from `R2DWidget` or future `QImage`

cpp/modmesh/pilot/R2DWidget.cpp
```diff
@@ -259,7 +194,10 @@ void R2DWidget::paintEvent(QPaintEvent * /*event*/)
     }
 
     // World geometry on top of the grid, under the origin marker.
-    paintWorld(painter);
+    if (m_world)
+    {
+        paint_world_2d(painter, *m_world, m_view);
+    }
```
tigercosmos, line 200: `paint_world_2d` is called here.

tigercosmos (commented):

cpp/modmesh/pilot/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ cmake_minimum_required(VERSION 4.0.1)
 set(MODMESH_PILOT_PYMODHEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/R3DWidget.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/R2DWidget.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/RPainter2d.hpp
```
yungyuc: I like the namd `RPainter2d`.  The names of `R3DWidget` and `R2DWidget` should follow.  But "widget" is not informative.  Do not rename hastily.  Let's give it more thoughts.

The only complain I have with "painter" is that it sounds like a pixel facility, like the Windows legacy suggested.  But it's not a big deal.  Just to share an instinctive feeling.

cpp/modmesh/pilot/RPainter2d.cpp
```diff
@@ -0,0 +1,108 @@
...
+namespace
+{
+
+QColor const GEOMETRY(120, 180, 240);
+
+// Cosmetic (zoom-independent) screen widths for world geometry.
+constexpr double GEOMETRY_LINE_WIDTH_PX = 1.5;
+constexpr int GEOMETRY_POINT_WIDTH_PX = 5;
+
+} // unnamed namespace
```
yungyuc: It's called anonymous namespace.  "Unnamed" is unconventional.

cpp/modmesh/pilot/RPainter2d.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+void paint_world_2d(QPainter & painter, WorldFp64 const & world, ViewTransform2dFp64 const & view)
+{
+    // Map math-convention world (x, y) to Qt screen pixels; z is dropped.
+    auto map = [&view](double world_x, double world_y)
```
yungyuc: When a lambda/closure is used but not for a functional purpose, it is a sign that the enclosing code should be made as a class and the lambda should be a member function.

But a function like this size (`paint_world_2d`) is short enough and readable.  It's OK to write code like this.

A more extensible design is to make it a class now.  By holding `WorldFp64 &` as a required member datum, the architecture explains itself.

yungyuc (approved):
LGTM.  If you do not want to enhance the architecture, I can just merge.  Let me know what I should do.

cpp/modmesh/pilot/RPainter2d.cpp
tigercosmos: Sure, let's make it a class directly.

cpp/modmesh/pilot/RPainter2d.cpp
tigercosmos: Got it.

cpp/modmesh/pilot/CMakeLists.txt
tigercosmos: I changed to `RWorldRenderer2d`.

cpp/modmesh/pilot/RWorldRenderer2d.cpp
```diff
@@ -0,0 +1,107 @@
...
+namespace
+{
+
+QColor const GEOMETRY(120, 180, 240);
+
+// Cosmetic (zoom-independent) screen widths for world geometry.
+constexpr double GEOMETRY_LINE_WIDTH_PX = 1.5;
+constexpr int GEOMETRY_POINT_WIDTH_PX = 5;
+
+} // anonymous namespace
```
tigercosmos, line 48: Fixed.

cpp/modmesh/pilot/RWorldRenderer2d.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+QPointF RWorldRenderer2d::map(double world_x, double world_y) const
```
tigercosmos, line 50: Refactored.

cpp/modmesh/pilot/RWorldRenderer2d.hpp
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Renders a world's live points, segments, and curves into a QPainter in
+ * screen space, mapping math-convention world coordinates through a 2D view
+ * transform. The world and view it draws are held as required construction
+ * arguments, so an instance can only exist for a renderable pair. This is the
+ * shared routine used by both the on-screen R2DWidget and the offscreen image
+ * renderer, so neither path can drift from the other. It paints only the
+ * geometry; the caller owns the background, grid, axes, and any origin marker.
+ */
+class RWorldRenderer2d
```
tigercosmos, line 52: Renamed to RWorldRenderer2d

tigercosmos (commented):

cpp/modmesh/pilot/RWorldRenderer2d.hpp
yungyuc: I am not sure `RWorldRenderer2d` is a good name.  It looks too much like `RWorld`, but it contains a `RWorld`.

It is just a concern, not an issue.  I will give it a shot, and see how it should be renamed, or stays unchanged.

yungyuc (approved):
LGTM

PR #930, opened 2026-06-20
Default-load common imports in the Pilot console

tigercosmos:
Install `solvcon`, `sc`, and `pilot` into the builtin namespace in `setup_process` so the Pilot console has them available without the user retyping the imports at every launch. The `pilot` name pulls in Qt/PySide6, so it is installed only when the build has the GUI.

### DIFF
--- a/solvcon/system.py
+++ b/solvcon/system.py
@@ -96,6 +96,10 @@ def setup_process(argv):
     builtins.solvcon = solvcon
     builtins.sc = solvcon
 
+    if solvcon.HAS_PILOT:
+        from . import pilot
+        builtins.pilot = pilot
+
 
 def enter_main(argv):
     args = _parse_command_line(argv)
--- a/tests/test_pilot.py
+++ b/tests/test_pilot.py
@@ -28,6 +28,18 @@ class PilotTC(unittest.TestCase):
         self.assertFalse(pilot.mgr.pycon.python_redirect)
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class SetupProcessTC(unittest.TestCase):
+
+    def test_namespace_includes_pilot(self):
+        import builtins
+        from solvcon import system, pilot
+        system.setup_process([])
+        self.assertIs(builtins.solvcon, solvcon)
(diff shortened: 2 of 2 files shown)


cpp/solvcon/pilot/RPythonConsoleDockWidget.cpp
```diff
@@ -297,6 +302,22 @@ void RPythonConsoleDockWidget::executeCommand()
     m_python_redirect.deactivate();
 }
 
+void RPythonConsoleDockWidget::runStartupImports()
+{
+    // Echoed in the history so the user can see what is already available.
+    // The plain "solvcon" and "sc" names are also installed as builtins by
+    // solvcon.system.setup_process, but importing them here keeps the console
```
yungyuc: Why don't you just add `pilot` in [`solvcon/system.py#L96`](https://github.com/solvcon/solvcon/blob/28176513e7441e0d61ba171f8da0b16d4055f506/solvcon/system.py#L96)?  `solvcon` and `sc` should already be available in the builtin namespace:

```python
def setup_process(argv):
    """Set up the runtime environment for the process."""
    # Install the namespace.
    builtins.solvcon = solvcon
    builtins.sc = solvcon
```

yungyuc (changes requested):
- [ ] Use the existing function `setup_process()` in [`solvcon/system.py#L96`](https://github.com/solvcon/solvcon/blob/28176513e7441e0d61ba171f8da0b16d4055f506/solvcon/system.py#L96) to implement.

solvcon/system.py
```diff
@@ -96,6 +96,10 @@ def setup_process(argv):
     builtins.solvcon = solvcon
     builtins.sc = solvcon
 
+    if solvcon.HAS_PILOT:
+        from . import pilot
```
yungyuc, line 100: This is a little bit dangerous.  If for some reason `HAS_PILOT` is true but `pilot` does not load, this will crash and no one can enter GUI.  A safer implementation is to catch `ImportError` and issue warning on exception.

I am OK with the flaky implementation for now.  Please find an opportunity to fix it in the future.

yungyuc (approved):

PR #937, opened 2026-06-22
Add front/back, pop, and empty helpers to BufferExpander and SimpleCollector

tigercosmos:
Add `empty()` and `pop_size()` to `BufferExpander`, and `empty()`,
`front()`, `back()`, and `pop_back()` to `SimpleCollector`, so it can act
as a LIFO stack of scalars.

Per review:

- Added `front()` as the companion of `back()`.
- `front()`/`back()` are unchecked accessors that mirror `operator[]`,
  following the `std::vector` convention (no `back_at()` bound-checking
  variant). They carry `noexcept` to match the existing `operator[]`.
- `pop_back()` throws `std::out_of_range` on an empty collector via
  `BufferExpander::pop_size`'s buffer-level guard.

Includes gtests.

### DIFF
--- a/cpp/solvcon/buffer/BufferExpander.hpp
+++ b/cpp/solvcon/buffer/BufferExpander.hpp
@@ -97,6 +97,8 @@ public:
         return static_cast<size_type>(this->m_end_cap - this->m_begin);
     }
 
+    bool empty() const noexcept { return size() == 0; }
+
     void reserve(size_type cap);
 
     void expand(size_type length)
@@ -124,6 +126,19 @@ public:
                             capacity()));
         }
         m_end += amount;
+    }
+
+    /**
+     * Pull the size down by amount, keeping the capacity intact.
+     * @param amount Number of bytes to pull down.
+     */
+    void pop_size(size_type amount)
+    {
+        if (amount > size())
+        {
+            throw std::out_of_range(std::format("{}: amount {} > size() {}", name(), amount, size()));
+        }
+        m_end -= amount;
     }
 
     std::shared_ptr<ConcreteBuffer> copy_concrete(size_type cap = 0) const;
@@ -200,4 +215,4 @@ private:
 
 } /* end namespace solvcon */
 
-/* vim: set et ts=4 sw=4: */
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/BufferExpander.hpp
```diff
@@ -126,6 +128,19 @@ class BufferExpander
         m_end += amount;
     }
 
+    /**
+     * Pull the size down by amount, keeping the capacity intact.
+     * @param amount Number of bytes to pull down.
+     */
+    void pop_size(size_type amount)
```
tigercosmos, line 135: Mirror push_size

cpp/solvcon/buffer/BufferExpander.hpp
```diff
@@ -97,6 +97,8 @@ class BufferExpander
         return static_cast<size_type>(this->m_end_cap - this->m_begin);
     }
 
+    bool empty() const noexcept { return size() == 0; }
```
tigercosmos, line 100: New added.

cpp/solvcon/buffer/SimpleCollector.hpp
```diff
@@ -117,6 +118,18 @@ class SimpleCollector
         (*this)[it] = std::move(value);
     }
 
+    value_type const & back() const { return at(size() - 1); }
+    value_type & back() { return at(size() - 1); }
+
+    void pop_back()
```
tigercosmos, line 126: Meat.

tigercosmos (commented):
@yungyuc  Please take a look.

cpp/solvcon/buffer/SimpleCollector.hpp
```diff
@@ -117,6 +118,18 @@ class SimpleCollector
         (*this)[it] = std::move(value);
     }
 
+    value_type const & back() const { return at(size() - 1); }
```
yungyuc: Add also `front()`.

For performance, please assess if we should call it `back_at()`, and leave `front()` and `back()` for non-bound-checking variant.

I am not sure if there is an established convention for the bound-checking and non-bound-checking variants in C++.

yungyuc (changes requested):
- [x] Add also `front()` as a companion of `back()`.
- [x] Discuss: use `back_at()` or not for the bound-checking variant.

cpp/solvcon/buffer/SimpleCollector.hpp
tigercosmos: Added `front`. std::vector only has `front` and `back` and doesn't have bound-checking. It is UB if using the API for an empty vector.

tigercosmos:

cpp/solvcon/buffer/SimpleCollector.hpp
tigercosmos: Decide to not have `*_at()` to follow std::vector.

yungyuc (commented):
LGTM

PR #948, opened 2026-06-26
Prime RHI composition so opening a domain viewer keeps the window

### LINKED ISSUE #934: Reimplement the pilot 3D visualizer using QRhi
The new implementation of the 3D visualizer will be **`RDomainWidget`**, a widget for interactive and programmable analysis of 2D and 3D spatial domains and fields on unstructured meshes, fully controllable from Python.

The implementaion will use **QRhi** in C++ without QML. The existing `R*` 3D classes are a prototype. The reimplementation will be done side by side, and the existing prototype works until it is retired.

This issue follows up issue #928. Also see https://doc.solvcon.net/devplan/rdomainwidget.html for more complete plan.

## Steps

Steps 1 to 8 are implemented in PR #943. Step 9 is in PR #946.

- [x] **1. Widget foundation**

  Stand up `RDomainWidget` as a [`QRhiWidget`](https://doc.qt.io/qt-6/qrhiwidget.html) with the render loop, clear, and depth buffer, drawing one trivial primitive through an `RMaterial` (pipeline plus a GLSL shader pair baked by [`qsb`](https://doc.qt.io/qt-6/qtshadertools-qsb.html)). Bind it so Python can construct it, embed it via PySide6, and call `grabImage()`, the binding over [`QRhiWidget::grabFramebuffer()`](https://doc.qt.io/qt-6/qrhiwidget.html). Settle the build wiring (`Qt6::GuiPrivate`, `#include <rhi/qrhi.h>`, the `qsb` step). This slice establishes both the render foundation and the Python control spine.

- [x] **2. Domain rendering (2D and 3D)**

  Add `RDrawable` and `RMeshFrame`, render the unstructured-mesh domain wireframe from the `SimpleArray` buffers for both 2D and 3D meshes, and bind `updateMesh` / `showMesh`. Validate against the unstructured sample mesh exercised in `solvcon/pilot/_oblique.py`, driven from Python.

- [x] **3. Field coloring and boundary highlight**

  Add `RField` and `RMeshBoundary` with the per-vertex-color `RMaterial` variant, and bind `updateColorField` / `showBoundary`. The field is swappable at runtime from Python.

- [x] **4. Scene and framing**

  Assemble `RDomainScene` with its domain bounding box and per-dimension projection, implement fit-to-scene, and bind `fitCameraToScene`.

- [x] **5. Camera interaction and control**

  Implement `RDomainCameraController` as a single controller with pan/zoom (2D) and first-person (FPS, 3D) modes (no per-mode camera classes), driven by [`QWidget`](https://doc.qt.io/qt-6/qwidget.html) mouse/key events feeding the [`QMatrix4x4`](https://doc.qt.io/qt-6/qmatrix4x4.html) view/projection. Bind the mode switch and programmatic pose get/set so Python navigates the domain as well as the mouse does.

- [x] **6. Orientation-guide overlay**

  Render `RAxisGizmo` as a second pass on its own camera in a corner viewport: a hand-built cone for the arrows and texture billboards for the labels (e.g. [`QPainter`](https://doc.qt.io/qt-6/qpainter.html) into a [`QImage`](https://doc.qt.io/qt-6/qimage.html) uploaded to a [`QRhiTexture`](https://doc.qt.io/qt-6/qrhitexture.html)), as a 2- or 3-axis guide. Bind `showAxis`.

- [x] **7. Consolidate the Python API and tests**

  Finalize and document the `RDomainWidget` pybind11 surface, point the `RManager` widget factory (`add3DWidget` and the MDI hosting) at `RDomainWidget`, update its Python callers (`solvcon/pilot/_mesh.py`, `_canvas_gui.py`) and the GUI tests, and route the screenshot path through `RDomainWidget.grabImage()`. Confirm `make run_pilot_pytest` and `tests/test_pilot_mesh_info.py` pass, exercising the widget through its Python control surface.

- [x] **8. Retire the 3D prototype**

  Delete the Qt 3D prototype classes (`R3DWidget`, `RScene`, `RStaticMesh`, `RColorField`, `RBoundary`, `RLines`, `RVertices`, `RAxisMark`, the `RCameraController` family, and the rest) and the `Qt6::3D*` `find_package` calls from `cpp/binary/pilot/CMakeLists.txt`, drop the build flag if one was used, and remove the Qt 3D modules from the dependency build. Only dead prototype code is removed; `RDomainWidget` has already taken over the 3D view.

- [x] **9. Reclaim class names**

  With the `R3DWidget` prototype retired, its names are free again. Reclaim the cleaner ones: rename `RMeshBoundary` to `RBoundary` and `RDomainScene` to `RScene`.

## Out of scope (later work)

Richer analytical operations (selection, query, measurement, slicing) additional dockable analysis widgets, and dedicated world-geometry primitive classes. The existing 2D viewer (`R2DWidget`, `RWorldRenderer2d`) is unaffected.

yungyuc:
`RDomainWidget` is a `QRhiWidget`, a render-to-texture widget. The first one to appear in the main window makes Qt switch the top-level into texture composition, which requires recreating the native platform window and reparenting every child onto it. On macOS that rebuild blinks the window and drops transient children, so opening the first mesh in a domain sub-window looked like the whole GUI restarted and closed other open widgets.

Park a hidden, zero-size `RDomainWidget` in the MDI area during `setUp`, so the one unavoidable native-window rebuild happens up front while the window is empty and invisible. Every later add3DWidget then reuses the same native window and disturbs nothing.

Guard it with two cases in `RDomainWidgetManagerTC`: one asserts `setUp` leaves the lone hidden primer parented to the MDI area, the other that the main window's winId survives the first `add3DWidget`.

### DIFF
--- a/tests/test_pilot_domain_widget.py
+++ b/tests/test_pilot_domain_widget.py
@@ -21,6 +21,7 @@ import solvcon
 try:
     from solvcon import pilot
     from PySide6.QtGui import QImage
+    from PySide6.QtWidgets import QWidget
 except ImportError:
     pilot = None
 
@@ -503,6 +504,40 @@ class RDomainWidgetManagerTC(unittest.TestCase):
         current = mgr.currentR3DWidget()
         self.assertIsNotNone(current)
         self.assertEqual(current.mesh.ncell, 1)
+
+    def test_setup_primes_rhi_composition(self):
+        """setUp parks a hidden RDomainWidget to fix the GUI-restart bug.
+
+        The first QRhiWidget in the main window makes Qt rebuild the top-level
+        native window so its backing store can flush through QRhi. On macOS
+        that tears down every open sub-window and dock, so opening a mesh looks
+        like the GUI restarts and other viewers vanish. The manager parks a
+        hidden primer viewer in the MDI area at setUp, so the rebuild happens
+        once up front and later viewers reuse the same native window.
+
+        This test checks there is one and only one primer."""
+        mgr = pilot.RManager.instance.setUp()
+        mdi = mgr.mdiArea
+        primers = [
+            w for w in mgr.mainWindow.findChildren(QWidget)
+            if w.metaObject().className().endswith("RDomainWidget")
+            and w.parent() is mdi and not w.isVisible()]
+        self.assertEqual(len(primers), 1)
+
+    def test_open_3d_keeps_native_window(self):
+        """A 3D viewer opened after setUp reuses the primed native window.
+
+        The native handle (winId) changing is the proxy for "the top-level was
+        rebuilt"; with the primer in place it must stay put across add3DWidget.
+        See :meth:`test_setup_primes_rhi_composition` for the mechanism."""
+        mgr = pilot.RManager.instance.setUp()
+        mw = mgr.mainWindow
+        mw.show()
+        before = int(mw.winId())
+        if not before:
+            raise unittest.SkipTest("no native window handle is available")
+        mgr.add3DWidget()
+        self.assertEqual(int(mw.winId()), before)
 
 
 @unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
--- a/cpp/solvcon/pilot/RManager.cpp
+++ b/cpp/solvcon/pilot/RManager.cpp
@@ -51,6 +51,7 @@ RManager & RManager::setUp()
     {
         this->setUpConsole();
         this->setUpCentral();
+        this->primeRhiComposition();
         this->setUpMenu();
 
         m_already_setup = true;
@@ -76,6 +77,7 @@ void RManager::reset()
     m_windowMenu = nullptr;
     m_pycon = nullptr;
     m_mdiArea = nullptr;
+    m_rhi_primer = nullptr;
 }
 
 RManager::~RManager()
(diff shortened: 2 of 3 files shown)


tests/test_pilot_domain_widget.py
```diff
@@ -504,6 +505,40 @@ def test_multiple_3d_widgets_coexist(self):
...
+        """setUp parks a hidden RDomainWidget to fix the GUI-restart bug.
+
+        The first QRhiWidget in the main window makes Qt rebuild the top-level
+        native window so its backing store can flush through QRhi. On macOS
+        that tears down every open sub-window and dock, so opening a mesh looks
+        like the GUI restarts and other viewers vanish. The manager parks a
+        hidden primer viewer in the MDI area at setUp, so the rebuild happens
+        once up front and later viewers reuse the same native window.
+
+        This test checks there is one and only one primer."""
```
yungyuc, line 518: Test for the existence of the unique primer (sub-window / widget).

tests/test_pilot_domain_widget.py
```diff
@@ -504,6 +505,40 @@ def test_multiple_3d_widgets_coexist(self):
...
+        rebuilt"; with the primer in place it must stay put across add3DWidget.
+        See :meth:`test_setup_primes_rhi_composition` for the mechanism."""
+        mgr = pilot.RManager.instance.setUp()
+        mw = mgr.mainWindow
+        mw.show()
+        before = int(mw.winId())
+        if not before:
+            raise unittest.SkipTest("no native window handle is available")
+        mgr.add3DWidget()
+        self.assertEqual(int(mw.winId()), before)
```
yungyuc, line 540: The window remains the same.

PR #959, opened 2026-06-28
Fix the Windows skip on the 3D-keep-native-window test

yungyuc:
The guard on test_open_3d_keeps_native_window (introduced in PR #952) was wrongly written as skipUnless(GITHUB_ACTIONS or non-Windows). Correct it and avoid the error-prone double negation by using skipIf.

### DIFF
--- a/tests/test_pilot_domain_widget.py
+++ b/tests/test_pilot_domain_widget.py
@@ -647,9 +647,9 @@ class RDomainWidgetManagerTC(unittest.TestCase):
     # Showing the main window with the QRhi primer through the MS WARP (Windows
     # Advanced Rasterization Platform) software rasterizer may fault the
     # headless Windows debug runner with an access violation.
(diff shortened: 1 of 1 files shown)


tests/test_pilot_domain_widget.py
```diff
@@ -647,9 +647,9 @@ def test_setup_primes_rhi_composition(self):
     # Showing the main window with the QRhi primer through the MS WARP (Windows
     # Advanced Rasterization Platform) software rasterizer may fault the
     # headless Windows debug runner with an access violation.
-    @unittest.skipUnless(os.getenv('GITHUB_ACTIONS', False) or
-                         platform.system() != "Windows",
-                         "MS WARP may fault headless Windows debug CI run")
+    @unittest.skipIf(os.getenv('GITHUB_ACTIONS', False) and
```
yungyuc, line 650: The mistake was introduced because I forgot to test it on CI nor ask CI to do it for me.  Use more AI.

PR #964, opened 2026-06-28
Shard the SimpleArray pybind11 wrappers by dtype

### LINKED ISSUE #953: Speed up build time by remove overly large translation unit
`wrap_SimpleArray.cpp` takes the 94% of the wall time in parallel building.  It should be sped up.

## Execution plan

Lift `WrapSimpleArray<T>` and `WrapSimpleCollector<T>` into a `wrap_SimpleArray.hpp`, then add per-dtype units (`wrap_SimpleArray_int.cpp`, `_uint.cpp`, `_float.cpp`, `_complex.cpp`, and so on) that each instantiate one slice and register it. A thin `wrap_SimpleArray` entry function calls each slice.

## Baseline measurement

Measured on a development machine: GCC 16, 16 hardware threads, 30 GB RAM, the devenv dependency prefix (Python 3.14.5, pybind11 2.13.6, Qt 6.11.1). Build type is the project default, `Release` with `DEBUG_SYMBOL=ON`, which compiles every unit with `-O3 -g -DNDEBUG -std=c++23`. The build ran with `-j16` (16 cores).

| Phase | Measurement |
| --- | --- |
| CMake configure | 3.9 s |
| Clean parallel build (`make`, `-j16`) | 208.7 s (3 min 29 s) |
| Average CPU use during the build | 585% (about 5.9 of 16 cores) |
| Peak resident memory | 3.9 GB |
| Sum of all per-unit compile times | 1243.6 s |
| Relink the `.so` after one source change | about 7 s |

The per-translation-unit (per-TU) compile times are extraordinarily skewed. The single slowest unit (`wrap_SimpleArray.cpp`) takes longer than the entire rest of the build combined could be compressed to.

| Translation unit | Compile time | Share of floor |
| --- | --- | --- |
| `buffer/pymod/wrap_SimpleArray.cpp` | 196.1 s | 94% |
| `spacetime/pymod/wrap_spacetime.cpp` | 44.1 s | |
| `universe/pymod/wrap_shape2d.cpp` | 42.5 s | |
| `buffer/pymod/wrap_SimpleArrayPlex.cpp` | 39.8 s | |
| `universe/pymod/wrap_shape1d.cpp` | 37.5 s | |
| `universe/pymod/wrap_World.cpp` | 31.5 s | |
| `toggle/pymod/wrap_Toggle.cpp` | 26.3 s | |
| `solvcon_primary_autogen/mocs_compilation.cpp` | 10.9 s | Qt MOC |

## Root cause is the template-heavy TU

`wrap_SimpleArray.cpp` defines a large pybind11 wrapper template, `WrapSimpleArray<T>`, spanning roughly 630 lines with on the order of 80 `.def(...)` calls, each generating its own lambdas and pybind11 dispatch machinery. At the bottom of the file it instantiates that template, plus the companion `WrapSimpleCollector<T>`, for **every supported dtype**:

```
WrapSimpleArray<bool, int8..int64, uint8..uint64, float, double,
                Complex<float>, Complex<double>>   // 13 dtypes
WrapSimpleCollector<... the same 13 dtypes ...>
```

That is **26 full instantiations of heavyweight pybind11 wrapper classes in one translation unit**, compiled serially by a single `cc1plus` process. GCC must instantiate all of it 26 times back to back. This is a textbook template-instantiation monster TU.

## Prototype results

A prototype sharded `wrap_SimpleArray.cpp` into five per-dtype-family units (bool, int, uint, float, complex) behind a thin entry, and the build was re-measured. Wall time follows `max(slowest unit, total work / cores)`, so the gain depends on core count.

### Many cores speed up

**On the 16-core baseline the clean build is 2x faster.** The monster unit is gone, and the new slowest unit (77 s) sits under the work-over-cores bound, so the build is core-bound rather than gated by one file.

| | Before | After |
| --- | --- | --- |
| Clean build (`-j16`) | 209 s | 106 s |
| Slowest single unit | 196 s | 77 s |
| CPU use | 585% | 1239% |
| Peak memory | 3.9 GB | 1.7 GB |

**But the speedup needs spare cores.** Splitting one unit into six re-parses the shared headers and re-emits the dtype-independent scaffolding several times, so total compile work rises 6.6% (1244 s to 1326 s) and the archive 6.8% (721 MB to 770 MB).

### Few cores slow down

On many cores that extra work hides behind parallelism; on few cores it does not. Measured cold on two pinned cores with `-j2`, the CI runner shape:

| | Pre-sharding | Sharded |
| --- | --- | --- |
| Wall (2 cores) | 596 s | 633 s |
| Peak memory | 3.9 GB | 1.7 GB |

So on a 2-core runner sharding costs 6.2% wall time, tracking the extra work almost exactly. The crossover is around six cores (1244 s of work over the 196 s longest unit): below it sharding costs a little, above it sharding wins and the gain grows with core count.

Two things favor sharding regardless of cores:

1. Peak memory halves at every core count, because no single process now instantiates all 26 wrappers at once, which protects the memory-tight runners (the macOS arm64 runner has 7 GB and swaps under an unbounded `make`).
2. `ccache` (already in CI) makes the duplicated work free after the first build, so the cold-build penalty rarely lands in practice.

yungyuc:
`wrap_SimpleArray.cpp` instantiated the `SimpleArray` and `SimpleCollector` wrappers for all 13 dtypes in one translation unit. That unit took 196 s to compile and, at 94% of the parallel-build floor, gated the whole build. Lift the two templates into wrap_SimpleArray.hpp and instantiate them from per-dtype-family units (bool, int, uint, float, complex) behind a thin `wrap_SimpleArray` entry, so the heavy instantiations compile in parallel instead of serially in one process. The contrib/standalone_buffer smoke build lists the new units too.

The split re-parses the shared headers a few times, so total compile work rises slightly, but no single process now instantiates all 26 wrappers at once: the slowest unit and the peak memory both drop, which also protects the memory-tight CI runners. No change to the module layout or any registered type. The buffer tests pass unchanged.

For issue #953.

### DIFF
--- a/cpp/solvcon/buffer/CMakeLists.txt
+++ b/cpp/solvcon/buffer/CMakeLists.txt
@@ -22,14 +22,24 @@ set(SOLVCON_BUFFER_SOURCES
 set(SOLVCON_BUFFER_PYMODHEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/array_common.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/buffer_pymod.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/TypeBroadcast.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/SimpleArrayCaster.hpp
     CACHE FILEPATH "" FORCE)
 
+# The SimpleArray and SimpleCollector wrappers are instantiated for every
+# dtype. Splitting the instantiations into per-dtype-family units lets the
+# heavy pybind11 template work compile in parallel instead of serially in a
+# single translation unit.
 set(SOLVCON_BUFFER_PYMODSOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/buffer_pymod.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_ConcreteBuffer.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray_bool.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray_int.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray_uint.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray_float.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray_complex.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArrayPlex.cpp
     CACHE FILEPATH "" FORCE)
 
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -1,0 +1,665 @@
+#pragma once
+
+/*
+ * Copyright (c) 2022, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
+
+#include <solvcon/buffer/pymod/array_common.hpp>
+
+namespace solvcon
+{
+
+namespace python
+{
+
+template <typename T>
+class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
+    : public WrapBase<WrapSimpleArray<T>, SimpleArray<T>>
+{
+
+    using root_base_type = WrapBase<WrapSimpleArray<T>, SimpleArray<T>>;
+    using wrapped_type = typename root_base_type::wrapped_type;
+    using wrapper_type = typename root_base_type::wrapper_type;
+    using value_type = typename wrapped_type::value_type;
+    using property_helper = ArrayPropertyHelper<T>;
+
+    friend root_base_type;
+
+    WrapSimpleArray(pybind11::module & mod, char const * pyname, char const * pydoc)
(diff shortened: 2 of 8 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.cpp
```diff
@@ -3,678 +3,21 @@
  * BSD 3-Clause License, see COPYING
  */
 
-#include <solvcon/buffer/pymod/buffer_pymod.hpp> // Must be the first include.
-
-#include <solvcon/buffer/pymod/array_common.hpp>
+#include <solvcon/buffer/pymod/wrap_SimpleArray.hpp> // Must be the first include.
```
yungyuc, line 6: Move code to `wrap_SimpleArray.hpp`.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc, line 1: Code moved from Move code to `wrap_SimpleArray.cpp`.

cpp/solvcon/buffer/pymod/wrap_SimpleArray_bool.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace solvcon
+{
+
+namespace python
+{
+
+void wrap_SimpleArray_bool(pybind11::module & mod)
+{
+    WrapSimpleArray<bool>::commit(mod, "SimpleArrayBool", "SimpleArrayBool");
```
yungyuc, line 16: Instantiate wrapper.

cpp/solvcon/buffer/CMakeLists.txt
```diff
@@ -22,14 +22,24 @@ set(SOLVCON_BUFFER_SOURCES
 set(SOLVCON_BUFFER_PYMODHEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/array_common.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/buffer_pymod.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/pymod/wrap_SimpleArray.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/TypeBroadcast.hpp
     ${CMAKE_CURRENT_SOURCE_DIR}/pymod/SimpleArrayCaster.hpp
     CACHE FILEPATH "" FORCE)
 
+# The SimpleArray and SimpleCollector wrappers are instantiated for every
```
yungyuc, line 30: Take a note for why we have multiple `wrap_SimpleArray*.cpp` (for speed).

tigercosmos:
LGTM

PR #967, opened 2026-06-28
Add the headless core of the solvcon.agent package

### LINKED ISSUE #966: Build the solvcon agent for the pilot controlling API
# Build the solvcon agent for the pilot controlling API

Related to issue #896 and discussion #951.

## Summary

This is the tracking issue for the solvcon agent: a natural-language driver for the pilot controlling API, usable headless (no GUI, driving a `World` it owns) and from the GUI. The design has been reworked to put the mechanism first: the core turn primitive is a bounded agentic loop where the model proposes commands, the session executes them, results (especially errors) return as an observation, and the model continues or stops under a step budget. One-shot is just `budget=1`. This plan supersedes `agent_plan_v3.md`; its still-unbuilt pieces (controllers, windowing, toggle gating, chat, docs) are carried as phases here.

The plan reconciles two directives. First, the agent must support both frontier backends (Claude CLI, hosted APIs) and small local models (Qwen, Gemma over an OpenAI-compatible server). Second, the end state adopts the frontier harness techniques (native tool calling, cache-stable composition, a replayable event log, render-back, an eval matrix). The resolution is phased adoption under a dual-support invariant: Phase 1 makes the mechanism work on the simplest protocol every backend can speak, and each later phase layers a frontier technique on top as a capability-probed enhancement. The simple path is never removed.

## Goal

Make the agent a bounded tool-using loop over the controlling API that behaves correctly on both backend classes, then raise the ceiling with frontier techniques that never break the baseline. Concretely: the same prompt ("draw three circles in a row") completes against the Claude CLI and against a local Qwen server with zero backend-specific code, a small model recovers from a malformed or invalid command through the retry loop, and where a frontier backend is present it gets native tool calls, resumed sessions, and rendered images through the same `Turn`, commands, policy, and transcript. Render-back (the model seeing and correcting its own drawing) is the thesis milestone and the PyCon demo path.

## Design decisions of record

1. The bounded agentic loop is the core turn primitive (Phase 1). One-shot is `budget=1`.
2. Dual-support is an invariant, not a phase. The baseline protocol (composed request rendered to flat text, JSON array out, `parse_tool_calls`) is permanent; frontier techniques are declared via capability descriptors and used only when present; the matrix runner (small local vs frontier, baseline vs native) exists from Phase 1 so a regression on either class is visible immediately.
3. Frontier techniques land as committed phases, not maybes. "Don't over-design" changes the order and the fallback rule, not the destination.
4. CLI backends are full agents and must be sandboxed at invocation. Pinning is two-tiered: tier 1 (tool-less) where the CLI supports it (Claude), tier 2 (read-only sandbox, no escalation) where it does not (`codex exec`), with the read-access disclosure recorded honestly.
5. Backend outcomes, parse status, and command validation are three distinct layers. Transport failures abort without retry; parse status distinguishes `commands | empty | prose | malformed`; validation failures are per-command results the model can fix within budget.
6. Step history is typed, protocol-neutral events with correlation ids (`UserGoal`, `AssistantReply`, `ToolResult`, `Observation`), so native backends reconstruct their messages array losslessly and aborted turns never leave a dangling tool call.
7. Artifacts are references, never bytes. Binary or oversized results (for example `render_png`) land in a session artifact store; prompts, transcript, and JSONL see a `{path, mime_type, width, height}` reference plus a one-line summary.

## PR list

Phase 0: harden what is landed

- [x] **PR0: Pin the CLI backend sandbox.** Tier-1 pinning on `ClaudeCliBackend`: tool-disabling and non-interactive permission flags, explicit config/settings isolation so HOME-scoped config cannot re-enable tools, an isolated scratch cwd with a minimal auth-only environment, plus a sentinel test (write isolation) and argv/config assertions (tool disablement). Lands first; it hardens code already running. #1157

Phase 1: the mechanism works on both

- [ ] **PR1: Observation layer, artifact store, compact composition.** Result and scene formatting per the composition rules, the artifact store (session temp dir, safe names, quotas, cleanup) with `render_png`'s base64 as its first customer, `scene_context` rewired, tool dump made compact and deterministic.
- [ ] **PR2a: Typed events, `AgentRequest`, abort semantics.** The event vocabulary with harness-generated ids on the baseline path, the frozen `AgentRequest` with a persistent `goal`, `as_text()` rendering, and the abort semantics (AssistantReply always kept, synthesized not-executed `ToolResult`s, marker `Observation`).
- [ ] **PR2b: Outcomes, parse status, factories, cancel, adaptation.** Backend outcome kinds and the separate parse status, unknown-op checking moved out of parsing into per-command execution, capability descriptors, the registry storing factories so each session owns a private backend instance, `cancel()` wired to both transports, the three landed backends adapted to `send(request)`. Single-step behavior unchanged.
- [ ] **PR3: The bounded turn loop.** `Turn` as a step-wise state machine (`next_request` freezes the concurrency token and scene on the main thread, `feed` does token compare, execution, observation, and layered stop/retry), `run_turn(prompt, budget=8)` rewritten over it, `ScriptedEchoBackend` for tests.
- [ ] **PR4: Console pumps the loop.** The Agent Console drives `Turn` with `next_request`/`feed` on the main thread and `send` in the worker, each step streamed into the transcript, a Stop button that calls `cancel()` mid-step, the widget-lifetime guard made explicit. The world-race debt is closed by the per-step token inside `Turn`. Related PRs: #1052 (console), #1135 (backend off the Qt thread, the non-blocking slice this extends).
- [ ] **PR5: Prove it on both backend classes, with the matrix runner.** The dual-support checkpoint: headless entry (`python -m solvcon.agent "..." --backend NAME --budget 8`), the matrix runner and result format, five scripted smoke fixtures in CI, a documented manual matrix against `claude` and a local OpenAI-compatible server, and the `_INSTRUCTIONS` (#966) reframing informed by those runs.

Phase 2: protocol enhancements (capability-probed, baseline kept)

- [ ] **PR6: Native tool calls and JSON envelope on the HTTP backend.** `OpenAIHttpBackend` builds the `tools` array from the per-op schema and reconstructs the messages array losslessly from typed events, or sends the `{"commands": [...]}` envelope in JSON mode; with descriptors off the baseline text path is byte-identical. Matrix evidence attached. Related PR: #1048 (the HTTP backend this extends).
- [ ] **PR7: Stream Claude Code, CLI resume, graded cancel, Codex CLI.** Single Claude Code intance for a agent session, `ClaudeCliBackend` resume via a token and event cursor in `backend_state` (delta rendering, transparent fallback), graded cancel escalation on both transports, and `CodexCliBackend` over `codex exec` with tier-2 pinning from day one. Matrix evidence attached. Related PR: #1032 (the Claude CLI backend this extends).

Phase 3: policy and the event log

- [ ] **PR8: CommandPolicy.** Replace `allow_destructive` with a per-CRUD-category policy (`allow | ask | deny`); `ask` calls an injected confirmer (headless default deny); the GUI installs a confirm dialog; `feed()` re-checks the concurrency token after any confirmer invocation.
- [ ] **PR9: Transcript JSONL and replay, artifact-safe.** Stable dict form for turns, events, and results, `save_jsonl`/`load_jsonl`, `replay(transcript, world)` with no backend; artifact references serialize as references, never bytes.

Phase 4: controllers and windowing

- [ ] **PR10: Controller protocol and `CanvasDrawController`.** The surface-agnostic core as a refactor: the `Controller` ABC (including `state_token`), `CanvasDrawController` over the Agent Draw public handles, the session unioning surfaces and routing by `owns`, `Turn` aggregating all controllers' tokens.
- [ ] **PR11: `agentwin` command surface (catalog + schema, stable ids).** The windowing catalog (`new_canvas`, `list_windows`, `activate_window`, `close_window`, `set_view`, `reset_view`, `save_image`) with JSON Schema and stable per-window ids. Qt-free, headless-testable. Related PRs: #1139 (Agent Window family: `new_canvas`, `list_windows`, `activate_window`, `close_window`, `save_image`, merged), #1146 (Agent View family: `get_view`, `pan`, `zoom_at`, `set_view`, `reset_view`, and screen/world mapping, open). Much of this item is already delivered by these two families.
- [ ] **PR12: `agentwin` executor and `WindowController`.** `Executor(mgr)` dispatching to `RManager`, the active canvas widget, and the `QMdiArea`, maintaining the id map; `WindowController` with a `state_token` over the id set, active window, and view. Related PR: #1139 (its headless `Executor` binds a duck-typed window manager; this is the GUI-side binding to `RManager` and `QMdiArea`).
- [x] **Expose `R2DWidget.world` to Python.** Landed: the readonly pybind property the canvas controller reads.

Phase 5: conversation and shipping posture

- [ ] **PR13: Multi-turn chat and conversation/backend lifecycle.** Conversation across prompts over the same append-only event machinery, a conversation reset only when the bound world actually changes, `backend_state` namespaced per backend, the console as a chat, and a REPL in `__main__`.
- [ ] **PR14: Feature toggle gating.** `apps.agent.enable`, startup `toggle.load` wiring, console default flipped off because it makes network calls.

Phase 6: the verification thesis

- [ ] **PR15: Offscreen render binding, adapter, and render limits.** The one C++ binding: a widget-independent offscreen entry over `RWorldRenderer2d::paint_canvas` into a QImage at a requested size and view, returned as PNG bytes; schema maxima plus a pre-allocation pixel-product check on `render_png`; the Python adapter and renderer wiring; headless under the Qt offscreen platform.
- [ ] **PR16: Multimodal observation.** Behind `image_transport`: the turn lists step artifacts on the request, `"message_parts"` (HTTP) encodes base64 image parts at send time, `"scratch_copy"` (CLI) copies the artifact into the child scratch cwd under the tested sandbox exception. Text-only models keep the text path untouched. Matrix evidence attached.

Phase 7: measurement and docs

- [ ] **PR17: Eval task expansion.** Grow the smoke set to about a dozen fixtures under `evals/agent/`, document the standing matrix procedure and its two readings (a stronger backend not scoring better means the harness is the ceiling; the small model collapsing only off baseline means an enhancement broke the fallback). CI keeps the scripted subset.
- [ ] **PR18: Documentation.** `doc/source/pilot/agent.md`: the loop, headless vs GUI, the controlling surfaces, backend classes and capability descriptors, the two-tier CLI sandbox pinning, the policy, transcripts and replay, artifacts and quotas, render-back, the toggle, the eval matrix.

Already landed on master (the base this plan builds on):

- [x] Headless core: `AgentSession` with one-shot `run_turn`, transcript, `allow_destructive` gate, the backend ABC, registry, and `EchoBackend` (#967).
- [x] `ClaudeCliBackend`, subprocess based with `parse_tool_calls` and a cancel mechanism (#1032).
- [x] `OpenAIHttpBackend`, stdlib-only with `parse_tool_calls` and a cancel mechanism (#1048).
- [x] The command framework with per-op JSON Schema and CRUD categories, the `World` command vocabulary, the executor, and session wiring, so `solvcon.agent.draw` (including `render_png` with the renderer-injection contract, unbounded and returning inline base64, unreachable today since no renderer is injected) is live (#1074, #1075, #1076, #1077, #1078).
- [x] The agent package imported as a module in its tests (#1106).
- [x] The Agent Console: single turn per prompt, apply plus `requestRepaint()` on the main thread (#1052).
- [x] The backend call moved off the Qt thread with stale-command race handling (#1135).
- [x] The `R2DWidget.world` binding (listed under Phase 4 above).

### Where master is today

Known debts in the landed code that this plan closes: the CLI backend runs unsandboxed from the repository cwd (PR0), `render_png` has no dimension limits (PR15), a world-revision race between submit and apply (PR3/PR4), a widget-lifetime `RuntimeError` swallow (PR4), no cancel in the GUI (PR4/PR7), no toggle gate (PR14), and the `_INSTRUCTIONS` reframing TODO from #966 (PR5). The 2D paint path (`RWorldRenderer2d`) is not bound to Python; the only bound offscreen render needs a live widget at its own size, which is why PR15 carries a dedicated C++ line item.

### The loop, the load-bearing design

```python
turn = session.start_turn(prompt, budget=8)
while not turn.done:
    request = turn.next_request()    # MAIN thread: freezes controller tokens,
                                     # returns an immutable AgentRequest
    response = backend.send(request) # SLOW: off the Qt thread in the GUI
    turn.feed(response)              # MAIN thread: token check,
                                     # validate + execute + observe
```

`Turn` is a step-wise state machine, not a blocking function, because the pilot needs the backend call off the Qt thread and world mutation on it. `next_request()` runs on the main thread (it reads live state to freeze the concurrency tokens and scene snapshot); the returned `AgentRequest` is immutable and safe to hand to a worker. `feed()` first compares the current aggregate token against the frozen one: if the state changed under the turn (user edit or window change between steps) the step's commands are dropped and the turn ends, while the agent's own mutations advance the expected tokens. Headless `run_turn(prompt, budget=8)` drives the same object synchronously.

The concurrency token is per-controller (`Controller.state_token()`). In Phase 1, before controllers exist, the single token is the canvas scene digest (a hash of `describe_state`) through the same seam. Two stated limits: `describe_state` serializes at 6 decimals so edits below 1e-6 escape the digest, and the digest is O(world) on the main thread twice per step. Both are acceptable for v1 canvas work; a cheap C++ `World` revision counter is the named fast-follow that removes both.

### The request contract: `AgentRequest` and typed events

One immutable request object and one small event vocabulary, introduced in PR2a, final from the start so they change once:

```python
ToolCall   = (id, name, arguments)       # provider id preserved when native
UserGoal   = (text,)                     # first event of every turn
AssistantReply = (text, tool_calls)      # what the model said/proposed
ToolResult = (call_id, result_ref)       # per-command result, id-correlated
Observation = (status_line,)             # scene/status tail for the step

@dataclasses.dataclass(frozen=True)
class AgentRequest:
    goal: str             # the user's request, present on EVERY step
    events: tuple         # append-only, starts with UserGoal
    tool_surface: tuple   # compact, deterministic tool definitions
    scene: str            # bounded scene snapshot, frozen at compose time
    artifacts: tuple      # artifact references visible this step (paths)
    backend_state: object # opaque per-conversation token (resume id, cursor)

    def as_text(self):    # the baseline rendering every backend can send
        ...
```

The baseline path (every backend, forever): `as_text()` renders the instruction block, tool definitions, the event sequence, scene, and goal into one flat text; the backend replies with text; the parse layer extracts the JSON array; schema validation rejects unknown ops and bad arguments per command at execution. Enhanced backends read the structured events instead: native mode renders each `ToolResult` as a tool message carrying its `tool_call_id` and the `Observation` status line as a user message, reconstructing the messages array losslessly with no double-sent content (PR6). Baseline batches get harness-generated `ToolCall` ids so the event stream is uniform in both modes.

Abort semantics on every abort path (token conflict, cancellation, policy deny, budget end): the `AssistantReply` event is always appended, every `ToolCall` without an executed result is closed by a synthesized `ToolResult(call_id, "not executed: <reason>")`, and the turn-end marker is an `Observation` event as well as a transcript entry. The event stream therefore never contains a dangling tool call, a native reconstruction is always a valid messages array, and a later turn's model sees why a proposal was dropped.

### Capability descriptors

Backends declare what they support; the harness probes, never assumes, and descriptors come from configuration or explicit detection, never from guessing a model name:

```python
class AgentBackend(abc.ABC):
    supports_tools = False      # native tool/function calling (PR6)
    supports_json_mode = False  # constrained JSON output (PR6)
    supports_resume = False     # backend-side session continuation (PR7)
    image_transport = None      # None | "message_parts" | "scratch_copy" (PR16)

    def cancel(self):           # uniform cancellation contract (PR2b)
        pass
```

The registry stores backend factories; each session owns a private backend instance with at most one in-flight call, so one session's `cancel()` can never reach another session's child process or connection. A small model behind Ollama with defaults gets the baseline and works; a frontier backend with descriptors set gets typed tool calls, resumed sessions, and rendered images. Same `Turn`, same commands, same policy, same transcript.

### Outcomes, parse status, and the retry policy

Three layers, kept distinct. Backend outcome (transport-level): `ok | transport | auth | timeout | cancelled`; anything but `ok` aborts the turn, records the error, and never retries. Parse status (protocol-level, on an `ok` response): `commands | empty | prose | malformed`; unknown-op checking is not part of parsing, so an unknown op parses fine and fails per command at execution. Command results (execution-level): unknown ops, schema failures, and runtime errors surface per command as observations, not outcomes.

| Layer result | Loop action |
| --- | --- |
| outcome not `ok` | abort turn, record, no retry |
| `commands` | execute, observe (including per-command failures), continue |
| `empty` | end turn (explicit completion) |
| `prose` | record the text as the reply, end turn |
| `malformed` | error observation, costs one step, retry |
| per-command failure | error line in the observation, costs the step it happened in, model retries within budget |

### CLI backends are agents; pin them at invocation

A CLI backend is a full agent harness with its own tools (file edit, shell, web). Left unpinned it bypasses the closed command surface, which is this design's strongest safety property. Tier 1 (tool-less), required where the CLI supports it (Claude CLI): the CLI's own tool use is disabled, permission escalation is denied non-interactively, the child runs in an isolated scratch working directory, and neither project-level nor user-level configuration can re-enable tools; only the minimal environment the CLI needs to authenticate is passed. Tier 2 (contained), where tool-less is impossible (`codex exec` always executes commands): read-only OS sandbox, no approval escalation, isolated scratch cwd and config, with the read-access disclosure recorded rather than pretended away. Verification is split honestly: the sentinel test proves write isolation only; argv and config inspection prove tool disablement.

### Artifacts: references, never bytes; bounded resources

Some command results are binary or oversized (`render_png` returns inline base64 today). The observation layer (PR1) routes any such value through the session artifact store: bytes land in a file under a session temp/cache directory (never the Git worktree), under a harness-generated safe name, subject to per-session quotas, cleaned up on session close. The value seen by prompts, transcript, and JSONL becomes a reference plus a one-line summary. Rendering is bounded before allocation: schema maxima on width and height plus a code-level pixel-product check (JSON Schema cannot express `width*height <= P`), enforced again in the renderer (PR15). A vision backend receives the artifact per its `image_transport`.

### Composition rules

Compact serialization and bounded observations serve small models; append-only history and a byte-stable prefix serve frontier prompt caches and local prefix caches equally. Rules: never rewrite earlier events, no varying tokens in the instruction block, tool surface fixed for the life of a session (ops gated by rejecting at execution, not by editing the surface), `sort_keys=True` and no indent on the tool dump. Observations per SWE-agent: one line per command, explicit wording for empty results, consecutive identical errors deduplicated, scene capped with an explicit "N more shapes, use read commands" tail.

### Sequencing

PR0 lands first (it hardens code already running). Phase 1 is the critical path across six PRs (PR1, PR2a, PR2b, PR3, PR4, PR5); Phases 2, 3, and 4 are mutually parallel after it; PR11 can start any time. Every PR stays under 500 net LOC (PR1 and PR3 carry a split rule if they overrun). The first demo checkpoint is the end of Phase 1: the same prompt completing against Claude CLI and a local Qwen, with a recorded mid-turn error recovery.

### Out of scope

Layered instruction files, an MCP front-end over the controller surface, context compaction (the append-only event history is its prerequisite and is in scope), subagents, conversation forking (reset is in PR13, fork is a later addition), and the spatial-visualization controller (part 3 of the controlling API). Each is one bounded addition later; nothing in this plan blocks them.

tigercosmos:
Introduce `solvcon.agent`, the GUI-free foundation an AI backend uses to drive the 2D `World` from natural language. This core lands and is tested on its own, ahead of the user-facing GUI pieces.

The package sits outside `solvcon.pilot` so the agent can also drive pure computation that needs no graphics. It imports no Qt and reaches the Agent Draw executor through a lazy import with a fallback, so it never requires the pilot GUI by default.

What it adds:
- `AgentSession`: binds a `World`, an optional backend, and a command runner, then records every applied command into a transcript.
- `AgentBackend`: the pluggable backend interface, backed by a process-wide registry and an offline `EchoBackend` default.

Later PRs build on this core without changing it: the Claude and Codex backends, the background worker, and the dock panel.

Related to issue #966 and discussion #951.

### DIFF
--- a/solvcon/agent/__init__.py
+++ b/solvcon/agent/__init__.py
@@ -1,0 +1,53 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Agent: drive the 2D ``World`` with an AI backend, with or without a GUI.
+
+This package lives outside :mod:`solvcon.pilot` so it can also drive pure
+computation that needs no graphics.  The headless core (:mod:`_core`) and the
+backend abstraction (:mod:`_backend`) load without Qt, so they run in CI and a
+headless build.
+"""
+
+from . import _core  # noqa: F401
+from . import _backend  # noqa: F401
+
+# _core.py
+list_of_core = [
+    'AgentSession',
+    'TranscriptTurn',
+]
+
+# _backend.py
+list_of_backend = [
+    'AgentBackend',
+    'BackendResponse',
+    'EchoBackend',
+    'register',
+    'all_backends',
+    'available_backends',
+    'get_backend',
+]
+
+# TODO: when the Qt dock module exists in solvcon.pilot, point this at its
+# Agent class, guarded by _pilot_core.enable like the airfoil sub-package.
+Agent = None
+
+__all__ = list_of_core + list_of_backend + [  # noqa: F822
+    'Agent',
+]
+
+
+def _load(module, symbol_list):
+    for name in symbol_list:
+        globals()[name] = getattr(module, name)
+
+
+_load(_core, list_of_core)
+_load(_backend, list_of_backend)
+
+del _load
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -1,0 +1,107 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Pluggable AI backend abstraction for the Agent.
+
+A backend turns a prompt (plus scene context and the Agent Draw tool surface)
+into a :class:`BackendResponse`: prose and a list of Agent Draw command dicts.
+Backends register in a process-wide registry so a caller can list the usable
+ones and let the user pick.  The module imports no Qt.  The offline
+:class:`EchoBackend` keeps the registry non-empty so there is always a working
+default.
+"""
+
+import abc
+import dataclasses
+
+
+@dataclasses.dataclass
+class BackendResponse:
+    """One backend reply: ``text`` prose, the proposed ``commands`` (Agent
+    Draw dicts the session applies; empty means no drawing), and an ``error``
+    reason or ``None``."""
+
+    text: str = ""
+    commands: list = dataclasses.field(default_factory=list)
+    error: str = None
+
+
+class AgentBackend(abc.ABC):
+    """Interface every AI backend implements: a stable :attr:`name`, an
+    :meth:`available` check, and :meth:`send`.  The tiny surface lets a caller
+    drive any backend from a background thread."""
+
+    @property
+    @abc.abstractmethod
+    def name(self):
+        """Short, stable identifier shown in the backend selector."""
+
+    @abc.abstractmethod
+    def available(self):
+        """Whether this backend can run now (CLI on PATH, key set, ...)."""
+
+    @abc.abstractmethod
+    def send(self, prompt, scene_context, tool_surface):
+        """Run the backend and return a :class:`BackendResponse`.
+
+        :param prompt: the user's natural-language request.
+        :param scene_context: a short text summary of the current world.
+        :param tool_surface: the Agent Draw tool definitions the model may
+            call.
+        """
+
+
+_REGISTRY = []
+
+
+def register(backend):
+    """Add a backend, replacing any with the same name (so a re-import does
+    not duplicate the built-in entries)."""
+    for index, existing in enumerate(_REGISTRY):
+        if existing.name == backend.name:
+            _REGISTRY[index] = backend
+            return backend
+    _REGISTRY.append(backend)
+    return backend
+
+
+def all_backends():
+    """Every registered backend, in registration order (a copy)."""
+    return list(_REGISTRY)
+
+
+def available_backends():
+    """Registered backends whose ``available()`` returns True."""
+    return [b for b in _REGISTRY if b.available()]
+
+
+def get_backend(name):
+    """The registered backend with ``name``, or ``None`` if absent."""
+    for backend in _REGISTRY:
+        if backend.name == name:
+            return backend
+    return None
+
+
+class EchoBackend(AgentBackend):
+    """Offline backend that proposes no drawing and echoes the prompt.
+
+    It is always :meth:`available` and fully deterministic, so a caller, the
+    tests, and a no-key demo always have a backend that exercises the whole
+    pipeline without any external process.
+    """
+
+    name = "echo (offline)"
+
+    def available(self):
+        return True
+
+    def send(self, prompt, scene_context, tool_surface):
+        return BackendResponse(text="echo: %s" % prompt, commands=[])
+
+
+register(EchoBackend())
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/solvcon/agent/_core.py
+++ b/solvcon/agent/_core.py
@@ -1,0 +1,126 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Headless core of the Agent.
+
+:class:`AgentSession` binds the active ``World``, an optional backend, and a
+command *runner* (the Agent Draw ``Executor`` by default), and records every
+applied command into a transcript a caller can render.  No Qt is imported.
+"""
+
+import json
+import dataclasses
+
+
+@dataclasses.dataclass
+class TranscriptTurn:
+    """One transcript entry: a ``role`` with its text, commands, and results.
+
+    ``results`` holds Agent Draw ``CommandResult`` objects (or any object with
+    an ``ok`` attribute).
+    """
+
+    role: str
+    text: str = ""
+    commands: list = dataclasses.field(default_factory=list)
+    results: list = dataclasses.field(default_factory=list)
+
+
+@dataclasses.dataclass
+class _OutcomeStub:
+    """Failed-result stand-in for when a runner raises instead of returning."""
+
+    op: str
+    ok: bool = False
+    error: str = None
+
+
+def _make_executor(world, renderer=None):
+    """Build an Agent Draw ``Executor`` for ``world``."""
+    # TODO: when the agentdraw package ships in-tree, import it at module
+    # level (here and in tool_surface) and drop the empty-list fallback.
+    from solvcon.pilot import agentdraw
+    return agentdraw.Executor(world, renderer)
+
+
+def tool_surface():
+    """The Agent Draw tool definitions for a backend, or ``[]`` if the
+    ``agentdraw`` package is unavailable."""
+    try:
+        from solvcon.pilot import agentdraw
+    except ImportError:
+        return []
+    return agentdraw.tool_definitions()
+
+
+class AgentSession:
+    """Bind a ``World``, a backend, and a runner; record a transcript.
+
+    ``runner`` is any object exposing ``run(command) -> result``; it defaults
+    to a lazily built Agent Draw ``Executor(world, renderer)``.  ``backend`` is
+    an :class:`~solvcon.agent.AgentBackend` or ``None``.
+    """
+
+    def __init__(self, world=None, backend=None, runner=None, renderer=None):
+        self.world = world
+        self.backend = backend
+        self._renderer = renderer
+        self._runner = runner
+        self._transcript = []
+
+    @property
+    def transcript(self):
+        """The recorded turns, oldest first (a copy)."""
+        return list(self._transcript)
+
+    @property
+    def runner(self):
+        """The command runner, built from ``agentdraw`` on first use."""
+        if self._runner is None:
+            self._runner = _make_executor(self.world, self._renderer)
+        return self._runner
+
+    def tool_surface(self):
+        """The Agent Draw tool definitions to hand the backend."""
+        return tool_surface()
+
+    def scene_context(self, level="basic"):
+        """A short text summary of the world for the model: the shape count
+        and distinct types from ``world.describe_state(...)`` (JSON), or a
+        plain count when it cannot be described."""
+        world = self.world
+        if world is None:
+            return "no active world"
+        try:
+            state = json.loads(world.describe_state(level=level))
+        except Exception:
+            return "world with %s shapes" % getattr(world, "nshape", "?")
+        shapes = state.get("shapes", [])
+        types = sorted({s["type"] for s in shapes if "type" in s})
+        kinds = ", ".join(types) if types else "none"
+        return "world with %d shapes (types: %s)" % (len(shapes), kinds)
+
+    def apply_commands(self, commands):
+        """Run each command, recording one transcript turn.  An empty batch is
+        a no-op that builds no runner; a runner that raises is captured into a
+        failed :class:`_OutcomeStub`, so one bad command never aborts a batch.
+        """
+        if not commands:
+            return []
+        runner = self.runner
+        results = []
+        for command in commands:
+            try:
+                results.append(runner.run(command))
+            except Exception as exc:
+                op = command.get("op", "?") if isinstance(command, dict) \
+                    else "?"
+                results.append(_OutcomeStub(
+                    op, error="%s: %s" % (type(exc).__name__, exc)))
+        self._transcript.append(TranscriptTurn(
+            role="agent", commands=list(commands), results=results))
+        return results
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/tests/test_agent_backend.py
+++ b/tests/test_agent_backend.py
@@ -1,0 +1,85 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""Tests for the Agent backend abstraction and registry.
+
+GUI-free: only the pure-Python backend module is imported, never an
+``RManager`` or a Qt widget, so these run in CI without a built GUI.
+"""
+
+import unittest
+
+from solvcon.agent import (
+    AgentBackend,
+    BackendResponse,
+    EchoBackend,
+    all_backends,
+    available_backends,
+    get_backend,
+    register,
+)
+
+
+class AgentBackendABCTC(unittest.TestCase):
+    def test_abstract_cannot_instantiate(self):
+        with self.assertRaises(TypeError):
+            AgentBackend()
+
+    def test_partial_subclass_cannot_instantiate(self):
+        # Missing send() leaves an abstract method, guarding the contract that
+        # every concrete backend (Claude, Codex, ...) fills all three.
+        class Partial(AgentBackend):
+            name = "partial"
+
+            def available(self):
+                return True
+
+        with self.assertRaises(TypeError):
+            Partial()
+
+
+class EchoBackendTC(unittest.TestCase):
+    def test_available_true_without_config(self):
+        # Needs no key or process, so it is the guaranteed default backend.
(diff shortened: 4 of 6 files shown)


solvcon/agent/_core.py
```diff
@@ -0,0 +1,126 @@
...
+    """The Agent Draw tool definitions for a backend, or ``[]`` if the
+    ``agentdraw`` package is unavailable."""
+    try:
+        from solvcon.pilot import agentdraw
+    except ImportError:
+        return []
+    return agentdraw.tool_definitions()
+
+
+class AgentSession:
```
tigercosmos, line 58: Interface of the AI session. We will bind a real AI backend later, e.g. Claude.

tests/test_agent_backend.py
```diff
@@ -0,0 +1,85 @@
...
+
+class EchoBackendTC(unittest.TestCase):
+    def test_available_true_without_config(self):
+        # Needs no key or process, so it is the panel's guaranteed default.
+        self.assertTrue(EchoBackend().available())
+
+    def test_send_is_deterministic_and_safe(self):
+        backend = EchoBackend()
+        first = backend.send("hello", "scene", [])
+        second = backend.send("hello", "scene", [])
```
tigercosmos, line 50: Usage example of a backend.

tests/test_agent_core.py
```diff
@@ -0,0 +1,132 @@
...
+@unittest.skipUnless(HAS_AGENTDRAW, "requires the agentdraw package")
+class AgentDrawIntegrationTC(unittest.TestCase):
+    def test_default_runner_mutates_world(self):
+        world = solvcon.WorldFp64()
+        session = AgentSession(world=world)
+        results = session.apply_commands([
+            {"op": "add_circle", "cx": 0.0, "cy": 0.0, "r": 1.0},
+            {"op": "add_circle", "cx": 3.0, "cy": 0.0, "r": 1.0}])
+        self.assertTrue(all(r.ok for r in results))
+        self.assertEqual(world.nshape, 2)
```
tigercosmos, line 125: A usage example of the agent session.

solvcon/agent/_backend.py
```diff
@@ -0,0 +1,107 @@
...
+    """One backend reply: ``text`` prose, the proposed ``commands`` (Agent
+    Draw dicts the session applies; empty means no drawing), and an ``error``
+    reason or ``None``."""
+
+    text: str = ""
+    commands: list = dataclasses.field(default_factory=list)
+    error: str = None
+
+
+class AgentBackend(abc.ABC):
```
tigercosmos, line 31: Interface of all AI agent backend.

tigercosmos (commented):

yungyuc:
> Related to #966 and #951

When referencing github items, prefix type, like: issue #966 and discussion #951, so that readers do not need to click to check.

solvcon/pilot/agent/__init__.py
yungyuc, line 1: I like this name.

But I don't think this belongs to `pilot`.  The agent should also drive for pure calculation that does not use graphics or visualization.

solvcon/agent/_backend.py
yungyuc, line 1: You are in `agent` module, why prefixing `_agent` in sub module names?

yungyuc (changes requested):
- [x] The `agent` package should be outside `pilot` so that it can control things with and without GUI.

When taking the agent outside the GUI, be careful to avoid requiring the GUI (pilot) by default.  The architecture will need some design and experimenting.

tigercosmos (commented):
@yungyuc  The file name issue and the module path issue is fixed. Please take a look, thanks!

solvcon/agent/__init__.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Agent: drive the 2D drawing world with an AI backend, with or without GUI.
```
yungyuc: Are you sure the docstring is accurate?

yungyuc (changes requested):
- [x] Review the docstrings.
- [x] `_load()` pattern.

I feel excited witnessing the inception of an agent in solvcon.

tigercosmos:
@yungyuc  Fixed the comments and used `_load` pattern for pythong packages.

yungyuc (approved):
LGTM.

Review the code added with issue #989 before adding the GUI widget.

Issue #526, opened 2025-04-26
Improve `.clang-format` to avoid long function declaration in macro


KHLee529:
When implementing SIMD library in #489, I found that current clang-format formats the function declaration in macro into a extreme long line as following. The `.clang-format` setup should be refined.

> before clang-format

```cpp
#define DECL_MM_IMPL_VGETQ(N)                                                           \
    template <typename base_type, size_t n,                                             \
        typename std::enable_if_t<                                                      \
            n < type::vector_lane<utype_t(N)> && std::is_same_v<utype_t(N), base_type>  \
        > * = nullptr>                                                                  \
    base_type vgetq(type::vector_t<base_type> vec)                                      \
    {                                                                                   \
        return vgetq_lane_u##N(vec, n);                                                 \
    }                                                                                   \
    template <typename base_type, size_t n,                                             \
        typename std::enable_if_t<                                                      \
            n < type::vector_lane<stype_t(N)> && std::is_same_v<stype_t(N), base_type>  \
        > * = nullptr>                                                                  \
    base_type vgetq(type::vector_t<base_type> vec)                                      \
    {                                                                                   \
        return vgetq_lane_s##N(vec, n);                                                 \
    }
```

> after clang-format

```cpp
#define DECL_MM_IMPL_VGETQ(N)                                                                                                                                                                               \
    template <typename base_type, size_t n, typename std::enable_if_t<n<type::vector_lane<utype_t(N)> && std::is_same_v<utype_t(N), base_type>> * = nullptr> base_type vgetq(type::vector_t<base_type> vec) \
    {                                                                                                                                                                                                       \
        return vgetq_lane_u##N(vec, n);                                                                                                                                                                     \
    }                                                                                                                                                                                                       \
    template <typename base_type, size_t n, typename std::enable_if_t<n<type::vector_lane<stype_t(N)> && std::is_same_v<stype_t(N), base_type>> * = nullptr> base_type vgetq(type::vector_t<base_type> vec) \
    {                                                                                                                                                                                                       \
        return vgetq_lane_s##N(vec, n);                                                                                                                                                                     \
    }
```

rockleona:
Hi @yungyuc, I've created a PR on #970, please take a look, thank you.

PR #970, opened 2026-06-29
Refine .clang-format to wrap long macro declarations

rockleona:
Refine .clang-format to avoid extremely long single-line function declarations produced inside macros by clang-format. Set ColumnLimit from 0 to 79 so long template/macro declarations wrap sensibly and preserve readability.

Related to #526.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -59,7 +59,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeComma
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     0
+ColumnLimit:     79
 CommentPragmas:  '^ IWYU pragma:'
 CompactNamespaces: false
 ConstructorInitializerAllOnOneLineOrOnePerLine: false


.clang-format
```diff
@@ -59,7 +59,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeComma
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     0
+ColumnLimit:     79
```
rockleona, line 62: From no limitation to 79 characters per line, I tried it on Visual Studio Code format feature, the result will be like this:

```diff
diff --git a/gtests/test_nopython_callprofiler.cpp b/gtests/test_nopython_callprofiler.cpp
index 4a8068b..3ccdb87 100644
--- a/gtests/test_nopython_callprofiler.cpp
+++ b/gtests/test_nopython_callprofiler.cpp
@@ -21,10 +21,7 @@ protected:
         pProfiler = &profiler;
     }

-    RadixTree<CallerProfile> & radix_tree()
-    {
-        return pProfiler->m_radix_tree;
-    }
+    RadixTree<CallerProfile> & radix_tree() { return pProfiler->m_radix_tree; }

     CallProfiler * pProfiler;
 };
@@ -37,7 +34,9 @@ void foo3()
 {
     SOLVCON_PROFILE_FUNCTION();
     auto start_time = std::chrono::high_resolution_clock::now();
-    while (std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time).count() < uniqueTime1)
+    while (std::chrono::duration_cast<std::chrono::milliseconds>(
+               std::chrono::high_resolution_clock::now() - start_time)
+               .count() < uniqueTime1)
     {
         // use busy loop to get a precise duration
     }
```

rockleona:
So I change this PR into draft now, since I forgot this change will cause lint issue during CI, for now I have no clue to make this change.

Or should I format all C++ code files, but it will make this PR into XL size I think.

.clang-format
tigercosmos: @rockleona I think we currently use about 120 for C++ code. I guess we only check by eye.

yungyuc: [`STYLE.md:56-57`](https://github.com/solvcon/solvcon/blob/e2c2ee00fc9ef8afad5f03eceeba0787597867d0/STYLE.md?plain=1#L56-L57):

> C++ files do not have a text width limit, but it is good for a line to be less than 120 characters. Python files should use a text width of 79 characters.

We may not impose the hard constraint on line width in C++ source code.  `ColumnLimit: 0` is a correct setting that must not change.

yungyuc (changes requested):
The issue #526 is strictly about the automatic format for CPP (C Pre-Processor) macros in clang-format.  Keep the scope without creeping.

.clang-format
tigercosmos: @yungyuc 
> We may not impose the hard constraint on line width in C++ source code.

May I ask the reason?

yungyuc: It is occasionally good to have long C++ lines, even though they need soft wrap.

PR #973, opened 2026-06-30
Balance line economy with breathing-room guidance

tigercosmos:
The Line Economy guidance in STYLE.md and its CLAUDE.md summary leaned hard on compactness ("prefer fewer lines", "do not add blank lines"), with vertical space framed only as a grudging exception. Read literally, that pushes code toward an unbroken wall of statements with no seams, which is hard to scan even when every line is necessary.

This rebalances the rule around two forces that do not conflict. Line economy keeps a single thought on as few lines as it honestly needs; breathing room separates the distinct steps of a longer body with one blank line so the structure stays visible. The section is renamed "Line Economy and Breathing Room" and split into "keep one thought compact" and "let the code breathe", with a worked C++ example and an explicit list of where a blank line earns its place and where it does not. The CLAUDE.md at-a-glance bullet is updated to match.

[skip-ci]

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -203,10 +203,13 @@ See "Build, Test, Lint, Format" above for the `make` invocations.
 
 `STYLE.md` is the canonical source. At a glance:
 
-- **Line economy**: Prefer fewer lines for better human readability. Dense
-  code within the line-width limits is easier to scan. Do not add unnecessary
-  blank lines or spread simple logic across many lines. Always respect the
-  linting line-width limits -- never sacrifice them to shorten line count.
+- **Line economy and breathing room**: Keep a single thought compact (group
+  short declarations, do not spread simple logic across many lines), but let a
+  longer body breathe by separating its distinct steps with one blank line.
+  Do not pad short blocks, and do not compress a function into an unbroken
+  wall of statements. Always respect the linting line-width limits; never
+  sacrifice them to shorten the line count. See `STYLE.md` "Line Economy and
+  Breathing Room".
 - **C++**: 4-space indent, `m_` prefix on member vars, angle-bracket includes,
   C++23, prefer `SimpleCollector` / `small_vector` over STL for fundamentals.
 - **Python**: PEP-8, 79-char hard limit, flake8.
--- a/STYLE.md
+++ b/STYLE.md
@@ -22,32 +22,70 @@ style when adding new code and changing existing code. The rules of thumb are:
    file and follow it. Familiar with the code in the module(s) if time permits.
 3. Use the style guide.
 
-## Line Economy
+## Line Economy and Breathing Room
 
-Prefer fewer lines for better human readability. Dense code that fits within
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -203,10 +203,13 @@ See "Build, Test, Lint, Format" above for the `make` invocations.
 
 `STYLE.md` is the canonical source. At a glance:
 
-- **Line economy**: Prefer fewer lines for better human readability. Dense
-  code within the line-width limits is easier to scan. Do not add unnecessary
-  blank lines or spread simple logic across many lines. Always respect the
-  linting line-width limits -- never sacrifice them to shorten line count.
+- **Line economy and breathing room**: Keep a single thought compact (group
```
tigercosmos, line 206: Meat.

tigercosmos (commented):

STYLE.md
```diff
@@ -22,32 +22,70 @@ style when adding new code and changing existing code. The rules of thumb are:
...
+void Solver::march_one_step()
+{
+    SimpleArray<double> left = gather_left();
+    SimpleArray<double> right = gather_right();
+
+    SimpleArray<double> flux = solve_riemann(left, right);
+
+    apply_flux(flux);
+    advance_time();
+}
```
tigercosmos, line 72: Good example.

CLAUDE.md
```diff
@@ -203,10 +203,13 @@ See "Build, Test, Lint, Format" above for the `make` invocations.
...
+  short declarations, do not spread simple logic across many lines), but let a
+  longer body breathe by separating its distinct steps with one blank line.
```
yungyuc, line 208: Blank lines are good when used wisely.

STYLE.md
yungyuc: I get the idea, but this is not a good example. Blank lines only make sense when there are many lines.  An example should make it explicit:

```cpp
void Solver::march_one_step()
{
    SimpleArray<double> left = gather_left();
    SimpleArray<double> right = gather_right();
    SimpleArray<double> flux = solve_riemann(left, right);

    // A couple of more lines ...

    apply_flux(flux);
    advance_time();
}
```

And `left`, `right`, and `flux` are of the same construct and should be group together.

yungyuc (changes requested):
- [x] Improve the example.

STYLE.md
yungyuc, line 72:
```suggestion
```cpp
void Solver::march_one_step()
{
    SimpleArray<double> left = gather_left();
    SimpleArray<double> right = gather_right();
    SimpleArray<double> flux = solve_riemann(left, right);

    // A couple of more lines ...

    apply_flux(flux);
    advance_time();
}
```

yungyuc: Change like this.

tigercosmos:
@yungyuc  Fixed. Please take a look. Thanks.

yungyuc (approved):
LGTM

PR #974, opened 2026-06-30
Document the skip-ci rule in the create-pr skill

tigercosmos:
The create-pr skill now ends the PR body with the [skip-ci] control string when a change touches only agent tooling. Those changes cannot affect the build or tests, so they skip the heavy CI jobs.

Agent tooling means the .claude/ and .cursor/ trees, the root CLAUDE.md and AGENTS.md, and contrib/prompt/. The control string only works on its own line and only for a trusted author. The skill puts it on the last line of the body.

[skip-ci]

### DIFF
--- a/.claude/skills/create-pr/SKILL.md
+++ b/.claude/skills/create-pr/SKILL.md
@@ -38,6 +38,13 @@ cross-reference; flag any drift between the two.
    assistance is OK, but the user must know what the text says. When the
    skill drafts text, present it for the user's review and edits before
    posting.
+8. **Skip CI for agent-only changes.** When the diff touches *only*
+   agent tooling (the `.claude/` and `.cursor/` trees, root `CLAUDE.md` /
+   `AGENTS.md`, and `contrib/prompt/`), end the body with `[skip-ci]` on
+   its own line so the `check_skip_ci` workflow skips the heavy CI jobs.
+   Omit it if the diff touches any other file. The control string works
+   only on its own line and only for a PR opened by a repository owner,
+   member, or collaborator.
 
 ## Workflow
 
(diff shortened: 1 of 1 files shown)


.claude/skills/create-pr/SKILL.md
```diff
@@ -38,6 +38,13 @@ cross-reference; flag any drift between the two.
    assistance is OK, but the user must know what the text says. When the
    skill drafts text, present it for the user's review and edits before
    posting.
+8. **Skip CI for agent-only changes.** When the diff touches *only*
```
tigercosmos, line 41: With this rule, I don't need to teach AI how to use [skip-ci].

tigercosmos (commented):

.claude/skills/create-pr/SKILL.md
yungyuc: Please take a look at [`check_skip_ci.yml:111`](https://github.com/solvcon/solvcon/blob/6dae34622acd7f0ff89f7a390eb8879a66951d50/.github/workflows/check_skip_ci.yml#L111):

```
            const isDoc = p =>
              p.startsWith('doc/') || p.endsWith('.md') ||
              p.endsWith('.rst') || p.startsWith('contrib/prompt/');

...

            if (paths.length && paths.every(isDoc)) {
              core.notice('CI skipped: documentation-only pull request.');
              core.setOutput('skip', 'true');
              core.setOutput('cpp', 'false');
              return;
            }
```

It's more reliable to handle it in CI gating.

yungyuc (commented):
- [x] Review the CI gating.

I am fine to have agent files to duplicate the CI gating.  But CI gating is the authoritative behavior.  If it does not work correctly, it should be fixed.

.claude/skills/create-pr/SKILL.md
tigercosmos: I see. I agree this is a better way.

tigercosmos: Let's take some time and monitor the CI behavior.

yungyuc (approved):
LGTM

PR #977, opened 2026-07-01
Add SimpleArray argwhere

### LINKED ISSUE #519: Implement `SimpleArray` member functions for searching

Yuli-ytd:
Add `SimpleArray::argwhere()` as the first part of the second stage of search-method work.

The new method returns the coordinates of non-zero entries as a two-dimensional `SimpleArray<uint64_t>` with shape `{count, ndim}`. It scans the array storage and converts each matching flat index back into multi-dimensional coordinates.

Expose `argwhere()` through the Python bindings for the existing `SimpleArray` specializations, and add Python unittest coverage for both one-dimensional and N-dimensional arrays.

`SimpleArray::where()` is left for follow-up work.

Related to #519

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -1350,6 +1350,9 @@ public:
         }
         return max_index;
     }
+
+    SimpleArray<uint64_t> argwhere() const;
+
 }; /* end class SimpleArrayMixinSearch */
 
 template <typename A, typename T>
@@ -2634,6 +2637,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
         ++src;
     }
     return ret;
+}
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
+{
+    auto athis = static_cast<A const *>(this);
+
+    size_t count = 0;
+    for (size_t i = 0; i < athis->size(); ++i)
+    {
+        if (athis->data(i) != value_type(0))
+        {
+            ++count;
+        }
+    }
+
+    size_t const ndim = athis->ndim();
+    shape_type res_shape{count, ndim};
+    SimpleArray<uint64_t> result(res_shape);
+
+    size_t idx = 0;
+    for (size_t i = 0; i < athis->size(); ++i)
+    {
+        if (athis->data(i) == value_type(0)) { continue; }
+
+        size_t offset = i;
+        for (size_t dim = 0; dim < ndim; ++dim)
+        {
+            size_t stride = athis->stride(dim);
+            result(idx, dim) = static_cast<uint64_t>(offset / stride);
+            offset %= stride;
+        }
+        ++idx;
+    }
+
+    return result;
 }
 
 // The declaration lives in namespace detail and is excluded from the API
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2660,7 +2660,10 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
     size_t idx = 0;
     for (size_t i = 0; i < athis->size(); ++i)
     {
-        if (athis->data(i) == value_type(0)) { continue; }
+        if (athis->data(i) == value_type(0))
+        {
+            continue;
+        }
 
         size_t offset = i;
         for (size_t dim = 0; dim < ndim; ++dim)
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2654,7 +2654,7 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
     }
 
     size_t const ndim = athis->ndim();
-    shape_type res_shape{count, ndim};
+    shape_type const res_shape{count, ndim};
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1350,6 +1350,9 @@ class SimpleArrayMixinSearch
         }
         return max_index;
     }
+
+    SimpleArray<uint64_t> argwhere() const;
```
Yuli-ytd, line 1354: I placed `argwhere()` under `SimpleArrayMixinSearch` to keep it next to the existing index-search methods. 

Since issue #519 also lists follow-up conditional APIs such as `where()`, this may be worth moving into a separate `SimpleArrayMixinConditional` once that group of methods is added. 

Feedback on the mixin boundary would be useful.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -584,6 +584,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         (*this)
             .def("argmin", &wrapped_type::argmin)
             .def("argmax", &wrapped_type::argmax)
+            .def("argwhere", &wrapped_type::argwhere)
```
Yuli-ytd, line 587: This exposes `argwhere()` through the shared `WrapSimpleArray` template.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2636,6 +2639,42 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
```
Yuli-ytd, line 2643: This implementation first counts non-zero entries to allocate the `count` of  `{count, ndim}` result shape, then scans again to convert each matching flat index into coordinates using the existing strides. 

The return type is `SimpleArray<uint64_t>` because `SimpleArray<size_t>` is not exposed to Python, and the existing `SimpleArray::argsort()` indexing API already returns `SimpleArray<uint64_t>`. 
Feedback on the return type would be useful.

tests/test_buffer.py
```diff
@@ -3843,6 +3843,27 @@ def test_argminmax(self):
         self.assertEqual(sarr.argmax(), 9)
         self.assertEqual(narr.argmin(), sarr.argmin())
         self.assertEqual(narr.argmax(), sarr.argmax())
+    
+    def test_argwhere(self):
```
Yuli-ytd, line 3847: The test compares against `numpy.argwhere()` for both 1-D and N-D inputs.

The calls use `eq(...).argwhere()` because this PR adds the coordinate extraction step, while the conditional array construction already exists.

tests/test_buffer.py
```diff
@@ -3843,6 +3843,27 @@ def test_argminmax(self):
...
+        data = [
+            [1, 3, 5, 7, 9],
+            [2, 4, 6, 8, 10],
+            [1, 10, 1, 10, 1]
+        ]
+        narr = np.array(data, dtype='float64')
+        sarr = solvcon.SimpleArrayFloat64(array=narr)
+
+        ret = sarr.eq(10).argwhere()
+        np.testing.assert_array_equal(ret.ndarray, np.argwhere(narr == 10))
```
tigercosmos, line 3866: Why the syntax is different? I expect:
```
sarr.argwhere(narr == 10) == np.argwhere(narr == 10)
```

yungyuc:
@Yuli-ytd This PR is still in draft mode.  Are you ready for review?  Please follow the PR protocol in https://github.com/solvcon/solvcon/blob/master/.github/pull_request_template.md?plain=1 .

Yuli-ytd:
@yungyuc Thank you for the reminder. 

I read the PR protocol before creating this draft PR last night. My plan was to keep it as a draft and post a general comment requesting review after CI passed.

Since CI is failed, I will fix the failures first.
After that, I will update the PR and leave a comment to formally request review.

yungyuc:
> I read the PR protocol before creating this draft PR last night. My plan was to keep it as a draft and post a general comment requesting review after CI passed.
>
> Since CI is failed, I will fix the failures first. After that, I will update the PR and leave a comment to formally request review.

Sounds good.

Yuli-ytd:
This PR is ready for review.

tests/test_buffer.py
Yuli-ytd: I think the current API follows the member-function mapping of `np.argwhere(a)`: the input array `a` becomes the receiver, so `np.argwhere(narr)` maps to `sarr.argwhere()`.

For the conditional case, `np.argwhere(narr == 10)` first builds a boolean array and then calls `argwhere()` on that boolean array. In the current `SimpleArray` API, the equivalent is therefore `sarr.eq(10).argwhere()`, because `eq(10)` returns a `SimpleArrayBool`.

That is why the syntax is different from `sarr.argwhere(narr == 10)`. I would like to confirm whether we want to support `sarr.argwhere(condition)` as an additional API form. If so, should it be included in this PR, or should we leave it for a follow-up?

tigercosmos: Personally, I prefer having the API design as same as Numpy. It should be a small change.

yungyuc: For the comparison and selection behavior, we should match numpy.  Tracked in issue #1034.

yungyuc (approved):
LGTM.

Issue #976, opened 2026-07-01
Basic features for 3D mesh visualization


yungyuc:
Mesh and solution data need more styles (surface, lines, wireframes) of visualization for analysis and debugging.

All features need to be exposed to Python for console control, GUI, and testing.

### Group 1: Mesh rendering

- [x] Item 1. Lit surface rendering and representations
  - PR #988.
- [x] Item 2. Surface with edges and per-object opacity
  - PR #995.
- [x] Item 3. Feature edges and normal glyphs
  - PR #996.

#### Item 1: Lit surface rendering and representations

**Summary.** Render mesh cells as a lit, shaded surface, with a representation switch between surface, wireframe, and points.

**Why.** A 3D mesh drawn only as a wireframe is unreadable: depth and shape do not come across. A shaded surface is the baseline every other 3D view builds on.

**Scope.**
- Add `RSurface`, a triangle drawable over the mesh skin (2D cells, or the 3D boundary faces), with per-vertex normals computed in C++ on the `SimpleArray` path (area-weighted from the face normals `fcnml` and cell nodes `clnds`).
- Add a lit `RMaterial` variant (`lit.vert` / `lit.frag`, a Lambert plus headlight model with one directional light following the camera).
- Add a `Representation` enum (`Surface`, `Wireframe`, `Points`) on `RDrawable`; default to `Surface` for 3D and `Wireframe` for 2D.

**Python.** `widget.setRepresentation("surface"|"wireframe"|"points")`, lighting on by default for 3D.

**Done when.** A 3D sample mesh renders as a shaded solid whose pixels vary with surface orientation; a pytest compares the lit frame against the flat wireframe frame and asserts the shading gradient.

#### Item 2: Surface with edges and per-object opacity

**Summary.** Draw the wireframe over the shaded surface as one representation, and give a drawable an adjustable opacity.

**Why.** Surface-with-edges is the standard mesh-inspection view (shape plus cell structure at once). Opacity lets a user see through an outer surface to the structure behind it.

**Scope.**
- Add a `SurfaceWithEdges` representation that draws M1's surface and the wireframe together, with a small depth bias so the edges do not z-fight.
- Add per-drawable opacity through the material's alpha-blend path, with a back-to-front draw order for the transparent pass.

**Python.** `widget.setRepresentation("surface_edges")`, `widget.setOpacity(value)`.

[...]

PR #995, opened 2026-07-04
Surface with edges and per-object opacity (basic mesh visualization item 2)

yungyuc:
The mesh already draws its surface, wireframe, and points as independent layers, so the standard surface-with-edges inspection view is just the surface and the wireframe shown together. Give each drawable an opacity so a caller can fade an outer surface and read the structure behind it.

RDrawable gains a clamped opacity that scales the color alpha written into the per-object uniform, and its pipeline now enables source-over blending unconditionally so a runtime opacity change takes effect without rebuilding the pipeline; an alpha of 1 keeps the blend an identity, so opaque drawables render unchanged. The per-vertex-color shader reads the uniform alpha so the shaded surface honors the opacity too.

RDomainWidget exposes setMeshOpacity and setFieldOpacity, both bound to Python, and the domain visualizer page documents the layering.

For item 2 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RDomainWidget.hpp
+++ b/cpp/solvcon/pilot/RDomainWidget.hpp
@@ -68,6 +68,13 @@ public:
     void showMeshStyle(std::string const & name, bool show);
     bool meshStyleShown(std::string const & name) const;
 
+    /// Set the mesh wireframe opacity, a [0, 1] fraction (1 opaque).
+    void setMeshOpacity(float opacity);
+
+    /// Set the color-field surface opacity, a [0, 1] fraction (1 opaque).
+    /// Lower it to read the wireframe drawn over the shaded surface.
+    void setFieldOpacity(float opacity);
+
     /// Replace the colored field: per-vertex-colored triangles from a vertex
     /// table (nvert, 3), a matching color table (nvert, 3), and a triangle
     /// index table (ntri, 3). Swappable at runtime.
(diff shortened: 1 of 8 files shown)


cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -68,6 +68,13 @@ class RDomainWidget
     void showMeshStyle(std::string const & name, bool show);
     bool meshStyleShown(std::string const & name) const;
 
+    /// Set the mesh wireframe opacity, a [0, 1] fraction (1 opaque).
+    void setMeshOpacity(float opacity);
```
yungyuc, line 72: Provide opacity settings in the engine.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -152,6 +152,8 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRDomainWidget
             .def_property_readonly("mesh", &wrapped_type::mesh)
             .def("updateMesh", &wrapped_type::updateMesh, py::arg("mesh"))
             .def("showMesh", &wrapped_type::showMesh, py::arg("show"))
+            .def("setMeshOpacity", &wrapped_type::setMeshOpacity, py::arg("opacity"))
```
yungyuc, line 155: Provide opacity setting (Python) API.

tests/test_pilot_domain_widget.py
```diff
@@ -465,6 +465,55 @@ def test_color_field_rejects_out_of_range_index(self):
             _update_field(widget, vertices, colors, indices)
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetOpacityTC(unittest.TestCase):
```
yungyuc, line 469: Test opacity API.

PR #999, opened 2026-07-04
Mesh quality metrics visualization (item 6 of basic mesh visualization)

yungyuc:
Add a per-cell geometric quality visualization: compute a quality metric for every cell over the SimpleArray mesh path and color the surface by it through the continuous colormap and scalar bar, reporting the metric range.

The metrics are volume (the cell volume), aspect_ratio (the ratio of the largest to smallest node distance), and the node-polygon angle metrics skewness, min_angle, and max_angle. colorByQuality colors the surface, giving a boundary face its owning cell's value in 3D; qualityRange reports the metric (min, max) over the cells.

Factor the surface-scalar geometry out of the categorical coloring into a shared collectSurfaceScalars, so the categorical and continuous fields build the same fan-triangulated, per-primitive-flat surface. Expose colorByQuality and qualityRange to Python, and cover the metric values against a hand computation, the coloring, and the unknown-metric rejection with tests.

For item 6 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RDomainWidget.cpp
+++ b/cpp/solvcon/pilot/RDomainWidget.cpp
@@ -20,11 +20,116 @@
 #include <QWheelEvent>
 
 #include <algorithm>
+#include <cmath>
 #include <limits>
 #include <stdexcept>
+#include <vector>
 
 namespace solvcon
 {
+
+namespace
+{
+
+constexpr float PI = 3.14159265358979323846f;
+
+/// Per-cell value of the named geometric quality metric, over the SimpleArray
+/// mesh path. The angle-based metrics read the cell node polygon, so they are
+/// meaningful for 2D cells; "volume" and "aspect_ratio" work in any dimension.
+/// Throws std::invalid_argument for an unknown metric name.
+std::vector<float> cell_quality(StaticMesh const & mh, std::string const & metric)
+{
+    static char const * const known[] = {
+        "volume", "aspect_ratio", "skewness", "min_angle", "max_angle"};
+    if (std::none_of(std::begin(known), std::end(known), [&metric](char const * n)
+                     { return metric == n; }))
+    {
+        throw std::invalid_argument(
+            "RDomainWidget: unknown quality metric \"" + metric +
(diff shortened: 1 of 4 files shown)


cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -132,6 +133,14 @@ class RDomainWidget
     void colorByBoundary();
     void clearCellColoring();
 
+    /// Color the mesh by a per-cell geometric quality metric through the
+    /// continuous colormap and scalar bar. @p metric is one of "volume",
+    /// "aspect_ratio", "skewness", "min_angle", "max_angle".
+    void colorByQuality(std::string const & metric);
```
yungyuc, line 139: Visualize mesh by predefined quality metrics.

cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -20,12 +20,117 @@
...
+constexpr float PI = 3.14159265358979323846f;
+
+/// Per-cell value of the named geometric quality metric, over the SimpleArray
+/// mesh path. The angle-based metrics read the cell node polygon, so they are
+/// meaningful for 2D cells; "volume" and "aspect_ratio" work in any dimension.
+/// Throws std::invalid_argument for an unknown metric name.
+std::vector<float> cell_quality(StaticMesh const & mh, std::string const & metric)
+{
+    static char const * const known[] = {
+        "volume", "aspect_ratio", "skewness", "min_angle", "max_angle"};
```
yungyuc, line 43: Use strings for now.  The quality metrics calculation is heavy.  We will evaluate when to replace them to be enum or more proper constructs.

See the follow-up issue #1000.

cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -20,12 +20,117 @@
...
 #include <stdexcept>
+#include <vector>
 
 namespace solvcon
 {
 
+namespace
+{
+
+constexpr float PI = 3.14159265358979323846f;
```
yungyuc, line 34: Follow-up issue #1001.

PR #1002, opened 2026-07-04
View presets and projection toggle (item 7 of basic mesh visualization)

yungyuc:
Add axis-aligned and isometric view presets and an explicit parallel-versus-perspective projection toggle, so the projection is no longer tied to the domain dimension.

RScene gains a projection override (auto, parallel, perspective) that viewProjection honors over the per-dimension default, and setViewPreset that points the camera along a named direction ("front", "back", "left", "right", "top", "bottom", the axis names "+x".."-z", or "iso") and frames the scene at a distance sized to the bounding sphere and viewport.

RDomainWidget exposes setView and a projection property, both bound to Python. Tests cover the projection round trip, the preset framing, that two presets view from different directions, and that the projection toggle changes the rendered frame.

For item 7 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RScene.cpp
+++ b/cpp/solvcon/pilot/RScene.cpp
@@ -6,6 +6,7 @@
 #include <solvcon/pilot/RScene.hpp> // Must be the first include.
 
 #include <algorithm>
+#include <cmath>
 
 namespace solvcon
 {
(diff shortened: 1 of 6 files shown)


cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -248,6 +248,25 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRDomainWidget
                 "cells.")
             .def("showAxis", &wrapped_type::showAxis, py::arg("show"))
             .def("fitCameraToScene", &wrapped_type::fitCameraToScene)
+            .def(
```
yungyuc, line 251: Python API for different view angles and positions.  Only exercise in unit testing because there is not a convenient interactive console and GUI for the feature.

cpp/solvcon/pilot/RScene.cpp
```diff
@@ -93,6 +94,98 @@ void RScene::fitCameraToScene(float aspect)
     m_camera.fitToBoundingBox(m_bbox_lo, m_bbox_hi, m_ndim, aspect);
 }
 
+void RScene::setProjection(std::string const & name)
```
yungyuc, line 97: The visual engine code for the view is in `RScene`.

PR #1003, opened 2026-07-04
Trackball orbit and pivot control (item 8 of basic mesh visualization)

yungyuc:
Add a free trackball orbit beside the turntable, and let the orbit pivot move off the scene center.

The turntable orbit holds the up axis fixed so the horizon stays level. The trackball orbit instead rolls the up axis with the drag: it yaws and pitches the eye offset and applies the same rotation to the up axis, with no pole guard, so the eye can tumble freely over the top the way a virtual-trackball drag does. setPivot points the orbit at an explicit center, keeping the eye put so the next drag swings around the new point.

RDomainWidget exposes an orbitStyle property, setOrbitStyle, setPivot, and frameSelected (which frames the whole scene until picking supplies a selection), all bound to Python. Tests check the style round trip, that the turntable keeps the up axis level while the trackball rolls it, and that a set pivot becomes the center of rotation.

For item 8 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RCameraController.cpp
+++ b/cpp/solvcon/pilot/RCameraController.cpp
@@ -54,6 +54,20 @@ std::string RCameraController::modeName(Mode mode)
     }
 }
 
+RCameraController::OrbitStyle RCameraController::orbitStyleFromName(std::string const & name)
+{
+    if ("trackball" == name)
+    {
+        return OrbitStyle::Trackball;
+    }
+    return OrbitStyle::Turntable;
+}
+
+std::string RCameraController::orbitStyleName(OrbitStyle style)
+{
+    return (OrbitStyle::Trackball == style) ? "trackball" : "turntable";
+}
+
 QVector3D RCameraController::forward() const
 {
     QVector3D const dir = m_target - m_position;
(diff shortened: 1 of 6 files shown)


cpp/solvcon/pilot/RCameraController.cpp
```diff
@@ -110,7 +124,22 @@ QMatrix4x4 RCameraController::viewMatrix() const
 
 void RCameraController::rotate(float dx, float dy)
 {
-    if (Mode::Orbit == m_mode)
+    if (Mode::Orbit == m_mode && OrbitStyle::Trackball == m_orbit_style)
```
yungyuc, line 127: Add an alternate camera control for different yaw.

cpp/solvcon/pilot/RCameraController.hpp
```diff
@@ -64,6 +64,27 @@ class RCameraController
     static Mode modeFromName(std::string const & name);
     static std::string modeName(Mode mode);
 
+    /**
+     * The orbit style. Turntable holds the up axis fixed so the horizon never
+     * rolls; Trackball tumbles freely, rolling the up axis with the drag.
+     *
+     * @ingroup group_domain
+     */
+    enum class OrbitStyle
```
yungyuc, line 73: Two orbiting styles.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -267,6 +267,35 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRDomainWidget
                 },
                 "Projection: \"auto\" (orthographic 2D, perspective 3D), "
                 "\"parallel\", or \"perspective\".")
+            .def_property(
```
yungyuc, line 270: Python API.  Only done unit testing without visual inspection.

tests/test_pilot_domain_widget.py
```diff
@@ -1045,6 +1045,75 @@ def test_projection_toggle_changes_the_frame(self):
         self.assertTrue((para_frame != persp_frame).any())
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetTrackballTC(unittest.TestCase):
```
yungyuc, line 1049: Unit testing for the trackball camera control.

PR #1006, opened 2026-07-04
Mesh entity picking and inspection (item 10 of basic mesh visualization)

yungyuc:
Add interactive picking to the domain viewer: a click resolves to the cell, node, or face under the cursor, reports its identity and geometry, and highlights it.

Picking back-projects the pixel through the same view-projection the renderer uses and casts the ray on the CPU: a cell is the nearest surface hit (the boundary shell in 3D, a point-in-cell test in 2D), a node is the nearest node to the ray, and a face is the nearest boundary face (3D). pickCell / pickNode / pickFace return the id, element type, a measure (cell volume or face area), and the centroid, and record the entity as the selection with its bounding box for later framing. A picked cell gets a bright surface highlight.

The pick methods are bound to Python, returning a dict or None on a miss, alongside clearSelection and hasSelection. Tests pick a known pixel of a unit quad and check the returned id and geometry, cover the background miss, the no-mesh case, node and 3D-face picks, and the selection clear.

For item 10 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RDomainWidget.cpp
+++ b/cpp/solvcon/pilot/RDomainWidget.cpp
@@ -17,6 +17,7 @@
 #include <QMouseEvent>
 #include <QNativeGestureEvent>
 #include <QPinchGesture>
+#include <QVector4D>
 #include <QWheelEvent>
 
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -781,6 +782,406 @@ void RDomainWidget::showNormals(bool show)
     update();
 }
 
+namespace
+{
+
+/// Moeller-Trumbore ray/triangle intersection. On a hit, @p t is the ray
+/// parameter of the front-facing or back-facing intersection.
+bool ray_triangle(
```
yungyuc, line 790: Get intersection between the visual ray and the cell face.

cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -124,6 +124,31 @@ class RDomainWidget
     /// Show or hide a short arrow at every face center along its normal.
     void showNormals(bool show);
 
+    /// The result of a pick: the entity kind ("cell", "node", "face", or
+    /// "none" for a miss), its id, its element type (for a cell), a measure
+    /// (cell volume or face area), and its centroid.
+    struct PickResult
```
yungyuc, line 130: Put the picking helper class in `RDomainWidget` for now.  Refactor in the future.

See follow-up issue #1007.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -123,6 +123,23 @@ namespace solvcon
 namespace python
 {
 
+/// Convert a pick result to a Python dict, or None on a miss.
+static pybind11::object pick_to_py(RDomainWidget::PickResult const & r)
```
yungyuc, line 127: We'll need a wrapper in the future.

tests/test_pilot_domain_widget.py
```diff
@@ -1114,6 +1114,110 @@ def test_frame_selected_frames_the_scene(self):
...
+    mh.cltpn.ndarray[:] = [Q]
+    mh.clnds.ndarray[:, :5] = [(4, 0, 1, 2, 3)]
+    mh.build_interior()
+    mh.build_boundary()
+    mh.build_ghost()
+    return mh
+
+
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetPickTC(unittest.TestCase):
```
yungyuc, line 1133: Only exercised in unit testing.  There is not a GUI control yet.

PR #1008, opened 2026-07-04
Measurement tools, zoom to selection, and camera reset (items 11 and 12 of basic mesh visualization)

yungyuc:
Measurement tools and Zoom to selection and camera reset.

## Item 11: Measurement tools

Add measurement tools: `measureDistance` returns the distance between two world points, and `measureAngle` returns the angle at a vertex between the arms to two other points, each drawing the measured geometry as a ruler over the scene.

`RSegments` is a small flat-colored line drawable that takes points in pairs and draws each pair as one line; the ruler is a magenta segment for a distance or two arms for an angle. `measureDistance` and `measureAngle` install the ruler and return the value; `clearMeasurements` removes it. All three are bound to Python, taking `(x, y, z)` sequences.

Tests check a 3-4-5 distance, the distance between two known mesh nodes, a right angle and a straight angle, and that clearing the ruler is harmless.

## Item 12: Zoom to selection and camera reset

Add the two most-used camera shortcuts: frame the current selection, and return to a known view.

`RScene` gains a framing box: `frameBox` points the camera at an arbitrary box and sizes the projection to it, so the view actually zooms in rather than only recentering. `fitCameraToScene` clears it and frames the whole scene again. `zoomToSelection` frames the picked entity's bounding box (or the whole scene when nothing is selected), and `resetCamera` restores the fit-to-scene default: auto projection, turntable orbit, the whole scene framed. `frameSelected` now routes through `zoomToSelection`.

Both are bound to Python. Tests pick the right-hand quad and check that zooming moves the camera target to the cell centroid while reset returns it to the scene center, and that zoom with no selection frames the scene.

For items 11 and 12 of issue #976.


tests/test_pilot_domain_widget.py
```diff
@@ -1283,6 +1283,104 @@ def swept(factor):
         self.assertGreater(swept(2.0), swept(1.0))
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetZoomTC(unittest.TestCase):
```
yungyuc, line 1287: Testing zoom.

tests/test_pilot_domain_widget.py
```diff
@@ -1283,6 +1283,104 @@ def swept(factor):
...
+        widget.pickCell(240, 120)
+        widget.zoomToSelection()
+        widget.resetCamera()
+        target = tuple(widget.cameraTarget)
+        self.assertAlmostEqual(target[0], 1.0, places=3)
+        self.assertAlmostEqual(target[1], 0.5, places=3)
+
+
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetMeasureTC(unittest.TestCase):
```
yungyuc, line 1337: Testing measurement.

cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -149,6 +149,17 @@ class RDomainWidget
     void clearSelection();
     bool hasSelection() const { return m_has_selection; }
 
+    /// Measure and draw the distance between two world points; returns the
+    /// distance.
+    double measureDistance(QVector3D const & p0, QVector3D const & p1);
```
yungyuc, line 154: Visual engine for measurement.

cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -190,6 +201,14 @@ class RDomainWidget
     /// Recenter and frame the whole scene (the selection once picking lands).
     void frameSelected();
 
+    /// Frame the current selection (from a pick), or the whole scene when
+    /// nothing is selected.
+    void zoomToSelection();
```
yungyuc, line 206: Visual engine for zooming.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -258,6 +265,28 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRDomainWidget
                 "None.")
             .def("clearSelection", &wrapped_type::clearSelection)
             .def_property_readonly("hasSelection", &wrapped_type::hasSelection)
+            .def(
```
yungyuc, line 268: Python API exposed for measurement and zooming.

PR #1010, opened 2026-07-04
Geometric slice and clip of the mesh (item 14 of basic mesh visualization)

yungyuc:
Add two geometric cuts that reveal a mesh interior without any field coloring.

addClip keeps the surface primitives (2D cells, or 3D boundary faces) whose centroid lies on one side of a plane and drops the rest, drawing the kept side as a shaded surface. addSlice cuts every cell by the plane: it intersects the plane with the cell edges (the polygon rim in 2D, every node pair in 3D) and draws the crossing points as a cross-section outline. Both re-evaluate from the plane on each call, and clearFilters removes them.

addClip and addSlice are bound to Python, taking (origin, normal) sequences and returning the kept-primitive and segment counts. Tests check that a plane keeps the expected cells, that a slice cuts the 2D mesh and the 3D tetrahedron, and that clearing is harmless.

For item 14 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RDomainWidget.cpp
+++ b/cpp/solvcon/pilot/RDomainWidget.cpp
@@ -1231,6 +1231,220 @@ void RDomainWidget::clearMeasurements()
     update();
 }
 
+int RDomainWidget::addClip(QVector3D const & origin, QVector3D const & normal)
+{
+    m_scene.removeDrawable(m_clip);
+    m_clip = nullptr;
+    if (nullptr == m_mesh)
+    {
+        return 0;
+    }
+    StaticMesh const & mh = *m_mesh;
+    uint32_t const ndim = mh.ndim();
+    QVector3D const n = normal.normalized();
+    auto coord = [&mh, ndim](int32_t ind, uint32_t dm) -> float
+    { return (dm < ndim) ? static_cast<float>(mh.ndcrd(ind, dm)) : 0.0f; };
+    auto side = [&](QVector3D const & p) -> float
+    { return QVector3D::dotProduct(p - origin, n); };
+
+    std::vector<float> verts;
+    std::vector<float> cols;
+    std::vector<uint32_t> tris;
+    auto add_polygon = [&](auto node, int32_t nnd)
+    {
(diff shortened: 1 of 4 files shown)


cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -1231,6 +1231,220 @@ void RDomainWidget::clearMeasurements()
     update();
 }
 
+int RDomainWidget::addClip(QVector3D const & origin, QVector3D const & normal)
```
yungyuc, line 1234: Clipping keeps the cell shape.

cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -1231,6 +1231,220 @@ void RDomainWidget::clearMeasurements()
...
+        {
+            m_clip = clip.get();
+            m_scene.addDrawable(std::move(clip));
+        }
+    }
+    update();
+    return kept;
+}
+
+int RDomainWidget::addSlice(QVector3D const & origin, QVector3D const & normal)
```
yungyuc, line 1347: Slicing uses a flat cutting surface.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -287,6 +287,27 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRDomainWidget
                 "Measure and draw the angle (degrees) at p1 between the arms "
                 "to p0 and p2.")
             .def("clearMeasurements", &wrapped_type::clearMeasurements)
+            .def(
```
yungyuc, line 290: Python API.

tests/test_pilot_domain_widget.py
```diff
@@ -1283,6 +1283,50 @@ def swept(factor):
         self.assertGreater(swept(2.0), swept(1.0))
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetFilterTC(unittest.TestCase):
```
yungyuc, line 1287: Unit testing for clipping and slicing.

PR #1012, opened 2026-07-04
Scene of multiple meshes and offscreen capture (items 15 and 16 of basic mesh visualization)

yungyuc:
Scene of multiple meshes with transforms and an outliner and High-resolution and transparent offscreen capture.

## Item 15: Scene of multiple meshes with transforms and an outliner

Let the viewer hold several named mesh objects at once, each with its own model transform and visibility, the basis an outliner would list.

- `RDrawable` gains a model transform, applied before the view-projection so a single object still renders in place, and a name for the scene registry.
- `RDomainWidget` registers objects by name: `addObject` adds a mesh as a named lit surface and frames the union; `setObjectTransform` sets a translate and scale and grows the framing box to the transformed corners; `setObjectVisible` and `setObjectOpacity` drive one object by name; `objectNames` lists them.
- The API is bound to Python.

Tests check that two named meshes register, that re-adding a name replaces it, that the by-name setters accept a known name and ignore an unknown one, and that two transformed objects render and toggle.

## Item 16: High-resolution and transparent offscreen capture

Add an offscreen capture at an arbitrary resolution decoupled from the widget size, with an optional transparent background, for publication figures that should not depend on the on-screen window.

- `renderToImage` renders the scene at a requested width and height by resizing the (possibly hidden) widget, grabbing the frame, and restoring the size; a transparent flag clears the frame's alpha so the background composites cleanly.
- Bound to Python as `renderToImage(path, width, height, transparent)`, writing the capture to a file.

Tests check that a capture is produced at the requested size regardless of the widget size, that a transparent capture leaves the corner not fully opaque, and that a non-positive size writes nothing.

For items 15 and 16 of issue #976.

### DIFF
--- a/cpp/solvcon/pilot/RDomainWidget.cpp
+++ b/cpp/solvcon/pilot/RDomainWidget.cpp
@@ -223,6 +223,128 @@ void RDomainWidget::updateMesh(std::shared_ptr<StaticMesh> const & mesh)
     update();
 }
 
+namespace
+{
+
+/// The axis-aligned node bounds of a mesh.
+void mesh_bounds(StaticMesh const & mh, QVector3D & lo, QVector3D & hi)
+{
+    bool const is_3d = (3 == mh.ndim());
+    lo = QVector3D(
+        std::numeric_limits<float>::max(),
+        std::numeric_limits<float>::max(),
+        std::numeric_limits<float>::max());
+    hi = QVector3D(
+        std::numeric_limits<float>::lowest(),
+        std::numeric_limits<float>::lowest(),
+        std::numeric_limits<float>::lowest());
+    for (uint32_t ind = 0; ind < mh.nnode(); ++ind)
+    {
+        float const x = static_cast<float>(mh.ndcrd(ind, 0));
+        float const y = static_cast<float>(mh.ndcrd(ind, 1));
+        float const z = is_3d ? static_cast<float>(mh.ndcrd(ind, 2)) : 0.0f;
+        lo = QVector3D(std::min(lo.x(), x), std::min(lo.y(), y), std::min(lo.z(), z));
+        hi = QVector3D(std::max(hi.x(), x), std::max(hi.y(), y), std::max(hi.z(), z));
+    }
+}
+
+} /* end namespace */
+
+void RDomainWidget::addObject(
+    std::string const & name, std::shared_ptr<StaticMesh> const & mesh)
+{
+    auto const it = m_objects.find(name);
+    if (it != m_objects.end())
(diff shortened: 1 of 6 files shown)


cpp/solvcon/pilot/RDomainWidget.cpp
```diff
@@ -157,6 +157,24 @@ QImage RDomainWidget::grabImage()
     return grabFramebuffer();
 }
 
+QImage RDomainWidget::renderToImage(int width, int height, bool transparent)
```
yungyuc, line 160: Use GPU / graphical library to render to image.

cpp/solvcon/pilot/RDomainWidget.hpp
```diff
@@ -361,6 +383,15 @@ class RDomainWidget
...
     RDrawable * m_slice = nullptr; ///< The slice cross-section outline.
+
+    /// A named scene object: its drawable (owned by the scene) and its mesh,
+    /// kept so a transform can re-extend the framing box.
+    struct ObjectEntry
+    {
+        RDrawable * drawable = nullptr;
+        std::shared_ptr<StaticMesh> mesh;
+    };
+    std::map<std::string, ObjectEntry> m_objects;
```
yungyuc, line 394: Allows to hold more than one mesh in a widget (sub-window).

tests/test_pilot_domain_widget.py
```diff
@@ -1283,6 +1283,117 @@ def swept(factor):
         self.assertGreater(swept(2.0), swept(1.0))
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetCaptureTC(unittest.TestCase):
```
yungyuc, line 1287: Unit testing image rendering.

tests/test_pilot_domain_widget.py
```diff
@@ -1283,6 +1283,117 @@ def swept(factor):
...
+        widget.resize(320, 240)
+        widget.updateMesh(_make_2d_mesh())
+        with tempfile.TemporaryDirectory() as folder:
+            path = os.path.join(folder, "bad.png")
+            self.assertFalse(widget.renderToImage(path, 0, 0))
+            self.assertFalse(os.path.exists(path))
+
+
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RDomainWidgetSceneObjectsTC(unittest.TestCase):
```
yungyuc, line 1337: Unit testing multiple mesh objects in a scene.

PR #1014, opened 2026-07-04
Action placement, identity, and groups in the menu model (step 2 of menu model)

### LINKED ISSUE #989: Restructure the pilot menu system
Restructure the pilot menu system to let any pilot feature, whether written in C++ or in Python, contribute a menu item at a named location without editing the window manager, and make the menu bar's structure and ordering a declared, queryable property of the running program rather than an accident of construction order. 

## Goal

1. **One shared placement path.** A feature places an action by menu path ("Mesh", "View/Panels") and weight, through one API, from either language. Action *creation* stays in the language that owns the callback (see "Why actions are created per language" below), but placement, identity, and ordering are shared.
2. **A live, Python-scriptable menu model.** The menus are not a build-time input: the model can be queried and edited at runtime from the embedded Python console. Look up an action by id, look up or create a menu by path, place, remove, and reorder items. The pilot ships a console dock precisely so the GUI can be scripted; the menu bar should not be the exception.
3. **Honest, discoverable shortcuts.** Shortcuts live in `QKeySequence` where Qt displays them, never in label text, and a menu item never advertises a key it does not own.
4. **One source of truth for exclusive selections.** The camera mode and the canvas draw tool each become a `QActionGroup` that drives every surface showing that state (menu radio items, the Painter toolbox buttons).

The planned steps come right below. Every section after them exists to support the steps: what the current code does and why it must change, the practice the design borrows from, the design itself, and how the result is tested.

## Planned steps

Each step is one pull request of fewer than 300 changed lines, independently reviewable, leaving the pilot working. The vocabulary used here (`RMenuModel`, `place`, weights, ids) is defined in "Target design"; the evidence for each step is in "What we have today".

1. **Step 1. `RMenuModel` skeleton: menus by path.** Add the class, owned by `RManager` and cleared from `RManager::reset()`, with `menu(path)` resolution that creates ancestors on demand and the menu-node weight table that orders menus among their siblings and on the bar. Bind it and test path creation, idempotency, and node ordering from Python. The same binding pull request also adds the missing one-line `editMenu` binding recorded under "Problems". No consumer changes yet. (~255 lines)  PR #1013.
3. **Step 2. Action placement and identity.** Add `place`, `placeSeparator`, item weights, the id registry (`action(id)`, `remove`), and `group(id)`; add the one-line `QT_TYPE_CASTER(QAction, ...)` so PySide actions cross. Test weight-ordered insertion, equal-weight stability, id round-trip, and removal from Python. (~250 lines)  PR #1015.
4. **Step 3. Seed the bar from the model.** Replace the `setUpMenu` body with the menu-node table reproducing today's bar exactly, turn the eight getters into thin adapters over `menu(path)`, and route the Edit items through `place`. Behavior-preserving by construction; no Python call site changes. (~150 lines)  PR #1015.
5. **Step 4. The camera menu as data, with a working Escape.** Convert the eleven camera-movement blocks and `createCameraMovementItemHandler` into the `CameraMove` table, delete the false "(key)" label hints, and register the camera-mode `QActionGroup` under a group id; mostly deletions. In the same pull request, make the Escape reset real: attach the reset action to each new viewer in `add3DWidget` (skipping the hidden RHI primer), which is the missing association that lets its `Qt::WidgetShortcut` fire, and test the association. (~240 lines)  PR #1015.
6. **Step 5. One Python item builder.** Add `PilotFeature.add_action` over `place`, reimplement `_add_menu_item` and the controller's `_addAction` on it, and give "Exit" an explicit `QuitRole` through the `menu_role` knob, replacing the darwin special case. Existing call signatures keep working, so no feature changes yet. (~120 lines)  PR #1016.
7. **Step 6. Migrate the flat features.** Give the non-toggle `populate_menu` methods explicit paths, ids, and weights: the mesh items, the oblique-shock solver, the three 1D apps, the SVG and Gmsh dialogs, the profiling items, and the Canvas sample and working items. (~200 lines)
8. **Step 7. Panels as a path.** Convert the Painter and MeshInfo toggles to `add_action`, replace the hand-built "Panels" submenu with the "View/Panels" path, drop the `menu=` constructor injection, and rewrite `tests/test_pilot_mesh_info.py` against `mgr.menu_model`, since it constructs and injects its own menu today. (~250 lines)  PR #1016.
9. **Step 8. The draw-tool action group.** One checkable action per registered tool in an exclusive `QActionGroup`, surfaced as "Canvas/Draw tool" radio items; the Painter toolbox buttons become `QToolButton::setDefaultAction` views of the same actions; the Canvas samples move under "Canvas/Samples". (~250 lines)  PR #1016.
10. **Step 9. Delete the getters.** Remove the eight `QMenu*` members, the adapter getters, and their bindings; everything reaches menus through the model. (~80 lines)Steps 1 and 2 are the load-bearing ones; every later step is small because they exist. Step 4 is an independent fix that can land in any order relative to its neighbors.  PR #1016.

## What we have today

This section records the machinery the steps above replace, and the evidence that it must change.

The pilot menu bar is assembled in phases. C++ `RManager::setUp` creates the top-level menus and the built-in Edit and View items; the Python controller then inserts the "Panels" submenu into View, constructs the features, and calls each feature's `populate_menu` in a fixed sequence. Menu structure therefore accretes across three phases whose orders are all implicit.

`RManager::setUpMenu` builds eight top-level menus imperatively, and their order on the menu bar is exactly the order of these statements (`cpp/solvcon/pilot/RManager.cpp:288`):

```cpp
m_fileMenu      = m_mainWindow->menuBar()->addMenu(QString("File"));
m_editMenu      = m_mainWindow->menuBar()->addMenu(QString("Edit"));
setUpEditMenuItems();
m_viewMenu      = m_mainWindow->menuBar()->addMenu(QString("View"));
{ setUpCameraControllersMenuItems(); setUpCameraMovementMenuItems(); }
m_oneMenu       = m_mainWindow->menuBar()->addMenu(QString("One"));
m_meshMenu      = m_mainWindow->menuBar()->addMenu(QString("Mesh"));
m_canvasMenu    = m_mainWindow->menuBar()->addMenu(QString("Canvas"));
m_profilingMenu = m_mainWindow->menuBar()->addMenu(QString("Profiling"));
m_windowMenu    = m_mainWindow->menuBar()->addMenu(QString("Window"));
```

Each menu is backed by a raw `QMenu*` member and exposed through a getter, eight of each (`cpp/solvcon/pilot/RManager.hpp:81` and `:132`). The Python side reads those getters, for example `self._mgr.meshMenu` in the mesh features.

Python features register items through a shared helper that builds a bare `QAction` against a menu handed in by the caller (`solvcon/pilot/_gui_common.py:44`):

```python
def _add_menu_item(self, menu, text, tip, func):
    act = QtGui.QAction(text, self._mainWindow)
    act.setStatusTip(tip)
    act.triggered.connect(func)
    menu.addAction(act)
```

The controller instantiates the features in one order (`solvcon/pilot/_gui.py:91`) and calls their `populate_menu` methods in a second, different order (`solvcon/pilot/_gui.py:140`), a hand-maintained list of twelve calls followed by two more items added inline.

The recent Mesh-menu rework (PR #986) is part of the picture. It collapsed the sample meshes into a `SampleMeshDialog` opened from a single Mesh item (`solvcon/pilot/_mesh.py:277`), fed by per-feature `mesh_sample_dialog_entries()` methods that return `(category, label, tip, func)` tuples which the controller concatenates (`solvcon/pilot/_gui.py:113`). That is a declarative spec invented ad hoc for one dialog: the codebase is already converging on menu-items-as-data, one private convention at a time. It also landed without its tests.

### Problems

Each problem names the step that removes it.

- **Three parallel item builders that have drifted** (step 5). `PilotFeature._add_menu_item` (`solvcon/pilot/_gui_common.py:44`) makes only flat, non-checkable items. The controller carries its own `_addAction` (`solvcon/pilot/_gui.py:125`) that additionally supports `checkable`, `checked`, and a dotted-string callable. Because the shared helper cannot make a checkable item, the two dock toggles hand-roll `QAction` + `setCheckable` + `toggled.connect` inline (`solvcon/pilot/_painter_gui.py:44` and `solvcon/pilot/_mesh_info.py:192`). Four ways to do one thing, and the `SampleMeshDialog` entry tuples add a fifth, parallel item-spec format.
- **Layout is implicit construction order, in three phases at once** (steps 1, 3, 6). The top-level order lives in `setUpMenu`; the "Panels" position is a special `insertMenu` in the controller (`solvcon/pilot/_gui.py:83`); the item order within each menu is whichever `populate_menu` ran first. Reasoning about the final bar means reading all three and mentally merging them.
- **A getter per menu does not scale, and is already inconsistent** (steps 1, 9). Adding a top-level menu means editing the header (member and getter), the source (creation), and the pybind11 wrapper. The eight getters exist in C++, but only seven are bound: `editMenu` is missing from the bindings (`cpp/solvcon/pilot/wrap_pilot.cpp:442`), so Python cannot reach the Edit menu at all. The boilerplate has already fallen out of sync.
- **No general submenu path** (steps 1, 7). The only Python submenu, "Panels", is built by hand in the controller and threaded into the two features that need it through a `menu=` constructor argument (`solvcon/pilot/_gui.py:83`, injected at `:94` and `:104`). There is no way for a feature to say "put this under View/Panels" and have the submenu created on demand.
- **The camera-movement menu misdescribes the keyboard** (step 4). The menu labels claim keys, for example `"Move camera up (W/UP)"` (`cpp/solvcon/pilot/RManager.cpp:390`), but those keys are handled elsewhere with different behavior: `RDomainWidget::keyPressEvent` (`cpp/solvcon/pilot/RDomainWidget.cpp:306`) maps W to `camera().moveForward(0.1)` while the menu item claiming "(W/UP)" calls `panCamera(0, 40)`. The key switch also ignores modifiers, so the item labeled "(Ctrl+W/UP)" describes a combination that is handled identically to plain W. The hints are not merely unwired; they are wrong today.
- **The Escape reset shortcut can never fire** (step 4). `reset_camera` gets `QKeySequence(Qt::Key_Escape)` with `Qt::WidgetShortcut` context (`cpp/solvcon/pilot/RManager.cpp:449`), but the action is parentless and is only ever added to the menu, never to a viewer via `QWidget::addAction`, so no widget it is associated with can have focus. The one shortcut the camera menu tries to wire is inert.
- **Actions leak** (step 2). The C++ `RAction`s are constructed with no parent (`cpp/solvcon/pilot/RManager.cpp:312`), and `QMenu::addAction` does not take ownership, so nothing ever deletes them.
- **Repetition in the camera menu** (step 4). Eleven camera-movement actions are built from near-identical blocks, each wrapping one call in `createCameraMovementItemHandler` (`cpp/solvcon/pilot/RManager.cpp:377`). This is data (label, tip, movement) expressed as code.
- **Exclusive state is scattered** (step 8). The camera modes have a proper `QActionGroup` (`cpp/solvcon/pilot/RManager.cpp:361`), but the draw-tool selection has no action representation at all: it lives in a `QButtonGroup` of toolbox buttons (`solvcon/pilot/_painter_gui.py:70`) that call `mgr.setDrawTool`, with `present()` manually re-checking the default button to keep the UI in step. A second surface showing the current tool would have to invent its own synchronization.
- **Almost nothing is tested** (steps 1, 2, 7 and "Testing"). Only `tests/test_pilot_mesh_info.py` asserts on a menu, and only that one feature's action lands in a `QMenu` the test itself constructs and injects. There is no test of the real bar's structure or ordering, and upstream PR 986 (solvcon)'s dialog shipped with no tests at all.

## Best practices from the research

The design the steps build follows patterns surveyed from Qt's own facilities and the KDE XMLGUI action system, the canonical reference for plugin-style menu contribution.

**Separate action identity from menu placement (KDE XMLGUI).** KDE stores every `QAction` in a `KActionCollection` keyed by a stable string name and describes placement separately, so creation (which may come from any plugin) is fully decoupled from layout. See <https://develop.kde.org/docs/getting-started/using_actions/>. This is the shape of steps 1 and 2.

**Qt gives the primitives to build a small live model.** `QMenu::addMenu` creates and returns a submenu, so a path resolver can walk "View/Panels", create each missing level, and cache it. `insertAction`, `insertMenu`, and `insertSeparator` place entries at a defined position rather than by append order (<https://doc.qt.io/qt-6/qmenu.html>). `objectName` gives every action a stable handle for lookup and testing (<https://doc.qt.io/qt-6/qaction.html>).

**Shortcuts belong in `QKeySequence`.** `QKeySequence::StandardKey` yields the platform-correct binding for Undo, Redo, Open, and Quit, and Qt renders the shortcut in the menu automatically (<https://doc.qt.io/qt-6/qkeysequence.html>). Shortcut context matters: a `Qt::WidgetShortcut` action is active only while a widget it was added to (via `QWidget::addAction`) has focus, which is exactly why the current Escape binding is dead and how step 4 fixes it. Platform roles matter too: on macOS Qt relocates actions that look like Quit, About, or Preferences into the native application menu, which is why the controller special-cases "Exit" off darwin (`solvcon/pilot/_gui.py:153`); step 5's `menu_role` knob keeps that explicit (<https://doc.qt.io/qt-6/qaction.html#menuRole-prop>).

**`QActionGroup` for exclusive sets.** A group of checkable actions gives radio behavior and a queryable `checkedAction` (<https://doc.qt.io/qt-6/qactiongroup.html>), and the same `QAction`s can drive toolbox buttons through `QToolButton::setDefaultAction`, so one group can be the single source of truth for every surface (<https://doc.qt.io/qt-6/qtoolbutton.html#setDefaultAction>). This is what step 8 applies to the draw tools.

A note on scope: the pilot code is the modmesh codebase, and modmesh has not grown a registry or declarative menu system of its own. This is a forward step for the project, not a catch-up to an upstream that already solved it.

## Why actions are created per language

The steps deliberately never move a callback across the language boundary. An earlier draft of this plan proposed a single C++ `add(spec)` taking a `std::function` callback that Python would also call. Checking that against the codebase kills it:

- The Qt bridge in `wrap_pilot.cpp` is a `qt_type_caster` that works only for QObject-derived types via PySide (`cpp/solvcon/pilot/wrap_pilot.cpp:39`). `QString` and `QKeySequence` are not QObjects and do not cross; every existing binding uses `std::string`.
- A Python callable stored in a C++ `std::function` changes failure modes: a PySide signal connection that raises prints a traceback and the app continues, while an exception thrown through the pybind11 functional caster out of a Qt signal dispatch terminates the process.
- Teardown becomes a knife's edge. The atexit path calls `RManager::reset()` while Python is alive, but the singleton's static destructor runs after interpreter finalization (`cpp/solvcon/python/module.cpp:81`), and a QAction holding a `py::object` destroyed there decrefs a dead interpreter. The registry would also keep Python feature objects alive invisibly to Python's GC through captured bound methods.

The viable architecture keeps action construction in the language that owns the callback and shares everything else. Python builds `QAction`s with PySide (normal PySide lifetime, exception, and signal semantics, including `toggled(bool)` for checkable items and reverse wiring like `dock.visibilityChanged.connect(action.setChecked)` that needs a live PySide object anyway). C++ keeps `RAction`, extended for shortcuts, checkable state, and groups, with a `std::function<void(bool)>` callback so toggles carry their state. Both hand the finished action to one C++ placement API. Crossing the boundary needs exactly one new line next to the existing casters: `QT_TYPE_CASTER(QAction, _("QAction"))` (step 2).

## Target design

This section defines what steps 1 through 9 build. One class, `RMenuModel`, owned by `RManager` and fully bound to Python, holds both halves: a registry of actions by id and a live model of menus by path.

### A live menu model, built incrementally

Menu structure arrives in phases (C++ setUp, then the Python controller, then each feature, then anything the user scripts from the console), so there is no moment for a one-shot build. The core primitive is an idempotent, incremental `place`:

```cpp
class RMenuModel : public QObject
{
public:
    // Declare a menu node; creates ancestors on demand. The weight orders
    // the node among its siblings (and on the bar for top-level paths).
    QMenu * menu(std::string const & path, int weight = -1);

    // Place an action (or, with an empty action, a separator) in the menu
    // at `path`, ordered by weight among its current siblings.
    void place(std::string const & path, QAction * action, int weight);
    void placeSeparator(std::string const & path, int weight);

    // Identity: every placed action is registered under its objectName.
    QAction * action(std::string const & id) const;
    void remove(std::string const & id);

    QActionGroup * group(std::string const & id);

    void clear(); // called from RManager::reset()
};
```

Ordering is owned by the model, not scanned off widgets: each menu node keeps a sorted list of `(weight, QPointer<QAction>)` entries (separators included, since a separator cannot be queried back out of a `QMenu` by id), and an insert lands before the first sibling of strictly greater weight, so equal weights keep arrival order. Item weights order entries within a menu; the menu nodes themselves carry their own weights, declared once, which is what fixes the bar order and each submenu's position independently of which feature happens to populate first. Step 3 seeds the current bar as one table:

```cpp
// path, weight
{"File", 0}, {"Edit", 10}, {"View", 20}, {"One", 30}, {"Mesh", 40},
{"Canvas", 50}, {"Profiling", 60}, {"Window", 70},
{"View/Panels", 0}, {"View/Camera", 10}, {"View/Camera move", 20},
```

Weight bands leave room to slot a menu between two others without renumbering. A menu materializes when declared or when its first item lands, so the "populate every menu or Windows may crash" note on today's `setUpMenu` (`cpp/solvcon/pilot/RManager.cpp:291`) is satisfied by construction: the invariant the current code actually relies on is "populated before the bar is realized", and an on-demand menu is never empty at all.

The model owns what it holds: placed C++ actions are parented to the model (fixing today's leak), Python-created actions keep their PySide parent, and `QPointer` entries make removal from either side safe. `RManager::reset()` calls `clear()` so the singleton survives repeated setUp cycles in tests.

### Scriptable from Python, end to end

Per the project decision, the model is bound as live, editable data, not a build-time input. From the embedded console (or a test):

```python
mgr = solvcon.pilot.RManager.instance
mgr.menu_model.menu("Mesh/Debug", weight=90)      # create a submenu
act = mgr.menu_model.action("mesh.sample_dialog")  # look up by id
mgr.menu_model.remove("canvas.sample.parabola")    # take an item out
mgr.menu_model.place("Canvas", act2, weight=15)    # put one in
```

`menu(path)` returns the `QMenu` through the existing caster, so anything the model does not cover directly is still reachable as an ordinary PySide object. The eight per-menu getters become unnecessary; during the migration (steps 3 through 8) they survive as thin adapters over `menu(path)` and step 9 deletes them.

### One Python helper over the shared placement

Step 5 gives `PilotFeature` the single item builder the three drifted helpers wanted to be, implemented in PySide and finished by the shared placement:

```python
def add_action(self, path, text, tip, func, *, id, weight=50,
               checkable=False, checked=False, shortcut=None,
               menu_role=None):
    act = QtGui.QAction(text, self._mainWindow)
    act.setObjectName(id)
    act.setStatusTip(tip)
    ...
    self._mgr.menu_model.place(path, act, weight)
    return act
```

It returns the live action so toggle features can keep their reverse wiring. `_add_menu_item`, the controller's `_addAction`, and the two hand-rolled toggles all converge on it (steps 5 through 7). The `menu_role` knob replaces the darwin special case for "Exit" with an explicit `QuitRole`.

### Shortcuts: real ones only

Undo, Redo, and Quit keep (or gain) `QKeySequence::StandardKey` bindings. The camera-movement items stop claiming keys (step 4): the "(W/UP)" hints are deleted because the keys belong to `RDomainWidget::keyPressEvent`, which does something different with them. Attaching menu shortcuts to those keys is explicitly rejected: the shortcut map would consume the keys before `keyPressEvent` and silently swap `moveForward(0.1)` for `panCamera(40)`, and `Ctrl+W` is Cmd+W on macOS, the platform close-window gesture. Reconciling the two camera-control paths into one is real work on camera semantics, out of scope here and left to a follow-up plan.

The Escape reset is fixed for real (step 4): the action keeps `Qt::WidgetShortcut` context and `add3DWidget` attaches it to each new viewer with `QWidget::addAction` (skipping the hidden RHI primer), which is the missing association that makes a widget-context shortcut live.

### The camera menu as data

Step 4 reduces the eleven camera-movement actions to a table walked in a loop:

```cpp
struct CameraMove { std::string id; QString text;
                    std::function<void(RDomainWidget *)> act; };
```

One loop registers each row and places it under "View/Camera move", replacing eleven copied blocks and `createCameraMovementItemHandler` with one list plus one registration call. The camera-mode radio set keeps its `QActionGroup`, now owned by the model under a group id and queryable from Python.

### The Canvas menu and the draw tools

Per the project decision, the Canvas menu is in scope; step 8 makes two changes:

- **The draw tools become a `QActionGroup`.** One checkable action per registered tool (`draw_tool_names()` already provides the list and order), exclusive, with the group's `triggered` calling `mgr.setDrawTool`. The group is the single source of truth: the Canvas menu shows the actions as radio items under "Canvas/Draw tool", and the Painter toolbox buttons become `QToolButton::setDefaultAction` views of the same actions, so checking one surface updates every other and `present()` stops hand-synchronizing button state. `RManager::setDrawTool` stays the programmatic entry point and checks the matching action, so console scripting keeps working.
- **The sample items group under a path.** The seven geometry samples move to "Canvas/Samples", leaving the working items (blank canvas, open-2D) at the top level with the draw-tool group, mirroring what upstream PR 986 (solvcon) did for Mesh.

The `SampleMeshDialog` entry tuples stay as they are in this plan; folding dialog entries into the action registry (register once, surface in a menu or a dialog) is a natural follow-up once the registry exists, and is noted as such rather than bundled here.

## Testing

Steps 1, 2, 4, 7, and 8 carry their tests in the same pull request. Tests are written in Python (the project's preferred layer) under `tests/`, run with `make run_pilot_pytest`, skipping on CI runners without a GUI the same way `test_pilot_mesh_info.py` does today. The registry and model are QWidget-bound, so a Qt-linked gtest target would be new CI plumbing for code Python reaches directly; not worth it.

- Model behavior (steps 1 and 2): `menu(path)` creates ancestors on demand and is idempotent; weights order items and menus regardless of arrival order; equal weights keep arrival order; `action(id)` round-trips; `remove` works from both languages; `clear()` plus repeated `setUp()` on the singleton stays sound.
- Bar structure (step 7): after a full `launch`-equivalent setup, assert the top-level menus and their order, that a known feature's item sits under the intended path, and that "View/Panels" holds the two dock toggles.
- Draw tools (step 8): checking a menu radio item drives `mgr.drawTool` and the toolbox button state, and `setDrawTool` from the console checks the menu.
- The Escape reset (step 4): the action is associated with each new viewer (`associatedObjects` includes the widget).
- macOS: steps 3, 7, and 8 must be exercised on a macOS machine before merging, since the native menu bar relocates role-tagged actions and has a history of dynamic-insertion quirks.

## Risks and alternatives

- **Behavior drift during the port.** Step 3 reproduces the current bar from the seeded table with the getters still in place, so the diff is behavior-preserving before any consumer migrates.
- **A registry can over-engineer a small menu.** The scope is one class with a handful of methods, not a plugin framework. The counter-evidence for "too small to matter" is already in-tree: the getter pattern shipped a missing binding, five item-builder conventions exist, and upstream PR 986 (solvcon) had to invent a private spec format for one dialog.
- **Native menu bars are the riskiest surface.** Incremental insertion into a live macOS bar works in Qt 6 but has a bug history; the mitigation is the explicit macOS verification item and keeping `menuRole` explicit.
- **Alternative: unify only the Python builders.** Rejected as the end state because it cannot deliver the scriptable model, paths, or ordering, but step 5's helper is deliberately shaped so that even if later steps stalled, the codebase is still better off.

## Appendix: resolved and open questions

- **Scriptability (resolved).** The model is exposed to Python as live, editable data; a console user can reorder, remove, and add items at runtime. This is a project decision and shapes the API above.
- **Canvas menu (resolved).** In scope; step 8 covers the draw-tool group and the Samples path.
- **Action ids (recommended).** Flat dotted ids ("mesh.sample_dialog", "camera.reset") independent of menu paths, so moving an item never changes its identity. Ids double as `objectName` for tests.
- **Open: dialog entries.** Should `SampleMeshDialog`'s `(category, label, tip, func)` tuples become registered actions surfaced in a dialog view? Natural follow-up after step 6; not bundled here.
- **Open: camera key semantics.** Unifying `keyPressEvent` movement with the menu movement actions (one handler, real shortcuts) is a camera-behavior change deserving its own plan.
- **Open: menu layout as a JSON document.** The architecture allows it by construction: the placement layer is pure data (menu nodes are path and weight, items are path, action id, and weight), so a JSON document can express the whole bar layout while callbacks stay in code and are referenced by action id, the same split KDE XMLGUI draws with its `.rc` files. The cheap route is a small Python loader that replays the document onto `menu_model` (it falls out of steps 1 and 2 with no C++ change); a `QJsonDocument`-loaded resource could later replace step 3's seeded table. JSON has in-tree precedent (the toggle system round-trips it); YAML would add a new C++ dependency and is not worth it. Either route needs a pending-placement extension: the model remembers unresolved `(path, id, weight)` rows until the action registers, and warns about ids that never resolve. Not bundled here.

yungyuc:
Turn the path-addressed menu tree into something that actually carries Qt actions: a menu's entries become one weight-ordered list, actions gain stable identities, and named groups are supported.

## What changed

- **Unified, weight-ordered entries.** A menu's submenus and items now live in a single weight-ordered list, so the two interleave instead of being kept in separate buckets. `place()` drops a `QAction` at a path and `placeSeparator()` drops a separator at a path.
- **Action identity.** A placed action registers under its `objectName`, so `action(id)` looks one up, `remove(id)` takes it out, and `clear()` also drops the id registry.
- **Named groups.** `group(id)` returns a named `QActionGroup`, created on first use; `clear()` drops the groups as well.
- **Type casters across the boundary.** Added `QAction` and `QActionGroup` type casters so PySide actions and groups pass through the pybind11 boundary. `action()` returns `None` for an unknown id rather than dereferencing a null pointer through the caster.
- **Python bindings.** Bound `place`, `place_separator`, `action`, `remove`, and `group` on the menu model.

For step 2 of issue #989.

### DIFF
--- a/cpp/solvcon/pilot/RMenuModel.hpp
+++ b/cpp/solvcon/pilot/RMenuModel.hpp
@@ -7,13 +7,14 @@
 
 /**
  * @file
- * Live model of the pilot menu bar: menus addressed by path and weight.
+ * Live model of the pilot menu bar: menus by path, actions by id and weight.
  *
  * @ingroup group_domain
  */
 
 #include <solvcon/pilot/common_detail.hpp> // Must be the first include.
 
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -23,6 +24,7 @@
 #include <QString>
 
 class QAction;
+class QActionGroup;
 class QMainWindow;
 class QMenu;
 class QMenuBar;
@@ -36,12 +38,18 @@ namespace solvcon
  * @ingroup group_domain
  *
  * Menus are addressed by a slash-separated path ("View/Panels"). menu()
- * resolves a path, creating any missing ancestor on demand, and orders each
- * node among its siblings (and on the bar for a top-level path) by a weight
- * declared once. A smaller weight sits earlier; equal weights keep arrival
- * order; a negative weight leaves an existing node untouched and appends a new
- * one. The model owns the menus it creates, so clear() empties the bar and
- * lets a fresh setUp rebuild it.
+ * resolves a path, creating any missing ancestor on demand, and place() drops
+ * an action (or a separator) into the menu at a path. Every entry in a menu,
+ * whether a submenu or an item, is ordered by a weight: a smaller weight sits
+ * earlier, equal weights keep arrival order, and a negative weight appends.
+ * The weight-with-gaps ordering follows the convention plugin systems use for
+ * contributed items (for example Drupal's #weight): a declared integer key,
+ * with room left between values to slot an entry in without renumbering.
+ *
+ * Placed actions are registered under their objectName, so action(id) looks
+ * one up and remove(id) takes it out. group(id) returns a named QActionGroup
+ * for exclusive selections. The model owns the menus and groups it creates, so
+ * clear() empties the bar and lets a fresh setUp rebuild it.
  */
 class RMenuModel
     : public QObject
@@ -59,10 +67,37 @@ public:
     /// leaves an existing node's order untouched and appends a new one.
     QMenu * menu(std::string const & path, int weight = -1);
 
-    /// Delete every menu the model created and empty the bar.
+    /// Place @p action in the menu at @p path, ordered by @p weight among the
+    /// menu's current entries. The action is registered under its objectName
+    /// when that name is set.
+    void place(std::string const & path, QAction * action, int weight = 50);
+
+    /// Place a separator in the menu at @p path, ordered by @p weight.
+    void placeSeparator(std::string const & path, int weight = 50);
+
+    /// The action registered under @p id, or nullptr when none is.
+    QAction * action(std::string const & id) const;
+
+    /// Remove and unregister the action registered under @p id, if any.
+    void remove(std::string const & id);
+
+    /// The named action group, created on first use.
+    QActionGroup * group(std::string const & id);
+
+    /// Delete every menu and group the model created and empty the bar.
     void clear();
 
 private:
+
+    struct Node;
+
+    struct Entry
+    {
+        int weight = -1;
+        QPointer<QAction> action; // the entry's QAction (item, separator, or
+                                  // a submenu's menuAction)
+        std::unique_ptr<Node> submenu; // non-null only for a submenu entry
+    };
 
     struct Node
     {
@@ -70,19 +105,25 @@ private:
         int weight = -1;
         QPointer<QMenu> menu; // nullptr for the root, which stands for the bar
         Node * parent = nullptr;
-        std::vector<std::unique_ptr<Node>> children;
+        std::vector<Entry> entries;
     };
 
+    Node * resolve(std::string const & path, int weight);
     Node * ensureChild(Node * parent, QString const & name, int weight);
-    void reweight(Node * parent, Node * node, int weight);
-    int insertionIndex(Node * parent, int weight) const;
-    void containerInsert(Node * parent, QMenu * menu, QAction * before);
-    void containerRemove(Node * parent, QMenu * menu);
+    void placeAction(Node * node, QAction * action, int weight);
+    int insertionIndex(Node * node, int weight) const;
+    void reweightEntry(Node * node, int index, int weight);
+    QAction * beforeAt(Node * node, int index) const;
+    void containerInsert(Node * node, QAction * action, QAction * before);
+    void containerRemove(Node * node, QAction * action);
+    bool removeAction(Node * node, QAction * target);
 
     QMainWindow * m_mainWindow = nullptr;
     QMenuBar * m_bar = nullptr;
     Node m_root;
     std::vector<QPointer<QMenu>> m_menus;
+    std::map<std::string, QPointer<QAction>> m_actions;
+    std::map<std::string, QPointer<QActionGroup>> m_groups;
 }; /* end class RMenuModel */
 
 } /* end namespace solvcon */
--- a/tests/test_pilot_menu_model.py
+++ b/tests/test_pilot_menu_model.py
@@ -9,6 +9,7 @@ import solvcon
 
 try:
     from solvcon import pilot
+    from PySide6 import QtGui
 except ImportError:
     pilot = None
 
@@ -86,6 +87,73 @@ class MenuModelTC(unittest.TestCase):
         self.assertIn("Temp", self._top_level())
         self.model.clear()
         self.assertNotIn("Temp", self._top_level())
+
+
+@unittest.skipIf(GITHUB_ACTIONS or not solvcon.HAS_PILOT,
+                 "GUI is not available in GitHub Actions")
+class MenuPlacementTC(unittest.TestCase):
+    def setUp(self):
+        self.mgr = pilot.RManager.instance.setUp()
+        self.model = self.mgr.menu_model
+        self.model.clear()
+
+    def _act(self, text, oid):
+        act = QtGui.QAction(text, self.mgr.mainWindow)
+        act.setObjectName(oid)
+        return act
+
+    def _items(self, path):
+        return [a.text() for a in self.model.menu(path).actions()]
+
+    def test_place_orders_items_by_weight(self):
+        self.model.place("Box", self._act("Later", "b.later"), weight=40)
+        self.model.place("Box", self._act("Early", "b.early"), weight=10)
+        self.model.place("Box", self._act("Middle", "b.middle"), weight=25)
+        self.assertEqual(self._items("Box"), ["Early", "Middle", "Later"])
+
+    def test_place_equal_weight_keeps_arrival_order(self):
+        self.model.place("Box", self._act("First", "b.1"), weight=10)
+        self.model.place("Box", self._act("Second", "b.2"), weight=10)
+        self.model.place("Box", self._act("Third", "b.3"), weight=10)
+        self.assertEqual(self._items("Box"), ["First", "Second", "Third"])
+
+    def test_items_and_submenu_interleave_by_weight(self):
+        self.model.place("Box", self._act("Top", "b.top"), weight=5)
+        self.model.menu("Box/Sub", weight=10)
+        self.model.place("Box", self._act("Bottom", "b.bottom"), weight=20)
+        self.assertEqual(self._items("Box"), ["Top", "Sub", "Bottom"])
+
+    def test_action_id_round_trip(self):
+        act = self._act("Named", "box.named")
+        self.model.place("Box", act, weight=10)
+        self.assertEqual(self.model.action("box.named"), act)
+        self.assertIsNone(self.model.action("box.missing"))
+
+    def test_remove_takes_item_out(self):
+        self.model.place("Box", self._act("Gone", "box.gone"), weight=10)
+        self.assertIn("Gone", self._items("Box"))
+        self.model.remove("box.gone")
+        self.assertIsNone(self.model.action("box.gone"))
+        self.assertNotIn("Gone", self._items("Box"))
+
+    def test_remove_finds_item_in_a_submenu(self):
+        self.model.place("Box/Sub", self._act("Deep", "box.deep"), weight=10)
+        self.assertIn("Deep", self._items("Box/Sub"))
+        self.model.remove("box.deep")
+        self.assertNotIn("Deep", self._items("Box/Sub"))
+
+    def test_place_separator(self):
+        self.model.place("Box", self._act("A", "box.a"), weight=10)
+        self.model.place_separator("Box", weight=15)
+        self.model.place("Box", self._act("B", "box.b"), weight=20)
+        actions = self.model.menu("Box").actions()
+        self.assertTrue(actions[1].isSeparator())
+        self.assertEqual([actions[0].text(), actions[2].text()], ["A", "B"])
+
+    def test_group_is_created_once(self):
+        first = self.model.group("modes")
+        second = self.model.group("modes")
+        self.assertEqual(first, second)
 
 
 # vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/pilot/RMenuModel.cpp
+++ b/cpp/solvcon/pilot/RMenuModel.cpp
@@ -8,6 +8,7 @@
 #include <limits>
 
 #include <QAction>
+#include <QActionGroup>
 #include <QMainWindow>
 #include <QMenu>
 #include <QMenuBar>
@@ -20,7 +21,7 @@ namespace solvcon
 namespace
 {
 
-/// A negative weight sorts last, so a node declared without a weight lands
+/// A negative weight sorts last, so an entry declared without a weight lands
 /// after every explicitly weighted sibling.
 int effective_weight(int weight)
 {
@@ -43,6 +44,97 @@ RMenuModel::~RMenuModel()
 
 QMenu * RMenuModel::menu(std::string const & path, int weight)
 {
+    Node * node = resolve(path, weight);
+    return node ? node->menu.data() : nullptr;
+}
+
+void RMenuModel::place(std::string const & path, QAction * action, int weight)
+{
+    Node * node = resolve(path, -1);
+    if (node == nullptr || action == nullptr)
+    {
+        return;
+    }
+    placeAction(node, action, weight);
+
+    QString const id = action->objectName();
+    if (!id.isEmpty())
+    {
+        m_actions[id.toStdString()] = action;
+    }
+}
+
+void RMenuModel::placeSeparator(std::string const & path, int weight)
+{
+    Node * node = resolve(path, -1);
+    if (node == nullptr)
+    {
+        return;
+    }
+    // Parent the separator to its menu so it is deleted with the menu; a
+    // separator carries no id and stays out of the registry.
+    auto * separator = new QAction(node->menu);
+    separator->setSeparator(true);
+    placeAction(node, separator, weight);
+}
+
+QAction * RMenuModel::action(std::string const & id) const
+{
+    auto const it = m_actions.find(id);
+    return it != m_actions.end() ? it->second.data() : nullptr;
+}
+
+void RMenuModel::remove(std::string const & id)
+{
+    auto const it = m_actions.find(id);
+    if (it == m_actions.end())
+    {
+        return;
+    }
+    if (QAction * target = it->second.data())
+    {
+        removeAction(&m_root, target);
+    }
+    m_actions.erase(it);
+}
+
+QActionGroup * RMenuModel::group(std::string const & id)
+{
+    QPointer<QActionGroup> & slot = m_groups[id];
+    if (!slot)
+    {
+        slot = new QActionGroup(this);
+    }
+    return slot.data();
+}
+
+void RMenuModel::clear()
+{
+    for (auto const & entry : m_groups)
+    {
+        if (entry.second)
+        {
+            delete entry.second;
+        }
+    }
+    m_groups.clear();
+
+    // QPointer entries drop to null when Qt already deleted a menu (for
+    // example when the whole main window is torn down), so guard each delete.
+    for (QPointer<QMenu> const & menu : m_menus)
+    {
+        if (menu)
+        {
+            delete menu;
+        }
+    }
+    m_menus.clear();
+    m_actions.clear();
+    m_root.entries.clear();
+}
+
+RMenuModel::Node * RMenuModel::resolve(std::string const & path, int weight)
+{
     QStringList const parts =
         QString::fromStdString(path).split('/', Qt::SkipEmptyParts);
     if (parts.isEmpty())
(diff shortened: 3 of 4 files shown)


cpp/solvcon/pilot/RMenuModel.hpp
```diff
@@ -36,12 +38,18 @@ namespace solvcon
...
- * declared once. A smaller weight sits earlier; equal weights keep arrival
- * order; a negative weight leaves an existing node untouched and appends a new
- * one. The model owns the menus it creates, so clear() empties the bar and
- * lets a fresh setUp rebuild it.
+ * resolves a path, creating any missing ancestor on demand, and place() drops
+ * an action (or a separator) into the menu at a path. Every entry in a menu,
+ * whether a submenu or an item, is ordered by a weight: a smaller weight sits
+ * earlier, equal weights keep arrival order, and a negative weight appends.
+ * The weight-with-gaps ordering follows the convention plugin systems use for
+ * contributed items (for example Drupal's #weight): a declared integer key,
```
yungyuc, line 46: Use weight to order the menu items.

tests/test_pilot_menu_model.py
```diff
@@ -88,4 +89,71 @@ def test_clear_removes_model_menus(self):
         self.assertNotIn("Temp", self._top_level())
 
 
+@unittest.skipIf(GITHUB_ACTIONS or not solvcon.HAS_PILOT,
+                 "GUI is not available in GitHub Actions")
+class MenuPlacementTC(unittest.TestCase):
```
yungyuc, line 94: Test for placement ordering.

yungyuc:
@tigercosmos @j8xixo12 please take a look at the new menu system code.

PR #1018, opened 2026-07-05
Make REPL and curate namespace (steps 1, 2 of console)

### LINKED ISSUE #1005: Improve console behaviors
Keep the in-process interpreter and the zero-copy access to live mesh and solver objects underneath the console.  Make the behaviors more predictable.

Change to the look and feel will be future work: issue #1023.

## Goal

A scientist running the pilot should be able to drop into the console and drive the whole application from Python: load a mesh, push it to a viewer, inspect fields, and script repetitive actions, with the editing comforts of a modern REPL. Today the console runs commands but stops short of that. The work below turns it into a first-class control and analysis surface.

## Planned steps

The work splits into six steps. Each is scoped to roughly a 300-line diff, implementation plus tests, so it reviews and lands on its own while keeping the tree building and the suite green. The line estimate on each step is a target for how much it should change, not a budget to fill.

1. **Step 1: Persistent interpreter and correct errors.** Replace the per-submission `exec()` with a per-environment `code.InteractiveConsole` (`solvcon/apputil.py`, `solvcon/system.py`), so `push(line)` drives the `>>> ` and `... ` prompts, `"single"`-mode compilation auto-displays bare expression results and binds `_`, and exceptions surface through the interpreter's own `showtraceback`/`showsyntaxerror` instead of the host `traceback.print_stack()`. Guard `SystemExit` so `exit()` does not tear down the pilot. The widget (`RPythonConsoleDockWidget.cpp`) gains the two-prompt display. This is the highest-leverage step and unblocks the rest. Estimated diff: ~300 lines (backend plus prompt handling and tests). PR #1018.
2. **Step 2: Curated namespace, banner, and housekeeping.** Seed the execution globals with the handles a scientist reaches for (`mgr`, `mesh`, `viewer`) and a few helpers (`show_mesh`, `meshes`, `viewers`), and print a banner that lists them. Fix the two latent bugs in the same module while here: `get_current_appenv` (which walks the dict only to take the last key) and `stop_code` (which indexes the dict by integer). Mostly `solvcon/apputil.py` and the pilot wiring. Estimated diff: ~280 lines (seeding, banner, fixes, tests).  PR #1018.
3. **Step 3: Persistent, searchable history.** Persist the command history to a file under the solvcon profile directory, load it at startup, and add reverse incremental search (Ctrl-R) over it in the command editor. Replaces the in-memory-only ring in `RPythonConsoleDockWidget.{hpp,cpp}`. Estimated diff: ~320 lines (file store, search mode, tests).  PR #1019.
4. **Step 4: Syntax highlighting and multiline editing.** Add a Python `QSyntaxHighlighter` to the command editor, plus auto-indent and bracket matching for multiline blocks. All in the widget. Estimated diff: ~280 lines (highlighter, editing polish, tests).  PR #1020.
5. **Step 5: Call tips and richer completion.** Keep the introspective `rlcompleter` path, which reflects real runtime state, and add signature and docstring call tips on `(`. Optionally back completion with Jedi where its dependency weight is acceptable. Widget plus a small Python helper. Estimated diff: ~300 lines (tip extraction, popup, tests).  PR #1021.
6. **Step 6: Busy state and a worker-thread helper.** Show a busy state around execution and provide a helper that runs heavy solver work on a worker `QThread` so the GUI stays responsive, since true in-console interrupt is not achievable in-process. Widget plus the thread helper. Estimated diff: ~280 lines (busy state, worker helper, tests).  PR #1021.

Steps 1 and 2 are mostly Python in `solvcon/`, with a small widget change for the prompts; steps 3 through 6 mainly touch the C++ widget in `cpp/solvcon/pilot/`.

### A persistent interpreter

Replace the per-submission `exec()` with a persistent `code.InteractiveConsole` bound to the same `AppEnvironment` namespace, the move ParaView and Blender both made. `push(line)` returns whether more input is needed, which drives the `>>> ` and `... ` prompts and makes multiline blocks and incomplete input work correctly. Compiling in `"single"` mode auto-displays bare expression results and sets `_` to the last value, so the console becomes a real read-eval-print loop. Guard `SystemExit` so that `exit()` or `quit()` typed in the console does not tear down the pilot.

### Correct error reporting

Route exceptions through the interpreter's own `showtraceback`/`showsyntaxerror`, which format the user's traceback with the console's input as the source, and drop the current `traceback.print_stack()` path in `system.py:exec_code`.

### A curated, documented namespace

Seed the execution globals with the handles a scientist actually reaches for, and document them in the console banner:

- `mgr`: the running `RManager` (windows, MDI area, 3D and 2D widgets).
- `mesh`: the current mesh, when one is loaded.
- `viewer`: the current 3D widget (`mgr.currentR3DWidget()`).
- A few helpers, for example `show_mesh(m)` to push a mesh into a fresh viewer, and `meshes()` / `viewers()` to list what is open.

Keep the `AppEnvironment` isolation from `__main__` that exists today.

### Persistent, searchable history

Persist history to a file under the solvcon profile directory, load it at startup, and add reverse search (Ctrl-R) over it. This matches the SQLite history that IPython keeps and removes the daily annoyance of losing recall across sessions.

### Editing comforts

- **Syntax highlighting** in the command editor (Pygments or a Qt highlighter), low risk and high perceived polish.
- **Call tips and richer completion.** Keep the live-object `rlcompleter` path, which reflects real runtime state, and add signature and docstring tips on `(`. Jedi is an option for type-aware, side-effect-free completion if its dependency weight is acceptable.

### Honest behavior under long commands

At a minimum, show a busy state around execution and document that the GUI blocks during a command. Provide a helper to run heavy solver work on a worker `QThread` so the GUI stays responsive, since true in-console interrupt is not achievable in-process.

### Proposed behavior surface

| Area | Today | Target |
| --- | --- | --- |
| Submission | fresh `exec()` per line | persistent `InteractiveConsole.push()` |
| Expression result | silent | auto-displayed, bound to `_` |
| Multiline | Shift+Enter only | `... ` continuation, incomplete detection |
| Errors | host call stack | user traceback |
| `exit()` | risks killing pilot | guarded |
| Namespace | `sc`, `appenv` | `mgr`, `mesh`, `viewer`, helpers |
| History | in-memory | persistent + reverse search |
| Long command | silent freeze | busy state + worker-thread helper |

### Keep the console in-process

The whole design above rests on one choice: the interpreter stays inside the pilot's own process. That is deliberate. The console earns its keep by handing the scientist live references to the `StaticMesh`, solver, `World`, and `RDomainWidget` objects and the running `RManager`, with nothing copied. An out-of-process kernel, the Spyder model, cannot see those objects at all; every value would have to cross a process boundary as serialized data. A Jupyter `qtconsole` kernel run in-process could see them, but it would stack a heavy IPython, ipykernel, tornado, and pyzmq dependency chain on top of the pilot's already unusual pybind11-embedded interpreter, adopt a path that Jupyter itself documents as not well supported in-process, and still inherit the same GUI-freeze and no-interrupt behavior the current console already has. The features that stack would buy (Jedi completion, SQLite history, Pygments highlighting, inline plots) are all reachable step by step on the widget we own, which is exactly what the steps above do.

Image

*The console keeps its in-process pipeline; each stage gains the missing REPL behavior rather than being replaced by an external kernel.*

The one goal that would justify `qtconsole` is making inline matplotlib and rich HTML output first-class. Even then, the only variant that preserves live-object access is an in-process kernel seeded from the existing namespace, so treat it as a possible future front-end, not the near-term path.

## More information

More information is moved to discussion #1024 to keep the contents in this issue concise.

yungyuc:
This turns the pilot's Python console from a one-shot `exec()` of each submission into a real read-eval-print loop, then seeds that loop with the handles a scientist actually reaches for and greets the session with a banner that lists them. The console now behaves like a standard interpreter: it carries state across submissions, auto-displays bare expressions, binds `_`, drives the `>>> ` and `... ` prompts, and survives `exit()`. On top of that it opens with `mgr`, `viewer`, `mesh`, and a few helpers already in scope, with `viewer` and `mesh` kept current as the focus changes. Two latent bookkeeping bugs in the same module are fixed along the way.

The UI does not change.  There are still two text boxes for output and input, respectively.  I will use more console to learn how to improve the UI.

## Step 1: make the pilot console a read-eval-print loop

Replace the per-submission `exec()` with a per-environment `code.InteractiveConsole`, so the console behaves like the standard read-eval-print loop instead of a fresh exec of each command.

`AppEnvironment` now owns one interpreter bound to a single namespace, shared by globals and locals, so names injected by the pilot and names bound by executed code are both visible. `run_code` feeds a command line by line through `push()`, which reports whether more input is needed and drives the `>>> ` and `... ` prompts. Compiling in `single` mode auto-displays a bare expression and binds it to `_`, and exceptions are formatted by the interpreter's own `showtraceback` and `showsyntaxerror` against the console input rather than the host call stack that `traceback.print_stack()` printed. `SystemExit` from `exit()` or `quit()` is caught so it cannot tear down the interpreter the pilot embeds.

The console dock widget echoes a submitted command with the two prompts so the transcript reads like a real interpreter session.

The console test is un-skipped and rebuilt around the backend: the bare expression auto-display and `_` binding, the multiline more-input report, the user traceback, the guarded `SystemExit`, and the `exec_code` entry point the C++ side calls.

## Step 2: curate the console namespace and fix bookkeeping

Seed the console with the handles a scientist reaches for and greet with a banner that lists them, so the live session is discoverable without knowing to walk the pilot module.

`install_pilot_namespace` seeds the namespace with `mgr` (the running manager), `viewer` (the current 3D viewer), `mesh` (its mesh), and the helpers `show_mesh(m)`, `viewers()`, and `meshes()`. `viewer` and `mesh` track the focused viewer through a namespace refresher that runs before each command, so they never go stale at the value captured when the console opened. The builder is duck-typed on the manager, and the banner is written to the transcript when the pilot builds its window. `RManager` gains `list3DWidgets`, the 3D counterpart of `list2DWidgets`, resolved through `domainWidgetOf` because a 3D viewer sits inside a host container.

Fix the two latent bookkeeping bugs in the same module: `get_current_appenv` walked the environment dict only to take the last key, and `stop_code` compared a dict key against the environment object and deleted by integer index, so it never removed anything.

Tests drive the curation with a stand-in manager and cover the seeded handles, `show_mesh` opening a viewer, the refresher tracking the current viewer, and both bookkeeping fixes.

For steps 1 and 2 of issue #1005.

### DIFF
--- a/solvcon/apputil.py
+++ b/solvcon/apputil.py
@@ -14,6 +14,7 @@ Tools to run applications
 # Use flake8 http://flake8.pycqa.org/en/latest/user/error-codes.html
 
 
+import code
 import importlib
 import rlcompleter
 
@@ -32,28 +33,71 @@ __all__ = [
 environ = {}
 
 
+class _ConsoleInterpreter(code.InteractiveConsole):
+    """
+    A :class:`code.InteractiveConsole` bound to an application namespace.
+
+    The console compiles in ``"single"`` mode, so a bare expression is
+    auto-displayed and bound to ``_`` the way the standard read-eval-print
+    loop does. Exceptions are formatted against the user's own input by
+    :meth:`showtraceback` and :meth:`showsyntaxerror`, not the host stack.
+
+    ``exit()`` or ``quit()`` typed in the console raises :exc:`SystemExit`,
+    which would tear down the interpreter that the pilot embeds. Guard it
+    here so the console reports the request without ending the process.
+    """
+    def push(self, line):
+        try:
+            return super().push(line)
+        except SystemExit:
+            self.write("SystemExit: use the window controls to quit.\n")
+            self.resetbuffer()
+            return False
+
+
 class AppEnvironment:
     """
     Collects the environment for an application.
 
+    The read-eval-print loop uses a single namespace, so :ivar:`globals`
+    and :ivar:`locals` are the same dict: names bound by executed code and
+    names injected by the pilot are both visible to the interpreter.
+
     :ivar globals:
-        The global namespace of the application.
+        The namespace of the application.
     :ivar locals:
-        The local namespace of the application.
+        An alias of :ivar:`globals`.
     """
     def __init__(self, name):
-        self.globals = {
+        namespace = {
             # Give the application an alias of the top package.
             'sc': importlib.import_module('solvcon'),
             'appenv': self,
         }
-        self.locals = {}
+        self.globals = namespace
+        self.locals = namespace
         self.name = name
+        self.console = _ConsoleInterpreter(namespace)
         # Each run of the application appends a new environment.
         environ[name] = self
 
-    def run_code(self, code):
-        exec(code, self.globals, self.locals)
+    def run_code(self, source):
+        """
+        Feed a possibly multi-line command to the persistent interpreter.
+
+        Each line drives ``push()``, which returns whether more input is
+        needed to complete the statement. A block left open after the last
+        line is closed with a blank line, the way a blank line ends a
+        compound statement in the interactive interpreter.
+
+        :return: True when the interpreter is still waiting for more input.
+        """
+        more = False
+        for line in source.split('\n'):
+            more = self.console.push(line)
+        if more:
+            more = self.console.push('')
+        return more
 
 
 def get_appenv(name=None):
@@ -100,9 +144,9 @@ def get_completions(text):
     return completions
 
 
-def run_code(code):
+def run_code(source):
     aenv = get_current_appenv()
-    aenv.run_code(code)
+    return aenv.run_code(source)
 
 
 def stop_code(appenvobj=None):
--- a/solvcon/apputil.py
+++ b/solvcon/apputil.py
@@ -26,6 +26,9 @@ __all__ = [
     'get_completions',
     'run_code',
     'stop_code',
+    'build_pilot_namespace',
+    'format_banner',
+    'install_pilot_namespace',
 ]
 
 
@@ -77,9 +80,23 @@ class AppEnvironment:
         self.globals = namespace
         self.locals = namespace
         self.name = name
+        self.namespace_refreshers = []
         self.console = _ConsoleInterpreter(namespace)
         # Each run of the application appends a new environment.
         environ[name] = self
+
+    def seed(self, **handles):
+        """Install curated handles into the interpreter namespace."""
+        self.globals.update(handles)
+
+    def add_namespace_refresher(self, refresh):
+        """
+        Register a callback that refreshes the namespace before each command.
+
+        The callback receives the namespace dict. Use it for handles whose
+        value tracks live session state, such as the current viewer.
+        """
+        self.namespace_refreshers.append(refresh)
 
     def run_code(self, source):
         """
@@ -92,6 +109,8 @@ class AppEnvironment:
 
         :return: True when the interpreter is still waiting for more input.
         """
+        for refresh in self.namespace_refreshers:
+            refresh(self.globals)
         more = False
         for line in source.split('\n'):
             more = self.console.push(line)
@@ -118,13 +137,9 @@ get_appenv(name='master')
 
 
 def get_current_appenv():
-    has_key = False
-    for k in reversed(environ):
-        has_key = True
-        break
-    if not has_key:
+    if not environ:
         raise KeyError("No AppEnviron is available")
-    return environ[k]
+    return environ[next(reversed(environ))]
 
 
 def get_completions(text):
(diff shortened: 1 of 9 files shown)


solvcon/apputil.py
```diff
@@ -25,35 +26,97 @@
...
+    'format_banner',
+    'install_pilot_namespace',
 ]
 
 
 # All environment objects of this process.
 environ = {}
 
 
+class _ConsoleInterpreter(code.InteractiveConsole):
```
yungyuc, line 39: Use `InteractiveConsole`.

solvcon/apputil.py
```diff
@@ -25,35 +26,97 @@
...
+    def add_namespace_refresher(self, refresh):
+        """
+        Register a callback that refreshes the namespace before each command.
+
+        The callback receives the namespace dict. Use it for handles whose
+        value tracks live session state, such as the current viewer.
+        """
+        self.namespace_refreshers.append(refresh)
+
+    def run_code(self, source):
```
yungyuc, line 101: `run_code` now uses `InteractiveConsole`, not `exec`.

solvcon/apputil.py
```diff
@@ -100,18 +159,93 @@ def get_completions(text):
...
-        indices = [i for i, o in enumerate(environ) if o == appenvobj]
-        indices = reversed(indices)
-        for i in indices:
-            del environ[i]
+        names = [name for name, env in environ.items() if env is appenvobj]
+        for name in names:
+            del environ[name]
+
+
+def build_pilot_namespace(mgr):
```
yungyuc, line 176: Explicit namespace.

Issue #1023, opened 2026-07-05
Add terminal-like I/O to console


yungyuc:
The Python console currently uses two text areas for output and input. It can be kept, but add a new terminal style UI for the console.

This issue is to follow up issue #1005.  See discussion #1123 for additional research materials.

## Goal

A scientist in the pilot should be able to open a console that reads the way a terminal reads. The prompt, their command, and its result should sit together in one column, in chronological order, so a session scrolls back as a single legible record. Typing should feel like a terminal: the cursor lives on the current input line after the prompt, the already-committed transcript above is not editable, and Enter runs the current line while an incomplete block keeps its continuation prompt.

The existing two-pane console stays available and unchanged. The terminal is a new dock the user opens when they want it. Everything the console already does must work in the terminal too: recall with Up and Down, reverse search with Ctrl-R, Tab completion over live objects, call tips on `(`, syntax highlighting as you type, the persistent history file, and the busy cursor with the worker-thread helper for long commands. The terminal shares those pieces with the existing console instead of copying them.

## Planned steps

The work splits into four steps, each scoped to roughly a 300-line diff, implementation plus tests, so each reviews and lands on its own while the tree keeps building and the suite stays green. The line estimate on each step is a target for how much it should change, not a budget to fill. The new widget lives in `cpp/solvcon/pilot/RPythonTerminalDockWidget.{hpp,cpp}`; the existing `RPythonConsoleDockWidget` is left alone except for the menu wiring that opens the terminal. Steps 2 and 4 also touch the Python entry points in `solvcon/system.py` and `solvcon/apputil.py`, which both consoles then share.

[...]

PR #1026, opened 2026-07-05
Guard the console against blocking stdin reads

yungyuc:
When running `help()` without argument in the new console that uses `InteractiveConsole`, it hangs the whole pilot.

A bare `help()` in the pilot console starts pydoc's interactive browser, which loops reading `sys.stdin`. The console runs commands on the GUI thread with no interactive input, so the read never returns and the window freezes. `input()` and any other stdin read hang the same way.

Present an exhausted `sys.stdin` while a console command runs, so such a read hits EOF at once and the command returns. The guard lives in the shared `run_code` driver, so it covers the pilot and any other embedding.

It will need improvement after implementing a terminal-based UI that has both stdin and stdout (issue #1023).

### DIFF
--- a/solvcon/apputil.py
+++ b/solvcon/apputil.py
@@ -16,10 +16,13 @@ Tools to run applications
 
 import code
 import concurrent.futures
+import contextlib
 import importlib
 import inspect
+import io
 import re
 import rlcompleter
+import sys
 
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


solvcon/apputil.py
```diff
@@ -74,6 +77,25 @@ def push(self, line):
             return False
 
 
+@contextlib.contextmanager
+def _stdin_at_eof():
```
yungyuc, line 81: This context manager provide a fake EOF stdin.

tests/test_apputil.py
```diff
@@ -44,4 +47,48 @@ def _check(i, basenum):
...
+
+    def _run_with_timeout(self, source, timeout=10):
+        env = solvcon.apputil.AppEnvironment('stdin-test')
+        done = threading.Event()
+
+        # Run on a daemon thread so a regression that hangs fails the
+        # assertion instead of stalling the whole test process.
+        def _run():
+            with contextlib.redirect_stdout(io.StringIO()):
+                env.run_code(source)
```
yungyuc, line 70: Run the code in a thread to detect regression.

PR #1031, opened 2026-07-06
Split the profiler out of the toggle module

### LINKED ISSUE #1022: Unify the toggle system into one store

### What is entangled

### Planned steps

PR #1031.

### Testing

### Goal

### Planned steps

## Study and report

yungyuc:
The toggle directory mixed the toggles with the call profiler that had no home of its own. Give the profiler its own `cpp/solvcon/profiling/` module so the toggle module holds only toggles. This is a pure relocation: every class name, the `SOLVCON_PROFILE_*` macros, and the pybind names stay unchanged.

Move `profile.hpp`, `RadixTree.hpp`, `RadixTree.cpp`, `SerializableProfiler.hpp`, and pymod/wrap_profile.cpp into the new module, and repoint every include from `solvcon/toggle/...` to `solvcon/profiling/...` in the buffer headers, the toggle header, and the profiler gtests. The new module registers its wrapper through `initialize_profiling`, called from the main pybind module right after the toggle wrapper, so the exposed Python surface is identical.

Track the move in the build wiring: a new `profiling/CMakeLists.txt` owns the profiler source lists, the top-level list picks up `SOLVCON_PROFILING_FILES`, the gtest binary picks up `SOLVCON_PROFILING_SOURCES` for `RadixTree.cpp`, and the standalone_buffer `Makefile` copies `solvcon/profiling` instead of `solvcon/toggle` for the `RadixTree` unit the buffer depends on.

For section 1 of issue #1022.


cpp/solvcon/profiling/pymod/profiling_pymod.cpp
yungyuc, line 1: Move from `toggle/` to the new directory `profiling/`.

PR #1037, opened 2026-07-08
Implement ordering comparison operators for SimpleArray

### LINKED ISSUE #1036: Design comparison operators for `SimpleArray`
Design and implement all [comparison operators](https://en.cppreference.com/cpp/language/operator_comparison) for `SimpleArray` in C++ and Python.  Implement the three-way comparison `operator <=>` in C++ but not Python.

Design the API to be consistent between C++ and Python, and between `SimpleArray` and numpy ndarray.

This issue follows up issue #1034.

tigercosmos:
Implement the element-wise ordering comparisons (the lt/le/gt/ge methods and the <, <=, >, >= operators returning a boolean array) for `SimpleArray`, alongside the existing eq and ne. The shared element-wise comparison loop is factored into a single compare_with template, and ordering is left unbound for complex arrays so Python raises TypeError like numpy.

Following the review, the C++-only three-way operator<=> was dropped: whole-array three-way comparison returns a single std::partial_ordering rather than an element-wise array, so it is inconsistent with the other comparison operators.

Related to #1036.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -741,11 +741,34 @@ public:
         return A(*static_cast<A const *>(this)).idiv(scalar);
     }
 
+private:
+
+    // Element-wise comparison kernel shared by eq/ne/lt/le/gt/ge. The array
+    // overload requires matching shapes; both produce a bool per element.
+    template <typename Cmp>
+    SimpleArray<bool> compare_with(A const & other, Cmp cmp, char const * op) const;
+    template <typename Cmp>
+    SimpleArray<bool> compare_with(value_type scalar, Cmp cmp) const;
+
+public:
+
     SimpleArray<bool> eq(A const & other) const;
     SimpleArray<bool> eq(value_type scalar) const;
 
     SimpleArray<bool> ne(A const & other) const;
     SimpleArray<bool> ne(value_type scalar) const;
+
+    SimpleArray<bool> lt(A const & other) const;
+    SimpleArray<bool> lt(value_type scalar) const;
+
+    SimpleArray<bool> le(A const & other) const;
+    SimpleArray<bool> le(value_type scalar) const;
+
+    SimpleArray<bool> gt(A const & other) const;
+    SimpleArray<bool> gt(value_type scalar) const;
+
+    SimpleArray<bool> ge(A const & other) const;
+    SimpleArray<bool> ge(value_type scalar) const;
 
     A & iadd(A const & other)
     {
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2286,54 +2321,116 @@ SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(value_type scala
...
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(A const & other) const
+{
+    return compare_with(other, "eq", std::equal_to<>{});
+}
+
+template <typename A, typename T>
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(value_type scalar) const
+{
+    return compare_with(scalar, std::equal_to<>{});
+}
```
tigercosmos, line 2334: Extract to common functions.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2286,54 +2321,116 @@ SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(value_type scala
...
+
+template <typename A, typename T>
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::ge(value_type scalar) const
+{
+    return compare_with(scalar, std::greater_equal<>{});
+}
+
+template <typename A, typename T>
+std::partial_ordering
+detail::SimpleArrayMixinCalculators<A, T>::operator<=>(A const & other) const
```
tigercosmos: Implementation for `<=>`.

tigercosmos (commented):

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -741,12 +742,49 @@ class SimpleArrayMixinCalculators
         return A(*static_cast<A const *>(this)).idiv(scalar);
     }
 
+private:
+
+    // Element-wise comparison kernel shared by eq/ne/lt/le/gt/ge. The array
+    // overload requires matching shapes; both produce a bool per element.
+    template <typename Cmp>
+    SimpleArray<bool> compare_with(A const & other, char const * op, Cmp cmp) const;
```
yungyuc: The argument order is not intuitive.  For scalar it's

```cpp
SimpleArray<bool> compare_with(value_type scalar, Cmp cmp) const
```

But for array `op` is inserted between `other` and `cmp:

```cpp
SimpleArray<bool> compare_with(A const & other, char const * op, Cmp cmp) const
```

Since the scalar does not need `op`, an intuitive order for array is:

```cpp
SimpleArray<bool> compare_with(A const & other, Cmp cmp, char const * op) const
```

yungyuc:
```suggestion
    SimpleArray<bool> compare_with(A const & other, Cmp cmp, char const * op) const;
```

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2286,54 +2329,116 @@ SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::eq(value_type scala
...
+}
+
+template <typename A, typename T>
+SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::ge(value_type scalar) const
+{
+    return compare_with(scalar, std::greater_equal<>{});
+}
+
+template <typename A, typename T>
+std::partial_ordering
```
yungyuc: The return type of three-way comparison does not make sense for array.  Array comparison should return an array, but `std::partial_order` is not an array.

We should not implement the three-way comparison for array.

yungyuc (changes requested):
- [x] We should not implement the three-way comparison for array because the return type is not an array.
- [x] Reorder array `compare_with` arguments.

tigercosmos:
@yungyuc  All fixed, please take a look. Thanks.

yungyuc (approved):
LGTM

PR #1038, opened 2026-07-08
Merge the mesh and entity tree docks into one Inspector panel

### LINKED ISSUE #1029: `EntityTree` to include `MeshInfoTree`
`MeshInfo`/`MeshInfoTree` should be merged into `EntityTree` and use the latter's framework.

`MeshInfo`/`MeshInfoTree` implements the entity tree as well, but for 2D/3D mesh rendering.  `EntityTree` is for the drawn shapes.  The two kinds of entities can share the same widget, but occupy different sub-trees.

This is to follow up PR #975.

j8xixo12:
The pilot GUI had two dock panels presenting the same kind of foldable tree, one for the 3D mesh viewer and one for the 2D canvas world. This factors the shared presentation into a base widget and merges them into a single Inspector dock that follows the active sub-window -- the 3D viewer shows the mesh-information tree, the 2D canvas shows the world entity tree -- so there is one panel and one View/Panels toggle instead of two. The old MeshInfo and EntityTreePanel feature classes are removed and their tests are consolidated into the tree-panel test.

Related to #1029.


j8xixo12:
Focus on mesh viewer 
image

Focus on 2D canvas
image

solvcon/pilot/_tree_panel.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        widget3d = self._mgr.currentR3DWidget()
+        if widget3d is not None:
+            self._stack.setCurrentWidget(self._mesh_tree)
+            self._mesh_tree.set_mesh(widget3d.mesh)
+            return
+        widget2d = self._mgr.currentR2DWidget()
+        if widget2d is not None:
+            self._stack.setCurrentWidget(self._entity_tree)
+            self._entity_tree.set_world(widget2d.world)
```
tigercosmos, line 528: Discussion, what if we want to see both 2D and 3D dianostic info?

j8xixo12: Good point, that means I need to find another way to detect the viewer.

j8xixo12: Currently, the 3D canvas is not yet available. Can we create an issue to track this requirement and use another PR to improve the detection method ?

solvcon/pilot/_tree_panel.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(*args, **kw)
+        self._action = None
+        self._dock = None
+        self._stack = None
+        self._mesh_tree = None
+        self._entity_tree = None
+
+    def populate_menu(self):
+        self._action = self.add_action(
+            "View/Panels", "Inspector", "Toggle the inspector panel",
```
j8xixo12, line 475: Rename the panel to `Inspector`.

solvcon/pilot/_tree_panel.py
tigercosmos: I think a simple way is to separate the 2D and the 3D inspector. My gut feeling tells me that they suppose to be two panels. What do you think?

tigercosmos: Or maybe a single inspector panel with many sub-panels, just like Chrome Debugger.

solvcon/pilot/_tree_panel.py
yungyuc: Inspector is a good name.  But please elaborate why it is named as inspector.

solvcon/pilot/_tree_panel.py
yungyuc: There will be only one `world` per sub-window, so the scenario to inspect both 2D and 3D world objects does not exist.

The use case of managing both 2D and 3D geometry entities exist, but they will need a 3D world.  We should not over-engineer before the code to create 3D geometry is written.

solvcon/pilot/_entity_tree.py
yungyuc, line 1: Remove `_entity_tree.py`, `_mesh_info.py`, and `_tree_panel.py` and move the code into `_tree_panel.py`.  The files are short and host similar code.  The many files are hard to maintain.

In the future, when the code grows longer, we can evaluate 2 approaches for better structure: (1) a sub-package and (2) a plug-in system.  For many widgets, option 2 may be required.  But at the moment, we have only short code and should not over-engineer.

yungyuc (changes requested):
- [x] Remove `_entity_tree.py`, `_mesh_info.py`, and `_tree_panel.py` and move the code into `_tree_panel.py`.
- [x] Respond in the PR comments to elaborate the reason behind naming the widget as inspector.

solvcon/pilot/_tree_panel.py
j8xixo12: The panel content updates dynamically based on the user's current widget focus to display the object information like mesh information or canvas information, the process known as inspection. Based on this behavior, Claude suggested the term Inspector. I think the name is intuitive, therefore I choice it.

yungyuc:
@j8xixo12 Please resolve the conflict.  When ready for review, leave a global comment.  Review the [review protocol](https://github.com/solvcon/solvcon/blob/0024a665a3e14c311184b3639215f03e0330ca54/.github/pull_request_template.md?plain=1#L9).

j8xixo12:
@yungyuc @tigercosmos This PR is ready for review.

tigercosmos (approved):
LGTM, please squash the commits. Thanks.

solvcon/pilot/_tree_panel_base.py
yungyuc, line 1: If there is not a reason to leave `_tree_panel_base.py` as a standalone file, move it to `_tree_panel.py`.

yungyuc (changes requested):
- [x] Merge `_tree_panel_base.py` into `_tree_panel.py`.  One file suffices unless there is a reason to use two.
- [x] Use fewer commits.  One or two are OK.

j8xixo12:
> * [x]  Merge `_tree_panel_base.py` into `_tree_panel.py`.  One file suffices unless there is a reason to use two.
> * [x]  Use fewer commits.  One or two are OK.

@yungyuc This PR is ready for review.

tigercosmos (approved):
I think these two commits are similar enough that they could be squashed, but I’m also fine keeping them separate. LGTM.

yungyuc (approved):
LGTM

PR #1040, opened 2026-07-09
Add typed handles and typed evaluation to the dynamic toggle store (steps 1, 2 of toggle)

### LINKED ISSUE #1022: Unify the toggle system into one store

yungyuc:
The dynamic toggle table stores runtime feature toggles keyed by string. Before this change it had two weaknesses. Its values lived in seven parallel `std::vector`s (one per type), so a stored value moved in memory whenever an unrelated toggle grew a vector past its capacity, which rules out handing a hot path a stable reference. And it could only be read by a string key that re-hashes on every access and returns a silent sentinel (`false`, `0`, `NaN`, `""`) on a missing or wrong-typed key, so a misspelled key read as `false` with no error.

This branch redesigns the store's storage and read contract in two reviewable steps: first stable-address register storage, then the typed handle and typed read API built on top of it. The existing `get`/`set`/`clear`/`keys` surface keeps its behavior, and the old silent-sentinel getters are kept but deprecated.

## Step 1: store dynamic toggles in stable-address registers

Replace the seven per-type value vectors with segmented columns whose element addresses stay put for the column's lifetime.

- `ToggleRegister<T>` boxes a single stored value, named after the atomic register of shared-memory theory. For now it holds a plain value; a later step can make the scalar read and write atomic without changing its shape.
- `ToggleRegisterColumn<T>` is a per-type column backed by a `std::deque`, so an appended register keeps its address as the column grows, unlike a reallocating `std::vector`. Indices are dense and start at zero, matching the offsets held by `DynamicToggleIndex`.
- The existing `get`/`set` route through the columns unchanged in behavior: a get still returns the sentinel on a missing or wrong-typed key; a set still overwrites a matching type, ignores a mismatched one, and appends a new register otherwise.
- A table generation counter is bumped on every `clear`, so a handle taken before a clear can be told from one taken after. Nothing reads it yet; it is the seam the next step builds the handle on. The public `get`/`set`/`clear`/`keys` surface is unchanged.

## Step 2: add typed handles and typed evaluation

Build the resolved handle and the typed read contract on top of the register storage.

- `ToggleRef<T>` binds a key to its register once (table, index, and the table generation at resolution). Because register addresses are stable it keeps reading through unrelated growth, and because it stamps the generation it is refused after a `dynamic_clear` instead of aliasing a reused register. It exposes `valid()`, `load()`, and `store()`.
- `declare<T>(key, default)` creates a toggle and returns a handle. Re-declaring the same type is idempotent and returns a handle to the existing register; a conflicting type throws `std::invalid_argument`.
- `ref<T>(key)` resolves an existing key to a handle, or returns an invalid handle when the key is missing or typed differently.
- `get<T>(key, default)` returns the caller default on a miss or mismatch, and `at<T>(key)` throws `std::out_of_range`, replacing the silent sentinels (the OpenFeature typed contract).
- `ToggleTypeTraits<T>` maps each supported value type to its `DynamicToggleIndex` tag; the primary template is left undefined so an unsupported type is a compile error rather than a silent mismatch.

Python gains `declare_bool` / `declare_int8` / `declare_int16` / `declare_int32` / `declare_int64` / `declare_real` / `declare_string`, a generic `get(key, default)` and `at(key)`, and the old `get_TYPE` sentinel getters now emit a `DeprecationWarning` pointing at `get(key, default)` and `at(key)`.

## Notes

- The register boxing and the generation-stamped handle are deliberately laid down as seams for later work (atomic scalar access, hot-path binding) without changing today's behavior.
- No caller is migrated off the deprecated `get_TYPE` getters in this branch; they still work and only warn.

For steps 1, 2 of issue #1022.

### DIFF
--- a/cpp/solvcon/toggle/toggle.hpp
+++ b/cpp/solvcon/toggle/toggle.hpp
@@ -19,7 +19,10 @@
 
 #include <cstdint>
 #include <deque>
+#include <format>
+#include <stdexcept>
 #include <string>
+#include <type_traits>
 #include <unordered_map>
 #include <vector>
 
@@ -27,6 +30,9 @@ namespace solvcon
 {
 
 int setenv(const char * name, const char * value, int overwrite);
+
+template <typename T>
+class ToggleRef;
 
 /**
  * A single stored toggle value.
@@ -127,6 +133,33 @@ struct DynamicToggleIndex
     Type type = TYPE_NONE;
 
 }; /* end struct DynamicToggleIndex */
+
+/**
+ * Maps a C++ value type to its DynamicToggleIndex type tag.
+ *
+ * The primary template is left undefined so an unsupported type is a
+ * compile error rather than a silent mismatch.
+ *
+ * @ingroup group_core
+ */
+template <typename T>
+struct ToggleTypeTraits;
+
+#define MM_TOGGLE_TYPE_TRAITS(CTYPE, TAG)               \
+    template <>                                         \
+    struct ToggleTypeTraits<CTYPE>                      \
+    {                                                   \
+        static constexpr DynamicToggleIndex::Type tag = \
+            DynamicToggleIndex::TAG;                    \
+    };
+MM_TOGGLE_TYPE_TRAITS(bool, TYPE_BOOL)
+MM_TOGGLE_TYPE_TRAITS(int8_t, TYPE_INT8)
+MM_TOGGLE_TYPE_TRAITS(int16_t, TYPE_INT16)
+MM_TOGGLE_TYPE_TRAITS(int32_t, TYPE_INT32)
+MM_TOGGLE_TYPE_TRAITS(int64_t, TYPE_INT64)
+MM_TOGGLE_TYPE_TRAITS(double, TYPE_REAL)
+MM_TOGGLE_TYPE_TRAITS(std::string, TYPE_STRING)
+#undef MM_TOGGLE_TYPE_TRAITS
 
 class DynamicToggleTable;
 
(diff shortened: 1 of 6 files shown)


cpp/solvcon/toggle/toggle.hpp
```diff
@@ -26,6 +31,62 @@ namespace solvcon
...
+        m_registers.emplace_back();
+        m_registers.back().value = value;
+        return index;
+    }
+
+    void clear() { m_registers.clear(); }
+
+private:
+
+    std::deque<ToggleRegister<T>> m_registers;
```
yungyuc, line 86: Use `std::deque` whose element address is stable: it does not change as the size grows.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -26,6 +31,62 @@ namespace solvcon
...
+ * A single stored toggle value.
+ *
+ * Named after the atomic register of shared-memory theory: a single-value
+ * shared location. For now it boxes a plain value; a later step makes the
+ * scalar read and write atomic without changing this shape.
+ *
+ * @ingroup group_core
+ */
+template <typename T>
+struct ToggleRegister
```
yungyuc, line 47: Single-value (scalar) location for atomic toggle datum.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -184,16 +272,55 @@ class DynamicToggleTable
...
+
     keymap_type m_key2index;
-    std::vector<bool> m_vector_bool;
-    std::vector<int8_t> m_vector_int8;
-    std::vector<int16_t> m_vector_int16;
-    std::vector<int32_t> m_vector_int32;
-    std::vector<int64_t> m_vector_int64;
-    std::vector<double> m_vector_real;
-    std::vector<std::string> m_vector_string;
+    ToggleRegisterColumn<bool> m_column_bool;
```
yungyuc, line 316: Use `ToggleRegisterColumn` by type to replace `std::vector`.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -202,6 +329,194 @@ inline DynamicToggleIndex HierarchicalToggleAccess::get_index(std::string const
...
+ * generation at resolution) so a hot path reads the value without touching
+ * the string map again. Because the storage keeps register addresses
+ * stable, the bound pointer survives unrelated toggles being added; because
+ * the handle stamps the table generation, a handle taken before a
+ * dynamic_clear is refused afterward instead of aliasing a reused register.
+ *
+ * @ingroup group_core
+ */
+template <typename T>
+class ToggleRef
```
yungyuc, line 345: Reference to a toggle by remembering its address and it will not need to look up again for the hot path.

PR #1041, opened 2026-07-09
Make the toggle store thread-safe and observable (steps 3, 4 of toggle)

### LINKED ISSUE #1022: Unify the toggle system into one store

yungyuc:
`Toggle::instance()` is a process-global that a solver thread and the UI thread both reach, and until now nothing synchronized it and nothing could subscribe to a change. A dynamic set could grow a value column while another thread read it, the read itself was an unguarded plain load, and the UI could only learn of a change by polling. This branch closes both gaps: it makes the store safe under concurrent access, then adds an observable change hook with a defined contract. It continues the toggle redesign on top of the register storage and typed handles from the earlier steps.

## Step 3: safe under concurrent access

Make the scalar register a `std::atomic` and serialize the rest behind a mutex.

- `ToggleRegister<T>` boxes a `std::atomic` for scalars and exposes get/set with `memory_order_relaxed`, which is sound because a toggle publishes nothing but itself. The `std::string` register stays a plain string guarded by the table mutex, and is documented as not a hot-path value.
- Each register is heap-owned in the column (a `deque` of `unique_ptr`), so a non-movable atomic lives in the container and register addresses stay stable for the handles.
- The key map, structural growth, string access, and clear run under the table mutex. The table generation is a `std::atomic` bumped on clear, so a handle's validity check stays lock-free on the hot path: a `ToggleRef` scalar read is one relaxed atomic load with no lock.
- The store now holds an atomic and a mutex, so it is no longer movable or assignable. `clone()` still deep-copies through a private constructor that holds the source lock for the whole copy.

## Step 4: notify observers when a toggle changes

Add an observable change hook so the UI can stop polling and a solver can be told to re-read.

- `on_change(key, callback)` returns an RAII `ToggleSubscription`; dropping it unsubscribes, so a callback never outlives the widget it belongs to.
- A write fires the callback through one mutating entry point (`set`), and only when the stored value actually changes. The no-op guard compares under the write lock; doubles compare by bit pattern so `NaN` and `-0.0` behave, which also keeps a two-way UI binding from echoing.
- The callback runs after the write lock is released, so it may call back into the store (including `set`) without deadlock, and firing outside the lock keeps a Python callback from inverting lock order against the GIL. A throwing callback is caught so it neither propagates nor stops the other observers.
- Callbacks are held by `shared_ptr` so a firing thread copies a pointer, not the `std::function`; copying a `std::function` that wraps a Python callable would touch a Python refcount without the GIL. The Python binding runs the callback under a reacquired GIL and reports a raising callback through the unraisable hook.

Python gains `on_change(key, callback)` returning a `ToggleSubscription` object (with `unsubscribe()` and an `active` property).

## Notes

- The design was checked clean under ThreadSanitizer for both the concurrent read/write paths and the concurrent notify path.
- The Qt bridge to a queued UI-thread signal is intentionally not here; it lands with the menu wiring that consumes it.

For steps 3, 4 of issue #1022.

### DIFF
--- a/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
+++ b/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
@@ -499,6 +499,40 @@ WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const *
         .def("declare_real", &detail::declare_toggle<double>, py::arg("key"), py::arg("default"))
         .def("declare_string", &detail::declare_toggle<std::string>, py::arg("key"), py::arg("default"))
         .def(
+            "on_change",
+            [](wrapped_type & self, std::string const & key, py::function const & callback)
+            {
+                // Run the Python callback under the GIL (it may fire from a
+                // solver thread that does not hold it) and never let it
+                // propagate a C++ exception out of the store.
+                // The whole body is wrapped so nothing escapes into the
+                // store's notify loop (which also guards callbacks); the
+                // analysis cannot always see that through the GIL guard.
+                // NOLINTNEXTLINE(bugprone-exception-escape)
+                auto wrapped = [callback]()
+                {
+                    try
+                    {
+                        py::gil_scoped_acquire const acquire;
+                        try
+                        {
+                            callback();
+                        }
+                        catch (py::error_already_set & e)
+                        {
+                            e.discard_as_unraisable("solvcon.Toggle on_change callback");
+                        }
+                    }
+                    // NOLINTNEXTLINE(bugprone-empty-catch)
+                    catch (...)
+                    {
+                    }
+                };
+                return self.on_change(key, std::move(wrapped));
+            },
+            py::arg("key"),
+            py::arg("callback"))
+        .def(
             "__getattr__",
             [](wrapped_type & self, std::string const & key)
             {
@@ -691,6 +725,14 @@ WrapProcessInfo::WrapProcessInfo(pybind11::module & mod, char const * pyname, ch
 
 void wrap_Toggle(pybind11::module & mod)
 {
+    namespace py = pybind11;
+
+    // The RAII subscription token returned by Toggle.on_change. Dropping the
+    // Python handle (or calling unsubscribe) removes the callback.
+    py::class_<ToggleSubscription>(mod, "ToggleSubscription")
+        .def("unsubscribe", &ToggleSubscription::reset)
+        .def_property_readonly("active", &ToggleSubscription::active);
+
     WrapSolidToggle::commit(mod, "SolidToggle", "SolidToggle");
     WrapFixedToggle::commit(mod, "FixedToggle", "FixedToggle");
     WrapHierarchicalToggleAccess::commit(mod, "HierarchicalToggleAccess", "HierarchicalToggleAccess");
--- a/cpp/solvcon/toggle/toggle.cpp
+++ b/cpp/solvcon/toggle/toggle.cpp
@@ -79,33 +79,42 @@ MM_DECL_DYNGET(std::string const &, string, sentinel_string)
 
 /* The macro gives debuggers a hard time. Manually expand it if you need to
  * step in a debugger. */
-#define MM_DECL_DYNSET(CTYPE, MTYPE, MTYPEC)                                                                                   \
-    /* NOLINTNEXTLINE(bugprone-easily-swappable-parameters) */                                                                 \
-    void DynamicToggleTable::set_##MTYPE(std::string const & key, CTYPE value)                                                 \
-    {                                                                                                                          \
-        std::scoped_lock const guard(m_mutex);                                                                                 \
-        auto it = m_key2index.find(key);                                                                                       \
-        if (it != m_key2index.end())                                                                                           \
-        {                                                                                                                      \
-            if (it->second.is_##MTYPE())                                                                                       \
-            {                                                                                                                  \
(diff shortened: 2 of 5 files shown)


cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -498,6 +498,40 @@ WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const *
         .def("declare_int64", &detail::declare_toggle<int64_t>, py::arg("key"), py::arg("default"))
         .def("declare_real", &detail::declare_toggle<double>, py::arg("key"), py::arg("default"))
         .def("declare_string", &detail::declare_toggle<std::string>, py::arg("key"), py::arg("default"))
+        .def(
+            "on_change",
+            [](wrapped_type & self, std::string const & key, py::function const & callback)
+            {
+                // Run the Python callback under the GIL (it may fire from a
```
yungyuc, line 505: The locking behavior is tricky.  We will need to test it when using.

cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -498,6 +498,40 @@ WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const *
...
+                // analysis cannot always see that through the GIL guard.
+                // NOLINTNEXTLINE(bugprone-exception-escape)
+                auto wrapped = [callback]()
+                {
+                    try
+                    {
+                        py::gil_scoped_acquire const acquire;
+                        try
+                        {
+                            callback();
```
yungyuc, line 519: Calling the callback from the observer.

cpp/solvcon/toggle/toggle.cpp
```diff
@@ -78,32 +79,42 @@ MM_DECL_DYNGET(std::string const &, string, sentinel_string)
...
+                DynamicToggleIndex const idx{static_cast<uint32_t>(m_column_##MTYPE.size()), DynamicToggleIndex::TYPE_##MTYPEC}; \
+                m_key2index.insert({key, idx});                                                                                  \
+                m_column_##MTYPE.append(value);                                                                                  \
+                changed = true;                                                                                                  \
+            }                                                                                                                    \
+        }                                                                                                                        \
+        if (changed)                                                                                                             \
+        {                                                                                                                        \
+            notify(key);                                                                                                         \
```
yungyuc, line 112: Notify the observer.

cpp/solvcon/toggle/toggle.cpp
```diff
@@ -155,7 +169,38 @@ void DynamicToggleTable::clear()
...
+            }
+        }
+    }
+    for (auto const & callback : callbacks)
+    {
+        try
+        {
+            (*callback)();
+        }
+        // A throwing observer must not corrupt the store or stop the other
```
yungyuc, line 197: I feel that we may need more tailored treatment for exceptions in the future.

But we have not tested it in applications yet, so leave it here.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -38,25 +44,46 @@ class ToggleRef;
...
 {
-    T value;
+    std::atomic<T> value;
+    T get() const { return value.load(std::memory_order_relaxed); }
+    void set(T v) { value.store(v, std::memory_order_relaxed); }
 }; /* end struct ToggleRegister */
 
 /**
- * Segmented column of ToggleRegister<T> with stable element addresses.
+ * String register.
```
yungyuc, line 62: Add a string register for dynamic toggles.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -66,24 +93,38 @@ class ToggleRegisterColumn
...
+        m_registers.back()->set(value);
         return index;
     }
 
     void clear() { m_registers.clear(); }
 
 private:
 
-    std::deque<ToggleRegister<T>> m_registers;
+    std::deque<std::unique_ptr<ToggleRegister<T>>> m_registers;
```
yungyuc, line 127: We will need to study if using unique pointer for toggle store hurts memory performance by too much.

A million toggles will be an absurd number, and the heap may be OK with it.  Created an issue #1042 to track it.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -161,6 +202,117 @@ MM_TOGGLE_TYPE_TRAITS(double, TYPE_REAL)
...
+
+    void swap(ToggleSubscription & other) noexcept
+    {
+        m_observers.swap(other.m_observers);
+        m_key.swap(other.m_key);
+        std::swap(m_id, other.m_id);
+        std::swap(m_active, other.m_active);
+    }
+
+    std::weak_ptr<detail::ToggleChangeObservers> m_observers;
```
yungyuc, line 309: Use a weak pointer because `ToggleChangeObservers` hold shared pointer to callbacks.

tests/test_toggle.py
```diff
@@ -332,6 +332,96 @@ def test_deprecated_getter_warns(self):
         self.assertTrue(issubclass(caught[0].category, DeprecationWarning))
 
 
+class ToggleOnChangeTC(unittest.TestCase):
```
yungyuc, line 335: Test for toggle callback.

@tigercosmos @j8xixo12 @ThreeMonth03 this will be very useful to GUI.

PR #1043, opened 2026-07-09
Consolidate the toggle store, migrate off the facades, and bind it to the pilot menu (steps 5, 6, 7 of toggle)

### LINKED ISSUE #1022: Unify the toggle system into one store

yungyuc:
This branch continues the toggle redesign on top of the concurrency guard and change-notification hook. It folds the last two parallel storage mechanisms (the compile-time solid toggle and the fixed-name toggle) onto the one store, migrates the remaining callers off the compatibility facades and retires the deprecated getters, and then gives the pilot a reusable way to back a checkable menu action with a store toggle through a Qt bridge. The end state is a single storage model with one read contract that a widget or a solver can subscribe to, wired end to end into the GUI menu.

## Step 5: consolidate the classes onto one store

Fold the three parallel storage mechanisms into one. The single solid toggle, `use_pyside`, becomes an inline `constexpr` build switch in `build_config.hpp` that the optimizer folds away, with `SolidToggle` reduced to a stateless facade over it. `python_redirect` and `show_axis` become declared toggles created at `Toggle` construction with startup defaults and a category, so they gain change notification and serialization like any other toggle, and `FixedToggle` becomes a facade that reads and writes them through the store; `to_python` still reports them under the "fixed" section and skips them in the dynamic section, so the serialization format is unchanged. A `ToggleCategory` (release, ops, experiment) is recorded per key at declare time for a later lint or test to flag a stale release toggle; it enforces nothing by itself. This step also fixes a `HierarchicalToggleAccess` lifetime bug (a subkey access holds a raw table pointer, so a returned access is now tied to its owner with `keep_alive`; a leaf read returns a plain value and cannot carry the reference, so the owner is tied only when the result is itself an access), which un-skips `test_load_bad_lifecycle`. It also makes `solvcon.toggle.load` generic, walking the JSON tree and inferring the type of each value for any app rather than only euler1d.

## Step 6: migrate call sites off the facades

With the store unified, move the remaining readers off the solid and fixed facades and retire the deprecated getters. `wrap_pilot` reads `use_pyside` directly from `build_config` instead of `Toggle::instance().solid()`, so the branch folds at compile time. The Python stream redirect and the console dock read and write `python_redirect` through the typed store (`get<bool>(key, default)` and `set_bool`) instead of `Toggle::instance().fixed()`, so they observe the same value as everything else and pick up change notification. The silent-sentinel getters are removed: the Python `get_bool`/`get_int8`/.../`get_string` on `Toggle` (deprecated in an earlier step) and the unused C++ `Toggle::get_TYPE` forwards. Callers use `get<T>(key, default)` or `at<T>(key)`, while the `set_TYPE` writers stay as the mutating entry point; the lower-level `get_TYPE` on the dynamic access and the table remain, since the type-dispatched read still uses them. Tests that encoded the old sentinel behavior now assert the typed contract, and the `Toggle` class documentation is rewritten around the one-store model.

## Step 7: bind the pilot menu to the store

Give the pilot a reusable way to back a checkable menu action with a store toggle. `build_action` and `PilotFeature.add_action` gain an optional `toggle_key`; when set, the action is bound two-way to that toggle through a `ToggleActionBridge` so the action's `toggled` writes the toggle and the toggle's `on_change` re-checks the action. `RMenuModel` stays the placement layer while the store owns the value, so the state persists, round-trips to JSON, and can be observed by a solver or a second widget. The bridge is the change hook's Qt bridge: `on_change` fires a queued signal to `setChecked`, so a change from a solver thread lands on the UI thread. The store's no-op guard plus Qt emitting `toggled` only on a real change break the echo, so no `blockSignals` dance is needed; the bridge is parented to its action and drops its subscription when the action is destroyed, so a later change never calls back into a dead widget. `Toggle.clear_change_observers` is added and registered with `atexit`, so a callback holding a Python object is released while the interpreter is still alive rather than when the store's static is destroyed after finalization.

For steps 5, 6, 7 of issue #1022.

### DIFF
--- a/cpp/solvcon/pilot/RPythonConsoleDockWidget.cpp
+++ b/cpp/solvcon/pilot/RPythonConsoleDockWidget.cpp
@@ -288,7 +288,7 @@ RPythonConsoleDockWidget::RPythonConsoleDockWidget(const QString & title, QWidge
     : QDockWidget(title, parent, flags)
     , m_history_edit(new RPythonHistoryTextEdit)
     , m_command_edit(new RPythonCommandTextEdit)
-    , m_python_redirect(Toggle::instance().fixed().get_python_redirect())
+    , m_python_redirect(Toggle::instance().get<bool>("python_redirect", true))
 {
     auto * container = new QWidget;
     auto * layout = new QVBoxLayout;
--- a/cpp/solvcon/pilot/wrap_pilot.cpp
+++ b/cpp/solvcon/pilot/wrap_pilot.cpp
@@ -1034,7 +1034,7 @@ void initialize_pilot(pybind11::module & mod)
         wrap_pilot(mod);
     };
 
-    if (Toggle::instance().solid().use_pyside())
+    if (build_config::use_pyside)
     {
         try
         {
--- a/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
+++ b/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
@@ -756,6 +756,14 @@ void wrap_Toggle(pybind11::module & mod)
         .value("Ops", ToggleCategory::Ops)
         .value("Experiment", ToggleCategory::Experiment);
 
+    // Release every change callback held by the global store at interpreter
+    // shutdown, while Python is still alive, so a callback holding a Python
+    // object (for example a pilot menu bridge) is not freed after
+    // finalization when the store's static destructor runs.
+    py::module_::import("atexit").attr("register")(
+        py::cpp_function([]()
+                         { Toggle::instance().clear_change_observers(); }));
+
     WrapSolidToggle::commit(mod, "SolidToggle", "SolidToggle");
     WrapFixedToggle::commit(mod, "FixedToggle", "FixedToggle");
     WrapHierarchicalToggleAccess::commit(mod, "HierarchicalToggleAccess", "HierarchicalToggleAccess");
--- a/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
+++ b/cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
@@ -448,17 +448,6 @@ void declare_toggle(Toggle & self, std::string const & key, T const & value)
     self.declare<T>(key, value);
 }
 
-void warn_deprecated_getter(char const * name)
-{
-    PyErr_WarnEx(
-        PyExc_DeprecationWarning,
-        std::format("Toggle.{} returns a silent sentinel on a missing or wrong-typed "
-                    "key and is deprecated; use get(key, default) or at(key)",
-                    name)
-            .c_str(),
-        1);
-}
-
 } /* end namespace detail */
 
 WrapToggle::WrapToggle(pybind11::module & mod, char const * pyname, char const * pydoc)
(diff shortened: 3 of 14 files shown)


cpp/solvcon/pilot/RPythonConsoleDockWidget.cpp
```diff
@@ -288,7 +288,7 @@ RPythonConsoleDockWidget::RPythonConsoleDockWidget(const QString & title, QWidge
     : QDockWidget(title, parent, flags)
     , m_history_edit(new RPythonHistoryTextEdit)
     , m_command_edit(new RPythonCommandTextEdit)
-    , m_python_redirect(Toggle::instance().fixed().get_python_redirect())
+    , m_python_redirect(Toggle::instance().get<bool>("python_redirect", true))
```
yungyuc, line 291: Use the unified getter, which indicates type too.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -1034,7 +1034,7 @@ void initialize_pilot(pybind11::module & mod)
         wrap_pilot(mod);
     };
 
-    if (Toggle::instance().solid().use_pyside())
+    if (build_config::use_pyside)
```
yungyuc, line 1037: Compile-time determined configuration.

cpp/solvcon/toggle/pymod/wrap_Toggle.cpp
```diff
@@ -135,7 +153,18 @@ WrapHierarchicalToggleAccess::WrapHierarchicalToggleAccess(pybind11::module & mo
...
+            "__getattr__",
+            [](py::object const & self_obj, std::string const & key)
+            {
+                auto & self = self_obj.cast<wrapped_type &>();
+                // A subkey access holds a raw table pointer, so a chained
+                // access off a temporary must keep its owner alive.
+                // __getattr__ may also return a plain value (a leaf toggle),
+                // which cannot carry a keep_alive reference, so tie the parent
+                // only when the result is itself an access.
+                return detail::tie_subkey_owner(getattr(self, key), self_obj);
```
yungyuc, line 166: Make sure the lifecycle is correct.

cpp/solvcon/toggle/toggle.hpp
```diff
@@ -37,6 +38,23 @@ namespace solvcon
...
+ * so a lint or test can flag, for example, a release toggle that outlived its
+ * feature. Ops and UI settings are long-lived; a release toggle is removed
+ * once its feature lands.
+ *
+ * @ingroup group_core
+ */
+enum class ToggleCategory : uint8_t
+{
+    Release,
+    Ops,
```
yungyuc, line 54: This `Ops` (operational) is the default.  It is the long-living toggle that we will frequently use in code.

solvcon/pilot/_gui_common.py
```diff
@@ -3,17 +3,70 @@
 
 
 from . import _pilot_core as _pcore
+from ..core import Toggle
 from PySide6 import QtCore, QtGui
 
 
+class ToggleActionBridge(QtCore.QObject):
```
yungyuc, line 10: With this, we will be ready to define parameters for GUI.  @j8xixo12

PR #1047, opened 2026-07-09
Add make run_pilot to build and launch Pilot

tigercosmos:
Add a make run_pilot target that builds pilot and launches it with PYTHONPATH set, resolving the platform binary path so it works on macOS, Linux, and Windows.


tigercosmos:

yungyuc (approved):
LGTM

PR #1054, opened 2026-07-11
Own `QApplication` only when the pilot creates it

### LINKED ISSUE #990: Review the use of `std::unique_ptr` for Qt code
Qt has its own ownership system.  In the new `QRhi`-based GUI design (`R`-classes), we are using a lot of unique pointers.  Study whether or not we are doing it in a right way, and evaluate how to improve.

yungyuc:
`RManager` held the application object in a single owning `std::unique_ptr<QCoreApplication>` that adopted whatever `QApplication::instance()` returned. When the pilot is embedded in Python, PySide6 has already created that application and owns it, so the owning handle made RManager a second owner. Not good.

For issue #990.

### DIFF
--- a/cpp/solvcon/pilot/RManager.cpp
+++ b/cpp/solvcon/pilot/RManager.cpp
@@ -33,13 +33,19 @@ RManager & RManager::instance()
 RManager::RManager()
     : QObject()
 {
-    m_core.reset(QApplication::instance());
-    static int argc = 1;
-    static char exename[] = "pilot";
-    static char * argv[] = {exename};
-    if (!m_core)
+    if (QCoreApplication * const existing = QApplication::instance())
     {
-        m_core.reset(new QApplication(argc, argv));
+        // Borrow an application another runtime owns, for example the one
+        // PySide6 creates when the pilot is embedded in Python.
+        m_core = existing;
+    }
+    else
+    {
+        static int argc = 1;
+        static char exename[] = "pilot";
+        static char * argv[] = {exename};
+        m_owned_core = std::make_unique<QApplication>(argc, argv);
+        m_core = m_owned_core.get();
     }
 
     m_mainWindow = new QMainWindow;
@@ -74,7 +80,8 @@ void RManager::reset()
     {
         m_menuModel->clear();
     }
-    m_core.reset();
+    m_owned_core.reset(); // Deletes the application only if the pilot made it.
+    m_core = nullptr;
     m_mainWindow = nullptr;
     m_menuModel = nullptr;
     m_pycon = nullptr;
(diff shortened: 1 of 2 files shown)


cpp/solvcon/pilot/RManager.cpp
```diff
@@ -33,13 +33,19 @@ RManager & RManager::instance()
...
-    static int argc = 1;
-    static char exename[] = "pilot";
-    static char * argv[] = {exename};
-    if (!m_core)
+    if (QCoreApplication * const existing = QApplication::instance())
     {
-        m_core.reset(new QApplication(argc, argv));
+        // Borrow an application another runtime owns, for example the one
+        // PySide6 creates when the pilot is embedded in Python.
+        m_core = existing;
```
yungyuc, line 40: Borrow from existing.

cpp/solvcon/pilot/RManager.cpp
```diff
@@ -33,13 +33,19 @@ RManager & RManager::instance()
...
+    }
+    else
+    {
+        static int argc = 1;
+        static char exename[] = "pilot";
+        static char * argv[] = {exename};
+        m_owned_core = std::make_unique<QApplication>(argc, argv);
```
yungyuc, line 47: There is not a `QApplication` instantiated, so the `RManager` constructor should.

This implementation is not thread-safe.  Let's worry about concurrency later.

cpp/solvcon/pilot/RManager.cpp
```diff
@@ -33,13 +33,19 @@ RManager & RManager::instance()
...
+        m_core = m_owned_core.get();
```
yungyuc, line 48: Own the constructed `QApplication`.

cpp/solvcon/pilot/RManager.hpp
```diff
@@ -122,7 +122,11 @@ class RManager
 
     bool m_already_setup = false;
 
-    std::unique_ptr<QCoreApplication> m_core = nullptr;
+    /// Non-owning handle, possibly borrowed from another runtime (PySide6).
+    QCoreApplication * m_core = nullptr;
```
yungyuc, line 126: `m_core` should be a raw pointer.

cpp/solvcon/pilot/RManager.hpp
```diff
@@ -122,7 +122,11 @@ class RManager
...
+
+    /// Owns the application only when the pilot created it, null otherwise.
+    std::unique_ptr<QApplication> m_owned_core = nullptr;
```
yungyuc, line 129: Only own the `QApplication` instance if pilot creates it.

PR #1055, opened 2026-07-11
Add `SimpleArray::where`

### LINKED ISSUE #519: Implement `SimpleArray` member functions for searching

Yuli-ytd:
This PR adds `SimpleArray::where()` with NumPy-style three-argument semantics.

The Python API uses `condition.where(x, y)`, where the condition is a `SimpleArrayBool`.

The implementation dispatches `x` and `y` by their `SimpleArray` dtype in the Python wrapper, and returns a result array with the same dtype as `x` and `y`.

The first implementation supports existing numeric `SimpleArray` dtypes and requires `condition`, `x`, and `y` to have the same shape.

Broadcasting and scalar `x` or `y` inputs are left for follow-up work.

Tests compare the result against `numpy.where()` for 1-D and N-D arrays, and cover shape mismatch handling.

Related to #519.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -1381,6 +1381,9 @@ public:
 
     SimpleArray<uint64_t> argwhere() const;
 
+    template <typename U>
+    SimpleArray<U> where(SimpleArray<U> const & x, SimpleArray<U> const & y) const;
+
 }; /* end class SimpleArrayMixinSearch */
 
 template <typename A, typename T>
@@ -2788,6 +2791,32 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
             offset %= stride;
         }
         ++idx;
+    }
+
+    return result;
+}
+
+template <typename A, typename T>
+template <typename U>
+SimpleArray<U> detail::SimpleArrayMixinSearch<A, T>::where(SimpleArray<U> const & x, SimpleArray<U> const & y) const
+{
+    auto athis = static_cast<A const *>(this);
+
+    static_assert(std::is_same_v<T, bool>, "SimpleArray::where() requires a boolean array");
+
+    if (athis->shape() != x.shape() || athis->shape() != y.shape())
+    {
+        throw std::invalid_argument(std::format(
+            "SimpleArray::where(): shape mismatch: condition={} x={} y={}",
+            format_shape(athis->shape()),
+            format_shape(x.shape()),
+            format_shape(y.shape())));
+    }
+
+    SimpleArray<U> result(athis->shape());
+    for (size_t i = 0; i < athis->size(); ++i)
+    {
+        result.data(i) = athis->data(i) ? x.data(i) : y.data(i);
     }
 
     return result;
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2813,11 +2813,30 @@ SimpleArray<U> detail::SimpleArrayMixinSearch<A, T>::where(SimpleArray<U> const 
             format_shape(y.shape())));
     }
 
-    SimpleArray<U> result(athis->shape());
-    for (size_t i = 0; i < athis->size(); ++i)
+    if (athis->nghost() != x.nghost() || athis->nghost() != y.nghost())
     {
-        result.data(i) = athis->data(i) ? x.data(i) : y.data(i);
+        throw std::invalid_argument(std::format(
+            "SimpleArray::where(): nghost mismatch: condition={} x={} y={}",
+            athis->nghost(),
+            x.nghost(),
+            y.nghost()));
     }
+
+    SimpleArray<U> result(athis->shape());
+    result.set_nghost(athis->nghost());
+
+    if (athis->shape().empty() || athis->size() == 0)
+    {
+        return result;
+    }
+
+    auto const range = IndexRange(*athis);
+    sshape_type idx = range.first();
+
+    do
+    {
+        result.at(idx) = athis->at(idx) ? x.at(idx) : y.at(idx);
+    } while (range.next(idx));
 
     return result;
 }
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -698,7 +698,70 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             //
             ;
 
+        if constexpr (std::is_same_v<T, bool>)
+        {
+            (*this).def(
+                "where",
+                &where,
+                py::arg("x"),
+                py::arg("y"));
+        }
+
         return *this;
+    }
+
+    static pybind11::object where(wrapped_type const & self, pybind11::object const & x, pybind11::object const & y)
+    {
+
+        std::string py_typename_x(pybind11::detail::obj_class_name(x.ptr()));
+        std::string py_typename_y(pybind11::detail::obj_class_name(y.ptr()));
+        const std::size_t found_x = py_typename_x.find("_solvcon.SimpleArray");
+        const std::size_t found_y = py_typename_y.find("_solvcon.SimpleArray");
+
+        if (found_x == std::string::npos || found_y == std::string::npos)
+        {
+            throw std::invalid_argument(
+                "SimpleArray::where(): x and y must be SimpleArray");
+        }
+
+        py_typename_x.replace(0, strlen("_solvcon.SimpleArray"), "");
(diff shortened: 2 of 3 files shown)


tests/test_buffer.py
```diff
@@ -4175,6 +4175,49 @@ def test_argwhere(self):
         ret = sarr.eq(10).argwhere()
         np.testing.assert_array_equal(ret.ndarray, np.argwhere(narr == 10))
 
+    def test_where(self):
```
Yuli-ytd, line 4185: The test keeps the public syntax close to `numpy.where(condition, x, y)` by forming the condition first and then calling `condition.where(x, y)`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2781,6 +2784,32 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
     return result;
 }
 
+template <typename A, typename T>
+template <typename U>
+SimpleArray<U> detail::SimpleArrayMixinSearch<A, T>::where(SimpleArray<U> const & x, SimpleArray<U> const & y) const
+{
+    auto athis = static_cast<A const *>(this);
+
+    static_assert(std::is_same_v<T, bool>, "SimpleArray::where() requires a boolean array");
```
Yuli-ytd, line 2805: `where()` is implemented on the search mixin as `condition.where(x, y)`. The static assertion keeps the C++ API limited to boolean condition arrays, matching the Python wrapper.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -698,8 +698,71 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+        const DataType dt_x(py_typename_x);
+        const DataType dt_y(py_typename_y);
+
+        if (dt_x != dt_y)
+        {
+            throw std::invalid_argument(
+                "SimpleArray::where(): x and y must have the same dtype");
+        }
+
+#define DECL_MM_WHERE_TYPED(DataTypeName) \
```
Yuli-ytd, line 741: This dispatch follows the existing `take_along_axis()` wrapper pattern: the condition is always `SimpleArrayBool`, while `x` and `y` determine the result dtype.

Yuli-ytd:
This PR is ready for review.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2781,6 +2784,32 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
...
+            "SimpleArray::where(): shape mismatch: condition={} x={} y={}",
+            format_shape(athis->shape()),
+            format_shape(x.shape()),
+            format_shape(y.shape())));
+    }
+
+    SimpleArray<U> result(athis->shape());
+    for (size_t i = 0; i < athis->size(); ++i)
+    {
+        result.data(i) = athis->data(i) ? x.data(i) : y.data(i);
```
yungyuc: This will not work when `x` and/or `y` is not contiguous.  We need to test them.  Two easy ways to create discontiguous arrays are (1) use negative stride and (2) having the two arrays to be C (row-major) and F (column-major) contiguity.

See [`test_buffer.py#L285`](https://github.com/solvcon/solvcon/blob/4101eb8fc6068d45706e25bd83fe462c3012b16a/tests/test_buffer.py#L285) for how to use negative stride.  Also test ghost array [`test_buffer.py#L465`](https://github.com/solvcon/solvcon/blob/4101eb8fc6068d45706e25bd83fe462c3012b16a/tests/test_buffer.py#L465).

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -698,8 +698,71 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+            DECL_MM_WHERE_TYPED(Float32)
+            DECL_MM_WHERE_TYPED(Float64)
+        default:
+            break;
+        }
+
+#undef DECL_MM_WHERE_TYPED
+
+        throw std::invalid_argument(
+            "SimpleArray::where(): unsupported dtype");
```
yungyuc: This is not a long line. Use a single line for the code:

```suggestion
        throw std::invalid_argument("SimpleArray::where(): unsupported dtype");
```

tests/test_buffer.py
```diff
@@ -4175,6 +4175,49 @@ def test_argwhere(self):
...
+        data = [1, 3, 5, 7, 9]
+        narr = np.array(data, dtype='uint64')
+        sarr = solvcon.SimpleArrayUint64(array=narr)
+
+        np.testing.assert_array_equal(
+            ((sarr < 5).where(sarr.add(1), sarr.mul(10))).ndarray,
+            np.where(narr < 5, narr + 1, narr * 10)
+        )
+
+        # test N-D data
```
yungyuc: Test at least 4-D.

yungyuc (changes requested):
- [x] Add unit tests for non-contiguous arrays.  Need to test for ghost arrays too.
- [x] Unit test should cover from 1D to at least 4D.
- [x] Use a single line when it fits.

Yuli-ytd:
> * [ ]  Add unit tests for non-contiguous arrays.  Need to test for ghost arrays too.
> * [ ]  Unit test should cover from 1D to at least 4D.
> * [ ]  Use a single line when it fits.

Thanks for the review. I will add the requested test coverage and update the inline annotations based on your suggestions.

I plan to push the revision before 7/ 14 (Tue).

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Updated. `SimpleArray::where()` now iterates over logical indices with `IndexRange` and `at(idx)`, instead of matching condition/x/y by data(i).

I also added coverage in `test_buffer.py::SimpleArraySearchTC::test_where` for positive-stride non-contiguous arrays, C/F-contiguous x/y layouts, and ghost arrays.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
Yuli-ytd: I folded this throw statement into one line.

tests/test_buffer.py
Yuli-ytd: test_where now covers 1-D, 2-D, 3-D, and 4-D contiguous arrays.

Yuli-ytd:
- [x] Add unit tests for non-contiguous arrays. Need to test for ghost arrays too.
- [x] Unit test should cover from 1D to at least 4D.
- [x] Use a single line when it fits.

I finished the requested updates.

**One note:** 
SimpleArray can already expose negative-stride NumPy views in some paths, but using `where()` with `at(idx)` over a negative-stride NumPy view appears to expose a separate wrapped-view/base-pointer issue. 

I used positive-stride non-contiguous and C/F-contiguous tests for this change, and I think the negative-stride `at(idx)` path should be handled as a separate follow-up.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2793,6 +2796,51 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
...
+
+    if (athis->shape().empty() || athis->size() == 0)
+    {
+        return result;
+    }
+
+    auto const range = IndexRange(*athis);
+    sshape_type idx = range.first();
+
+    do
```
yungyuc, line 2836: Somewhere near this loop a comment should remind maintainers that the special index object is needed for non-contiguous access.  It can be added in a follow-up PR.

tests/test_buffer.py
```diff
@@ -4182,6 +4182,113 @@ def test_argwhere(self):
...
+        cond_sarr = solvcon.SimpleArrayBool(array=cond_view)
+        x_sarr = solvcon.SimpleArrayFloat64(array=x_view)
+        y_sarr = solvcon.SimpleArrayFloat64(array=y_view)
+
+        np.testing.assert_array_equal(
+            ((cond_sarr).where(x_sarr, y_sarr)).ndarray,
+            np.where(cond_view, x_view, y_view)
+        )
+
+        # test c-contiguous and f-contiguous arrays
```
yungyuc, line 4234: Good addition.

The sections in this long test function may be organized in [sub-tests](https://docs.python.org/3.14/library/unittest.html#unittest.TestCase.subTest).  It can be done in a follow-up PR.

yungyuc (approved):
LGTM

PR #1051, opened 2026-07-10
Add worktree skill for isolated task worktrees

tigercosmos:
Add a worktree skill that enters or creates an isolated git worktree for the task and stops there, with agent-specific entry (Claude EnterWorktree, Cursor /worktree, otherwise git worktree add).

[skip-ci]

PR #1056, opened 2026-07-11
Ignore .claude/worktrees to fit the skill directive

yungyuc:
For PR #1051.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 CLAUDE.local.md
+.claude/worktrees/
 .vscode
 .vs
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 CLAUDE.local.md
+.claude/worktrees/
```
yungyuc, line 2: We forgot to ignore the worktrees directory.

tigercosmos: Thanks, not sure why but my git ignores worktree for me.

PR #1057, opened 2026-07-11
Add a basic window manager to the Window menu

### LINKED ISSUE #1030: Need a basic window manager for the "window" menu
Canvas and mesh create sub-windows.  There is not a list under the "window" menu to list the sub-windows and select them to the foreground.  Add a basic window manager using the "window" menu.

Also move the console toggle from window menu to view->panel.

j8xixo12:
Canvas and mesh viewers open MDI sub-windows, but the Window menu had no way to see them or bring one to the foreground. A WindowManager feature now rebuilds, each time the Window menu is about to show, a checkable action per visible sub-window; triggering one activates it and the active sub-window is shown checked. Closed viewers fall out of the list and an empty list shows a disabled placeholder. The feature is wired into the controller beside the existing Console toggle.

Related to #1030

### DIFF
--- a/solvcon/pilot/_base_app.py
+++ b/solvcon/pilot/_base_app.py
@@ -372,6 +372,9 @@ class OneDimBaseApp(_gui_common.PilotFeature):
     use_grid_layout: bool = False
     adjust_region: bool = False
 
+    #: Title of the plot sub-window; the Window menu lists windows by it.
+    TITLE = "Base Window"
+
     def populate_menu(self):
         """
         Set menu item for GUI.
@@ -394,6 +397,7 @@ class OneDimBaseApp(_gui_common.PilotFeature):
 
         # A new sub-window (`QMdiSubWindow`) for the plot area
         self._subwin = self._mgr.addSubWindow(QWidget())
+        self._subwin.setWindowTitle(self.TITLE)
         self._subwin.setWidget(PlotArea(self))
         self._subwin.showMaximized()
         self._subwin.show()
--- a/solvcon/pilot/_window_manager.py
+++ b/solvcon/pilot/_window_manager.py
@@ -1,0 +1,99 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Window manager feature for pilot.
+
+List every open MDI sub-window under the "Window" menu and bring one to
+the foreground when its entry is chosen.
+"""
+
+from PySide6 import QtGui
+
+from . import _gui_common
+
+__all__ = [
+    'WindowManager',
+]
+
+
+class WindowManager(_gui_common.PilotFeature):
+    """List open MDI sub-windows under the "Window" menu.
+
+    One checkable action per open sub-window, labelled by its title.
+    Triggering an action activates that sub-window; the active one is
+    checked. The list is rebuilt each time the menu is about to show.
+    """
+
+    #: objectName tagging every dynamic per-sub-window action.
+    ITEM_ID = "window.subwindow"
+
+    def __init__(self, *args, **kw):
+        super(WindowManager, self).__init__(*args, **kw)
+        self._menu = None
+        self._items = []
+
+    def populate_menu(self):
+        """Anchor the dynamic list on the "Window" menu.
+
+        The menu holds nothing static (panel toggles live under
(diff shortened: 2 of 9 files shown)


j8xixo12:
Without any window:
image

With multiple windows: 
image

j8xixo12:
@yungyuc @tigercosmos This PR is ready for review.

solvcon/pilot/_base_app.py
```diff
@@ -372,6 +372,9 @@ class OneDimBaseApp(_gui_common.PilotFeature):
     use_grid_layout: bool = False
     adjust_region: bool = False
 
+    #: Title of the plot sub-window; the Window menu lists windows by it.
+    TITLE = "1D solver"
```
tigercosmos: Should be `Base Window` or similar?

tigercosmos (commented):
I think we can have more window types, e.g. 2D canvas, Painter, Agent, etc, but I don't see them. Could you help clarify if this PR has covered those already?

yungyuc (changes requested):
- [x] Leave inline annotations to help review.
- [x] Move the console toggle from "window" to "view->panel".

> I think we can have more window types, e.g. 2D canvas, Painter, Agent, etc, but I don't see them. Could you help clarify if this PR has covered those already?

@j8xixo12 Please leave comments to discuss your plan.  It is not necessary to do everything in one PR.  Feel free to bring it to discord to discuss.

solvcon/pilot/_base_app.py
j8xixo12: I don't get you. Do you mean the window's title should be called `Base Window`. ?

j8xixo12:
> I think we can have more window types, e.g. 2D canvas, Painter, Agent, etc, but I don't see them. Could you help clarify if this PR has covered those already?

@tigercosmos The window manager only manages the mdi sub-windows such as the 2D canvas or domain viewer, whereas the painter and agent are dockable widgets. Therefore, the window manager will not contain these dockable widgets. Also, echoing @yungyuc point on moving console to `View/Panel`,  the dockable widgets are already placed under `View/Panel` via `add_action` inside `populate_menu`, for example https://github.com/solvcon/solvcon/blob/master/solvcon/pilot/_agent_gui.py#L111-L115, I think this is a good way to proceed.

solvcon/pilot/_base_app.py
tigercosmos: `TITLE` is in a base class, so "1D solver" seems incorrect.

tigercosmos:
Thanks, it's now clear to me.

solvcon/pilot/_window_manager.py
```diff
@@ -0,0 +1,97 @@
...
+        one. A disabled placeholder is shown when none are open.
+        """
+        for act in self._items:
+            self._menu.removeAction(act)
+            act.deleteLater()
+        self._items = []
+
+        mdi = self._mgr.mdiArea
+        active = mdi.activeSubWindow()
+        subwins = [s for s in mdi.subWindowList() if s.isVisible()]
```
j8xixo12, line 63: Record all opened sub-windows

solvcon/pilot/_window_manager.py
```diff
@@ -0,0 +1,97 @@
...
+        action per visible sub-window in area order, checking the active
+        one. A disabled placeholder is shown when none are open.
+        """
+        for act in self._items:
+            self._menu.removeAction(act)
+            act.deleteLater()
+        self._items = []
+
+        mdi = self._mgr.mdiArea
+        active = mdi.activeSubWindow()
```
j8xixo12, line 62: Record which sub-window is focused by user.

solvcon/pilot/_window_manager.py
```diff
@@ -0,0 +1,99 @@
...
+        subwins = [s for s in mdi.subWindowList() if s.isVisible()]
+
+        if not subwins:
+            self._append_placeholder()
+            return
+
+        for index, subwin in enumerate(subwins):
+            self._append_item(index, subwin, subwin is active)
```
j8xixo12: Build the menubar dropped down list and use `activate` to indicate which sub-window is focused by user.

j8xixo12:
@tigercosmos @yungyuc This PR is ready for the next review.

yungyuc (approved):
LGTM

PR #1063, opened 2026-07-12
Support mermaid diagrams in the myst documentation

yungyuc:
Enable sphinxcontrib.mermaid and map a "```mermaid" fenced block to the mermaid directive through MyST. It renders on both GitHub and through Sphinx.

[skip-ci]


doc/source/conf.py
```diff
@@ -56,6 +57,10 @@
     "deflist",
 ]
 
+# Treat a ```mermaid fence as the mermaid directive, so the same fenced
```
yungyuc, line 60: Support mermaid in myst.

PR #1069, opened 2026-07-12
Show Claude.ai subscription usage in the status line

tigercosmos:
Append the Claude.ai 5-hour and weekly subscription usage windows to the status line, after the existing estimated-cost segment and on the same single line, each rendered only when the session JSON exposes that window so accounts without subscription usage degrade cleanly.

[skip-ci]


yungyuc (commented):
LGTM

PR #965, opened 2026-06-28
Add the Agent Draw command schema for the World API

tigercosmos:
This adds `solvcon.pilot.agentdraw`, a prototype schema defining the input and output contract for driving a World from an AI agent, intended as the shared command language for the planned MCP server and tool-calling harness.

The schema is authored as JSON Schema in one place: each op declares its arguments and, symmetrically, the shape of the value it returns, so a front-end can publish both an `inputSchema` and an `outputSchema` from a single source. An Executor applies validated commands to a `World`, captures errors into structured results, and offers opt-in result validation against the declared output contract.

Related to #966

PR #1074, opened 2026-07-12
Agent Draw Command 1: generic command framework

tigercosmos:
Add the Qt-free Agent command framework that domain families (drawing, mesh, pilot control) build on. A `Command` declares its op, category, and JSON Schema for its arguments and returns; a `CommandSet` collects commands and derives the schema, validators, and MCP-style tool definitions from them; a `CommandProcessor` runs validated commands against a target and captures failures as results; and a `CommandDispatcher` routes ops across several families so one session can expose them together. The framework mirrors the existing backend registry and imports no Qt, so it stays headless and unit-testable.

## How a command flows

A command is a plain dict such as `{"op": "add_item", "name": "x"}`, usually proposed by an AI backend against the tool definitions the `CommandSet` derives. The `CommandProcessor` is the single entry point that turns that untrusted dict into a real, recorded change to the target. It validates the dict against the op's JSON Schema, fills omitted optional arguments from their defaults, calls the command's `apply`, and wraps the outcome in a `CommandResult`. It never raises, so one bad command cannot abort a batch.

Responsibilities are split so families stay reusable: validation and the schema live on the stateless `CommandSet`, while the target, the log, and the run policy live on the `CommandProcessor` that is bound to one target. Many targets can therefore share one family definition.

```mermaid
flowchart TD
    A["Command dict (op + arguments)"] --> B["CommandProcessor.run"]
    B --> C{"validate against JSON Schema"}
    C -- invalid --> F["CommandResult: ok=False, error"]
    C -- valid --> D["apply_defaults: fill optional args"]
    D --> E["Command.apply(target, args, ctx)"]
    E -- "returns mapping" --> G["CommandResult: ok=True, value"]
    E -- "raises" --> F
    G --> H["appended to log"]
    F --> H
```

When one session exposes several families at once, a `CommandDispatcher` sits in front and routes each op to the `CommandProcessor` that owns it. Ops stay unprefixed but must be unique across the group, and the tool surface is the concatenation of the members'.

```mermaid
flowchart LR
    I["Command dict"] --> J["CommandDispatcher"]
    J -- "draw op" --> K["CommandProcessor (draw World)"]
    J -- "pilot op" --> L["CommandProcessor (pilot app)"]
    J -- "unknown op" --> M["CommandResult: ok=False"]
```

### DIFF
--- a/solvcon/agent/_command.py
+++ b/solvcon/agent/_command.py
@@ -1,0 +1,361 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Generic command framework for the Agent.
+
+``Command`` subclasses are collected in a ``CommandSet`` that derives the
+JSON Schema, validators, and tool definitions; a ``CommandProcessor`` runs
+them against a target, and a ``CommandDispatcher`` routes ops across
+families. Imports no Qt; concrete commands live in domain packages such as
+:mod:`solvcon.agent.draw`.
+"""
+
+import copy
+import dataclasses
+
+
+class CommandError(ValueError):
+    """A command failed schema validation or could not be applied."""
+
+
+CRUD_CATEGORIES = ("create", "read", "update", "delete", "log")
+
+
+class Command:
+    """One command: op name, JSON Schema, and behavior."""
+
+    op = ""
+    category = ""
+    summary = ""
+    arguments = {}
+    optional = ()
+    returns = {}
+
+    def apply(self, target, args, ctx):
+        """Apply the command to ``target`` and return its result mapping."""
+        raise NotImplementedError
+
+
+def _command_schema(cmd):
+    # ``op`` is a required const; ``category`` is an annotation
+    # validators ignore.
+    required = [name for name in cmd.arguments if name not in cmd.optional]
+    return {
+        "title": cmd.op,
+        "description": cmd.summary,
+        "category": cmd.category,
+        "type": "object",
+        "properties": {"op": {"const": cmd.op}, **cmd.arguments},
+        "required": ["op"] + required,
+        "additionalProperties": False,
+    }
+
+
+def _result_schema(cmd):
+    return {
+        "title": f"{cmd.op}_result",
+        "type": "object",
+        "properties": dict(cmd.returns),
+        "required": list(cmd.returns),
+        "additionalProperties": False,
+    }
+
+
+def _merge_default(default, value):
+    """Recursively overlay ``value`` onto ``default``."""
+    if not (isinstance(default, dict) and isinstance(value, dict)):
+        return value
+    out = {}
+    for name, prop in default.items():
+        out[name] = (_merge_default(prop, value[name]) if name in value
+                     else copy.deepcopy(prop))
+    for name, val in value.items():
+        if name not in default:
+            out[name] = val
+    return out
+
+
+class CommandSet:
+    """A named family of commands with its derived schema and validators."""
+
+    def __init__(self, title, description, categories=CRUD_CATEGORIES):
+        self.title = title  # family name, used as the schema title
+        self.description = description  # human summary of the family
+        self.categories = tuple(categories)  # allowed op categories
+        self.commands = {}
+        self._built = None
+
+    def register(self, command):
+        """Add a command to the set; usable as a class decorator."""
+        cls = command if isinstance(command, type) else type(command)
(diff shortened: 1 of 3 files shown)


solvcon/agent/_command.py
```diff
@@ -0,0 +1,445 @@
...
+    for name, prop in default.items():
+        out[name] = (_merge_default(prop, value[name]) if name in value
+                     else copy.deepcopy(prop))
+    for name, val in value.items():
+        if name not in default:
+            out[name] = val
+    return out
+
+
+class CommandSet:
```
tigercosmos, line 80: The class for making commands.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,445 @@
...
+        for name, prop in command.arguments.items():
+            if (isinstance(prop, dict) and "default" in prop
+                    and name not in command.optional):
+                raise ValueError(
+                    f"{command.op}.{name} is defaulted but required")
+        self.commands[command.op] = cls() if command is cls else command
+        self._built = None
+        return command
+
+    def _build(self):
```
tigercosmos, line 113: Build the command.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,445 @@
...
+            },
+            "validators": {op: jsonschema.Draft202012Validator(schema)
+                           for op, schema in command_schemas.items()},
+            "result_validators": {op: jsonschema.Draft202012Validator(schema)
+                                  for op, schema in result_schemas.items()},
+        }
+
+    @property
+    def _cache(self):
+        # TODO: the lazy build races benignly when two threads hit a cold
```
tigercosmos: TODO.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,445 @@
...
+            if not isinstance(prop, dict) or "default" not in prop:
+                continue
+            default = prop["default"]
+            if name not in out:
+                out[name] = copy.deepcopy(default)
+            else:
+                out[name] = _merge_default(default, out[name])
+        return out
+
+    def tool_definitions(self):
```
tigercosmos, line 204: MCP style tool.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,445 @@
...
+
+def _pos(description):
+    return {**POSITIVE, "description": description}
+
+
+def _int(description):
+    return {**INTEGER, "description": description}
+
+
+class Command:
```
tigercosmos, line 26: Defination of the command.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,361 @@
...
+            return CommandResult(op, False, error=str(exc))
+        except Exception as exc:
+            if self.reraise:
+                raise
+            return CommandResult(
+                op, False, error=f"{type(exc).__name__}: {exc}")
+        return CommandResult(op, True, value=value)
+
+
+class CommandDispatcher:
```
tigercosmos, line 303: Dispatch the command to different processors. Each processor handles a different target.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,361 @@
...
+@dataclasses.dataclass
+class CommandResult:
+    """The outcome of applying one command."""
+    op: str
+    ok: bool
+    value: object = None
+    error: str = None
+
+
+class CommandProcessor:
```
tigercosmos, line 247: The processor will handle the command on a target: validating, filling defaults, running the command, getting result and appending logs.

solvcon/agent/_command.py
```diff
@@ -0,0 +1,361 @@
...
+    """Run a ``CommandSet``'s commands against a ``target``, keeping a log."""
+
+    def __init__(self, target, command_set, validate_results=False,
```
tigercosmos, line 250: A target is the object that will be applied with the commands, e.g. Pilot, Canvas

tigercosmos:

solvcon/agent/_command.py
yungyuc: Could you please also annotate other code involving the MCP support in this PR?  Or is this the only place?

yungyuc (approved):
Looks like a reasonable foundation.

solvcon/agent/_command.py
tigercosmos: This is the only place for now. For MCP full support, we will need to implement a MCP server in the future. For harness engineering, a MCP style tool call is just allowing AI to understand how to use tools.

PR #1075, opened 2026-07-12
Agent Draw Command 2: command framework tests

tigercosmos:
The command framework merged in #1074 as a basic go-through whose docstring deferred the full test suite to a later change. This is that follow-up: test-only, additive, with no change under solvcon/agent.

It covers the branches the go-through leaves untested: the register() guards, the processor's result-schema check, the log and its copy semantics, error capture, invalid-input rejection, schema cache invalidation, apply_defaults immutability, the tool-definition shape, and the base apply contract. All 35 tests pass; lint is clean.

### DIFF
--- a/tests/test_command_framework.py
+++ b/tests/test_command_framework.py
@@ -252,4 +252,182 @@ class CommandDispatcherTC(unittest.TestCase):
             self.dispatcher.run_script([{"op": "nope"}])
 
 
+class BadCount(agent.Command):
+    op = "bad_count"
+    category = "read"
+    summary = "Return a value that violates the result schema."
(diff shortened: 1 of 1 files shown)


tests/test_command_framework.py
```diff
@@ -252,4 +252,182 @@ def test_run_script_across_families(self):
...
+    category = "log"
+    summary = "Record a note in the processor log."
+    arguments = {"text": {"type": "string", "description": "Note text."}}
+
+    def apply(self, target, args, ctx):
+        ctx.append_log(args["text"])
+        return {}
+
+
+class CommandRegistrationTC(unittest.TestCase):
```
tigercosmos, line 276: Add more basic tests.

PR #1076, opened 2026-07-12
Agent Draw Command 3: World command vocabulary

tigercosmos:
Add the Agent Draw command vocabulary for the World as the `solvcon.agent.draw` package: one `Command` subclass per World operation, collected in a single `CommandSet`. The schema documents, validators, and tool definitions are all derived from the registered commands by the generic framework merged in Agent Draw Command 1 and 2 (#1074, #1075), so this package declares only the World-specific fragments and apply behavior. A mesh or pilot family would declare its commands the same way.

The vocabulary spans the CRUD categories the framework understands:

- create: bare primitives (`add_point`, `add_segment`, `add_bezier`) and shapes (`add_line`, `add_triangle`, `add_rectangle`, `add_square`, `add_ellipse`, `add_circle`, `add_bezier_shape`).
- read: `get_shape`, `shape_type_of`, `nshape`, `query_visible`, `describe_state`, and `render_png`.
- update: `translate_shape`.
- delete: `remove_shape` and `clear`.
- log: `log`, a free-text note in the command log.

The result fragments mirror `WorldState` and `WorldShapeState` in `cpp/solvcon/universe/World.hpp`, so the published schema tracks the C++ side. Coordinates are world coordinates in the World's native units, math convention with +Y up. `render_png` renders through a renderer injected on the command context, and returns a base64 PNG rather than raw bytes so a result stays JSON-serializable over any transport; with no renderer configured it fails cleanly instead of raising, and never touches a GUI. By-id commands preflight the id so they fail the one clean way `get_shape` does, rather than leaking a C++-flavored exception. The JSON Schema property builders the fragments need live in the draw package, since the merged framework is domain-agnostic and carries none.

A basic go-through test walks the family once against a real `WorldFp64` through the generic `CommandProcessor`: schema derivation from the commands, a create-read-update-delete flow that mutates the world, clean failure paths for a schema violation and a dead-id operation, the renderer-less `render_png` failure, and the log note. It covers a representative slice only; the comprehensive per-command suite lands in Agent Draw Command 4 (#1077).

Related to #966.

### DIFF
--- a/solvcon/agent/draw/command.py
+++ b/solvcon/agent/draw/command.py
@@ -1,0 +1,463 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The command vocabulary for driving a ``World`` from an AI agent.
+
+Each command is one ``Command`` subclass registered in ``DRAW``; the JSON
+Schema documents and validators are derived from them (see
+:mod:`solvcon.agent._command`). Only the ``World``-specific fragments and
+behavior live here, so a mesh or pilot family declares commands the same way.
+``render_png`` returns a transport-ready image (base64 PNG), never raw bytes,
+so a result is JSON-serializable across any transport.
+
+Coordinates are world coordinates in the ``World``'s native units, math
+convention with +Y pointing up.
+"""
+
+import json
+import base64
+
+from .. import _command as _cmd
+
+
+# Shared JSON Schema fragments, used by reference; treat as immutable.
+NUMBER = {"type": "number"}
+POSITIVE = {"type": "number", "exclusiveMinimum": 0}
+INTEGER = {"type": "integer"}
+POSITIVE_INT = {"type": "integer", "exclusiveMinimum": 0}
+BOOLEAN = {"type": "boolean"}
+STRING = {"type": "string"}
+
+
+def _num(description):
+    return {**NUMBER, "description": description}
+
+
+def _pos(description):
+    return {**POSITIVE, "description": description}
+
+
+def _int(description):
+    return {**INTEGER, "description": description}
+
+
+def _point(description):
+    return {"type": "array", "items": {"type": "number"},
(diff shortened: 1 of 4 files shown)


solvcon/agent/draw/command.py
```diff
@@ -0,0 +1,452 @@
...
+        "Total segments in the world after the add, "
+        "including those owned by shapes.")}
+
+    def apply(self, world, args, ctx):
+        world.add_segment(_world_point(args["p0"]), _world_point(args["p1"]))
+        return {"nsegment": world.nsegment}
+
+
+@_command_set.register
+class AddLine(_cmd.Command):
```
tigercosmos, line 163: Example of a draw command

tests/test_agent_draw_command.py
```diff
@@ -0,0 +1,83 @@
...
+        grouped = draw.commands_by_category()
+        self.assertIn("add_circle", grouped["create"])
+        self.assertIn("translate_shape", grouped["update"])
+        self.assertIn("clear", grouped["delete"])
+        tool = next(t for t in draw.tool_definitions()
+                    if t["name"] == "add_circle")
+        self.assertIn("r", tool["inputSchema"]["properties"])
+        self.assertIn("shape_id", tool["outputSchema"]["properties"])
+
+    def test_crud_flow_mutates_the_world(self):
```
tigercosmos, line 37: Usage example.

solvcon/agent/draw/command.py
```diff
@@ -0,0 +1,452 @@
...
+so a result is JSON-serializable across any transport.
+
+Coordinates are world coordinates in the ``World``'s native units, math
+convention with +Y pointing up.
+"""
+
+import json
+import base64
+
+from .. import _command as _cmd
```
yungyuc, line 22: It's a little bit weird to import a private module from upper level.  It may or may not break encapsulation, but I will keep an eye on this pattern.

No need to change anything.  This is just a feeling.

yungyuc (commented):
LGTM

PR #1077, opened 2026-07-12
Agent Draw Command 4: executor and tests

tigercosmos:
Add the Executor that binds the Agent Draw command family and an injected renderer to a World, so a front-end runs validated commands through one entry point. The Executor subclasses the generic CommandProcessor; render_png reads the renderer from its context and returns a failed result when none is configured, so it never touches a GUI.

Add the comprehensive test suite: the schema documents are checked for well-formedness and CRUD coverage, pure-JSON cases pin what the schema accepts and rejects, and the Executor is exercised end to end against a real World.


solvcon/agent/draw/executor.py
```diff
@@ -0,0 +1,32 @@
...
+The run loop, validation, logging, and result checking are the generic
+``CommandProcessor`` in :mod:`solvcon.agent._command`; this subclass binds the
+Agent Draw command set and the injected renderer.
+"""
+
+from .. import _command as _cmd
+from . import command
+
+
+class Executor(_cmd.CommandProcessor):
```
tigercosmos, line 17: Add Executor.

tests/data/agent_draw_command_cases.json
```diff
@@ -0,0 +1,50 @@
+{
```
tigercosmos, line 1: Increase the JSON schema test coverage.

tests/test_agent_draw_schema.py
```diff
@@ -0,0 +1,328 @@
...
+    def test_script_validation(self):
+        script = [c["command"] for c in _CASES["valid"]]
+        self.assertIs(validate_script(script), script)
+        with self.assertRaises(CommandError):
+            validate_script([{"op": "clear"}, {"op": "fly"}])
+        with self.assertRaises(CommandError):
+            validate_script({"op": "clear"})
+
+
+class ExecutorTC(unittest.TestCase):
```
tigercosmos, line 127: Example of executor with the real world.

solvcon/agent/draw/__init__.py
```diff
@@ -18,6 +19,7 @@
     CommandResult,
 )
 from .command import _command_set
+from .executor import Executor  # noqa: F401
```
yungyuc, line 22: We should move the executor to be working on any GUI operations in  the future.

After that, the concept should extend to any operations.  But this will be longer term.

PR #1078, opened 2026-07-12
Agent Draw Command 5: wire the session, drop the fallback

tigercosmos:
Rebase the first in-tree Agent command-family wiring on current master now that Agent Draw has landed. Replace the obsolete fallback bridge with the Draw executor, but keep the session and backend vocabulary generic so upcoming windowing and other command families can share it. The surface-agnostic multi-controller routing remains a follow-up in #966.

Keep delete commands disabled unless a session explicitly opts in, declare the jsonschema runtime dependency, and update the headless and Pilot integration tests to exercise the real command path. jsonschema is already installed via pip across the CI jobs, and the Agent Draw tests exercise the validator directly, so the dependency is covered without a separate packaging test.


solvcon/agent/_core.py
```diff
@@ -37,52 +37,43 @@ class _OutcomeStub:
...
-            from solvcon.pilot import agentdraw
-            _draw_module = agentdraw
-        except ImportError:
-            from . import _draw
-            _draw_module = _draw
-    return _draw_module
+def _make_executor(world, renderer=None):
+    """Build the current ``World`` command executor."""
+    from . import draw
+    return draw.Executor(world, renderer)
```
tigercosmos, line 43: We will have more executors later.

solvcon/agent/_draw.py
```diff
@@ -1,63 +0,0 @@
-# Copyright (c) 2026, solvcon team <contact@solvcon.net>
```
tigercosmos, line 1: Removed the old fallback draw command.

solvcon/agent/_core.py
```diff
@@ -128,19 +119,31 @@ def scene_context(self, level="basic"):
...
         command), so a bad runner or command never aborts the batch and the
-        results always line up with the commands.  This does not touch the
-        transcript.
+        results always line up with the commands.  Delete commands are rejected
+        before reaching the runner unless this session allows them.  This does
+        not touch the transcript.
         """
         if not commands:
             return []
+        blocked = set() if self.allow_destructive else _destructive_ops()
```
tigercosmos, line 139: only the allowed sessions can call deletors.

solvcon/agent/_core.py
```diff
@@ -37,52 +37,43 @@ class _OutcomeStub:
...
+def _make_executor(world, renderer=None):
+    """Build the current ``World`` command executor."""
+    from . import draw
+    return draw.Executor(world, renderer)
 
 
-def _make_executor(world, renderer=None):
-    """Build the command runner for ``world`` from the resolved draw module."""
-    return _resolve_draw().Executor(world, renderer)
+def tool_surface(allow_destructive=False):
```
tigercosmos, line 46: only the draw tools for now.

yungyuc: Already a big leap.

Issue #1067, opened 2026-07-12
Windows rel build failed `test_SimpleArrayPlex_typed_roundtrip`


yungyuc:
Issue to track the CI run [build_windows-2022_Release](https://github.com/solvcon/solvcon/actions/runs/29196124241/job/86659408956#logs) failure at [`86659408956#step:15:306`](https://github.com/solvcon/solvcon/actions/runs/29196124241/job/86659408956#step:15:306):

```
  ================================== FAILURES ===================================
  __ SimpleArrayPlexTC.test_SimpleArrayPlex_typed_roundtrip (dtype='float32') ___
  
  self = <test_buffer.SimpleArrayPlexTC testMethod=test_SimpleArrayPlex_typed_roundtrip>
  
      def test_SimpleArrayPlex_typed_roundtrip(self):
          for dtype, (_, typed_name, _) in self._DTINFO.items():
              with self.subTest(dtype=dtype):
                  plex = solvcon.SimpleArray((2, 3, 4), dtype=dtype)
                  typed = plex.typed
                  plex2 = typed.plex
      
                  self.assertEqual(str(type(plex)),
                                   "<class '_solvcon.SimpleArray'>")
                  self.assertEqual(str(type(typed)),
                                   "<class '{}'>".format(typed_name))
                  self.assertEqual(str(type(plex2)),
                                   "<class '_solvcon.SimpleArray'>")
      
                  # Verify data survives the roundtrip:
                  # write through typed, read back from plex2
                  ndarr = np.array(plex, copy=False)
                  ndarr.flat[0] = 1
  >               self.assertEqual(plex2[0, 0, 0], typed[0, 0, 0])
  E               AssertionError: nan != nan
  
  ..\..\..\..\tests\test_buffer.py:4430: AssertionError
  =========================== short test summary info ===========================
  SUBFAILED(dtype='float32') ..\..\..\..\tests\test_buffer.py::SimpleArrayPlexTC::test_SimpleArrayPlex_typed_roundtrip - AssertionError: nan != nan
```

tigercosmos:
Working on it.

tigercosmos:
Fixed by #1083

Issue #1065, opened 2026-07-12
Reorganize pilot file structure in C++ and Python


yungyuc:
We have many new features for pilot GUI.  The flat structure needs to be changed to hierarchical for better code organization.

Use two sub-issues: #1112 for C++ and #1113 for Python.

PR #1079, opened 2026-07-12
Add the Qt-free pilot keymap core (step 1 of shortcut)

### LINKED ISSUE #1070: Pilot keyboard-shortcut system
# Pilot keyboard-shortcut system

Develop a keyboard-shortcut system that makes the pilot feel native on Linux, macOS, and Windows: one shared vocabulary of commands, and a phrasebook per platform that spells each command in that platform's accent, so tuning a binding on one platform never disturbs the others.

## Problem

The pilot binds keys by hand, one call at a time, and binds almost nothing. `RManager` wires undo, redo, and the camera reset directly; every other command carries no shortcut at all, whether it moves the camera, picks a draw tool, or operates the pilot itself: opening the console, toggling the agent panel, making a new canvas. Three things follow:

- The bindings are scattered rather than gathered in one place, and they are scattered across two layers: `RManager` sets a few in C++, while most of the pilot's commands are registered by Python feature classes through `PilotFeature.add_action`. There is no one place to see what the pilot binds or to reason about it as a whole.
- Most commands are silent. Operating the pilot is mouse-only today: opening the console (`window.console`), toggling the agent panel (`panel.agent_console`), and making a canvas all sit on the menu with no key behind them. A user who reaches for a key finds nothing bound, and adding one means another ad-hoc edit.
- There is no way to catch two commands that claim the same chord while their shortcut contexts are active together. A collision is found by a user pressing a key, not by a test.

The bindings the pilot does carry are native almost by accident. Undo and redo use `QKeySequence::Undo` and `::Redo`, which Qt already spells in each platform's accent (Command+Z on macOS, Control+Z elsewhere), and reset is plain Escape, which reads the same on every platform. The real exposure is the pilot's own commands, the camera moves and draw tools that no standard sequence covers: written as a plain chord they would use Control everywhere and feel foreign on the Mac. This plan gathers every binding into one vocabulary and lets each platform pronounce the non-standard ones in its own accent, so a curated chord is as native as a standard one, and every platform's table is checked by a test rather than a keystroke.

## Goal

Make the pilot's keys feel native on each platform while keeping one place to reason about them. Concretely:

- **Native accent per platform.** Honor each platform's command modifier (Command on macOS, Control on Windows and Linux) and its standard sequences. Preserve the existing macOS application-menu role on the Quit action (`file.exit`); Preferences and About have no action in the pilot yet, so they get the same treatment if and when it grows them.
- **Covers operating the pilot, not just editing.** The vocabulary spans both layers where commands are born: the C++ commands in `RManager` (undo, redo, camera) and the Python feature commands (console, agent panel, a new canvas, the other panels). A Python command gets its shortcut within this plan: Step 4 adds one small helper that applies a command's binding by its objectName, so a Python-registered action such as `panel.agent_console` gets its native key the same way a C++ one does, with no per-command wiring by hand.
- **Default bindings now; user rebinding is a follow-up.** This plan furnishes the built-in defaults and the machinery to apply them from both C++ and Python. Letting a user pick or rebind keys, a Keyboard Shortcuts view, and saving a keymap to a file are explicitly out of scope here. The vocabulary and lookup leave room for that layer so it can be added later without reshaping the system (see the appendix).
- **A common vocabulary.** The heart of the system is a shared vocabulary of commands and the roles a binding is written against: the `Primary` modifier that each platform pronounces its own way, the shortcut context, and the per-platform tables in which every binding lives. Tuning the Windows accent never touches the macOS accent, and adding a command does not reshape the shared core.
- **Testable at both layers.** Every platform's declarative binding table stays Qt-free and is unit tested on every CI runner. Collisions involving Qt standard keys are checked after resolution on that platform's CI runner, because Qt owns the platform-specific standard-key bindings.

How native the pilot can be differs per platform, and the system does not force one answer globally. It makes the per-platform table the single place that decides, with the curated bindings as the fallback every platform shares.

## Development steps

Each step is a self-contained pull request of roughly 300 changed lines that leaves the pilot working. The architecture sections below detail the pieces each step names.

- [x] **Step 1: the Qt-free core.** Build the keymap core: the command vocabulary, the context, standard-action, menu-role, and modifier-role enums, the platform axis, the per-platform tables seeded from today's bindings (macOS carrying the Quit role), the capability record, and `findDeclaredConflicts`. Compile it into `test_nopython` and cover it with Qt-free cases. No pilot behavior changes. Split by platform table if it overflows the budget. #1079
- [x] **Step 2: the roof.** Add `RShortcutManager`: detect the platform, resolve a command id to `QKeySequence` values (standard-key lookup or `Primary`-to-`Qt::CTRL` chord conversion), apply them via `applyTo`, add the Qt-backed `findResolvedConflicts`, and expose the resolution to Python as portable values. Wire it into the pilot but route no action through it yet. https://github.com/solvcon/solvcon/pull/1103
- [x] **Step 3: adopt the existing bindings.** Route today's undo, redo, and reset shortcuts through the manager instead of the three hardcoded `setShortcut` calls, keeping reset's widget context and its `add3DWidget` attachment. Cover the live bindings with end-to-end Python tests. Behavior is identical, now table-driven. #1114 
- [x] **Step 4: bind the silent commands, both layers.** Give every command that carries no shortcut its platform accent through table entries, guarded by the conflict test. C++ commands bind through `RManager`; one Python helper applies a command's binding by objectName, so `PilotFeature.add_action` and the direct `build_action` callers (console, Quit, panels) all get it. Add a "new 2D canvas" command. Leave the `keyPressEvent` keys alone (console editing, and `RDomainWidget`'s W, A, S, D and arrows). Split by layer if needed. #1115
- [x] **Step 5: tune each platform's accent.** Fill the per-platform reserved-sequence and curated tables, annotate each capability record, and verify every resolved sequence and conflict check on that platform's CI. Preferences and About wait until those menu items exist; the capability record already names them. #1116

## The system

The system is laid out as one structure: a shared foundation, a phrasebook per platform held inside it, and a single roof over all of them.

```mermaid
flowchart TD
    RM["<b>RManager (coordinator, C++)</b><br/>binds the C++ commands (undo, redo, camera)<br/>via applyTo and RMenuModel"]
    PY["<b>Python action builders</b><br/>PilotFeature.add_action and direct build_action callers<br/>ask the roof for each command's binding by id"]
    SM["<b>RShortcutManager (Qt adapter: the roof)</b><br/>detects the running platform<br/>resolves a command id to native sequences<br/>checks standard-key and curated collisions<br/>reachable from C++ and Python<br/>emits shortcutsChanged"]
    CORE["<b>Keymap core (Qt-free foundation)</b><br/>vocabulary: ShortcutCommand, KeyChord<br/>ShortcutContext, StandardAction, MenuRole, KeyMod<br/>ShortcutCapabilities, findDeclaredConflicts()<br/>unit tested in test_nopython on every runner"]
    L["<b>Linux phrasebook</b><br/>Control accent<br/>standard sequences<br/>curated fallback"]
    M["<b>macOS phrasebook</b><br/>Command accent<br/>Quit menu role (Preferences / About later)<br/>curated fallback"]
    W["<b>Windows phrasebook</b><br/>Control accent<br/>standard sequences<br/>curated fallback"]

    RM --> SM
    PY --> SM
    SM --> CORE
    CORE --> L
    CORE --> M
    CORE --> W

    classDef coord fill:#f3f3f4,stroke:#888,color:#111;
    classDef roof fill:#eef3fb,stroke:#3574f0,color:#111;
    classDef book fill:#fbfdff,stroke:#3574f0,color:#111;
    class RM,PY coord;
    class SM,CORE roof;
    class L,M,W book;
```

One foundation, three phrasebooks. The Qt-free core holds the shared vocabulary and every platform's binding table; the roof picks the running platform's phrasebook and turns each command into a real key sequence; the coordinator wires it to the actions on the menu.

### The foundation: the keymap core (Qt-free)

The core holds the vocabulary the pilot binds against and the rules that need no Qt: a `ShortcutCommand` naming each command by id, a `KeyChord` that spells a binding, the `ShortcutContext` (Application, Window, Widget)[^context] and `MenuRole` (none, Quit, Preferences, About)[^menurole] enums, a `StandardAction` enum that names the Qt standard sequences the pilot uses, and a `KeyMod` role set whose `Primary` flag stands for the platform command modifier rather than a physical key. A command id is the action objectName the pilot already uses (`edit.undo`, `window.console`, `panel.agent_console`), so the vocabulary names existing commands across both layers rather than inventing ids. Three more Qt-free pieces make it a system rather than a single table:

- **Per-platform tables.** The lookup carries a platform axis: `bindingFor(platform, id)`, backed by a curated table per platform. A `PlatformId` enum (Linux, Mac, Windows) names the axis; detecting the running platform is a thin Qt call made higher up and passed in, so the tables stay Qt-free.
- **A capability record.** A plain `ShortcutCapabilities` struct states, per platform, what the keymap can honor: whether it moves items into an application menu, which sequences the platform reserves. This is data, so it lives in the core and is tested there.
- **Declared conflict detection.** `findDeclaredConflicts(platform)` is a pure function that rejects duplicate curated chords in contexts that can be active together. The pilot has one main window and its widget shortcuts live below that window, so the default tables use a conservative rule: Application overlaps every context, Window overlaps Window and Widget, and Widget overlaps Widget. The system does not reuse a chord between two widget scopes until the vocabulary can identify and prove those scopes are mutually exclusive. Standard actions remain symbolic in this layer and are checked after Qt resolves them in the roof.

Because it never mentions Qt, all of this compiles into `test_nopython` and is unit tested without a display on every CI runner. Every runner therefore checks the structure and curated chords in all three tables. The Qt-backed suite on Linux, macOS, and Windows separately resolves that runner's standard keys and checks the complete native binding set against the curated chords. This split does not pretend a Linux Qt build can resolve Windows standard keys.

### The phrasebooks: per-platform tables

Unlike a theme system, a keyboard-shortcut system needs almost no native code. Qt's standard sequences already carry the platform accent for the common commands, and Qt maps `Qt::CTRL` to the Command key on macOS, so a `Primary` binding pronounces itself correctly without a platform call. What differs between platforms is therefore data, not machinery: which standard sequence a command uses, which chord it falls back to, which items become application-menu roles, and which sequences the platform reserves. So the per-platform half is a Qt-free table, one per `PlatformId`, compiled everywhere through the core and read by the roof. Editing the Windows accent touches one table and nothing else; the compiler and the conflict test keep that change out of the other platforms.

### The roof: `RShortcutManager`

`RShortcutManager` is the Qt adapter. It detects the running platform and folds it into a `PlatformId`, resolves a command id to its native sequences (using `QKeySequence::keyBindings` when the command names a `StandardAction`, and otherwise converting the platform `KeyChord`, mapping `KeyMod::Primary` to `Qt::CTRL` and each `Key` to its `Qt::Key`), and applies the result to an action through `applyTo`. A standard action uses `QAction::setShortcuts(StandardKey)`; a curated chord uses `setShortcut`. Both paths set the `Qt::ShortcutContext` and menu role. `findResolvedConflicts()` compares every resolved sequence under the same conservative context-overlap rule as the core, so a standard Undo binding can collide neither with another standard action nor with a curated `Primary+Z` chord. Because half the pilot's commands are born in Python, the resolution is also reachable from Python: the manager returns a command's binding as portable values (a `StandardAction` id or a sequence string, plus the context and menu role), which a shared Python helper applies with the corresponding PySide `setShortcuts` or `setShortcut` call. That keeps one vocabulary across both layers without bridging a `QAction` across the pybind and PySide boundary. It exposes `capabilities()` so callers ask what the current platform supports rather than guessing, and emits `shortcutsChanged`, which no caller consumes in Steps 1 through 5; it is scaffolding for the customization the appendix describes.

### The coordinator: `RManager`

`RManager` owns the manager and wires the C++ side to the menu. It replaces the three hardcoded `setShortcut` calls with `applyTo(action, id)` lookups and binds each C++ action by id as it builds the menu. Python mirrors that through one binding helper: `PilotFeature.add_action` uses it for feature actions, and `_Controller.populate_menu` uses it for the Console, Quit, and other actions it constructs directly with `build_action`. `RMenuModel` stays a pure placement layer, unaware of shortcuts; both layers resolve through the one manager. Menu roles are applied before placement, as Qt requires on macOS. The coordinator leans on the capability record for platform-shaped behavior, so the interface never offers a binding the platform cannot honor.

## Per-platform native levers

Each phrasebook pulls the same set of levers, tuned to its platform. Naming them keeps the phrasebooks honest and the capability record grounded.

| Lever | Linux | macOS | Windows |
| --- | --- | --- | --- |
| Command modifier | Control | Command (Qt maps `Qt::CTRL`)[^ctrl] | Control |
| Standard sequences | `QKeySequence::StandardKey`[^standardkey] | `QKeySequence::StandardKey`[^standardkey] | `QKeySequence::StandardKey`[^standardkey] |
| Menu roles | none | Quit into the app menu; Preferences, About when added[^menurole] | none |
| Reserved sequences | curated table | curated table | curated table |

The common commands lean on the standard sequences, which Qt already spells in each platform's accent. Where no standard sequence exists, the curated chord carries the binding, written against `Primary` so it pronounces itself correctly on every platform, and the capability record reports what each platform reserves so the interface can stay clear of it.

## Verification

- **Unit (C++, no GUI).** The `PilotKeymap*` cases in `test_nopython` cover every platform table (each is well-formed, its standard-action and chord entries are consistent, and its contexts and menu roles are correct), the resolution inputs, and `findDeclaredConflicts` returning empty for each platform's default table. Add cross-context cases proving that Application, Window, and Widget bindings are rejected when their activation can overlap. These tests run on every runner, so all three phrasebooks' declarative data and curated chords are checked everywhere.
- **End-to-end (Python).** The `ShortcutTC` cases assert the live `QAction` carries the platform-correct sequences and context, for both a C++ command (undo) and Python actions from both construction paths: a feature action such as `panel.agent_console` and a direct `build_action` action such as `window.console`. Assert that Quit (`file.exit`) has its standard sequence and that the macOS role remains set. Drive them headlessly under `QT_QPA_PLATFORM=offscreen`.
- **Per-platform accent and resolved conflicts.** On each platform's CI runner, resolve every command, compare all `QKeySequence::keyBindings` alternatives, and require `findResolvedConflicts()` to return empty. Confirm Command on macOS, Control elsewhere, and the standard sequences where they apply. Add a negative case in which a curated chord duplicates Undo's primary or alternate native binding and prove that the resolved checker reports it.

## Appendix: risks and open questions

- **Standard sequence coverage.** `QKeySequence::StandardKey` covers the common commands but not the pilot's own (camera moves, draw tools). Those fall to curated chords, and the open question is which chords read as natural on each platform without colliding with a reserved sequence; settled per platform in Step 5. A standard key may expose more than one native binding, so application and conflict checks operate on the full `keyBindings()` list rather than only its primary entry.
- **Reserved sequences.** The set a platform reserves is a moving target and is seeded here as a curated list rather than read from the OS. The list is a judgment call, kept in the capability record so it is data and testable, and revisited if a collision surfaces on real hardware.
- **Keys handled outside the action system.** The Python console and the 3D domain widget each handle keys in their own `keyPressEvent` rather than through actions: the console its editing keys, `RDomainWidget` the W, A, S, D and arrow-key camera moves. Folding these into the vocabulary is out of scope here and left to a later step. Until then the tables must leave those keys alone, since neither conflict checker can see a binding that lives in a `keyPressEvent`.
- **Coordination with the theme system.** The theme system is not in the tree yet, so this system defines its own small `PlatformId` and running-platform detection. If the theme core lands first, converge on a shared axis then.

## Appendix: a future path to custom bindings

The vocabulary already defines what a binding is: a command id, a `KeyChord` or a `StandardAction`, a context, and a menu role. A user-chosen binding is therefore a data extension, not a redesign. The likely path, deferred until the defaults are furnished:

- Add a user-override layer that the roof consults before the per-platform table, so a chosen binding wins over the default while the declared and resolved conflict checks still guard the merged result.
- Surface the bindings in a Keyboard Shortcuts view, listing each command with its resolved sequence and letting the user rebind it, gated by the capability record like everything else.
- Persist the chosen bindings across sessions, and optionally load a keymap from a file at startup during development, so tuning becomes a save-and-restart loop instead of a rebuild.

[^standardkey]: `QKeySequence::StandardKey` maps a standard action to the platform-native sequence (for example Undo to Command+Z on macOS and Control+Z elsewhere). <https://doc.qt.io/qt-6/qkeysequence.html>

[^ctrl]: Qt spells modifiers portably: `Qt::CTRL` maps to the Command key on macOS by default, and `Qt::META` maps to Control there, so a binding written against the command modifier pronounces itself correctly on every platform. <https://doc.qt.io/qt-6/qt.html#KeyboardModifier-enum>

[^menurole]: `QAction::MenuRole` moves Quit, Preferences, and About into the macOS application menu with their native placement. <https://doc.qt.io/qt-6/qaction.html#MenuRole-enum>

[^context]: `Qt::ShortcutContext` controls whether a shortcut fires application-wide, per window, or only while a widget has focus. <https://doc.qt.io/qt-6/qt.html#ShortcutContext-enum>

tigercosmos:
Introduce the Qt-free keymap core, the foundation of the pilot keyboard-shortcut system: a shared command vocabulary, per-platform binding tables seeded from the pilot's current keys (each binding a KeySpelling sum type of a Qt StandardAction, a curated KeyChord, or Unbound), capability records, and the findDeclaredConflicts rule, compiled into test_nopython and covered by PilotKeymap* cases on every CI runner, with no pilot behavior change yet.

Related to #1070.

### DIFF
--- a/cpp/solvcon/pilot/keymap.cpp
+++ b/cpp/solvcon/pilot/keymap.cpp
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/pilot/keymap.hpp>
+
+#include <cstddef>
+#include <stdexcept>
+#include <variant>
+
+namespace solvcon
+{
+
+namespace
+{
+
+std::vector<ShortcutBinding> tableWith(std::vector<ShortcutBinding> extra)
+{
+    std::vector<ShortcutBinding> rows = {
+        {ShortcutCommand::Undo, StandardAction::Undo, ShortcutContext::Window},
+        {ShortcutCommand::Redo, StandardAction::Redo, ShortcutContext::Window},
+        {ShortcutCommand::CameraReset, KeyChord{KeyMod::None, Key::Escape}, ShortcutContext::Widget},
+    };
+    rows.insert(rows.end(), extra.begin(), extra.end());
+    return rows;
+}
+
+std::vector<ShortcutBinding> const & linuxTable()
+{
+    static std::vector<ShortcutBinding> const table = tableWith({});
+    return table;
+}
+
+std::vector<ShortcutBinding> const & windowsTable()
+{
+    static std::vector<ShortcutBinding> const table = tableWith({});
+    return table;
+}
+
+std::vector<ShortcutBinding> const & macTable()
+{
+    // macOS carries the Quit role from the start; the key sequence for Quit
+    // is added when file.exit begins routing through the manager.
+    static std::vector<ShortcutBinding> const table =
+        tableWith({{ShortcutCommand::Exit, Unbound{}, ShortcutContext::Application, MenuRole::Quit}});
+    return table;
+}
+
+} /* end namespace */
+
+std::string_view commandId(ShortcutCommand command)
+{
+    switch (command)
+    {
+    case ShortcutCommand::Undo:
+        return "edit.undo";
+    case ShortcutCommand::Redo:
+        return "edit.redo";
+    case ShortcutCommand::CameraReset:
+        return "camera.reset";
+    case ShortcutCommand::Exit:
+        return "file.exit";
+    case ShortcutCommand::Console:
+        return "window.console";
+    case ShortcutCommand::AgentPanel:
+        return "panel.agent_console";
+    }
+    throw std::logic_error("Unexpected command");
+}
+
+std::vector<ShortcutBinding> const & bindingTable(PlatformId platform)
+{
+    switch (platform)
+    {
+    case PlatformId::Mac:
+        return macTable();
+    case PlatformId::Windows:
+        return windowsTable();
+    case PlatformId::Linux:
+        return linuxTable();
+    }
+    throw std::logic_error("Unexpected platform");
+}
+
+ShortcutBinding const * bindingFor(PlatformId platform, ShortcutCommand command)
+{
+    for (auto const & binding : bindingTable(platform))
+    {
+        if (binding.command == command)
+        {
+            return &binding;
+        }
+    }
+    return nullptr;
+}
+
+ShortcutCapabilities capabilitiesFor(PlatformId platform)
+{
+    bool const appMenu = platform == PlatformId::Mac;
+    return {platform, appMenu, {}};
+}
+
+bool contextsOverlap(ShortcutContext lhs, ShortcutContext rhs)
+{
+    auto covers = [](ShortcutContext wide, ShortcutContext narrow)
+    {
+        switch (wide)
+        {
+        case ShortcutContext::Application:
+            return true;
+        case ShortcutContext::Window:
+            return narrow == ShortcutContext::Window || narrow == ShortcutContext::Widget;
+        case ShortcutContext::Widget:
+            return narrow == ShortcutContext::Widget;
+        }
+        return false;
+    };
+    return covers(lhs, rhs) || covers(rhs, lhs);
+}
+
+std::vector<ShortcutConflict> findDeclaredConflicts(PlatformId platform)
+{
+    std::vector<ShortcutConflict> conflicts;
+    auto const & table = bindingTable(platform);
+    for (std::size_t i = 0; i < table.size(); ++i)
+    {
+        auto const * chord_i = std::get_if<KeyChord>(&table[i].key);
+        if (chord_i == nullptr)
+        {
+            continue;
+        }
+        for (std::size_t j = i + 1; j < table.size(); ++j)
+        {
+            auto const * chord_j = std::get_if<KeyChord>(&table[j].key);
+            if (chord_j == nullptr)
+            {
+                continue;
+            }
+            if (*chord_i == *chord_j && contextsOverlap(table[i].context, table[j].context))
+            {
+                conflicts.push_back({table[i].command, table[j].command});
+            }
+        }
+    }
+    return conflicts;
+}
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/pilot/keymap.hpp
+++ b/cpp/solvcon/pilot/keymap.hpp
@@ -1,0 +1,149 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Qt-free core of the pilot keyboard-shortcut system: the shared command
+ * vocabulary, the per-platform binding tables, the capability records, and
+ * the declared-conflict rule. Kept free of Qt so every CI runner can unit
+ * test all three platforms' tables without a display. The Qt adapter that
+ * resolves these into live QKeySequence values lives in RShortcutManager.
+ *
+ * @ingroup group_domain
+ */
+
+#include <string_view>
+#include <variant>
+#include <vector>
+
+namespace solvcon
+{
+
+enum class PlatformId
+{
+    Linux,
+    Mac,
+    Windows
+};
+
+enum class ShortcutContext
+{
+    Application,
+    Window,
+    Widget
+};
+
+enum class MenuRole
+{
+    None,
+    Quit,
+    Preferences,
+    About
+};
+
+enum class StandardAction
+{
+    Undo,
+    Redo
+};
+
+/// A modifier role, written against the command modifier rather than a
+/// physical key: Primary is Command on macOS and Control elsewhere. A flag
+/// set so a chord can name more than one modifier; the roof adds the bitwise
+/// combine and test helpers when it resolves a chord into Qt modifiers.
+enum class KeyMod : unsigned
+{
+    None = 0,
+    Primary = 1u << 0,
+    Shift = 1u << 1,
+    Alt = 1u << 2
+};
+
+/// A physical key a curated chord names. The vocabulary stays small: only the
+/// keys the pilot binds today, growing as later steps add curated chords.
+/// Arrow keys are deliberately absent; they belong to
+/// RDomainWidget::keyPressEvent, not the action system.
+enum class Key
+{
+    Escape
+};
+
+struct KeyChord
+{
+    KeyMod mods;
+    Key key;
+};
+
+constexpr bool operator==(KeyChord lhs, KeyChord rhs)
+{
+    return lhs.mods == rhs.mods && lhs.key == rhs.key;
+}
+
+struct Unbound
+{
+};
+
+/// A key spelling is either a standard action, a curated chord, or unbound.
+/// Unbound is a placeholder for a command that has no binding on a platform yet.
+/// StandardAction is a placeholder for a command that defers to Qt's standard sequence for that action.
+/// KeyChord is a curated chord that the pilot binds to a command.
+using KeySpelling = std::variant<Unbound, StandardAction, KeyChord>;
+
+/// A command named by the objectName the pilot already gives its action, so
+/// the vocabulary names existing commands across both the C++ and Python
+/// layers rather than inventing ids.
+enum class ShortcutCommand
+{
+    Undo,
+    Redo,
+    CameraReset,
+    Exit,
+    Console,
+    AgentPanel
+};
+
+struct ShortcutBinding
+{
+    ShortcutCommand command;
+    KeySpelling key;
+    ShortcutContext context;
+    MenuRole role = MenuRole::None;
+};
+
+struct ShortcutCapabilities
+{
+    PlatformId platform;
+    bool movesItemsToApplicationMenu;
+    std::vector<KeyChord> reservedSequences;
+};
+
+struct ShortcutConflict
+{
+    ShortcutCommand first;
+    ShortcutCommand second;
+};
+
+std::string_view commandId(ShortcutCommand command);
+
+std::vector<ShortcutBinding> const & bindingTable(PlatformId platform);
+
+/// The binding for @p command on @p platform, or nullptr when the command
+/// carries none on that platform yet.
+ShortcutBinding const * bindingFor(PlatformId platform, ShortcutCommand command);
+
+ShortcutCapabilities capabilitiesFor(PlatformId platform);
+
+/// Whether two contexts can be active together, under the conservative rule:
+/// Application overlaps every context, Window overlaps Window and Widget, and
+/// Widget overlaps Widget. Symmetric.
+bool contextsOverlap(ShortcutContext lhs, ShortcutContext rhs);
+
+std::vector<ShortcutConflict> findDeclaredConflicts(PlatformId platform);
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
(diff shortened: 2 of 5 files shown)


cpp/solvcon/pilot/keymap.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+namespace solvcon
+{
+
+namespace
+{
+
+std::vector<ShortcutBinding> tableWith(std::vector<ShortcutBinding> extra)
+{
+    std::vector<ShortcutBinding> rows = {
```
tigercosmos, line 20: We will have more bindings.

cpp/solvcon/pilot/keymap.cpp
```diff
@@ -0,0 +1,151 @@
...
+    return nullptr;
+}
+
+ShortcutCapabilities capabilitiesFor(PlatformId platform)
+{
+    bool const appMenu = platform == PlatformId::Mac;
+    return {platform, appMenu, {}};
+}
+
+bool contextsOverlap(ShortcutContext lhs, ShortcutContext rhs)
```
tigercosmos, line 104: Deal with the overlap.

cpp/solvcon/pilot/keymap.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+struct Unbound
+{
+};
+
+/// A key spelling is either a standard action, a curated chord, or unbound.
+/// Unbound is a placeholder for a command that has no binding on a platform yet.
+/// StandardAction is a placeholder for a command that defers to Qt's standard sequence for that action.
+/// KeyChord is a curated chord that the pilot binds to a command.
+using KeySpelling = std::variant<Unbound, StandardAction, KeyChord>;
```
tigercosmos, line 94: Three types of the key spelling.

gtests/test_nopython_pilot_keymap.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    auto const * exit = solvcon::bindingFor(solvcon::PlatformId::Mac, solvcon::ShortcutCommand::Exit);
+    ASSERT_NE(exit, nullptr);
+    EXPECT_TRUE(std::holds_alternative<solvcon::Unbound>(exit->key));
+    EXPECT_EQ(exit->role, solvcon::MenuRole::Quit);
+}
+
+TEST(PilotKeymapCapabilities, OnlyMacMovesItemsToApplicationMenu)
+{
+    EXPECT_TRUE(solvcon::capabilitiesFor(solvcon::PlatformId::Mac).movesItemsToApplicationMenu);
```
tigercosmos, line 56: Only mac has the application menu.

tigercosmos (commented):

yungyuc:
Please help fix the CI issue #1067 first.  It is biting this PR too.

cpp/solvcon/pilot/keymap.hpp
yungyuc, line 1: There is not a class named `RKeyMap`.  The file names `RKeyMap.[hc]pp` are misleading.  Rename them.

CC made this kind of naming mistakes a couple of times.  It is hard to detect it during planning phase, but when there is a PR the issue is obvious.  It will be better to create PRs after you fully review the full commit streak.

gtests/test_nopython_pilot_keymap.cpp
yungyuc: Good test.

yungyuc (changes requested):
- [x] The file names `RKeyMap.[hc]pp` are wrong and should be renamed because there is not a class named `RKeyMap`.

tigercosmos (commented):
@yungyuc  the file is renamed. Please take a look. Thanks

yungyuc (approved):
We will need to work out issue #1065 to organize the files soon.

yungyuc:
@tigercosmos Please merge once master CI cools down.

PR #1080, opened 2026-07-12
Inspector Label 3: Add advanced geometric labels to the 2D canvas overlay

tigercosmos:
Add an advanced label mode that restates each shape's stored geometry (line endpoints and length, triangle vertices and sides, rectangle span and width/height, circle center and radius, ellipse center and semi-axes, Bezier endpoints), backed by new World shape_segment and shape_curve accessors, and wire the inspector and save-dialog mode selector to it. Stacked on #1073.

Related to https://github.com/solvcon/solvcon/issues/896

PR #1081, opened 2026-07-12
Inspector Label 4: Add grid coordinate labels to the 2D canvas overlay

tigercosmos:
Add the final overlay member, world-coordinate labels along the grid, factoring the grid-spacing math out of paint_chrome into shared helpers (grid_spacing_world and for_each_grid_line) so the labels land on the same grid lines and never drift. Labels format at the grid's own precision and name the two axes.

Placement keeps the set readable at any zoom: every label, the grid numerals and the two axis letters alike, is drawn only where it stays within the canvas and clear of the labels already placed, so none is clipped at an edge or painted over another. This also fixes the "y" axis letter, which previously sat too close to the top edge and clipped out of view entirely.

Coordinate labels are their own switch, in the inspector's Labels section and in the save dialog, independent of the shape-label switch and its normal/advanced mode, so either set can be shown alone or together.

Stacked on #1080.

Related to https://github.com/solvcon/solvcon/issues/896

image

### DIFF
--- a/.claude/skills/cpp-style-review/SKILL.md
+++ b/.claude/skills/cpp-style-review/SKILL.md
@@ -61,6 +61,12 @@ it briefly but don't re-implement the check here.
 
 **Comments**
 - **Comments are very important.** Check all comments in the diff for clarity, accuracy, and relevance. Flag any comment that is unclear, misleading, trivial, or outdated.
+- **Doxygen markers (check every changed comment, not just its content).**
+  Per STYLE.md, use `/** ... */` for any block of two or more lines, `///`
+  for a one-line brief on a declaration, and `///<` for a trailing member or
+  enumerator brief. Flag a multi-line `///` block (two or more consecutive
+  `///` lines forming one brief) and tell the author to convert it to
+  `/** */`. Judging a comment's wording is not enough; verify the marker too.
 - Refer to "C++ Comment" in STYLE.md for what counts as a comment and how to judge it.
 
 **pybind11**
--- a/cpp/solvcon/pilot/RWorldRenderer2d.cpp
+++ b/cpp/solvcon/pilot/RWorldRenderer2d.cpp
@@ -40,17 +40,15 @@ QColor const OVERLAY_HIGHLIGHT(240, 180, 60);
 constexpr double BASE_GRID_SPACING_PX = 64.0;
 constexpr double MIN_GRID_SPACING_PX = 16.0;
 constexpr double MAX_GRID_SPACING_PX = 256.0;
+constexpr double COORD_TICK_PX = 6.0;
 
 // Cosmetic (zoom-independent) screen widths for world geometry.
 constexpr double GEOMETRY_LINE_WIDTH_PX = 1.5;
 constexpr int GEOMETRY_POINT_WIDTH_PX = 5;
 
-void paint_chrome(QPainter & painter, ViewTransform2dFp64 const & view, int width, int height)
+/// Return a grid spacing in world units that maps to a screen spacing in pixels.
+double grid_spacing_world(ViewTransform2dFp64 const & view)
 {
-    double const widget_w = static_cast<double>(width);
-    double const widget_h = static_cast<double>(height);
-
-    // Snap grid spacing to a power of ten times {1, 2, 5} for a readable band.
     double const target_world = BASE_GRID_SPACING_PX / view.zoom();
     double const exponent = std::floor(std::log10(target_world));
     double const base = std::pow(10.0, exponent);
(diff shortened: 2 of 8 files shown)


cpp/solvcon/pilot/RWorldRenderer2d.cpp
```diff
@@ -61,50 +59,193 @@ void paint_chrome(QPainter & painter, ViewTransform2dFp64 const & view, int widt
...
+
+/// Format a grid coordinate at the grid's own precision, snapping away floating-point noise.
+QString format_coordinate(double value, double spacing_world, int decimals)
+{
+    double const snapped = std::round(value / spacing_world) * spacing_world;
+    return QString::number(snapped, 'f', decimals);
+}
+
+/// Label the visible grid lines with their world coordinates and name the two axes.
+void paint_coordinate_labels(QPainter & painter, ViewTransform2dFp64 const & view, int width, int height)
```
tigercosmos: Core logic for coordinate.

solvcon/pilot/_canvas_gui.py
```diff
@@ -77,10 +79,12 @@ def _build_label_controls(self):
         group.addButton(self._normal_radio)
         group.addButton(self._advanced_radio)
         self._label_group = group
+        self._coords_check = QtWidgets.QCheckBox("Include coordinates")
```
tigercosmos, line 82: Add a checkbox for coordinate. We shall beautify the UI later.

tigercosmos (commented):

cpp/solvcon/pilot/RWorldRenderer2d.cpp
```diff
@@ -61,50 +59,193 @@ void paint_chrome(QPainter & painter, ViewTransform2dFp64 const & view, int widt
...
+    QFontMetricsF const metrics(painter.font());
+    int const decimals = coordinate_decimals(spacing_world);
+    double const ascent = metrics.ascent();
+    double const text_h = metrics.height();
+    constexpr double PAD = 8.0;
+
+    // Every label placed so far, x row and y column and the two axis letters
+    // alike, so none is drawn over another or clipped past the canvas edge.
+    std::vector<QRectF> placed;
+    auto fits = [&](QRectF const & r)
```
yungyuc: The use of many lambdas is a sign that the helper function should be reimplemented as a helper class.

The change can be done with this PR or later (tracking with a new issue).

yungyuc (approved):
LGTM.

A side note.  The ruler should use a toggle for on/off.  Maybe it's there but I do not know how to find it.  UX issue to be considered later.

cpp/solvcon/pilot/RWorldRenderer2d.cpp
tigercosmos: @yungyuc Sure, let me fix the lambda issue.
Regarding the "ruler", I think you are mentioning the "coordinate" in this case. I can make it to a toggle switch instead of a checkbox.
Just to clarify, there is no real "ruler" for now, which supposes to be a tool to measure the distance of two points.

yungyuc: By ruler I meant the ticks around the window.  Rulers of o measure distance are a different thing.

I was just recording what I saw to align before we improve the UIUX later.

cpp/solvcon/pilot/RWorldRenderer2d.cpp
```diff
@@ -108,6 +160,178 @@ void paint_chrome(QPainter & painter, ViewTransform2dFp64 const & view, int widt
...
+
+/**
+ * Labels the visible grid lines with their world coordinates and names the two
+ * axes. Numerals land on the same lines as paint_chrome's grid; each label,
+ * the numerals and the two axis letters alike, is drawn only where it stays on
+ * the canvas and clear of the labels already placed, so the set stays readable
+ * at any zoom. The per-frame state the placement steps share (metrics, the
+ * placed rectangles) lives in members rather than a web of captured lambdas.
+ */
+class CoordinateLabelPainter
```
tigercosmos, line 189: New helper class.

solvcon/pilot/_tree_panel.py
```diff
@@ -334,6 +335,61 @@ def set_expanded(self, expanded):
         self._toggle.setChecked(expanded)
 
 
+class _ToggleSwitch(QAbstractButton):
```
tigercosmos, line 338: Add a new toggle.

.claude/skills/cpp-style-review/SKILL.md
```diff
@@ -61,6 +61,12 @@ it briefly but don't re-implement the check here.
 
 **Comments**
 - **Comments are very important.** Check all comments in the diff for clarity, accuracy, and relevance. Flag any comment that is unclear, misleading, trivial, or outdated.
+- **Doxygen markers (check every changed comment, not just its content).**
```
tigercosmos, line 64: A piggyback skill update.

cpp/solvcon/pilot/RWorldRenderer2d.cpp
```diff
@@ -108,6 +160,178 @@ void paint_chrome(QPainter & painter, ViewTransform2dFp64 const & view, int widt
...
+    double m_widget_w;
+    double m_widget_h;
+    double m_spacing_world;
+    double m_spacing_px;
+    QFontMetricsF m_metrics;
+    int m_decimals;
+    double m_ascent;
+    double m_text_h;
+    std::vector<QRectF> m_placed;
+};
```
yungyuc, line 333: There should be a class ending mark.

tigercosmos: Sure, let me fix it.

PR #1083, opened 2026-07-13
Fix SimpleArrayPlex typed roundtrip test

tigercosmos:
Related to #1067.

This changes tests/test_buffer.py so the roundtrip check no longer depends on copying uninitialized buffer contents through typed and plex conversions.

What changed:
- Initialize the source array through the NumPy view before taking the roundtrip copies.
- Assert the copied views preserve the initialized value.

Why:
- The previous test mutated the source buffer after the copies were already made. That exposed uninitialized bytes in the copies, which could surface as nan on Windows for float32.

Checks:
- python3 -m py_compile tests/test_buffer.py
- make flake8
- make checkascii
- make checktws
- git diff --check
- Windows verification on windev: targeted test, full SimpleArrayPlexTC, and full tests/test_buffer.py

### DIFF
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -4413,6 +4413,8 @@ class SimpleArrayPlexTC(unittest.TestCase):
         for dtype, (_, typed_name, _) in self._DTINFO.items():
             with self.subTest(dtype=dtype):
                 plex = solvcon.SimpleArray((2, 3, 4), dtype=dtype)
+                ndarr = np.array(plex, dtype=dtype, copy=False)
+                ndarr.flat[0] = 1
                 typed = plex.typed
                 plex2 = typed.plex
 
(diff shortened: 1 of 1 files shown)


tests/test_buffer.py
```diff
@@ -4423,11 +4425,10 @@ def test_SimpleArrayPlex_typed_roundtrip(self):
                 self.assertEqual(str(type(plex2)),
                                  "<class '_solvcon.SimpleArray'>")
 
-                # Verify data survives the roundtrip:
-                # write through typed, read back from plex2
-                ndarr = np.array(plex, copy=False)
```
yungyuc, line 4428: Is the missing dtype the culprit?

tigercosmos: Also possible. I will verify it.

tigercosmos: @yungyuc  Try on Windows, confirmed that `dtype` is not the culprit.

yungyuc: Then it is not clear what is the root cause.  Please clarify.

tigercosmos: Root cause: the old test asserted on uninitialized memory. `.typed` and `.plex` deep-copy the buffer (the copy ctor clones it), so writing `ndarr.flat[0] = 1` into plex after `typed` and `plex2` were cloned never reached them. The final `assertEqual` compared two clones of the same uninitialized bytes: identical on Linux/macOS, but on Windows those bytes formed a `NaN` for float32 and `NaN != NaN` failed. The missing dtype was not the cause; the fix seeds plex before the round-trip and asserts each copy carries the value.

yungyuc: I see.  `.typed` and `.plex` are hard to read.  It'll be better to spell out the full code:

```python
                typed = plex.typed
                plex2 = typed.plex
```

PR #1084, opened 2026-07-13
Build the Qt-free theme foundation for the pilot (step 1 of theme)

### LINKED ISSUE #1062: Pilot theme system

## Problem

## Goal

## Development steps

----

yungyuc:
Step 1 of issue #1062.

The pilot rides whatever style and palette each platform hands it, with no way to choose a light or dark look and no designated place for a theme to live. Laying the groundwork for a native-feeling theme system starts with a foundation that needs no display, so it can be tuned and tested on every platform at once.

Add the theme model: the color-role vocabulary (ThemePalette and the console's SyntaxColors), the ThemeMode and ThemeVariant enums, the rule that resolves a requested mode against the operating system color scheme, a PlatformId axis so each platform's colors live in their own table, and a ThemeCapabilities record stating what each platform's theme can honor. Every platform table is seeded from one curated set for now; the per-platform rooms tune their own copies later.

Because none of this mentions Qt, it compiles into the no-GUI test target and every platform's tables are checked on every runner. Cover the resolution rule, the mode identifiers, the per-platform palette and syntax lookups, and the capability record with test_nopython cases. No pilot behavior changes yet: the core is compiled but not yet consumed.

### DIFF
--- a/cpp/solvcon/pilot/theme.hpp
+++ b/cpp/solvcon/pilot/theme.hpp
@@ -1,0 +1,206 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Qt-free theme foundation: the color-role vocabulary, the light and dark
+ * color tables carried on a per-platform axis, the rule that resolves a
+ * requested mode against the operating system color scheme, and the record of
+ * what each platform's theme can honor.
+ *
+ * Nothing here mentions Qt, so the whole foundation compiles into the
+ * no-GUI test target and every platform's tables are checked on every CI
+ * runner. The Qt adapter (RThemeManager) turns a table into a QPalette by a
+ * straight field-by-field copy, and the per-platform backends read the
+ * capability record to decide what native levers to pull.
+ *
+ * @ingroup group_domain
+ */
+
+#include <cstdint>
+
+namespace solvcon
+{
+
+/**
+ * @brief The source a theme draws its variant from.
+ *
+ * System follows the operating system color scheme and tracks changes to it;
+ * Light and Dark pin the palette regardless of the operating system.
+ */
+enum class ThemeMode
+{
+    System,
+    Light,
+    Dark,
+};
+
+/// The two concrete palettes a mode resolves to.
+enum class ThemeVariant
+{
+    Light,
+    Dark,
+};
+
+/**
+ * @brief The platform a color table and a capability record are written for.
+ *
+ * The palette lookup carries this axis so each platform's look is tuned in its
+ * own table without disturbing the others. Detecting the running platform is a
+ * thin Qt call made in the adapter and passed in here, which keeps this
+ * foundation free of Qt.
+ */
+enum class PlatformId
+{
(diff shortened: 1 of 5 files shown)


cpp/solvcon/pilot/theme.hpp
```diff
@@ -0,0 +1,206 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Qt-free theme foundation: the color-role vocabulary, the light and dark
```
yungyuc, line 10: Add helper functions and data structure for pilot GUI theme.

cpp/solvcon/pilot/theme.hpp
```diff
@@ -0,0 +1,206 @@
...
+ * Light and Dark pin the palette regardless of the operating system.
+ */
+enum class ThemeMode
+{
+    System,
+    Light,
+    Dark,
+};
+
+/// The two concrete palettes a mode resolves to.
```
yungyuc, line 42: We will extend in the future, after the basic theme system works.

cpp/solvcon/pilot/theme.hpp
```diff
@@ -0,0 +1,206 @@
...
+    bool can_force_variant = false;
+    /// Supplies a platform accent color read from the operating system.
+    bool has_native_accent = false;
+    /// Can theme the window's title bar, not just its client area.
+    bool controls_titlebar = false;
+    /// Installs a platform-native widget style rather than a portable one.
+    bool has_native_style = false;
+};
+
+/// The curated light palette, the shared fallback every platform table seeds
```
yungyuc, line 160: Free helper functions.

gtests/test_nopython_pilot_theme.cpp
```diff
@@ -0,0 +1,151 @@
...
+    EXPECT_EQ(std::string("system"), themeModeId(ThemeMode::System));
+    EXPECT_EQ(std::string("light"), themeModeId(ThemeMode::Light));
+    EXPECT_EQ(std::string("dark"), themeModeId(ThemeMode::Dark));
+
+    EXPECT_EQ(themeModeFromId("system"), ThemeMode::System);
+    EXPECT_EQ(themeModeFromId("light"), ThemeMode::Light);
+    EXPECT_EQ(themeModeFromId("dark"), ThemeMode::Dark);
+}
+
+TEST(PilotThemeId, UnknownIdFallsBackToSystem)
```
yungyuc, line 53: Fall back to system.

cpp/solvcon/pilot/theme.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * @brief The colors a palette assigns to the widget roles the pilot uses.
+ *
+ * Each field is named after the QPalette::ColorRole it maps to, so the Qt
+ * adapter copies the struct into a QPalette one field at a time. The trailing
+ * disabled_* fields feed the QPalette::Disabled color group, which keeps
+ * greyed-out text legible in both variants.
+ */
+struct ThemePalette
```
yungyuc, line 94: Color palette.

Issue #884, opened 2026-06-11
Flaky test: SimpleArrayPlexTC.test_SimpleArrayPlex_typed_roundtrip (nan != nan) on Windows Release


tigercosmos:
## Summary

`tests/test_buffer.py::SimpleArrayPlexTC::test_SimpleArrayPlex_typed_roundtrip` flakes on the **Windows-2022 Release** CI job (`run_pilot_pytest`) with:

```
E   AssertionError: nan != nan
tests/test_buffer.py:4100: AssertionError
SUBFAILED(dtype='float32') ... test_SimpleArrayPlex_typed_roundtrip
```

It is intermittent and platform-specific: Windows **Debug** passes, and macOS / Ubuntu (Release and Debug) pass. On one PR branch I saw it fail, pass, then fail across three identical CI runs (no source change between them), so it is roughly a coin flip on Windows Release. Because `run_pilot_pytest` uses `-x`, this one failure stops the whole Windows-Release suite.

## Root cause

The test builds an **uninitialized** array and then checks that two views of the same buffer agree:

```python
plex = modmesh.SimpleArray((2, 3, 4), dtype=dtype)   # not initialized
typed = plex.typed
plex2 = typed.plex
...
ndarr = np.array(plex, copy=False)   # intended to be a view
ndarr.flat[0] = 1                    # seed element [0,0,0]
self.assertEqual(plex2[0, 0, 0], typed[0, 0, 0])
```

Two things combine to make it flaky:

1. `np.array(plex, copy=False)` does not *guarantee* a view — `copy=False` means "avoid a copy if possible", not "never copy". When it copies (apparently the case on Windows Release), the `ndarr.flat[0] = 1` seed never reaches the shared buffer, so `[0,0,0]` stays uninitialized.
2. Uninitialized `float32` memory is sometimes a NaN bit pattern. The assertion then compares `nan == nan`, which is always false by IEEE-754 — so it fails even though `plex2` and `typed` read the *same* bytes. The roundtrip the test means to verify is actually fine; the test is just reading garbage and comparing NaN to NaN.

This is purely a test-robustness problem; the buffer/view code is not implicated.

## Suggested fix

Make the element deterministic before comparing, and don't depend on `copy=False` being a view. For example, seed through the SimpleArray API (or `np.asarray` after asserting a view), assert against the written value rather than view-vs-view, and use a NaN-safe comparison. e.g.:

```python
[...]

PR #1086, opened 2026-07-13
Build SimpleArrayPlex roundtrip test source without a copy=False view

tigercosmos:
master's 80de6cc already fixes the Windows Release nan != nan flake, but it seeds the source through np.array(plex, copy=False), which only avoids a copy when possible and is not guaranteed to be a view. This builds the source with np.full instead, so the roundtrip compares deterministic data without relying on view semantics.

Related to #884.

### DIFF
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -4410,11 +4410,19 @@ class SimpleArrayPlexTC(unittest.TestCase):
                     wrong_cls(array=ndarr)
 
     def test_SimpleArrayPlex_typed_roundtrip(self):
-        for dtype, (_, typed_name, _) in self._DTINFO.items():
+        # Seed values for each dtype to fill the SimpleArray with, so that
+        # we can check that the data survives the roundtrip.
+        seed = {
+            "bool": True,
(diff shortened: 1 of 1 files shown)


tigercosmos:
@yungyuc  Please take a look. Thanks. Honestly, this is just a nice to have improvement.

tests/test_buffer.py
```diff
@@ -4410,11 +4410,19 @@ def test_SimpleArrayPlex_cast_dtype_mismatch(self):
                     wrong_cls(array=ndarr)
 
     def test_SimpleArrayPlex_typed_roundtrip(self):
-        for dtype, (_, typed_name, _) in self._DTINFO.items():
+        # Seed values for each dtype to fill the SimpleArray with, so that
+        # we can check that the data survives the roundtrip.
+        seed = {
```
yungyuc, line 4415: I think this is a sentinel, but seed works too.

yungyuc (approved):
LGTM

yungyuc:
Subsequent devbuild run cancels the previous one for workflow run with commit.  Before we fix this, merge after the previous run completes.

PR #1088, opened 2026-07-13
Furnish the per-platform theme rooms for macOS, Windows, and Linux (steps 5, 6, 7 of theme)

### LINKED ISSUE #1062: Pilot theme system

yungyuc:
Steps 5, 6, 7 of issue #1062.

Furnish all three platforms of mac, linux, and windows, the placeholder default backend is gone. Linux becomes the fallback for any platform without a room of its own.

### Step 5: macOS

Add a macOS color table tuned to Aqua conventions: a near-white light window and a deep neutral dark window, both a touch cooler than the curated greys, with the system blue as the fallback accent. Route `themePaletteFor(Mac, ...)` to it. Add `RMacThemeBackend`: it installs the native macos style, reads the user's system accent from the style's standard palette so the highlight follows the platform, and reports its capabilities. Pinning a variant rides the color-scheme hint the manager already sets, which on Qt 6.8 and newer flips the whole native appearance, title bar included; on older Qt the capability record drops title-bar control and the curated palette carries the forced variant. The whole backend body compiles only on macOS, and `makeThemeBackend()` selects it there; on other platforms it is an empty translation unit.

### Step 6: Windows

Add a Windows color table in the Fluent conventions of Windows 11: a light neutral window and a near-black dark window, with the default system blue as the fallback accent. Route `themePaletteFor(Windows, ...)` to it. Add `RWindowsThemeBackend`: it installs the windows11 style, falling back to windowsvista where that style is absent, reads the operating system accent from the style's standard palette, and switches the title bar to its dark variant through the desktop window manager. The chrome call loads dwmapi at runtime so the room needs no extra link dependency, and it is skipped under the offscreen and minimal platforms where there is no native window. The whole backend body compiles only on Windows, and `makeThemeBackend()` selects it there; on other platforms it is an empty translation unit.

### Step 7: Linux

Linux is the one with the most variety across desktops, so it leans hardest on the shared fallback. Add `RLinuxThemeBackend`: it keeps the desktop's own Qt style so the pilot looks native, honors the desktop accent when a recognized platform theme exposes one, and otherwise falls back to the curated palettes and reports no native accent. Whether a desktop is recognized is decided by `linuxDesktopHasNativeTheme`, a Qt-free reader of `XDG_CURRENT_DESKTOP` that reports true for GNOME or KDE. With all three rooms furnished, the factory no longer needs the placeholder `DefaultThemeBackend`, so drop it and select the Linux backend as the default for any platform without a room of its own.

### DIFF
--- a/cpp/solvcon/pilot/RLinuxThemeBackend.hpp
+++ b/cpp/solvcon/pilot/RLinuxThemeBackend.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * The Linux room: the native theme backend for Linux and other desktops.
+ *
+ * Keeps the desktop's own Qt style, honors the desktop accent when a
+ * recognized platform theme exposes one, and otherwise leans on the curated
+ * palettes. It is the room with the most variety across desktops, so it leans
+ * hardest on the shared fallback. This is also the factory's default, selected
+ * for any platform without a room of its own.
+ *
+ * @ingroup group_domain
+ */
+
+#include <solvcon/pilot/RThemeBackend.hpp>
+
+#include <optional>
+#include <string>
+
+namespace solvcon
+{
+
+/**
+ * @brief The Linux native theme backend.
+ *
+ * @ingroup group_domain
+ */
+class RLinuxThemeBackend
+    : public RThemeBackend
+{
+
+public:
+
+    PlatformId platform() const override;
+    std::string styleName() const override;
+    std::optional<ThemeColor> accentColor(ThemeVariant variant) const override;
+    void applyNativeChrome(QWidget * window, ThemeVariant variant) override;
+    ThemeCapabilities capabilities() const override;
+
+}; /* end class RLinuxThemeBackend */
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/pilot/RMacThemeBackend.hpp
+++ b/cpp/solvcon/pilot/RMacThemeBackend.hpp
@@ -1,0 +1,49 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * The macOS room: the native theme backend for macOS.
+ *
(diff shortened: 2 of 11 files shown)


cpp/solvcon/pilot/RLinuxThemeBackend.hpp
yungyuc, line 1: Theme backend for Linux.

cpp/solvcon/pilot/RLinuxThemeBackend.hpp
```diff
@@ -0,0 +1,50 @@
...
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * The Linux room: the native theme backend for Linux and other desktops.
+ *
+ * Keeps the desktop's own Qt style, honors the desktop accent when a
```
yungyuc, line 12: Use desktop Qt style on Linux.

cpp/solvcon/pilot/RMacThemeBackend.hpp
yungyuc, line 1: Theme backend for mac.

cpp/solvcon/pilot/RWindowsThemeBackend.hpp
yungyuc, line 1: Theme backend for windows.

cpp/solvcon/pilot/RWindowsThemeBackend.hpp
```diff
@@ -0,0 +1,49 @@
...
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * The Windows room: the native theme backend for Windows.
+ *
+ * Installs the windows11 style (falling back to windowsvista), reads the
```
yungyuc, line 12: Window 11 style.

cpp/solvcon/pilot/theme.cpp
```diff
@@ -66,6 +68,117 @@ static ThemePalette makeDarkPalette()
...
 
+// The macOS tables follow the Aqua conventions: a near-white light window and
+// a deep neutral dark window, both a touch cooler than the shared curated
+// greys, paired with the system blue the platform defaults to. A running mac
+// backend replaces the highlight with the user's chosen accent; these values
+// are the fallback when no accent is read.
+
+static ThemePalette makeMacLightPalette()
+{
+    ThemePalette p;
```
yungyuc, line 79: Hard-code colors for now.  Once the theme system stabilizes, we can extract the value to json files.

PR #1089, opened 2026-07-13
 Add the system-versus-curated look choice and the final theme polish (steps 8, 9 of theme)

### LINKED ISSUE #1062: Pilot theme system

yungyuc:
Steps 8, 9 of issue #1062.

Add a look preference (System or Curated) alongside the light or dark mode, so the user chooses between a native feel and a controlled look that travels between machines.

### Step 8: the system-versus-curated look choice

Add a look preference alongside the light or dark mode. System lets the platform's own colors show through the native style untouched, while Curated paints the plan's palettes over that style for a controlled look that carries between machines. Curated stays the default.

Add the `ThemeLook` vocabulary to the Qt-free core and route it through the manager: `apply()` paints the curated palette or the style's native standard palette by the current look, caching the effective window color so the MDI backdrop matches whichever is in force. Surface the choice as a second group in the Python theme menu, gated by the capability record now exposed as `theme_has_native_style`, so System greys out where no native style can carry it. Expose `set_look` and `theme_look` to Python, and extend the C++ menu sync so the look radio also keeps step when the theme moves from outside the menu.

### Step 9: accent role, chrome, overlays, and persistence

Round off the theme system with the touches a palette alone does not reach.

Set the dedicated `QPalette::Accent` role (Qt 6.6+) from the highlight, so widgets that want the platform accent distinct from a selection highlight pick it up. Add a thin supplemental stylesheet for the one thing a palette cannot draw, a tooltip border and a focus ring on text inputs, colored from the theme under the curated look and cleared under the system look so the native style stays untouched.

Recolor the two remaining hardcoded overlays, the axis text (`RTextOverlay`) and the scalar bar (`RScalarBar`), from the application palette so their labels stay legible when the theme turns dark instead of forcing black.

Persist the chosen mode and look through `QSettings` and read them back when the manager is built, so a theme carries across restarts.

### DIFF
--- a/cpp/solvcon/pilot/RManager.cpp
+++ b/cpp/solvcon/pilot/RManager.cpp
@@ -343,15 +343,11 @@ void RManager::setUpCentral()
                      { applyDrawTool(); });
 
     // The MDI area paints its own backdrop instead of reading the palette, so
-    // drive it from the theme's window color and follow theme changes;
-    // otherwise a stale slab shows through under the dark theme. Take the color
-    // from the theme model, not m_mainWindow->palette(): when themeChanged fires
-    // the freshly set application palette has not yet propagated to the window,
-    // so palette() would lag one switch behind.
-    auto paintBackdrop = [this](ThemeVariant variant)
+    // drive it from the theme's effective window color and follow theme
+    // changes; otherwise a stale slab shows through under the dark theme.
+    auto paintBackdrop = [this](ThemeVariant)
     {
-        ThemeColor const c = themePaletteFor(m_themeManager->platform(), variant).window;
-        m_mdiArea->setBackground(QColor(c.r, c.g, c.b));
+        m_mdiArea->setBackground(m_themeManager->windowColor());
     };
     paintBackdrop(m_themeManager->currentVariant());
     QObject::connect(m_themeManager, &RThemeManager::themeChanged, m_mdiArea, paintBackdrop);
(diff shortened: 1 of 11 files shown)


cpp/solvcon/pilot/RManager.cpp
```diff
@@ -343,15 +343,11 @@ void RManager::setUpCentral()
...
-    // so palette() would lag one switch behind.
-    auto paintBackdrop = [this](ThemeVariant variant)
-    {
-        ThemeColor const c = themePaletteFor(m_themeManager->platform(), variant).window;
-        m_mdiArea->setBackground(QColor(c.r, c.g, c.b));
+    // drive it from the theme's effective window color and follow theme
+    // changes; otherwise a stale slab shows through under the dark theme.
+    auto paintBackdrop = [this](ThemeVariant)
+    {
+        m_mdiArea->setBackground(m_themeManager->windowColor());
```
yungyuc, line 350: Get background color from theme manager.

cpp/solvcon/pilot/RThemeManager.hpp
```diff
@@ -109,8 +128,27 @@ class RThemeManager
...
+    void restorePersisted();
+
+    /// Write the current mode and look so the next session starts on them.
+    void persist() const;
+
     ThemeMode m_mode = ThemeMode::System;
 
+    /// Where the colors come from. Curated is the controlled default; System is
+    /// opted into for the platform's own colors.
+    ThemeLook m_look = ThemeLook::Curated;
```
yungyuc, line 146: Allow to select between curated and system coloring.

solvcon/pilot/_theme.py
```diff
@@ -30,6 +30,18 @@ def populate_menu(self):
             ],
             base_weight=10, setter=mgr.set_theme, current=mgr.theme_mode)
 
+        self._build_group(
+            "theme.look",
```
yungyuc, line 34: Curated/system menu items.

PR #1093, opened 2026-07-13
Add an Intel MKL BLAS/LAPACK option on Windows and a nightly CI job

### LINKED ISSUE #1060: Add pre-built vendor blas/lapack on windows
PR #1059 uses [openblas](https://github.com/solvcon/solvcon/pull/1059#discussion_r3565822583).  We should use an optimized, pre-built vendor blas/lapack on windows.  The target ISA is x86 because most windows machines run on x86.

The option should be added to [`build-scdv-windows.ps1`](https://github.com/solvcon/solvcon/blob/bb696a3be9d54667de07416bbabc429c1ce21f57/contrib/dependency/windows/build-scdv-windows.ps1).

tigercosmos:
## What this does

Windows links OpenBLAS as solvcon's LP64 BLAS/LAPACK today. This change adds Intel oneAPI MKL as an optimized vendor alternative for x86, entirely opt-in. The from-source dependency script `build-scdv-windows.ps1` gains a `SCDV_WIN_BLAS` selector (`openblas` by default, or `mkl`), and the build gains a `SOLVCON_USE_MKL` CMake option (off by default). When neither is set the existing OpenBLAS path is untouched, so nothing changes for current users or for the default build and CI.

MKL slots in with almost no new C++. Its LP64 runtime exports the same reference CBLAS entry points and the same Fortran LAPACK symbols solvcon already declares, so `matmul_blas()` and `EigenSystem` reach it through the very same code paths as OpenBLAS; the only source change is a one-line `#include` that picks MKL's CBLAS header. An MKL scdv also exports `SOLVCON_USE_MKL` when activated, so a solvcon build inside that scdv selects MKL automatically without an explicit flag.

## What we tested

This PR adds `tests/test_blas_lapack.py`, an acceptance test that checks both backends against NumPy's own LAPACK rather than against another solvcon path. That matters because comparing one solvcon path to another cannot catch a miscompiled or ABI-mismatched vendor library that returns consistently-wrong values; comparing to an independent implementation does. It covers:

- `matmul_blas()` (CBLAS `?gemm`) against NumPy for float32, float64, complex64, and complex128. The complex `cgemm`/`zgemm` path had no prior test.
- `EigenSystem` (LAPACK `?geev`) eigenvalues against `numpy.linalg.eigvals` for the same four dtypes, plus an independent `A @ vr == vr @ diag(w)` eigenvector reconstruction for the complex types (real types pack conjugate pairs across two columns, so that relation is left to the eigenvalue check).

## Validation

The change was validated three ways. The full `devbuild` CI matrix is green with the new test on ubuntu (OpenBLAS), macOS (Accelerate), and the new `build_windows_mkl` job on `windows-2022` against real MKL: https://github.com/solvcon/solvcon/actions/runs/29268780837. Separately, a hands-on build on a standalone Windows machine linked `_solvcon` against MKL and the test passed (3 passed, 10 subtests). Finally, a negative control confirmed the test fails when the computed result is perturbed, so it genuinely has teeth rather than passing vacuously.

## CI coverage

CI coverage is a new nightly-only `build_windows_mkl` job that builds solvcon against MKL with `SOLVCON_USE_MKL=ON` and runs the gtest and pilot suites. It is nightly-only so the large MKL wheels do not slow pull-request CI, while the existing OpenBLAS `build_windows` job continues to gate every pull request.

Related to #1060.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -335,26 +335,28 @@ jobs:
           # cold).
           max-size: 500M
 
-      - name: Cache vcpkg (install artifacts)
-        uses: actions/cache@v4
-        with:
-          path: |
-            C:\vcpkg\installed
-            C:\vcpkg\downloads
-            C:\vcpkg\buildtrees
-            C:\vcpkg\packages
-          key: vcpkg-${{ runner.os }}-openblas-lapack
-          restore-keys: vcpkg-${{ runner.os }}-
-
-      - name: install BLAS and LAPACK
+      - name: install Intel MKL (vendor BLAS/LAPACK)
         run: |
-          echo "::group::install BLAS and LAPACK"
-          vcpkg install openblas:x64-windows lapack:x64-windows
-          $vcpkg_bin_path = @(
-            "C:\vcpkg\installed\x64-windows\bin",
-            "C:\vcpkg\installed\x64-windows\debug\bin"
-          ) -join ";"
-          echo "$vcpkg_bin_path" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          echo "::group::install Intel MKL"
+          # Pre-built vendor BLAS/LAPACK from Intel's official wheels, the same
+          # source build-scdv-windows.ps1 uses for SCDV_WIN_BLAS=mkl.
+          pip install mkl mkl-devel mkl-include intel-openmp
+          $prefix = python -c "import sys; sys.stdout.write(sys.prefix)"
+          # Locate the wheel-installed pieces by name (the layout varies across
+          # wheel versions) and hand CMake their dirs so find_path(mkl_cblas.h)
+          # and find_library(mkl_rt) succeed.
+          $mkl = Get-ChildItem -LiteralPath $prefix -Recurse -File -Include 'mkl_cblas.h', 'mkl_rt.lib', 'mkl_rt*.dll'
+          $hdr = $mkl | Where-Object Name -EQ 'mkl_cblas.h' | Select-Object -First 1
+          $lib = $mkl | Where-Object Name -EQ 'mkl_rt.lib' | Select-Object -First 1
+          $dll = $mkl | Where-Object Name -Like 'mkl_rt*.dll' | Select-Object -First 1
+          if (-not ($hdr -and $lib -and $dll)) {
+            throw "MKL pieces not found (hdr=$hdr lib=$lib dll=$dll)"
+          }
+          echo "MKL_INCLUDE_DIR=$($hdr.DirectoryName)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          echo "MKL_LIBRARY_DIR=$($lib.DirectoryName)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          # mkl_rt loads the mkl_* kernels and libiomp5md at runtime; put their
+          # dir on PATH so the test executables find them.
+          echo "$($dll.DirectoryName)" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
           echo "::endgroup::"
 
       - name: dependency by pip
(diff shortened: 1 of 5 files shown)


cpp/solvcon/math/blas_compat.cpp
```diff
@@ -14,6 +14,8 @@
 #endif
 #include <vecLib/cblas_new.h>
 
+#elifdef MM_HAS_MKL
+#include <mkl_cblas.h>
```
tigercosmos, line 18: Include MLK.

tests/test_blas_lapack.py
```diff
@@ -0,0 +1,95 @@
...
+# name -> (SimpleArray class, numpy dtype, tolerance)
+_DTYPES = {
+    "float32": (sc.SimpleArrayFloat32, np.float32, 1e-4),
+    "float64": (sc.SimpleArrayFloat64, np.float64, 1e-12),
+    "complex64": (sc.SimpleArrayComplex64, np.complex64, 1e-4),
+    "complex128": (sc.SimpleArrayComplex128, np.complex128, 1e-12),
+}
+
+
+class MatmulBlasTC(unittest.TestCase):
```
tigercosmos, line 29: Some simple checks.

.github/workflows/devbuild.yml
```diff
@@ -241,6 +241,145 @@ jobs:
         make run_pilot_pytest VERBOSE=1
         echo "::endgroup::"
 
+  build_windows_mkl:
```
tigercosmos: Only nightly build.

tigercosmos (commented):

.github/workflows/devbuild.yml
yungyuc: Having alternate windows build looks like an overkill.  I prefer to just use MKL for windows build.

@ExplorerRay @chestercheng What do you think?

.github/workflows/devbuild.yml
```diff
@@ -550,7 +689,7 @@ jobs:
         echo "::endgroup::"
 
   send_email_on_failure:
-    needs: [standalone_buffer, build, build_windows, sanitizer]
+    needs: [standalone_buffer, build, build_windows, build_windows_mkl, sanitizer]
```
yungyuc: An additional MKL run looks like too many.

contrib/dependency/windows/build-scdv-windows.ps1
```diff
@@ -64,6 +65,9 @@
 #     LIBCLANG_VERSION: Qt prebuilt libclang version for shiboken.
 #   Build settings:
 #     SCDV_NP: Parallel build jobs.
+#     SCDV_WIN_BLAS: LP64 BLAS/LAPACK provider for solvcon's matmul_blas() /
```
yungyuc, line 68: It's correct to use LP64 for now.

contrib/dependency/windows/build-scdv-windows.ps1
```diff
@@ -289,6 +293,14 @@ $ScdvSrcDir = Join-Path $ScdvBase 'src'
 # Install root (user-space).
 $ScdvUsrDir = Join-Path $ScdvBase 'usr'
 
+# LP64 BLAS/LAPACK provider solvcon links for SimpleArray::matmul_blas() and
+# EigenSystem: "openblas" or "mkl".  Distinct from the ILP64 scipy-openblas64
+# wheel numpy links (see Get-OpenblasPkgConfig).
+$WinBlas = (Get-EnvOrDefault 'SCDV_WIN_BLAS' 'openblas').ToLower()
```
yungyuc: Make MKL the default `WinBlas` because it's faster.  MKL is very standard cool on Intel.

yungyuc (changes requested):
- [x] Just use MKL for windows build without duplicating the runs.
- [x] Use MKL as default.

----

The PR description is too verbose.  Please make it concise and less AI-pilled next time.

.github/workflows/devbuild.yml
tigercosmos: Removed it.

contrib/dependency/windows/build-scdv-windows.ps1
tigercosmos: Fixed it.

tigercosmos:
@yungyuc  The issues are fixed. Please take a look, thanks!

yungyuc (approved):
LGTM

@ExplorerRay @chestercheng please take a look.

PR #978, opened 2026-07-01
Make shape as a signed vector in SimpleArray

ThreeMonth03:
Previous [discussion](https://github.com/solvcon/solvcon/issues/856#issuecomment-4686103124) points out third-party libraries usually implement array with signed vector, which could avoid unnecessary casting and wierd bugs. 

To unify the type in `SimpleArray` and reduce type casting, this pull request changes shapes and indices to signed vector in SimpleArray.

Related to #856.

PR #1094, opened 2026-07-13
Make SimpleArray treat size-1 axes as contiguous like NumPy

### LINKED ISSUE #603: Bug: F-contigous check in SimpleArray breaks

### LINKED ISSUE #856: SimpleArray should handle negative NumPy strides and Python slice semantics consistently

rockleona:
Make SimpleArray treat size-1 axes as contiguous like NumPy and add tests.

This changes contiguous-detection logic to follow NumPy semantics: when a dimension has length 1 its stride may be arbitrary and does not break C-contiguity or F-contiguity. Also made check_f_contiguous static to match check_c_contiguous. Added tests that exercise 1x2, 2x1, and 2x2 Fortran views and assert contiguity flags and ndarray equality.

Related to #603.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2078,16 +2078,18 @@ private:
     static bool is_c_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (stride[stride.size() - 1] != 1)
+        if (shape.empty())
         {
-            return false;
+            return true;
         }
-        for (size_t it = 0; it < shape.size() - 1; ++it)
+        ssize_t expected = 1;
+        for (ssize_t i = static_cast<ssize_t>(shape.size()) - 1; i >= 0; --i)
         {
-            if (stride[it] != shape[it + 1] * stride[it + 1])
+            if (stride[i] != expected && shape[i] > 1)
             {
                 return false;
             }
+            expected *= shape[i];
         }
         return true;
     }
@@ -2095,16 +2097,18 @@ private:
     static bool is_f_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (stride[0] != 1)
+        if (shape.empty())
         {
-            return false;
+            return true;
         }
-        for (size_t it = 0; it < shape.size() - 1; ++it)
+        ssize_t expected = 1;
+        for (size_t i = 0; i < shape.size(); ++i)
         {
-            if (stride[it + 1] != shape[it] * stride[it])
+            if (stride[i] != expected && shape[i] > 1)
             {
                 return false;
             }
+            expected *= shape[i];
         }
         return true;
     }
@@ -2118,8 +2122,8 @@ private:
         }
     }
 
-    void check_f_contiguous(shape_type const & shape,
-                            sshape_type const & stride) const
+    static void check_f_contiguous(shape_type const & shape,
+                                   sshape_type const & stride)
     {
         if (!is_f_contiguous(shape, stride))
         {
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -289,6 +289,35 @@ class SimpleArrayBasicTC(unittest.TestCase):
         ):
             solvcon.SimpleArrayFloat64((-1, 2))
 
+    def test_SimpleArray_fortran_empty(self):
+        ndarr = np.empty((0, 0), order='F')
+        sarr = solvcon.SimpleArrayFloat64(array=ndarr)
+
+        self.assertTrue(sarr.is_c_contiguous)
+        self.assertTrue(sarr.is_f_contiguous)
+        np.testing.assert_array_equal(ndarr, sarr.ndarray)
+
+    def test_SimpleArray_fortran_1x2(self):
+        ndarr = np.asfortranarray([[1.0, 0.1]])
+        sarr = solvcon.SimpleArrayFloat64(array=ndarr)
+        self.assertTrue(sarr.is_c_contiguous)
+        self.assertTrue(sarr.is_f_contiguous)
+        np.testing.assert_array_equal(ndarr, sarr.ndarray)
+
+    def test_SimpleArray_fortran_2x1(self):
+        ndarr = np.asfortranarray([[1.0], [0.0]])
+        sarr = solvcon.SimpleArrayFloat64(array=ndarr)
+        self.assertTrue(sarr.is_c_contiguous)
+        self.assertTrue(sarr.is_f_contiguous)
+        np.testing.assert_array_equal(ndarr, sarr.ndarray)
+
+    def test_SimpleArray_fortran_2x2(self):
+        ndarr = np.asfortranarray([[1.0, 0.1], [0.0, 1.0]])
+        sarr = solvcon.SimpleArrayFloat64(array=ndarr)
+        self.assertFalse(sarr.is_c_contiguous)
+        self.assertTrue(sarr.is_f_contiguous)
+        np.testing.assert_array_equal(ndarr, sarr.ndarray)
+
     def test_SimpleArray_from_numpy_negative_stride(self):
         ndarr = np.arange(2 * 3, dtype='float64').reshape((2, 3))
         view = ndarr[::-1, ::-1]


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2084,33 +2084,43 @@ class SimpleArray
...
+            if (stride[i] != expected)
             {
-                return false;
+                if (static_cast<ssize_t>(shape[i]) != 1)
+                {
+                    return false;
+                }
             }
+            expected *= static_cast<ssize_t>(shape[i]);
         }
```
rockleona, line 2093: Iterate from last axis inward because C layout packs the last axis as contiguous (stride 1), then multiplies by axis lengths moving outward.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2084,33 +2084,43 @@ class SimpleArray
...
```
rockleona, line 2112: iterate from first axis outward because Fortran packs the first axis as contiguous.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2124,8 +2134,8 @@ class SimpleArray
...
     }
 
-    void check_f_contiguous(shape_type const & shape,
-                            sshape_type const & stride) const
+    static void check_f_contiguous(shape_type const & shape,
+                                    sshape_type const & stride)
```
rockleona: Not really sure if this change is needed or this had some other  reason, if no, make it static as the symmetry with `check_c_contiguous`

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2084,33 +2084,43 @@ class SimpleArray
     static bool is_c_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (stride[stride.size() - 1] != 1)
+        if (shape.empty())
         {
-            return false;
+            return true;
         }
```
rockleona, line 2084: Empty (0-d) arrays are trivially contiguous; allow both C/F semantics.

tests/test_buffer.py
```diff
@@ -282,6 +282,27 @@ def test_SimpleArray(self):
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
         self.assertEqual((2, 2, 2, 3), sarr.reshape((2, 2, 2, 3)).shape)
 
+    def test_SimpleArray_fortran_1x2(self):
+        ndarr = np.asfortranarray([[1.0, 0.1]])
+        sarr = solvcon.SimpleArrayFloat64(array=ndarr)
+        self.assertEqual(bool(ndarr.flags.c_contiguous), sarr.is_c_contiguous)
```
yungyuc: Assertion for the same results as numpy is good and what we need, but also add assertions to explicitly spell out `True`/`False`, e.g.,
```python
self.assertFalse(sarr.is_c_contiguous)
self.assertTrue(sarr.is_f_contiguous)
```
so that we can tell the behaviors without needing to run the tests.

tests/test_buffer.py
```diff
@@ -282,6 +282,27 @@ def test_SimpleArray(self):
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
         self.assertEqual((2, 2, 2, 3), sarr.reshape((2, 2, 2, 3)).shape)
 
+    def test_SimpleArray_fortran_1x2(self):
```
yungyuc, line 300: Why not test empty arrays?

yungyuc (changes requested):
Please squash for the simple PR.

- [ ] Add assertions to explicitly spell out `True`/`False` for readability.
- [ ] Test empty arrays.

@KHLee529 @tigercosmos Please help take a look.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2084,33 +2084,43 @@ class SimpleArray
...
-        for (size_t it = 0; it < shape.size() - 1; ++it)
+        ssize_t expected = 1;
+        for (ssize_t i = static_cast<ssize_t>(shape.size()) - 1; i >= 0; --i)
         {
-            if (stride[it] != static_cast<ssize_t>(shape[it + 1]) * stride[it + 1])
+            if (stride[i] != expected)
             {
-                return false;
+                if (static_cast<ssize_t>(shape[i]) != 1)
```
KHLee529: Please leave a comment on why this condition matters.

KHLee529 (commented):
Maybe you can add some test case for sliced array which are neither c-contiguous nor f-contiguous.

cpp/solvcon/buffer/SimpleArray.hpp
rockleona: The condition ensures the constructor only accepts arrays with a C-contiguous layout and a final stride of 1. SimpleArray depends on that memory layout for correct element indexing and safe zero-copy sharing with NumPy; allowing arrays with mismatched strides (e.g., Fortran-order or inconsistent shape/stride) can produce incorrect reads/writes or hard-to-debug crashes later.

Here is the example to demonstrate stride-mismatch case:

```python3
import numpy as np
from numpy.lib.stride_tricks import as_strided

import solvcon


array = np.arange(6, dtype='float64') # [0..5]

# force a view with element-stride (2,1) in items -> bytes strides (16,8)
view = as_strided(array, shape=(2,3), strides=(16,8))

print("shape:", view.shape, "item-stride:", (view.strides[0]//8, view.strides[1]//8))
print("numpy C-contiguous:", view.flags['C_CONTIGUOUS'])

print("SimpleArray stride:", sarr.stride, "is_c_contiguous:", sarr.is_c_contiguous)
```

tests/test_buffer.py
rockleona: Added at L292

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2075,33 +2075,43 @@ class SimpleArray
...
         {
-            if (stride[it + 1] != shape[it] * stride[it])
+            if (stride[i] != expected)
             {
-                return false;
+                if (static_cast<ssize_t>(shape[i]) != 1)
+                {
+                    return false;
+                }
             }
```
ThreeMonth03, line 2110: We might simplify the nested branch to improve readability.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2075,33 +2075,43 @@ class SimpleArray
...
 
     static bool is_f_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (stride[0] != 1)
+        if (shape.empty())
         {
-            return false;
+            return true;
         }
```
ThreeMonth03, line 2103: By the way, I'm not sure whether the following case could be constructed in Solvcon. If it could be initialized, we might seen them as contiguous array, since numpy docs points out the following example is a contiguous array.

```python3
def test_SimpleArray_fortran_empty(self):
    ndarr = np.empty((0, 0), order='F')
    sarr = solvcon.SimpleArrayFloat64(array=ndarr)

    self.assertTrue(sarr.is_c_contiguous)
    self.assertTrue(sarr.is_f_contiguous)
    np.testing.assert_array_equal(ndarr, sarr.ndarray)
```

Reference: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html
```
Even for contiguous arrays a stride for a given dimension arr.strides[dim] may be arbitrary if arr.shape[dim] == 1 or the array has no elements.
```

ThreeMonth03 (changes requested):
- [ ] Remove nested if-else branch.
- [ ] Clarify the behavior if shape vector is `(0, ..., 0)`.

cpp/solvcon/buffer/SimpleArray.hpp
rockleona: Got it, since we need to follow the behavior of numpy, I will try to implement this.

rockleona:
Thank you @KHLee529 and @ThreeMonth03, I've update the change with your suggestion, please review them again.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2078,33 +2078,37 @@ class SimpleArray
...
-        for (size_t it = 0; it < shape.size() - 1; ++it)
+        ssize_t expected = 1;
+        for (size_t i = 0; i < shape.size(); ++i)
         {
-            if (stride[it + 1] != shape[it] * stride[it])
+            if (stride[i] != expected && static_cast<ssize_t>(shape[i]) > 1)
             {
                 return false;
             }
+            expected *= shape[i];
```
rockleona, line 2111: From #978, the type of shape had changed to `small_vector<ssize_t>` so there's no need to transform in this line.

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03: Good modification.

ThreeMonth03 (approved):
LGTM

rockleona:
Hi @yungyuc, I've refined the code as the comments above, please review, thank you.

cpp/solvcon/buffer/SimpleArray.hpp
tigercosmos, line 2121: Please fix the error message for the new relaxed rule.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2078,33 +2078,37 @@ class SimpleArray
...
     {
-        if (stride[0] != 1)
+        if (shape.empty())
         {
-            return false;
+            return true;
         }
-        for (size_t it = 0; it < shape.size() - 1; ++it)
+        ssize_t expected = 1;
+        for (size_t i = 0; i < shape.size(); ++i)
```
tigercosmos, line 2105: Nit: use the same type for L2086 and L2105 for `i`.

tigercosmos (approved):
LGTM, there are some minor errors, @rockleona please fix in a following PR. Thanks.

yungyuc (commented):
LGTM

PR #1097, opened 2026-07-14
Tell agents not to hard-wrap GitHub issue and PR prose

tigercosmos:
Add a rule to the "Writing Style for Prose" section so agents write one line per paragraph when they create, update, or comment on issues and pull requests, since manual mid-paragraph line breaks render as ragged text on GitHub while source files still keep their linting line-width limits.

[skip-ci]

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -218,6 +218,11 @@ messages, PR and issue descriptions and comments, and documentation.
   em-dash, en-dash, or "smart quotes" in them. The same restraint is expected
   in GitHub prose even though GitHub accepts Unicode.
 - Write plainly. Short sentences beat long ones strung together with dashes.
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -218,6 +218,11 @@ messages, PR and issue descriptions and comments, and documentation.
   em-dash, en-dash, or "smart quotes" in them. The same restraint is expected
   in GitHub prose even though GitHub accepts Unicode.
 - Write plainly. Short sentences beat long ones strung together with dashes.
+- Do not hard-wrap GitHub prose. When you create, update, or comment on an
```
tigercosmos, line 221: It happens too many times, so I decide to put it in the general guideline.

yungyuc: It happens all the time.  I grew a habit to tell it to unwrap in every prompt.

PR #1101, opened 2026-07-14
Preserve NumPy view origins in SimpleArray buffers

### LINKED ISSUE #856: SimpleArray should handle negative NumPy strides and Python slice semantics consistently

ThreeMonth03:
`SimpleArray` used the logical origin of a NumPy view as the start of its `ConcreteBuffer`. For a negative-strided view, valid elements may lie before that pointer.

```python
ndarr = np.arange(6, dtype="float64").reshape((2, 3))
view = ndarr[::-1, ::-1]

clone = sarr.clone()
```

The original `sarr` could still read the view because the complete NumPy allocation remained alive. However, `clone()` copied only the incomplete range declared by `ConcreteBuffer` while retaining the negative strides. Reading `clone.ndarray` then accessed memory before the cloned allocation.

This PR stores the complete reachable storage span in `ConcreteBuffer` and records the byte offset to the logical origin. It also validates every strided `SimpleArray` constructor, preserves `data_offset` in copied transposes, and rejects NumPy views whose byte strides or data alignment cannot be represented safely.

Regression tests cover cloning and transposing a negative-strided view, and rejecting unrepresentable NumPy strides and alignment.

Related to #856.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -270,7 +270,7 @@ public:
         {
             return sum_contiguous(athis->data(), n);
         }
-        return sum_strided(athis->data(), athis->shape(), athis->stride());
+        return sum_strided(athis->logical_data(), athis->shape(), athis->stride());
     }
 
 private:
@@ -1610,6 +1610,24 @@ public:
     using sshape_type = typename internal_types::sshape_type;
     using buffer_type = typename internal_types::buffer_type;
 
+    enum class ArrayOrder : std::uint8_t
+    {
+        Unspecified = 0,
+        CType = 1U << 0,
+        FType = 1U << 1,
+    };
+
+    friend constexpr ArrayOrder operator|(ArrayOrder lhs, ArrayOrder rhs) noexcept
+    {
+        return static_cast<ArrayOrder>(std::to_underlying(lhs) | std::to_underlying(rhs));
+    }
+
+    friend constexpr ArrayOrder & operator|=(ArrayOrder & lhs, ArrayOrder rhs) noexcept
+    {
+        lhs = lhs | rhs;
+        return lhs;
+    }
+
     static constexpr size_t ITEMSIZE = sizeof(value_type);
 
     static constexpr size_t itemsize() { return ITEMSIZE; }
@@ -1637,7 +1655,7 @@ public:
         if (!m_shape.empty())
         {
             m_buffer = buffer_type::construct(static_cast<size_t>(storage_size(m_shape, m_stride)) * ITEMSIZE, 0);
-            m_body = m_buffer->template data<T>();
+            update_data_pointers();
         }
     }
 
@@ -1649,7 +1667,7 @@ public:
         if (!m_shape.empty())
         {
             m_buffer = buffer_type::construct(static_cast<size_t>(storage_size(m_shape, m_stride)) * ITEMSIZE, alignment);
-            m_body = m_buffer->template data<T>();
+            update_data_pointers();
         }
     }
 
@@ -1697,7 +1715,7 @@ public:
             m_shape = shape_type{static_cast<ssize_t>(nitem)};
             m_stride = sshape_type{1};
             m_buffer = buffer;
-            m_body = m_buffer->template data<T>();
+            update_data_pointers();
         }
         else
         {
@@ -1706,6 +1724,13 @@ public:
     }
 
     explicit SimpleArray(shape_type const & shape, std::shared_ptr<buffer_type> const & buffer)
+        : SimpleArray(shape, buffer, 0)
+    {
+    }
+
+    explicit SimpleArray(shape_type const & shape,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         size_t data_offset)
         : SimpleArray(buffer)
     {
         if (buffer)
@@ -1713,32 +1738,50 @@ public:
             m_shape = shape;
             m_stride = calc_stride(m_shape);
             const size_t nbytes = static_cast<size_t>(storage_size(m_shape, m_stride)) * ITEMSIZE;
-            if (nbytes != buffer->nbytes())
+            if (data_offset % ITEMSIZE != 0)
+            {
+                throw std::runtime_error("SimpleArray: data offset must be divisible by item size");
+            }
+            if (data_offset > buffer->nbytes())
             {
                 throw std::runtime_error(
-                    std::format("SimpleArray: shape byte count {} differs from buffer {}",
-                                nbytes,
+                    std::format("SimpleArray: data offset {} exceeds buffer byte count {}",
+                                data_offset,
                                 buffer->nbytes()));
             }
+            size_t const available_nbytes = buffer->nbytes() - data_offset;
+            if (nbytes != available_nbytes)
+            {
+                throw std::runtime_error(
+                    std::format("SimpleArray: shape byte count {} differs from available buffer byte count {} at data offset {}",
+                                nbytes,
+                                available_nbytes,
+                                data_offset));
+            }
+            update_data_pointers(data_offset);
         }
     }
 
     explicit SimpleArray(shape_type const & shape,
                          sshape_type const & stride,
                          std::shared_ptr<buffer_type> const & buffer,
-                         bool c_contiguous,
-                         bool f_contiguous)
-        : SimpleArray(shape, stride, buffer)
+                         ArrayOrder array_order)
+        : SimpleArray(shape, stride, buffer, 0, array_order)
     {
-        if (shape.size() != stride.size())
-        {
-            throw std::runtime_error("SimpleArray: shape and stride size mismatch");
-        }
-        if (c_contiguous)
+    }
+
+    explicit SimpleArray(shape_type const & shape,
+                         sshape_type const & stride,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         size_t data_offset,
+                         ArrayOrder array_order)
+        : SimpleArray(shape, stride, buffer, data_offset)
+    {
+        if (has_order(array_order, ArrayOrder::CType))
         {
             check_c_contiguous(shape, stride);
         }
-        if (f_contiguous)
+        if (has_order(array_order, ArrayOrder::FType))
         {
             check_f_contiguous(shape, stride);
         }
@@ -1747,12 +1790,22 @@ public:
     explicit SimpleArray(shape_type const & shape,
                          sshape_type const & stride,
                          std::shared_ptr<buffer_type> const & buffer)
+        : SimpleArray(shape, stride, buffer, 0)
+    {
+    }
+
+    explicit SimpleArray(shape_type const & shape,
+                         sshape_type const & stride,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         size_t data_offset)
         : SimpleArray(buffer)
     {
         if (buffer)
         {
+            validate_layout(shape, stride, buffer->nbytes(), data_offset);
             m_shape = shape;
             m_stride = stride;
+            update_data_pointers(data_offset);
         }
     }
 
@@ -1772,14 +1825,15 @@ public:
         , m_shape(other.m_shape)
         , m_stride(other.m_stride)
         , m_nghost(other.m_nghost)
-        , m_body(calc_body(m_buffer->template data<T>(), m_stride, other.m_nghost))
     {
+        update_data_pointers(other.logical_data_offset());
     }
 
     SimpleArray(SimpleArray && other) noexcept
         : m_buffer(std::move(other.m_buffer))
         , m_shape(std::move(other.m_shape))
         , m_stride(std::move(other.m_stride))
+        , m_logical_data(other.m_logical_data)
         , m_nghost(other.m_nghost)
         , m_body(other.m_body)
     {
@@ -1793,7 +1847,7 @@ public:
             m_shape = other.m_shape;
             m_stride = other.m_stride;
             m_nghost = other.m_nghost;
-            m_body = calc_body(m_buffer->template data<T>(), m_stride, other.m_nghost);
+            update_data_pointers(other.logical_data_offset());
         }
         return *this;
     }
@@ -1805,6 +1859,7 @@ public:
             m_buffer = std::move(other.m_buffer);
             m_shape = std::move(other.m_shape);
             m_stride = std::move(other.m_stride);
+            m_logical_data = other.m_logical_data;
             m_nghost = other.m_nghost;
             m_body = other.m_body;
         }
@@ -1889,13 +1944,13 @@ public:
     {
         sshape_type const idx{normalize_index(it)};
         ssize_t const offset = buffer_offset(m_stride, idx);
-        return *(data() + offset);
+        return *(logical_data() + offset);
     }
     value_type & at(ssize_t it)
     {
         sshape_type const idx{normalize_index(it)};
         ssize_t const offset = buffer_offset(m_stride, idx);
-        return *(data() + offset);
+        return *(logical_data() + offset);
     }
 
     value_type const & at(std::vector<ssize_t> const & idx) const { return at(sshape_type(idx)); }
@@ -1905,13 +1960,13 @@ public:
     {
         sshape_type const idx = normalize_index(sidx);
         ssize_t const offset = buffer_offset(m_stride, idx);
-        return *(data() + offset);
+        return *(logical_data() + offset);
     }
     value_type & at(sshape_type const & sidx)
     {
         sshape_type const idx = normalize_index(sidx);
         ssize_t const offset = buffer_offset(m_stride, idx);
-        return *(data() + offset);
+        return *(logical_data() + offset);
     }
 
     ssize_t ndim() const noexcept { return static_cast<ssize_t>(m_shape.size()); }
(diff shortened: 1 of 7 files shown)


tests/test_buffer.py
```diff
@@ -308,8 +308,13 @@ def test_SimpleArray_from_numpy_negative_stride(self):
         self.assertTrue(row_major.is_c_contiguous)
         np.testing.assert_array_equal(rhs, row_major.ndarray)
 
-        sarr.ndarray[0, 0] = 100
-        self.assertEqual(100, ndarr[1, 2])
+        clone = sarr.clone()
+        self.assertFalse(clone.is_from_python)
+        np.testing.assert_array_equal(rhs, clone.ndarray)
```
ThreeMonth03, line 346: Add unittest for cloned buffer with negative strides array. With reallocation, this testcase can raise segment fault if the program dereference an unallocated pointer.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2284,10 +2353,59 @@ class SimpleArray
     /// bytes) to skip for advancing an index in the corresponding dimension.
     sshape_type m_stride;
 
+    /// Byte offset from the buffer start to the array's logical origin.
+    size_t m_data_offset = 0;
```
ThreeMonth03: Add a new member variable `m_data_offset`, which records slice offset.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2038,8 +2094,16 @@ class SimpleArray
...
+    value_type const * data() const
+    {
+        value_type const * ret = buffer().template data<value_type>();
+        return ret ? ret + m_data_offset / ITEMSIZE : nullptr;
+    }
+    value_type * data()
+    {
+        value_type * ret = buffer().template data<value_type>();
+        return ret ? ret + m_data_offset / ITEMSIZE : nullptr;
+    }
```
ThreeMonth03: `data()` will return the pointer of first element in a sliced array now.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2004,11 +2060,11 @@ class SimpleArray
...
     std::span<value_type> as_span()
     {
-        return std::span<value_type>(data(), buffer().nbytes() / ITEMSIZE);
+        return std::span<value_type>(buffer().template data<value_type>(), buffer().nbytes() / ITEMSIZE);
     }
     std::span<value_type const> as_span() const
     {
-        return std::span<value_type const>(data(), buffer().nbytes() / ITEMSIZE);
+        return std::span<value_type const>(buffer().template data<value_type>(), buffer().nbytes() / ITEMSIZE);
     }
```
ThreeMonth03, line 2068: `as_span()` still returns a span of whole buffer.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -58,74 +60,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-                        const size_t view_nbytes = true_owner_nbytes - static_cast<size_t>(offset_bytes);
-                        auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(owner);
-                        const std::shared_ptr<ConcreteBuffer> buffer =
-                            ConcreteBuffer::construct(
-                                view_nbytes,
-                                view_ptr,
-                                std::move(remover));
-                        return wrapped_type(shape, stride, buffer, is_c_contiguous, is_f_contiguous);
-                    }),
-                py::arg("array"))
```
ThreeMonth03, line 128: Modify lambda function to a static member function, since it triggers lint error after modification.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -764,6 +699,91 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+
+    char * view_ptr = static_cast<char *>(arr_in.mutable_data());
+    char * storage_ptr = view_ptr + storage_begin_offset * itemsize;
+    const size_t storage_nbytes = has_element
+                                      ? static_cast<size_t>(storage_end_offset - storage_begin_offset + 1) *
+                                            wrapped_type::itemsize()
+                                      : 0;
+    const auto data_offset = static_cast<size_t>(-storage_begin_offset) * wrapped_type::itemsize();
+    auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(owner);
+    const auto buffer = ConcreteBuffer::construct(storage_nbytes, storage_ptr, std::move(remover));
```
ThreeMonth03, line 807: `ConcreteBuffer` is constructed by `storage_begin_offset` and `storage_end_offset` now, which are calculated based on negative and positive strides.

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03, line 1: SimpleArray is constructed based on `m_data_offset` now.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1706,20 +1706,33 @@ class SimpleArray
...
             {
                 throw std::runtime_error(
                     std::format("SimpleArray: shape byte count {} differs from buffer {}",
                                 nbytes,
                                 buffer->nbytes()));
             }
+            m_data_offset = data_offset;
+            m_body = data();
         }
     }
```
ThreeMonth03, line 1763: `SimpleArray::reshape()` still uses these functions, so I decide to keep these constructors now, even though they don't consider negative strides. We might open a new issue for fixing `SimpleArray::reshape()`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2284,10 +2353,59 @@ class SimpleArray
...
     ssize_t m_nghost = 0;
     value_type * m_body = nullptr;
 }; /* end class SimpleArray */
 
+template <typename T>
+void SimpleArray<T>::validate_layout(shape_type const & shape,
+                                     sshape_type const & stride,
+                                     size_t buffer_nbytes,
+                                     size_t data_offset)
+{
```
ThreeMonth03, line 2392: This function ensures the elements are in the buffer, given shape, stride and data_offset.

ThreeMonth03 (commented):
@j8xixo12 @yungyuc Please take a look, thanks.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -764,6 +701,106 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+    {
+        storage_begin_offset = 0;
+        storage_end_offset = 0;
+    }
+
+    const bool is_c_contiguous = (arr_in.flags() & py::array::c_style) == py::array::c_style;
+    const bool is_f_contiguous = (arr_in.flags() & py::array::f_style) == py::array::f_style;
+
+    py::array owner = arr_in;
+    /*
```
tigercosmos, line 767: use `//` instead of `/** */`. The later one is for multiple-line doxygron docstring.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1744,15 +1757,48 @@ class SimpleArray
         }
     }
 
+    explicit SimpleArray(shape_type const & shape,
+                         sshape_type const & stride,
+                         std::shared_ptr<buffer_type> const & buffer,
+                         size_t data_offset,
+                         bool c_contiguous,
+                         bool f_contiguous)
```
tigercosmos: I wonder if a enum can be better: `SimpleArrary::ArrayOrder::CType`, `SimpleArrary::ArrayOrder::FType`

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2038,8 +2089,16 @@ class SimpleArray
     /* Backdoor */
     value_type const & data(size_t it) const { return data()[it]; }
     value_type & data(size_t it) { return data()[it]; }
-    value_type const * data() const { return buffer().template data<value_type>(); }
-    value_type * data() { return buffer().template data<value_type>(); }
+    value_type const * data() const
+    {
+        value_type const * ret = buffer().template data<value_type>();
+        return ret ? ret + m_data_offset / ITEMSIZE : nullptr;
+    }
```
tigercosmos: `data()` is called frequently after this PR, but `ret + m_data_offset / ITEMSIZE` sounds heavy. Worth doing a profiling.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1706,20 +1706,33 @@ class SimpleArray
...
             const size_t nbytes = static_cast<size_t>(storage_size(m_shape, m_stride)) * ITEMSIZE;
-            if (nbytes != buffer->nbytes())
+            if (data_offset % ITEMSIZE != 0)
+            {
+                throw std::runtime_error("SimpleArray: data offset must be divisible by item size");
+            }
+            if (data_offset > buffer->nbytes() || nbytes != buffer->nbytes() - data_offset)
             {
                 throw std::runtime_error(
                     std::format("SimpleArray: shape byte count {} differs from buffer {}",
```
tigercosmos: The message seems to be misleading now.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc: The lambda is really long.  It is good to move it to a static helper.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc: This is `/* */` so it's OK.  It's good to use block comments here.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: This should be fixed.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: This can be address in this PR or use a follow-up one.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: Agree.  The problem is the overload `data(size_t)` that always use `data()`.  There should not be arithmetics in `data()`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2038,8 +2089,16 @@ class SimpleArray
     /* Backdoor */
     value_type const & data(size_t it) const { return data()[it]; }
     value_type & data(size_t it) { return data()[it]; }
-    value_type const * data() const { return buffer().template data<value_type>(); }
-    value_type * data() { return buffer().template data<value_type>(); }
+    value_type const * data() const
+    {
+        value_type const * ret = buffer().template data<value_type>();
+        return ret ? ret + m_data_offset / ITEMSIZE : nullptr;
```
yungyuc: I don't get it.  This prevents to retrieve the true pointer from outside and looks wrong.

If you need an alternate address, is there a reason that you do not use an additional helper?

yungyuc (changes requested):
- [ ] `SimpleArray::data()` can only return the pointer without any arithmetics.  And clarify why it is changed to use the arithmetic.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
tigercosmos: I see. I rarely see this pattern in the codebase, though.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1610,6 +1610,24 @@ class SimpleArray
...
+    friend constexpr ArrayOrder operator|(ArrayOrder lhs, ArrayOrder rhs) noexcept
+    {
+        return static_cast<ArrayOrder>(std::to_underlying(lhs) | std::to_underlying(rhs));
+    }
+
+    friend constexpr ArrayOrder & operator|=(ArrayOrder & lhs, ArrayOrder rhs) noexcept
+    {
+        lhs = lhs | rhs;
+        return lhs;
+    }
```
ThreeMonth03, line 1629: Create enum class `ArrayOrder` for contiguous flags.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2284,10 +2355,78 @@ class SimpleArray
     /// bytes) to skip for advancing an index in the corresponding dimension.
     sshape_type m_stride;
 
+    /// Pointer to the array's logical origin within the buffer.
+    value_type * m_logical_data = nullptr;
```
ThreeMonth03, line 2363: Add a new member `m_logical_data` to distinguish the beginning of the buffer storage from the logical origin of the array. These pointers are identical for ordinary positive-stride arrays, but differ for negative-stride views.

For example:

```python
import numpy as np
import solvcon

base = np.array([10, 20, 30, 40], dtype=np.float64)
view = base[::-1]
```

The NumPy layout is:

```text
address       base+0    base+8    base+16    base+24
value           10        20         30         40
                 ^                                ^
          base.ctypes.data                 view.ctypes.data
```

Previously, `ConcreteBuffer` and `data()` started at `view.ctypes.data`. The original NumPy allocation contains the preceding elements, so the negative stride can appear to work before cloning. However, the cloned allocation starts at the copied logical origin:

```text
Previous cloned buffer:

outside allocation     cloned allocation
        X            |       40
        ^                    ^
   data() - 1              data()
    index 1                index 0
```

For the reversed view, accessing index 1 calculates `data() - 1`, which points before the cloned allocation and may cause a segmentation fault.

The new layout keeps `data()` at the beginning of the complete storage span and records the logical origin separately:

```text
address              p       p+8      p+16      p+24
storage             10        20        30        40
                     ^                             ^
                   data()                   m_logical_data

logical index         3         2         1         0
                                   stride = -1 <---
```

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2284,10 +2355,78 @@ class SimpleArray
...
+    return static_cast<size_t>(m_logical_data - data_ptr) * ITEMSIZE;
+}
+
+template <typename T>
+void SimpleArray<T>::update_data_pointers(size_t data_offset)
+{
+    value_type * data_ptr = m_buffer ? m_buffer->template data<value_type>() : nullptr;
+    m_logical_data = data_ptr ? data_ptr + data_offset / ITEMSIZE : nullptr;
+    m_body = calc_body(m_logical_data, m_stride, m_nghost);
+}
```
ThreeMonth03, line 2385: To avoid repetitive calculation in `data()` and `logical_data()`, the pointer `m_logical_data` is computed when `SimpleArray` is constructed.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1706,39 +1724,64 @@ class SimpleArray
...
             }
+            size_t const available_nbytes = buffer->nbytes() - data_offset;
+            if (nbytes != available_nbytes)
+            {
+                throw std::runtime_error(
+                    std::format("SimpleArray: shape byte count {} differs from available buffer byte count {} at data offset {}",
+                                nbytes,
+                                available_nbytes,
+                                data_offset));
+            }
```
ThreeMonth03, line 1760: Fix the misleading comment. There are three branches for checking legal input now.

gtests/test_nopython_buffer.cpp
ThreeMonth03, line 1: Test whether cloned `SimpleArray` works with negative stride.

ThreeMonth03 (commented):
@yungyuc @tigercosmos Thanks for reviewing this pull request, please take a look again.
- [x] Fix misleading error message.
- [x] Enumerate contiguous flags by enum class.
- [x] SimpleArray::data() can only return the pointer without any arithmetics. And clarify why it is changed to use the arithmetic.
The previous `data()` precomputed sliced offset when constructing `SimpleArray`. However, it might cause segmentation fault in some circumstance, see [this comment](https://github.com/solvcon/solvcon/pull/1101/changes#r3609005345). Therefore, I decided not to precompute sliced offset when constructing `SimpleArray`, and compute the offset when accessing data. Thus, you saw a lot of arithmetic in last review.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: This explanation is clear.

yungyuc (approved):
The code looks OK, but the system has grown to be pretty complex.  To keep it simple enough to be named simple array, we will need to review the design.

I would like to merge the code to get it running for a period of time.  After that, we should meet online to discuss how to describe the status quo.

PR #1104, opened 2026-07-14
Extract the shared platform axis and fix Doxygen comment markers

tigercosmos:
Preparatory cleanup split out of the shortcut-manager work (#1070): moves the duplicated PlatformId axis out of theme.hpp and keymap.hpp into a shared Qt-free platform.hpp both cores include, and converts the multi-line /// briefs in the pilot cores to /** */ as STYLE.md requires.

Related to #1070.

### DIFF
--- a/cpp/solvcon/pilot/platform.hpp
+++ b/cpp/solvcon/pilot/platform.hpp
@@ -1,0 +1,51 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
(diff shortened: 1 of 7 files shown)


cpp/solvcon/pilot/platform.hpp
```diff
@@ -0,0 +1,51 @@
...
+ * result passed in, which keeps this header free of Qt.
+ *
+ * @ingroup group_domain
+ */
+
+namespace solvcon
+{
+
+/// The platform a per-platform table or capability record is written for.
+enum class PlatformId
```
yungyuc, line 24: Good catch.  I didn't notice it was in cpp.

PR #1108, opened 2026-07-14
Run AddressSanitizer on Windows via the gtest binary

tigercosmos:
This adds a way to run the project under AddressSanitizer on Windows.

AddressSanitizer has to run through the C++ gtest binary, because that binary links the sanitizer runtime directly and starts it at launch, while a normal python.exe loading an instrumented module starts it too late and crashes.

build.ps1 gets a -Sanitize switch that builds and runs the gtest suite under AddressSanitizer, reusing the existing -Gtest path with USE_SANITIZER=ON and BUILD_QT=OFF.

A new sanitizer_windows CI job does the same on an MSVC toolchain, gated to the nightly schedule like the Linux sanitizer job.

The build.ps1 path was checked on a Windows Server 2022 machine with MSVC 19.44, where all 213 gtest tests passed under AddressSanitizer with no diagnostics.

The sanitizer_windows CI job was also run once on this pull request and passed on a windows-2022 runner: https://github.com/solvcon/solvcon/actions/runs/29443432649

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -554,8 +554,105 @@ jobs:
           CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3) -DUSE_SANITIZER=ON"
         echo "::endgroup::"
 
+  sanitizer_windows:
+
+    needs: [check_skip]
+
+    # The MSVC counterpart of the sanitizer job. MSVC implements only
+    # AddressSanitizer (no leak or undefined), and the instrumented gtest
+    # binary links the ASan runtime directly so ASan initializes at process
+    # start; a stock python.exe loading an instrumented .pyd cannot. It is
+    # slow, so it runs on the nightly schedule only.
+    if: |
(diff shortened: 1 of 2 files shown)


.github/workflows/devbuild.yml
```diff
@@ -554,8 +554,105 @@ jobs:
...
+    needs: [check_skip]
+
+    # The MSVC counterpart of the sanitizer job. MSVC implements only
+    # AddressSanitizer (no leak or undefined), and the instrumented gtest
+    # binary links the ASan runtime directly so ASan initializes at process
+    # start; a stock python.exe loading an instrumented .pyd cannot. It is
+    # slow, so it runs on the nightly schedule only.
+    if: |
+      needs.check_skip.outputs.skip != 'true' &&
+      github.event_name == 'schedule' && vars.MMGH_NIGHTLY == 'enable'
```
tigercosmos, line 568: Only run Windows sanitize build in nightly.

yungyuc (commented):
It will be used to debug the nightly crash.

PR #1110, opened 2026-07-15
Add missing C++ class, struct, and namespace ending marks

tigercosmos:
Sweep cpp/ and gtests/ to add the STYLE.md-required class, struct, and namespace ending marks, correct marks that named the wrong type or keyword, and normalize stale wording ("//", "end of ...", dropped "end"), plus add a cpp-style-review skill rule so future diffs get checked; the spacetime/ reference implementation is left untouched.

### DIFF
--- a/cpp/solvcon/mesh/StaticMesh.hpp
+++ b/cpp/solvcon/mesh/StaticMesh.hpp
@@ -255,7 +255,7 @@ private:
 
     class ctor_passkey
     {
-    };
+    }; /* end class ctor_passkey */
 
 public:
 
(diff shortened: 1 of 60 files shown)


cpp/solvcon/mesh/StaticMesh.hpp
```diff
@@ -255,7 +255,7 @@ class StaticMesh
 
     class ctor_passkey
     {
-    };
+    }; /* end class ctor_passkey */
```
yungyuc, line 258: Simple things like a passkey can be allowed to omit the ending mark.  But let agent work.

yungyuc (commented):
Good agent work

PR #1114, opened 2026-07-15
Route undo, redo, and camera reset through the shortcut manager

tigercosmos:
Step 3 of the pilot shortcut work: undo, redo, and camera-reset bindings now resolve from the keymap tables through RShortcutManager::applyTo instead of hardcoded setShortcut calls, reset keeps its widget context and add3DWidget attachment, and new Python tests cover the live bindings. Related to #1070.

### DIFF
--- a/cpp/solvcon/pilot/RManager.cpp
+++ b/cpp/solvcon/pilot/RManager.cpp
@@ -20,7 +20,6 @@
 #include <QAction>
 #include <QActionGroup>
 #include <QColor>
-#include <QKeySequence>
 #include <QVBoxLayout>
 #include <QWidget>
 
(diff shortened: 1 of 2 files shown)


cpp/solvcon/pilot/RManager.cpp
```diff
@@ -420,7 +419,7 @@ void RManager::setUpEditMenuItems() const
         { undoCanvas(); },
         m_menuModel);
     undo_action->setObjectName("edit.undo");
-    undo_action->setShortcut(QKeySequence::Undo);
+    m_shortcutManager->applyTo(undo_action, ShortcutCommand::Undo);
```
tigercosmos, line 422: Use new shortcut system.

PR #1116, opened 2026-07-15
Tune pilot shortcut accents per platform

tigercosmos:
This continues the pilot shortcut work by teaching each platform which chords are reserved (camera WASD, plus the OS window-manager and macOS system keys) and by giving Console a Mac-specific chord (Primary+Shift+C) so Primary+Grave stays free for window cycling. Linux and Windows keep Primary+` for the console.

The conflict checks now run over an explicit binding table, so one set of declared and resolved helpers serves both the live tables and tests that rebind a command to prove a clash with Undo is reported. A standard action that resolves to no sequence on a platform (Quit has none on Windows) is treated as bound rather than missing, so every vocabulary command still resolves there.

The Python boundary gains resolve_all_shortcuts and the platform capability record, and both bindings share the resolved-shortcut and conflict-pair converters instead of repeating them.

Related to #1070.

### DIFF
--- a/cpp/solvcon/pilot/RShortcutManager.cpp
+++ b/cpp/solvcon/pilot/RShortcutManager.cpp
@@ -78,12 +78,26 @@ Qt::Key toQtKey(Key key)
         return Qt::Key_Escape;
     case Key::Grave:
         return Qt::Key_QuoteLeft;
+    case Key::Tab:
+        return Qt::Key_Tab;
+    case Key::Space:
+        return Qt::Key_Space;
+    case Key::F4:
+        return Qt::Key_F4;
     case Key::A:
         return Qt::Key_A;
+    case Key::D:
+        return Qt::Key_D;
     case Key::I:
         return Qt::Key_I;
     case Key::P:
         return Qt::Key_P;
+    case Key::S:
+        return Qt::Key_S;
+    case Key::W:
+        return Qt::Key_W;
+    case Key::Z:
+        return Qt::Key_Z;
     }
     throw std::logic_error("Unexpected key");
 }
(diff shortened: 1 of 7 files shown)


cpp/solvcon/pilot/keymap.cpp
```diff
@@ -35,35 +36,42 @@ std::vector<ShortcutBinding> tableWith(std::vector<ShortcutBinding> extra)
          KeyChord{KeyMod::Primary | KeyMod::Shift, Key::P},
          ShortcutContext::Window},
         {ShortcutCommand::New2DCanvas, StandardAction::New, ShortcutContext::Window},
+
+        // Like VSCode, the pilot binds Console to physical Ctrl+Grave on every platform.
+        {ShortcutCommand::Console, KeyChord{KeyMod::Control, Key::Grave}, ShortcutContext::Window},
```
tigercosmos, line 41: Special one.

cpp/solvcon/pilot/keymap.cpp
```diff
@@ -165,8 +173,51 @@ ShortcutBinding const * bindingFor(PlatformId platform, ShortcutCommand command)
 
 ShortcutCapabilities capabilitiesFor(PlatformId platform)
 {
-    bool const appMenu = platform == PlatformId::Mac;
-    return {platform, appMenu, {}};
+    ShortcutCapabilities caps;
+    caps.platform = platform;
+    caps.movesItemsToApplicationMenu = platform == PlatformId::Mac;
+    // Plain W/A/S/D belong to RDomainWidget::keyPressEvent on every platform.
+    caps.reservedSequences = {
```
tigercosmos, line 180: Reserve some shortcuts in advanced.

cpp/solvcon/pilot/RShortcutManager.cpp
```diff
@@ -78,12 +78,26 @@ Qt::Key toQtKey(Key key)
         return Qt::Key_Escape;
     case Key::Grave:
         return Qt::Key_QuoteLeft;
+    case Key::Tab:
```
tigercosmos, line 81: More keys. Maybe we will extract these key declarations into another file.

gtests/test_nopython_pilot_keymap.cpp
```diff
@@ -111,6 +162,23 @@ TEST(PilotKeymapConflicts, DefaultTablesHaveNoDeclaredConflicts)
     }
 }
 
+TEST(PilotKeymapConflicts, DeclaredCheckerReportsDuplicateCuratedChord)
```
tigercosmos, line 165: Test the conflict.

cpp/solvcon/pilot/RShortcutManager.cpp
yungyuc: The settings in C++ will be the default fallback.  User-configuration will later go to Python, and then json.

yungyuc:
@j8xixo12 please take a look.

PR #1118, opened 2026-07-16
Add a terminal-style console on a guarded surface (step 1 of terminal)

### LINKED ISSUE #1023: Add terminal-like I/O to console

yungyuc:
Step 1 of issue #1023.

The pilot console presents the interpreter in two panes, a read-only transcript above a separate input box, so the prompt a user types against never sits next to what they type. Add a second console, RPythonTerminalDockWidget, that presents the same in-process interpreter as a terminal: one document where the prompt, the typed command, and the captured output interleave, with the committed transcript read-only and the current input editable after the prompt.

RPythonTerminalTextEdit holds the surface. An input-start anchor splits the document into a read-only head and an editable tail, and the key and mouse handlers keep every edit at or past the anchor. Typing or pasting into the head relocates the caret to the end, Backspace and Left stop at the anchor, and Home lands after the prompt. Enter runs the input through the shared interpreter and writes stdout and stderr back above the next prompt; Up and Down recall from the history file that both consoles share.

The manager builds the terminal next to the two-pane console, tabbed and hidden until opened, and a Terminal toggle sits under View/Panels beside the Console toggle. Both consoles run the one interpreter, so the terminal opens with the same handles-in-scope banner the two-pane console prints. The existing console is unchanged, so a user chooses either console from the menu.

Cover the widget with tests that drive it through synthesized key events. The events are built by hand with QKeyEvent and QApplication.sendEvent rather than QtTest, whose module is not importable on every platform the CI matrix covers. A bar-structure test checks that the Terminal toggle joins the other panels under View/Panels, and a widget test checks the terminal opens with the shared banner.

### DIFF
--- a/cpp/solvcon/pilot/RManager.hpp
+++ b/cpp/solvcon/pilot/RManager.hpp
@@ -19,6 +19,7 @@
 #include <solvcon/pilot/R2DWidget.hpp>
 #include <solvcon/pilot/RAction.hpp>
 #include <solvcon/pilot/RPythonConsoleDockWidget.hpp>
+#include <solvcon/pilot/RPythonTerminalDockWidget.hpp>
 
 #include <vector>
 
@@ -76,6 +77,8 @@ public:
 
     RPythonConsoleDockWidget * pycon() { return m_pycon; }
 
+    RPythonTerminalDockWidget * pyterm() { return m_pyterm; }
+
     QMainWindow * mainWindow() { return m_mainWindow; }
 
     QMdiArea * mdiArea() { return m_mdiArea; }
@@ -98,12 +101,14 @@ public:
     void reset();
 
     void toggleConsole();
+    void toggleTerminal();
 
 private:
 
     RManager();
 
     void setUpConsole();
+    void setUpTerminal();
     void setUpCentral();
     void setUpMenu();
 
@@ -161,6 +166,7 @@ private:
     RShortcutManager * m_shortcutManager = nullptr;
 
     RPythonConsoleDockWidget * m_pycon = nullptr;
+    RPythonTerminalDockWidget * m_pyterm = nullptr;
     QMdiArea * m_mdiArea = nullptr;
 
     /**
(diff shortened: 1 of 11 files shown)


cpp/solvcon/pilot/RManager.hpp
```diff
@@ -161,6 +166,7 @@ class RManager
     RShortcutManager * m_shortcutManager = nullptr;
 
     RPythonConsoleDockWidget * m_pycon = nullptr;
+    RPythonTerminalDockWidget * m_pyterm = nullptr;
```
yungyuc, line 169: The second interpreter interface.  It work independently from the first `m_pycon` console.

cpp/solvcon/pilot/RPythonTerminalDockWidget.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * The dockable terminal console.
+ *
+ * It runs each submitted command through the embedded interpreter and
+ * writes the captured stdout and stderr back onto the same surface, above
+ * the next prompt. It shares the persistent history file with the two-pane
+ * console, so recall spans both. The two-pane console is unaffected.
+ *
+ * @ingroup group_domain
+ */
+class RPythonTerminalDockWidget
```
yungyuc, line 105: Mimic the existing Python console but present in the interactive terminal style.

cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -834,6 +881,12 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRManager
                 {
                     return self.pycon();
                 })
+            .def_property_readonly(
+                "pyterm",
```
yungyuc, line 885: Expose to Python as `pyterm`.

tests/test_pilot_terminal.py
yungyuc, line 1: New test file for the terminal

PR #1119, opened 2026-07-16
Drive terminal multiline from the interpreter's push (step 2 of terminal)

### LINKED ISSUE #1023: Add terminal-like I/O to console

yungyuc:
Step 2 of issue #1023.

The terminal collects a multiline statement by splitting the input on newlines and running it at once, so it cannot show a continuation prompt as a block is entered. Make it collect the statement the way a read-eval-print loop does, one line per Enter, with the interpreter deciding when the block is complete.

Add a per-line entry through the stack: AppEnvironment.push_line feeds one line and reports whether the statement is still open without closing it, apputil.push_code and system.push_code expose it, and Interpreter::push_code surfaces the flag to the widget. reset_console abandons a partly entered block.

On Enter the terminal pushes the current line; an open statement switches
to the continuation prompt with the block's indentation carried forward, and a whitespace-only line closes the block the way a blank line does in the interactive interpreter. The completed statement is recorded in history as one entry, and Up and Down recall only at the primary prompt, not in the middle of a block.

Guard executeCommand and resetInput with the GIL: when Enter arrives through the Qt event dispatch the caller may not hold it, the same reason the completion path already acquires it.

### DIFF
--- a/cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
+++ b/cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
@@ -5,6 +5,8 @@
 
 #include <solvcon/pilot/RPythonTerminalDockWidget.hpp>
 
+#include <solvcon/pilot/RPythonSyntaxRules.hpp>
+
 #include <algorithm>
 
 #include <QFont>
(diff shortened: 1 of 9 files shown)


cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
```diff
@@ -257,11 +269,53 @@ void RPythonTerminalDockWidget::executeCommand()
     }
     m_python_redirect.deactivate();
 
+    if (!m_pending_statement.empty())
+    {
+        m_pending_statement += '\n';
+    }
+    m_pending_statement += input.toStdString();
+
+    if (more)
```
yungyuc, line 278: Handle multi-line input to the Python terminal.

PR #1120, opened 2026-07-16
Add terminal completion, call tips, search, highlighting (step 3 of terminal)

### LINKED ISSUE #1023: Add terminal-like I/O to console

yungyuc:
Step 3 of issue #1023.

Bring the terminal's editing comforts to parity with the two-pane console, wired onto the bounded input region and reusing the same backends.

Tab requests completion for the identifier prefix behind the caret, inserting a single match directly and showing a popup for several; a '(' after a callable raises its signature tip; and Ctrl-R runs reverse incremental search over the shared history, walking toward older matches on repeated presses. The completion and call-tip handlers take the GIL, since they run from the Qt event dispatch.

The syntax highlighter now paints only the input region: it takes an input-start provider and skips any block that ends at or before the anchor, so the committed transcript above stays uncolored, and executeCommand and resetInput rehighlight after the anchor moves. The two-pane console leaves the provider unset, so it still paints every block.

This mirrors the two-pane console's proven logic rather than sharing a base class. Unifying the two consoles is a worthwhile follow-up, but a larger and riskier change than this step should carry.

### DIFF
--- a/cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
+++ b/cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
@@ -9,12 +9,17 @@
 
 #include <algorithm>
 
+#include <QAbstractItemView>
+#include <QColor>
 #include <QFont>
 #include <QKeyEvent>
 #include <QMimeData>
 #include <QPalette>
+#include <QScrollBar>
 #include <QStandardPaths>
+#include <QTextBlock>
 #include <QTextCursor>
+#include <QToolTip>
 
 namespace solvcon
 {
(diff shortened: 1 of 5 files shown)


cpp/solvcon/pilot/RPythonTerminalDockWidget.cpp
```diff
@@ -90,11 +95,182 @@ void RPythonTerminalTextEdit::moveToEnd()
     setTextCursor(cursor);
 }
 
+void RPythonTerminalTextEdit::setCompleter(QCompleter * completer)
```
yungyuc, line 98: Code completion.

tests/test_pilot_terminal.py
```diff
@@ -170,6 +170,52 @@ def test_write_to_history_appends_above_the_prompt(self):
         # The in-progress input survives the external write.
         self.assertEqual(self.term.command, "keep")
 
+    def test_tab_completes_a_single_match(self):
```
yungyuc, line 173: Test code completion.

PR #1124, opened 2026-07-16
Speed up builds with precompiled third-party headers

tigercosmos:
Adds `SOLVCON_USE_PCH` (enabled by default) to precompile expensive pybind11 and Qt headers. Across five clean builds, compile CPU fell by about 6% (465s to 436s). Wall time at full parallelism changed from 51.0s to 50.2s, which is within run-to-run noise, so the main benefit is for core-limited or thermally constrained machines.

| rep (-j14) | PCH ON wall | PCH ON CPU | PCH OFF wall | PCH OFF CPU |
|---|---|---|---|---|
| 1 | 50.3s | 442.5s | 53.2s | 463.2s |
| 2 | 49.9s | 432.7s | 48.4s | 460.3s |
| 3 | 48.0s | 428.1s | 52.8s | 466.6s |
| 4 | 50.9s | 438.8s | 50.9s | 471.5s |
| 5 | 51.8s | 438.1s | 49.8s | 463.3s |
| mean | 50.2s | 436.0s | 51.0s | 465.0s |

| rep (-j4) | PCH ON wall | PCH OFF wall |
|---|---|---|
| 1 | 105.1s | 105.3s |
| 2 | 101.2s | 105.3s |

PCH is automatically disabled with clang-tidy, ccache, and sccache, keeping CI behavior unchanged. It includes only caster-neutral pybind11 headers, and `python/common.cpp` opts out to avoid a Qt `slots` macro conflict.

Verified with `make lint`, pytest (1588 passed), gtest (213 passed), pilot builds, and `BUILD_QT=OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,25 @@ if(SOLVCON_PROFILE)
     endif()
 endif()
 
+# Precompiled headers cut the repeated parse of the heavy third-party
+# headers (pybind11, Qt) in every translation unit. They do not compose
+# with compiler caches (ccache/sccache fall back to uncached real compiles
+# on PCH translation units) nor with clang-tidy (it cannot replay another
+# compiler's PCH), so force OFF whenever either is in play. The set()
+# shadows the cached option with a normal variable so the gate also holds
+# when an existing build tree is reconfigured into an incompatible mode.
+option(SOLVCON_USE_PCH "precompile heavy third-party headers" ON)
+file(REAL_PATH "${CMAKE_CXX_COMPILER}" SOLVCON_CXX_COMPILER_REALPATH)
+if(SOLVCON_USE_PCH
+   AND (USE_CLANG_TIDY
+        OR CMAKE_CXX_COMPILER_LAUNCHER
+        OR SOLVCON_CXX_COMPILER_REALPATH MATCHES "ccache|sccache"))
+    message(STATUS
+        "force SOLVCON_USE_PCH=OFF: clang-tidy or a compiler cache is on")
+    set(SOLVCON_USE_PCH OFF)
+endif()
+message(STATUS "SOLVCON_USE_PCH: ${SOLVCON_USE_PCH}")
+
 if(BUILD_METAL)
 include_directories(SYSTEM "${CMAKE_CURRENT_LIST_DIR}/thirdparty/install/metal-cpp")
 endif()
--- a/cpp/solvcon/CMakeLists.txt
+++ b/cpp/solvcon/CMakeLists.txt
@@ -293,4 +293,30 @@ else () # MSVC
     )
 endif () # MSVC
 
+# Only the caster-neutral pybind11 headers go into the PCH. Force-including
+# a caster-adding header (stl.h, functional.h, complex.h) would silently
+# change conversion semantics in translation units that deliberately omit
+# it.
+if (SOLVCON_USE_PCH)
+    target_precompile_headers(
+        solvcon_primary PRIVATE
+        <pybind11/pybind11.h>
+        <pybind11/numpy.h>
+        <pybind11/embed.h>
+    )
+    if (BUILD_QT)
+        target_precompile_headers(
+            solvcon_primary PRIVATE
+            <QtWidgets/QtWidgets>
+        )
+        # python/common.cpp includes the raw numpy C API headers, whose
+        # struct member named "slots" collides with the Qt keyword macro
+        # that the QtWidgets PCH force-includes.
+        set_source_files_properties(
+            ${CMAKE_CURRENT_SOURCE_DIR}/python/common.cpp
+            PROPERTIES SKIP_PRECOMPILE_HEADERS ON
+        )
+    endif () # BUILD_QT
+endif () # SOLVCON_USE_PCH
+
 # vim: set ff=unix fenc=utf8 nobomb et sw=4 ts=4 sts=4:
--- a/.github/workflows/pch_experiment.yml
+++ b/.github/workflows/pch_experiment.yml
@@ -1,0 +1,71 @@
+name: pch_experiment
+
+# Temporary workflow to validate the SOLVCON_USE_PCH build-time effect on
+# the real CI runners. It runs the same dependency setup as devbuild but
+# deliberately omits the ccache step, because PR 1124 force-disables the
+# PCH whenever a compiler cache is present. Each configuration is built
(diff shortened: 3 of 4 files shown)


tigercosmos:
@yungyuc  Please help take a look. Thanks

cpp/solvcon/CMakeLists.txt
```diff
@@ -293,4 +293,30 @@ else () # MSVC
     )
 endif () # MSVC
 
+# Only the caster-neutral pybind11 headers go into the PCH. Force-including
+# a caster-adding header (stl.h, functional.h, complex.h) would silently
+# change conversion semantics in translation units that deliberately omit
+# it.
+if (SOLVCON_USE_PCH)
```
tigercosmos, line 300: Flag for header precomilation

cpp/solvcon/CMakeLists.txt
```diff
@@ -293,4 +293,30 @@ else () # MSVC
...
+    target_precompile_headers(
+        solvcon_primary PRIVATE
+        <pybind11/pybind11.h>
+        <pybind11/numpy.h>
+        <pybind11/embed.h>
```
tigercosmos, line 305: Heavy headers.

CMakeLists.txt
```diff
@@ -199,6 +199,25 @@ if(SOLVCON_PROFILE)
     endif()
 endif()
 
+# Precompiled headers cut the repeated parse of the heavy third-party
```
yungyuc, line 202: It will be good to have the comment block to be less verbose.

CMakeLists.txt
```diff
@@ -199,6 +199,25 @@ if(SOLVCON_PROFILE)
...
+# headers (pybind11, Qt) in every translation unit. They do not compose
+# with compiler caches (ccache/sccache fall back to uncached real compiles
+# on PCH translation units) nor with clang-tidy (it cannot replay another
+# compiler's PCH), so force OFF whenever either is in play. The set()
+# shadows the cached option with a normal variable so the gate also holds
+# when an existing build tree is reconfigured into an incompatible mode.
+option(SOLVCON_USE_PCH "precompile heavy third-party headers" ON)
+file(REAL_PATH "${CMAKE_CXX_COMPILER}" SOLVCON_CXX_COMPILER_REALPATH)
+if(SOLVCON_USE_PCH
+   AND (USE_CLANG_TIDY
```
yungyuc, line 212: It's good to avoid PCH in clang-tidy or other static analyzers.

yungyuc (commented):
> Adds `SOLVCON_USE_PCH` (enabled by default) to precompile expensive pybind11 and Qt headers. Across five clean builds, compile CPU fell by about 6% (465s to 436s). Wall time at full parallelism changed from 51.0s to 50.2s, which is within run-to-run noise, so the main benefit is for core-limited or thermally constrained machines.

The benefit is marginal.  I don't feel we should do it right away. 

I'd like to have input from @chestercheng @ExplorerRay.

Having said that, it's OK to make PCH as an opt-in.

CMakeLists.txt
```diff
@@ -199,6 +199,25 @@ if(SOLVCON_PROFILE)
...
 endif()
 
+# Precompiled headers cut the repeated parse of the heavy third-party
+# headers (pybind11, Qt) in every translation unit. They do not compose
+# with compiler caches (ccache/sccache fall back to uncached real compiles
+# on PCH translation units) nor with clang-tidy (it cannot replay another
+# compiler's PCH), so force OFF whenever either is in play. The set()
+# shadows the cached option with a normal variable so the gate also holds
+# when an existing build tree is reconfigured into an incompatible mode.
+option(SOLVCON_USE_PCH "precompile heavy third-party headers" ON)
```
chestercheng, line 209: The result shows only a small improvement, also our current CI does not test the PCH enabled path.
This feels more like an experimental optimization.

Maybe keep it `OFF` by default for now, and enable it after we have better results and CI coverage.

yungyuc: > Maybe keep it `OFF` by default for now, and enable it after we have better results and CI coverage.

This suggests the change does not add noticeable (by most contributors) value.

tigercosmos: Let me do more experiment on different VM with different cores.

ExplorerRay: > Maybe keep it OFF by default for now, and enable it after we have better results and CI coverage.

I agree with this. If we really want to turn it on by default, I think we should add CI to check whether the default setting is available or not.

yungyuc: Before adding it to CI, it should demonstrate value.

tigercosmos:
I ran a from-scratch A/B of `SOLVCON_USE_PCH` on the GitHub-hosted runners to see how the build-time effect holds up across toolchains.

Method: a temporary workflow builds each configuration from scratch with `BUILD_QT=ON` and no ccache (so the PCH gate stays active), timing only the `_solvcon_py` target since `solvcon_primary` is the one target the precompiled headers attach to. The A/B therefore isolates the PCH and nothing else. Three reps per configuration, `--parallel 2` to match CI's `MAKE_PARALLEL=-j2`. Wall time is a monotonic clock; CPU time is `getrusage(RUSAGE_CHILDREN)` (user + system across every compiler child), the portable equivalent of `/usr/bin/time`.

Means over 3 reps:

| runner | compiler | wall ON | wall OFF | CPU ON | CPU OFF | PCH effect |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| macos-26 | Apple clang | 269.6s | 277.8s | 506.0s | 530.9s | -3.0% wall, -4.7% CPU |
| ubuntu-24.04 | GCC 16 | 568.5s | 507.0s | 1065.8s | 960.7s | +12.1% wall, +10.9% CPU |

The effect splits by compiler, and the sign is consistent in every rep on each runner (not noise):

- Apple clang gets the modest saving the description promises, close to the ~6% CPU from the Apple-silicon benchmark.
- GCC 16 is a net regression on both wall and CPU.

Since `SOLVCON_USE_PCH` defaults ON for every compiler, the default currently slows the Linux/GCC build path. It may be worth gating the default by compiler, or defaulting OFF for GCC.

One caveat on magnitude: because the measurement isolates `solvcon_primary` plus the Python module, these percentages are the effect on the PCH-targeted translation units, not the whole-project build. Including the non-PCH pilot and gtest sources would shrink the numbers but not flip the sign.

tigercosmos:
Summary: this PR doesn't give us much benifit, so I will close it.

PR #1125, opened 2026-07-16
Note the C++ 120-char line width in the agent style guidance

tigercosmos:
STYLE.md already says C++ has no hard line-width limit and that a line is good under about 120 characters, but the agent-facing guidance did not repeat it, so generated C++ was being wrapped down to the Python 79/80 width for no reason.

This states the rule where code generation actually reads it: a C++ bullet in CLAUDE.md and a line-width note in the cpp-style-review skill, both saying a line up to about 120 characters is fine and that C++ should not be wrapped to the Python limit. The Python 79-character rule is unchanged.

[skip-ci]

### DIFF
--- a/.claude/skills/cpp-style-review/SKILL.md
+++ b/.claude/skills/cpp-style-review/SKILL.md
@@ -66,6 +66,12 @@ it briefly but don't re-implement the check here.
 - Trivial bodies (single `return`, single assignment) as one-liners.
 
 **Line economy**
+- **Line width (C++).** STYLE.md sets no hard width limit for C++; a line is
+  fine up to ~120 characters. Do not wrap C++ down to the Python 79/80 limit,
+  and do not flag a line under ~120 for width. Prefer one readable statement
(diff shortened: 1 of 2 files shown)


.claude/skills/cpp-style-review/SKILL.md
```diff
@@ -66,6 +66,12 @@ it briefly but don't re-implement the check here.
 - Trivial bodies (single `return`, single assignment) as one-liners.
 
 **Line economy**
+- **Line width (C++).** STYLE.md sets no hard width limit for C++; a line is
```
tigercosmos, line 69: Prevent AI outputs short lines.

.claude/skills/cpp-style-review/SKILL.md
```diff
@@ -66,6 +66,12 @@ it briefly but don't re-implement the check here.
...
+  fine up to ~120 characters. Do not wrap C++ down to the Python 79/80 limit,
```
yungyuc, line 70: Good to remind to not follow Python for C++ code.

PR #1127, opened 2026-07-16
Combine the per-OS mmdv dependency build scripts into one cross-platform script

### LINKED ISSUE #871: Combine the per-OS mmdv build scripts into one cross-platform script
modmesh ships two separate mmdv (development-environment) build scripts ([`contrib/dependency/ubuntu2404/build-mmdv-ubuntu24.sh` for Ubuntu](https://github.com/solvcon/modmesh/blob/c6b4e6484f69bc6c248e6fde5de1ab844809aba3/contrib/dependency/ubuntu2404/build-mmdv-ubuntu24.sh), [`contrib/dependency/macos26/build-mmdv-macos26.sh` for macOS](https://github.com/solvcon/modmesh/blob/c6b4e6484f69bc6c248e6fde5de1ab844809aba3/contrib/dependency/macos26/build-mmdv-macos26.sh)) that are nearly identical but maintained independently.  Consider to have a single mmdv build script that works on both Ubuntu and macOS, so the shared behavior lives in one place and only the genuinely platform-specific parts differ.

Pros:
- Single source of truth: the shared logic is written, reviewed, and tested once instead of twice.
- No more duplicating the same change across two files; this removes the drift that appears whenever an edit lands in one script but not the other.
- Lower long-term maintenance cost as dependency versions and recipes evolve.

Cons:
- One file covering two platforms is harder to read than a focused per-OS script.
- Per-OS branching adds conditional complexity, and a careless edit to the shared path can regress both platforms at once.
- Every change must be validated on both platforms, and anything that references the old per-OS script paths (CI, docs) needs a migration path.

This issue is created to track the discussion and implementation.  Patches are welcome.

tigercosmos:
This replaces the per-OS dependency build scripts with one cross-platform entrance, contrib/dependency/build-scdv.sh, plus a small per-OS module (ubuntu2404/platform.sh and macos26/platform.sh).

The entrance holds the shared build logic and detects the platform from uname, while each module carries only its real differences: prerequisites, compiler and BLAS choices, Qt's libclang source, and the activation script.

The old per-OS scripts are removed, and the Windows script, the Linux CI action, and the docs are updated to match.
This stops the trouble of keeping three near-identical scripts in sync.

There is no automated test yet, so it was checked by hand on two VMs.
The full build ran to the end on Ubuntu 24.04, and on macOS Sequoia it built everything with NumPy, SciPy, and PySide6 all importing and running.

Testing also found and fixed two bugs.
- The Qt step ignored the job count and used every core, running the small Ubuntu VM out of memory.
- An old, unused package set (nose, boto, paramiko) was still installed and broke the macOS build, so it was dropped.

Related to #871.


contrib/dependency/macos26/platform.sh
```diff
@@ -0,0 +1,409 @@
+#!/bin/bash
```
tigercosmos: Extract the platform-specifc functions

contrib/dependency/build-scdv.sh
```diff
@@ -1,61 +1,69 @@
...
-# built with --with-readline=editline, which uses libedit (libedit-dev).
-# libmpdec is not packaged on Ubuntu 24.04; Python is built with
-# --with-system-libmpdec=no to use its bundled copy.
+# Before running this script, install the system prerequisites. The script
+# never invokes the OS package manager itself; run it with --print-deps to see
+# the exact commands, review them, and run them yourself.
 #
 # Usage:
-#   ./build-scdv-ubuntu24.sh
+#   ./build-scdv.sh
```
tigercosmos: Single entry now.

contrib/dependency/ubuntu2404/platform.sh
yungyuc, line 1: The build script should stay single file.  Sourcing and calling an additional file breaks its self-containment.

The scdv scripts changed to use the single-file approach from the multi-file used by the old devenv script set.  I do not want to go back.

yungyuc (changes requested):
- [x] Use single file per platform without sourcing or calling an additional script in the repository.

contrib/dependency/ubuntu2404/platform.sh
tigercosmos: Fixed.

tigercosmos:
@yungyuc Please take a look. I’m not sure whether you prefer having a single file for all Unix-like operating systems, or keeping a common entry file while maintaining a separate full script for each platform.

yungyuc:
> I’m not sure whether you prefer having a single file for all Unix-like operating systems, or keeping a common entry file while maintaining a separate full script for each platform.

I was still evaluating.  Having a single file helps reduce the difference between mac and ubuntu, but makes the script pretty long (~1,500 lines).

I can use more opinions.  Is length a concern?  @chestercheng @iefiru @j8xixo12 @KHLee529 @ThreeMonth03 @ExplorerRay @rockleona ?

tigercosmos:
@yungyuc I guess we can use the current design and merge it?

contrib/dependency/build-scdv.sh
```diff
@@ -79,177 +75,677 @@
...
-      *" ${name} "*) ;;
-      *) echo "warning: --skip '${name}' is not a known package (known: ${SCDV_KNOWN_PKGS})" >&2 ;;
-    esac
-    SCDV_SKIP_LIST="${SCDV_SKIP_LIST} ${name}"
-  done
+# Define the platform-specific parts inline (nothing is sourced).  Each branch
+# sets SCDV_OS_TAG and the plat_* hook functions; only the selected platform's
+# block is defined.
+case "${SCDV_OS}" in
+  ubuntu)
```
yungyuc, line 135: This part sets up variables and functions for ubuntu.

contrib/dependency/build-scdv.sh
```diff
@@ -79,177 +75,677 @@
...
+  unset SCDV_BASE SCDV_USRDIR
+  unset -f scdv_deactivate
+}
+ACTIVATE_EOF
+  # Do not set the execution bit of "${target}"; it should be sourced.
+  echo "wrote activation script: ${target}"
+}
+
+    ;;
+  macos)
```
yungyuc, line 465: This part sets up variables and functions for macos.

contrib/dependency/build-scdv.sh
```diff
@@ -79,177 +75,677 @@
...
-    esac
-    SCDV_SKIP_LIST="${SCDV_SKIP_LIST} ${name}"
-  done
+# Define the platform-specific parts inline (nothing is sourced).  Each branch
+# sets SCDV_OS_TAG and the plat_* hook functions; only the selected platform's
+# block is defined.
+case "${SCDV_OS}" in
+  ubuntu)
+
+#
```
yungyuc, line 137: I dislike this empty comment line, but it does no harm except visual ungliness.

contrib/dependency/build-scdv.sh
```diff
@@ -370,14 +866,157 @@ ACTIVATE_EOF
   echo "wrote activation script: ${target}"
 }
 
+    ;;
+  *) echo "unknown SCDV_OS='${SCDV_OS}'; expected ubuntu or macos" >&2
+     exit 1 ;;
+esac
+
+SCDV_WRITE_ACTIVATE_ONLY=0
```
yungyuc, line 874: Common logic after platform specific code.

contrib/dependency/build-scdv.sh
```diff
@@ -611,23 +1250,7 @@ build_python() {
...
-    # -headerpad_max_install_names so install_name_tool has room to rewrite
-    # paths later (a no-op when we don't, harmless when we do).
-    local ldflags="-Wl,-headerpad_max_install_names"
-    ldflags="${ldflags} -Wl,-rpath,${SCDV_USRDIR}/lib"
-    ldflags="${ldflags} -L${SCDV_USRDIR}/lib"
-    if [ -n "${BREW_PREFIX}" ] && [ -d "${BREW_PREFIX}/opt/xz/lib" ] ; then
-      ldflags="${ldflags} -L${BREW_PREFIX}/opt/xz/lib"
-    fi
-    export LDFLAGS="${ldflags} ${LDFLAGS:-}"
+    plat_python_env
```
yungyuc, line 1253: The platform-specific code is moved up.

yungyuc (approved):
> @yungyuc I guess we can use the current design and merge it?

Right.  We can give it a try.

ExplorerRay:
In my opinion, the combination of different OS in build script is for modularize code and reuse common parts. If this operation can do this, I will prefer to merge into single file.  

It's a long script to review, but since changes like this will be rare, the length is an acceptable trade-off for me.

KHLee529:
I personally don't think the length is a big concern, and making all into single entry points looks good to me. Everyone can start their exploration from the same point.

j8xixo12:
In my opinion, while the initial review effort will be higher, but a lot of code can be reused, and build script should not be modified frequently. Since this is a one-time effort, putting per OS build process into a single script makes sense to me.

PR #1135, opened 2026-07-17
Run the agent turn off the Qt thread to keep the pilot responsive

tigercosmos:
The Agent Console applied each turn synchronously on the Qt main thread, so the whole pilot froze for the length of the backend call, a subprocess or HTTP round trip, while the model was thinking.

This moves only the backend call onto an AgentBackendWorker QThread. That call touches neither Qt nor the world, so it is safe off the main thread. The reply returns through a signal and the resulting commands are applied back on the main thread, serialized with the user's own edits through the event loop, so nothing races the world's C++ buffers. The prompt box locks for the duration of a turn so turns never overlap, but the canvas stays live and the user can keep drawing while the model thinks. On quit the panel joins an in-flight worker so its thread is never destroyed while still running.

While a turn runs the console shows an animated "Agent is working ..." line and clears it when the turn ends, so a slow or misconfigured backend reads as busy and its eventual reply or error is always shown, rather than the console looking frozen and silent.

A command that names a shape the user removed while the model was thinking now fails cleanly as a not-live shape and shows in the transcript rather than crashing the batch. This is the basic race workaround; a richer scheme that diffs the scene the model saw against the live world, and responsive cancellation that reaches the backend subprocess, are noted as follow-ups.

To support this without duplicating logic, AgentSession.run_turn is split into record_prompt, complete_turn, and fail_turn so the GUI can record the prompt, run the backend off the main thread, and finish the turn later. run_turn keeps its previous behavior for headless callers.

Related to #966.

### DIFF
--- a/solvcon/pilot/_agent_gui.py
+++ b/solvcon/pilot/_agent_gui.py
@@ -13,7 +13,7 @@ prompt is an independent single turn.
 
 from itertools import zip_longest
 
-from PySide6.QtCore import Qt, Signal
+from PySide6.QtCore import Qt, QCoreApplication, QThread, QTimer, Signal
 from PySide6.QtWidgets import (QWidget, QVBoxLayout, QHBoxLayout, QDockWidget,
                                QLabel, QComboBox, QTextEdit, QLineEdit,
                                QPushButton)
@@ -22,9 +22,41 @@ from ..agent import AgentSession, available_backends
 from . import _gui_common
 
 __all__ = [  # noqa: F822
+    'AgentBackendWorker',
     'AgentConsoleWidget',
     'AgentPanel',
 ]
+
+
+class AgentBackendWorker(QThread):
+    """Run one backend call off the Qt thread.
+
+    Only the backend call runs here, the slow subprocess or HTTP round trip;
+    it reads neither Qt nor the world, so it is safe off the main thread.  The
+    reply returns through :attr:`succeeded` (a ``BackendResponse``) or
+    :attr:`failed` (an error string); the owning panel applies the commands and
+    repaints on the main thread, where the connected slots run.
+    """
+
+    succeeded = Signal(object)
+    failed = Signal(str)
+
+    def __init__(self, backend, prompt, scene_context, tool_surface,
+                 parent=None):
+        super().__init__(parent)
+        self._backend = backend
+        self._prompt = prompt
+        self._scene_context = scene_context
+        self._tool_surface = tool_surface
+
+    def run(self):
+        try:
+            response = self._backend.send(
+                self._prompt, self._scene_context, self._tool_surface)
+        except Exception as exc:
+            self.failed.emit("%s: %s" % (type(exc).__name__, exc))
+        else:
+            self.succeeded.emit(response)
 
 
 class AgentConsoleWidget(QWidget):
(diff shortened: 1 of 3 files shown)


solvcon/pilot/_agent_gui.py
```diff
@@ -147,26 +218,57 @@ def _ensure_panel(self):
...
         self._panel.set_busy(True)
-        try:
-            turn = session.run_turn(prompt)
-        except Exception as exc:
-            self._panel.append_message("agent", "[error] %s" % exc)
-        else:
-            self._panel.append_message("agent", self._format_turn(turn))
-            if widget is not None:
+        self._panel.start_working()
+        self._worker = AgentBackendWorker(
```
tigercosmos, line 237: Create a worker for the agent session so we don't block the main thread.

tests/test_pilot_agent.py
```diff
@@ -91,18 +126,94 @@ def test_drives_the_active_canvas_world(self):
...
+        self._select_echo(panel)
+        panel._input.setText("draw a circle")
+        panel._emit()
+        worker = feature._worker
+        self.assertIsNotNone(worker)
+        feature._join_worker()
+        self.assertTrue(worker.isFinished())
+        QCoreApplication.processEvents()
+
+    def test_stale_by_id_command_fails_cleanly(self):
```
tigercosmos, line 199: A race condition test case.

solvcon/pilot/_agent_gui.py
```diff
@@ -147,26 +218,59 @@ def _ensure_panel(self):
...
+            session.tool_surface(), parent=self._panel)
+        self._worker.succeeded.connect(self._on_backend_succeeded)
+        self._worker.failed.connect(self._on_backend_failed)
+        self._worker.finished.connect(self._on_worker_finished)
+        self._worker.start()
+
+    def _on_backend_succeeded(self, response):
+        """Apply the backend's commands to the world and repaint the canvas.
+        """
+        # TODO(#966): potential race condition here. capture a world revision
```
tigercosmos, line 248: TODO

solvcon/pilot/_agent_gui.py
```diff
@@ -147,26 +218,59 @@ def _ensure_panel(self):
...
+        self._panel.append_message("agent", self._format_turn(turn))
+        widget = self._active_widget
+        if widget is not None:
+            try:
                 widget.requestRepaint()
-        finally:
-            self._panel.set_busy(False)
-            self._panel.clear_input()
+            except RuntimeError:
+                # TODO: the widget may have been deleted while the backend was
```
tigercosmos, line 257: TODO

solvcon/pilot/_agent_gui.py
```diff
@@ -22,11 +22,43 @@
 from . import _gui_common
 
 __all__ = [  # noqa: F822
+    'AgentBackendWorker',
     'AgentConsoleWidget',
     'AgentPanel',
 ]
 
 
+class AgentBackendWorker(QThread):
```
yungyuc, line 31: We will need a thread manager for all workers including an agent one like this.

@tigercosmos Do you have an idea when should we start to design it?

tigercosmos: @yungyuc I concur. I guess we can think it much later when there are more threads.

yungyuc: We should plan for the manager when we are about to have the third thread.  The most likely scenario is the computing thread for the solver.

PR #1139, opened 2026-07-17
Add the Agent Window command family for canvas windows

tigercosmos:
The Agent Draw family lets an AI backend edit what lives in the World. This adds the family that manages the canvas windows around it: open a canvas, list the open ones, activate, close, and save an image.

Agent Window drives a duck-typed window manager, the surface the GUI WindowController will later bind to the pilot's RManager and QMdiArea (tests bind an in-memory fake). The vocabulary is new_canvas, list_windows, activate_window, close_window, and save_image. Commands that take a window id fail as a clean result when the id is unknown rather than leaking a KeyError from the manager or the Qt layer.

The family mirrors solvcon.agent.draw exactly: one Command subclass per op, a private CommandSet that derives the JSON Schema, validators, and MCP-style tool definitions, and an Executor that binds the set to a target. The re-export boilerplate that lets a module double as its own command set had been copied verbatim into every front-end package; it now lives once in _command.install_command_api, and the draw package is migrated onto the shared helper in this PR.

Wiring the family into AgentSession alongside Draw needs a multi-family dispatcher and stays a follow-up. The Agent View family, which drives the 2D pan/zoom transform (ViewTransform2dFp64), builds on the same install_command_api helper and lands in its own PR stacked on this one.

The command set is covered by tests/test_agent_window_command.py, which exercises schema derivation, JSON conformance, and each command end to end against an in-memory fake manager.

Related to #966.

### DIFF
--- a/solvcon/agent/_command.py
+++ b/solvcon/agent/_command.py
@@ -235,6 +235,45 @@ class CommandSet:
         return grouped
 
 
+# CommandSet attributes a fronting module re-exports so it doubles as its set.
+COMMAND_API = (
+    "apply_defaults",
+    "categories",
+    "command_from_tool_call",
+    "command_schemas",
+    "commands",
+    "commands_by_category",
+    "description",
+    "result_schemas",
+    "schema",
+    "title",
+    "tool_definitions",
+    "validate_command",
+    "validate_result",
+    "validate_script",
+)
+
+
+def install_command_api(namespace, command_set):
+    """Wire a module to double as ``command_set``'s command API.
+
+    Pass the module's ``globals()`` as ``namespace``; installs
+    ``__getattr__``/``__dir__`` and returns the name tuple for ``__all__``.
+    """
+    def __getattr__(name):
+        if name in COMMAND_API:
+            return getattr(command_set, name)
+        raise AttributeError(
+            f"module {namespace['__name__']!r} has no attribute {name!r}")
+
+    def __dir__():
+        return sorted(set(namespace) | set(COMMAND_API))
+
+    namespace["__getattr__"] = __getattr__
+    namespace["__dir__"] = __dir__
+    return COMMAND_API
+
+
 @dataclasses.dataclass
 class CommandResult:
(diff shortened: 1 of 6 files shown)


solvcon/agent/window/command.py
```diff
@@ -0,0 +1,124 @@
...
+    if not any(w["id"] == window_id for w in manager.list_windows()):
+        raise _cmd.CommandError(f"no open window with id {window_id}")
+
+
+_command_set = _cmd.CommandSet(
+    "Agent Window command", "Any single command in the Agent Window schema.")
+
+
+@_command_set.register
+class NewCanvas(_cmd.Command):
```
tigercosmos, line 60: The agent will create a new window if no window is opened.

solvcon/agent/window/command.py
```diff
@@ -0,0 +1,124 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The Agent Window command family: the canvas windowing surface for an agent.
```
tigercosmos, line 6: Add window command family.

tests/test_agent_window_command.py
```diff
@@ -0,0 +1,240 @@
...
+    commands_by_category,
+    result_schemas,
+    schema,
+    tool_definitions,
+    validate_command,
+    validate_script,
+)
+
+
+class FakeWindowManager:
```
tigercosmos, line 28: For headless testing. The real manager support will be in the following PR.

solvcon/agent/_command.py
```diff
@@ -235,6 +235,45 @@ def commands_by_category(self):
...
+    "schema",
+    "title",
+    "tool_definitions",
+    "validate_command",
+    "validate_result",
+    "validate_script",
+)
+
+
+def install_command_api(namespace, command_set):
```
tigercosmos, line 257: Helper funciton to install common API for the command's module.

Issue #1136, opened 2026-07-17
Improve Agent Draw output quality: coordinate framing, richer primitives, and a feedback loop


tigercosmos:
## Summary

Asking the Agent Draw feature (Claude backend) to "draw a truck", "draw a house", "draw Elon Musk", or "draw solvcon" produces poor results in the Pilot canvas. I drove the real `claude (cli)` backend through the actual `AgentSession` -> `Executor` -> `World` pipeline and rendered the output both auto-fit and in the true Pilot view. The surprising result: the geometry the model emits is usually reasonable, but it lands in the wrong place because the model is never told the canvas coordinate frame. The dominant problem is framing, not the shapes. Secondary problems are a thin prompt, a starved and bloated context (one-line scene summary next to a 25k-char tool dump), a wireframe-only primitive set, and the absence of any visual feedback loop, which together cap how good the output can get.

## Goal

Make natural-language drawing produce output that is centered, correctly scaled, and legible by default, then raise the quality ceiling with richer primitives, better context, and a self-correction loop. Concretely: a first-time user who types "draw a truck" should see a well-framed truck without touching pan or zoom, "draw solvcon" should render legible text, and a follow-up like "make the truck bigger" should work because the model can see what is on the canvas.

## Root cause analysis

The Pilot 2D canvas puts the world origin at the center of the widget with zoom 1 and +Y up (`R2DWidget::resetView` plus the pan recentering in `R2DWidget::updateWorld`, and `ViewTransform2d`). The model, however, is told none of this. `AgentBackend._INSTRUCTIONS` in `solvcon/agent/_backend.py` is four sentences ("You drive a 2D drawing canvas ...") that never mention the coordinate system, and `AgentSession.scene_context` in `solvcon/agent/_core.py` reports only the shape count and types. So the model falls back to an SVG/screen habit and draws inside a first-quadrant box such as x in [0, 150] or [0, 400]. In the real canvas that box sits in the upper-right quadrant, off-center, and often partly clipped. After a turn the Agent panel calls only `widget.requestRepaint()` (`solvcon/pilot/_agent_gui.py`), so nothing ever frames what was drawn.

[...]

PR #1140, opened 2026-07-17
Add polygon and polyline drawing primitives to World

tigercosmos:
This is step 1 of a three-step series that grows the Agent Draw primitive set from issue #1136 (polygon/polyline, then text, then arc and scale). Each step is a separate sub-500-LOC PR against master; the next opens after this one lands.

A polyline is an open chain of straight segments through a vertex list; a polygon is the closed chain, with a final segment back to the first vertex. Both are segment-backed shapes, so they translate, rotate, remove, and undo through the existing shape machinery, and they let the agent draw true silhouettes instead of stitching rectangles.

Binds add_polyline/add_polygon on World, adds the matching Agent Draw commands, and covers the open/closed segment counts and the minimum-vertex validation.

Related to #1136.

### DIFF
--- a/cpp/solvcon/universe/World.hpp
+++ b/cpp/solvcon/universe/World.hpp
@@ -22,6 +22,7 @@
 #include <solvcon/universe/rtree.hpp>
 
 #include <algorithm>
+#include <array>
 #include <cmath>
 #include <limits>
 #include <optional>
@@ -55,6 +56,8 @@ enum class ShapeType : uint8_t
     SQUARE = 6, ///< Specialization of RECTANGLE with equal side lengths.
     ELLIPSE = 7,
     CIRCLE = 8, ///< Specialization of ELLIPSE with equal radii.
+    POLYLINE = 9, ///< Open chain of straight segments.
+    POLYGON = 10, ///< Closed chain of straight segments.
 }; /* end of enum class ShapeType */
 
 inline std::string shape_type_name(ShapeType st)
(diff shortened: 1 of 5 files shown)


cpp/solvcon/universe/World.hpp
```diff
@@ -416,6 +421,18 @@ class World
      */
     int32_t add_bezier_shape(bezier_type const & bezier);
 
+    /**
+     * Add an open chain of straight segments through the given vertices.
+     * Needs at least two vertices; consecutive vertices form one segment each.
+     */
+    int32_t add_polyline(std::vector<std::array<T, 2>> const & vertices);
```
tigercosmos, line 428: New API.

cpp/solvcon/universe/World.hpp
```diff
@@ -416,6 +421,18 @@ class World
...
+
+    /**
+     * Add a closed chain of straight segments through the given vertices, with
+     * a final segment back to the first. Needs at least three vertices.
+     */
+    int32_t add_polygon(std::vector<std::array<T, 2>> const & vertices);
```
tigercosmos, line 434: New API.

cpp/solvcon/universe/World.hpp
```diff
@@ -55,6 +56,8 @@ enum class ShapeType : uint8_t
     SQUARE = 6, ///< Specialization of RECTANGLE with equal side lengths.
     ELLIPSE = 7,
     CIRCLE = 8, ///< Specialization of ELLIPSE with equal radii.
+    POLYLINE = 9, ///< Open chain of straight segments.
```
yungyuc, line 59: Please reference the coding system of other drawing systems.  I am thinking to keep straight lines in a group and curvilinear in another.

It is for future development.

PR #1092, opened 2026-07-13
Stop master CI runs from cancelling each other

yungyuc:
An older CI run is cancelled by a latter one when merged to master, because the lint, devbuild, and nouse_install workflows share a concurrency group keyed on the pull request number when one is present and on the ref otherwise.

Gate the cancellation on the ref instead, so a push to master is never cancelled.

PR #1142, opened 2026-07-17
Give each master CI run its own concurrency group

yungyuc:
This is an additional fix for master CI cancallation because PR #1092 was not a complete fix.

The lint, devbuild, and nouse_install workflows shared the concurrency group refs/heads/master across every push to master, relying on cancel-in-progress being false there to preserve each merge's result. That is not enough. With cancel-in-progress disabled GitHub serialises runs in a group and, when a new run queues behind the one in progress, cancels whatever run was previously pending. A merge that lands while an earlier master run is still going therefore drops that earlier run.

Key master pushes on the commit SHA so each merge lands in its own group and never queues behind or cancels another master run. Pull requests still key on the PR number and other branches on the ref, so their superseded runs cancel as before.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -8,10 +8,14 @@ on:
     - cron: '34 17 * * *'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  # A push to master must never cancel an earlier master run, so every
-  # merge keeps its own CI result. Superseded pull request and branch
-  # runs still cancel.
+  # Key each master push on its commit SHA so every merge lands in its own
+  # concurrency group. Sharing one group serializes master runs, and with
+  # cancel-in-progress disabled GitHub still cancels a pending run when the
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -8,10 +8,14 @@ on:
...
-  # merge keeps its own CI result. Superseded pull request and branch
-  # runs still cancel.
+  # Key each master push on its commit SHA so every merge lands in its own
+  # concurrency group. Sharing one group serializes master runs, and with
+  # cancel-in-progress disabled GitHub still cancels a pending run when the
+  # next merge queues behind the one in progress, so a later merge would
+  # drop an earlier merge's result. A unique group per SHA keeps every
+  # master run independent. Pull requests key on the PR number and other
+  # branches on the ref, so their superseded runs still cancel.
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || (github.ref == 'refs/heads/master' && github.sha) || github.ref }}
```
yungyuc, line 18: For master branch, use sha as group, not branch ref name.

Issue #1126, opened 2026-07-16
Define SimpleArray reshape semantics for strided and ghost arrays


ThreeMonth03:
`SimpleArray::reshape()` reuses the existing buffer and assigns new C-contiguous strides. For a strided source, this can silently change the logical element order.

Reproduced on `master` at `980e8883`.

## Reproduce

```python
import numpy as np
import solvcon

ndarr = np.arange(6, dtype="float64").reshape((2, 3))
view = ndarr.T
sarr = solvcon.SimpleArrayFloat64(array=view)

expected = view.reshape((6,))
actual = sarr.reshape((6,)).ndarray

print("expected:", expected)
print("actual:  ", actual)
np.testing.assert_array_equal(actual, expected)
```

## Output

```text
expected: [0. 3. 1. 4. 2. 5.]
actual:   [0. 1. 2. 3. 4. 5.]
AssertionError:
Arrays are not equal
Mismatched elements: 4 / 6 (66.7%)
```

`reshape()` should preserve the source logical element order, or reject a layout when a zero-copy reshape cannot represent it. The intended behavior should be defined for C-contiguous, F-contiguous, arbitrary-strided, negative-strided, and ghost arrays.

This decision should include whether to retain `SimpleArray(shape, buffer, data_offset)`. The overload is currently used only by `reshape()` and combines the existing logical origin with newly calculated C-contiguous strides.

Related to #856.

yungyuc:
Good catch.

tigercosmos:
I will fix this

PR #1145, opened 2026-07-18
Preserve logical element order in SimpleArray reshape

tigercosmos:
SimpleArray::reshape() recomputed C-contiguous strides over the shared buffer and discarded the source layout, so reshaping a non-C-contiguous array (transposed, F-contiguous, or strided) silently returned the raw buffer order rather than the logical element order.

reshape() now returns a zero-copy view only when the source is C-contiguous and its buffer holds exactly the logical elements; otherwise it copies the elements in C order into a new dense buffer, matching numpy's reshape(order='C'). An array with ghost cells is rejected because reshaping across the ghost/body split of the first axis has no well-defined element mapping.

Related to #1126.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2027,20 +2027,62 @@ public:
         }
     }
 
+    /**
+     * Reshape to a new shape, preserving the logical element order.
+     *
+     * Returns a zero-copy view when the source is C-contiguous and its
+     * buffer holds exactly the logical elements; otherwise copies them in
+     * C order into a new dense buffer, matching numpy's reshape(order='C').
+     * Rejects an array with ghost cells, whose ghost/body split has no
+     * well-defined element mapping.
+     *
+     * @param shape The new shape; its element count must match the source.
+     * @return A view when zero-copy is possible, otherwise a copy.
+     */
     template <typename U>
     SimpleArray<U> reshape(shape_type const & shape) const
     {
-        return SimpleArray<U>(shape, m_buffer, logical_data_offset());
+        if (has_ghost())
+        {
+            throw std::runtime_error(
+                std::format("SimpleArray: cannot reshape an array with {} ghost cells", m_nghost));
+        }
+        validate_shape(shape);
+
+        // numpy reshape keeps the type; we allow a cross-type reshape but only to a type of the same item size,
+        // so the element count is what must match either way.
+        if constexpr (!std::is_same_v<U, value_type>)
+        {
+            if (sizeof(U) != ITEMSIZE)
+            {
+                throw std::runtime_error(
+                    std::format("SimpleArray: cannot reshape to a different item size {} vs {}", sizeof(U), ITEMSIZE));
+            }
+        }
+        check_reshape_size(shape);
+
+        // Reuse the buffer only when its logical region holds exactly the
+        // logical element and is C-contiguous; otherwise the buffer order
+        // differs from the logical order, so copy in C order instead.
+        const size_t offset = logical_data_offset();
+        const bool exact_buffer = m_buffer && m_buffer->nbytes() - offset == size() * ITEMSIZE;
+        if (is_c_contiguous() && exact_buffer)
+        {
+            return SimpleArray<U>(shape, m_buffer, offset);
+        }
+        SimpleArray row_major(m_shape);
+        copy_logical_into(row_major);
+        return SimpleArray<U>(shape, row_major.m_buffer);
     }
 
     SimpleArray reshape(shape_type const & shape) const
     {
-        return SimpleArray(shape, m_buffer, logical_data_offset());
+        return reshape<value_type>(shape);
     }
 
     SimpleArray reshape() const
     {
-        return SimpleArray(m_shape, m_buffer, logical_data_offset());
+        return reshape(m_shape);
     }
 
     void swap(SimpleArray & other) noexcept
@@ -2366,6 +2408,21 @@ private:
             return 0;
         }
         return shape[0] * stride[0];
+    }
+
+    /// Reject a target shape whose element count differs from the source.
+    void check_reshape_size(shape_type const & shape) const
+    {
+        ssize_t target = 1;
+        for (ssize_t const dim : shape)
+        {
+            target *= dim;
+        }
+        if (static_cast<size_t>(target) != size())
+        {
+            throw std::runtime_error(
+                std::format("SimpleArray: cannot reshape size {} into size {}", size(), target));
+        }
     }
 
     /// Contiguous data buffer for the array.
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -76,6 +76,42 @@ TEST(SimpleArray, abs)
     EXPECT_EQ(brr.sum(), 10.0);
 }
 
+TEST(SimpleArray, reshape_cross_type_preserves_logical_order)
+{
+    using namespace solvcon;
+
+    SimpleArray<int64_t> array(small_vector<ssize_t>{2, 3}, 0);
+    int64_t v = 0;
+    for (auto & it : array)
+    {
+        it = v++;
+    }
+
+    array.transpose(false);
+    auto result = array.reshape<uint64_t>(small_vector<ssize_t>{6});
+
+    const uint64_t expected[6] = {0, 3, 1, 4, 2, 5};
+    for (ssize_t i = 0; i < 6; ++i)
+    {
+        EXPECT_EQ(result(i), expected[i]);
+    }
+}
+
+TEST(SimpleArray, reshape_rejects_mismatched_size)
+{
+    using namespace solvcon;
+
+    SimpleArray<uint64_t> array(small_vector<ssize_t>{3}, 0);
+
+    // A cross-type reshape to a different item size has no well-defined
+    // element mapping and is rejected.
+    EXPECT_THROW(array.reshape<uint32_t>(small_vector<ssize_t>{6}), std::runtime_error);
+
+    // A same-type reshape must keep the element count.
+    EXPECT_THROW(array.reshape(small_vector<ssize_t>{5}), std::runtime_error);
+    EXPECT_THROW(array.reshape(small_vector<ssize_t>{}), std::runtime_error);
+}
+
 TEST(SimpleArray, iterator)
 {
     using namespace solvcon;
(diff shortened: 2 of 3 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1960,14 +1960,36 @@ class SimpleArray
...
+     * @param shape The new shape; its element count must match the source.
+     * @return A view when C-contiguous, otherwise a copy.
+     */
     SimpleArray reshape(shape_type const & shape) const
     {
-        return SimpleArray(shape, m_buffer);
+        if (has_ghost())
+        {
+            throw std::runtime_error(
+                std::format("SimpleArray: cannot reshape an array with {} ghost cells", m_nghost));
```
tigercosmos, line 2048: Don't support ghost (for now).

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1960,14 +1960,36 @@ class SimpleArray
...
+        }
+        // Reuse the buffer only when it holds exactly the logical element
+        const bool exact_buffer = m_buffer && size() * ITEMSIZE == m_buffer->nbytes();
+        if (is_c_contiguous() && exact_buffer)
+        {
+            return SimpleArray(shape, m_buffer);
+        }
+        return reshape_by_copy(shape);
```
tigercosmos: Meat.

tests/test_buffer.py
```diff
@@ -282,6 +282,67 @@ def test_SimpleArray(self):
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
         self.assertEqual((2, 2, 2, 3), sarr.reshape((2, 2, 2, 3)).shape)
 
+    def test_SimpleArray_reshape_transposed(self):
```
tigercosmos, line 285: Test case from the original issue with more test cases.

tigercosmos (commented):
@ThreeMonth03  Please take a look, thanks!

yungyuc:
@KHLee529 please help take a look.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1960,14 +1960,36 @@ class SimpleArray
...
+        {
+            throw std::runtime_error(
+                std::format("SimpleArray: cannot reshape an array with {} ghost cells", m_nghost));
+        }
+        // Reuse the buffer only when it holds exactly the logical element
+        const bool exact_buffer = m_buffer && size() * ITEMSIZE == m_buffer->nbytes();
+        if (is_c_contiguous() && exact_buffer)
+        {
+            return SimpleArray(shape, m_buffer);
+        }
```
ThreeMonth03, line 2072: It might exclude a zero-size array. In numpy, a zero-size array cannot be reshape.
```python3
np.empty((0,)).reshape(())
# ValueError: cannot reshape array of size 0 into shape ()
```

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1960,14 +1960,36 @@ class SimpleArray
         return SimpleArray<U>(shape, m_buffer);
     }
```
ThreeMonth03: `reshape<U>()` still ignores logical order. For a transposed `(2, 3)` array containing `0..5`, `reshape<double>({6})` returns `[0, 1, 2, 3, 4, 5]` instead of `[0, 3, 1, 4, 2, 5]`.

```cpp
SimpleArray<double> array({2, 3}, 0.0);
for (size_t i = 0; i < array.size(); ++i)
{
    array[i] = static_cast<double>(i);
}

array.transpose(false);
auto result = array.reshape<double>({6});

// result:   [0, 1, 2, 3, 4, 5]
// expected: [0, 3, 1, 4, 2, 5]
```

ThreeMonth03 (commented):
I notice that this pull request conflicts with pr #1101, so this pull request might be rebased and merged after #1101.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: It's good to follow numpy zero-sized arrays.

yungyuc (changes requested):
LGTM.  We can selectively follow numpy behaviors, but the zero-sized array is trivial and we can follow now:

- [x] Follow numpy zero-size array behavior.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2027,20 +2027,60 @@ class SimpleArray
...
+        {
+            throw std::runtime_error(
+                std::format("SimpleArray: cannot reshape an array with {} ghost cells", m_nghost));
+        }
+        validate_shape(shape);
+
+        // source and target shapes size must match
+        if constexpr (std::is_same_v<U, value_type>)
+        {
+            check_reshape_size(shape);
```
tigercosmos: If the value type is the same, check size.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2027,20 +2027,60 @@ class SimpleArray
...
+        }
+        else
+        {
+            check_reshape_bytes(shape, sizeof(U));
```
tigercosmos: If the value type is different, check the size of bytes.

gtests/test_nopython_buffer.cpp
```diff
@@ -76,6 +76,40 @@ TEST(SimpleArray, abs)
...
+
+    SimpleArray<int64_t> array(small_vector<ssize_t>{2, 3}, 0);
+    int64_t v = 0;
+    for (auto & it : array)
+    {
+        it = v++;
+    }
+
+    array.transpose(false);
+    auto result = array.reshape<uint64_t>(small_vector<ssize_t>{6});
```
tigercosmos, line 91: test `reshape` with different value type

tests/test_buffer.py
```diff
@@ -283,6 +282,87 @@ def test_SimpleArray(self):
...
+    def test_SimpleArray_reshape_ghost_raises(self):
+        sarr = solvcon.SimpleArrayFloat64((6,))
+        sarr.nghost = 2
+        with self.assertRaisesRegex(
+                RuntimeError,
+                "SimpleArray: cannot reshape an array with 2 ghost cells"
+        ):
+            sarr.reshape((3, 2))
+
+    def test_SimpleArray_reshape_zero_size(self):
```
tigercosmos, line 354: Add the zero size test case.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2027,20 +2027,60 @@ class SimpleArray
...
+        }
+        // Reuse the buffer only when its logical region holds exactly the
+        // logical element and is C-contiguous; otherwise the buffer order
+        // differs from the logical order, so copy in C order instead.
+        const size_t offset = logical_data_offset();
+        const bool exact_buffer = m_buffer && m_buffer->nbytes() - offset == size() * ITEMSIZE;
+        if (is_c_contiguous() && exact_buffer)
+        {
+            return SimpleArray<U>(shape, m_buffer, offset);
+        }
```
ThreeMonth03, line 2072: A cross-type reshape may enter this zero-copy branch even when the offset is invalid for U:
```c++
auto buffer = ConcreteBuffer::construct(5 * sizeof(uint32_t));
SimpleArray<uint32_t> array(
    small_vector<ssize_t>{4}, buffer, sizeof(uint32_t));
auto result = array.reshape<uint64_t>(
    small_vector<ssize_t>{2});
```
The byte counts match, but this throws because the offset is not divisible by `sizeof(uint64_t)`. Should this case fall back to the copy path?

ThreeMonth03 (commented):
@tigercosmos There is an edge case need to discuss.

cpp/solvcon/buffer/SimpleArray.hpp
tigercosmos: Thanks for the question. I notice that numpy doesn't allow `reshape` for different value types. Therefore I suggest that we can make a tradeoff where solvcon only accepts `reshape` iff the target data type has the same size of the source.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: Why not checking for bytes anyway for both same or different types?

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: @tigercosmos could you please also point out which line guards 0-sized array?

yungyuc (approved):
LGTM

cpp/solvcon/buffer/SimpleArray.hpp
tigercosmos: Yes, you can, but I explicitly separate them for a clearer error message (size for the same type, bytes for different types)

yungyuc: Good point.  Error messages matter.  I overlooked it.

cpp/solvcon/buffer/SimpleArray.hpp
tigercosmos: `test_SimpleArray_reshape_zero_size` already covers this case.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2022,20 +2027,62 @@ class SimpleArray
...
+        // numpy reshape keeps the type; we allow a cross-type reshape but only to a type of the same item size,
+        // so the element count is what must match either way.
+        if constexpr (!std::is_same_v<U, value_type>)
+        {
+            if (sizeof(U) != ITEMSIZE)
+            {
+                throw std::runtime_error(
+                    std::format("SimpleArray: cannot reshape to a different item size {} vs {}", sizeof(U), ITEMSIZE));
+            }
+        }
```
tigercosmos, line 2061: We allow different type reshape, but the size of the type needs to be the same.

gtests/test_nopython_buffer.cpp
```diff
@@ -76,6 +76,42 @@ TEST(SimpleArray, abs)
...
+
+TEST(SimpleArray, reshape_rejects_mismatched_size)
+{
+    using namespace solvcon;
+
+    SimpleArray<uint64_t> array(small_vector<ssize_t>{3}, 0);
+
+    // A cross-type reshape to a different item size has no well-defined
+    // element mapping and is rejected.
+    EXPECT_THROW(array.reshape<uint32_t>(small_vector<ssize_t>{6}), std::runtime_error);
```
tigercosmos, line 108: Example of the type itme size rejection.

Issue #1138, opened 2026-07-17
Agent command families accept non-finite numbers (NaN/Inf) and report success


tigercosmos:
The agent command families in solvcon.agent validate numeric arguments with plain JSON Schema number and positive-number fragments. jsonschema treats NaN and Inf as passing these constraints: exclusiveMinimum rejects a value only when it is <= the bound, and NaN <= 0 is false, so a non-finite value clears validation.

For mutating commands this corrupts the target while still reporting success. In the Window family, set_view with zoom NaN sets ViewTransform2dFp64.zoom to NaN and returns ok=True, and pan with an infinite delta poisons pan_x the same way. The Draw family shares the gap: add_circle with cx NaN and r Inf, and add_point with x Inf, both return ok=True. Some ops such as zoom_at and zoom_at_clamped are defended in C++ and silently no-op on non-finite input, but the command still reports ok=True, so an agent cannot tell the request did nothing.

The clean fix is a shared finite-number constraint applied to both families rather than per-command guards, so draw and window stay identical. Options include a reusable schema fragment that adds a finite check, or a small validation helper in solvcon.agent._command that every mutating command routes its numeric arguments through.

Found while adding the Window command family, which mirrors Draw. Related to #966.

yungyuc:
@tigercosmos Please elaborate when nan and inf are illegal.

Some numerical calculation may use nan and inf as sentinel and/or input.  Sometimes they are illegal but not always.

yungyuc:
@tigercosmos comments?

tigercosmos:
Currently, the scope is only 2D drawing, so nan and inf are invalid. When there are new numerical commands, we can further adjust the value validation based on different types of commands.

yungyuc:
Please elaborate in what scenarios nan and inf are invalid/illegal.  In many cases nan and inf are valid.  A simple example is to represent intersection point of two lines or line segments.  It is wrong to assume nan and inf are invalid by default.

tigercosmos:
> It is wrong to assume nan and inf are invalid by default.

Yes, you are right. An intersection calculation can use nan to present no unique intersection or use inf to present a point at infinity, so nan and inf should not be rejected globally.
However, for Draw, Window, and View commands, the values of the arguments should be always valid; therefore the proposed contrains should apply only to these specific commands.

yungyuc:
> However, for Draw, Window, and View commands, the values of the arguments should be always valid

[...]

PR #1146, opened 2026-07-18
Add the Agent View command family for the 2D view transform

tigercosmos:
The Agent Window family (#1139) manages the canvas windows. This adds the family that aims the 2D view inside one: panning, zooming, and mapping between screen and world coordinates.

Agent View lives in solvcon/agent/window/view.py and targets ViewTransform2dFp64, the pan/zoom transform Draw already references when it renders. Reads are get_view, screen_from_world, and world_from_screen; updates are pan, zoom_at, zoom_at_clamped, set_view, and reset_view. Positive-zoom and positive-factor bounds live in the schema so bad input fails as a clean result, and zoom_at_clamped additionally rejects inverted bounds (max_zoom < min_zoom) rather than silently no-opping.

The family mirrors the Window family: one Command subclass per op, a private CommandSet that derives the JSON Schema, validators, and MCP-style tool definitions, and an Executor that binds the set to a target. It reuses the _command.install_command_api helper introduced in #1139.

Like the Draw family, these commands still accept non-finite numbers (NaN, Inf) that clear JSON Schema validation and reach the transform; that shared gap is tracked in #1138 and is best addressed once at the schema layer for all families. The command set is covered by tests/test_agent_view_command.py, which exercises schema derivation, JSON conformance, and each command end to end against a real ViewTransform2dFp64.

Related to #966 and #1136

### DIFF
--- a/solvcon/agent/window/view.py
+++ b/solvcon/agent/window/view.py
@@ -1,0 +1,181 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The Agent View command family: driving a ``ViewTransform2dFp64`` from an agent.
+
+The commands aim the 2D view by panning, zooming, and mapping between screen
+and world coordinates. Screen coordinates are Qt pixels (+Y down); world
+coordinates are the ``World``'s native units, +Y up. The affine map is
+``screen = zoom * world + pan`` with a +Y flip on the y axis.
+"""
+
+from .. import _command as _cmd
+
+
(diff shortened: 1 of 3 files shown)


solvcon/agent/window/view.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The Agent View command family: driving a ``ViewTransform2dFp64`` from an agent.
```
tigercosmos, line 6: Add the view family commands so Agent can control the view.

PR #1102, opened 2026-07-14
Support negative strides in SimpleArray mdspan

ThreeMonth03:
`std::layout_stride` requires positive strides, so `SimpleArray::as_mdspan()` cannot expose reversed or mixed-sign views.

Add an internal `SignedStrideLayout` that normalizes signed offsets into a non-negative mapped span. Use it in `SimpleArray::as_mdspan()` and anchor the data handle at `logical_data() - origin_offset()` so logical index zero and the storage span remain consistent.

Related to #856.

PR #1147, opened 2026-07-19
Use signed-stride mapping in SimpleArray mdspan

ThreeMonth03:
`SimpleArray::as_mdspan()` currently uses `std::layout_stride`, which cannot represent negative strides. Reversed NumPy views therefore cannot be exposed as mdspan.

This PR uses `SignedStrideLayout` and anchors the mdspan data handle at `logical_data() - origin_offset()`. Tests cover a shifted positive-stride view, a reversed view, and a four-dimensional mixed-sign layout with padding.

This draft depends on #1101 and #1102. It will remain draft until both are merged.

Related to #856.


ThreeMonth03:
According to the discussion in #1102, this pull request will be intergrated to #1102 and close.

PR #1152, opened 2026-07-19
Remove unnecessary ghost metadata changes from array assignment

### LINKED ISSUE #1148: Preserve SimpleArray ghost metadata when array assignment fails
## Description

`ArrayPropertyHelper::broadcast_array_using_ellipsis()` and `broadcast_array_using_slice()` temporarily set `SimpleArray::nghost` to zero before copying data from a NumPy array, then restore it after the copy. If `TypeBroadcast::broadcast()` throws, the restoration is skipped and the destination array is left with modified ghost metadata.

The relevant code is in [array_common.hpp](https://github.com/solvcon/solvcon/blob/55365d263b7540b92d6f147e3b0f33bd0907ee75/cpp/solvcon/buffer/pymod/array_common.hpp#L164-L182).

## Reproducible example

```python
import numpy as np
import solvcon

sarr = solvcon.SimpleArrayFloat64(shape=(2, 2), value=0)
sarr.nghost = 1

print("nghost before:", sarr.nghost)

try:
    sarr[...] = np.ones((2, 2), dtype=np.complex128)
except RuntimeError as exc:
    print("error:", exc)

print("nghost after:", sarr.nghost)
```

## Bug result

```text
nghost before: 1
error: Cannot convert between complex and non-complex types
nghost after: 0
```

The failed assignment unexpectedly changes `nghost` from 1 to 0.

## Expected behavior

A failed assignment must leave the destination array metadata unchanged. In this example, `sarr.nghost` should remain 1 after the exception.

ThreeMonth03:
Array assignment resets `SimpleArray::nghost` around broadcasting, although broadcasting computes destination addresses from `logical_data()` and does not depend on `body()`. This unnecessary metadata change can become permanent when assignment raises an exception.

Remove the `nghost` mutation from the shared slice-assignment path and route ellipsis assignment through the same path.

Related to #1148.

### DIFF
--- a/cpp/solvcon/buffer/pymod/array_common.hpp
+++ b/cpp/solvcon/buffer/pymod/array_common.hpp
@@ -165,21 +165,7 @@ public:
     static void broadcast_array_using_ellipsis(SimpleArray<T> & arr_out, pybind11::array const & arr_in)
     {
         auto slices = make_default_slices(arr_out);
-
-        TypeBroadcast<T>::check_shape(arr_out, slices, arr_in);
-
-        const ssize_t nghost = arr_out.nghost();
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(0);
-        }
-
-        TypeBroadcast<T>::broadcast(arr_out, slices, arr_in);
-
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(nghost);
-        }
+        broadcast_array_using_slice(arr_out, slices, arr_in);
     }
 
     // FIXME: NOLINTNEXTLINE(readability-function-cognitive-complexity)
@@ -458,19 +444,7 @@ private:
                                             pybind11::array const & arr_in)
     {
         TypeBroadcast<T>::check_shape(arr_out, slices, arr_in);
-
-        const ssize_t nghost = arr_out.nghost();
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(0);
-        }
-
         TypeBroadcast<T>::broadcast(arr_out, slices, arr_in);
-
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(nghost);
-        }
     }
 }; /* end class ArrayPropertyHelper */
 
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -999,6 +999,21 @@ class SimpleArrayBasicTC(unittest.TestCase):
                     self.assertEqual(v, sarr[i, j, k])
                     v += 1
 
+    def test_SimpleArray_broadcast_error_preserves_nghost(self):
+        sarr = solvcon.SimpleArrayFloat64(shape=(2, 2), value=0)
+        rhs = np.ones((2, 2), dtype='complex128')
+
+        for key in (Ellipsis, np.s_[:, :]):
+            with self.subTest(key=key):
+                sarr.nghost = 1
+                with self.assertRaisesRegex(
+                        RuntimeError,
+                        r"^Cannot convert between complex and "
+                        r"non-complex types$"
+                ):
+                    sarr[key] = rhs
+                self.assertEqual(1, sarr.nghost)
+
     def test_SimpleArray_broadcast_slice_basic(self):
         ndarr_input = np.arange(
             1 * 2 * 3 * 4 * 5, dtype='float64').reshape((1, 2, 3, 4, 5))


cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -165,21 +165,7 @@ class ArrayPropertyHelper
...
-            arr_out.set_nghost(0);
-        }
-
-        TypeBroadcast<T>::broadcast(arr_out, slices, arr_in);
-
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(nghost);
-        }
+        broadcast_array_using_slice(arr_out, slices, arr_in);
```
ThreeMonth03, line 168: The deleted code are same as `broadcast_array_using_slice(arr_out, slices, arr_in);`.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -465,7 +451,18 @@ class ArrayPropertyHelper
...
+            TypeBroadcast<T>::broadcast(arr_out, slices, arr_in);
+        }
+        catch (...)
+        {
+            if (0 != nghost)
+            {
+                arr_out.set_nghost(nghost);
+            }
+            throw;
+        }
```
ThreeMonth03: If broadcast raises error, `nghost` would be modified to the original value.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/pymod/array_common.hpp
tigercosmos: Nice catch

cpp/solvcon/buffer/pymod/array_common.hpp
tigercosmos: Not sure if we can move `arr_out.set_nghost(nghost);` before `TypeBroadcast<T>::broadcast`. Current it seems okay.

However, I am thinking that we should define the contract first for the slice assignment and broadcasting for ghost.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -458,19 +444,7 @@ class ArrayPropertyHelper
...
-        {
-            arr_out.set_nghost(0);
-        }
-
         TypeBroadcast<T>::broadcast(arr_out, slices, arr_in);
-
-        if (0 != nghost)
-        {
-            arr_out.set_nghost(nghost);
-        }
```
ThreeMonth03, line 473: Remove `set_nghost()` here. There are legacy codes, since we assign slice value based on `logical_data()`, instead of `body()` now.

ThreeMonth03 (commented):
Thanks. TypeBroadcast computes destination addresses from `logical_data()`, so changing `nghost` only moves `body()` and does not affect assignment. I removed the unnecessary mutation and restoration.
I also found a separate indexing contract mismatch:
```
logical index:      -2        -1         0        1        2
                 [ ghost0 ][ ghost1 ][ body0 ][ body1 ][ body2 ]
                     ^                    ^
                     |                    |
                 sarr[0:1]             sarr[0]
```
Scalar indexing is body-relative, while slice assignment currently treats the slice as an offset from logical_data(). I propose making the first-axis slice domain `[-nghost, nbody)`, so `sarr[0:1]` targets the same element as `sarr[0]`. I will track that semantic change in a follow-up issue and keep this PR focused on removing the unnecessary metadata mutation.

tests/test_buffer.py
```diff
@@ -999,6 +999,17 @@ def test_SimpleArray_broadcast_ellipsis_dtype(self):
                     self.assertEqual(v, sarr[i, j, k])
                     v += 1
 
+    def test_SimpleArray_broadcast_error_preserves_nghost(self):
+        sarr = solvcon.SimpleArrayFloat64(shape=(2, 2), value=0)
+        rhs = np.ones((2, 2), dtype='complex128')
```
j8xixo12, line 1004: Why is the dtype different ?

ThreeMonth03: The dtype mismatch is intentional. The shapes must match so `check_shape()` succeeds, while `broadcast()` throws after the legacy code changes `nghost`. A shape mismatch would fail before that mutation, and matching dtypes would not exercise the error path.

tigercosmos (approved):
@ThreeMonth03  LGTM, but please fix the PR title and body.

cpp/solvcon/buffer/pymod/array_common.hpp
yungyuc: Do we still need `body()` after having `logical_data()`?

tests/test_buffer.py
```diff
@@ -999,6 +999,17 @@ def test_SimpleArray_broadcast_ellipsis_dtype(self):
...
+
+        for key in (Ellipsis, np.s_[:, :]):
+            with self.subTest(key=key):
+                sarr.nghost = 1
+                with self.assertRaises(RuntimeError):
```
yungyuc: Check for error messages along with the exception type.

yungyuc (changes requested):
- [x] In unit test, check for error messages along with exception type.

tests/test_buffer.py
```diff
@@ -999,6 +999,21 @@ def test_SimpleArray_broadcast_ellipsis_dtype(self):
...
+            with self.subTest(key=key):
+                sarr.nghost = 1
+                with self.assertRaisesRegex(
+                        RuntimeError,
+                        r"^Cannot convert between complex and "
+                        r"non-complex types$"
+                ):
+                    sarr[key] = rhs
+                self.assertEqual(1, sarr.nghost)
+
```
ThreeMonth03, line 1016: Check the error message for broadcast error.

ThreeMonth03 (commented):
- [x] In unit test, check for error messages along with exception type.

@yungyuc Pleae take a look.

yungyuc (approved):
LGTM

Issue #1149, opened 2026-07-19
Validate operand shapes before SimpleArray binary arithmetic


ThreeMonth03:
## Description

The array overloads of the `SimpleArray` arithmetic methods do not validate that their operands have matching shapes. They traverse the left-hand buffer and advance a pointer into the right-hand buffer without checking its shape or size.

As a result, operands with different shapes but the same element count are silently combined in physical buffer order. If the right-hand operand is smaller, the same traversal may read beyond its buffer.

The relevant implementations are in [SimpleArray.hpp](https://github.com/solvcon/solvcon/blob/55365d263b7540b92d6f147e3b0f33bd0907ee75/cpp/solvcon/buffer/SimpleArray.hpp#L704-L1069).

## Reproducible example

```python
import solvcon

lhs = solvcon.SimpleArrayFloat64(shape=(2, 3), value=1)
rhs = solvcon.SimpleArrayFloat64(shape=(3, 2), value=2)

result = lhs.add(rhs)

print("shape:", result.shape)
print("values:", [result[i, j] for i in range(2) for j in range(3)])
```

## Bug result

```text
shape: (2, 3)
values: [3.0, 3.0, 3.0, 3.0, 3.0, 3.0]
```

The operation succeeds even though `(2, 3)` and `(3, 2)` are incompatible shapes.

## Expected behavior

Binary array arithmetic should reject operands with different shapes and report both operand shapes, consistent with the existing comparison operations.

Shape validation should apply consistently to `add`, `sub`, `mul`, `div`, their in-place variants, and the explicit SIMD entry points.

ThreeMonth03:
@yungyuc Please assign this issue to me.

ThreeMonth03:
Fix by #1153

PR #1153, opened 2026-07-19
Validate SimpleArray arithmetic operand shapes

ThreeMonth03:
SimpleArray array arithmetic currently accepts mismatched shapes. Equal-sized operands are combined in physical buffer order, while a smaller right-hand operand can be read past its buffer.

Add shared same-shape validation to `add`, `sub`, `mul`, and `div`, including their in-place and SIMD variants. The error reports both operand shapes.

Related to #1149.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -705,6 +705,7 @@ public:
 
     A add(A const & other) const
     {
+        validate_same_shape(other, "add");
         return A(*static_cast<A const *>(this)).iadd(other);
     }
 
@@ -715,6 +716,7 @@ public:
 
     A sub(A const & other) const
     {
+        validate_same_shape(other, "sub");
         return A(*static_cast<A const *>(this)).isub(other);
     }
 
@@ -725,6 +727,7 @@ public:
 
     A mul(A const & other) const
     {
+        validate_same_shape(other, "mul");
         return A(*static_cast<A const *>(this)).imul(other);
     }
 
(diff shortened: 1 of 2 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2445,6 +2462,15 @@ SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::compare_with(
             format_shape(athis->shape()),
             format_shape(other.shape())));
     }
```
ThreeMonth03, line 2464: Add a helper function for checking shape. It will be refactored after we create an abstract helper class for traverse elements.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2432,9 +2450,8 @@ void SimpleArray<T>::validate_layout(shape_type const & shape,
 }
 
 template <typename A, typename T>
-template <typename Cmp>
-SimpleArray<bool> detail::SimpleArrayMixinCalculators<A, T>::compare_with(
-    A const & other, Cmp cmp, char const * op) const
+void detail::SimpleArrayMixinCalculators<A, T>::validate_same_shape(
```
tigercosmos, line 2453: I wonder if we should make this chekc inline.

tigercosmos (approved):
LGTM

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: It's OK either way.

PR #1158, opened 2026-07-21
Add axis support to SimpleArray argmin and argmax

### LINKED ISSUE #519: Implement `SimpleArray` member functions for searching

Yuli-ytd:
This PR add axis-aware reductions back to `SimpleArray::argmin()` and `SimpleArray::argmax()` while keeping the existing no-axis flat-index behavior.

The implementation now walks logical indices with stride-aware access, so flat and axis reductions work for non-contiguous NumPy-backed arrays, including sliced and negative-stride views. Empty reductions now raise `ValueError` through the Python binding.

The Python wrapper accepts `axis=None` for the existing scalar result and integer axes for reduced `SimpleArrayUint64` results. The 1D axis case returns a scalar to match NumPy behavior.

Tests cover flat reductions, empty arrays, non-contiguous views, negative axes, and axis reductions across multiple dimensions.

Related to #519 .

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -1369,42 +1369,21 @@ class SimpleArrayMixinSearch
 private:
 
     using internal_types = detail::SimpleArrayInternalTypes<T>;
+    using shape_type = typename internal_types::shape_type;
+    using sshape_type = typename internal_types::sshape_type;
+
+    ssize_t normalize_axis(ssize_t axis, char const * op) const;
+    shape_type reduced_shape(ssize_t axis) const;
 
 public:
 
     using value_type = typename internal_types::value_type;
 
-    size_t argmin() const
-    {
-        size_t min_index = 0;
-        value_type min_value = std::numeric_limits<value_type>::max();
-        auto athis = static_cast<A const *>(this);
-        for (size_t i = 0; i < athis->size(); ++i)
-        {
-            if (athis->data(i) < min_value)
-            {
-                min_value = athis->data(i);
-                min_index = i;
-            }
-        }
-        return min_index;
-    }
+    size_t argmin() const;
+    size_t argmax() const;
 
-    size_t argmax() const
-    {
-        size_t max_index = 0;
-        value_type max_value = std::numeric_limits<value_type>::lowest();
-        auto athis = static_cast<A const *>(this);
-        for (size_t i = 0; i < athis->size(); ++i)
-        {
-            if (athis->data(i) > max_value)
-            {
-                max_value = athis->data(i);
-                max_index = i;
-            }
-        }
-        return max_index;
-    }
+    SimpleArray<uint64_t> argmin(ssize_t axis) const;
+    SimpleArray<uint64_t> argmax(ssize_t axis) const;
 
     SimpleArray<uint64_t> argwhere() const;
 
@@ -3002,6 +2981,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+template <typename A, typename T>
+ssize_t detail::SimpleArrayMixinSearch<A, T>::normalize_axis(ssize_t axis, char const * op) const
+{
+    auto const * athis = static_cast<A const *>(this);
+    ssize_t const ndim = athis->ndim();
+
+    if (axis < 0)
+    {
+        axis += ndim;
+    }
+    if (axis < 0 || axis >= ndim)
+    {
+        throw std::invalid_argument(std::format(
+            "SimpleArray::{}(): axis {} is out of bounds for array of dimension {}",
+            op,
+            axis,
+            ndim));
+    }
+    if (athis->shape(axis) == 0)
+    {
+        throw std::invalid_argument(std::format(
+            "SimpleArray::{}(): axis {} has size 0, cannot compute",
+            op,
+            axis));
+    }
+
+    return axis;
+}
+
+template <typename A, typename T>
+typename detail::SimpleArrayMixinSearch<A, T>::shape_type
+detail::SimpleArrayMixinSearch<A, T>::reduced_shape(ssize_t axis) const
+{
+    auto const * athis = static_cast<A const *>(this);
+
+    shape_type result_shape;
+    for (ssize_t dim = 0; dim < athis->ndim(); ++dim)
+    {
+        if (dim != axis)
+        {
+            result_shape.push_back(athis->shape(dim));
+        }
+    }
+
+    return result_shape;
+}
+
+template <typename A, typename T>
+size_t detail::SimpleArrayMixinSearch<A, T>::argmin() const
+{
+    auto athis = static_cast<A const *>(this);
+
+    if (athis->size() == 0)
+    {
+        throw std::invalid_argument("SimpleArray::argmin(): "
+                                    "attempt to get argmin of an empty sequence");
+    }
+
+    auto const range = IndexRange(*athis);
+    sshape_type idx = range.first();
+
+    value_type min_value = athis->at(idx);
+    size_t min_index = 0;
+    size_t flat_index = 0;
+
+    do
+    {
+        value_type const current_value = athis->at(idx);
+        if (current_value < min_value)
+        {
+            min_value = current_value;
+            min_index = flat_index;
+        }
+        ++flat_index;
+    } while (range.next(idx));
+
+    return min_index;
+}
+
+template <typename A, typename T>
+size_t detail::SimpleArrayMixinSearch<A, T>::argmax() const
+{
+    auto athis = static_cast<A const *>(this);
+
+    if (athis->size() == 0)
+    {
+        throw std::invalid_argument("SimpleArray::argmax(): "
+                                    "attempt to get argmax of an empty sequence");
+    }
+
+    auto const range = IndexRange(*athis);
+    sshape_type idx = range.first();
+
+    value_type max_value = athis->at(idx);
+    size_t max_index = 0;
+    size_t flat_index = 0;
+
+    do
+    {
+        value_type const current_value = athis->at(idx);
+        if (current_value > max_value)
+        {
+            max_value = current_value;
+            max_index = flat_index;
+        }
+        ++flat_index;
+    } while (range.next(idx));
+
+    return max_index;
+}
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argmin(ssize_t axis) const
+{
+    auto athis = static_cast<A const *>(this);
+
+    axis = normalize_axis(axis, "argmin");
+    SimpleArray<uint64_t> result(reduced_shape(axis));
+
+    auto const out_range = IndexRange(result);
+    sshape_type out_idx = out_range.first();
+
+    do
+    {
+        sshape_type in_idx(athis->ndim(), 0);
+
+        for (ssize_t dim = 0, out_dim = 0; dim < athis->ndim(); ++dim)
+        {
+            if (dim == axis)
+            {
+                continue;
+            }
+            in_idx[dim] = out_idx[out_dim++];
+        }
+
+        in_idx[axis] = 0;
+        value_type min_value = athis->at(in_idx);
+        uint64_t min_index = 0;
+
+        for (ssize_t i = 1; i < athis->shape(axis); ++i)
+        {
+            in_idx[axis] = i;
+            value_type const current_value = athis->at(in_idx);
+            if (current_value < min_value)
+            {
+                min_value = current_value;
+                min_index = static_cast<uint64_t>(i);
+            }
+        }
+        result.at(out_idx) = min_index;
+    } while (out_range.next(out_idx));
+
+    return result;
+}
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argmax(ssize_t axis) const
+{
+    auto athis = static_cast<A const *>(this);
+
+    axis = normalize_axis(axis, "argmax");
+    SimpleArray<uint64_t> result(reduced_shape(axis));
+
+    auto const out_range = IndexRange(result);
+    sshape_type out_idx = out_range.first();
+
+    do
+    {
+        sshape_type in_idx(athis->ndim(), 0);
+
+        for (ssize_t dim = 0, out_dim = 0; dim < athis->ndim(); ++dim)
+        {
+            if (dim == axis)
+            {
+                continue;
+            }
+            in_idx[dim] = out_idx[out_dim++];
+        }
+
+        in_idx[axis] = 0;
+        value_type max_value = athis->at(in_idx);
+        uint64_t max_index = 0;
+
+        for (ssize_t i = 1; i < athis->shape(axis); ++i)
+        {
+            in_idx[axis] = i;
+            value_type const current_value = athis->at(in_idx);
+            if (current_value > max_value)
+            {
+                max_value = current_value;
+                max_index = static_cast<uint64_t>(i);
+            }
+        }
+        result.at(out_idx) = max_index;
+    } while (out_range.next(out_idx));
+
+    return result;
+}
 template <typename A, typename T>
 SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argwhere() const
 {
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -3050,7 +3050,7 @@ template <typename A, typename T>
 ssize_t detail::SimpleArrayMixinSearch<A, T>::unchecked_buffer_index(
     A const & array, shape_type const & idx)
 {
-    ssize_t const origin = static_cast<ssize_t>(
+    auto const origin = static_cast<ssize_t>(
         array.logical_data() - array.data());
     return origin + unchecked_logical_offset(array, idx);
 }
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -3034,8 +3034,7 @@ detail::SimpleArrayMixinSearch<A, T>::reduced_shape(ssize_t axis) const
 }
 
 template <typename A, typename T>
-ssize_t detail::SimpleArrayMixinSearch<A, T>::unchecked_logical_offset(
-    A const & array, shape_type const & idx)
+ssize_t detail::SimpleArrayMixinSearch<A, T>::unchecked_logical_offset(A const & array, shape_type const & idx)
 {
     ssize_t offset = 0;
     for (ssize_t dim = 0; dim < array.ndim(); ++dim)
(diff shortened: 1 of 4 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -630,8 +630,32 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                 {
+                    if (axis.is_none())
+                    {
+                        return py::cast(self.argmin());
+                    }
+                    auto const axis_value = axis.cast<ssize_t>();
+                    if (self.ndim() == 1 && (axis_value == 0 || axis_value == -1))
+                    {
+                        return py::cast(self.argmin());
+                    }
```
Yuli-ytd, line 644: The 1D axis case returns the existing scalar result to match NumPy behavior and avoid relying on 0D `SimpleArray` storage semantics.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
     return ret;
 }
 
+template <typename A, typename T>
+ssize_t detail::SimpleArrayMixinSearch<A, T>::normalize_axis(ssize_t axis, char const * op) const
```
Yuli-ytd, line 2990: Keep axis normalization in the C++ layer so the public C++ overloads share the same axis validation rules, including negative axes.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+{
+    auto athis = static_cast<A const *>(this);
+
+    if (athis->size() == 0)
+    {
+        throw std::invalid_argument("SimpleArray::argmin(): "
+                                    "attempt to get argmin of an empty sequence");
+    }
+
+    auto const range = IndexRange(*athis);
```
Yuli-ytd, line 3093: The flat-index path now walks logical indices and reads through `at(idx)`, so the returned index follows NumPy's logical C-order traversal instead of physical buffer order.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    return max_index;
+}
+
+template <typename A, typename T>
+SimpleArray<uint64_t> detail::SimpleArrayMixinSearch<A, T>::argmin(ssize_t axis) const
+{
+    auto athis = static_cast<A const *>(this);
+
+    axis = normalize_axis(axis, "argmin");
+    SimpleArray<uint64_t> result(reduced_shape(axis));
```
Yuli-ytd, line 3202: Axis reductions reuse the same axis validation and build an output shape with the reduced axis removed, matching NumPy's reduction shape.

tests/test_buffer.py
```diff
@@ -4314,6 +4314,65 @@ def test_argminmax(self):
...
+            sarr = solvcon.SimpleArrayFloat64(array=narr)
+            with self.assertRaises(ValueError):
+                sarr.argmin()
+            with self.assertRaises(ValueError):
+                sarr.argmax()
+
+        with self.subTest(name='non_contiguous'):
+            data = np.zeros((4, 6), dtype='float64')
+            data[0, 0] = -1
+            data[0, 1] = 999
```
Yuli-ytd, line 4456: The out-of-view large and small values make this fail under physical-buffer traversal, so the test specifically guards the stride-aware flat-index behavior.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+
+    auto const out_range = IndexRange(result);
+    sshape_type out_idx = out_range.first();
+
+    do
+    {
+        sshape_type in_idx(athis->ndim(), 0);
+
+        for (ssize_t dim = 0, out_dim = 0; dim < athis->ndim(); ++dim)
```
Yuli-ytd, line 3218: `out_idx` is expanded back into an input logical index before reading with `at(in_idx)`, which keeps strided views correct.

Yuli-ytd:
This PR is ready for review.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1361,42 +1361,21 @@ class SimpleArrayMixinSearch
...
-        {
-            if (athis->data(i) < min_value)
-            {
-                min_value = athis->data(i);
-                min_index = i;
-            }
-        }
-        return min_index;
-    }
+    size_t argmin() const;
```
yungyuc, line 1387: It's good to take this opportunity to move the function body outside the class declaration.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1361,42 +1361,21 @@ class SimpleArrayMixinSearch
 private:
 
     using internal_types = detail::SimpleArrayInternalTypes<T>;
+    using shape_type = typename internal_types::shape_type;
+    using sshape_type = typename internal_types::sshape_type;
+
+    ssize_t normalize_axis(ssize_t axis, char const * op) const;
```
yungyuc, line 1376: @ThreeMonth03 @KHLee529 Could you please help evaluate if it is good to factor out the shape and axis helpers `normalize_axis` and `reduced_shape` to a common mixin to be shared among all mixins?

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    auto const range = IndexRange(*athis);
+    sshape_type idx = range.first();
+
+    value_type min_value = athis->at(idx);
+    size_t min_index = 0;
+    size_t flat_index = 0;
+
+    do
+    {
+        value_type const current_value = athis->at(idx);
```
yungyuc: Avoid bound-checking which is slow.  Use `(*athis)[idx]` and carefully calculate index instead.

tests/test_buffer.py
```diff
@@ -4314,6 +4314,65 @@ def test_argminmax(self):
...
+            data[0, 3] = -999
+            data[2, 4] = 10
+            narr = data[::2, ::2]
+            sarr = solvcon.SimpleArrayFloat64(array=narr)
+            self.assertEqual(sarr.argmin(), 0)
+            self.assertEqual(sarr.argmax(), 5)
+            self.assertEqual(narr.argmin(), sarr.argmin())
```
yungyuc, line 4463: Add constants fixture in the test function so that we do not need to run the code to see the value.

yungyuc (changes requested):
- [ ] Avoid bound-checking which is slow.  Use `(*athis)[idx]` and carefully calculate index instead.  Update all occurrences.
- [ ] Add constants fixture in the test function so that we do not need to run the code to see the value.  Update all occurrences.

yungyuc:
@ThreeMonth03 @KHLee529 please help take a look.

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03: I suggest keeping these helpers local to `SimpleArrayMixinSearch` for now. `normalize_axis` currently also enforces the argmin/argmax non-empty-axis policy, while value reductions have different empty-domain and axis-selection rules. reduced_shape is also single-axis, whereas other reductions may accept multiple axes. Once another reduction consumer has a concrete implementation, we can compare the contracts and extract stateless axis or domain helpers if they actually align.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    {
+        sshape_type in_idx(athis->ndim(), 0);
+
+        for (ssize_t dim = 0, out_dim = 0; dim < athis->ndim(); ++dim)
+        {
+            if (dim == axis)
+            {
+                continue;
+            }
+            in_idx[dim] = out_idx[out_dim++];
```
ThreeMonth03: Axis traversal assumes every logical dimension starts at zero, but axis 0 starts at `-nghost`. Both reducing and retaining axis 0 therefore fail when ghosts are present.
```python3
import numpy as np
import solvcon

narr = np.array([
    [9.0, 1.0, 8.0],
    [7.0, 6.0, 0.0],
    [5.0, 4.0, 3.0],
])
sarr = solvcon.SimpleArrayFloat64(array=narr)
sarr.nghost = 1

print(sarr.argmin(axis=0).ndarray)
print(sarr.argmin(axis=1).ndarray)
```
Current result:
```
IndexError: SimpleArray: dim 0 in [2, 0] >= nbody: 2
Expected:
[2 0 1]
[1 2 2]
```
Please map iteration offsets to the input's logical coordinates, including the axis-0 ghost offset. The mirrored argmax loop needs the same correction and regression tests.

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03: The public C++ axis overload cannot handle a 1D input. Removing its only axis creates a rank-zero shape, but SimpleArray cannot store or index that result.
```cpp
solvcon::SimpleArrayFloat64 array(3);
array[0] = 2.0;
array[1] = 1.0;
array[2] = 3.0;

auto result = array.argmin(0);
```
Current result:
```
SimpleArray::normalize_index(): empty index
```
The Python wrapper bypasses this overload, but the newly added public C++ API remains invalid. Please define a valid C++ scalar/all-axis route or reject this case explicitly before constructing the rank-zero result. argmax(0) has the same problem.

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03: This unconditionally replaces the existing linear contiguous traversal with runtime-rank `IndexRange` plus checked `at()` access.
Reproducer, run on the base commit and this PR with a Release build:
```python3
import statistics
import timeit

import numpy as np
import solvcon

narr = np.arange(1_000_000, dtype=np.float64)
narr[-1] = -1
sarr = solvcon.SimpleArrayFloat64(array=narr)

samples = timeit.repeat(sarr.argmin, repeat=7, number=10)
print(statistics.median(samples) / 10)
```
Measured result:
base: 0.00117 s/call
PR:   0.00378 s/call
The 10k-element case regressed from approximately 8.0 us to 50.2 us. Please preserve direct linear traversal for eligible contiguous arrays and use unchecked signed-stride traversal as the generic fallback.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    auto athis = static_cast<A const *>(this);
+
+    axis = normalize_axis(axis, "argmin");
+    SimpleArray<uint64_t> result(reduced_shape(axis));
+
+    auto const out_range = IndexRange(result);
+    sshape_type out_idx = out_range.first();
+
+    do
+    {
```
ThreeMonth03, line 3215: `normalize_axis()` only checks whether the reduced axis is empty. A retained axis may still have size zero, in which case the result is empty and this do/while must not execute.
```python3
import numpy as np
import solvcon

sarr = solvcon.SimpleArrayFloat64(
    shape=(0, 3),
    value=0.0,
)

print(sarr.argmin(axis=1).ndarray)
```
Current result:
```
IndexError: SimpleArray: dim 0 in [0, 0] >= nbody: 0
```
NumPy returns an empty result:
```
np.argmin(np.empty((0, 3)), axis=1).shape
# (0,)
```
Please return the empty result without entering the output loop. (2, 0, 3) reduced along axis 0 is another reproducer.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2940,6 +2919,204 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+
+        for (ssize_t i = 1; i < athis->shape(axis); ++i)
+        {
+            in_idx[axis] = i;
+            value_type const current_value = athis->at(in_idx);
+            if (current_value < min_value)
+            {
+                min_value = current_value;
+                min_index = static_cast<uint64_t>(i);
+            }
```
ThreeMonth03: The strict comparison ignores a NaN encountered after the first element, while NumPy argmin and argmax select the first NaN position.
```python3
import numpy as np
import solvcon

narr = np.array([[1.0, np.nan, 0.0]])
sarr = solvcon.SimpleArrayFloat64(array=narr)

print(np.argmin(narr, axis=1))
print(sarr.argmin(axis=1).ndarray)
print(np.argmax(narr, axis=1))
print(sarr.argmax(axis=1).ndarray)
```
Current result:
```
NumPy argmin:   [1]
SimpleArray:    [2]
NumPy argmax:   [1]
SimpleArray:    [0]
```
Please either implement the NumPy NaN policy for the new axis overloads or document and test an intentional semantic difference.

ThreeMonth03 (changes requested):
I find some bugs in this pull request. Please take a look.

Yuli-ytd:
Thanks for the detailed review and suggestions. I have received the feedback and will address the requested changes as soon as possible.

I plan to push the updates by 7/26 (Sun).

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Done. Flat reductions use direct pointer traversal for contiguous arrays and unchecked signed-stride offsets for other layouts. Axis reductions use explicitly calculated buffer indices with `operator[]` for both input reads and result writes. No argmin or argmax traversal uses checked `at()` access.

tests/test_buffer.py
Yuli-ytd: Done. I added explicit expected values to the new argmin and argmax tests so the intended results are visible without running NumPy. The NumPy comparisons remain as secondary cross-checks.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Thanks for the discussion. I kept the two helpers local for now.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Done. Iteration offsets are now mapped back to the input logical coordinates through `IndexRange::index_from_offset`. Regression tests cover both reducing and retaining axis 0 for argmin and argmax.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: The axis overloads now return immediately when the reduced result has size zero, before entering the output loop. Regression tests cover shapes `(0, 3)` reduced along axis 1 and `(2, 0, 3)` reduced along axis 0.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Done. Both flat and axis argmin and argmax now select the first encountered NaN, matching NumPy. The tests include consecutive NaNs to verify that the first position is retained.

Yuli-ytd:
@yungyuc Thanks for the review. Both requested changes have been addressed:

- [x] Replaced checked access in all `argmin` and `argmax` traversal paths with direct pointer access or explicitly calculated unchecked buffer indices.
- [x] Added explicit expected constants throughout the new `argmin` and `argmax` tests, while retaining the NumPy comparisons as secondary cross-checks.

The implementation details are included in the corresponding inline replies. Could you please take a look?

Yuli-ytd:
@ThreeMonth03 Thanks for the detailed review. All requested changes have been addressed. Could you please take another look?

The implementation and regression-test details are included in the corresponding inline replies.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: The contiguous path now preserves direct linear traversal, while the generic fallback uses unchecked signed-stride traversal.

I reran the provided benchmark on `master` and the latest PR branch using the same Release environment:

- `master`: 0.000766 s/call
- Latest PR ([bc4908b](https://github.com/solvcon/solvcon/pull/1158/commits/bc4908bbbf500a5972a5c8dd53fa4c87706bb284)): 0.000768 s/call

The difference is approximately 0.3%, indicating that the reported performance regression is no longer present.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3002,6 +2986,346 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+
+    value_type const * ptr = athis->logical_data();
+    value_type min_value = *(ptr + unchecked_logical_offset(*athis, idx));
+    size_t min_index = 0;
+    size_t flat_index = 0;
+
+    do
+    {
+        value_type const current_value = *(
+            ptr + unchecked_logical_offset(*athis, idx));
```
yungyuc: Good change.

If clang-format allows, use one line for similar code:

```suggestion
        value_type const current_value = *(ptr + unchecked_logical_offset(*athis, idx));
```

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3002,6 +2986,346 @@ A detail::SimpleArrayMixinSort<A, T>::take_along_axis(SimpleArray<I> const & ind
...
+    {
+        axis += ndim;
+    }
+    if (axis < 0 || axis >= ndim)
+    {
+        throw std::invalid_argument(std::format(
+            "SimpleArray::{}(): axis {} is out of bounds for array of dimension {}",
+            op,
+            axis,
+            ndim));
```
yungyuc, line 3005: If clang-format allows, use one line for `op, axis, ndim`.  It's not easy to read to put them in multiple lines.  Change other similar places.

```suggestion
            op, axis, ndim));
```

yungyuc (approved):
- [x] Save lines if clang-format allows.

LGTM.  @ThreeMonth03 what do you think?

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: Done. I consolidated this expression and similar occurrences where clang-format allows it.

cpp/solvcon/buffer/SimpleArray.hpp
Yuli-ytd: I tried the suggested one-line argument list, but `make FORCE_CLANG_FORMAT=inplace cformat` restores the multi-line form, so I kept the formatter-generated result.

Yuli-ytd:
@yungyuc Thanks for the follow-up.

- [x] Save lines if clang-format allows.

I pushed the formatting changes that clang-format allows. Details are included in the inline replies.

I will apply this guideline to future changes.

ThreeMonth03 (approved):
LGTM

PR #1165, opened 2026-07-23
Reuse the shared shape formatter in linalg

### LINKED ISSUE #893: Refactor repeated format_shape() helper in linalg into a mix-in
Follow-up from review of #815.

A `format_shape()` helper -- formatting a `SimpleArray` shape as `"(d0, d1, ...)"` for error messages -- is duplicated across several linalg classes (e.g. `LuFactorization` in `cpp/modmesh/linalg/lu_factorization.hpp`, and other LA code).

Scope: extract the shared `format_shape()` logic into a reusable mix-in (or comparable shared base) so the LA classes no longer each carry their own copy.

Raised by @yungyuc in https://github.com/solvcon/modmesh/pull/815.

ThreeMonth03:
Several linalg classes independently format `SimpleArray` shapes for diagnostic messages, and the LLT and Kalman paths repeat the same formatting loop at each validation site.

Reuse the existing shared shape formatter in `Llt`, `LuFactorization`, `KalmanFilter`, and `EigenSystem`. This removes the duplicated helpers and loops without adding another inheritance layer or changing the messages or public API.

Related to #893.

### DIFF
--- a/cpp/solvcon/linalg/EigenSystem.hpp
+++ b/cpp/solvcon/linalg/EigenSystem.hpp
@@ -95,8 +95,6 @@ public:
 
 private:
 
-    static std::string format_shape(array_type const & arr);
-
     array_type const & m_matrix;
     array_type m_colmajor;
     real_array_type m_wr;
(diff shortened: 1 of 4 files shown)


cpp/solvcon/linalg/EigenSystem.hpp
```diff
@@ -125,7 +123,7 @@ EigenSystem<T>::EigenSystem(array_type const & matrix, bool do_vl, bool do_vr, c
     {
         throw std::invalid_argument(std::format(
             "EigenSystem: matrix must be a square 2D SimpleArray, but got shape {}",
-            format_shape(matrix)));
+            detail::format_shape(matrix.shape())));
```
ThreeMonth03, line 126: Replace `format_shape(matrix)` with `detail::format_shape(matrix.shape())`.

ThreeMonth03 (commented):
@yungyuc Please take a look.

yungyuc (approved):
Beautiful cleanup.

PR #1168, opened 2026-07-23
Add SVG export for the pilot 2D canvas

### LINKED ISSUE #1050: Export SVG from 2D canvas
Per https://github.com/solvcon/solvcon/pull/1046#discussion_r3559041506, we shall think how to export SVG from the 2D canvas.

tychuang1211:
The pilot 2D canvas can currently only be exported as a raster PNG or JPEG. This adds a vector SVG export option alongside them, useful for anyone who wants to include a canvas snapshot in a document or diagram without losing resolution when scaled.

R2DWidget gains a saveSvg method that renders through the same paint_canvas path the existing PNG/JPEG export uses, but into a QSvgGenerator instead of a QImage, and is exposed to Python through the existing pybind11 binding pattern. Getting QSvgGenerator to produce a correctly scaled file took one extra step beyond setting the output size: without an explicit viewBox, viewers map the drawing coordinates onto the physical canvas size using the CSS convention of 96 DPI, which does not match the DPI QSvgGenerator itself used internally, so content (including the background fill) rendered shrunk into the canvas's top-left corner with blank space around it. Setting an identity viewBox fixes this by pinning pixel coordinates 1:1 to the canvas regardless of DPI.

On the Python side, the "Save 2D canvas" dialog now offers an SVG filter alongside PNG and JPEG, and dispatches to saveSvg when the resolved path ends in .svg.

Related to https://github.com/solvcon/solvcon/issues/1050.

### DIFF
--- a/cpp/solvcon/CMakeLists.txt
+++ b/cpp/solvcon/CMakeLists.txt
@@ -30,6 +30,8 @@ if (BUILD_QT)
     find_package(Qt6 QUIET OPTIONAL_COMPONENTS GuiPrivate)
     # ShaderTools provides qt6_add_shaders() and the qsb tool.
     find_package(Qt6 REQUIRED COMPONENTS ShaderTools)
+    # Svg provides QSvgGenerator, used by R2DWidget::saveSvg.
+    find_package(Qt6 REQUIRED COMPONENTS Svg)
 endif () # BUILD_QT
 
 add_subdirectory(buffer)
@@ -120,6 +122,7 @@ if (BUILD_QT)
         Qt::Widgets
         Qt::Core
         Qt::Gui
+        Qt::Svg
     )
 
     # <rhi/qrhi.h> lives in the QtGui private headers. Prefer the GuiPrivate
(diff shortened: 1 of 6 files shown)


tychuang1211:
image

yungyuc (changes requested):
LGTM.  This change is smaller than I thought.  Perhaps 1 or 2 commits would be enough.  Please clarify if you really need 5 commits for the PR.

tigercosmos (approved):
`QSvgGenerator` could be a good start. Later I expect to have a `SvgExporter` directly from `World` that generate shapes from the world's shapes, e.g. `<circle>` for the circle shape in the world.

yungyuc (approved):
LGTM

> `QSvgGenerator` could be a good start. Later I expect to have a `SvgExporter` directly from `World` that generate shapes from the world's shapes, e.g. `<circle>` for the circle shape in the world.

Good point.  @tychuang1211 , could you please create a followup issue to track the work to create a SVG writer for `World` without depending on Qt?

PR #1170, opened 2026-07-23
Refuse to build scdv from an activated environment

### LINKED ISSUE #871: Combine the per-OS mmdv build scripts into one cross-platform script

yungyuc:
This PR follows up PR #1127 for issue #871.

The activation script exports `SCDV_BASE`, and the builder honored an inherited value as a deliberate override, so running it from a shell with an active scdv (or old-generation mmdv) installed the new build into the active environment's tree.  Detect the activation through the `_SCDV_OLD_PATH`/`_MMDV_OLD_PATH` snapshot variables, which exist only after an activate file has been sourced, and abort with instructions to deactivate first.  `--allow-write-to-active-env` overrides the fence for a deliberate in-place rebuild, with a warning echoed when it takes effect; a `SCDV_BASE` override from a clean shell keeps working and the print-only modes stay usable.

The from-scratch verification build also surfaced that the documentation toolchain had drifted from `doc/requirements.txt`: install furo, the theme the docs actually use, instead of pydata-sphinx-theme.

Also tightens comments throughout the script. State the why without walking through the mechanics.

### DIFF
--- a/contrib/dependency/build-scdv.sh
+++ b/contrib/dependency/build-scdv.sh
@@ -1,59 +1,47 @@
 #!/bin/bash
 #
-# Build solvcon's runtime dependencies from source into a user-space prefix,
-# with no dependency on the devenv tool. This is a single self-contained
-# cross-platform script: it holds the shared build logic and the
-# platform-specific parts for every supported OS inline, so nothing else in the
-# repository is sourced or called. The target platform is auto-detected from
-# `uname -s` (Linux -> ubuntu, Darwin -> macos) and can be forced with
-# SCDV_OS=ubuntu|macos.
+# Build solvcon's runtime dependencies from source into a user-space prefix.
+# This is a single self-contained cross-platform script: the shared build
+# logic and every per-OS part live inline, so nothing else is sourced or
+# called.  The platform is auto-detected from `uname -s` and can be forced
+# with SCDV_OS=ubuntu|macos.
 #
-# The platform-specific parts live in the per-OS `case` blocks below (search for
-# 'case "${SCDV_OS}"'). Each block sets SCDV_OS_TAG and the plat_* hook
-# functions this shared logic calls (see "Platform block contract" below), and
-# carries its own notes (apt/brew prerequisites, libclang, etc.). The
-# per-package build recipes are inlined translations of the scripts under
-# devenv/scripts/build.d/ (zlib, openssl, sqlite, python, pybind11, cython,
-# numpy, scipy, qt, pyside6).
+# The per-OS parts live in the `case "${SCDV_OS}"` blocks below; each sets
+# SCDV_OS_TAG and the plat_* hooks (see "Platform block contract") and
+# carries its own notes.  The build recipes are inlined translations of the
+# devenv/scripts/build.d/ scripts.
 #
-# 4 sections (BASE, PYTHON, NUMPY, QT) are guarded by the corresponding
-# SCDVBUILD_* environment variables.
-#
-# Before running this script, install the system prerequisites. The script
-# never invokes the OS package manager itself; run it with --print-deps to see
-# the exact commands, review them, and run them yourself.
+# 4 sections (BASE, PYTHON, NUMPY, QT) are guarded by the SCDVBUILD_*
+# environment variables.  Install the system prerequisites first: the script
+# never invokes the package manager; --print-deps shows the commands.
 #
 # Usage:
 #   ./build-scdv.sh
-#       Builds everything (BASE + PYTHON + NUMPY + QT).  The default is "build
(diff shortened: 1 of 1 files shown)


contrib/dependency/build-scdv.sh
```diff
@@ -1,59 +1,47 @@
 #!/bin/bash
 #
-# Build solvcon's runtime dependencies from source into a user-space prefix,
```
yungyuc, line 3: Tighten comments.

contrib/dependency/build-scdv.sh
```diff
@@ -1,59 +1,47 @@
...
-#       qt pyside6.  The flag accepts a single name, a comma-separated list
-#       ("--skip openssl,sqlite"), and may be repeated.  The --skip=PKG form
-#       also works.
+#       Skip a package within the selected sections.  PKG is one of: zlib
+#       openssl sqlite python pybind11 cython numpy scipy qt pyside6.
+#       Accepts a comma-separated list, repetition, and the --skip=PKG form.
 #   ./build-scdv.sh --no-confirm
 #       Skip the "Press Enter to start the build" prompt that fires after the
 #       startup echo block.  Use in non-interactive runs (CI, scripts).
+#   ./build-scdv.sh --allow-write-to-active-env
```
yungyuc, line 35: New command line argument `--allow-write-to-active-env`.  If unset, the script does not do any building and writing when a scdv is activated.

contrib/dependency/build-scdv.sh
```diff
@@ -965,15 +866,32 @@ fi
...
   echo "SCDV_PREFIX=${SCDV_PREFIX}"
   exit 0
 fi
 
+# An activated scdv/mmdv exports SCDV_BASE, which the assignment below would
+# honor and install into the active tree.  The _SCDV_OLD_PATH/_MMDV_OLD_PATH
+# snapshot vars exist only after sourcing an activate, so a deliberate
+# SCDV_BASE override from a clean shell still passes.
+if [ -n "${_SCDV_OLD_PATH+1}${_MMDV_OLD_PATH+1}" ] ; then
+  if [ "${SCDV_ALLOW_WRITE_TO_ACTIVE_ENV}" = "1" ] ; then
```
yungyuc, line 881: If `SCDV_ALLOW_WRITE_TO_ACTIVE_ENV` is 1, show a warning and continue building and writing into the active scdv.

contrib/dependency/build-scdv.sh
```diff
@@ -1467,9 +1375,10 @@ if [[ "${SCDVBUILD_ALL:-}" == "1" || "${SCDVBUILD_PYTHON:-}" == "1" ]] ; then
 echo "Python build uses PGO; expect ~20 min"
 scdv_time build_python python
 "${PY}" -m pip install -U flake8 autopep8 black pytest jsonschema certifi
-# Documentation toolchain (Sphinx-based; see doc/README.md). doxygen, the
-# system half of the C++ API path, is in the base prerequisites above.
-"${PY}" -m pip install -U sphinx myst-parser pydata-sphinx-theme \
+# Documentation toolchain; keep in sync with doc/requirements.txt (the
+# self-contained script does not read it).  doxygen comes from the base
+# prerequisites.
+"${PY}" -m pip install -U sphinx myst-parser furo \
```
yungyuc, line 1381: Correct the document dependency packages.

PR #1179, opened 2026-07-25
Accept zero-sized NumPy array layouts

### LINKED ISSUE #1178: Allow zero-sized NumPy arrays in SimpleArray construction
SimpleArray rejects valid zero-sized NumPy arrays when another axis has length greater than one.

```python
import numpy as np
import solvcon

for shape in ((3, 0), (0, 3), (2, 1, 0)):
    array = np.empty(shape, dtype="float64")
    print(shape, array.flags.c_contiguous, array.flags.f_contiguous)
    solvcon.SimpleArrayFloat64(array=array)
```

Current result:

```text
RuntimeError: SimpleArray: F contiguous stride must match shape and start with 1
```

Depending on the shape, the C-contiguous check may fail instead. NumPy marks zero-sized arrays as contiguous even when their strides do not follow the ordinary non-empty recurrence. The wrapper preserves those flags and then applies the stricter SimpleArray C/F stride checks.

Construction should succeed and preserve the original zero-sized shape. The contiguity checks should treat any shape containing a zero-length axis as contiguous, consistently with the NumPy flags supplied to the wrapper.

ThreeMonth03:
NumPy marks zero-sized arrays as both C- and F-contiguous even when their strides do not follow the recurrence used for non-empty arrays. SimpleArray preserved those flags but then rejected valid shapes such as `(3, 0)` and `(0, 3)`.

This change treats any shape with a zero-length axis as contiguous for order validation and adds constructor regressions for zero-sized shapes in different axis positions.

Related to #1178.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -2133,7 +2133,7 @@ private:
     static bool is_c_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (shape.empty())
+        if (std::ranges::contains(shape, 0))
         {
             return true;
         }
@@ -2152,7 +2152,7 @@ private:
     static bool is_f_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (shape.empty())
+        if (std::ranges::contains(shape, 0))
         {
             return true;
         }
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -290,13 +290,18 @@ class SimpleArrayBasicTC(unittest.TestCase):
         ):
             solvcon.SimpleArrayFloat64((-1, 2))
 
-    def test_SimpleArray_fortran_empty(self):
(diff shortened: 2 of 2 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03, line 2169: Replace `shape.empty()` with `std::ranges::contains(shape, 0)`. We don't need to check whether shape is empty, since it always return true in the last line.

ThreeMonth03 (commented):

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2133,7 +2133,7 @@ class SimpleArray
     static bool is_c_contiguous(shape_type const & shape,
                                 sshape_type const & stride)
     {
-        if (shape.empty())
+        if (std::ranges::contains(shape, 0))
```
yungyuc, line 2136: Learned a trick.

yungyuc (approved):
LGTM

PR #1180, opened 2026-07-25
Check small_vector size for equality test

### LINKED ISSUE #1177: Require equal sizes in small_vector equality
`small_vector::operator==` currently compares the lhs element range with `rhs.begin()` without first checking the two sizes. A shorter lhs that matches the rhs prefix is therefore treated as equal, while a longer lhs may read past the rhs range.

```cpp
small_vector<int> prefix{1, 2};
small_vector<int> longer{1, 2, 3};

prefix == longer; // incorrectly true
longer == prefix; // may read past prefix
```

Equality should require equal sizes before comparing corresponding elements.

`SimpleArray` shape validation exposes this bug because operand shapes with different ranks may be treated as equal. Regression coverage should therefore include unequal `small_vector` sizes in both directions and a `SimpleArray` shape-validation case.

ThreeMonth03:
`small_vector::operator==` compares the lhs element range with `rhs.begin()` without first checking the two sizes. A shorter lhs that matches the rhs prefix may therefore compare equal, while a longer lhs may read past the rhs range.

This change requires equal sizes before comparing corresponding elements. Regression coverage includes unequal `small_vector` sizes in both directions and the `SimpleArray` shape-validation case that exposed the bug.

Related to #1177.

### DIFF
--- a/cpp/solvcon/buffer/small_vector.hpp
+++ b/cpp/solvcon/buffer/small_vector.hpp
@@ -446,7 +446,8 @@ small_vector<T, N>::quick_select(iterator first, iterator last, size_t k)
 template <typename T>
 bool operator==(small_vector<T> const & lhs, small_vector<T> const & rhs)
 {
-    return std::equal(lhs.begin(), lhs.end(), rhs.begin());
+    return lhs.size() == rhs.size() &&
+           std::equal(lhs.begin(), lhs.end(), rhs.begin());
 }
 
 static_assert(sizeof(small_vector<size_t>) == 40, "small_vector<size_t> should use 40 bytes");
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -252,6 +252,17 @@ TEST(small_vector, select_kth_random)
     }
 }
 
+TEST(small_vector, equality_requires_same_size)
+{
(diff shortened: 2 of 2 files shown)


cpp/solvcon/buffer/small_vector.hpp
```diff
@@ -446,7 +446,8 @@ small_vector<T, N>::quick_select(iterator first, iterator last, size_t k)
 template <typename T>
 bool operator==(small_vector<T> const & lhs, small_vector<T> const & rhs)
 {
-    return std::equal(lhs.begin(), lhs.end(), rhs.begin());
+    return lhs.size() == rhs.size() &&
+           std::equal(lhs.begin(), lhs.end(), rhs.begin());
```
ThreeMonth03, line 450: Ensure the sizes of `small_vector` are same.

ThreeMonth03 (commented):
@yungyuc Please take a look.

yungyuc (approved):
LGTM.

PR should use a title slightly different from the issue to make it easier to distinguish.

cpp/solvcon/buffer/small_vector.hpp
yungyuc: Nice catch.

Issue #1172, opened 2026-07-24
Add NumPy-compatible batched matmul with layout-aware HPC execution


ThreeMonth03:
## Motivation

`SimpleArray::matmul()` accepts rank-1 and rank-2 operands but rejects leading batch dimensions. This prevents NumPy-style batched matrix multiplication:

```python
lhs.shape == (2, 1, 3, 4)
rhs.shape == (1, 5, 4, 6)

result = lhs @ rhs
assert result.shape == (2, 5, 3, 6)
```

Only the leading batch axes should broadcast. The trailing axes retain their vector and matrix roles, and the contracted `K` dimensions must match. Existing rank-1 and rank-2 result conventions remain unchanged.

The implementation must also be layout-aware. A loop over broadcast coordinates is correct but can repeatedly copy the same non-contiguous matrix before calling BLAS. On Apple M1, NumPy and the prototype both use Accelerate. For `(1,256,256) @ (64,256,256)`, the dense route is at parity, while a broadcast negative-stride lhs exposes the cost of repeated packing:

| Lhs supplied to the call | Lhs element strides | NumPy | Prototype |
| --- | --- | ---: | ---: |
| Dense | `(65536,256,1)` | 11.960 ms | 11.873 ms |
| Reversed columns | `(65536,256,-1)` | 1997.845 ms | 12.099 ms |
| Same reversed values, prepacked before timing | `(65536,256,1)` | 11.967 ms | 11.950 ms |

The prepacked control shows that broadcasting itself is not the slow operation. The prototype packs the one supplied lhs inside the timed call and reuses it for 64 GEMMs. Full platform results and controls are kept in the linked fork study.

The prototype also found an existing signed-stride correctness defect that must be fixed before optimized routes are reused:

```python
values = np.arange(8.0)[::-1]
expected = np.matmul(values, values)  # 140.0

array = solvcon.SimpleArrayFloat64(array=values)
actual = array.matmul(array)           # 56.0 in the current path
```

## Proposed design

Callers continue to use `lhs.matmul(rhs)` or `lhs @ rhs`. Planning and execution remain internal:

```text
SimpleArray::matmul(rhs)
        |
        v
MatmulPlan::make(lhs, rhs)
        |
        +-- validate vector roles and contracted K
        +-- compute broadcast batch and output shapes
        `-- retain signed matrix and batch strides
        |
        v
MatmulExecutor::execute(plan, lhs, rhs)
        +-- generic signed-stride contraction
[...]

PR #1181, opened 2026-07-25
Traverse equal matmul batches

ThreeMonth03:
This PR adds batched matrix multiplication to `matmul_planned()` when both matrix operands have the same leading batch shape. For example:

```text
(4, 1000, 2000) @ (4, 2000, 3000) -> (4, 1000, 3000)
```

Each batch coordinate is evaluated with the existing generic signed-stride contraction. This PR intentionally does not add NumPy-style batch broadcasting, rank-1 DOT/GEVM/GEMV roles, or BLAS dispatch. Broadcasting and rank-1 roles are handled by the stacked follow-up #1208, while optimized dispatch remains later work in #1172.

To traverse the batch coordinates without constructing temporary `SimpleArray` views, `LoopDomain` records the runtime-rank batch shape, `OperandMapping` records the signed output, lhs, and rhs batch strides, and `MappedOffsetCursor` advances their offsets together. `MatmulPlan` retains these mappings, and `MatmulExecutor` writes every matrix product into one allocated output.

Python differential tests cover rank-3 and rank-4 equal batches with C-contiguous, negative-stride, and step-two batch layouts, plus empty batch domains. The profiler documents the end-to-end performance of this generic route, not the final BLAS-optimized implementation.

## Profiling

The macOS measurements used an Apple M1 with NumPy 2.5.1 and `_solvcon` both linked to Accelerate. The benchmark used Python 3.14.6, fixed `B = 10`, two warmups, and 15 measured calls per case. The generic planned route is already competitive for small batched matrices. Larger C-contiguous matrices will use size- and layout-aware dispatch to optimized BLAS or tiled routes in later tasks.

### Non-contiguous behavior

`np.flip()` produces batched views with lhs strides `(a*a, a, -1)` and rhs strides `(a*a, -a, 1)`. In NumPy 2.5.1 these layouts still fail the direct [`is_blasable2d()` check](https://github.com/numpy/numpy/blob/v2.5.1/numpy/_core/src/umath/matmul.c.src#L39-L61), but matrix-matrix dispatch now [allocates reusable temporary buffers, packs non-BLASable operands, and invokes the BLAS matrix-matrix kernel](https://github.com/numpy/numpy/blob/v2.5.1/numpy/_core/src/umath/matmul.c.src#L439-L585). The planned executor continues to contract the supplied signed strides directly. Consequently, NumPy 2.5.1 is substantially faster than the earlier NumPy 2.2.4 non-BLAS path for the larger non-contiguous cases; the larger NumPy-normalized ratios do not indicate an equivalent absolute-time regression in `planned_sa`.

### float32 scaling by input layout (NumPy 2.5.1)

Float32 batched matrix multiplication scaling at fixed B=10 with NumPy 2.5.1

### float64 scaling by input layout (NumPy 2.5.1)

Float64 batched matrix multiplication scaling at fixed B=10 with NumPy 2.5.1

<details>
<summary>Complete raw profile_matrix_ops Markdown (NumPy 2.5.1)</summary>

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 4, 4)`, element strides: `(16, 4, 1)`
- rhs shape: `(10, 4, 4)`, element strides: `(16, 4, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 2.355E-03       | 1.000           |
| planned_sa           | 2.367E-03       | 1.005           |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 4, 4)`, element strides: `(16, 4, -1)`
- rhs shape: `(10, 4, 4)`, element strides: `(16, -4, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 2.389E-03       | 1.000           |
| planned_sa           | 1.897E-03       | 0.794           |

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 9, 9)`, element strides: `(81, 9, 1)`
- rhs shape: `(10, 9, 9)`, element strides: `(81, 9, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 4.161E-03       | 1.000           |
| planned_sa           | 6.036E-03       | 1.451           |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 9, 9)`, element strides: `(81, 9, -1)`
- rhs shape: `(10, 9, 9)`, element strides: `(81, -9, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 5.372E-03       | 1.000           |
| planned_sa           | 6.031E-03       | 1.123           |

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 16, 16)`, element strides: `(256, 16, 1)`
- rhs shape: `(10, 16, 16)`, element strides: `(256, 16, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 5.022E-03       | 1.000           |
| planned_sa           | 2.742E-02       | 5.461           |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 16, 16)`, element strides: `(256, 16, -1)`
- rhs shape: `(10, 16, 16)`, element strides: `(256, -16, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 9.383E-03       | 1.000           |
| planned_sa           | 2.744E-02       | 2.925           |

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 27, 27)`, element strides: `(729, 27, 1)`
- rhs shape: `(10, 27, 27)`, element strides: `(729, 27, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 7.978E-03       | 1.000           |
| planned_sa           | 1.207E-01       | 15.126          |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 27, 27)`, element strides: `(729, 27, -1)`
- rhs shape: `(10, 27, 27)`, element strides: `(729, -27, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 1.842E-02       | 1.000           |
| planned_sa           | 1.183E-01       | 6.423           |

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 64, 64)`, element strides: `(4096, 64, 1)`
- rhs shape: `(10, 64, 64)`, element strides: `(4096, 64, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 1.452E-02       | 1.000           |
| planned_sa           | 2.340E+00       | 161.165         |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 64, 64)`, element strides: `(4096, 64, -1)`
- rhs shape: `(10, 64, 64)`, element strides: `(4096, -64, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 6.121E-02       | 1.000           |
| planned_sa           | 2.344E+00       | 38.300          |

## Equal batch: `c_contiguous`, dtype: `float32`
- lhs shape: `(10, 81, 81)`, element strides: `(6561, 81, 1)`
- rhs shape: `(10, 81, 81)`, element strides: `(6561, 81, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 4.301E-02       | 1.000           |
| planned_sa           | 5.641E+00       | 131.153         |

## Equal batch: `non_contiguous`, dtype: `float32`
- lhs shape: `(10, 81, 81)`, element strides: `(6561, 81, -1)`
- rhs shape: `(10, 81, 81)`, element strides: `(6561, -81, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 1.492E-01       | 1.000           |
| planned_sa           | 5.714E+00       | 38.310          |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 4, 4)`, element strides: `(16, 4, 1)`
- rhs shape: `(10, 4, 4)`, element strides: `(16, 4, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 2.294E-03       | 1.000           |
| planned_sa           | 1.856E-03       | 0.809           |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 4, 4)`, element strides: `(16, 4, -1)`
- rhs shape: `(10, 4, 4)`, element strides: `(16, -4, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 2.492E-03       | 1.000           |
| planned_sa           | 1.808E-03       | 0.726           |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 9, 9)`, element strides: `(81, 9, 1)`
- rhs shape: `(10, 9, 9)`, element strides: `(81, 9, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 4.139E-03       | 1.000           |
| planned_sa           | 6.733E-03       | 1.627           |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 9, 9)`, element strides: `(81, 9, -1)`
- rhs shape: `(10, 9, 9)`, element strides: `(81, -9, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 5.222E-03       | 1.000           |
| planned_sa           | 6.800E-03       | 1.302           |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 16, 16)`, element strides: `(256, 16, 1)`
- rhs shape: `(10, 16, 16)`, element strides: `(256, 16, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 4.872E-03       | 1.000           |
| planned_sa           | 2.757E-02       | 5.658           |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 16, 16)`, element strides: `(256, 16, -1)`
- rhs shape: `(10, 16, 16)`, element strides: `(256, -16, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 9.619E-03       | 1.000           |
| planned_sa           | 2.747E-02       | 2.856           |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 27, 27)`, element strides: `(729, 27, 1)`
- rhs shape: `(10, 27, 27)`, element strides: `(729, 27, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 1.470E-02       | 1.000           |
| planned_sa           | 1.255E-01       | 8.540           |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 27, 27)`, element strides: `(729, 27, -1)`
- rhs shape: `(10, 27, 27)`, element strides: `(729, -27, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 2.473E-02       | 1.000           |
| planned_sa           | 1.205E-01       | 4.873           |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 64, 64)`, element strides: `(4096, 64, 1)`
- rhs shape: `(10, 64, 64)`, element strides: `(4096, 64, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 3.682E-02       | 1.000           |
| planned_sa           | 2.377E+00       | 64.543          |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 64, 64)`, element strides: `(4096, 64, -1)`
- rhs shape: `(10, 64, 64)`, element strides: `(4096, -64, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 8.156E-02       | 1.000           |
| planned_sa           | 2.368E+00       | 29.029          |

## Equal batch: `c_contiguous`, dtype: `float64`
- lhs shape: `(10, 81, 81)`, element strides: `(6561, 81, 1)`
- rhs shape: `(10, 81, 81)`, element strides: `(6561, 81, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 9.307E-02       | 1.000           |
| planned_sa           | 5.494E+00       | 59.030          |

## Equal batch: `non_contiguous`, dtype: `float64`
- lhs shape: `(10, 81, 81)`, element strides: `(6561, 81, -1)`
- rhs shape: `(10, 81, 81)`, element strides: `(6561, -81, 1)`

| func                 | per call (ms)   | cmp to np       |
| -------------------- | --------------- | --------------- |
| np                   | 1.950E-01       | 1.000           |
| planned_sa           | 5.512E+00       | 28.274          |

</details>

Matrix correctness passes all 46 tests in `tests/test_matrix.py`. Full clang-format, flake8, include-ordering, ASCII, and trailing-whitespace lint also passes.

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/loop.hpp
+++ b/cpp/solvcon/buffer/loop.hpp
@@ -1,0 +1,195 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/buffer/small_vector.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <functional>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+
+namespace solvcon
+{
+
+namespace detail
+{
+
+/**
+ * @brief Describe a shared runtime-rank iteration space.
+ *
+ * A higher-level operation plan creates one LoopDomain for every set of
+ * coordinates traversed together. The domain owns only the extents and
+ * defines their rank, bounds, and empty-domain behavior. Operand-specific
+ * strides remain in OperandMapping, while operation-specific axes such as
+ * M, N, K, or kept and reduced axes remain in the higher-level plan.
+ *
+ * For `(2,1,3,4) @ (1,5,4,6)`, a broadcast-capable MatmulPlan uses
+ * `LoopDomain({2,5})` for the ten result-batch coordinates. M=3, N=6, and
+ * K=4 remain matrix metadata rather than becoming axes of this domain.
+ *
+ * @note The current consumer is MatmulPlan with equal leading batch shapes.
+ * Batch broadcasting and plans for other operation families are follow-up
+ * work.
+ */
+class LoopDomain
+{
+public:
+    using shape_type = small_vector<ssize_t>;
+
+    explicit LoopDomain(shape_type shape)
+        : m_shape(std::move(shape))
+    {
+    }
+
+    shape_type const & shape() const noexcept { return m_shape; }
+    ssize_t extent(size_t axis) const noexcept { return m_shape[axis]; }
+    size_t rank() const noexcept { return m_shape.size(); }
+    size_t size() const noexcept { return std::ranges::fold_left(m_shape, size_t{1}, std::multiplies<size_t>{}); }
+
+private:
+    shape_type m_shape;
+}; /* end class LoopDomain */
+
+/**
+ * @brief Describe how one operand advances over a LoopDomain.
+ *
+ * Each OperandMapping records one signed stride for every domain axis. This
+ * separates the shared coordinate space from each operand layout: positive
+ * and negative strides traverse supplied layouts, while a zero stride reuses
+ * one operand position along a broadcast axis. contiguous_blocks() constructs
+ * the mapping for adjacent fixed-size blocks.
+ *
+ * For `(2,1,3,4) @ (1,5,4,6)`, the domain is `(2,5)`. The output, lhs, and
+ * rhs mappings are `{90,18}`, `{12,0}`, and `{0,24}`. The zero lhs stride
+ * reuses it across axis 1, and the zero rhs stride reuses it across axis 0.
+ *
+ * @note The current MatmulPlan constructs mappings for equal batch shapes,
+ * so it does not create zero broadcast strides yet.
+ */
+class OperandMapping
+{
+public:
+    using stride_type = small_vector<ssize_t>;
+
+    OperandMapping() = default;
+    explicit OperandMapping(stride_type strides)
+        : m_strides(std::move(strides))
+    {
+    }
+
+    static OperandMapping contiguous_blocks(LoopDomain const & domain, ssize_t block_size);
+
+    size_t rank() const noexcept { return m_strides.size(); }
+    ssize_t stride(size_t axis) const noexcept { return m_strides[axis]; }
+
+private:
+    stride_type m_strides;
+}; /* end class OperandMapping */
+
+/**
+ * @brief Advance all operand offsets with one runtime-rank cursor.
+ *
+ * MappedOffsetCursor owns one coordinate counter for the LoopDomain and one
+ * relative offset for each OperandMapping. advance() updates every offset
+ * together, so executors do not need rank-specific nested loops or separate
+ * coordinate traversal for each operand. The cursor does not dereference
+ * data or execute an operation.
+ *
+ * With domain `(2,5)` and mappings `{90,18}`, `{12,0}`, and `{0,24}`, the
+ * output, lhs, and rhs offsets start at `(0,0,0)`. They advance to
+ * `(18,0,24)` for coordinate `(0,1)` and `(90,12,0)` for coordinate `(1,0)`.
+ *
+ * @note The current consumer is MatmulExecutor with three equal-batch
+ * mappings. The cursor borrows its domain and mappings, which must outlive
+ * it. Broadcast and other operation executors are follow-up work.
+ */
+class MappedOffsetCursor
+{
+public:
+    using mapping_type = small_vector<OperandMapping>;
+
+    MappedOffsetCursor(LoopDomain const & domain, mapping_type const & mappings);
+
+    explicit operator bool() const noexcept { return m_valid; }
+    ssize_t offset(size_t operand) const noexcept { return m_offsets[operand]; }
+
+    template <typename Operand>
+    ssize_t offset(Operand operand) const noexcept
+    {
+        static_assert(std::is_enum_v<Operand>, "cursor operand must be an enum");
+        return offset(static_cast<size_t>(std::to_underlying(operand)));
+    }
+
+    void advance();
+
+private:
+    LoopDomain const & m_domain;
+    mapping_type const & m_mappings;
+    LoopDomain::shape_type m_index;
+    small_vector<ssize_t> m_offsets;
+    bool m_valid = false;
+}; /* end class MappedOffsetCursor */
+
+inline OperandMapping OperandMapping::contiguous_blocks(LoopDomain const & domain, ssize_t block_size)
+{
+    stride_type strides(domain.rank(), block_size);
+    for (size_t axis = domain.rank(); axis > 1; --axis)
+    {
+        strides[axis - 2] = strides[axis - 1] * domain.extent(axis - 1);
+    }
+    return OperandMapping(std::move(strides));
+}
+
+inline MappedOffsetCursor::MappedOffsetCursor(LoopDomain const & domain, mapping_type const & mappings)
+    : m_domain(domain)
+    , m_mappings(mappings)
+    , m_index(domain.rank(), 0)
+    , m_offsets(mappings.size(), 0)
+    , m_valid(domain.size() != 0)
+{
+    for (OperandMapping const & mapping : mappings)
+    {
+        if (mapping.rank() != domain.rank())
+        {
+            throw std::invalid_argument(
+                "operand mapping rank does not match its loop domain");
+        }
+    }
+}
+
+inline void MappedOffsetCursor::advance()
+{
+    for (size_t axis_plus_one = m_domain.rank(); axis_plus_one > 0; --axis_plus_one)
+    {
+        size_t const axis = axis_plus_one - 1;
+        ssize_t const extent = m_domain.extent(axis);
+        ++m_index[axis];
+        if (m_index[axis] < extent)
+        {
+            for (size_t operand = 0; operand < m_mappings.size(); ++operand)
+            {
+                m_offsets[operand] += m_mappings[operand].stride(axis);
+            }
+            return;
+        }
+
+        m_index[axis] = 0;
+        for (size_t operand = 0; operand < m_mappings.size(); ++operand)
+        {
+            m_offsets[operand] -= m_mappings[operand].stride(axis) * (extent - 1);
+        }
+    }
+    m_valid = false;
+}
+
+} /* end namespace detail */
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 nobomb et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -5,12 +5,15 @@
  * BSD 3-Clause License, see COPYING
  */
 
+#include <solvcon/buffer/loop.hpp>
 #include <solvcon/buffer/small_vector.hpp>
 #include <solvcon/math/math.hpp>
 
 #include <algorithm>
 #include <cstddef>
+#include <cstdint>
 #include <format>
+#include <ranges>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
@@ -43,14 +46,18 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
  * the broadcast axes. It does not allocate the output or evaluate the ten
  * matrix pairs.
  *
- * @note This implementation only plans rank-2 matrix-matrix operands. It
- * validates K and records M, N, K, the output shape, and signed matrix
- * strides. It does not have vector roles or a batch domain yet.
+ * @note This implementation plans matrix-matrix operands with equal ranks
+ * and equal leading batch shapes. It records the batch domain and mappings,
+ * M, N, K, the output shape, and signed matrix strides. Vector roles and
+ * broadcast batch axes are follow-up work.
  */
 class MatmulPlan
 {
 public:
     using shape_type = small_vector<ssize_t>;
+
+    template <typename Array>
+    static MatmulPlan make(Array const & lhs, Array const & rhs);
 
     shape_type const & output_shape() const noexcept { return m_output_shape; }
     ssize_t rows() const noexcept { return m_contraction.m_rows; }
@@ -61,10 +68,14 @@ public:
     ssize_t rhs_inner_stride() const noexcept { return m_strides.m_rhs_inner_stride; }
     ssize_t rhs_column_stride() const noexcept { return m_strides.m_rhs_column_stride; }
 
-    template <typename Array>
-    static MatmulPlan make(Array const & lhs, Array const & rhs);
+    bool has_batch_axes() const noexcept { return m_batch.m_domain.rank() != 0; }
+    MappedOffsetCursor batch_cursor() const & { return MappedOffsetCursor(m_batch.m_domain, m_batch.m_mappings); }
+    MappedOffsetCursor batch_cursor() const && = delete;
 
 private:
+    using batch_stride_type = OperandMapping::stride_type;
+    using mapping_type = MappedOffsetCursor::mapping_type;
+
     struct Contraction
     {
         ssize_t m_rows;
@@ -80,11 +91,29 @@ private:
         ssize_t m_rhs_column_stride;
     }; /* end struct MatrixStrides */
 
-    MatmulPlan(shape_type output_shape, Contraction contraction, MatrixStrides strides);
+    struct BatchMappings
+    {
+        LoopDomain m_domain;
+        mapping_type m_mappings;
+    }; /* end struct BatchMappings */
+
+    MatmulPlan(shape_type output_shape, Contraction contraction, MatrixStrides strides, BatchMappings batch);
+
+    template <typename Array>
+    static Contraction make_contraction(Array const & lhs, Array const & rhs);
+
+    template <typename Array>
+    static BatchMappings make_batch_mappings(Array const & lhs, Array const & rhs, ssize_t output_matrix_size);
+
+    static shape_type make_output_shape(BatchMappings const & batch, Contraction const & contraction);
+
+    template <typename Array>
+    static std::string shape_string(Array const & array);
 
     shape_type m_output_shape;
     Contraction m_contraction;
     MatrixStrides m_strides;
+    BatchMappings m_batch;
 }; /* end class MatmulPlan */
 
 /**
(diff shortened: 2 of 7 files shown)


cpp/solvcon/buffer/loop.hpp
ThreeMonth03, line 1: The class in this files could be reused in other operations, like elementwise operation or statistic operation. Therefore, I moved them out of `matmul.hpp`.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -114,54 +144,147 @@ class MatmulExecutor
...
+    };
+    BatchMappings batch = make_batch_mappings(lhs, rhs, contraction.m_rows * contraction.m_columns);
+    shape_type output_shape = make_output_shape(batch, contraction);
+    return MatmulPlan{
+        std::move(output_shape),
+        contraction,
+        strides,
+        std::move(batch),
+    };
+}
```
ThreeMonth03, line 194: `MatmulPlan::make(Array const & lhs, Array const & rhs)` rearrange the shapes and strides information.

`make_contraction(lhs, rhs);` validates and records unbatched matrix shape `m_rows`, `m_columns` and `m_inner_size`.

`make_batch_mappings(lhs, rhs, contraction.m_rows * contraction.m_columns)` validates and records `class LoopDomain;` and `small_vector<OperandMapping>;`. `class LoopDomain;` represents the result batch coordinates, and `small_vector<OperandMapping>;` represents the strides of batch coordinates of output, LHS and RHS.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -175,19 +298,41 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan const & plan, Array & output, A
...
+    }
+
+    for (MappedOffsetCursor cursor = m_plan.batch_cursor(); cursor; cursor.advance())
+    {
+        execute_generic(
+            cursor.offset(MappingSlot::Output),
+            cursor.offset(MappingSlot::Lhs),
+            cursor.offset(MappingSlot::Rhs));
+    }
+}
```
ThreeMonth03, line 315: If `Matmulplan::has_batch_axes()` is false, it means that LHS and RHS are 2 dimensional array, and the program calculates matrix multiplication without considering batch offset.

Otherwise, the executor iterates and calculates batch offset of output, LHS and RHS by `class MappedOffsetCursor`.

cpp/solvcon/buffer/loop.hpp
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * @note The current consumer is MatmulExecutor with three equal-batch
+ * mappings. The cursor borrows its domain and mappings, which must outlive
+ * it. Broadcast and other operation executors are follow-up work.
+ */
+class MappedOffsetCursor
+{
+public:
+    using mapping_type = small_vector<OperandMapping>;
+
```
ThreeMonth03, line 116: Although `mapping_type` is nested `small_vector` with the shape `small_vector<small_vector<ssize_t>>(3, small_vector<ssize_t>(batch_corrdinates_axis_size))`, there is no evidence proves that it is a bottleneck. I think we don't need to optimize the datastructure now.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -114,54 +144,147 @@ class MatmulExecutor
...
+    {
+        if (axis > 0)
+        {
+            result += ",";
+        }
+        result += std::to_string(array.shape(axis));
+    }
+    result += ")";
+    return result;
 }
```
ThreeMonth03, line 288: See #1165, there is a same static function in `SimpleArray.hpp`. However, we cannot use that function directly, since it would cause nested include. Maybe the follow-up issue is to make the static function as a member function of `SimpleArray` in `SimpleArray.hpp`.

ThreeMonth03 (commented):
@yungyuc Please review this pr. Thanks.
I think the latency of batches traversal is acceptable now.
You can see the following graph, which is profiled and compared with numpy 2.2.3, and recorded in the eighth-edition PR description.

When calculating non contiguous array with `matmul()`, numpy 2.2.3 calculates the result without packing and SIMD, so our program is slightly better than numpy. 
Image

yungyuc:
> @yungyuc Please review this pr. Thanks. I think the latency of batches traversal is acceptable now. You can see the following graph, which is profiled and compared with numpy 2.2.3, and recorded in the eighth-edition PR description.
> 
> When calculating non contiguous array with `matmul()`, numpy 2.2.3 calculates the result without packing and SIMD, so our program is slightly better than numpy.

But numpy 2.5.1 has sped up from 2.2.3.  Why do we compare with the old and slower version?

yungyuc:
> `matmul_planned()` currently handles one rank-2 matrix pair. This change extends it to equal leading batch shapes such as `(10,M,K) @ (10,K,N) -> (10,M,N)` while retaining the existing generic signed-stride contraction.
> 
> `LoopDomain` records the runtime-rank batch shape. `OperandMapping` records signed output, lhs, and rhs batch strides, and `MappedOffsetCursor` advances their offsets together. `MatmulPlan` retains these mappings, the caller allocates one output, and `MatmulExecutor` writes each matrix product without constructing temporary `SimpleArray` views. NumPy-style batch broadcasting remains follow-up work.
> 
> Python differential tests cover rank-3 and rank-4 equal batches with C-contiguous, negative-stride, and step-two batch layouts, plus empty batch domains. The profiler fixes `B = 10` and sweeps `a = 4, 9, 16, 27, 64, 81` for `(10,a,a) @ (10,a,a)`. It compares C-contiguous inputs with non-contiguous views that reverse lhs columns and rhs rows.

It is not clear what do you want to achieve with the PR.  Please respond to this comment to discuss.

ThreeMonth03:
> > @yungyuc Please review this pr. Thanks. I think the latency of batches traversal is acceptable now. You can see the following graph, which is profiled and compared with numpy 2.2.3, and recorded in the eighth-edition PR description.
> > When calculating non contiguous array with `matmul()`, numpy 2.2.3 calculates the result without packing and SIMD, so our program is slightly better than numpy.
> 
> But numpy 2.5.1 has sped up from 2.2.3. Why do we compare with the old and slower version?

Old version calculates non-contiguous array without packing and cblas, and new version accelerates generic matrix multiplication with packing and cblas, which will be implemented in the following pull request. I don't implement and compare packing and cblas algorithm now, otherwise the difference of this pull request will be too large to review.

Besides, I've checked that the prototype outperforms numpy 2.5.1 if we accelerate operation with packing and cblas. It will be implemented in the later PRs.

ThreeMonth03:
> > `matmul_planned()` currently handles one rank-2 matrix pair. This change extends it to equal leading batch shapes such as `(10,M,K) @ (10,K,N) -> (10,M,N)` while retaining the existing generic signed-stride contraction.
> > `LoopDomain` records the runtime-rank batch shape. `OperandMapping` records signed output, lhs, and rhs batch strides, and `MappedOffsetCursor` advances their offsets together. `MatmulPlan` retains these mappings, the caller allocates one output, and `MatmulExecutor` writes each matrix product without constructing temporary `SimpleArray` views. NumPy-style batch broadcasting remains follow-up work.
> > Python differential tests cover rank-3 and rank-4 equal batches with C-contiguous, negative-stride, and step-two batch layouts, plus empty batch domains. The profiler fixes `B = 10` and sweeps `a = 4, 9, 16, 27, 64, 81` for `(10,a,a) @ (10,a,a)`. It compares C-contiguous inputs with non-contiguous views that reverse lhs columns and rhs rows.
> 
> It is not clear what do you want to achieve with the PR. Please respond to this comment to discuss.

This PR is to support naive batch matrix multiplication without considering broadcasting and cblas, like `SimpleArray(shape={4, 1000, 2000}) @ SimpleArray(shape={4, 2000, 3000}) =  SimpleArray(shape={4, 1000, 3000})`. Originally, I want to implement naive batch matrix multiplication with broadcasting, but I notice there would be around 600 lines of difference if we implement batch matrix multiplication with broadcasting. 

Therefore, I decide to divide broadcasting program to follow-up PR #1208, and implement dot/GEVM/ GEMV in #1208. This PR implements part of helper classes about naive broadcasting program.

yungyuc:
> > > `matmul_planned()` currently handles one rank-2 matrix pair. This change extends it to equal leading batch shapes such as `(10,M,K) @ (10,K,N) -> (10,M,N)` while retaining the existing generic signed-stride contraction.
> > > `LoopDomain` records the runtime-rank batch shape. `OperandMapping` records signed output, lhs, and rhs batch strides, and `MappedOffsetCursor` advances their offsets together. `MatmulPlan` retains these mappings, the caller allocates one output, and `MatmulExecutor` writes each matrix product without constructing temporary `SimpleArray` views. NumPy-style batch broadcasting remains follow-up work.
> > > Python differential tests cover rank-3 and rank-4 equal batches with C-contiguous, negative-stride, and step-two batch layouts, plus empty batch domains. The profiler fixes `B = 10` and sweeps `a = 4, 9, 16, 27, 64, 81` for `(10,a,a) @ (10,a,a)`. It compares C-contiguous inputs with non-contiguous views that reverse lhs columns and rhs rows.
> > 
> > 
> > It is not clear what do you want to achieve with the PR. Please respond to this comment to discuss.
> 
> This PR is to support naive batch matrix multiplication without considering broadcasting and cblas, like `SimpleArray(shape={4, 1000, 2000}) @ SimpleArray(shape={4, 2000, 3000}) = SimpleArray(shape={4, 1000, 3000})`. Originally, I want to implement naive batch matrix multiplication with broadcasting, but I notice there would be around 600 lines of difference if we implement batch matrix multiplication with broadcasting.
> 
> Therefore, I decide to divide broadcasting program to follow-up PR #1208, and implement dot/GEVM/ GEMV in #1208. This PR implements part of helper classes about naive broadcasting program.

This clarifies what you want to achieve.  Please update the PR description to make it obvious there.

cpp/solvcon/buffer/loop.hpp
yungyuc: Agree.  Let profiling speak.

cpp/solvcon/buffer/loop.hpp
```diff
@@ -0,0 +1,195 @@
...
+private:
+    LoopDomain const & m_domain;
+    mapping_type const & m_mappings;
+    LoopDomain::shape_type m_index;
+    small_vector<ssize_t> m_offsets;
+    bool m_valid = false;
+}; /* end class MappedOffsetCursor */
+
+inline OperandMapping OperandMapping::contiguous_blocks(LoopDomain const & domain, ssize_t block_size)
```
yungyuc, line 139: It does seem to be necessary to inline these member functions, while I am OK to keep them in the header file.  If you want to keep them here, leave a comment to let me know.

yungyuc (approved):
LGTM.  Please let me know what you think:

- [x] Evaluate if some functions should be in header or cpp file.

cpp/solvcon/buffer/loop.hpp
ThreeMonth03: Yes, some hot spots should be keep in `loop.hpp`, like `MappedOffsetCursor::advance()`, and other functions could be moved to cpp files. However, there are only 200 lines in `loop.hpp` , so I decide not to divide function definitions to cpp files now.

ThreeMonth03 (commented):
- [x] Evaluate if some functions should be in header or cpp file.

@yungyuc Please take a look.

yungyuc (approved):
LGTM

Issue #353, opened 2024-06-07
Implement an initial UI configuration file in the home directory


yungyuc:
Currently the viewer does not remember any UI configuration and it is inconvenient for production use.  An initial UI configuration file placed in the user home directory should be read upon starting to configure the UI.

PR #1182, opened 2026-07-25
Keep the pilot window size and location in a user configuration file

### LINKED ISSUE #1044: Remember window size and location in a user configuration file
After the unified toggle system is done with issue #1022, we can define the user configuration file.  The first feature is to remember the window size and location.

We are not ready for a global configuration file.  First try to design the user-level configuration file, and then extend to a global.  A set of hierarchical configuration files (global -> user) requires an overriding mechanism that is outside the scope of this issue.

Follows up issue #1022.

yungyuc:
For issue #1044.

This helps the pilot to start remember things between runs.  With the memory in json, pilot remembers the window size and position.

The work splits into the store and its user. `solvcon.config.Config` represents the config file. `solvcon.pilot.base._ui_state.UiState` is the pilot feature that reads the interface state on startup and writes it back when the application is about to quit. Keeping the two apart means a later non-GUI setting can use `Config` without going through the pilot, and a later piece of interface state can join `UiState` without touching the file format.

The configuration file name is `pilot.json` in solvcon's configuration directory, resolved from `SOLVCON_CONFIG_HOME` when that is set, otherwise `solvcon` under `XDG_CONFIG_HOME`, otherwise `~/.config/solvcon`.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -199,6 +199,8 @@ some_object.do_something(with_some_information)
 ### Acronym
 
 Treat acronyms like a word. Do not make them all-upper-cases in names.
+The rule holds in both C++ and Python, and applies no matter how short
+the acronym is or how habitually it is written in capitals elsewhere.
 
 ```cpp
 // "Http" is treated like a word in CamelCase.
@@ -207,6 +209,14 @@ class HttpRequest
     // "http" is treated like a word in snake_case.
     void update_http_header();
 } /* end class HttpRequest */
+```
+
+```python
+# "Ui" is treated like a word in CamelCase, so it is not "UIState".
+class UiState(object):
+    # "ui" is treated like a word in snake_case.
+    def apply_ui_section(self, section):
+        pass
 ```
 
 ### Qt
--- a/doc/source/pilot/settings.md
+++ b/doc/source/pilot/settings.md
@@ -1,0 +1,36 @@
+# Settings
+
+Pilot keeps its settings in `pilot.json` in solvcon's configuration directory,
+resolved in this order:
+
+1. `$SOLVCON_CONFIG_HOME`, taken as the directory itself when it is set. This
+   keeps a test or a self-contained checkout from touching the settings of the
+   user who runs it.
+2. `$XDG_CONFIG_HOME/solvcon`, when `XDG_CONFIG_HOME` is set.
+3. `~/.config/solvcon` otherwise.
+
+Pilot starts from its defaults when the file is missing or cannot be parsed,
+so removing the file or breaking a hand edit does not stop the application.
+
+Settings are grouped by section. The `ui` section holds the state of the user
+interface, one entry per part of it that is remembered:
+
+```json
+{
+  "ui": {
+    "window": {
+      "height": 600,
+      "width": 1000,
+      "x": 100,
+      "y": 80
+    }
+  }
+}
+```
+
+The `window` entry is the main window's size and location, applied at startup
+and written back when the application quits. A saved location on a screen that
+is no longer connected is ignored, so the window cannot come back out of sight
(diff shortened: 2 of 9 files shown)


solvcon/pilot/base/_gui.py
```diff
@@ -125,6 +127,7 @@ def build(self, name="pilot", size=(1000, 600)):
         self.agent = _agent_gui.AgentPanel(mgr=self._rmgr)
         self.theme_menu = _theme.ThemeMenu(mgr=self._rmgr)
         self.window_manager = _window_manager.WindowManager(mgr=self._rmgr)
+        self.ui_state = _ui_state.UiState(mgr=self._rmgr)
```
yungyuc, line 130: Keep the UI state.

solvcon/pilot/base/_ui_state.py
```diff
@@ -0,0 +1,130 @@
...
+    def apply(self, section):
+        """Resize and move the window to match ``section``."""
+        apply_geometry(self._window, section)
+
+    def capture(self):
+        """Return the window's present size and location."""
+        return capture_geometry(self._window)
+
+
+class UiState(_gui_common.PilotFeature):
```
yungyuc, line 78: The state objects are ad hoc before we have enough parameters to reveal the pattern.

solvcon/config.py
```diff
@@ -0,0 +1,94 @@
...
+    def config_home():
+        """The directory holding solvcon's configuration files.
+
+        ``SOLVCON_CONFIG_HOME`` names the directory outright and wins when
+        it is set, which lets a test or a self-contained checkout keep its
+        settings away from the invoking user's. Otherwise the directory is
+        ``solvcon`` under ``XDG_CONFIG_HOME``, or under ``~/.config`` when
+        that is unset, the convention on the platforms solvcon targets.
+        """
+        home = os.environ.get("SOLVCON_CONFIG_HOME", "")
```
yungyuc, line 51: First try `$SOLVCON_CONFIG_HOME`.

solvcon/config.py
```diff
@@ -0,0 +1,94 @@
...
+        if home:
+            return home
+        base = os.environ.get("XDG_CONFIG_HOME", "")
```
yungyuc, line 54: Then the POSIX `$XDG_CONFIG_HOME`.

solvcon/config.py
```diff
@@ -0,0 +1,94 @@
...
+        if not base:
+            base = os.path.join(os.path.expanduser("~"), ".config")
```
yungyuc, line 56: Finally the default path.

doc/source/pilot/settings.md
```diff
@@ -0,0 +1,36 @@
+# Settings
```
yungyuc, line 1: Document to describe the configuration file.

STYLE.md
```diff
@@ -209,6 +211,14 @@ class HttpRequest
 } /* end class HttpRequest */
 ```
 
+```python
+# "Ui" is treated like a word in CamelCase, so it is not "UIState".
+class UiState(object):
```
yungyuc, line 216: Emphasize the style guide to remind CC.

solvcon/config.py
```diff
@@ -0,0 +1,94 @@
...
+        return os.path.join(base, "solvcon")
+
+    @classmethod
+    def default_path(cls):
+        """The default configuration path, ``<config_home>/<FILENAME>``."""
+        return os.path.join(cls.config_home(), cls.FILENAME)
+
+    def load(self):
```
tigercosmos, line 64: Maybe validate the json first before loading it?

yungyuc: Absolutely.  Updated issue #353.

PR #1186, opened 2026-07-26
Run the pilot GUI tests on the Linux and macOS CI jobs

tigercosmos:
Twenty-five pilot test classes are skipped whenever `GITHUB_ACTIONS` is set, most with the reason "GUI is not available in GitHub Actions". That reason is stale: the Linux build job now starts Xvfb and selects the xcb platform with the llvmpipe rasterizer, and the macOS build job selects cocoa. Both already drive real QRhi rendering through `test_pilot_domain_widget`, so the blanket skip only hid coverage that works.

Skip on what these classes actually cannot survive instead of on CI detection. The offscreen platform cannot back a live window, and the headless Windows runner faults the same way through its WARP software rasterizer, so Windows CI keeps skipping them. A Windows desktop with a real display still runs them.

This enables 141 tests on each of Linux and macOS: Linux goes from 1621 passed / 153 skipped to 1762 / 12, and macOS from 1622 / 152 to 1763 / 11.

### DIFF
--- a/tests/test_pilot_2d.py
+++ b/tests/test_pilot_2d.py
@@ -20,7 +20,10 @@ try:
 except ImportError:
     pilot = None
 
-GITHUB_ACTIONS = os.getenv('GITHUB_ACTIONS', False)
+# The offscreen platform cannot back a live window; neither can the headless
+# Windows CI runner, whose WARP software rasterizer faults on one.
+NO_LIVE_WINDOW = ((os.getenv('QT_QPA_PLATFORM') or '').startswith('offscreen')
+                  or ('nt' == os.name and bool(os.getenv('GITHUB_ACTIONS'))))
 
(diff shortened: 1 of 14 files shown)


tests/test_pilot_2d.py
```diff
@@ -350,8 +353,8 @@ def test_circle_renders_as_hollow_loop_on_locus(self):
                 "circle interior should be hollow")
 
 
-@unittest.skipIf(GITHUB_ACTIONS or not solvcon.HAS_PILOT,
-                 "live-GUI interaction is unstable under GitHub Actions")
+@unittest.skipIf(NO_LIVE_WINDOW or not solvcon.HAS_PILOT,
+                 "live-GUI interaction needs a real window surface")
```
tigercosmos, line 357: Enable GUI test for macOS and Linux.

PR #1192, opened 2026-07-26
Clarify profiler units and balance timed wrappers

### LINKED ISSUE #1187: Keep wrapper profiler state balanced when timed calls throw
## Description

A timed pybind11 call starts profiling in `precall` and ends it in `postcall`. If the C++ call throws, `postcall` is skipped and the failed node remains current.

See [common.hpp](https://github.com/solvcon/solvcon/blob/59031d387931f3d6ad6f9f7b682a39b9de83c380/cpp/solvcon/python/common.hpp#L100-L119).

## Reproducible example

```python
import solvcon

solvcon.call_profiler.reset()
try:
    solvcon.ConcreteBuffer(1, alignment=16)
except ValueError as exc:
    print("error:", exc)

failed = solvcon.call_profiler.result()["children"][0]
print("current after error:", failed.get("current_node", False))

solvcon.ConcreteBuffer(16, alignment=16)
failed = solvcon.call_profiler.result()["children"][0]
print("failed-call children:", [
    node["name"] for node in failed["children"]
])
```

## Bug result

```text
error: ConcreteBuffer::allocate: size 1 must be a multiple of alignment 16
current after error: True
failed-call children: ['ConcreteBuffer.__init__']
```

The later successful constructor is incorrectly nested under the failed call.

## Expected behavior

Exception unwinding must balance the profiler entry. No node should remain current, and the next constructor should be a top-level call. Add a regression test for this sequence.

Related to #302.

### LINKED ISSUE #1190: Use a monotonic clock and explicit units for profiler timing
## Description

Profiler intervals use `high_resolution_clock`, which is not monotonic in the prime Linux devenv. `result()` exports milliseconds while `serialize()` exports nanoseconds, but neither output identifies its unit. Consumers therefore cannot interpret `total_time` without knowing implementation details.

See [CallerProfile](https://github.com/solvcon/solvcon/blob/master/cpp/solvcon/profiling/RadixTree.hpp), [result()](https://github.com/solvcon/solvcon/blob/master/cpp/solvcon/profiling/pymod/wrap_profile.cpp), and [serialize()](https://github.com/solvcon/solvcon/blob/master/cpp/solvcon/profiling/SerializableProfiler.hpp).

## Reproducible example

```python
import json
import time
import solvcon

solvcon.call_profiler.reset()
probe = solvcon.CallProfilerProbe("unit-check")
time.sleep(0.01)
del probe

result = solvcon.call_profiler.result()
serialized = json.loads(solvcon.call_profiler.serialize())
print("result:", result["children"][0]["total_time"], result.get("time_unit"))
print("serialized:", serialized["radix_tree"]["children"][0]["total_time"], serialized.get("time_unit"))
```

The clock property can be checked with:

```cpp
#include <chrono>
#include <iostream>

int main()
{
    std::cout << std::boolalpha
              << std::chrono::high_resolution_clock::is_steady << ' '
              << std::chrono::steady_clock::is_steady << '\n';
}
```

## Bug result

```text
result: 10.1266 None
serialized: 10126600.0 None
false true
```

The factor of 1,000,000 reflects milliseconds versus nanoseconds, but neither output states its unit. The interval clock is also not steady.

## Expected behavior

Use `steady_clock` for profiler intervals. Preserve milliseconds in `result()` and nanoseconds in `serialize()`, and add an explicit `time_unit` field to each output. Update the Python printer, Pilot labels, schema, and tests to follow those contracts.

Related to #302 and #831.

ThreeMonth03:
`CallProfiler.result()` reports `total_time` in milliseconds while `CallProfiler.serialize()` reports it in nanoseconds, but neither output identifies its unit. This change adds `time_unit: "ms"` to `result()` and `time_unit: "ns"` to `serialize()` without changing either numeric representation, and labels millisecond values in the Python printer and Pilot UI.

Timed pybind11 wrappers start a profiler frame in `mmtag::precall()` and previously ended it in `postcall()`. pybind11 skips `postcall()` when the wrapped C++ call or return conversion throws, leaving the failed frame current and nesting later calls below it. This change moves frame retirement to a `pybind11::call_guard` destructor so successful and exceptional calls both balance the profiler stack while `precall()` still supplies the wrapper name.

Related to #1190.

Related to #1187.

### DIFF
--- a/cpp/solvcon/profiling/SerializableProfiler.hpp
+++ b/cpp/solvcon/profiling/SerializableProfiler.hpp
@@ -114,12 +114,14 @@ public:
     }
 
     MM_DECL_SERIALIZABLE(
+        register_member("time_unit", m_time_unit);
         register_member("radix_tree", m_root);
         register_member("id_map", m_id_map);
         register_member("unique_id", m_unique_id);)
 
 private:
 
+    std::string m_time_unit = "ns";
     SerializableRadixTreeNode m_root;
     std::unordered_map<std::string, key_type> m_id_map;
     key_type m_unique_id;
--- a/cpp/solvcon/profiling/pymod/wrap_profile.cpp
+++ b/cpp/solvcon/profiling/pymod/wrap_profile.cpp
@@ -114,6 +114,7 @@ pybind11::dict WrapCallProfiler::result(CallProfiler & profiler)
         return {};
     }
     py::dict result;
+    result["time_unit"] = "ms";
     std::queue<const RadixTreeNode<CallerProfile> *> node_queue;
     std::unordered_map<const RadixTreeNode<CallerProfile> *, py::dict> dict_storage;
 
--- a/cpp/solvcon/python/common.hpp
+++ b/cpp/solvcon/python/common.hpp
@@ -85,6 +85,36 @@ private:
 
 }; /* end class WrapperProfilerStatus */
 
+class WrapperProfilerGuard
+{
+
+public:
+
+    WrapperProfilerGuard()
+        : m_enabled(WrapperProfilerStatus::me().enabled())
+    {
+    }
+
+    WrapperProfilerGuard(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard(WrapperProfilerGuard &&) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard &&) = delete;
+
+    // NOLINTNEXTLINE(bugprone-exception-escape)
+    ~WrapperProfilerGuard()
+    {
+        if (m_enabled)
+        {
+            CallProfiler::instance().end_caller();
+        }
+    }
+
+private:
+
+    bool m_enabled;
+
+}; /* end class WrapperProfilerGuard */
+
 struct mmtag
 {
 }; /* end struct mmtag */
@@ -107,14 +137,6 @@ struct process_attribute<solvcon::python::mmtag>
         if (solvcon::python::WrapperProfilerStatus::me().enabled())
         {
             solvcon::CallProfiler::instance().start_caller(get_name(call), nullptr);
-        }
-    }
-
-    static void postcall(function_call &, handle &)
-    {
-        if (solvcon::python::WrapperProfilerStatus::me().enabled())
-        {
-            solvcon::CallProfiler::instance().end_caller();
         }
     }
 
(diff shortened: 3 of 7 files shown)


cpp/solvcon/profiling/pymod/wrap_profile.cpp
```diff
@@ -114,6 +114,7 @@ pybind11::dict WrapCallProfiler::result(CallProfiler & profiler)
         return {};
     }
     py::dict result;
+    result["time_unit"] = "ms";
```
ThreeMonth03, line 117: Explicitily clarify `time unit` as `ms` in `WrapCallProfiler::result(CallProfiler & profiler)`.

solvcon/profiling/_result.py
```diff
@@ -137,7 +137,7 @@ class ProfilingResultPrinter:
 
     Each dictionary must contain at least the following keys:
         - "name": Name of the function (str)
-        - "total_time": Total execution time in seconds (float)
+        - "total_time": Total execution time in milliseconds (float)
         - "count": Number of times the function was called (int)
         - "children": A list of child profiling results (list)
```
ThreeMonth03, line 142: Fix docstring typo in `class ProfilingResultPrinter`.

cpp/solvcon/profiling/SerializableProfiler.hpp
```diff
@@ -114,12 +114,14 @@ class SerializableRadixTree : public SerializableItem
...
 
     MM_DECL_SERIALIZABLE(
+        register_member("time_unit", m_time_unit);
         register_member("radix_tree", m_root);
         register_member("id_map", m_id_map);
         register_member("unique_id", m_unique_id);)
 
 private:
 
+    std::string m_time_unit = "ns";
```
ThreeMonth03, line 124: Explicitily clarify time unit as `ns` in `SerializableRadixTree`.

ThreeMonth03 (commented):
@tigercosmos @yungyuc Please review this pull request. Thank you.

cpp/solvcon/python/common.hpp
```diff
@@ -281,11 +304,13 @@ WrapBase
...
+// ref: https://pybind11.readthedocs.io/en/stable/advanced/functions.html#call-guard
 #define DECL_MM_PYBIND_CLASS_METHOD_TIMED(METHOD)                             \
     template <class... Args> /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
     wrapper_type & METHOD##_timed(Args &&... args)                            \
     {                                                                         \
-        m_cls.METHOD(std::forward<Args>(args)..., mmtag());                   \
+        using guard_type = pybind11::call_guard<WrapperProfilerGuard>;        \
+        m_cls.METHOD(std::forward<Args>(args)..., mmtag(), guard_type());     \
         return *static_cast<std::add_pointer_t<wrapper_type>>(this);          \
     }
```
ThreeMonth03, line 314: Replace `postcall()` with RAII guard, since `postcall()` is only called when a method exits normally.

cpp/solvcon/python/common.hpp
```diff
@@ -85,6 +85,37 @@ class WrapperProfilerStatus
...
+            CallProfiler::instance().end_caller();
+        }
+    }
+
+private:
+
+    bool m_enabled;
+
+}; /* end class WrapperProfilerGuard */
+
```
ThreeMonth03, line 117: RAII pattern.

ThreeMonth03 (commented):
@yungyuc @j8xixo12 Please review this pull request.

yungyuc (approved):
LGTM.

PR #1195, opened 2026-07-26
Balance timed wrapper profiling on exceptions

ThreeMonth03:
Timed pybind11 wrappers enter `CallProfiler` from the pre-call hook and leave it from the post-call hook. When the wrapped C++ call throws, pybind11 skips the post-call hook, so the failed frame remains current and the next call is nested below it.

```python
solvcon.wrapper_profiler_status.enable()
solvcon.call_profiler.reset()

try:
    solvcon.ConcreteBuffer(1, alignment=16)
except ValueError:
    pass

solvcon.ConcreteBuffer(16, alignment=16)
result = solvcon.call_profiler.result()
result["children"][0]["children"]  # contained the next constructor before this change
```

This change moves frame cleanup into a pybind11 call guard. Its destructor retires the frame after a successful call and during exception unwinding, while the existing pre-call hook still supplies the wrapper name.

The regression test runs a throwing constructor followed by a successful one and checks that both samples remain in one completed top-level node without a nested constructor.

Related to #1187.

### DIFF
--- a/cpp/solvcon/python/common.hpp
+++ b/cpp/solvcon/python/common.hpp
@@ -85,6 +85,37 @@ private:
 
 }; /* end class WrapperProfilerStatus */
 
+class WrapperProfilerGuard
+{
+
+public:
+
+    WrapperProfilerGuard()
+        : m_enabled(WrapperProfilerStatus::me().enabled())
+    {
+    }
+
+    WrapperProfilerGuard(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard(WrapperProfilerGuard &&) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard &&) = delete;
+
+    // FIXME: Remove the suppression after CallProfiler::end_caller() becomes noexcept.
+    // NOLINTNEXTLINE(bugprone-exception-escape)
+    ~WrapperProfilerGuard()
+    {
+        if (m_enabled)
+        {
+            CallProfiler::instance().end_caller();
+        }
+    }
+
+private:
+
+    bool m_enabled;
+
+}; /* end class WrapperProfilerGuard */
+
 struct mmtag
 {
 }; /* end struct mmtag */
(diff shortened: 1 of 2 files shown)


cpp/solvcon/python/common.hpp
```diff
@@ -281,11 +304,13 @@ WrapBase
...
+// ref: https://pybind11.readthedocs.io/en/stable/advanced/functions.html#call-guard
 #define DECL_MM_PYBIND_CLASS_METHOD_TIMED(METHOD)                             \
     template <class... Args> /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
     wrapper_type & METHOD##_timed(Args &&... args)                            \
     {                                                                         \
-        m_cls.METHOD(std::forward<Args>(args)..., mmtag());                   \
+        using guard_type = pybind11::call_guard<WrapperProfilerGuard>;        \
+        m_cls.METHOD(std::forward<Args>(args)..., mmtag(), guard_type());     \
         return *static_cast<std::add_pointer_t<wrapper_type>>(this);          \
     }
```
ThreeMonth03, line 315: Replace `postcall()` with RAII guard, since `postcall()` is only called when a method exits normally.

cpp/solvcon/python/common.hpp
```diff
@@ -85,6 +85,37 @@ class WrapperProfilerStatus
...
+        {
+            CallProfiler::instance().end_caller();
+        }
+    }
+
+private:
+
+    bool m_enabled;
+
+}; /* end class WrapperProfilerGuard */
```
ThreeMonth03, line 117: RAII pattern.

ThreeMonth03 (commented):
@yungyuc @tigercosmos Please review this pull request. Thank you.

cpp/solvcon/python/common.hpp
```diff
@@ -85,6 +85,37 @@ class WrapperProfilerStatus
...
+        : m_enabled(WrapperProfilerStatus::me().enabled())
+    {
+    }
+
+    WrapperProfilerGuard(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard(WrapperProfilerGuard &&) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard const &) = delete;
+    WrapperProfilerGuard & operator=(WrapperProfilerGuard &&) = delete;
+
+    // FIXME: Remove the suppression after CallProfiler::end_caller() becomes noexcept.
```
yungyuc, line 103: A FIXME for another ongoing PR means the two PRs should be combined.

yungyuc (changes requested):
Combine this PR #1195 and the PR #1192.  It's OK to use a single PR for two issues when they are tightly coupled.

yungyuc (changes requested):
Combine with PR #1195.

ThreeMonth03:
This change has been moved to #1192 in commit 1239827e. Closing this PR so both profiler corrections can be reviewed together.

Issue #1189, opened 2026-07-26
Keep profiler probes consistent during cancellation and reset


ThreeMonth03:
## Description

`CallProfilerProbe.cancel()` leaves its call-tree frame active. `CallProfiler.reset()` clears active profiler state without invalidating live probes, so their destructors later call `end_caller()` against an empty stack. Both APIs are exposed to Python.

See [CallProfilerProbe](https://github.com/solvcon/solvcon/blob/59031d387931f3d6ad6f9f7b682a39b9de83c380/cpp/solvcon/profiling/RadixTree.hpp#L329-L366) and [the binding](https://github.com/solvcon/solvcon/blob/59031d387931f3d6ad6f9f7b682a39b9de83c380/cpp/solvcon/profiling/pymod/wrap_profile.cpp#L150-L170).

## Reproducible examples

```python
import solvcon

solvcon.call_profiler.reset()
probe = solvcon.CallProfilerProbe("cancelled")
probe.cancel()
del probe
later = solvcon.CallProfilerProbe("later")
del later
root = solvcon.call_profiler.result()
cancelled = root["children"][0]
print("cancelled current:", cancelled.get("current_node", False))
print("cancelled children:", [node["name"] for node in cancelled["children"]])
```

```text
cancelled current: True
cancelled children: ['later']
```

```python
import solvcon

solvcon.call_profiler.reset()
probe = solvcon.CallProfilerProbe("active")
solvcon.call_profiler.reset()
del probe
print("survived")
```

Before the fix:

```text
Segmentation fault (exit status 139)
```

After the fix:

```text
survived
```

## Expected behavior

A probe may discard only the most recent active frame (LIFO), including its callback registration. Reset must invalidate all live probes before clearing profiler state. Later probes must start at the root, and destroying invalidated probes must be safe.

Related to #302.

tigercosmos:
LIFO makes sense to me.

ThreeMonth03:
Resolved by #1196.

PR #1196, opened 2026-07-26
Keep profiler probes consistent during cancellation and reset

ThreeMonth03:
`CallProfilerProbe.cancel()` suppresses destructor cleanup but leaves its call-tree frame and callback active. A direct `CallProfiler.reset()` has the inverse lifecycle problem: it clears profiler state without invalidating live probes, so their destructors later operate on an empty stack.

This change records each probe's frame identity and only discards the most recent active frame. Reset now marks live probes canceled before clearing profiler state, skips empty callbacks used by wrapper frames, and provides the safe path used by profiler-wide cancellation.

Related to #1189.

### DIFF
--- a/cpp/solvcon/profiling/RadixTree.cpp
+++ b/cpp/solvcon/profiling/RadixTree.cpp
@@ -5,11 +5,21 @@
 
 #include <solvcon/profiling/RadixTree.hpp>
 
+#include <stdexcept>
+
 namespace solvcon
 {
 
 void CallProfiler::reset()
 {
+    for (auto & cancel_callback : m_cancel_callbacks)
+    {
+        if (cancel_callback)
+        {
+            cancel_callback();
+        }
+    }
+
     while (!m_radix_tree.is_root())
     {
         CallerProfile & profile = m_radix_tree.get_current_node()->data();
@@ -40,8 +50,7 @@ void CallProfiler::end_caller()
     CallerProfile & call_profile = m_radix_tree.get_current_node()->data();
     call_profile.stop_stopwatch(); // Update profiling information to the pending time and count
     m_pending_nodes.insert(m_radix_tree.get_current_node());
-    m_radix_tree.move_current_to_parent(); // Pop the caller from the call stack
-    m_cancel_callbacks.pop_back();
+    pop_caller();
 
     if (m_radix_tree.get_current_node() == m_radix_tree.get_root()) // If the root function ends, update all pending nodes and stable items
     {
@@ -49,6 +58,22 @@ void CallProfiler::end_caller()
         m_radix_tree.update_stable_items();
     }
 }
+
+void CallProfiler::discard_caller(RadixTreeNode<CallerProfile> const * frame_node)
+{
+    if (frame_node != m_radix_tree.get_current_node())
+    {
+        throw std::runtime_error("CallProfilerProbe::cancel: only the most recent active probe can be canceled");
+    }
+    pop_caller();
+}
+
+void CallProfiler::pop_caller()
+{
+    m_radix_tree.move_current_to_parent();
+    m_cancel_callbacks.pop_back();
+}
+
 // NOLINTNEXTLINE(misc-no-recursion)
 void CallProfiler::print_profiling_result(const RadixTreeNode<CallerProfile> & node, const int depth, std::ostream & outstream) const
 {
--- a/cpp/solvcon/profiling/RadixTree.hpp
+++ b/cpp/solvcon/profiling/RadixTree.hpp
@@ -280,6 +280,9 @@ public:
     /// Called when a function ends
     void end_caller();
 
+    /// Discard the most recent caller without recording it
+    void discard_caller(RadixTreeNode<CallerProfile> const * frame_node);
+
     /// Print the profiling information
     void print_profiling_result(std::ostream & outstream) const
     {
@@ -290,18 +293,12 @@ public:
     void reset();
 
     /// Cancel the profiling from all probes
-    void cancel()
-    {
-        for (auto & cancel_callback : m_cancel_callbacks)
-        {
-            cancel_callback();
-        }
-        reset();
-    }
+    void cancel() { reset(); }
 
     const RadixTree<CallerProfile> & radix_tree() const { return m_radix_tree; }
 
 private:
+    void pop_caller();
     void print_profiling_result(const RadixTreeNode<CallerProfile> & node, int depth, std::ostream & outstream) const;
     static void print_statistics(const RadixTreeNode<CallerProfile> & node, std::ostream & outstream);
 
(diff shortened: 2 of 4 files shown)


cpp/solvcon/profiling/RadixTree.hpp
```diff
@@ -356,11 +360,16 @@ class CallProfilerProbe
 
     void cancel()
     {
-        m_cancel = true;
+        if (!m_cancel)
+        {
+            m_profiler.discard_caller(m_frame_depth);
+            m_cancel = true;
+        }
     }
```
ThreeMonth03, line 362: When people cancel probe actively, we dafaultly think they don't want the cancelled data. Therefore, `cancel()` should call `discard_caller()`, which only move the pointer to it's parent.

cpp/solvcon/profiling/RadixTree.cpp
```diff
@@ -40,15 +43,30 @@ void CallProfiler::end_caller()
...
+        throw std::runtime_error("CallProfilerProbe::cancel: only the most recent active probe can be canceled");
+    }
+    pop_caller();
+}
+
+void CallProfiler::pop_caller()
+{
+    m_radix_tree.move_current_to_parent();
+    m_cancel_callbacks.pop_back();
+}
```
ThreeMonth03, line 75: `end_caller()` and `discard_caller()` calls `pop_caller()` now.

cpp/solvcon/profiling/RadixTree.cpp
```diff
@@ -40,15 +43,30 @@ void CallProfiler::end_caller()
...
         m_radix_tree.update_stable_items();
     }
 }
+
+void CallProfiler::discard_caller(size_t frame_depth)
+{
+    if (frame_depth != m_cancel_callbacks.size())
+    {
+        throw std::runtime_error("CallProfilerProbe::cancel: only the most recent active probe can be canceled");
+    }
```
ThreeMonth03: `discard_caller()` would raise error if people delete the probe of the outer scope.

cpp/solvcon/profiling/RadixTree.hpp
```diff
@@ -335,9 +339,9 @@ class CallProfilerProbe
     {
         auto cancel_callback = [&]()
         {
-            cancel();
+            m_cancel = true;
```
ThreeMonth03, line 335: Tag whether the probe should be cancelled.

cpp/solvcon/profiling/RadixTree.cpp
```diff
@@ -5,11 +5,21 @@
...
 
 void CallProfiler::reset()
 {
+    for (auto & cancel_callback : m_cancel_callbacks)
+    {
+        if (cancel_callback)
+        {
+            cancel_callback();
+        }
+    }
```
ThreeMonth03, line 21: In `precall(function_call & call)`, the program calls `start_caller` with null `std::function`. After modification, it raises error now. A simple workaround is skipping null `std::function`. The problem might be fixed in the following PR.
```c++
static void precall(function_call & call)
    {
        if (solvcon::python::WrapperProfilerStatus::me().enabled())
        {
            solvcon::CallProfiler::instance().start_caller(get_name(call), nullptr);
        }
    }
```

cpp/solvcon/profiling/RadixTree.hpp
```diff
@@ -290,18 +293,12 @@ class CallProfiler
...
     /// Cancel the profiling from all probes
-    void cancel()
-    {
-        for (auto & cancel_callback : m_cancel_callbacks)
-        {
-            cancel_callback();
-        }
-        reset();
-    }
+    void cancel() { reset(); }
```
ThreeMonth03, line 296: The behavior of `cancel()` and `reset()` are same now.

ThreeMonth03 (commented):
@yungyuc @tigercosmos Please review this pull request.

yungyuc:
> @yungyuc @tigercosmos Please review this pull request.

It seems that you are still working on it.  Please leave a comment again to request for review when you are ready.

tests/test_callprofiler.py
```diff
@@ -324,4 +324,50 @@ def test_wrapper_exception_keeps_profiler_balanced(self):
...
+        del inner
+        del outer
+
+        later = solvcon.CallProfilerProbe("later")
+        del later
+
+        result = solvcon.call_profiler.result()
+        later_result, = result["children"]
+        self.assertEqual(later_result["name"], "later")
+        self.assertEqual(later_result["count"], 1)
```
ThreeMonth03, line 371: Test whether the profiler works when delete probe after reset.

tests/test_callprofiler.py
```diff
@@ -324,4 +324,50 @@ def test_wrapper_exception_keeps_profiler_balanced(self):
...
+        inner_result, = outer_result["children"]
+        self.assertEqual(outer_result["name"], "outer")
+        self.assertEqual(outer_result["count"], 0)
+        self.assertNotIn("current_node", outer_result)
+        self.assertEqual(inner_result["name"], "inner")
+        self.assertEqual(inner_result["count"], 0)
+        self.assertNotIn("current_node", inner_result)
+        self.assertEqual(later_result["name"], "later")
+        self.assertEqual(later_result["count"], 1)
+
```
ThreeMonth03, line 355: Test whether the profiler raises error when delete the outer probe.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/profiling/RadixTree.cpp
```diff
@@ -25,13 +35,14 @@ void CallProfiler::reset()
 }
 
 // Called when a function starts
-void CallProfiler::start_caller(const std::string & caller_name, const std::function<void()> & cancel_callback)
+size_t CallProfiler::start_caller(const std::string & caller_name, const std::function<void()> & cancel_callback)
```
tigercosmos: It's weird to return `int` for `start_caller`. I think it's better to modify the `m_frame_depth` counter directly instead.

cpp/solvcon/profiling/RadixTree.hpp
```diff
@@ -335,9 +332,9 @@ class CallProfilerProbe
     {
         auto cancel_callback = [&]()
         {
-            cancel();
+            m_cancel = true;
         };
-        m_profiler.start_caller(m_caller_name, cancel_callback);
+        m_frame_depth = m_profiler.start_caller(m_caller_name, cancel_callback);
```
tigercosmos: I would write:
```
m_profiler.start_caller(m_caller_name, cancel_callback);
m_frame_depth++;
```

tigercosmos (changes requested):
- [ ] discuss the return type of `CallProfiler::start_caller`.

yungyuc (approved):
LGTM.  @tigercosmos please help merge once you think it's OK.

tigercosmos (approved):
Cleaner. LGTM

cpp/solvcon/profiling/RadixTree.cpp
ThreeMonth03: Yes, `start_caller()` should not return anything. I afterwards revert the modification, and record `radix_tree` node as a frame.

Issue #1175, opened 2026-07-24
New UI design for Pilot 2D canvas and Painter (Root Issue)


tigercosmos:
The current Pilot 2D canvas and painter is just a prototype for development and testing. Make a new UI version, which have a better UI/UX.

### Dark
<img width="1270" height="849" alt="Image" src="https://github.com/user-attachments/assets/018544c7-e455-48d6-9832-500e3c7c2938" />
<img width="1271" height="847" alt="Image" src="https://github.com/user-attachments/assets/9e02ab52-06bb-4846-ac34-b99cbf6e034d" />
<img width="1269" height="850" alt="Image" src="https://github.com/user-attachments/assets/ff3372a5-1f9f-4ae5-8e35-c635d384862f" />

### Light

<img width="1177" height="784" alt="Image" src="https://github.com/user-attachments/assets/f42182b4-c06b-483a-a50f-31a09bf53902" />

<img width="1173" height="780" alt="Image" src="https://github.com/user-attachments/assets/3fa2cbc1-af5b-4c93-b787-0f21c1a93d86" />

<img width="1176" height="790" alt="Image" src="https://github.com/user-attachments/assets/93a8dd0d-ed8b-47b2-aad7-94bf34bcf528" />

yungyuc:
@j8xixo12 visualizer will need a similar improvement.  We should do it after @tigercosmos gets it work for the canvas.

yungyuc:
@tigercosmos Please make both light and dark modes.

tigercosmos:
@yungyuc  Just added.

yungyuc:
@tychuang1211 please take a look

tigercosmos:
First MVP is done in #1240, the follow-ups are at #1241.

PR #1197, opened 2026-07-26
Rebuild the Painter dock into the redesigned frame

tigercosmos:
The Painter dock now has the two-part body the redesign specifies: a 64px tool rail beside an inspector whose Design, Layers, and Canvas pages sit in a stack under a segmented tab row. This is the frame only. The rail keeps the existing draw-tool actions, and every rail button is still a view of the shared `draw.tool` action, so the rail and the Canvas/Draw tool menu cannot diverge. The three inspector pages are greyed-out placeholders naming what the later steps fill in. Rail entries carry short names (Pan, Tri, Rect) because the menu labels do not fit a 64px rail; each button keeps its menu text as its tooltip.

The rail shade, the accent pill on the active tool, and the segmented tabs all take their colors from the application palette rather than from fixed values, so the panel follows a light/dark switch. One Qt detail worth the reviewer's attention: a color written into a child widget's palette from a palette-change handler does not survive Qt finishing the propagation into that child, so the rail paints its own background and the styled pieces rebuild their style sheet on a palette event instead.

The tests cover the dock toggle, tab switching, rail-to-menu sync, that a dock wider than the design grows the inspector rather than leaving bands beside the panel, and that the rail shade tracks a light/dark switch.

Related to #1175.

Screenshot 2026-07-27 at 3 23 46 AM
Screenshot 2026-07-27 at 3 24 08 AM

### DIFF
--- a/solvcon/pilot/canvas/_painter_gui.py
+++ b/solvcon/pilot/canvas/_painter_gui.py
@@ -2,24 +2,325 @@
 # BSD 3-Clause License, see COPYING
 
 """
-Painter toolbox for the 2D canvas.
+Painter toolbox for the 2D canvas: the tool rail and the inspector.
 """
 
-from PySide6 import QtCore, QtWidgets
+from PySide6 import QtCore, QtGui, QtWidgets
 
 from ..base import _gui_common
 from .._pilot_core import draw_tool_names, default_draw_tool_name
 
 __all__ = [
+    'PainterPanel',
     'Painter',
 ]
 
 
+def _rule(shape):
+    """A hairline divider between two areas of the panel."""
+    line = QtWidgets.QFrame()
+    line.setFrameShape(shape)
+    line.setFrameShadow(QtWidgets.QFrame.Sunken)
+    return line
+
+
+def _blend(color, other, ratio):
+    """Mix ``ratio`` of ``other`` into ``color``."""
+    keep = 1.0 - ratio
+    return QtGui.QColor(
+        round(color.red() * keep + other.red() * ratio),
+        round(color.green() * keep + other.green() * ratio),
+        round(color.blue() * keep + other.blue() * ratio))
+
+
+def _shade(widget, ratio):
+    """The panel color of ``widget`` moved ``ratio`` toward its text color.
+
+    Blending toward the text is what makes a shade follow the theme: it steps
+    darker under a light palette and lighter under a dark one, which is how the
+    design separates the rail from the inspector and the checked tab from both.
+    """
+    palette = widget.palette()
+    return _blend(palette.color(QtGui.QPalette.Window),
+                  palette.color(QtGui.QPalette.WindowText), ratio)
+
+
+class _PaletteStyled(QtWidgets.QWidget):
+    """A widget whose child controls are styled from the palette.
+
+    A subclass builds its style sheet in :meth:`_apply_style` and calls it once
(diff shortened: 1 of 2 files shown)


solvcon/pilot/canvas/_painter_gui.py
```diff
@@ -2,24 +2,325 @@
...
+        if event.type() in self._RESTYLE_EVENTS:
+            self._apply_style()
+            self.update()
+        return super().event(event)
+
+    def _apply_style(self):
+        raise NotImplementedError
+
+
+class _ToolRail(_PaletteStyled):
```
tigercosmos, line 73: Only make the frame for the tool rail. The details will be implemented in the next PR.

solvcon/pilot/canvas/_painter_gui.py
```diff
@@ -2,24 +2,325 @@
...
+            }}
+            QToolButton:checked {{
+                background: {accent.name()};
+                color: {on_accent.name()};
+            }}
+            """)
+
+
+class _SegmentedTabs(_PaletteStyled):
+    """The Design / Layers / Canvas selector as one segmented control.
```
tigercosmos, line 155: Segemented tab component.

solvcon/pilot/canvas/_painter_gui.py
```diff
@@ -2,24 +2,325 @@
...
+            }}
+            QPushButton:checked {{
+                background: {pill.name()};
+                color: {text.name()};
+                font-weight: 500;
+            }}
+            """)
+
+
+class PainterPanel(QtWidgets.QWidget):
```
tigercosmos, line 224: The new panel for the whole Painter.

tests/test_pilot_draw_tool.py
```diff
@@ -60,5 +65,75 @@ def test_console_set_draw_tool_checks_the_menu(self):
...
+    def test_widening_the_panel_grows_the_inspector(self):
+        # The rail keeps its designed width and the inspector takes the rest,
+        # so a dock wider than the design leaves no bands beside the panel.
+        panel = self._panel()
+        panel.resize(500, 400)
+        panel.layout().activate()
+        self.assertEqual(panel._rail.width(), 64)
+        self.assertGreater(panel._inspector.width(), 264)
+
+    def test_tab_row_restyles_with_the_theme(self):
```
tigercosmos, line 126: Basic tests.

solvcon/pilot/canvas/_painter_gui.py
```diff
@@ -2,24 +2,325 @@
 # BSD 3-Clause License, see COPYING
 
 """
-Painter toolbox for the 2D canvas.
+Painter toolbox for the 2D canvas: the tool rail and the inspector.
```
yungyuc, line 5: I don't like the word `rail`.  It's not informative.  What does it really mean?   I'd like to have a description shedding insight.

solvcon/pilot/canvas/_painter_gui.py
yungyuc: `_ToolRail` is a name that lacks information.

yungyuc (commented):
It's good to see everything about the GUI logics is in Python.

solvcon/pilot/canvas/_painter_gui.py
tigercosmos: Sure, let me fix it.

PR #1193, opened 2026-07-26
Make the 2D canvas follow the application theme

tigercosmos:
The pilot 2D canvas painted from color constants compiled into RWorldRenderer2d and R2DWidget, so it stayed dark under the light theme while every other widget switched. A canvas fills every pixel itself and cannot inherit a QPalette, so the colors have to be handed to it.

This adds Canvas2dPalette beside ThemePalette in the Qt-free theme foundation, with a light and a dark table, and exposes the resolved one from RThemeManager. The renderer, the widget, and the rubber-band preview now read every color from it, and RManager hands the table to each canvas as the canvas is created and again on every theme change. theme_qt.hpp carries the one ThemeColor to QColor conversion the Qt consumers share.

R2DWidget.canvasPalette is exposed to Python as a name to (r, g, b) mapping, which is what the new tests in test_pilot_2d.py and test_pilot_theme.py check the canvas against. A gtest covers the two tables directly in the no-GUI target.

Related to #1062.

PR #1199, opened 2026-07-27
Follow up the canvas theme and Painter naming reviews

tigercosmos:
Two review comments from the canvas theme (#1193) and Painter frame (#1197) pull requests. Both were left after the branches merged, so they are addressed here instead.

RManager::add2DWidget() used to set up the canvas theme connection itself: it had to know that a canvas fills its own pixels and therefore cannot pick a theme switch up from the application palette the way an ordinary widget does. That knowledge belongs to the canvas rather than to whoever happens to create one, and a second creation site would have had to repeat it to get a canvas that follows the theme. R2DWidget::followTheme() now owns it, taking the current colors and connecting to themeChanged with the widget as the connection context, exactly as the old lambda did. The manager is left with one call. The console, the terminal, and the MDI backdrop still wire their own theme handlers in RManager; only the canvas moves here.

The Painter panel's _ToolRail named where the widget sits in the mockup rather than what it is, and "tool" reads as ambiguous in a file full of QToolButton. It is the exclusive selector for the active draw tool, so it becomes _DrawToolSelector, and the vocabulary around it follows: rail_labels becomes short_labels, RAIL_LABELS becomes SHORT_LABELS, and the comments say "selector" for the widget and "column" where the sentence is about its shape. No behavior changes; the existing tests are renamed to match and keep their assertions.

Related to #1175, and to the review comments on #1193 and #1197.

### DIFF
--- a/cpp/solvcon/pilot/app/RManager.cpp
+++ b/cpp/solvcon/pilot/app/RManager.cpp
@@ -159,16 +159,7 @@ R2DWidget * RManager::add2DWidget()
     if (m_mdiArea)
     {
         viewer = new R2DWidget(/*parent*/ m_mdiArea);
-        // The canvas fills its own pixels, so it cannot inherit a theme switch
-        // from the application palette the way an ordinary widget does. Hand it
-        // the colors now and again on every switch, with the canvas itself as
-        // the connection context so the link dies with it.
-        auto applyCanvasPalette = [this, viewer](ThemeVariant)
-        {
-            viewer->setCanvasPalette(m_themeManager->canvas2dPalette());
-        };
-        applyCanvasPalette(m_themeManager->currentVariant());
-        QObject::connect(m_themeManager, &RThemeManager::themeChanged, viewer, applyCanvasPalette);
+        viewer->followTheme(m_themeManager);
         viewer->setWindowTitle("2D canvas");
         viewer->show();
         auto * subwin = this->addSubWindow(viewer);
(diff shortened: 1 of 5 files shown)


cpp/solvcon/pilot/app/RManager.cpp
```diff
@@ -159,16 +159,7 @@ R2DWidget * RManager::add2DWidget()
...
-        // from the application palette the way an ordinary widget does. Hand it
-        // the colors now and again on every switch, with the canvas itself as
-        // the connection context so the link dies with it.
-        auto applyCanvasPalette = [this, viewer](ThemeVariant)
-        {
-            viewer->setCanvasPalette(m_themeManager->canvas2dPalette());
-        };
-        applyCanvasPalette(m_themeManager->currentVariant());
-        QObject::connect(m_themeManager, &RThemeManager::themeChanged, viewer, applyCanvasPalette);
+        viewer->followTheme(m_themeManager);
```
tigercosmos, line 162: Follow the theme in the canvas object.

solvcon/pilot/canvas/_painter_gui.py
```diff
@@ -70,7 +71,7 @@ def _apply_style(self):
         raise NotImplementedError
 
 
-class _ToolRail(_PaletteStyled):
+class _DrawToolSelector(_PaletteStyled):
```
tigercosmos, line 74: Renamed to "draw tool selector".

PR #1200, opened 2026-07-27
Show the selection on the Painter inspector's Design page

tigercosmos:
The Painter dock's Design tab has been a greyed-out summary of what it would eventually hold. This gives it a real page: a header naming what the active canvas has selected, with the shape's icon in the accent color, a type-derived name, and the "1 selected" badge, and an empty state when nothing is picked.

The world changes in C++ without a change signal, so the page re-reads it on the entity tree's 500 ms poll, and only while it is on screen. What the poll compares is the selection itself rather than a serialized world: that is all the page shows, it costs no serialization, and a plain selection click moves no geometry for a world-wide fingerprint to notice. A selection whose shape was removed or undone away reads as no selection at all, because the canvas keeps the id it stored and every query on a dead one throws.

The canvas is resolved through a callable on every read rather than held. R2DWidget is a C++ widget its sub-window owns, so a Python reference kept across that window's close outlives the object behind it; the dock hands the page the manager's own accessor instead, and refreshes it when the active sub-window changes.

One part of the diff is a verbatim move worth reading as such: the palette helpers the tool selector introduced go to a module of their own, unchanged but for the names losing their underscore, because a page importing them from the panel that imports the page back would close an import cycle.

The X/Y/W/H position grid and the greyed Stroke, Fill, Grid & snap, and Layers sections follow in a stacked pull request.

Related to #1175.

### DIFF
--- a/solvcon/pilot/canvas/_painter_design.py
+++ b/solvcon/pilot/canvas/_painter_design.py
@@ -1,0 +1,197 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+The Design page of the Painter inspector: what is selected and where it sits.
+"""
+
+from PySide6 import QtCore, QtGui, QtWidgets
+
+from . import _painter_icons
+from ._painter_style import blend, shade, PaletteStyled
+
+__all__ = [
+    'DesignPage',
+]
+
+
+def _mono_font():
+    """The stand-in for the mono font the design gives every number."""
+    return QtGui.QFontDatabase.systemFont(QtGui.QFontDatabase.FixedFont)
+
+
+class DesignPage(PaletteStyled):
+    """The inspector's Design page: what the canvas has selected.
+
+    The page reads the canvas it is bound to on a timer, because the world
+    changes in C++ without a change signal. What the poll compares is the
+    selection itself, which is both what the page shows and cheap enough to
+    read every tick.
+    """
+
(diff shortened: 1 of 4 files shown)


solvcon/pilot/canvas/_painter_design.py
```diff
@@ -0,0 +1,197 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+The Design page of the Painter inspector: what is selected and where it sits.
```
tigercosmos, line 5: Add the design page implementation.

solvcon/pilot/canvas/_painter_style.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+Palette-derived styling shared by the pieces of the Painter panel.
```
tigercosmos, line 5: Share the common components.

tests/test_pilot_painter_design.py
```diff
@@ -0,0 +1,177 @@
...
+        self.world = world
+        self.selectedShape = -1
+        self.repaints = 0
+
+    def requestRepaint(self):
+        self.repaints += 1
+
+
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class PainterDesignPageTC(unittest.TestCase):
```
tigercosmos, line 59: Basic UI tests.

PR #1201, opened 2026-07-27
Give the Painter Design page its position grid

tigercosmos:
The Design page names the selection but says nothing about where it sits. This adds the design's X/Y/W/H grid, read from the selection's oriented bounding box: X and Y are its center, W and H the box's own sides, so a rotated shape reports the size it was drawn with rather than the wider span it covers along the axes. The editors ask for no width of their own, so opening the Painter no longer widens the dock past the inspector the design draws.

X and Y are editable. A field commits when the user has typed in it and the text parses to a number the shape does not already hold: the text is a rounded view of the value, so a focus change on an untouched field would otherwise write that rounding back as a move. One finished edit consumes the typing, so the text Enter leaves behind is not committed a second time over whatever moved since. The edit goes through translate_shape inside one begin_operation bracket, so undo takes it back in a single step. W and H stay read-only until the world grows a scale operation.

Arithmetic that leaves the double range is turned away rather than written to the world, since a shape with a non-finite coordinate is not brought back by an inverse translation that is non-finite too. That covers input that is not finite, a step between two finite coordinates that overflows, and a finite step that carries a wide shape's far corner past the range; the center is averaged corner by corner for the same reason.

Stroke, Fill, Grid & snap, and the mini Layers list take their designed places on the page, greyed out and naming what each one waits for.

Related to #1175.

### DIFF
--- a/solvcon/pilot/canvas/_painter_design.py
+++ b/solvcon/pilot/canvas/_painter_design.py
@@ -5,10 +5,12 @@
 The Design page of the Painter inspector: what is selected and where it sits.
 """
 
+import math
+
 from PySide6 import QtCore, QtGui, QtWidgets
 
 from . import _painter_icons
-from ._painter_style import blend, shade, PaletteStyled
+from ._painter_style import blend, rule, shade, PaletteStyled
 
 __all__ = [
     'DesignPage',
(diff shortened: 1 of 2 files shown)


solvcon/pilot/canvas/_painter_design.py
```diff
@@ -20,39 +22,211 @@ def _mono_font():
...
+
+    The shape lies inside its own box, so corners that survive the step bound
+    the geometry that follows them.
+    """
+    dx, dy = delta
+    return all(math.isfinite(x + dx) and math.isfinite(y + dy)
+               for x, y in zip(obb[0::2], obb[1::2]))
+
+
+class _Field(QtWidgets.QFrame):
```
tigercosmos, line 57: Field component for X, Y values.

solvcon/pilot/canvas/_painter_design.py
```diff
@@ -20,39 +22,211 @@ def _mono_font():
...
 
-    # Poll period in milliseconds, the cadence the entity tree already uses.
-    # The timer runs only while the page is on screen, so a hidden dock or
-    # another inspector tab costs nothing.
-    _POLL_MS = 500
+    # Poll period in milliseconds. The fields are read while a drag moves the
+    # shape under them, so the entity tree's half-second cadence reads as lag;
+    # one shape's box costs little enough to look at this often, and the timer
+    # runs only while the page is on screen.
+    _POLL_MS = 60
```
tigercosmos, line 200: Make the poll to 60ms to reflect the realtime position while moving the shape.

solvcon/pilot/canvas/_painter_design.py
```diff
@@ -175,22 +395,71 @@ def _show_icon(self):
             self._ICON_PX, self.devicePixelRatioF()))
         self._icon.setVisible(True)
 
+    def _on_committed(self, axis, value):
```
tigercosmos, line 398: on commit the position.

solvcon/pilot/canvas/_painter_design.py
yungyuc: It must not be named as `_Field` or `Field`.  It collides with solution field that will be used.

Please create a follow-up issue to rename it.  Not urgent though.

solvcon/pilot/canvas/_painter_design.py
```diff
@@ -175,22 +395,71 @@ def _show_icon(self):
...
+            world.translate_shape(shape, *delta)
+        finally:
+            world.end_operation()
+
+        widget.requestRepaint()
+        self.refresh()
+
     def _apply_style(self):
-        """Color the header and the badge from the palette."""
+        """Color the header, the badge, and the fields from the palette."""
```
yungyuc, line 426: It does not look like a good idea to hard-code style sheets in Python code.  When there is an opportunity, it should be changed to use a more organized structure.

tigercosmos: Sure

PR #1202, opened 2026-07-28
Keep the pilot GUI test windows off the screen

tigercosmos:
The pilot GUI tests build real top-level windows, and every one that maps takes the desktop and the keyboard from whoever is running the suite. This adds tests/conftest.py, which marks each top-level widget WA_DontShowOnScreen as it is shown: the window is still created, laid out, and painted into its backing store, and it never reaches the screen. The tests therefore keep running against a live window surface, which is what QT_QPA_PLATFORM=offscreen cannot give them, since their own NO_LIVE_WINDOW guards skip 217 of the 398 pilot tests there.

Exporting SOLVCON_TEST_SHOW_WINDOWS=ON maps the windows again, for watching a GUI test that is misbehaving. The CI jobs that run the GUI tests export it so their coverage is unchanged, except the Linux make pytest job, which stays at the default so one job runs them off the screen. Nothing else in the Makefile changes: the environment variable already passes through both pytest targets, so the diff there is comments only.

One result changes: a window that never maps is never activated, so the keyboard probe already in test_pilot_shortcut.py skips its two live key-delivery tests in a default local run. They still run in CI and under SOLVCON_TEST_SHOW_WINDOWS=ON.

Locally, make pytest and make run_pilot_pytest each report 1790 passed, 13 skipped, 3 xfailed, and make lint is clean.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -224,6 +224,8 @@ jobs:
           install_name_tool -add_rpath $(qtpaths --install-prefix)/lib ${PYSIDE6_PATH}/QtCore.abi3.so
           JOB_MAKE_ARGS="${JOB_MAKE_ARGS} BUILD_METAL=ON" ; \
         fi
+        # Left at the default so one job runs the GUI tests with their windows
+        # off the screen; run_pilot_pytest below maps its windows.
         make pytest ${JOB_MAKE_ARGS}
         echo "::endgroup::"
 
@@ -242,6 +244,9 @@ jobs:
       run: |
         echo "::group::make run_pilot_pytest"
         export LD_LIBRARY_PATH=$(python3 -c "import sys, os, shiboken6; sys.stdout.write(os.path.dirname(shiboken6.__file__))")
+        # A runner has no desktop to take over, so let this job map its windows
+        # and keep the tests that need a live surface out of the skips.
+        export SOLVCON_TEST_SHOW_WINDOWS=ON
         make run_pilot_pytest VERBOSE=1
         echo "::endgroup::"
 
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def _build_hider():
+    class WindowHider(QtCore.QObject):
+        def eventFilter(self, obj, event):
+            if (QtCore.QEvent.Type.Show == event.type()
+                    and isinstance(obj, QtWidgets.QWidget)
+                    and obj.isWindow()):
+                obj.setAttribute(
+                    QtCore.Qt.WidgetAttribute.WA_DontShowOnScreen, True)
```
tigercosmos, line 39: Not show windows on the screen.

.github/workflows/devbuild.yml
```diff
@@ -430,6 +435,9 @@ jobs:
       - name: cmake run_pilot_pytest
         run: |
           echo "::group::cmake run_pilot_pytest"
+          # A runner has no desktop to take over, so let this job map its
+          # windows, the way it has always run them.
+          $env:SOLVCON_TEST_SHOW_WINDOWS = "ON"
```
tigercosmos, line 440: Still enable `SOLVCON_TEST_SHOW_WINDOWS` to `ON` for CI.

Makefile
```diff
@@ -136,6 +136,8 @@ install: cmake
 #   make pytest PYTEST_OPTS='-k test_buffer.py'
 # Example for one class:
 #   make pytest PYTEST_OPTS='-v -k SimpleArrayBasicTC'
+# The GUI tests keep their windows off the screen (tests/conftest.py); export
+# SOLVCON_TEST_SHOW_WINDOWS=ON to watch them.
```
tigercosmos, line 140: `SOLVCON_TEST_SHOW_WINDOWS` = `off` by default.

yungyuc:
LGTM.

BTW, the unit tests for pilot GUI start to run slowly.  We should improve that later for better devexp.

Issue #1206, opened 2026-07-28
Agent MVP: multi-turn chat with a draw agent that frames its work


tigercosmos:
## Summary

This is the MVP re-cut of the agent harness plan in #966. That plan grew to 20 PRs because it committed to every frontier harness technique up front (typed events, capability descriptors, native tool calling, artifact store, policy engine, controllers, eval matrix). The MVP inverts the priority: ship the two things a user actually experiences, a multi-turn chat in the pilot and draw output that lands centered and legible (the #1136 findings), in 5 steps. Everything else in #966 remains valid as post-MVP work layered on top.

The plan was design-reviewed and re-audited against master: the prompt-engineering slice of #1136 (coordinate-frame instructions, system-prompt separation, baseline CLI pinning) has already landed, three formerly deferred pieces are pulled in as reduced forms because the steps cannot work without them, and the demo, concurrency-token, prompt-growth, and auto-fit semantics are specified up front.

## What the MVP delivers

1. A first-time user types "draw a truck" in the Agent console and sees a well-framed truck without touching pan or zoom (#1136's framing fix, already landed, plus this plan's auto-fit).
2. A follow-up edit works because the session keeps the conversation and the model sees a compact inventory of what is on the canvas. The guaranteed demo is "move the truck to the right" (`translate_shape` by inventory id), because the landed command surface has no scale and delete ops are hidden by default, so resizing is not expressible today. "Make the truck bigger" becomes a demo the day the staged `scale_shape` PR on the primitives track lands.
3. Within one prompt, the model can recover from its own failed commands: results and errors go back to the model under a small step budget instead of dying silently.
4. The same session runs headless (`python -m solvcon.agent`) against the Claude CLI or any OpenAI-compatible endpoint, with scripted smoke tests in CI.

## What the MVP keeps simple, and the three reduced forms it cannot skip

The wire protocol stays what is landed: composed flat text in, JSON command array out. No full typed event vocabulary, no `AgentRequest` object, no capability descriptors, no native tool calling, no artifact store, no MCP transport. The multi-turn history is the transcript replayed into the composed prompt under explicit byte budgets. This works identically on frontier backends and small local models.

Three items the original plan deferred turn out to be prerequisites in reduced form, and are in scope:

[...]

PR #1207, opened 2026-07-28
Compact the agent tool surface and pin the CLI environment

tigercosmos:
The command vocabulary reaches the model as a JSON Schema dump, so it costs about 27 KB on every request. This renders it as one signature line per op instead, keeping the required and optional arguments, defaults, numeric bounds, array lengths, nested shapes, and enumerations, and drops the argument descriptions the signature made redundant. Forty-six such descriptions go; twenty-six stay because each names a mistake the schema leaves open, such as radius against diameter. The composed payload for an empty world falls from 27000 to 3277 bytes.

Related to #1206, #1136, and #966.

### DIFF
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -18,6 +18,130 @@ import dataclasses
 import json
 
 
+class ToolSurfaceFormatter:
+    """The renderers that turn JSON Schema tool definitions into the compact
+    per-op signatures the model reads."""
+
+    SCALAR_TYPES = {"number": "num", "integer": "int", "string": "str",
+                    "boolean": "bool", "null": "null"}
+
+    BOUNDS = (("minimum", ">="), ("exclusiveMinimum", ">"),
+              ("maximum", "<="), ("exclusiveMaximum", "<"))
+
+    @classmethod
+    def literal(cls, value):
+        return json.dumps(value, separators=(",", ":"))
+
+    @classmethod
+    def one_line(cls, text):
+        return " ".join(str(text).split())
+
+    @classmethod
+    def array_type(cls, schema):
+        item = cls.type_name(schema.get("items", {}))
+        low, high = schema.get("minItems"), schema.get("maxItems")
+        if low is None and high is None:
+            return "[%s]" % item
+        if low == high:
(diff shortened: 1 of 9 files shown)


solvcon/agent/_backend.py
```diff
@@ -18,6 +18,130 @@
 import json
 
 
+class ToolSurfaceFormatter:
```
tigercosmos, line 21: Example output:
```
  [create]
  add_point(x: num, y: num, z?: num = 0.0) -> {npoint: int}
    Add a free point at world (x, y, z); z is dropped in 2D.
  add_circle(cx: num, cy: num, r: num>0) -> {shape_id: int}
    Add a circle shape centered at (cx, cy).
    r: Radius.
  add_bezier(p0: [num]x2..3, p1: [num]x2..3, p2: [num]x2..3, p3: [num]x2..3) -> {nbezier: int}
    Add a bare cubic Bezier from four control points.
    p0: Start anchor.
    ...
  add_polygon(vertices: [[num]x2]x3+) -> {shape_id: int}
    Add a closed polygon through a list of [x, y] vertices.
    vertices: The last vertex connects back to the first.
```

solvcon/agent/_backend.py
```diff
@@ -48,6 +172,20 @@ class AgentBackend(abc.ABC):
         "the operations below: draw on the canvas, open or arrange canvas "
         "windows, and pan or zoom the view. Reply with only that array.\n"
         "\n"
+        "Reading the operation list. Operations are listed under a "
```
tigercosmos, line 175: Add the prompt for the operation list.

solvcon/agent/_backends_impl.py
```diff
@@ -29,94 +29,100 @@
...
-            raise ValueError("command is not an object: %r" % (entry,))
-        op = entry.get("op")
-        if not isinstance(op, str):
-            raise ValueError("command needs a string \"op\": %r" % (entry,))
-        if valid and op not in valid:
-            raise ValueError("unknown op %r" % (op,))
-        commands.append(entry)
-    return commands
+    @classmethod
+    def load_json_payload(cls, text):
```
tigercosmos, line 64: Also move the free functions to helper classes.

PR #1208, opened 2026-07-28
Complete generic planned matmul operation

ThreeMonth03:
The parent #1181 traverses equal matrix batch shapes, but it neither broadcasts unequal leading axes nor accepts rank-1 operands. NumPy matmul needs both behaviors for `(2,1,M,K) @ (1,5,K,N)`, `(K,) @ (...,K,N)`, `(...,M,K) @ (K,)`, and `(K,) @ (K,)`.

This stacked change completes the generic planned roles. `MatmulPlan` aligns leading batch axes with zero strides and records whether each operand is a vector or matrix. Synthetic M=1 or N=1 lets one signed-stride executor evaluate DOT, GEVM, GEMV, and GEMM, while output-shape construction removes the synthetic axes and retains the existing `(1,)` dot-result convention. Operands are not expanded.

Differential tests cover rank alignment, empty domains, incompatible shapes, signed vector, batch, and contraction strides, and C-, F-, negative-, and step-two layouts. `profile_matrix_ops.py` measures broadcast batch matrix-matrix, DOT, single GEVM/GEMV, and batch GEVM/GEMV across all four independent LHS/RHS C-contiguous or negative-contraction-stride combinations.

## Profiling

The benchmark ran at `4e42c825` on an Apple M1 with Python 3.14.6, NumPy 2.5.1, and Accelerate. Each plotted point is the median of individual calls from five rounds. Each round uses two warmups and 15 measured calls per implementation.

At `S = 1024`, C-contiguous batch GEVM/GEMV planned-to-NumPy ratios are 14.044/7.102 for float32 and 6.412/3.929 for float64. The other three stride combinations range from 0.783 to 0.984; mixed or negative DOT at `S = 1048576` ranges from 0.999 to 1.000. NumPy remains faster for large C-contiguous inputs and side-256 broadcast batches.

### float32 scaling by operation and layout (NumPy 2.5.1)

Float32 planned matmul scaling with explicit LHS, RHS, and output shape and element stride labels

### float64 scaling by operation and layout (NumPy 2.5.1)

Float64 planned matmul scaling with explicit LHS, RHS, and output shape and element stride labels

[Download the complete profile output.](https://github.com/user-attachments/files/30590397/pr1208-profile-matrix-ops-4e42c825.txt)

Matrix correctness passes all 58 tests and 498 subtests in `tests/test_matrix.py`.

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -39,17 +39,21 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
  * the result shape, and records the signed-stride mappings used to locate
  * operands in the result batch domain. Broadcast batch axes are represented
  * by zero strides, so an executor can advance operand offsets without
- * reinterpreting ranks or broadcasting rules.
+ * reinterpreting operand dimensions or broadcasting rules.
  *
  * For `(2,1,3,4) @ (1,5,4,6)`, the plan records batch shape `(2,5)`, M=3,
  * N=6, K=4, output shape `(2,5,3,6)`, and zero-stride batch mappings for
  * the broadcast axes. It does not allocate the output or evaluate the ten
  * matrix pairs.
  *
- * @note This implementation plans matrix-matrix operands with equal ranks
- * and equal leading batch shapes. It records the batch domain and mappings,
- * M, N, K, the output shape, and signed matrix strides. Vector roles and
- * broadcast batch axes are follow-up work.
+ * For `(K,) @ (B,K,N)`, the plan keeps a synthetic M=1 for execution but
+ * omits that axis from the `(B,N)` output. Matrix-vector products similarly
+ * omit synthetic N=1, while vector-vector products retain SimpleArray's
+ * existing `(1,)` dot-result convention.
+ *
+ * @note This implementation plans vector and matrix operand roles with
+ * broadcast leading batch shapes. It records the batch domain and mappings,
+ * M, N, K, the output shape, and signed contraction strides.
  */
 class MatmulPlan
 {
@@ -78,18 +82,20 @@ private:
 
     struct Contraction
     {
+        bool m_lhs_vector;
+        bool m_rhs_vector;
         ssize_t m_rows;
         ssize_t m_columns;
         ssize_t m_inner_size;
     }; /* end struct Contraction */
 
-    struct MatrixStrides
+    struct ContractionStrides
     {
         ssize_t m_lhs_row_stride;
         ssize_t m_lhs_inner_stride;
         ssize_t m_rhs_inner_stride;
         ssize_t m_rhs_column_stride;
-    }; /* end struct MatrixStrides */
+    }; /* end struct ContractionStrides */
 
     struct BatchMappings
     {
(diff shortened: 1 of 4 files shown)


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -175,14 +193,7 @@ template <typename Array>
...
-    MatrixStrides const strides{
-        .m_lhs_row_stride = lhs.stride(lhs_matrix_axis),
-        .m_lhs_inner_stride = lhs.stride(lhs_matrix_axis + 1),
-        .m_rhs_inner_stride = rhs.stride(rhs_matrix_axis),
-        .m_rhs_column_stride = rhs.stride(rhs_matrix_axis + 1),
-    };
+    ContractionStrides const strides = make_contraction_strides(lhs, rhs, contraction);
     BatchMappings batch = make_batch_mappings(lhs, rhs, contraction.m_rows * contraction.m_columns);
     shape_type output_shape = make_output_shape(batch, contraction);
     return MatmulPlan{
```
ThreeMonth03, line 199: Calculate contraction strides in `make_contraction_strides(lhs, rhs, contraction);` now.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -196,78 +207,178 @@ MatmulPlan MatmulPlan::make(Array const & lhs, Array const & rhs)
...
+        --rhs_inner_axis;
+        rhs_column_stride = rhs.stride(rhs.ndim() - 1);
     }
+    return ContractionStrides{
+        .m_lhs_row_stride = lhs_row_stride,
+        .m_lhs_inner_stride = lhs.stride(lhs.ndim() - 1),
+        .m_rhs_inner_stride = rhs.stride(rhs_inner_axis),
+        .m_rhs_column_stride = rhs_column_stride,
+    };
+}
```
ThreeMonth03, line 277: `MatmulPlan::make_contraction(Array const & lhs, Array const & rhs)` and `MatmulPlan::make_contraction_strides(
    Array const & lhs,
    Array const & rhs,
    Contraction const & contraction)` can deal with 1D array now. They will treat 1D array as a 2D array, which extra dimension has shape 1 and stride 0.

For example:
```
lhs shape   = (4,), stride = (-1,)
rhs shape   = (4,), stride = (2,)
```
`make_contraction` and `make_contraction_strides` process and record shape and stride to:
```
lhs_vector = true
rhs_vector = true

M = 1
N = 1
K = 4

lhs_row_stride    = 0
lhs_inner_stride  = -1
rhs_inner_stride  = 2
rhs_column_stride = 0

output shape = (1,)
```

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -196,78 +207,178 @@ MatmulPlan MatmulPlan::make(Array const & lhs, Array const & rhs)
...
+    {
+        size_t const domain_axis = batch_axis_offset + axis;
+        if (operand.shape(axis) == domain.extent(domain_axis))
+        {
+            strides[domain_axis] = operand.stride(axis);
+        }
+    }
+    return OperandMapping(std::move(strides));
+}
+
```
ThreeMonth03, line 351: `MatmulPlan::make_batch_shape(Array const & lhs, Array const & rhs)` and `MatmulPlan::make_batch_mapping(Array const & operand, LoopDomain const & domain)` can now handle broadcasting over leading batch axes. They right-align the batch shapes, preserve signed strides for matching axes, and use stride 0 for broadcast axes.

For example:

```text
lhs shape   = (2, 1, 3, 4), strides = (12, 12, 4, 1)
rhs shape   = (1, 5, 4, 6), strides = (120, 24, 6, 1)
```

`make_batch_shape` and `make_batch_mapping` process and record the batch shape and mappings as:

```text
batch shape = (2, 5)

lhs batch mapping = (12, 0)
rhs batch mapping = (0, 24)

output shape = (2, 5, 3, 6)
```

ThreeMonth03: By the way, broadcasting function may be moved to `loop.hpp` in the future, but I think it could be done in the follow-up issue, since different operation broacast different axis. 

For example, when broadcast arrays for matrix multiplication operation, it should not consider the contraction part. Whereas, it should broadcast whole array for elementwise operations.

ThreeMonth03 (commented):
@yungyuc Please take a look.

profiling/profile_matrix_ops.py
yungyuc, line 1: @KHLee529 Please take a look at the current profiling code.

yungyuc (approved):
LGTM.  The routing looks reasonable.

PR #1209, opened 2026-07-28
Add direct BLAS routes for DOT, unbatched GEVM/GEMV, and BLAS-compatible GEMM

ThreeMonth03:
## Motivation

Before this PR, `matmul_planned()` evaluated every contraction with generic C++ loops, even when the operand layout could be passed to CBLAS without copying. On the tested Apple M1 system, NumPy uses Accelerate for the same compatible workloads. As the contraction grows, the generic path can become hundreds of times slower than NumPy.

## Approach

`MatmulExecutor` sends sufficiently large DOT, unbatched vector products, and matrix-matrix products directly to CBLAS when their strides can be represented without copying. GEMM dispatch happens after batch offsets are resolved, so regular batch and batch-broadcast products use the same route. Smaller operations and unsupported layouts continue to use the generic signed-stride loops.

## Profiling

The plots compare full-call execution time among NumPy 2.5.1, the generic `matmul_planned()` implementation before this PR, and the automatic dispatch added by this PR. The vertical line marks the measured dispatch threshold; points to its left remain on the generic route as small-workload controls.

Above the thresholds, the after-PR route tracks NumPy across direct DOT, unbatched vector products, 2D GEMM, equal-batch GEMM, and cross-broadcast GEMM. At the largest plotted GEMM sizes, it is hundreds of times faster than the before-PR generic loops.

### float32 scaling by operation and layout

pr1209-dot-float32

pr1209-matrix-vector-float32

pr1209-gemm-2d-float32

pr1209-gemm-batch-broadcast-float32

### float64 scaling by operation and layout

pr1209-dot-float64

pr1209-matrix-vector-float64

pr1209-gemm-2d-float64

pr1209-gemm-batch-broadcast-float64

<details>
<summary>Measurement details</summary>

- Environment: native Apple M1 arm64, macOS 26.5.1, Python 3.14.6, NumPy 2.5.1, and Accelerate. BLAS-related thread variables were set to `1` before importing NumPy.

- Baselines: the before-PR series uses the generic `matmul_planned()` implementation from the parent of this PR. The after-PR series uses the automatic dispatch introduced here. Both use the same workloads, seed, dependencies, and sampling procedure.

- Timing: input construction, conversion, correctness checks, and warmups were outside timing. Each timed call includes wrapper dispatch, computation, result allocation, and return.

- Sampling: seed `1209`, five rounds per case, two warmups, and 15 individually profiled calls per implementation per round. Plots use the untrimmed median of 75 calls. The four C-contiguous `S=256` batch GEMM cases were independently repeated five times; their ratio ranges were `0.951-1.031` and their medians were `0.995-1.008`.

- Thresholds: forced generic and CBLAS passes used identical inputs in generic, CBLAS, CBLAS, generic order. Calibrated timing chunks and 1,632 vector-matrix and matrix-vector layout and shape combinations selected `K >= 128` for DOT, `K * N >= 729` for compact C-layout vector-matrix products, a minimum matrix dimension of `32` for other vector products, and a minimum M/N/K dimension of `8` for GEMM.

</details>

<details>
<summary>After-PR raw profile output</summary>

[Download the complete after-PR profile output.](https://github.com/user-attachments/files/30631476/pr1209-profile-matrix-ops-2f5f055b.txt)

</details>

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -13,6 +13,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <format>
+#include <optional>
 #include <ranges>
 #include <stdexcept>
 #include <string>
@@ -72,6 +73,8 @@ public:
     ssize_t rhs_inner_stride() const noexcept { return m_strides.m_rhs_inner_stride; }
     ssize_t rhs_column_stride() const noexcept { return m_strides.m_rhs_column_stride; }
 
+    bool lhs_is_vector() const noexcept { return m_contraction.m_lhs_vector; }
+    bool rhs_is_vector() const noexcept { return m_contraction.m_rhs_vector; }
     bool has_batch_axes() const noexcept { return m_batch.m_domain.rank() != 0; }
     MappedOffsetCursor batch_cursor() const & { return MappedOffsetCursor(m_batch.m_domain, m_batch.m_mappings); }
     MappedOffsetCursor batch_cursor() const && = delete;
@@ -147,9 +150,9 @@ private:
  * evaluates one `(3,4) @ (4,6)` contraction at each offset. The results are
  * written into the allocated `(2,5,3,6)` output.
  *
- * @note This implementation provides generic signed-stride DOT, GEMV, GEVM,
- * and GEMM routes with broadcast batch shapes. Optimized routes are
- * follow-up work.
+ * @note This implementation provides generic signed-stride routes and direct
+ * BLAS routes for unit-stride DOT, positive-increment GEMV and GEVM, and
+ * C/F-compatible GEMM. Packing and batched-vector tuning are follow-up work.
  */
 template <typename Array>
 class MatmulExecutor
@@ -161,6 +164,8 @@ public:
 
 private:
     using value_type = typename Array::value_type;
+    using matrix_view_type = BlasMatrixView<value_type>;
+    using vector_view_type = BlasVectorView<value_type>;
 
     enum class MappingSlot : std::uint8_t
     {
@@ -169,6 +174,23 @@ private:
         Rhs,
     };
 
+    static constexpr ssize_t BLAS_DOT_MIN_LENGTH = 128;
+    static constexpr ssize_t BLAS_COMPACT_GEVM_MIN_ELEMENTS = 729;
+    static constexpr ssize_t BLAS_GEMV_MIN_DIMENSION = 32;
+    static constexpr ssize_t BLAS_GEMM_MIN_DIMENSION = 8;
+
+    void execute_at(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base);
+    bool try_execute_blas(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base);
+    bool try_dot(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+    bool try_gevm(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+    bool try_gemv(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+    bool try_gemm(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+    static std::optional<matrix_view_type> make_matrix_view(
+        value_type const * data,
+        ssize_t row_stride,
+        ssize_t column_stride,
+        ssize_t rows,
+        ssize_t columns);
     void execute_generic(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base);
 
     MatmulPlan const & m_plan;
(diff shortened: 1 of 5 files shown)


profiling/profile_matrix_ops.py
ThreeMonth03, line 1: In order to make the profiling result more stable, I align the repetitive times from 5 to 15 for GEMM.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -169,6 +174,22 @@ class MatmulExecutor
         Rhs,
     };
 
+    static constexpr ssize_t BLAS_DOT_MIN_LENGTH = 512;
+    static constexpr ssize_t BLAS_GEMV_MIN_ELEMENTS = 4096;
+    static constexpr ssize_t BLAS_GEMM_MIN_DIMENSION = 8;
```
ThreeMonth03, line 179: These thresholds affect route selection only. Workloads below a threshold still use the generic signed-stride executor, so tuning them does not change MatmulPlan or numerical behavior.

We might research the adjustment of thrshold after this issue close.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -412,19 +433,151 @@ void MatmulExecutor<Array>::execute()
...
 }
 
+template <typename Array>
+void MatmulExecutor<Array>::execute_at(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base)
+{
+    if (!try_execute_blas(output_base, lhs_base, rhs_base))
+    {
+        execute_generic(output_base, lhs_base, rhs_base);
+    }
+}
```
ThreeMonth03, line 456: Dispatch happens after the batch cursor resolves one output, lhs, and rhs base offset. GEMM therefore supports unbatched, equal-batch, and broadcast-batch execution here, while batched GEVM/GEMV remain generic in this PR.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -412,19 +433,151 @@ void MatmulExecutor<Array>::execute()
...
+
+    ssize_t lhs_stride = m_plan.lhs_inner_stride();
+    ssize_t rhs_stride = m_plan.rhs_inner_stride();
+    if (lhs_stride == -1 && rhs_stride == -1)
+    {
+        lhs_data += (m_plan.inner_size() - 1) * lhs_stride;
+        rhs_data += (m_plan.inner_size() - 1) * rhs_stride;
+        lhs_stride = -lhs_stride;
+        rhs_stride = -rhs_stride;
+    }
```
ThreeMonth03, line 501: Reversing both vectors preserves their element pairing. Rebasing both pointers converts (-1, -1) to (1, 1) without copying; mixed directions remain generic because the same transformation would change the pairing.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -412,19 +433,151 @@ void MatmulExecutor<Array>::execute()
...
+    if (column_stride == 1 && row_stride >= columns)
+    {
+        return BlasMatrixView<value_type>{data, row_stride, BlasTranspose::None};
+    }
+    if (row_stride == 1 && column_stride >= rows)
+    {
+        return BlasMatrixView<value_type>{data, column_stride, BlasTranspose::Transpose};
+    }
+    return std::nullopt;
+}
```
ThreeMonth03, line 579: This is the no-copy BLAS compatibility boundary. C-contiguous and row-padded cores use NoTrans, F-contiguous cores use a transposed storage descriptor, and unsupported negative or step-two core strides return nullopt.

cpp/solvcon/math/blas_compat.cpp
```diff
@@ -60,6 +60,16 @@ static CBLAS_TRANSPOSE to_cblas_transpose(bool transpose_matrix)
...
 
+static CBLAS_TRANSPOSE to_cblas_transpose(BlasTranspose transpose)
+{
+    return transpose == BlasTranspose::Transpose ? CblasTrans : CblasNoTrans;
+}
+
+static BlasTranspose compose_transposes(BlasTranspose storage, BlasTranspose requested)
+{
+    return storage == requested ? BlasTranspose::None : BlasTranspose::Transpose;
+}
```
ThreeMonth03, line 71: storage describes how the matrix is physically represented, while requested describes whether matmul needs matrix-vector or vector-matrix semantics. Equal flags cancel; different flags require a CBLAS transpose.

cpp/solvcon/math/blas_compat.cpp
```diff
@@ -305,6 +315,162 @@ void gemm_blas(ssize_t m,
...
+    blas_int_type const bldb = to_blas_int(rhs.m_leading_dimension, "ldb");
+    CBLAS_TRANSPOSE const opa = to_cblas_transpose(lhs.m_transpose);
+    CBLAS_TRANSPOSE const opb = to_cblas_transpose(rhs.m_transpose);
+    std::complex<double> const alpha{1.0, 0.0};
+    std::complex<double> const beta{0.0, 0.0};
+    auto const * lhs_data = as_std_complex_pointer(lhs.m_data);
+    auto const * rhs_data = as_std_complex_pointer(rhs.m_data);
+    auto * result_data = as_std_complex_pointer(result);
+    cblas_zgemm(CblasRowMajor, opa, opb, bm, bn, bk, &alpha, lhs_data, blda, rhs_data, bldb, &beta, result_data, bn);
+}
```
ThreeMonth03, line 470: blda and bldb are physical row pitches in elements, not logical shapes or batch strides. Reading them from the descriptors lets the same wrapper handle compact C, row-padded, and F-contiguous storage without copying.

tests/test_matrix.py
```diff
@@ -405,21 +417,24 @@ def test_batch_strides(self):
...
+                lhs = self.SimpleArray(array=case_lhs)
+                rhs = self.SimpleArray(array=case_rhs)
+                expected = np.matmul(case_lhs, case_rhs)
 
-            with self.subTest(lhs=lhs_case, rhs=rhs_case):
-                self.assert_matmul_planned(lhs, rhs, expected)
+                with self.subTest(
+                        shape=(m, k, n),
+                        lhs=lhs_case, rhs=rhs_case):
+                    self.assert_matmul_planned(lhs, rhs, expected)
```
ThreeMonth03, line 437: The same cross-broadcast topology is tested below and above the GEMM threshold. This exercises signed and zero batch mappings with both generic and direct-BLAS execution.

tests/test_matrix.py
```diff
@@ -364,21 +375,22 @@ def test_compare_with_numpy(self):
...
-        for (lhs_case, lhs_data), (rhs_case, rhs_data) in cases:
-            lhs = self.SimpleArray(array=lhs_data)
-            rhs = self.SimpleArray(array=rhs_data)
-            expected = np.matmul(lhs_data, rhs_data)
-
-            with self.subTest(lhs=lhs_case, rhs=rhs_case):
-                self.assert_matmul_planned(lhs, rhs, expected)
+                with self.subTest(
+                        shape=(m, k, n), lhs=lhs_case, rhs=rhs_case):
+                    self.assert_matmul_planned(lhs, rhs, expected)
```
ThreeMonth03, line 393: The two matrix sizes exercise both sides of the GEMM threshold. Their layout Cartesian products cover direct-BLAS candidates and generic fallback candidates against the same NumPy result.

ThreeMonth03 (commented):
@yungyuc Please take a look.

yungyuc:
> `matmul_planned()` uses generic signed-stride loops even when a contraction can run directly on vendor BLAS. This leaves optimized kernels unused for compatible products.
> 
> This change adds typed BLAS descriptors and direct routes for DOT, unbatched GEVM/GEMV, and GEMM. Direct GEMM requires both trailing matrix cores to be C-contiguous, F-contiguous, or row-padded; leading batch axes may still use signed or zero strides. Small or skinny products, batched vector products, and unsupported core strides remain generic. `MatmulPlan` stays independent of backend selection.

Please leave a comment stating what gets sped up, and update the summary in the PR description.  You list all the benchmark results and it is very difficult to follow without a clear and concise summary.

cpp/solvcon/math/blas_compat.cpp
yungyuc: It's really nice to have AI help these tedious maintenance.

ThreeMonth03:
> > `matmul_planned()` uses generic signed-stride loops even when a contraction can run directly on vendor BLAS. This leaves optimized kernels unused for compatible products.
> > This change adds typed BLAS descriptors and direct routes for DOT, unbatched GEVM/GEMV, and GEMM. Direct GEMM requires both trailing matrix cores to be C-contiguous, F-contiguous, or row-padded; leading batch axes may still use signed or zero strides. Small or skinny products, batched vector products, and unsupported core strides remain generic. `MatmulPlan` stays independent of backend selection.
> 
> Please leave a comment stating what gets sped up, and update the summary in the PR description. You list all the benchmark results and it is very difficult to follow without a clear and concise summary.

@yungyuc This pull request supports direct `matmul()` blas routes for big arrays without reordering array. Therefore, if `SimpleArray` are big and it's layout is compatible with blas, these cases could be accelerated through `blas` now.

I've updated the chart in PR description, the charts illustrate the improved cases, and the comparision among numpy, modification before and after pull request.

yungyuc:
> @yungyuc This pull request supports direct `matmul()` blas routes for big arrays without reordering array. Therefore, if `SimpleArray` are big and it's layout is compatible with blas, these cases could be accelerated through `blas` now.
> 
> I've updated the chart in PR description, the charts illustrate the improved cases, and the comparision among numpy, modification before and after pull request.

The new description looks clear.  Thank you.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -412,19 +434,160 @@ void MatmulExecutor<Array>::execute()
...
     {
-        execute_generic(0, 0, 0);
+        execute_at(0, 0, 0);
         return;
     }
 
     for (MappedOffsetCursor cursor = m_plan.batch_cursor(); cursor; cursor.advance())
     {
-        execute_generic(
+        execute_at(
```
yungyuc, line 443: I like the change to `execute_at()`.

yungyuc (approved):
LGTM

PR #1210, opened 2026-07-29
Drop the echo backend from the agent selector

tigercosmos:
The offline echo backend registered itself on import, so the pilot's Agent panel listed it first and opened on it. It answers by repeating the prompt and proposes no commands, which makes it a test double rather than something to offer a user.

This stops registering it. The class, its availability check, and its send() stay in place, and the tests that exercise the console pipeline now instantiate it explicitly. With it out of the registry the first entry is the Claude CLI backend, so that is what the selector defaults to.

Related to #966.

### DIFF
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -8,9 +8,9 @@ Pluggable AI backend abstraction for the Agent.
 A backend turns a prompt (plus context and a command tool surface) into a
 :class:`BackendResponse`: prose and a list of command dicts.
 Backends register in a process-wide registry so a caller can list the usable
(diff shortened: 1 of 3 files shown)


solvcon/agent/_backend.py
```diff
@@ -164,7 +166,4 @@ def available(self):
     def send(self, prompt, scene_context, tool_surface):
         return BackendResponse(text="echo: %s" % prompt, commands=[])
 
-
-register(EchoBackend())
```
tigercosmos, line 168: Don't registe echo backend in Pilot.

tigercosmos: Keep it for now. should remove it later when all other backend are done.

PR #1211, opened 2026-07-29
Painter Layers 1: give the Painter inspector its Layers page

tigercosmos:
The Layers tab in the Painter inspector was a greyed-out summary of what it would hold. This fills it with the list itself: one row per live shape on the active canvas, carrying the type icon the tool box already draws and a name made of the type and the shape id, since the world has no names of its own yet. Rows run newest first, the order the design lists and the order the canvas stacks them in, and a shape type the icon set does not draw keeps its row with an empty icon slot so the names stay in one column.

The row lands in its own module rather than as the page's private class, because the design draws it twice: this list, and the short list at the foot of the Design page that follows once the world can name a shape. The rules that color it travel with it, so the two lists cannot drift apart, and a page appends them to its own sheet rather than repeating them.

The page polls the bound canvas the way the Design page does, because the world changes in C++ without a change signal. What the poll compares is the row content itself, which is both what the page shows and the read the redraw needs, so the world is serialized once a tick rather than once to decide and again to draw. The rows scroll inside the page so a crowded world cannot stretch it past the dock, and surplus rows are hidden rather than destroyed.

This is the first of three stacked changes for step 4 of the Painter redesign plan. The measurements each row carries and the selection tint come next, then the search field, filter chips, and footer. Each was split out to keep every diff reviewable.

Tests: tests/test_pilot_painter_layers.py covers the rows against a stub canvas and against a real one bound by the Painter dock. make lint is clean and make pytest passes (1818 passed, 13 skipped, 3 xfailed).

Related to #1175

### DIFF
--- a/solvcon/pilot/canvas/_painter_layers.py
+++ b/solvcon/pilot/canvas/_painter_layers.py
@@ -1,0 +1,229 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+The Layers page of the Painter inspector: every object the world holds.
+"""
+
+import json
+
+from PySide6 import QtCore, QtGui, QtWidgets
+
+from . import _painter_icons
+from . import _painter_rows
+from ._painter_style import blend, PaletteStyled
+
+__all__ = [
+    'LayersPage',
+]
+
+
+class LayersPage(PaletteStyled):
+    """The inspector's Layers page: one row per object the world holds.
+
+    Like the Design page, the page reads the canvas it is bound to on a timer,
+    because the world changes in C++ without a change signal. What the poll
+    compares is the row content itself, which is both what the page shows and
+    the only fingerprint that cannot go stale as the rows grow what they say.
+
+    The rows are read-only: what a row shows comes from the world, and picking
+    an object stays the canvas's job until the list grows selection of its own.
+    """
+
+    # Poll period in milliseconds. Between the entity tree's half second and
+    # the Design page's 60ms: the whole world is serialized per tick, which is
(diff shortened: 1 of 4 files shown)


solvcon/pilot/canvas/_painter_layers.py
```diff
@@ -0,0 +1,225 @@
...
+from . import _painter_icons
+from . import _painter_rows
+from ._painter_style import blend, PaletteStyled
+
+__all__ = [
+    'LayersPage',
+]
+
+
+class LayersPage(PaletteStyled):
```
tigercosmos, line 21: Layer page.

solvcon/pilot/canvas/_painter_layers.py
```diff
@@ -0,0 +1,225 @@
...
+    def _objects(world):
+        """Every live shape the world holds, as the world describes it.
+
+        Newest first, which is the order the design lists: the world draws in
+        registry order, so the last shape added is the one on top of the
+        canvas and the one the list names first.
+        """
+        return json.loads(world.describe_state())["shapes"][::-1]
+
+    def _content(self):
```
tigercosmos, line 154: Get the layers (shapes).

solvcon/pilot/canvas/_painter_rows.py
```diff
@@ -0,0 +1,99 @@
...
+_MARGINS = (8, 0, 8, 0)
+_GAP = 8
+_NAME_PX = 12
+_RADIUS = 5
+
+# How far the type icon sits from the panel color.
+_MUTED_MIX = 0.35
+
+
+class ObjectRow(QtWidgets.QFrame):
```
tigercosmos, line 37: shared row component.

PR #1212, opened 2026-07-29
Painter Layers 2: measure and mark the selection

tigercosmos:
The Layers page listed the objects on the active canvas but said nothing about their size or which one the canvas had picked. Each row now carries the one measurement the design gives it, a circle's radius or the width and height of anything else, both read from the shape's own oriented box so a rotated shape reports the size it was drawn with rather than the wider span it covers along the axes.

The row the canvas has selected wears the accent tint and strokes its icon and metric in the accent. The sync is one way, canvas to list; picking from the list waits for the follow-up that gives it a selection of its own. A selection the world has since removed or undone away reads as no selection, because the canvas hands back the stored id without checking it and every query on a dead one throws.

Marking the row means polishing the metric with it: the rule that colors the metric reads the selected property off the row, and Qt resolves each widget's style rules against its own cache, so a label left unpolished keeps the color it had under the previous selection. A test covers it, and it fails without the fix.

Measuring the shape rather than its serialized box is also what keeps the poll honest. The world serializes an axis-aligned box, and a shape lying along the axes covers the very box of one of the other proportions turned onto them, so a fingerprint taken from those boxes reads the same for two canvases the page must tell apart.

The metric and the tint land on the shared row module, so the Design page's short list inherits them when it arrives. The font and oriented-box helpers move from the Design page into the shared style module for the same reason.

Second of three stacked changes for step 4 of the Painter redesign plan, on top of #1211. The search field, filter chips, and footer follow.

Tests: make lint is clean and make pytest passes (1825 passed, 13 skipped, 3 xfailed).

Related to #1175

### DIFF
--- a/solvcon/pilot/canvas/_painter_rows.py
+++ b/solvcon/pilot/canvas/_painter_rows.py
@@ -11,7 +11,7 @@ live together here rather than in either page.
 
 from PySide6 import QtGui, QtWidgets
 
-from ._painter_style import blend
+from ._painter_style import blend, mono_font
 
 __all__ = [
     'ICON_PX',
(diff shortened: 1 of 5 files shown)


solvcon/pilot/canvas/_painter_style.py
```diff
@@ -32,6 +37,27 @@ def rule(shape):
     return line
 
 
+def mono_font():
+    """The stand-in for the mono font the design gives every number."""
+    return QtGui.QFontDatabase.systemFont(QtGui.QFontDatabase.FixedFont)
+
+
+def obb_metrics(obb):
```
tigercosmos, line 45: Get tyhe metrics.

solvcon/pilot/canvas/_painter_rows.py
```diff
@@ -45,54 +53,99 @@ def __init__(self, parent=None):
         self._icon.setFixedWidth(ICON_PX)
         self._name = QtWidgets.QLabel(self)
         self._name.setObjectName("name")
+        self._metric = QtWidgets.QLabel(self)
+        self._metric.setObjectName("metric")
+        self._metric.setFont(mono_font())
```
tigercosmos, line 58: Add the metric for each shape row.

PR #1213, opened 2026-07-29
Painter Layers 3: give the page its search, filters, and footer

tigercosmos:
The Layers page listed and measured the objects on the active canvas, but stood alone above an empty page. This adds the rest of the design's Layers tab: a search field over the All/Shapes/Guides chips, and a footer that counts the objects the world holds between the add and remove buttons.

Everything but the count is greyed out and names what it waits for. The search field and the chips need object names and guides, neither of which the world has; the buttons need editing from the list, which waits on the selection the list does not own yet. They take their designed places now so the page keeps its shape and the follow-ups only fill them in.

The footer buttons carry their icon in the disabled state alone, wearing the greyed color the page mixed rather than whatever fade the style would otherwise apply. The search glass, plus, and minus join the icon set the tool box already draws from, and the rules that color the search field name it rather than every line edit the page may come to hold, so a later editor does not inherit a borderless field by accident.

Last of the three stacked changes for step 4 of the Painter redesign plan, after #1211 and #1212. With this the page matches the design's Layers tab; picking a shape from the list, object names, guides, and visibility remain follow-ups.

Tests: make lint is clean and make pytest passes (1826 passed, 13 skipped, 3 xfailed).

Related to #1175

Screenshot 2026-07-30 at 1 50 50 AM

### DIFF
--- a/solvcon/pilot/canvas/_painter_layers.py
+++ b/solvcon/pilot/canvas/_painter_layers.py
@@ -11,11 +11,127 @@ from PySide6 import QtCore, QtGui, QtWidgets
 
 from . import _painter_icons
 from . import _painter_rows
-from ._painter_style import blend, obb_metrics, PaletteStyled
+from ._painter_style import (blend, mono_font, obb_metrics, rule, shade,
+                             PaletteStyled)
 
 __all__ = [
     'LayersPage',
 ]
+
+
+class _Filters(QtWidgets.QWidget):
+    """The greyed-out search field over the filter chips.
+
+    Both wait on the same follow-ups, object names and guides, and the design
+    draws them as one block above the list, so they are built together.
+    """
+
+    _SEARCH_MARGINS = (10, 10, 10, 8)
+    _SEARCH_HEIGHT = 28
+    _SEARCH_PADDING = (8, 0, 8, 0)
+    _SEARCH_GAP = 8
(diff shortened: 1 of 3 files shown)


solvcon/pilot/canvas/_painter_layers.py
```diff
@@ -11,13 +11,129 @@
...
-from ._painter_style import blend, obb_metrics, PaletteStyled
+from ._painter_style import (blend, mono_font, obb_metrics, rule, shade,
+                             PaletteStyled)
 
 __all__ = [
     'LayersPage',
 ]
 
 
+class _Filters(QtWidgets.QWidget):
```
tigercosmos, line 22: Add a placeholder filter component

solvcon/pilot/canvas/_painter_layers.py
```diff
@@ -11,13 +11,129 @@
...
+            chip.setCheckable(True)
+            chip.setChecked(0 == index)
+            chip.setToolTip(f"{name}  (needs guides and object names)")
+            layout.addWidget(chip)
+            self.chips[name] = chip
+        layout.addStretch(1)
+        return layout
+
+
+class _Footer(QtWidgets.QWidget):
```
tigercosmos, line 90: Add placeholder footer component.

PR #1217, opened 2026-07-30
 Build and test through the presets, and drive the Windows CI jobs from them (cmake preset steps 2, 3)

### LINKED ISSUE #1215: Make `CMakePresets.json` the single source configuration
I got a new machine to set up and found that it is a pain to configure IDE, because solvcon hands CMake its build arguments from four places that nothing keeps in agreement: (1) the root `Makefile` expands twelve `-D` flags, (2) `CMakePresets.json` sets a handful of its own, (3) `build.ps1` adds more, and (4) `.github/workflows/devbuild.yml` spells out two `cmake -GNinja -D...` blocks of its own. In addition, `HIDE_SYMBOL` and `LINT_AS_ERRORS` are on in the `Makefile` and absent from the presets, so a preset build and a make build are not the same build.

The existing preset file `CMakePresets.json` was a by-product when making `build.ps1` on Windows. No preset is usable from an IDE at all: `build.ps1` and the `Makefile` inject the dependency-prefix paths on the command line, while an IDE runs `cmake --preset <name>` with nothing added, so configure fails at `find_package(pybind11 ...)`.

It's time to take advantage of it to organize the build configuration. Using IDE with git worktree will be made easier.

## Goal

Make `CMakePresets.json` the one place the build configuration is written down. The `Makefile`, `build.ps1`, and the workflows become thin drivers over it, and a contributor who opens the repository in VS Code or CLion gets a working build by selecting a preset.

- Cap the schema at version 10. Versions 11 and 12 raise the CMake floor above the project's own 4.0.1, and CLion reads no further than 10.
- CLion is the binding consumer. It reads configure and build presets only, and it supports only `equals` and `notEquals` in a `condition`, so nothing may depend on `inList`, `matches`, or `anyOf`.
- Filter host-specific presets twice, with a CMake `condition` on `${hostSystemName}` and with a `hostOS` entry in the Microsoft vendor map. The command line and CLion honour the first, VS Code the second, and neither alone hides a preset everywhere.
- Machine-specific values stay out of the checked-in file. Dependency-prefix paths belong in `CMakeUserPresets.json`, and values a CI runner owns reach the presets through `$env{}`.

## Steps

- [x] Step 1: Write the build configuration into the presets
  - PR #1216
- [x] Step 2: Build and test through the presets
  - PR #1217
- [x] Step 3: Drive the Windows CI jobs from workflow presets
  - PR #1217
- [x] Step 4: Document the presets and the user-preset seam
  - PR #1218

### Step 1: Write the build configuration into the presets

Raise `CMakePresets.json` to schema version 10 with a `$schema` key and typed cache variables. Add a hidden `base` preset carrying the knobs that hold on every host, at the values the `Makefile` passes, plus `PYTHONPATH` in its `environment` map so a target launched from an IDE finds the in-tree package. Add the visible leaves `dev-rel`, `dev-dbg`, `dev-noqt`, `win-rel`, and `win-dbg`, and `ci-` presets describing what the Windows CI jobs configure. Give the build presets their target lists, the test presets a label filter, and add workflow presets chaining the three.

Declare every knob a driver can pass, including `USE_SANITIZER`, or a preset rerun cannot reset it. Set the build tree to `${sourceDir}/build/${presetName}`, one per preset, which is what both IDEs assume.

This step changes no behaviour: nothing reads the new presets yet.

*Verify:* `cmake --list-presets` lists the developer presets on Linux and macOS and the MSVC presets only on Windows.

### Step 2: Build and test through the presets

Have the `Makefile` configure with `cmake --preset` and add only what a preset cannot carry: `-B` for the ABI-tagged build directory, `-G` for the generator a make build has always used, and a `-D` flag for each knob actually set on the command line, in the environment, or in `setup.mk`. Every target name and every `make VAR=value` override must keep working, and the lint targets must not be rewired, so they still run in a checkout that has never been configured.

Register the two Python suites with `add_test()` and move `enable_testing()` to the top level, so `ctest` covers the C++ cases, the Python suite, and the pilot suite. Give each suite its environment through a test property. Add a target that removes the built extension, so `make clean` stops recomputing the extension suffix with a second `python3-config` call.

Document `CMakeUserPresets.json` as the seam for the three machine paths, gitignore it, and ship a template per host family under `contrib/cmake`. A build preset inherits its configure preset's `condition`, so a single template would be refused on the other host. Give `build.ps1` a switch to build such a preset, have it read `binaryDir` out of the preset files instead of recomputing it, and let it pick among build presets rather than assembling a target list.

*Verify:* the documented make surface, including `PYTEST_OPTS` forwarding and `make flake8` without a configure; `ctest --preset <name>` runs every suite; a fresh checkout plus a user preset configures with a bare `cmake --preset`.

### Step 3: Drive the Windows CI jobs from workflow presets

Replace the two hand-written `cmake` blocks in `devbuild.yml` with one workflow preset per job. Export the pybind11 package directory and the MKL paths so the `ci-` presets pick them up through `$env{}`. Keep the coverage these jobs have today by filtering the test stage on the C++ and pilot labels.

*Verify:* a full CI run, compared against the previous run's configure log for cache-variable equality.

### Step 4: Document the presets and the user-preset seam

Add a developer guide page listing the presets each host offers, the build and test presets that go with them, and the IDE behavior that follows: what VS Code ignores once preset mode is on, that CLion reads configure and build presets only, and that switching interpreters against a preset tree needs `--fresh`. Walk through `CMakeUserPresets.json` with a worked example. Spell out that the C++ cases are enumerated by running the test binary, so it has to be built before `ctest` can see them. Add the `ctest` invocations to the testing guide and note in the build guide that the `Makefile` now configures through a preset.

*Verify:* the documentation builds with no reference warnings, and a contributor can follow the page on a clean clone with no IDE settings.

## Follow-ups in mind

Other cmake files need to be linted and reorganized for more flexibility and readability.  But it is outside the scope of this issue.

Test coverage and harness can be improved. Coherent interoperation between ctest and pytest will be a goal.

The current cmake setup generates warnings that should be clarified and fixed.

yungyuc:
Steps 2, 3 of issue #1215.

After the PR, the `Makefile` configures through `cmake --preset`, `build.ps1` learns to build a preset the developer wrote, `ctest` becomes the one test entry point on every platform, and the two Windows CI jobs each name a single workflow preset instead of spelling out their own `cmake -GNinja -D...` block. The build a make user gets and the build a preset user gets are now the same build, because there is one place where it is described.

### Step 2: Build and test through the presets

The `Makefile` configures with `cmake --preset` and adds only what a preset cannot carry: `-B` for the ABI-tagged build directory, `-G` for the generator a make build has always used, and a `-D` flag for each knob that is actually set on the command line, in the environment, or in `setup.mk`. `CMAKE_KNOBS` lists those knobs and `CMAKE_OVERRIDES` turns the ones with a non-default `origin` into flags. `CMAKE_BUILD_TYPE` picks `dev-rel` or `dev-dbg`, and `CMAKE_PRESET` overrides that choice, for instance to build a preset of your own. Every target name and every `make VAR=value` override keeps working. The lint targets are deliberately not rewired, so they still run in a checkout that has never been configured.

`ctest` becomes the one test entry point. `gtest_discover_tests` already registered the C++ cases and nothing used them; the new `tests/CMakeLists.txt` joins both Python suites through `add_test()`, as `pytest` and `pilot_pytest`, each carrying its own environment in a test property and labeled `python` or `pilot`. The C++ cases pick up the `cpp` label. `enable_testing()` moves to the top level so it no longer depends on `USE_GOOGLETEST`. `SOLVCON_TEST_SHOW_WINDOWS` stays an environment variable, because a test property overrides rather than defaults, and a developer and a runner want opposite answers.

The extension suffix is already computed in `cpp/binary/pymod_solvcon/CMakeLists.txt`, so a `remove_solvcon_py` target there replaces the `Makefile`'s second shell-out to `python3-config` in `clean` and `cmakeclean`.

The three cache variables that name one machine's directories, `PYTHON_EXECUTABLE`, `pybind11_path`, and `CMAKE_PREFIX_PATH`, belong in `CMakeUserPresets.json`, which CMake reads automatically and both IDEs pick up. It is gitignored and a template ships under `contrib/cmake`, one per host family, each named so that an editor recognizes it as JSON and applies the `$schema` it declares. A build preset inherits its configure preset's `condition`, so a single template whose build presets inherit `dev-rel*` is refused on Windows as a disabled build preset; the two templates keep every `inherits` line a matched set. This is what makes an IDE build work at all: an IDE runs `cmake --preset` with nothing added, and configure used to fail at `find_package(pybind11)`.

`build.ps1` gains `-Preset` to build such a preset, in which case it leaves the dependency-prefix paths to the preset. It reads `binaryDir` out of the preset files (`Get-ConfigurePresetBinaryDir`, resolving `inherits`) instead of recomputing `build\<preset>`, and picks among the `<preset>`, `<preset>-module`, and `<preset>-gtest` build presets instead of assembling a target list in PowerShell.

`LINT_AS_ERRORS` and `BLA_VENDOR` are reported with `message()` where they are decided. Only the clang-tidy branches read the first and only the OpenBLAS branch reads the second, so a preset that sets either for the other case drew an unused cache variable warning on every configure.

### Step 3: Drive the Windows CI jobs from workflow presets

`devbuild.yml` spelled out its own `cmake -GNinja -D...` block twice, once for the Windows build job and once for the Windows sanitizer job, and neither drove the `Makefile` or the presets. Nothing kept those blocks in agreement with the preset file.

Each job now names one workflow preset, `ci-win-rel`, `ci-win-dbg`, or `ci-win-asan`, which chains configure, build, and the CTest run. The values a runner owns stay out of the checked-in file: the pybind11 package directory and the MKL include and library paths reach the `ci-` presets through `$env{}`, which the jobs export. The portable-package step rebuilds the pilot with `cmake --build --preset ci-win-rel --target pilot` and reads `pilot.exe` from the runtime output directory the preset names.

The test stage filters on the `cpp` and `pilot` labels, the coverage these jobs have always had; the plain Python suite runs on Linux and macOS. Where the jobs used to build the `run_gtest` and `run_pilot_pytest` targets they now run `ctest`, so the C++ cases report individually.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -403,43 +403,21 @@ jobs:
           Get-Command flake8
           echo "::endgroup::"
 
-      - name: cmake ALL_BUILD
+      - name: cmake workflow
         run: |
-          echo "::group::cmake ALL_BUILD"
-          # Use the Ninja generator: CMAKE_<LANG>_COMPILER_LAUNCHER (the
-          # sccache hook) is honored only by Ninja/Makefile generators, not
-          # the Visual Studio (MSBuild) generator, which silently ignores it.
-          # No vcpkg toolchain: SOLVCON_USE_MKL makes solvcon link mkl_rt for
-          # both CBLAS and LAPACK, found via the CMAKE_*_PATH dirs above.
-          cmake -GNinja `
-            -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
-            -DCMAKE_C_COMPILER_LAUNCHER=sccache `
-            -DCMAKE_CXX_COMPILER_LAUNCHER=sccache `
-            -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded `
-            -DSOLVCON_USE_MKL=ON `
-            -DCMAKE_INCLUDE_PATH="$env:MKL_INCLUDE_DIR" `
-            -DCMAKE_LIBRARY_PATH="$env:MKL_LIBRARY_DIR" `
-            -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
-            -S${{ github.workspace }} `
-            -B${{ github.workspace }}/build
-          cmake --build ${{ github.workspace }}/build
-          echo "::endgroup::"
-
-      - name: cmake run_gtest
-        run: |
-          echo "::group::cmake run_gtest"
-          cmake --build ${{ github.workspace }}/build `
-            --target run_gtest
-          echo "::endgroup::"
-
-      - name: cmake run_pilot_pytest
-        run: |
-          echo "::group::cmake run_pilot_pytest"
+          echo "::group::cmake workflow"
+          # The ci-win-* presets in CMakePresets.json state the whole
+          # configure.
+          $env:PYBIND11_DIR = "$(pybind11-config --cmakedir)"
           # A runner has no desktop to take over, so let this job map its
           # windows, the way it has always run them.
           $env:SOLVCON_TEST_SHOW_WINDOWS = "ON"
-          cmake --build ${{ github.workspace }}/build `
-            --target run_pilot_pytest
+          $preset = if ('${{ matrix.cmake_build_type }}' -eq 'Debug') {
+            'ci-win-dbg'
+          } else {
+            'ci-win-rel'
+          }
+          cmake --workflow --preset $preset
           echo "::endgroup::"
 
       - name: generate portable
@@ -483,14 +461,14 @@ jobs:
           # Remove redundant Qt DLLs from PySide6
           Remove-Item -Path $pyside6_path\Qt*.dll
 
-          # Configure and build the pilot. The build dir is already
-          # configured with the Ninja single-config generator above, so reuse
-          # it without --config and read pilot.exe from the non-config path.
-          cmake -Dpybind11_DIR="$(pybind11-config --cmakedir)" -S . -B build
-          cmake --build build --target pilot
+          # The workflow preset above already configured and built this tree
+          # with the Ninja single-config generator, so rebuild the pilot in
+          # place and read pilot.exe from the runtime output directory the
+          # preset names.
+          cmake --build --preset ci-win-rel --target pilot
 
           # Deploy the Qt environment for the pilot executable and copy necessary files
-          Copy-Item -Path ".\build\cpp\binary\pilot\pilot.exe" -Destination $destination
+          Copy-Item -Path ".\build\ci-win-rel\pilot.exe" -Destination $destination
           windeployqt --release "$destination\pilot.exe"
           Copy-Item -Path ".\solvcon" -Destination $destination -Recurse
           # Also include potential thirdparty
@@ -674,39 +652,19 @@ jobs:
         echo "$pyside6_path;$shiboken6_path" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
         echo "::endgroup::"
 
-    - name: cmake ALL_BUILD USE_SANITIZER=ON
+    - name: cmake workflow USE_SANITIZER=ON
       run: |
-        echo "::group::cmake ALL_BUILD USE_SANITIZER=ON"
-        # Instrument both the gtest binary and the Qt pilot: BUILD_QT=ON and
-        # USE_GOOGLETEST=ON build both, USE_SANITIZER=ON adds /fsanitize=address.
+        echo "::group::cmake workflow USE_SANITIZER=ON"
+        # The ci-win-asan preset instruments both the gtest binary and the Qt
+        # pilot: BUILD_QT and USE_GOOGLETEST build both, USE_SANITIZER adds
+        # /fsanitize=address, and the vcpkg toolchain supplies BLAS and LAPACK.
         # The MSVC toolset bin dir that setup-msvc-dev puts on PATH also carries
         # the ASan runtime DLL the instrumented binaries load at start.
-        cmake -GNinja `
-          -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} `
-          -DCMAKE_TOOLCHAIN_FILE="$env:VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake" `
-          -DVCPKG_TARGET_TRIPLET="x64-windows" `
-          -Dpybind11_DIR="$(pybind11-config --cmakedir)" `
-          -DBUILD_QT=ON `
-          -DUSE_GOOGLETEST=ON `
-          -DUSE_SANITIZER=ON `
-          -S${{ github.workspace }} `
-          -B${{ github.workspace }}/build
-        cmake --build ${{ github.workspace }}/build
-        echo "::endgroup::"
-
-    - name: cmake run_gtest
-      run: |
-        echo "::group::cmake run_gtest"
-        cmake --build ${{ github.workspace }}/build --target run_gtest
-        echo "::endgroup::"
-
-    - name: cmake run_pilot_pytest
-      run: |
-        echo "::group::cmake run_pilot_pytest"
+        $env:PYBIND11_DIR = "$(pybind11-config --cmakedir)"
         # A runner has no desktop to take over, so let this job map its
         # windows, the way it has always run them.
         $env:SOLVCON_TEST_SHOW_WINDOWS = "ON"
-        cmake --build ${{ github.workspace }}/build --target run_pilot_pytest
+        cmake --workflow --preset ci-win-asan
         echo "::endgroup::"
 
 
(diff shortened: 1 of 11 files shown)


.github/workflows/devbuild.yml
```diff
@@ -403,43 +403,21 @@ jobs:
           Get-Command flake8
           echo "::endgroup::"
 
-      - name: cmake ALL_BUILD
+      - name: cmake workflow
```
yungyuc, line 406: Use the preset workflow to build on Windows.

contrib/cmake/CMakeUserPresets.example.json
```diff
@@ -0,0 +1,41 @@
...
+      "inherits": "dev-rel",
+      "displayName": "Local dependency prefix",
+      "description": "A checked-in preset plus the paths of a local dependency prefix.",
+      "cacheVariables": {
+        "PYTHON_EXECUTABLE": {
+          "type": "FILEPATH",
+          "value": "/path/to/prefix/bin/python3"
+        },
+        "pybind11_path": "/path/to/prefix/lib/python3.14/site-packages/pybind11/share/cmake/pybind11",
+        "CMAKE_PREFIX_PATH": "/path/to/prefix"
```
yungyuc, line 17: When scdv is used, `CMAKE_PREFIX_PATH` should be set to `${SCDV_BASE}/usr`.

cpp/binary/pymod_solvcon/CMakeLists.txt
```diff
@@ -52,12 +52,19 @@ cmake_print_variables(PYTHON_VENV_BIN)
...
+    OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
 add_custom_target(_solvcon_py
     COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:_solvcon> ${SOLVCON_PY_DIR}/../_solvcon${PYEXTSUFFIX}
     DEPENDS _solvcon)
 
+add_custom_target(remove_solvcon_py
+    COMMAND ${CMAKE_COMMAND} -E rm -f
+        ${SOLVCON_PY_DIR}/_solvcon${PYEXTSUFFIX}
```
yungyuc, line 64: CMake knows the Python extension suffix.

cpp/solvcon/CMakeLists.txt
```diff
@@ -212,6 +212,10 @@ if (NOT APPLE)
         "Use pre-built Intel oneAPI MKL for BLAS/LAPACK instead of OpenBLAS"
         "$ENV{SOLVCON_USE_MKL}")
 
+    # The find_package(LAPACK) below reads it, and MKL skips that branch, so
```
yungyuc, line 215: Report `BLA_VENDOR` here.

tests/CMakeLists.txt
yungyuc, line 1: Allow `ctest` harness Python tests.

Makefile
```diff
@@ -22,25 +22,37 @@ BUILD_PATH_EXT ?=
...
 RUNENV += PYTHONPATH=$(SOLVCON_ROOT)
 
-SKIP_PYTHON_EXECUTABLE ?= OFF
-HIDE_SYMBOL ?= ON
-DEBUG_SYMBOL ?= ON
-SOLVCON_PROFILE ?= OFF
-BUILD_METAL ?= OFF
-BUILD_QT ?= ON
-USE_CLANG_TIDY ?= OFF
+# The configure knobs are written down once, in CMakePresets.json; see
```
yungyuc, line 25: Filter CMake knobs.

Makefile
```diff
@@ -98,30 +108,16 @@ cmake: $(BUILD_PATH)/Makefile
...
-	env $(RUNENV) $(CMAKE_CMD)
-
-$(BUILD_PATH)_xcode/Makefile: CMakeLists.txt Makefile
-	mkdir -p $(BUILD_PATH)_xcode ; \
-	cd $(BUILD_PATH)_xcode ; \
-	env $(RUNENV) $(CMAKE_CMD) -G Xcode
+# The preset carries the configure; -B and -G layer on top of it. The build
+# tree keeps its ABI tag, which a preset cannot compute, and the generator
+# stays the one the clean target and the workflows expect from a make build.
+CMAKE_CMD = cmake --preset $(CMAKE_PRESET) $(CMAKE_OVERRIDES) $(CMAKE_ARGS)
```
yungyuc, line 114: Use CMake presets.

Makefile
```diff
@@ -314,12 +310,12 @@ format: pyformat
 
 .PHONY: clean
 clean:
-	rm -f $(SOLVCON_ROOT)/solvcon/_solvcon$(pyextsuffix)
-	rm -f $(SOLVCON_ROOT)/_solvcon$(pyextsuffix)
+	cmake --build $(BUILD_PATH) --target remove_solvcon_py
```
yungyuc, line 313: Use CMake to clean up extension modules.

yungyuc:
@ExplorerRay Please help take a look.

PR #1218, opened 2026-07-30
Document the CMake presets and the user-preset seam (cmake preset step 4)

### LINKED ISSUE #1215: Make `CMakePresets.json` the single source configuration

yungyuc:
Step 4 of issue #1215.

The presets are only half the story for a new contributor. The three cache variables that name one machine's directories cannot be checked in, so whether an IDE build works at all rests on documentation rather than on a file in the repository.

Add `doc/source/devguide/cmake.md`, which lists the presets each host offers, the build and test presets that go with them, and the IDE behaviour that follows: what VS Code ignores once preset mode is on, that CLion reads configure and build presets only, and that switching interpreters against a preset tree needs `--fresh`. It walks through `CMakeUserPresets.json` with a worked example, one per host family, and says that the `inherits` lines are a matched set, since a build preset inherited from another host is refused as disabled.

Both guides spell out the `<preset>-gtest` build. A plain `<preset>` build does not build `test_nopython`, and `gtest_discover_tests` enumerates the C++ cases by running that binary, so `ctest --preset <preset>` straight after a plain build fails on the `test_nopython_NOT_BUILT` placeholder.

The testing guide gains the `ctest` invocations alongside the make targets, and the build guide says that the `Makefile` configures through a preset and that `CMAKE_PRESET` selects another one.

### DIFF
--- a/doc/source/devguide/cmake.md
+++ b/doc/source/devguide/cmake.md
@@ -1,0 +1,190 @@
+# CMake Presets
+
+`CMakePresets.json` at the repository root holds solvcon's build
+configuration: the generator, the build type, the cache variables, and the run
+environment.  It is checked in, so every contributor and every continuous
+integration job configures the same way, and both supported IDEs read it
+directly.
+
+List what the file offers on this host, then configure and build:
+
+```bash
+cmake --list-presets
+cmake --preset <name>
+cmake --build --preset <name>
+```
+
+A preset that names a toolchain the current host cannot run carries a
+`condition`, so it does not appear in the listing there.  On Linux and macOS
+the make targets described in {doc}`/start/build_solvcon` remain the entry
+point every contributor uses, and they configure through these presets: the
+Makefile adds the ABI-tagged build directory, the generator, and whatever
+`make VAR=value` sets, and nothing else.
+
+| Configure preset | Host          | What it configures                    |
+|:-----------------|:--------------|:--------------------------------------|
+| `dev-rel`        | Linux, macOS  | optimized module and pilot            |
+| `dev-dbg`        | Linux, macOS  | debuggable module and pilot           |
+| `dev-noqt`       | Linux, macOS  | optimized module, no pilot            |
+| `win-rel`        | Windows       | optimized module and pilot, MSVC      |
+| `win-dbg`        | Windows       | debuggable module and pilot, MSVC     |
+
+Each configure preset comes with three build presets, one per thing that is
+actually built:
+
+| Build preset          | Targets                            |
+|:----------------------|:-----------------------------------|
+| `<preset>`            | the extension module and the pilot |
+| `<preset>-module`     | the extension module alone         |
+| `<preset>-gtest`      | the C++ gtest binary               |
+
+`cmake --build --preset <name>` therefore needs no `--target` argument.  The
+`dev-noqt` preset has no pilot to build, so it has only the plain and `-gtest`
+entries.
+
+The build tree is `build/<configure preset>`, named through the
+`${presetName}` macro so that a preset inheriting another gets its own tree
+rather than writing into its parent's.  It is one tree per preset, which is
+what both IDEs assume, and it is not the Makefile's `build/rel<pyvminor>`, so
+a preset build and a make build never share a cache.  They do share where the
+extension module lands, `solvcon/` and the repository root, so whichever built
+last is the one Python imports.
+
+Test presets run the suites CTest has registered.  There is one per configure
+preset, plus label-filtered ones on `dev-rel`:
+
+```sh
+ctest --preset dev-rel            # every suite
+ctest --preset dev-rel-cpp        # the C++ cases alone
+ctest --preset dev-rel-python     # the Python suite alone
+ctest --preset dev-rel-pilot      # the Python suite inside the pilot binary
+```
+
+The C++ cases come from `gtest_discover_tests`, which enumerates them by
+running the built binary, and the plain build preset does not build it.  Build
+`<preset>-gtest` before a run that includes them, or CTest reports the single
+placeholder case `test_nopython_NOT_BUILT` and fails:
+
+```sh
+cmake --preset dev-rel
+cmake --build --preset dev-rel
+cmake --build --preset dev-rel-gtest
+ctest --preset dev-rel
+```
+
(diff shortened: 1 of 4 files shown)


doc/source/devguide/cmake.md
yungyuc, line 1: Make it clear how to use the presets.

doc/source/devguide/testing.md
```diff
@@ -20,6 +20,34 @@ After `make gtest` has built the binary, a single C++ test can be run directly:
 
 where `<pyvminor>` is the active Python major and minor version, e.g. `314`.
 
+## Through CTest
```
yungyuc, line 23: Alternate way to run tests.

PR #1221, opened 2026-07-30
Move the Painter into a package of its own

tigercosmos:
The Painter has grown to six modules and a bit over two thousand lines, all sitting in the canvas package next to the canvas window and the SVG dialog, and all carrying a `_painter_` prefix to say which half of the package they belong to. They move to `solvcon.pilot.painter`, a peer of the other feature packages, where the prefix is the package name and the pages, the object row, the icons, and the palette-derived styling sit together.

Nothing changes but where the code lives and how it is reached. The new package exports `Painter` and `PainterPanel`; the canvas package keeps the 2D surface alone and no longer re-exports `Painter`; and the controller reaches the feature through the package rather than through its module, because a `from ..painter import _gui` would shadow the controller's own module name at every use site. The tests follow the same rule. `setup.py` gains the new package so an installed build ships it.

The diff is a pure move: git detects all six renames, and the only edited lines are imports, the two package docstrings, and the package list.

Related to #1175.


solvcon/pilot/painter/__init__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The Painter: the draw tool selector and the inspector that edits what a 2D
```
yungyuc, line 6: So a painter controls and draws on a canvas.  Let me know if I am wrong.

tigercosmos: Yes

PR #1223, opened 2026-07-30
Painter Canvas 2: fit, size, and center the view

tigercosmos:
The Painter's Canvas page gains the three view buttons the design draws under the zoom readout. Fit frames everything the world draws, shapes and bare segments, curves and points alike, and greys out when there is nothing to frame. 100% zooms to exactly one pixel per world unit about the middle the view already shows, and Center pans the content, or the origin, into that middle without touching the zoom.

Centering needs the size of the space the view maps into, which the wrapper did not expose, so `R2DWidget` hands its drawing area over as `viewportSize`. The canvas clamps the zoom it is given, so the pan is figured from the transform it hands back, and a center too far out for that pan to stay finite leaves the view alone rather than half written.

Fit only needs to know whether the world holds anything, so the poll compares counts of what it holds rather than the serialized state, which costs 4.6 ms against 0.0003 ms at a thousand shapes and would have slowed the zoom readout. Counts do not move when a shape is dragged, so Fit measures the geometry when it is pressed.

At 556 changed lines this is over the 500-line guideline, about 300 of them tests.

Related to #1175.

### DIFF
--- a/solvcon/pilot/painter/_canvas.py
+++ b/solvcon/pilot/painter/_canvas.py
@@ -5,10 +5,14 @@
 The Canvas page of the Painter inspector: how the canvas shows the world.
 """
 
+import json
+import math
+
 from PySide6 import QtCore, QtGui, QtWidgets
 
 from ._sections import Placeholder, Section
 from ._style import blend, mono_font, rule, shade, PaletteStyled
+from ..panel._tree_panel import EntityTreeWidget
 
 __all__ = [
     'CanvasPage',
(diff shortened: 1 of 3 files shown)


solvcon/pilot/painter/_canvas.py
```diff
@@ -155,44 +176,165 @@ def _build(self):
...
+        self._view.body.addLayout(self._build_actions())
         return self._view
 
+    def _build_actions(self):
+        """Build the view buttons, in design order."""
+        layout = QtWidgets.QHBoxLayout()
+        layout.setSpacing(self._ACTION_GAP)
+        for name, handler in (("Fit", self._on_fit),
+                              ("100%", self._on_actual),
+                              ("Center", self._on_center)):
```
tigercosmos, line 192: New features for the canvas page.

solvcon/pilot/painter/_canvas.py
```diff
@@ -155,44 +176,165 @@ def _build(self):
...
+        return counts is not None and any(counts)
+
+    @staticmethod
+    def _content_bounds(widget):
+        """The extent of what ``widget`` draws, or ``None`` for nothing."""
+        world = widget.world
+        if world is None:
+            return None
+        return EntityTreeWidget.world_bounds(
+            json.loads(world.describe_state()))
```
tigercosmos, line 245: How we get the bounded box of the world.

Issue #1222, opened 2026-07-30
Cut the pilot unittest time on CI


tigercosmos:
The devbuild workflow runs the Python suite three times per OS: `make pytest BUILD_QT=OFF`, `make pytest BUILD_QT=ON`, and `make run_pilot_pytest`, which runs the whole `tests/` tree once more inside the pilot binary. On the latest master run (warm ccache) the three steps cost 7 s, 98 s, and 73 s on ubuntu-24.04, and 24 s, 138 s, and 101 s on macos-26. The suite itself is not slow: the 1848 tests minus the pilot widget tests finish in about 3 seconds. Two files dominate everything else. `test_pilot_theme.py` and `test_pilot_terminal.py` together take 77 s of the 96 s `BUILD_QT=ON` test span on ubuntu (91 s of 124 s on macos), then run again inside the pilot binary for another 51 s (65 s on macos).

The cost concentrates in `set_theme`, which takes 0.5 to 1 second per call on the CI runners. `RThemeManager::setMode` writes QSettings, flips `QStyleHints::setColorScheme`, and then `apply()` installs an application palette and an application-wide stylesheet, and the stylesheet change forces Qt to unpolish and repolish every widget in the process. `TerminalWidgetTC.setUp` pins the theme and restores it in cleanup, so each of the 21 terminal tests pays two full switches before doing any work, and the 27 theme tests each pay one to three.

On top of that, every pilot GUI test runs twice, once under python with PySide6 and once inside the pilot binary, although the code under test is the same C++ compiled into the same libraries. The PR lane should not pay for the full cross product on every push when a nightly lane already exists to carry the exhaustive coverage.

The build half of the job is broken in its own way. On a pull request touching only Python files (run 30532183716), the ubuntu ccache step logged "No cache found" and the post-job stats showed 0 hits out of 231 cacheable compilations: a fully cold build at `-j2`, 571 s plus 781 s for the two `buildext` steps and 31 minutes for the job. The repository's 10 GB Actions cache quota is saturated by PR-scoped archives: every PR job saves a 0.5 to 1.4 GB ccache or sccache archive under its own `refs/pull/N/merge` ref, which no other PR can restore, and those archives keep evicting the master-branch caches that PRs can restore. The workflow comment already states the intent ("A master push or nightly run saves this cache; pull requests restore it"), but no step passes `save:` to `hendrikmuhs/ccache-action`, whose default saves on every run.

## Goal

[...]

PR #1224, opened 2026-07-30
Save the compiler cache only outside pull requests

tigercosmos:
Related to #1222.

The comments on the ccache steps already state the policy: a master push or nightly run saves the cache and a pull request restores it. Nothing enforced it, because `hendrikmuhs/ccache-action` saves on every run by default.

A cache written during a pull-request run is scoped to that run's `refs/pull/N/merge` ref. No other pull request, and no master push, can ever restore it. It is write-only, it is read by nobody, and it still counts against the 10 GB repository quota. GitHub evicts by least recent use, so these archives crowd out the master-branch entries that pull requests do restore from.

The quota is full of exactly that. Measured today from `gh api repos/solvcon/solvcon/actions/cache/usage` and `actions/caches`, at 10.2 GB against the 10 GB limit:

| ref | caches | size |
|---|---|---|
| `refs/heads/master` | 16 | 2.93 GB |
| `refs/pull/1223/merge` | 19 | 3.40 GB |
| `refs/pull/1219/merge` | 16 | 1.85 GB |
| `refs/pull/1225/merge` | 6 | 1.04 GB |
| `refs/pull/1226/merge` | 4 | 0.45 GB |

About 6.7 GB, two thirds of the quota, is pull-request-scoped and unrestorable, against 2.93 GB of master entries that every pull request can use. The last two rows are #1225 and #1226, two pull requests in this same series, which have already written 1.5 GB that nothing will ever read.

The consequence is intermittent rather than constant, and it is worth being precise about that. While a master entry survives, pull requests restore it and hit rates are high: #1225 measured 97.84% on Linux and 97.41% on macOS today. When quota pressure evicts a master entry, the next pull request that needs it compiles from scratch. That is run 30532183716, a pull request touching only Python files, which logged "No cache found" on both the Linux and macOS ccache steps, finished 0 hits out of 231 cacheable compilations, and spent about 31 minutes building. Cutting the write-only two thirds of the quota is what makes that outcome stop recurring.

This passes `save: ${{ github.event_name != 'pull_request' }}` to every ccache-action step: the `standalone_buffer`, `build`, `build_windows`, and `sanitizer` jobs in devbuild, the `lint` job, and `nouse_install`. The condition is not a no-op on the last two even though their `if:` normally excludes pull requests, because `MMGH_FORCE_SANITIZER` and `MMGH_FORCE_NOUSE_INSTALL` can put them on a pull-request event.

The profiling job already passes `save: false`, which is stricter, and is left alone: it reads the key the `build` job writes, so making it conditional would let nightly runs start writing into that namespace.

The vcpkg `actions/cache` step is also left alone. Its primary key is a constant, and `actions/cache` skips its post-run save on an exact key hit, so once master has written that entry every pull request restores it and writes nothing.

Verified on a pull request against a fork: the ccache step logged `save: false`, and the workflows parse and schedule unchanged.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -73,6 +73,7 @@ jobs:
           key: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           restore-keys: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           create-symlink: true
+          save: ${{ github.event_name != 'pull_request' }}
           # Bound the cache so it does not grow without limit across the rolling
           # key. A master push or nightly run saves this cache; pull requests
           # restore it (a PR on a non-default base branch cannot, so it runs
@@ -157,6 +158,7 @@ jobs:
         key: ${{ runner.os }}-${{ matrix.cmake_build_type }}
         restore-keys: ${{ runner.os }}-${{ matrix.cmake_build_type }}
         create-symlink: true
+        save: ${{ github.event_name != 'pull_request' }}
         # Bound the cache so it does not grow without limit across the rolling
         # key. A master push or nightly run saves this cache; pull requests
         # restore it (a PR on a non-default base branch cannot, so it runs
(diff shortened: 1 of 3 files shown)


.github/workflows/devbuild.yml
```diff
@@ -73,6 +73,7 @@ jobs:
           key: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           restore-keys: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           create-symlink: true
+          save: ${{ github.event_name != 'pull_request' }}
```
tigercosmos, line 76: Don't cache PR code.

yungyuc:
LGTM

@chestercheng @ExplorerRay please help take a look and share what you think.

ExplorerRay:
LGTM. I think this will keep frequent PR changes from evicting the shared cache.

PR #1225, opened 2026-07-30
Run each pilot GUI test once per operating system on CI

tigercosmos:
Related to #1222.

Every pilot GUI test runs twice on a pull request: once under python plus PySide6 (`make pytest BUILD_QT=ON`) and once inside the pilot binary (`make run_pilot_pytest`). The widget classes are where both passes spend their time, while the whole non-GUI suite finishes in a few seconds.

The pilot binary embeds an interpreter nothing else exercises, so it keeps the full GUI pass. The python pass now exports `SOLVCON_SKIP_PILOT_WIDGET_TESTS` and drops the live-window widget classes, while still running the import check, the headless pilot classes (the console backend, the namespace, the call tips, the domain widget, the painter pages), and every non-GUI test. A new `tests/pilot_ci.py` holds the flag and explains why it exists.

A schedule event leaves the guard empty, so the nightly `build` job still runs every GUI test in both Qt hosts and `send_email_on_fail.yml` reports its failures. Local runs are never guarded: a developer running `make pytest` leaves the variable unset and sees no change.

The guard is an environment variable rather than a pytest marker because both Qt hosts read it the same way, the test files stay pure unittest, and it keeps working under plain `python -m unittest`.

The Windows lane needs no equivalent change. `build_windows` runs `cmake --workflow --preset ci-win-rel`, whose test preset filters on the `cpp|pilot` labels, and per `tests/CMakeLists.txt` the plain `pytest` target carries the `python` label, so Windows never ran the duplicate pass.

Measured on a pull request against a fork, ubuntu-24.04, against a baseline run of the same commit range on the same runners:

| step | before | after |
|---|---|---|
| `make pytest BUILD_QT=ON` | 96.56 s | 11.78 s |
| `make run_pilot_pytest` | 70.06 s | 72.97 s |

The coverage is relocated rather than lost. The trimmed pass reports 1642 passed and 205 skipped where the baseline reported 1834 passed and 19 skipped, and the pilot binary pass still reports the same 12 skips as the baseline, so every test the python pass now drops is executed in the pilot binary.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -224,8 +224,10 @@ jobs:
           install_name_tool -add_rpath $(qtpaths --install-prefix)/lib ${PYSIDE6_PATH}/QtCore.abi3.so
           JOB_MAKE_ARGS="${JOB_MAKE_ARGS} BUILD_METAL=ON" ; \
         fi
-        # Left at the default so one job runs the GUI tests with their windows
-        # off the screen; run_pilot_pytest below maps its windows.
+        # The live-window widget classes run once per OS, in run_pilot_pytest
+        # below; what stays here needs no window, so SOLVCON_TEST_SHOW_WINDOWS
+        # is left at the default. A schedule run leaves the guard empty.
+        export SOLVCON_SKIP_PILOT_WIDGET_TESTS=${{ github.event_name != 'schedule' && '1' || '' }}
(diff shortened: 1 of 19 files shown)


tigercosmos:
Superseded by #1235, which does this with a tests/gui directory and named make targets instead of an environment variable, following the review comment on #1222.

PR #1226, opened 2026-07-30
Make the pilot theme and terminal tests cheap

tigercosmos:
Related to #1222.

A theme switch is expensive. `RThemeManager::setMode` writes QSettings, flips `QStyleHints::setColorScheme`, then `apply()` calls `QApplication::setPalette` plus `qApp->setStyleSheet(...)`, and an application-wide stylesheet change unpolishes and repolishes every widget in the process. The theme and terminal tests paid for that far more often than they needed to. This makes them cheap rather than skipping them, so local runs get faster too.

Three changes:

- `TerminalWidgetTC` pinned the light variant in `setUp` and restored it after every test, so each of the 21 tests paid two switches before doing any work. The pin moves to `setUpClass` and the restore to `tearDownClass`, taking 42 switches down to 2.
- `apply()` now caches the last stylesheet and calls `qApp->setStyleSheet` only when it changes, and `persist()` returns early when the mode and look already match what was written.
- `ThemeConsoleTC._console_edits` polished every QTextEdit in the process right after a full repolish. It now takes the two text edits under the console dock.

Measured on ubuntu-24.04, against a pull request that changes only workflow YAML so its test steps run the same suite on the same runners. Nothing is skipped on either side:

| step | before | after |
|---|---|---|
| `make pytest BUILD_QT=ON` | 115.04 s | 48.26 s |
| `make run_pilot_pytest` | 81.78 s | 39.50 s |

Two notes for review. The palette pass in `apply()` is deliberately left unguarded, because the canvas and console listeners hang off `themeChanged` and the palette is what the tests read. The persist cache is a `std::optional` that starts absent, so the first write always happens and a session that never leaves the defaults still records them.

Found while measuring, not caused by this change: `ThemePolishTC.test_mode_and_look_persist_across_sessions` is flaky on macOS at about the same rate on an unmodified tree. Filed as #1227.

### DIFF
--- a/cpp/solvcon/pilot/theme/RThemeManager.cpp
+++ b/cpp/solvcon/pilot/theme/RThemeManager.cpp
@@ -81,7 +81,7 @@ void RThemeManager::apply()
         m_window_color = native.color(QPalette::Window);
         QApplication::setPalette(native);
         // Leave the native look untouched, including any style the platform set.
-        qApp->setStyleSheet(QString());
+        setApplicationStyleSheet(QString());
     }
     else
     {
@@ -89,7 +89,7 @@ void RThemeManager::apply()
             buildPalette(themePaletteFor(m_backend->platform(), variant), variant);
         m_window_color = curated.color(QPalette::Window);
         QApplication::setPalette(curated);
-        qApp->setStyleSheet(supplementalStyleSheet(curated));
+        setApplicationStyleSheet(supplementalStyleSheet(curated));
     }
     if (m_window != nullptr)
     {
@@ -242,6 +242,19 @@ QPalette RThemeManager::buildPalette(ThemePalette const & spec, ThemeVariant var
     return pal;
 }
 
+void RThemeManager::setApplicationStyleSheet(QString const & sheet)
+{
+    // An application-wide stylesheet change unpolishes and repolishes every
+    // widget in the process, so pay for it only when the text really differs.
+    if (sheet == m_style_sheet)
+    {
+        return;
+    }
+
+    m_style_sheet = sheet;
+    qApp->setStyleSheet(sheet);
+}
+
 QString RThemeManager::supplementalStyleSheet(QPalette const & pal) const
 {
     // A QPalette cannot draw a tooltip border or a focus ring, so add just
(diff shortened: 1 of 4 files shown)


tests/test_pilot_terminal.py
```diff
@@ -42,14 +42,19 @@ class TerminalWidgetTC(unittest.TestCase):
...
     def setUpClass(cls):
         cls.mgr = pilot.RManager.instance.setUp()
+        # Pin a known variant so the syntax and output colors are
+        # deterministic, and restore the shared default afterward. A theme
+        # switch repolishes every widget, so pay for it once for the class
+        # rather than twice per test.
+        cls.mgr.set_theme("light")
+
+    @classmethod
+    def tearDownClass(cls):
```
tigercosmos, line 52: Reduce reverting from 42 times to 2 times.

cpp/solvcon/pilot/theme/RThemeManager.cpp
```diff
@@ -265,11 +278,20 @@ void RThemeManager::restorePersisted()
...
 }
 
-void RThemeManager::persist() const
+void RThemeManager::persist()
 {
+    // The store is only read when a session starts, so a write that repeats
+    // what was already written buys nothing and touches the disk.
+    if (m_persisted_mode == m_mode && m_persisted_look == m_look)
+    {
+        return;
```
tigercosmos, line 287: return if nothing changes.

tests/test_pilot_theme.py
```diff
@@ -278,14 +278,22 @@ def tearDown(self):
...
-        # instances in the pilot. Deliver the posted palette-change events and
-        # polish each so its palette resolves against the current application
-        # palette before it is read.
+        # instances under its dock. Deliver the posted palette-change events
+        # and polish each so its palette resolves against the current
+        # application palette before it is read. Polishing every text edit in
+        # the process instead would repolish widgets this test says nothing
+        # about, at the price of re-running their stylesheets.
         self.app.processEvents()
-        edits = [w for w in self.app.allWidgets()
```
tigercosmos, line 287: Avoid `allWidgets`.

cpp/solvcon/pilot/theme/RThemeManager.cpp
```diff
@@ -242,6 +242,19 @@ QPalette RThemeManager::buildPalette(ThemePalette const & spec, ThemeVariant var
     return pal;
 }
 
+void RThemeManager::setApplicationStyleSheet(QString const & sheet)
+{
+    // An application-wide stylesheet change unpolishes and repolishes every
+    // widget in the process, so pay for it only when the text really differs.
```
yungyuc, line 248:
```suggestion
    // Only pay for an application-wide stylesheet change when the text really differs.
```

"unpolishes and repolishes every widget" is excessive information.

tests/test_pilot_terminal.py
```diff
@@ -42,14 +42,19 @@ class TerminalWidgetTC(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.mgr = pilot.RManager.instance.setUp()
+        # Pin a known variant so the syntax and output colors are
+        # deterministic, and restore the shared default afterward. A theme
```
yungyuc, line 46: This is a good place to explain what is happening in the system.  When the system changes, this is the document for what it did in the past.  The discrepancy will signal for an action to review.

yungyuc (commented):
LGTM

PR #1228, opened 2026-07-30
Add GCC 16 apt helper for Ubuntu 24.04

### LINKED ISSUE #1220: Add information for how to install GCC 16 in Linux dependency build script
The current code requires C++23 which is only supported in version 16 when the compiler is GCC.  But the build script only [implicitly mentioning it in `build-scdv.sh`](https://github.com/solvcon/solvcon/blob/675a209d548dd30c7e919b382902cb56c22a795e/contrib/dependency/build-scdv.sh#L270) without clarity:

```
  # GCC 16 trunk on Ubuntu 24.04 rejects the AVX512 `evex512` target attribute
  # that numpy 2.5.x uses, so cap cpu-dispatch at AVX2 instead of MAX.  Pass
  # via --config-settings so the pip-driven meson rebuild honors it (not just
  # an out-of-tree `spin build`).
```

`build-scdv.sh` should add `scdv_apt_gcc` function to report how to install GCC 16 on Ubuntu 24.04.

yugalredhu097:
## Summary

This PR adds a new `scdv_apt_gcc_cmd()` helper to print the commands required to install and configure GCC 16.

It also updates `plat_print_deps()` to include the GCC 16 installation instructions in the printed dependency setup.

The commands match the existing GCC 16 installation steps used in `.github/actions/setup_linux/action.yml`.

Related to #1220

### DIFF
--- a/contrib/dependency/build-scdv.sh
+++ b/contrib/dependency/build-scdv.sh
@@ -233,9 +233,25 @@ sudo ln -fs "$(which clang-tidy-22)" /usr/local/bin/clang-tidy
 EOF
 }
 
+scdv_apt_gcc_cmd() {
+    # Print the apt commands for installing GCC 16.
+    # The script never runs apt itself; copy the output, review it, and run it.
+   cat <<'EOF'
+sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+sudo apt-get -qqy update
+sudo apt-get -qy install gcc-16 g++-16
+sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-16 100
+sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-16 100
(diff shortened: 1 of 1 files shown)


contrib/dependency/build-scdv.sh
```diff
@@ -233,10 +233,26 @@ sudo ln -fs "$(which clang-tidy-22)" /usr/local/bin/clang-tidy
 EOF
 }
 
+scdv_apt_gcc_cmd() {
```
yungyuc, line 236: We encourage the practice to always leave an inline annotation (like this) in the PR for anything that is one a one-liner change.  It helps guide the reviewers.

This PR is simple enough and it does not matter to annotate.  But it's a good habit to do it for future PRs.

yungyuc (approved):
LGTM.  Thanks for the contribution.

PR #1231, opened 2026-07-30
Add the plot color vocabulary and the xy plot series

### LINKED ISSUE #1049: Use native Qt to draw XY plot for 1D Euler solver
We used [matplotlib](https://matplotlib.org/) for 2D plotting in pilot.  To better interact with the 2D geometry canvas and 2/3D visualization, we will develop in-house, Qt-based plotting tools.  The Qt-based plotting code will allow us to interact tightly with the whole pilot GUI.

The existing 2D plotting code is abstracted in [`solvcon.pilot._base_app`](https://github.com/solvcon/solvcon/blob/3b4bc6e6ed6a4e2b24226ac4f87a9879177bae08/solvcon/pilot/_base_app.py#L10).  We should first analyze the code to create a plan for the renovation, and then replace the matplotlib code with native Qt plotting code step by step.

j8xixo12:
The pilot has no native xy plot: a plot of solver output is assembled in Python, so the samples cross the binding once per point.

This adds the two Qt-free pieces the rest of the plot core builds on. `PlotColor` is one sRGB color with alpha, together with the matplotlib C0-C9 cycle; reusing matplotlib's ten colors in its order means a pilot plot carries the same color meaning as the matplotlib plot the reader would otherwise have made. `RPlotSeries` holds one xy series as a contiguous `SimpleCollector<double>` pair with its label, color, and stroke width, plus lazily cached data limits that skip any sample whose x or y is not finite.

`set_data` copies its operands, the way matplotlib copies the arrays handed to plot, so a later write through the NumPy array cannot leave the samples and the cached extent disagreeing. A SimpleCollector wraps a whole buffer and carries no shape, stride, or ghost, so set_data rejects an operand that is not one-dimensional, ghost-free, or unit strided, and validates both before assigning either.

Neither plot header includes Qt; converting a PlotColor to a QColor stays in the Qt front end. Behavior is covered by `tests/test_pilot_plot.py`. The `std::span` accessors stay unbound because a Python view would dangle at the next set_data.

Related to #1049 step 1.

### DIFF
--- a/cpp/solvcon/pilot/plot/RPlotSeries.cpp
+++ b/cpp/solvcon/pilot/plot/RPlotSeries.cpp
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/pilot/plot/RPlotSeries.hpp>
+
+#include <cmath>
+#include <format>
+#include <stdexcept>
+
+namespace solvcon
+{
+
+namespace
+{
+
+void validate_operand(SimpleArray<double> const & arr, char const * name)
+{
+    if (arr.ndim() != 1)
+    {
+        throw std::invalid_argument(
+            std::format("RPlotSeries::set_data: {} must be 1-dimensional, but ndim is {}", name, arr.ndim()));
+    }
+    if (arr.nbody() > 1 && arr.stride(0) != 1)
+    {
+        throw std::invalid_argument(
+            std::format(
+                "RPlotSeries::set_data: {} must be contiguous with unit stride, but stride is {}",
+                name,
+                arr.stride(0)));
+    }
+    // The collector clones the whole buffer, so the ghost part would become
+    // samples that the nbody-based length check never saw.
+    if (arr.nghost() != 0)
+    {
+        throw std::invalid_argument(
+            std::format("RPlotSeries::set_data: {} must be ghost-free, but nghost is {}", name, arr.nghost()));
+    }
+}
+
+} /* end namespace */
+
+void RPlotSeries::set_data(SimpleArray<double> const & x, SimpleArray<double> const & y)
+{
+    validate_operand(x, "x");
+    validate_operand(y, "y");
+    if (x.nbody() != y.nbody())
+    {
+        throw std::invalid_argument(
+            std::format(
+                "RPlotSeries::set_data: x and y must have the same length, but they are {} and {}",
+                x.nbody(),
+                y.nbody()));
+    }
+
+    m_x = SimpleCollector<double>(x);
+    m_y = SimpleCollector<double>(y);
+    m_limits_stale = true;
+}
+
+void RPlotSeries::clear_data()
+{
+    m_x = SimpleCollector<double>();
+    m_y = SimpleCollector<double>();
+    m_limits_stale = true;
+}
+
+double RPlotSeries::x_at(std::size_t it) const
+{
+    if (it >= size())
+    {
+        throw std::out_of_range(
+            std::format("RPlotSeries::x_at: index {} is out of bounds with size {}", it, size()));
+    }
+    return x()[it];
+}
+
+double RPlotSeries::y_at(std::size_t it) const
+{
+    if (it >= size())
+    {
+        throw std::out_of_range(
+            std::format("RPlotSeries::y_at: index {} is out of bounds with size {}", it, size()));
+    }
+    return y()[it];
+}
+
+std::optional<std::array<double, 4>> RPlotSeries::data_limits() const
+{
+    if (!m_limits_stale)
+    {
+        return m_limits;
+    }
+
+    std::optional<std::array<double, 4>> limits;
+    std::span<double const> const xs = x();
+    std::span<double const> const ys = y();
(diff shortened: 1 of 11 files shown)


cpp/solvcon/pilot/plot/RPlotSeries.hpp
```diff
@@ -0,0 +1,89 @@
...
+    RPlotSeries(RPlotSeries &&) = default;
+    RPlotSeries & operator=(RPlotSeries const &) = default;
+    RPlotSeries & operator=(RPlotSeries &&) = default;
+    ~RPlotSeries() = default;
+
+    void set_data(SimpleArray<double> const & x, SimpleArray<double> const & y);
+
+    void clear_data();
+
+    std::size_t size() const { return m_x.size(); }
```
j8xixo12, line 46: Since the series length can change, we need to recalculate its size on every call.

cpp/solvcon/pilot/plot/RPlotSeries.cpp
```diff
@@ -0,0 +1,143 @@
...
+            std::format(
+                "RPlotSeries::set_data: x and y must have the same length, but they are {} and {}",
+                x.nbody(),
+                y.nbody()));
+    }
+
+    m_x = SimpleCollector<double>(x);
+    m_y = SimpleCollector<double>(y);
+    m_limits_stale = true;
+}
```
j8xixo12, line 60: Users need to call `set_data` to update the series content, which copies the data into RPlotSeries to prevent unexpected mutations. Calling `set_data` also set `m_limits_stale = true` to trigger `data_limits` to recalculate the series data range, ensuring autoscale works properly for future plots.

cpp/solvcon/pilot/plot/plot_style.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+std::span<PlotColor const> plot_color_cycle()
+{
+    return std::span<PlotColor const>(PLOT_COLOR_TABLE.data(), PLOT_COLOR_TABLE.size());
+}
+
+PlotColor plot_cycle_color(std::size_t index)
+{
+    return PLOT_COLOR_TABLE[index % PLOT_COLOR_TABLE.size()];
+}
```
j8xixo12, line 41: If the user doesn't specify a line color, a color cycle automatically assigns one for each newly added series. Currently, up to 10 distinct line colors are supported, if there are more than 10 lines, the color sequence will repeat. This api will be used for future PR.

yungyuc:
> Replace this paragraph with a clear prose description of what this PR changes and why.

It seems that the PR is not yet ready for review.  Please leave a global comment when it is, after all descriptions are cleaned up.

j8xixo12:
@yungyuc This PR is ready for review.

cpp/solvcon/pilot/plot/wrap_plot.cpp
```diff
@@ -0,0 +1,197 @@
...
+            //
+            ;
+
+        (*this)
+            .def("set_data", &wrapped_type::set_data, py::arg("x"), py::arg("y"))
+            .def("clear_data", &wrapped_type::clear_data)
+            .def_property_readonly("size", &wrapped_type::size)
+            .def("__len__", &wrapped_type::size)
+            .def(
+                "x_at",
```
yungyuc: How about naming them as `x()` and `y()`?  `at()` is a naming convention for bound-checking in C++.  When exposing to Python, it's OK to assume bound-checking is always there because Python element-wise runtime is slow.

yungyuc (changes requested):
- [x] Discuss: dropping `_at()` when exposing bound-checking accessors to Python.

cpp/solvcon/pilot/plot/wrap_plot.cpp
j8xixo12: I prefer keeping `_at`. I'm concerned it might make the naming style in solvcon inconsistent, since I noticed the following that C++ functions like `x_at` or `y_at` are exported to Python without any modifications.

https://github.com/solvcon/solvcon/blob/17c9def65e8b743404c6c7e5799925c6ef95afa8/cpp/solvcon/universe/pymod/wrap_shape0d.cpp#L366-L390

yungyuc: What you pointed out was a counter example.  `x_at()` was added because `x()` has been there:

```cpp
    (*this)
        DECL_WRAP(x_at)
        DECL_WRAP(y_at)
        DECL_WRAP(z_at)
        ;
// ...
    (*this)
        DECL_WRAP(x)
        DECL_WRAP(y)
        DECL_WRAP(z)
        ;
```

When there is not non-bound-checking element accessors, we should just use `x()` in Python for `x_at()` in C++.

It is not perfectly consistent, but is a convention I practiced.  I suggest to follow it before we run into an issue.

j8xixo12: Got it, let's drop `_at` for Python element accessors.

j8xixo12:
@yungyuc This PR is ready for the next review.

yungyuc (approved):
LGTM.  @j8xixo12 Please append the PR after the corresponding step in issue #1049.

yungyuc:
@tigercosmos Please take a look.  You will need to use (and improve) it when analyzing data.

PR #1235, opened 2026-07-30
Group the window-driving tests under tests/gui

tigercosmos:
Related to #1222. Replaces #1225, which did the same job with an environment variable folded into 31 `skipIf` conditions. This follows the review comment on #1222 asking for sub-directories under `tests/` and a way to select the slow tests that is clear to a human.

First of a short series. This one is the layout and the targets; the files that mix window-driving and headless classes follow separately.

A test belongs under `tests/gui/` when it needs a real top-level window surface. That is the property `NO_LIVE_WINDOW` already guards: the offscreen Qt platform and the headless Windows runner cannot back a live window, so these tests skip there, and driving real windows is what makes them slow. Those guards are untouched, because they answer a different question than the directory does. The directory says which lane a test runs in; the guard says whether the platform can run it at all.

Seven files move, the ones whose classes are all window-driving. They are pure renames with no content change. The names lose the now redundant `pilot` prefix, which also keeps every test module basename unique, something pytest needs while `tests/` has no `__init__.py`.

Three targets, so a developer can say what to run instead of what to skip:

| target | runs |
|---|---|
| `make pytest` | everything, unchanged |
| `make pytest-fast` | everything except `tests/gui/` |
| `make pytest-gui` | only `tests/gui/` |

The pull-request lane calls `pytest-fast`, because the same tests run seconds later inside the pilot binary in the same job, where the embedded interpreter is the environment nothing else exercises. A push and the nightly schedule call `pytest`, so master and nightly keep running the whole suite in both Qt hosts. #1225 got this part wrong: it keyed on `!= 'schedule'`, which also disabled the tests on master pushes.

`tests/gui/README.md` states the boundary and the commands.

Once the series is complete, the layout becomes machine-checkable: a live-window guard outside `tests/gui/` is then a lint failure rather than a convention someone has to remember.


Makefile
```diff
@@ -139,6 +139,21 @@ pytest: buildext
...
+# pytest-gui when the change is.
+.PHONY: pytest-fast
+pytest-fast: buildext
+	env $(RUNENV) \
+		$(PYTEST) $(PYTEST_OPTS) tests/ --ignore=tests/gui
+
+.PHONY: pytest-gui
+pytest-gui: buildext
+	env $(RUNENV) \
+		$(PYTEST) $(PYTEST_OPTS) tests/gui/
```
tigercosmos, line 155: Meat.

Makefile
```diff
@@ -139,6 +139,21 @@ pytest: buildext
 	env $(RUNENV) \
 		$(PYTEST) $(PYTEST_OPTS) tests/
 
+# The tests under tests/gui/ drive real top-level windows, so they are the
+# slow half of the suite and the half a headless machine cannot run at all.
+# pytest runs everything; these two run one side of that split. Use
+# pytest-fast while iterating on something that is not the GUI, and
+# pytest-gui when the change is.
```
yungyuc, line 146: The comment block is useless.  `tests/gui` is clear enough that it is for GUI (top-level window).

```suggestion
```

Makefile
yungyuc: Beautiful.

Makefile
tigercosmos: let me fix it.

PR #1236, opened 2026-07-30
Split the mixed pilot test files along the window line

tigercosmos:
Stacked on master now that #1235 has merged; GitHub cannot base a cross-fork pull request on a fork branch. Until #1235 merges, Files changed shows its commits too. **Review only commit `ba64453e`.**; the rest is #1235.

Related to #1222. Stacked on #1235, which created `tests/gui/` and moved the files whose classes were all window-driving. This one starts on the files that mix window-driving and headless classes, so they have to be split rather than moved.

Code motion only. Every moved class body is byte-identical to its original, checked mechanically rather than by eye, and no assertion, `setUp` body, or skip condition changed.

| source | stays | moves to `tests/gui/` |
|---|---|---|
| `test_pilot_2d.py` | 6 headless classes | `test_2d.py`: `R2DWidgetSelectToolTC`, `R2DWidgetThemeTC`, `PainterToolboxTC` |
| `test_pilot_draw_tool.py` | `PainterIconTC` | `test_draw_tool.py`: `DrawToolTC` |
| `test_pilot_painter_design.py` | `_StubCanvas`, `PainterDesignPageTC` | `test_painter_design.py`: `PainterDesignCanvasTC` |

`_load_rgba` and `_drawn_pixels` are used from both sides, so they are duplicated rather than imported across the directory boundary. A shared import between `tests/` and `tests/gui/` would recouple the two lanes this layout exists to separate. Helpers left with no user (`_send_mouse`, `_click`) and the imports they needed are deleted from the sources.

Two docstrings now name a class that lives in the other file: `PainterToolboxTC`'s cites `R2DWidgetWorldTC.test_draw_tool_round_trip`, and `_StubCanvas`'s cites `PainterDesignCanvasTC`. Both are left verbatim, since rewording them here would bury real edits inside a pure-motion change. Worth a follow-up if you want them chased down.


tests/test_pilot_2d.py
```diff
@@ -389,87 +364,6 @@ def test_circle_renders_as_hollow_loop_on_locus(self):
                 "circle interior should be hollow")
 
 
-@unittest.skipIf(NO_LIVE_WINDOW or not solvcon.HAS_PILOT,
```
tigercosmos, line 392: Move all `NO_LIVE_WINDOW` to `tests/gui/`

PR #1237, opened 2026-07-30
Split the mixed shortcut and inspector tests into tests/gui

tigercosmos:
Stacked on master now that #1236 has merged. **Review only commit `f40b0e23`.**

Related to #1222. Stacked on #1236. Continues splitting the files that mix window-driving and headless classes.

Code motion only. Every moved and every remaining class body is byte-identical to its original, checked mechanically.

| source | stays | moves to `tests/gui/` |
|---|---|---|
| `test_pilot_painter_layers.py` | `_StubCanvas`, `PainterLayersPageTC` | `test_painter_layers.py`: `PainterLayersCanvasTC` |
| `test_pilot_shortcut.py` | `ShortcutResolutionTC` | `test_shortcut.py`: `ShortcutTC`, `ApplyShortcutHelperTC`, `DrawToolShortcutTC` |
| `test_pilot_solution_info.py` | `ComputeFieldTC` | `test_solution_info.py`: `SolutionInfoTC`, `SolutionInspectorTC` |

`_click`, `_PANEL_CHORDS`, and `_DRAW_TOOL_KEYS` are read from both sides and are duplicated rather than imported across the directory boundary.

`DrawToolShortcutTC` is guarded by `NO_LIVE_WINDOW or not solvcon.HAS_PILOT or QTest is None`, and `QTest` comes from a separate import guard that deliberately sits outside the main one, with a comment explaining that macOS CI rewrites the Qt rpath for QtCore, QtGui, and QtWidgets only. That block moves verbatim with the class, comment included, and leaves the source since nothing there references `QTest` any more.

The single added line in a source file is `from PySide6 import QtGui`, narrowed from an import that also brought in `QtCore` and `QtWidgets`, which no longer have users there.

### DIFF
--- a/tests/gui/test_painter_layers.py
+++ b/tests/gui/test_painter_layers.py
@@ -1,0 +1,164 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+Tests for the Layers page of the Painter inspector.
+"""
+
+import os
+import unittest
+
+import solvcon
+
+try:
+    from solvcon import pilot
+    from solvcon.pilot.base import _gui
+    from PySide6 import QtCore, QtGui, QtWidgets
+except ImportError:
+    pilot = None
+
+# The offscreen platform cannot back a live window; neither can the headless
+# Windows CI runner, whose WARP software rasterizer faults on one.
+NO_LIVE_WINDOW = ((os.getenv('QT_QPA_PLATFORM') or '').startswith('offscreen')
+                  or ('nt' == os.name and bool(os.getenv('GITHUB_ACTIONS'))))
+
+
+def _click(widget, x, y):
+    """Post a synthetic left-button press and release to ``widget``."""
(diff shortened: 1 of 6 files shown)


tests/gui/test_painter_layers.py
```diff
@@ -0,0 +1,164 @@
...
+             QtCore.Qt.LeftButton, QtCore.Qt.LeftButton),
+            (QtCore.QEvent.Type.MouseButtonRelease,
+             QtCore.Qt.LeftButton, QtCore.Qt.NoButton)):
+        QtWidgets.QApplication.sendEvent(
+            widget,
+            QtGui.QMouseEvent(etype, pos, glob, button, buttons,
+                              QtCore.Qt.NoModifier))
+
+
+@unittest.skipIf(NO_LIVE_WINDOW or not solvcon.HAS_PILOT,
```
tigercosmos, line 41: Move all `NO_LIVE_WINDOW` test cases.

PR #1240, opened 2026-07-30
Give the world a change stamp for its readers to poll

tigercosmos:
The panels that show what a world holds had no cheap way to notice an edit: the Painter's Layers page serialized the whole world every 250 ms and the entity tree every 500 ms, both only to learn whether anything had happened. `World` now carries a stamp that moves whenever its geometry does, and both panels poll that instead, serializing only when there is something new to show.

Every mutation funnels through a handful of internal primitives, so the stamp is bumped in those rather than in each public entry point, and undo and redo move it as the edits they replay do. What it does not cover is written down where it is declared: a caller that reaches past the world into the pads `points()`, `segments()`, and `curves()` hand out moves geometry the stamp does not see, and the undo history is not geometry, so `can_undo()` can turn true without it moving. A test pins the first of those so the limit stays honest.

The entity tree's cache is keyed on the stamp together with the world it came from, and the Layers page's poll key likewise. Both hold the world as the object rather than as its `id`, since stamps count per world and an id a dead world freed can be handed out to the next.

Related to #1241.

### DIFF
--- a/cpp/solvcon/universe/World.hpp
+++ b/cpp/solvcon/universe/World.hpp
@@ -326,6 +326,7 @@ public:
     void add_point(point_type const & vertex)
     {
         m_points->append(vertex);
+        mark_changed();
     }
     void add_point(value_type x, value_type y, value_type z)
     {
@@ -344,6 +345,7 @@ public:
     {
         m_segments->append(segment);
         m_bare_segment_indices.push_back(m_segments->size() - 1);
+        mark_changed();
     }
     void add_segment(point_type const & p0, point_type const & p1)
     {
@@ -361,10 +363,12 @@ public:
     void add_bezier(bezier_type const & bezier)
     {
         m_curves->append(bezier);
+        mark_changed();
     }
     void add_bezier(point_type const & p0, point_type const & p1, point_type const & p2, point_type const & p3)
     {
         m_curves->append(p0, p1, p2, p3);
+        mark_changed();
     }
     size_t nbezier() const { return m_curves->size(); }
     bezier_type bezier(size_t i) const { return m_curves->get(i); }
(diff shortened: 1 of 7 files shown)


cpp/solvcon/universe/World.hpp
```diff
@@ -583,6 +602,9 @@ class World
 
 private:
 
+    /// Stamp the world as changed; see version().
+    void mark_changed() { ++m_version; }
```
tigercosmos, line 593: Every change bumps the version.

solvcon/pilot/painter/_layers.py
```diff
@@ -307,26 +307,32 @@ def _selection(self):
         selected = widget.selectedShape
         return world, selected if world.shape_is_live(selected) else -1
 
-    def _snapshot(self):
-        """Fingerprint and row content for the bound canvas, one world read.
+    def _state_key(self):
```
tigercosmos, line 310: Now `_state_key` is cheap.

cpp/solvcon/universe/World.hpp
```diff
@@ -700,6 +709,7 @@ class World
     std::vector<ShapeRecord> m_shape_registry;
 
     size_t m_nshape = 0; ///< Count of live (non-DEAD) shapes.
+    uint64_t m_version = 0; ///< Moves on every change; see version().
```
yungyuc, line 712: I am not sure it's a good idea to use the name `version`, but let's keep it for a while.

yungyuc (commented):
LGTM

cpp/solvcon/universe/World.hpp
tigercosmos: Let me change to a better name.

Issue #1233, opened 2026-07-30
Move the pilot's Qt style sheets out of Python widget code


tigercosmos:
Reviewing #1201, yungyuc noted that hard-coding Qt style sheets in Python code is not a good idea and that it should move to a more organized structure when there is an opportunity: https://github.com/solvcon/solvcon/pull/1201#discussion_r3682910201

Today every pilot widget that wants a look writes its own style sheet inline. The rules live as long f-strings in the middle of widget code, each page repeats the selector names it shares with its neighbors, and nothing outside the widget can see or override what it asked for. The current sites are:

- `solvcon/pilot/canvas/_painter_design.py:412` (Design page: name label, axis labels, badge, field frames, line edits)
- `solvcon/pilot/canvas/_painter_layers.py:401` (Layers page, appending `_painter_rows.rules()`)
- `solvcon/pilot/canvas/_painter_gui.py:188` and `:271` (tool buttons and page tabs)
- `solvcon/pilot/canvas/_painter_rows.py:135` (`rules()`, the one piece already factored out of a page)
- `solvcon/pilot/panel/_tree_panel.py:297`
- `solvcon/pilot/panel/_profiling.py:109` and `:305`

Two different problems are mixed in that list. The canvas pieces build their colors from the palette, so they follow a light/dark switch, and their trouble is only that the rules sit inline and are duplicated between pages. The panel pieces are worse: `_profiling.py` names dark hex values (`#2e2e2e`, `#dcdcdc`, `#555555`) directly, so the profiling tree stays dark under a light theme and ignores the curated palette entirely, and `_tree_panel.py` reaches for an `rgba(127, 127, 127, 40)` literal for the same reason.

There is already a model for the organized form on the C++ side. `RThemeManager::supplementalStyleSheet` (`cpp/solvcon/pilot/theme/RThemeManager.cpp:245`) builds the application-wide sheet from the curated `QPalette` in one place, and `RThemeManager::apply` installs it whenever the theme changes. `_painter_rows.rules()` is the small Python echo of the same idea: rules that travel with the thing they style, so the two lists that draw the same row cannot drift apart.

What this issue asks for:

- Give the pilot's Python side one place that owns style rules, so a widget asks for the rules it needs instead of spelling them out inline. `_painter_style.py` is the natural home for the canvas ones.
- Derive every color from the palette. No literal hex or rgba in widget code, which is what makes the profiling tree ignore the theme today.
[...]

PR #1242, opened 2026-07-30
Move the pilot's Qt style sheets out of widget code

tigercosmos:
Every pilot widget that wanted a look wrote its own style sheet inline, and the panel sheets named dark hex values directly, so the profiling tree stayed dark under a light theme.

`solvcon/pilot/_style` now holds the shared pieces: `Shades` for the colors a widget reads off its palette, `RuleCatalog` for composing a sheet out of named roles, and the `PaletteStyle` mixin that re-applies a sheet when the palette changes. `painter/_style` and `panel/_style` subclass `RuleCatalog` with one method per role, and a widget asks `Rules.sheet()` for the roles it draws.

The painter pages come out at the colors and metrics they had. The profiling result trees and the collapsible section header change on screen: they now mix every color from the palette and follow a theme switch.

Related to #1233.

### DIFF
--- a/solvcon/pilot/painter/_canvas.py
+++ b/solvcon/pilot/painter/_canvas.py
@@ -8,10 +8,11 @@ The Canvas page of the Painter inspector: how the canvas shows the world.
 import json
 import math
 
-from PySide6 import QtCore, QtGui, QtWidgets
+from PySide6 import QtCore, QtWidgets
 
+from .._style import PaletteStyled
 from ._sections import Placeholder, Section
-from ._style import blend, mono_font, rule, shade, PaletteStyled
+from ._style import Parts, Rules
 from ..panel._tree_panel import EntityTreeWidget
 
 __all__ = [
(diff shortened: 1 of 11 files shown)


solvcon/pilot/painter/_canvas.py
```diff
@@ -317,44 +308,7 @@ def _show(self, widget, zoom, center):
...
         """Color the section heads, the readout, and the buttons from the
         palette."""
-        palette = self.palette()
-        text = palette.color(QtGui.QPalette.WindowText)
-        panel = palette.color(QtGui.QPalette.Window)
-        muted = blend(text, panel, self._MUTED_MIX)
-        greyed = blend(text, panel, self._GREYED_MIX)
-        border = shade(self, self._BORDER_MIX)
-        base = palette.color(QtGui.QPalette.Base)
-        self.setStyleSheet(f"""
```
tigercosmos, line 327: Remove hardcoded styles into dedicated style classes.

solvcon/pilot/panel/_style.py
```diff
@@ -0,0 +1,68 @@
...
+]
+
+_TITLE_PX = 14
+
+# How far a hovered section header sits from the panel.
+_GRID_MIX = 0.12
+_HOVER_MIX = 0.12
+
+
+class Rules(RuleCatalog):
```
tigercosmos, line 23: Example of the style class.

tigercosmos (commented):
This PR can be a temporary solution for the QT style issues.
After investigating other QT projects, there is not the best practice yet.
We may need to do a better style system design specific for solvcon.
cc @yungyuc

PR #1246, opened 2026-07-31
Give the agent a scene inventory and a replayed conversation (step 1)

tigercosmos:
Step 1 of 2. This adds the machinery for replaying recorded turns; nothing calls it with a history yet, so the request this branch sends is byte-for-byte what it sent before. Step 2 wires the callers and adds the scene inventory.

`HistoryFormatter` renders recorded turns into a section that fits a character budget. Turns are taken newest first so growth drops the oldest, a recent failure is pinned so the error outlives the turn that raised it, and every dropped run is announced so what is left cannot read as one unbroken conversation. Foreign prose is flattened to a single line, and command arguments and results are clipped with the amount lost named.

`AgentBackend` gains one `_sections()` that both the composed payload and the new `history_section()` are built from, so a caller showing what was replayed cannot drift from what was sent. `send()` takes the turns to replay, defaulting to none.

The system prompt gains a paragraph on reading the scene and the conversation. It describes the per-shape inventory format that step 2 renders.

Related to #1206. Related to #1136. Related to #966.

### DIFF
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -17,8 +17,29 @@ import abc
 import dataclasses
 import json
 
+from . import _command
 
-class ToolSurfaceFormatter:
+
+class TextFormatter:
+    """Rendering helpers every request section shares.
+
+    :meth:`one_line` is what keeps a foreign string on the line it was given:
+    tool descriptions, model prose, and error text all land in a line-oriented
+    payload, where an embedded newline would read as the next section.
+    """
+
+    @classmethod
+    def literal(cls, value):
+        # allow_nan=False: a non-finite float must raise rather than dump as
+        # the NaN and Infinity tokens no JSON reader accepts.
+        return json.dumps(value, separators=(",", ":"), allow_nan=False)
+
+    @classmethod
+    def one_line(cls, text):
+        return " ".join(str(text).split())
+
+
+class ToolSurfaceFormatter(TextFormatter):
     """The renderers that turn JSON Schema tool definitions into the compact
     per-op signatures the model reads."""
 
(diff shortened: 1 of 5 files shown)


solvcon/agent/_backend.py
```diff
@@ -142,6 +155,176 @@ def format_tool_surface(tool_surface):
     return ToolSurfaceFormatter.render(tool_surface)
 
 
+class HistoryFormatter(TextFormatter):
```
tigercosmos, line 158: Format the raw JSON history. History will be used in the step 2 PR.

solvcon/agent/_backend.py
```diff
@@ -142,6 +155,176 @@ def format_tool_surface(tool_surface):
...
+        arguments = ({name: value for name, value in command.items()
+                      if name != "op"}
+                     if isinstance(command, dict) else command)
+        return "  %s %s -> %s" % (
+            _command.op_of(command),
+            cls.clip(cls.value_text(arguments), cls.PART_CAP),
+            cls.outcome_text(result))
+
+    @classmethod
+    def turn(cls, turn):
```
tigercosmos, line 216: Each turn includes a human requests, agent reponse, and command results.

solvcon/agent/_backend.py
```diff
@@ -186,6 +369,15 @@ class AgentBackend(abc.ABC):
         "operation and then its arguments. Pass no key an operation does "
         "not list: an unlisted key is rejected, not ignored.\n"
         "\n"
+        "Reading the scene and the conversation. The scene lists the shapes "
```
tigercosmos, line 372: Related instruction.

PR #1251, opened 2026-07-31
Give the agent a scene inventory and a replayed conversation (step 2)

tigercosmos:
Step 2 of 2, on top of #1246. That branch added `HistoryFormatter` and widened `send()`, but no caller passed a history, so the request never carried a conversation section. This branch is what puts it there.

`AgentSession.history()` returns the recorded turns, dropping the trailing user turn the composed request already carries, and both `run_turn` and the panel worker hand it to the backend. Binding the session to a different world marks the transcript, so shape ids from the canvas that was open then are never replayed unmarked beside the new world inventory.

The scene told the model only how many shapes were on the canvas and which types, so it had no id to edit and redrew instead. Each shape is now listed as `#id type [x_min, y_min, x_max, y_max]`, the format the step 1 system prompt describes, keeping the newest `INVENTORY_LIMIT` and counting the rest. The geometry `describe_state` also carries is left out of the prompt.

In the panel, the reply is the log messages of the commands that ran, followed by a line per command that failed or never ran, so a turn whose batch died cannot read as a turn that worked. The conversation being replayed goes to the Python console.

Related to #1206. Related to #1136. Related to #966.

### DIFF
--- a/solvcon/agent/_core.py
+++ b/solvcon/agent/_core.py
@@ -13,19 +13,24 @@ applied command into a transcript a caller can render.  No Qt is imported.
 import json
 import dataclasses
 
+from . import _backend
+from . import _command
+
 
 @dataclasses.dataclass
 class TranscriptTurn:
     """One transcript entry: a ``role`` with its text, commands, and results.
 
     ``results`` holds ``CommandResult`` objects (or any object with an ``ok``
-    attribute).
+    attribute).  ``failed`` marks a turn that went wrong before any command
+    ran, which no result can record.
     """
 
     role: str
     text: str = ""
     commands: list = dataclasses.field(default_factory=list)
     results: list = dataclasses.field(default_factory=list)
+    failed: bool = False
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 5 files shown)


solvcon/agent/_core.py
```diff
@@ -120,15 +194,9 @@ def scene_context(self, level="basic"):
...
-
-    @staticmethod
-    def _op_of(command):
-        """The command's declared ``op``, or ``"?"`` when it names none."""
-        if not isinstance(command, dict):
-            return "?"
-        op = command.get("op")
-        return op if isinstance(op, str) else "?"
+        lines = ["world with %d shapes (types: %s)" % (len(shapes), kinds)]
+        lines.extend(self._inventory(shapes))
```
tigercosmos, line 198: Insert world shape inventory.

solvcon/pilot/agent/_agent_gui.py
```diff
@@ -241,11 +246,15 @@ def _on_submitted(self, prompt):
         self._active_widget = widget
         self._panel.set_busy(True)
         self._panel.start_working()
+        history = session.history()
         scene = _agent_control.pilot_scene_context(
             session.runner, session.scene_context())
+        tool_surface = session.tool_surface()
+        self._write_history_payload(session.backend, prompt, scene,
```
tigercosmos, line 253: Now send the history as the context to the agent.

Issue #1250, opened 2026-07-31
Causal time-base alignment for the time-series data frame


tigercosmos:
Related to the time-series line of work: discussion #380 (time series analysis), #435 (data frame ordering helpers), and #685 (estimate the position and attitude of NASA flight using Kalman Filter). Every branch and pull request in this series references this issue.

`solvcon.track.DataFrame` can put one source in time order, but it cannot put two sources on the same time base. That gap was named in discussion #380 in October 2024, deferred on purpose, and never picked up. It is now in the way of the remaining work in #685.

The fix is small. `DataFrame.sort()` is already `argsort()` followed by `take_along_axis()`. Resampling one source onto another's timestamps is `searchsorted()` followed by `take_along_axis()`. The same shape, one missing primitive.

## 1. How the gap got here

Discussion #380 opened in June 2024 to build time-series analysis in this code base from the ground up, with the NASA flight dataset chosen as the model problem.

In October 2024 the requirements for the container were written down in that discussion. One of them was merging two series whose index columns differ, which needs the two index columns merged into one and needs a rule for the timestamps where a column has no value. The prototype was scoped to a single source first, and that item was set aside.

In November 2024 #435 built the ordering half. `SimpleArray.sort()` and `argsort()` landed, and `DataFrame.sort()` (`solvcon/track/dataframe.py:110`) composes them with `take_along_axis()` (`:136`). A data frame can now be put in index order. It still cannot relate two of them.

In March 2026 #685 made the model problem concrete: estimate a flight's position and attitude from IMU and lidar measurements and check the result against ground truth, three sources at three different sampling rates. Its second item, preprocessing the data, carries the constraint directly: "we should not align data with linear interpolation." The reason is specific to measurements entering the filter. A measurement placed at an instant by interpolating between the samples around it carries information from after that instant, and an estimate built from it is no longer a causal estimate.

PR #696 landed the loader that month. `NasaDataset._rebuild_timeline()` (`solvcon/track/dataset.py:231`) merges the three sources into one timestamp-ordered event stream with a Python `dict` and `sorted()`, one object per row, 120,791 of them for Flight 1. The union of the three index columns is computed there by hand, in Python, because there was nothing to call.

[...]

PR #1252, opened 2026-07-31
Add `SimpleArray::searchsorted()`

tigercosmos:
`SimpleArray::searchsorted()` returns the positions that keep a sorted receiver sorted, as `numpy.searchsorted` does. It returns a new array and does not change the receiver.

Issue #1285 needs this method to look up time-series samples. The zeroth-order-hold index is `searchsorted(side="right") - 1`, and a window is two calls plus a view. Related to #1250.

## Why this search

A binary search over the whole receiver, once per value, is always correct. The target workload allows something better. A resample walks a sorted time grid, so the operand values rise, and both bounds rise with them. The previous result is then a lower bound for the next search.

Three forms use that fact. This branch implements one of them. The other two exist only as microbenchmarks, and the numbers below say why they lost.

All numbers come from an Apple M-series machine, with macOS, Apple clang, `-O3`, and libc++. Standalone microbenchmarks reproduce each code path; they do not call the extension. Each number is the best of N runs, and the runs alternate between the forms. Treat a difference below 10% as noise. Every form returns identical indices, so only the cost differs.

### Gallop forward, do not narrow the range

A rejected form keeps the binary search and moves its left edge to the previous result. It is slower than no optimization at all, by 1.4x to 1.9x.

| receiver / operand, rising | full search | rejected: narrow to `[prev, end)` | this branch: gallop from prev |
| --- | ---: | ---: | ---: |
| 1e6 / 1e6 | 26.2 ms | 46.8 ms | **6.2 ms** |
| 1e7 / 1e7 | 389.5 ms | 547.2 ms | **63.0 ms** |
| 5e7 / 5e6 | 240.6 ms | 336.9 ms | **58.6 ms** |
| 1e5 / 2e7 | 413.3 ms | 777.2 ms | **30.2 ms** |
| 1e7 / 1e7, `float64` | 515.8 ms | 829.9 ms | **82.0 ms** |

The range cannot pay for itself. It falls only from `n` to `n - i`, and the mean comparison count `log2(n!)/n` is about `log2(n) - 1.44`. The narrower range therefore saves about 1.5 probes out of 20. Against that gain, each search must wait for the previous result, and each search starts at an address that the processor cannot predict. A fixed range instead re-reads the same hot cache lines for every value.

A gallop from the previous result wins because it changes the exponent, not the constant. It doubles the probe forward before it searches the bracket that the probe passes, so the cost follows the distance travelled. That cost is O(log d), not O(log n), and a resample moves a short distance each time. The gain is 4.1x to 13.7x over a full search.

### Track the run inside the loop, not before it

A rejected form calls `std::is_sorted` on the operand before the search, then picks one loop for the whole array. That pass reads the operand a second time, and it spends that time on the path that it must make faster. One out-of-order value also costs the whole operand its fast path. No call to `std::is_sorted` remains in the diff.

| `float64`, receiver 1e6 | rejected: `is_sorted` before the loop | this branch: run tracked in loop |
| --- | ---: | ---: |
| 1e6 values, sorted | 14.0 ms | **13.0 ms** |
| 4e6 values, sorted | 14.3 ms | **12.9 ms** |
| 1e6 values, one drop at 90% | 34.4 ms | **10.7 ms** |
| 4e6 values, one drop at 90% | 132.2 ms | **27.4 ms** |

The gain on a sorted operand is small, at 1.08x. The gain on a nearly sorted operand reaches 3.2x to 4.8x. The chosen form keeps the fast path for every value before the drop.

### Leave the forward path for good

A rejected form leaves the forward path at a drop, then returns to it as soon as the values rise again. Half of the pairs in a random operand rise, so this form gallops between unrelated positions and reaches its worst case.

| receiver / operand, random | this branch: full search after the first drop | rejected: return to the fast path |
| --- | ---: | ---: |
| 1e6 / 1e6 | **51.7 ms** | 97.0 ms |
| 1e7 / 1e7 | **2049 ms** | 6885 ms |
| 5e7 / 5e6 | **1848 ms** | 5743 ms |
| 1e5 / 2e7 | **742 ms** | 1240 ms |
| 1e7 / 1e7, `float64` | **1804 ms** | 5958 ms |

The chosen form therefore ends the run for good at the first drop. One out-of-order value costs the values after it, and nothing more.

## Tests

`tests/test_buffer.py` compares `searchsorted` against `numpy.searchsorted` across all thirteen dtypes, over random input with ties. The tests also cover these cases:

- an empty receiver, and an empty operand
- values outside the receiver range
- exact hits, on both sides
- a NaN in the receiver, and a NaN in the operand
- a complex NaN in either component
- a ghosted receiver
- an operand that drops out of order at the head, and near the end

The full module passes with 223 tests. `make gtest` passes with 238 tests. `make lint` and `clang-tidy` on the changed lines report nothing.

## Known limitation

These members read `begin()` to `end()`, so they use storage order and ignore `stride()`. A strided view is therefore searched as the wrong span. The limitation covers the whole sort mixin and predates this branch. A `TODO` in the mixin records it.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -108,6 +108,27 @@ inline ssize_t buffer_offset(small_vector<ssize_t> const & stride, small_vector<
     return offset;
 }
 
+/// Which end of a run of equal values a search lands on.
+enum class SearchSide : uint8_t
+{
+    Left = 0, ///< The first position where the value may be inserted.
+    Right = 1, ///< The position after the last equal element.
+}; /* end enum class SearchSide */
+
+inline SearchSide search_side_from_string(std::string const & side, char const * op)
+{
+    if (side == "left")
+    {
+        return SearchSide::Left;
+    }
+    if (side == "right")
+    {
+        return SearchSide::Right;
+    }
+    throw std::invalid_argument(
+        std::format(R"(SimpleArray::{}(): side must be "left" or "right" but got "{}")", op, side));
+}
+
 namespace detail
 {
 
@@ -1433,10 +1454,42 @@ struct NanAwareLess
     bool operator()(V const & lhs, V const & rhs) const { return nan_aware_less(lhs, rhs); }
 }; /* end struct NanAwareLess */
 
-// TODO: every member here walks begin() to end() and so reads the buffer in
-// storage order, ignoring stride(). A strided one-dimensional view is accepted
-// and then sorted as the wrong span. Reject a non-contiguous receiver the way
-// argmin() already does, in a change that covers the whole mixin at once.
+template <typename V>
+size_t search_bound(V const * first, V const * last, V const & value, SearchSide side)
+{
+    V const * const pos = (side == SearchSide::Left)
+                              ? std::lower_bound(first, last, value, NanAwareLess{})
+                              : std::upper_bound(first, last, value, NanAwareLess{});
+    return static_cast<size_t>(pos - first);
+}
+
+/**
+ * Answer the bound for a value that cannot precede the one answered at lo,
+ * doubling the probe forward from there so the cost follows the distance
+ * travelled rather than the size of the array.
+ */
+template <typename V>
+size_t gallop_bound(V const * first, V const * last, V const & value, SearchSide side, size_t lo)
+{
+    auto const before_bound = [&value, side](V const & probe)
+    { return (side == SearchSide::Left) ? nan_aware_less(probe, value) : !nan_aware_less(value, probe); };
+
+    auto const size = static_cast<size_t>(last - first);
+    size_t step = 1;
+    size_t hi = lo;
+    while (hi < size && before_bound(first[hi]))
+    {
+        lo = hi + 1;
+        hi += step;
+        step *= 2;
+    }
+
+    return lo + search_bound(first + lo, first + std::min(hi, size), value, side);
+}
+
+// TODO: every member here walks begin() to end(), ignoring stride(), so a
+// strided view is sorted or searched as the wrong span. Reject a
+// non-contiguous receiver the way argmin() does, across the whole mixin.
 template <typename A, typename T>
 class SimpleArrayMixinSort
 {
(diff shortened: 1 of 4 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1433,10 +1454,42 @@ struct NanAwareLess
...
+    return static_cast<size_t>(pos - first);
+}
+
+/**
+ * Answer the bound for a value that cannot precede the one answered at lo,
+ * doubling the probe forward from there so the cost follows the distance
+ * travelled rather than the size of the array.
+ */
+template <typename V>
+size_t gallop_bound(V const * first, V const * last, V const & value, SearchSide side, size_t lo)
```
tigercosmos, line 1472: Find the bound faster. PR body explains the reason.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1451,6 +1504,36 @@ class SimpleArrayMixinSort
...
+     *
+     * The receiver must already be sorted under the order that sort() uses.
+     * An unsorted receiver gives an unspecified position and does not raise.
+     *
+     * @param value The value to place.
+     * @param side Left gives the first such position, Right the position
+     *      after the last equal element.
+     * @return The position, counted from begin().
+     */
+    size_t searchsorted(value_type const & value, SearchSide side = SearchSide::Left) const;
```
tigercosmos, line 1519: `searchsorted` for single value.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1451,6 +1504,36 @@ class SimpleArrayMixinSort
...
+     *
+     * Each value is placed independently, so the operand needs no order. A
+     * rising run only searches faster, by stepping forward from the previous
+     * result.
+     *
+     * @param values The values to place, in one dimension and of this class.
+     * @param side As for the scalar overload.
+     * @return One position per value, in the operand's shape.
+     */
+    SimpleArray<uint64_t> searchsorted(A const & values, SearchSide side = SearchSide::Left) const;
```
tigercosmos, line 1535: `searchsorted` for values.

doc/source/compute/buffer/reduce.md
```diff
@@ -205,6 +206,29 @@ from the first ghost element, where a subscript counts from the first body
 element, so pass them to `take_along_axis()` rather than subscripting with
 them.
 
+### The `searchsorted` Method
```
tigercosmos, line 209: Doc for `searchsorted`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3089,6 +3172,47 @@ SimpleArray<uint64_t> detail::SimpleArrayMixinSort<A, T>::argsort()
...
+    values.validate_1d("searchsorted", "value array");
+
+    value_type const * const first = static_cast<A const *>(this)->begin();
+    value_type const * const last = static_cast<A const *>(this)->end();
+    SimpleArray<uint64_t> ret(values.shape());
+    value_type const * src = values.begin();
+    uint64_t * dst = ret.begin();
+
+    // Values keep rising until the first drop, and both bounds rise with
+    // them, so each search can gallop forward from the previous result.
```
tigercosmos, line 3196: Check PR body for more explanation of the decision choice of the implementation.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1433,10 +1454,42 @@ struct NanAwareLess
...
+    {
+        lo = hi + 1;
+        hi += step;
+        step *= 2;
+    }
+
+    return lo + search_bound(first + lo, first + std::min(hi, size), value, side);
+}
+
+// TODO: every member here walks begin() to end(), ignoring stride(), so a
```
tigercosmos, line 1490: TODO

PR #1253, opened 2026-07-31
Let the user configure the agent backend from the console

tigercosmos:
A backend now advertises its tunable knobs through settings_spec(), and the AgentBackend base stores and validates them. The Claude Code backend uses it for the model and the reasoning effort, each mapped to a CLI flag and left out on the shared "default" choice. Both lists are hard-coded, since the CLI cannot enumerate them, but every value was checked against the installed CLI.

The console grows a Settings button whose dialog is built from the selected backend's spec, so a backend with no knobs gets a placeholder instead of an empty form. The Claude selector label becomes "Claude Code", so SubprocessBackend now carries a name class attribute instead of a property.

Related to #1206.

### DIFF
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -325,6 +325,19 @@ def format_history(history, used=0):
     return HistoryFormatter.render(history, used)
 
 
+@dataclasses.dataclass(frozen=True)
+class BackendSetting:
+    """One user-tunable knob a backend advertises to a settings editor.  An
+    empty ``choices`` means free text; otherwise the value must be one of the
+    listed strings."""
+
+    name: str
+    label: str
+    choices: tuple = ()
+    default: str = ""
+    tooltip: str = ""
+
+
 @dataclasses.dataclass
 class BackendResponse:
     """One backend reply: ``text`` prose, the proposed ``commands`` the
@@ -342,6 +355,10 @@ class AgentBackend(abc.ABC):
     system instruction (:attr:`_INSTRUCTIONS`, sent as a real system prompt)
     and one user-payload layout (:meth:`_compose_user`), so a CLI and an HTTP
     backend never drift apart in what they ask the model.
+
+    A backend may also advertise user-tunable knobs through
+    :meth:`settings_spec`; the base class stores and validates their values, so
+    a settings editor works the same for every backend.
     """
 
     # TODO: solvcon is an application platform for geometry-based computation:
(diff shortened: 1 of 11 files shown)


solvcon/agent/_backend.py
```diff
@@ -486,6 +547,10 @@ class BackendRegistry:
 
     _BACKENDS = []
 
+    #: The configuration key holding every backend's settings, as a
+    #: ``backend name -> {knob: value}`` mapping.
+    CONFIG_KEY = "agent_backend_settings"
```
tigercosmos, line 552: The key for config.

solvcon/agent/_backend.py
```diff
@@ -517,6 +582,43 @@ def get(cls, name):
                 return backend
         return None
 
+    @classmethod
+    def load_settings(cls, config):
```
tigercosmos, line 586: Load settings.

solvcon/agent/_backend.py
```diff
@@ -517,6 +582,43 @@ def get(cls, name):
...
+            if not isinstance(values, dict):
+                continue
+            for name, value in values.items():
+                try:
+                    backend.set_setting(name, value)
+                except (KeyError, ValueError):
+                    continue
+
+    @classmethod
+    def save_settings(cls, config):
```
tigercosmos, line 608: Save settings.

solvcon/agent/_backends_impl.py
```diff
@@ -243,16 +247,37 @@ class ClaudeCliBackend(SubprocessBackend):
...
+    so :attr:`DEFAULT_CHOICE` leaves the flag out and the request runs on
+    whatever the CLI itself defaults to, which moves over time.
     """
 
     command = "claude"
+    name = "Claude Code"
+
+    DEFAULT_CHOICE = "default"
+
+    SETTINGS = (
```
tigercosmos, line 262: Settings for Claude Code.

solvcon/pilot/agent/_agent_settings.py
```diff
@@ -0,0 +1,85 @@
...
+
+from PySide6.QtWidgets import (QDialog, QDialogButtonBox, QComboBox, QLabel,
+                               QLineEdit, QFormLayout, QVBoxLayout)
+
+__all__ = [  # noqa: F822
+    'AgentBackendSettingsDialog',
+]
+
+
+class AgentBackendSettingsDialog(QDialog):
```
tigercosmos, line 19: New agent backend settings dialog.

PR #1254, opened 2026-08-01
CMake: Find PySide6 through `PYTHON_EXECUTABLE` instead of `PATH`

yungyuc:
Replace `python3` in `CMakeLists.txt` with `${PYTHON_EXECUTABLE}` so that the detection of Python related modules do not rely on `PATH`.

Also move from `sys.stdout.write` to `print` to read more plainly. `print` appends a newline that `execute_process` would otherwise capture into the path, so each call gains `OUTPUT_STRIP_TRAILING_WHITESPACE`, which also guards against stray whitespace.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,9 +241,11 @@ set(SOLVCON_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp" CACHE INTERNAL "")
 
 include_directories(${SOLVCON_INCLUDE_DIR})
 
+# Discovery has to use PYTHON_EXECUTABLE rather than a bare python3.
 execute_process(
-    COMMAND python3 -c "import sys, os, PySide6; sys.stdout.write(os.path.dirname(PySide6.__file__))"
+    COMMAND "${PYTHON_EXECUTABLE}" -c "import os, PySide6; print(os.path.dirname(PySide6.__file__))"
     OUTPUT_VARIABLE PYSIDE6_PYTHON_PACKAGE_PATH
+    OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 message(STATUS "PYSIDE6_PYTHON_PACKAGE_PATH: ${PYSIDE6_PYTHON_PACKAGE_PATH}")
 # Addressing an issue with the PySide6/Shiboken6 6.11 homebrew package:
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -241,9 +241,11 @@ set(SOLVCON_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp" CACHE INTERNAL "")
 
 include_directories(${SOLVCON_INCLUDE_DIR})
 
+# Discovery has to use PYTHON_EXECUTABLE rather than a bare python3.
 execute_process(
-    COMMAND python3 -c "import sys, os, PySide6; sys.stdout.write(os.path.dirname(PySide6.__file__))"
+    COMMAND "${PYTHON_EXECUTABLE}" -c "import os, PySide6; print(os.path.dirname(PySide6.__file__))"
```
yungyuc, line 246: The PySide6 and Shiboken6 discovery runs a bare `python3` taken from `PATH`. A desktop IDE inherits the session environment rather than the shell where the dependency prefix was activated, so that `python3` resolves to an interpreter without PySide6 installed. None of the `execute_process` calls captures a result code, so the failed imports are silent: every discovered path comes back empty, `PYSIDE6_LIBFILE` is empty, and `solvcon_primary` links against nothing. The build then fails much later with an undefined `PySide::getWrapperForQObject` referenced from `wrap_pilot.cpp.o`, which points at the wrapper code rather than at the missing environment that caused it.

The five discovery calls now run `${PYTHON_EXECUTABLE}`. That variable is resolved earlier in the same file and is set explicitly by the `scdv-rel` preset, so discovery no longer depends on the ambient `PATH` and behaves the same in a shell, in a desktop IDE, and in a headless build.

CMakeLists.txt
```diff
@@ -309,6 +315,15 @@ endif()
 message(STATUS "PYSIDE6_LIBFILE: ${PYSIDE6_LIBFILE}")
 message(STATUS "SHIBOKEN6_LIBFILE: ${SHIBOKEN6_LIBFILE}")
 
+# An empty PYSIDE6_LIBFILE links solvcon_primary against nothing and defers the
```
yungyuc, line 318: An empty `PYSIDE6_LIBFILE` is now a configure-time `FATAL_ERROR` when `BUILD_QT` is on, naming the interpreter that failed. The failure surfaces where the cause is instead of as an undefined symbol after a full compile.

ExplorerRay (commented):
LGTM.

PR #1256, opened 2026-08-01
Pack reused matmul matrices once

ThreeMonth03:
## Motivation

`matmul_planned()` sends C- and F-compatible matrices directly to CBLAS. A matrix layout that CBLAS cannot describe instead falls back to the generic strided executor, even when batch broadcasting reuses that matrix across multiple output coordinates. This PR optimizes that fallback by converting reusable unsupported layouts into a CBLAS-compatible representation once.

## Approach

`MatmulPlan` records which operands are reused by shape broadcasting. For sufficiently large float32 and float64 GEMM, `MatmulExecutor` packs each reused, CBLAS-incompatible operand once and sends every mapped contraction through the existing direct CBLAS route. Small matrices, vector products, explicit zero-stride views, and unsupported operands without broadcast reuse keep their existing routes.

## Profiling

The plots compare full-call times for NumPy 2.5.1, `matmul_planned()` before this PR, and the automatic pack-once route. They cover LHS, RHS, and cross-broadcast square GEMM with negative and step-two inner strides.

For `S >= 16`, the new route is faster at all 72 affected points: 3.30x-549.52x over the previous implementation and 1.11x-2.97x over NumPy. `S = 8` and `S = 12` remain on the generic route as controls.

### float32

pr1256-lhs-rhs-negative-float32

pr1256-cross-negative-float32

pr1256-lhs-rhs-step2-float32

pr1256-cross-step2-float32

### float64

pr1256-lhs-rhs-negative-float64

pr1256-cross-negative-float64

pr1256-lhs-rhs-step2-float64

pr1256-cross-step2-float64

<details>
<summary>Measurement details</summary>

- Environment: native macOS arm64, Python 3.14.6, NumPy 2.5.1, and Accelerate. BLAS-related thread limits were set to one.
- Workloads: float32 and float64 square GEMM with `S` in 8, 12, 16, 24, 32, 64, 128, and 256; LHS, RHS, and cross-broadcast reuse; and negative or step-two inner strides.
- Sampling: each method received a discarded preconditioning run, two warmup batches, and 15 timed batches in each of six balanced rounds. Each plotted value is the untrimmed median of 90 per-call estimates.
- Validation: all 96 comparison points passed correctness and sampling audits. The maximum relative Frobenius error was 1.32e-16.

</details>

<details>
<summary>Complete isolated benchmark output</summary>

[Download the 1,728-record JSONL raw data](https://github.com/user-attachments/files/30812808/pr1256-exact-head-profiling.jsonl.txt)

</details>

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -75,6 +75,10 @@ public:
 
     bool lhs_is_vector() const noexcept { return m_contraction.m_lhs_vector; }
     bool rhs_is_vector() const noexcept { return m_contraction.m_rhs_vector; }
+    bool lhs_is_broadcast() const noexcept { return m_batch.m_lhs_is_broadcast; }
+    bool rhs_is_broadcast() const noexcept { return m_batch.m_rhs_is_broadcast; }
+    bool lhs_has_zero_batch_stride() const noexcept { return m_batch.m_lhs_has_zero_stride; }
+    bool rhs_has_zero_batch_stride() const noexcept { return m_batch.m_rhs_has_zero_stride; }
     bool has_batch_axes() const noexcept { return m_batch.m_domain.rank() != 0; }
     MappedOffsetCursor batch_cursor() const & { return MappedOffsetCursor(m_batch.m_domain, m_batch.m_mappings); }
     MappedOffsetCursor batch_cursor() const && = delete;
@@ -100,10 +104,21 @@ private:
         ssize_t m_rhs_column_stride;
     }; /* end struct ContractionStrides */
 
+    struct BatchOperand
+    {
+        OperandMapping m_mapping;
+        bool m_is_broadcast = false;
+        bool m_has_zero_stride = false;
+    }; /* end struct BatchOperand */
+
     struct BatchMappings
     {
         LoopDomain m_domain;
         mapping_type m_mappings;
+        bool m_lhs_is_broadcast;
+        bool m_rhs_is_broadcast;
+        bool m_lhs_has_zero_stride;
+        bool m_rhs_has_zero_stride;
     }; /* end struct BatchMappings */
 
     MatmulPlan(shape_type output_shape, Contraction contraction, ContractionStrides strides, BatchMappings batch);
@@ -124,7 +139,7 @@ private:
     static shape_type make_batch_shape(Array const & lhs, Array const & rhs);
 
     template <typename Array>
-    static OperandMapping make_batch_mapping(Array const & operand, LoopDomain const & domain);
+    static BatchOperand make_batch_operand(Array const & operand, LoopDomain const & domain);
 
     static shape_type make_output_shape(BatchMappings const & batch, Contraction const & contraction);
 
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -142,23 +157,31 @@ class MatmulPlan
...
  */
 template <typename Array>
 class MatmulExecutor
 {
 public:
-    MatmulExecutor(MatmulPlan const & plan, Array & output, Array const & lhs, Array const & rhs);
+    MatmulExecutor(MatmulPlan plan, Array & output, Array const & lhs, Array const & rhs);
+    ~MatmulExecutor() = default;
+
+    MatmulExecutor(MatmulExecutor const &) = delete;
```
ThreeMonth03, line 181: `MatmulExecutor` owns optional packed arrays and caches raw pointers that may point into them. It therefore explicitly follows the Rule of Five: destruction is defaulted, while copy/move construction and assignment are deleted. This keeps the executor non-transferable and prevents stale pointers after copying or moving.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -431,6 +499,69 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan const & plan, Array & output, A
...
+    Packing packing;
+#if (defined(__APPLE__) && defined(__arm64__)) || defined(MM_HAS_CBLAS)
+    if constexpr (can_matmul_blas_v<value_type>)
+    {
+        if (m_plan.lhs_is_vector() || m_plan.rhs_is_vector() ||
+            std::min({m_plan.rows(), m_plan.columns(), m_plan.inner_size()}) < BLAS_GEMM_PACK_MIN_DIMENSION)
+        {
+            return packing;
+        }
+        if (!m_plan.lhs_is_broadcast() && !m_plan.rhs_is_broadcast())
```
ThreeMonth03, line 525: This early return defines the scope of pack-once: only shape broadcasting provides reusable matrix cores. Unsupported equal-batch matrices need streamed packing per contraction, so they remain generic in this PR.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -431,6 +499,69 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan const & plan, Array & output, A
...
+        m_packed_lhs.emplace(m_lhs.to_row_major());
+    }
+    if (packing.m_pack_rhs)
+    {
+        m_packed_rhs.emplace(m_rhs.to_row_major());
+    }
+
+    Array const & lhs = m_packed_lhs ? *m_packed_lhs : m_lhs;
+    Array const & rhs = m_packed_rhs ? *m_packed_rhs : m_rhs;
+    m_plan = MatmulPlan::make(lhs, rhs);
```
ThreeMonth03, line 558: `pack()` runs before `execute_contractions()`. Each selected operand is materialized by one `to_row_major()` call, and the packed array is reused at every batch coordinate. Rebuilding the plan updates the matrix and batch mappings for the new physical strides, so the contraction loop needs no packing-specific offset code.

tests/test_matrix.py
```diff
@@ -437,21 +437,32 @@ def test_broadcast_batch_strides(self):
...
-        rhs_data = np.arange(40, dtype=dtype).reshape(1, 5, 4, 2)
-        cases = itertools.product(
-            self.make_matrix_stride_cases(lhs_data, 3),
-            self.make_matrix_stride_cases(rhs_data, 2),
-        )
-        for (lhs_case, case_lhs), (rhs_case, case_rhs) in cases:
-            lhs = self.SimpleArray(array=case_lhs)
-            rhs = self.SimpleArray(array=case_rhs)
-            expected = np.matmul(case_lhs, case_rhs)
+        for m, k, n in ((3, 4, 2), (17, 18, 19)):
```
ThreeMonth03, line 442: These sizes cover both paths: `(3,4,2)` remains generic, while `(17,18,19)` crosses the minimum packing dimension. The layout product then exercises direct, one-sided packing, and two-sided packing combinations.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -174,17 +197,31 @@ class MatmulExecutor
         Rhs,
     };
 
+    struct Packing
```
yungyuc: IIUC, `Packing` means more like `PackingState`.  I am OK with either name.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -174,17 +197,31 @@ class MatmulExecutor
...
+    {
+        bool m_pack_lhs = false;
+        bool m_pack_rhs = false;
```
yungyuc: It's OK to name the member data without `m_` prefix in public POD class without accessors:

```suggestion
        bool lhs = false;
        bool rhs = false;
```

But it's also OK to use `m_` prefix.  Either way works.

yungyuc (approved):
LGTM.  It's great the new code becomes faster than numpy.

I assume no profiling code needs to be changed.  Let me know if you want me to merge as is.

ThreeMonth03:
Hi, @yungyuc. I've modified `PackingState` and `m_` prefix. Please merge this pull request.

PR #1259, opened 2026-08-01
Tune batched vector matmul routes

ThreeMonth03:
`matmul_planned()` already dispatches BLAS-compatible matrix-matrix products, but batched vector-matrix `(K,) @ (B,K,N)` and matrix-vector `(B,M,K) @ (K,)` products still use the generic signed-stride loop. The same vector is reused at every batch coordinate, so these calls repeat scalar traversal even when each matrix can be passed directly to CBLAS.

`MatmulExecutor` now records one `MatmulSelection` before batch traversal. It selects direct BLAS for a C- or F-compatible matrix and a positive vector stride, packs a reused negative-stride vector once before BLAS when the copy is amortized, and otherwise keeps generic execution. `MatmulTuning` stores the measured boundaries, while `MatmulPlan` remains independent of backend policy.

## Profiling

The benchmark compares head `7b40cb46`, its parent `f5824e1a`, and NumPy 2.5.1 on an Apple M1 using one Accelerate thread. All 112 points whose route changes are faster than the parent, with a median 5.03x speedup and a range of 1.15x to 31.54x.

All 32 negative-stride points that pack the reused vector once are faster than NumPy, with a median 6.24x speedup. All 80 positive-stride direct-BLAS points are faster than the parent; 16 are faster than NumPy, and their median runtime is 1.06 times NumPy.

### float32

pr1259-c-matrix-stride1-float32

pr1259-c-matrix-stride2-float32

pr1259-c-matrix-negative-float32

pr1259-f-matrix-stride1-float32

pr1259-f-matrix-stride2-float32

pr1259-f-matrix-negative-float32

### float64

pr1259-c-matrix-stride1-float64

pr1259-c-matrix-stride2-float64

pr1259-c-matrix-negative-float64

pr1259-f-matrix-stride1-float64

pr1259-f-matrix-stride2-float64

pr1259-f-matrix-negative-float64

<details>
<summary>Measurement method and raw data</summary>

- The compared implementations run in separate persistent processes. NumPy and solvcon both use Apple Accelerate with one BLAS thread.
- Workloads use batch size 4; `S` in 8, 12, 16, 24, 32, 64, 128, and 256; float32 and float64; C- and per-matrix F-compatible matrices; and vector strides 1, 2, and -1.
- Each point uses two warmups and five timed batches per round for 12 rounds. A timed batch is calibrated to about 15 ms, and the plotted value is the median of 60 per-call times.
- The 12 rounds use all six execution orders twice. Four `S=24`, vector-stride-1 points also received 12 confirmation rounds; their paired bootstrap 95% lower bounds are 1.044x to 1.239x.
- The maximum relative Frobenius error is `2.32e-7`.
- [Formal benchmark output: 6,912 JSONL records](https://github.com/user-attachments/files/30860020/pr1259-formal-profiling.jsonl.txt)
- [Targeted confirmation output: 144 JSONL records](https://github.com/user-attachments/files/30860041/pr1259-targeted-confirmation.jsonl.txt)

</details>

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -33,6 +33,87 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
                                           std::is_same_v<T, Complex<double>>;
 
 /**
+ * @brief Identify the contraction kernel fixed before batch traversal.
+ */
+enum class MatmulKernel : std::uint8_t
+{
+    Generic,
+    BlasDot,
+    BlasGevm,
+    BlasGemv,
+    BlasGemm,
+}; /* end enum class MatmulKernel */
+
+/**
+ * @brief Identify operands that must be materialized into row-major storage.
+ */
+struct PackingState
+{
+    bool lhs = false;
+    bool rhs = false;
+
+    explicit operator bool() const noexcept { return lhs || rhs; }
+}; /* end struct PackingState */
+
+/**
+ * @brief Group measured thresholds by matmul dispatch decision.
+ *
+ * A tuning table supplies the workload boundaries used to compare generic,
+ * direct BLAS, and packing routes. Keeping the values separate from dispatch
+ * code allows a backend or value type to provide another table.
+ *
+ * @note MatmulExecutor currently uses one compile-time table for every
+ * supported BLAS backend and value type.
+ */
+struct MatmulTuning
+{
+    /**
+     * @brief Select BLAS when the supplied operands have compatible layouts.
+     */
+    struct DirectBlas
+    {
+        ssize_t dot_min_length;
+        ssize_t compact_gevm_min_elements;
+        ssize_t gemv_min_dimension;
+        ssize_t gemm_min_dimension;
+    }; /* end struct DirectBlas */
+
+    /**
+     * @brief Pack reused matrices that cannot be described directly to BLAS.
+     */
+    struct MatrixPacking
+    {
+        ssize_t gemm_min_dimension;
+    }; /* end struct MatrixPacking */
+
+    /**
+     * @brief Select direct or packed BLAS for batched vector-matrix products.
+     */
+    struct BatchedVector
+    {
+        ssize_t direct_min_matrix_elements;
+        size_t always_pack_min_matrix_elements;
+        size_t pack_min_matrix_elements;
+        size_t pack_min_batch_size;
+        size_t reuse_min_matrix_elements;
+        size_t reuse_min_total_output_elements;
+    }; /* end struct BatchedVector */
+
+    DirectBlas direct_blas;
+    MatrixPacking matrix_packing;
+    BatchedVector batched_vector;
+}; /* end struct MatmulTuning */
+
+/**
+ * @brief Record the fixed kernel and packing for one matmul call.
+ */
+struct MatmulSelection
+{
+    MatmulKernel kernel = MatmulKernel::Generic;
+    PackingState packing;
+}; /* end struct MatmulSelection */
+
+/**
  * @brief Describe matmul operands as an execution-independent contraction.
  *
  * MatmulPlan interprets trailing axes as vector or matrix roles and leading
@@ -44,17 +125,12 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
  *
  * For `(2,1,3,4) @ (1,5,4,6)`, the plan records batch shape `(2,5)`, M=3,
  * N=6, K=4, output shape `(2,5,3,6)`, and zero-stride batch mappings for
- * the broadcast axes. It does not allocate the output or evaluate the ten
- * matrix pairs.
+ * the broadcast axes.
  *
  * For `(K,) @ (B,K,N)`, the plan keeps a synthetic M=1 for execution but
  * omits that axis from the `(B,N)` output. Matrix-vector products similarly
  * omit synthetic N=1, while vector-vector products retain SimpleArray's
  * existing `(1,)` dot-result convention.
- *
- * @note This implementation plans vector and matrix operand roles with
- * broadcast leading batch shapes. It records the batch domain and mappings,
- * M, N, K, the output shape, and signed contraction strides.
  */
 class MatmulPlan
 {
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -32,6 +32,87 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
...
+
+    DirectBlas direct_blas;
+    MatrixPacking matrix_packing;
+    BatchedVector batched_vector;
+}; /* end struct MatmulTuning */
+
+/**
+ * @brief Record the fixed kernel and packing for one matmul call.
+ */
+struct MatmulSelection
```
ThreeMonth03, line 110: This object separates dispatch policy from batch traversal. It records one kernel and any one-time packing for the full matmul call, so batch coordinates do not repeat threshold or route selection.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -32,6 +32,87 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
...
+     */
+    struct MatrixPacking
+    {
+        ssize_t gemm_min_dimension;
+    }; /* end struct MatrixPacking */
+
+    /**
+     * @brief Select direct or packed BLAS for batched vector-matrix products.
+     */
+    struct BatchedVector
```
ThreeMonth03, line 92: These thresholds cover direct BLAS for positive-stride vectors, unconditional packing for large reused vectors, and reuse-dependent packing for smaller vectors. A later backend-specific table can replace the values without adding platform branches to the selector.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -500,45 +593,214 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan plan, Array & output, Array con
...
+
+template <typename Array>
+PackingState MatmulExecutor<Array>::select_vector_packing(PackingState required) const
+{
+    if (!required)
+    {
+        return PackingState{};
+    }
+    bool const vector_reused = required.lhs ? m_plan.lhs_is_broadcast() : m_plan.rhs_is_broadcast();
+    return vector_reused && should_pack_vector() ? required : PackingState{};
```
ThreeMonth03, line 779: `select_vector_packing()` permits packing only when broadcasting reuses the same vector. Distinct vectors across the batch stay generic because one packed buffer cannot serve every coordinate.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -500,45 +593,214 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan plan, Array & output, Array con
...
 void MatmulExecutor<Array>::execute()
 {
-    PackingState const packing = select_packing();
+    MatmulSelection const selection = select_execution();
 
-    if (packing)
+    if (selection.packing)
     {
-        pack(packing);
+        pack(selection.packing);
```
ThreeMonth03, line 600: Packing is applied once before kernel dispatch. `pack()` rebuilds the plan and data pointers, so every batch coordinate then observes the packed storage.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -561,67 +823,69 @@ void MatmulExecutor<Array>::pack(PackingState const & packing)
...
 void MatmulExecutor<Array>::execute_contractions()
 {
     if (!m_plan.has_batch_axes())
     {
-        execute_at(0, 0, 0);
+        execute_at<Kernel>(0, 0, 0);
         return;
     }
 
     for (MappedOffsetCursor cursor = m_plan.batch_cursor(); cursor; cursor.advance())
```
ThreeMonth03, line 835: `Kernel` is fixed before this loop. The cursor advances batch offsets without repeating threshold or packing decisions for each matrix-vector product.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -32,6 +32,87 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
                                           std::is_same_v<T, Complex<float>> ||
                                           std::is_same_v<T, Complex<double>>;
 
+/**
+ * @brief Identify the contraction kernel fixed before batch traversal.
+ */
+enum class MatmulKernel : std::uint8_t
```
yungyuc, line 38: I like this summary enum class.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -32,6 +32,87 @@ inline constexpr bool can_matmul_blas_v = std::is_same_v<T, float> ||
...
+    BlasDot,
+    BlasGevm,
+    BlasGemv,
+    BlasGemm,
+}; /* end enum class MatmulKernel */
+
+/**
+ * @brief Identify operands that must be materialized into row-major storage.
+ */
+struct PackingState
```
yungyuc, line 50: Good to take this outside the inner.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -197,29 +273,46 @@ class MatmulExecutor
...
-
-        explicit operator bool() const noexcept { return lhs || rhs; }
-    }; /* end struct PackingState */
-
-    static constexpr ssize_t BLAS_DOT_MIN_LENGTH = 128;
-    static constexpr ssize_t BLAS_COMPACT_GEVM_MIN_ELEMENTS = 729;
-    static constexpr ssize_t BLAS_GEMV_MIN_DIMENSION = 32;
-    static constexpr ssize_t BLAS_GEMM_MIN_DIMENSION = 8;
-    static constexpr ssize_t BLAS_GEMM_PACK_MIN_DIMENSION = 16;
+    static constexpr MatmulTuning TUNING{
```
yungyuc: This is a smart way to create constants with hierarchy.  Learnt.

Better to keep the space before `{`:
```suggestion
    static constexpr MatmulTuning TUNING {
```

yungyuc (approved):
LGTM.  Let me know if you want it to be merged immediately.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -197,29 +273,46 @@ class MatmulExecutor
...
+        },
+        .batched_vector = {
+            .direct_min_matrix_elements = 512,
+            .always_pack_min_matrix_elements = 4096,
+            .pack_min_matrix_elements = 1024,
+            .pack_min_batch_size = 4,
+            .reuse_min_matrix_elements = 576,
+            .reuse_min_total_output_elements = 128,
+        },
+    };
```
ThreeMonth03, line 294: The following code will fail in clang-format 20, so I afterwards assign bracelet with assignment operator.

```c++
static constexpr MatmulTuning TUNING {
```

ThreeMonth03 (commented):
@yungyuc I've modified code according to the suggestion. You could merge this pull request after CI runner returns passsed.

yungyuc:
@KHLee529 Please take a look.

yungyuc:
@ThreeMonth03 Please help fix the linting failure: https://github.com/solvcon/solvcon/actions/runs/31297887466

PR #1260, opened 2026-08-01
Say what a backend reply is: transport outcome and parse status

tigercosmos:
A `BackendResponse` carried prose, commands, and an untyped `error`, which leaves the two questions a caller has to answer unanswerable. Whether the model replied at all was guesswork over error strings, and `ToolCallParser.parse` mapped an explicit `[]` and plain prose onto the same empty list, so "the request is done" and "the model talked instead of acting" were the same value.

The reply now names both. `TransportOutcome` says how the exchange ended (`ok`, `transport`, `timeout`, `cancelled`) and `ParseStatus` says what the text turned out to be (`commands`, `empty`, `prose`, `malformed`); both backends map their failure modes onto the first, and `parse_reply` returns the second beside the commands. A response that fills only the older fields still reports a usable status, so nothing existing has to change at once.

Cancellation becomes honest as part of this. `CancellableBackend` holds the flag both backends share, a cancel that lands before the child process or the socket is reachable is acted on once it is, and a call that was stopped but answered anyway returns `cancelled` with no commands rather than applying work the user asked to abandon.

Op names leave the parser. Rejecting an unadvertised op while parsing threw away every command that arrived with it; the runner already answers unknown ops with a per-command error, so one bad entry no longer voids its batch. What the session hides from the tool surface it now refuses at execution instead, and injecting a renderer stops hiding `render_png`, which is the op that renderer is for.

This is the first of two stacked changes: the turn loop that reads these statuses follows once this lands.

Related to #1136, #966.

### DIFF
--- a/solvcon/agent/_backend.py
+++ b/solvcon/agent/_backend.py
@@ -15,6 +15,7 @@ the registry: it is a test and demo double, not a backend to offer a user.
 
 import abc
 import dataclasses
+import enum
 import json
 
 from . import _command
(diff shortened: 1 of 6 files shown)


solvcon/agent/_backend.py
```diff
@@ -338,14 +339,56 @@ class BackendSetting:
     tooltip: str = ""
 
 
+class TransportOutcome(enum.Enum):
```
tigercosmos, line 342: Transport state.

solvcon/agent/_backend.py
```diff
@@ -338,14 +339,56 @@ class BackendSetting:
...
+    arrived, which is what tells a loop to abort instead of retry.
+    """
+
+    OK = "ok"
+    TRANSPORT = "transport"
+    TIMEOUT = "timeout"
+    CANCELLED = "cancelled"
+
+
+class ParseStatus(enum.Enum):
```
tigercosmos, line 355: Parsing state.

solvcon/agent/_backends_impl.py
```diff
@@ -119,13 +125,77 @@ def parse(cls, text, tool_surface=None):
...
+            return ParsedReply(status=status.PROSE)
+        try:
+            commands = cls.commands_of(data)
+        except ValueError as exc:
+            return ParsedReply(status=status.MALFORMED, error=str(exc))
+        return ParsedReply(
+            commands, status.COMMANDS if commands else status.EMPTY)
+
+
+class CancellableBackend:
```
tigercosmos, line 166: New base backend for handling canceling.

solvcon/agent/_backends_impl.py
```diff
@@ -119,13 +125,77 @@ def parse(cls, text, tool_surface=None):
...
+        tears down nothing, so the call can still succeed.  That answer is
+        unwanted: returning it would let commands land after the user asked
+        for none.
+        """
+        if not self._cancelled:
+            return None
+        return self.failure("cancelled")
+
+
+class SubprocessBackend(CancellableBackend, _backend.AgentBackend):
```
tigercosmos, line 198: All backends inherit `CancellableBackend`

solvcon/agent/_core.py
```diff
@@ -204,33 +211,26 @@ def _execute(self, commands):
...
-                   if _command.op_of(command) not in blocked]
-        if not allowed:
-            return [_OutcomeStub(
-                _command.op_of(command),
-                error="destructive command %r is disabled for this session"
-                % _command.op_of(command)) for command in commands]
+        blocked = self._blocked_ops()
+        gated = [_command.op_of(command) in blocked for command in commands]
+        if all(gated):
+            return [self._blocked_result(command) for command in commands]
```
tigercosmos, line 223: If all commands are blocked, return immediately. Otherwise, try to go to the runner first.

Issue #807, opened 2026-05-23
Add architecture fence in mmdv building scripts


yungyuc:
mmdv build scripts support only selected architecture.  For example (and currently the only one), [`contrib/dependency/ubuntu2404/build-mmdv-ubuntu24.sh`](https://github.com/solvcon/modmesh/blob/2cbc729a26cff2ad22c95f9e289bfdc0532f0658/contrib/dependency/ubuntu2404/build-mmdv-ubuntu24.sh) supports only x86-64.

The scripts should use a fence for the supported architectures.

_Originally posted by @tigercosmos in https://github.com/solvcon/modmesh/pull/805#pullrequestreview-4350446015_

OuChiaYun:
I interest to do.

PR #1264, opened 2026-08-02
Fence dependency builds by platform

OuChiaYun:
Reject unsupported OS and architecture combinations before starting the dependency build. Allow Ubuntu on 64-bit x86 and macOS on Apple Silicon.

Related to #807

### DIFF
--- a/contrib/dependency/build-scdv.sh
+++ b/contrib/dependency/build-scdv.sh
@@ -947,7 +947,22 @@ elif [ "${SCDV_ALLOW_UNSUPPORTED_PLATFORM}" = "1" ] ; then
        "'${SCDV_OS}/${SCDV_ARCH}'; continuing anyway." >&2
 else
   case "${SCDV_OS}:${SCDV_ARCH}" in
-    ubuntu:x86_64|macos:arm64)
+    macos:arm64)
+      ;;
+
+    ubuntu:x86_64)
+      # SCDV_OS=ubuntu covers every Linux (uname -s is just "Linux"), so
+      # confirm this is actually Ubuntu; other distros lack apt and the
+      # Debian/Ubuntu paths the recipes assume. Warn and proceed.
+      _scdv_id=""
+      if [ -r /etc/os-release ] ; then
+        _scdv_id=$(set +e ; . /etc/os-release 2>/dev/null ; printf '%s' "${ID:-}")
+      fi
+      if [ "${_scdv_id}" != "ubuntu" ] ; then
+        echo "warning: non-Ubuntu Linux (os-release ID='${_scdv_id:-unknown}');" \
+             ""the build assumes apt and Ubuntu-specific paths and may fail."" >&2
+      fi
+      unset _scdv_id
       ;;
 
     ubuntu:aarch64)
@@ -958,18 +973,6 @@ else
     macos:x86_64)
       # Intel Mac: some dependencies may not build. (e.g. numpy/scipy)
       echo "warning: some dependencies may not build correctly on macos/x86_64." >&2
-
-      if ! { : </dev/tty ; } 2>/dev/null ; then
-        echo "no controlling tty to confirm; rerun on a supported" \
-             "OS/architecture." >&2
-        exit 1
-      fi
-
-      read -r -p "Continue anyway? [y/N] " _ans </dev/tty
-      case "${_ans}" in
-        [Yy]*) ;;
-        *) echo "aborted." >&2 ; exit 1 ;;
-      esac
       ;;
 
     *)
--- a/contrib/dependency/build-scdv.sh
+++ b/contrib/dependency/build-scdv.sh
@@ -45,6 +45,8 @@
 #       ends with the two toolchains no build section installs: LaTeX, which
 #       the documentation needs to render its figures, then clang-format and
 #       clang-tidy for `make lint`, both from LLVM 22.
+#   ./build-scdv.sh --allow-unsupported-platform
+#       Proceed on an unsupported OS/architecture for platform testing.
 #
 # Overridable variables (search below for defaults):
 #   Platform:
@@ -861,6 +863,7 @@ SCDV_PRINT_PREFIX_ONLY=0
 SCDV_PRINT_DEPS_ONLY=0
 SCDV_NO_CONFIRM=0
 SCDV_ALLOW_WRITE_TO_ACTIVE_ENV=0
+SCDV_ALLOW_UNSUPPORTED_PLATFORM=${SCDV_ALLOW_UNSUPPORTED_PLATFORM:-0}
 SCDV_SKIP_LIST=""
 SCDV_KNOWN_PKGS="zlib openssl sqlite python pybind11 cython numpy scipy qt pyside6"
 
@@ -901,6 +904,9 @@ while [ $# -gt 0 ] ; do
     --allow-write-to-active-env)
       SCDV_ALLOW_WRITE_TO_ACTIVE_ENV=1
       ;;
+    --allow-unsupported-platform)
+      SCDV_ALLOW_UNSUPPORTED_PLATFORM=1
+      ;;
     --skip)
       shift
       if [ $# -eq 0 ] ; then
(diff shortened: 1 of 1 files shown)


contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +931,33 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
   exit 0
 fi
 
+# The per-OS blocks assume x86_64 on Ubuntu and arm64 on macOS.
+SCDV_ARCH=$(uname -m)
+
+case "${SCDV_OS}:${SCDV_ARCH}" in
+  ubuntu:x86_64|macos:arm64)
```
yungyuc: `ubuntu:arm64` should pass as well.

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +931,33 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
+      exit 1
+    fi
+    read -r -p "Continue anyway? [y/N] " _ans </dev/tty
+    case "${_ans}" in
+      [Yy]*) ;;
+      *) echo "aborted." >&2 ; exit 1 ;;
+    esac
+    ;;
+  *)
+    echo "unsupported OS/architecture '${SCDV_OS}/${SCDV_ARCH}'" >&2
```
yungyuc: Provide a command-line argument and an environment variable to override the non-support fencing to make it easy to test the script for new platform setup.

yungyuc (changes requested):
- [ ] `ubuntu:arm64` should be allowed.  It is also a popular setup.
- [ ] Provide a command-line argument and an environment variable to override the fencing.

yungyuc:
@ExplorerRay @ThreeMonth03 please help take a look.

contrib/dependency/build-scdv.sh
OuChiaYun: may i ask, i see L280-284 is hardcoded to x86_64:
```
library_dirs = /usr/lib/x86_64-linux-gnu:${SCDV_USRDIR}/lib
include_dirs = /usr/include/x86_64-linux-gnu/openblas-pthread:/usr/include/openblas:${SCDV_USRDIR}/include
runtime_library_dirs = /usr/lib/x86_64-linux-gnu
```

and the numpy dispatch is x86-only too:
```
with_log install.log "${PY}" -m pip install . --no-build-isolation \
  --config-settings="setup-args=-Dcpu-dispatch=SSE3 SSSE3 SSE41 POPCNT SSE42 AVX F16C FMA3 AVX2"
```
[code mean](https://numpy.org/doc/2.3/reference/simd/build-options.html#on-x86)

should I still add a soft fence for ubuntu:arm64, or is no warning needed?

yungyuc: > may i ask, i see L280-284 is hardcoded to x86_64:
> [deleted]
> should I still add a soft fence for ubuntu:arm64, or is no warning needed?

Good catch.  Still we should allow `ubuntu:arm64` to run.  Ubuntu is a popular OS and it is very likely that we may need to install scdv on it.  For example, a Ubuntu VM guest on a M1 Mac will be `ubuntu:arm64`.  Even if it fails with [`build-scdv.sh:280-284`](https://github.com/solvcon/solvcon/blob/a4142d7d00e222dca33b28ae19937fe91fd8a5b5/contrib/dependency/build-scdv.sh#L280-L284), we will see what happens instead of a blunt fencing message.

But you are right, a warning will be a good addition.

contrib/dependency/build-scdv.sh
OuChiaYun: fixed

contrib/dependency/build-scdv.sh
OuChiaYun: fixed

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,60 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
+
+if [ "${SCDV_ALLOW_UNSUPPORTED_PLATFORM}" = "1" ] ; then
+  echo "warning: platform support check overridden for" \
+       "'${SCDV_OS}/${SCDV_ARCH}'; continuing anyway." >&2
+else
+  case "${SCDV_OS}:${SCDV_ARCH}" in
+    ubuntu:x86_64|macos:arm64)
+      ;;
+
+    ubuntu:arm64)
```
ThreeMonth03: [P1] Match the architecture name Linux actually reports. `uname -m` returns `aarch64` on Ubuntu ARM64, so the default `SCDV_ARCH` never reaches this `ubuntu:arm64` branch and falls into the hard-failure case instead. I reproduced this with `SCDV_OS=ubuntu SCDV_ARCH=aarch64 --print-prefix`, which exits 1. Please recognize or normalize `aarch64`.

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,60 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
+      echo "warning: some build settings currently assume x86_64 and" \
+           "may not work correctly on ubuntu/arm64." >&2
+
+      if ! { : </dev/tty ; } 2>/dev/null ; then
```
ThreeMonth03: [P1] Let Ubuntu ARM continue after the warning. The maintainer clarification explicitly asks to allow this platform to run, but a non-interactive invocation has no `/dev/tty` and exits here before any build starts; `--no-confirm` does not bypass this separate prompt. Please keep the warning and proceed for Ubuntu ARM.

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,60 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
   exit 0
 fi
 
+# The per-OS blocks assume x86_64 on Ubuntu and arm64 on macOS.
+SCDV_ARCH=${SCDV_ARCH:-$(uname -m)}
+
+if [ "${SCDV_ALLOW_UNSUPPORTED_PLATFORM}" = "1" ] ; then
```
ThreeMonth03: [P2] Keep `--print-prefix` outside the platform fence. It is documented as a no-side-effect query that is safe to capture, but this block now fails or prompts before the existing print-prefix early exit on fenced platforms. Please exempt this query or handle it before fencing.

ThreeMonth03 (changes requested):
Ubuntu ARM remains fenced in two ways: Linux reports `aarch64`, which hits the hard-failure branch, and the `arm64` branch requires a controlling TTY instead of warning and proceeding. The fence also blocks the documented side-effect-free `--print-prefix` query. The CLI and environment overrides work as intended. I verified the behavior matrix with `--print-prefix`; `bash -n` and `git diff --check` pass.

contrib/dependency/build-scdv.sh
yungyuc: Oh you are right!  My bad.  It should have been `ubuntu:aarch64` instead of `arm64`, which does not belong to Linux.

contrib/dependency/build-scdv.sh
yungyuc: @OuChiaYun The confirmation prompt causes more trouble than we thought.  Let's get rid of the confirmation prompt and leave only the warning.

It's OK to remove the confirmation prompt here because there is an existing confirmation prompt at [`build-scdv.sh:L1048-L1057`](https://github.com/solvcon/solvcon/blob/4c5eb25ef0f0f1f59beccc8bd901de2ba2152694/contrib/dependency/build-scdv.sh#L1048-L1057):

```sh
# Final confirmation before any package is built; --no-confirm skips it for
# non-interactive runs.
if [ "${SCDV_NO_CONFIRM}" != "1" ] ; then
  # `test -r /dev/tty` returns true even with no controlling terminal, so try
  # to actually open it; that fails with ENXIO when there is no controlling tty
  # (e.g. under setsid or `< /dev/null` in CI).
  if ! { : </dev/tty ; } 2>/dev/null ; then
    echo "no controlling tty for confirmation prompt; rerun with --no-confirm" >&2
    exit 2
  fi
  read -r -p "Press Enter to start the build, Ctrl-C to abort: " _ </dev/tty
fi
```

yungyuc (changes requested):

yungyuc:
The points have not been addressed fully.  When it's ready, please leave a global comment to request for the next review.  Don't use the "request for review" button (see [`pull_request_template.md:L8-L11`](https://github.com/solvcon/solvcon/blob/4c5eb25ef0f0f1f59beccc8bd901de2ba2152694/.github/pull_request_template.md?plain=1#L8-L11)).

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,58 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
+       "'${SCDV_OS}/${SCDV_ARCH}'; continuing anyway." >&2
+else
+  case "${SCDV_OS}:${SCDV_ARCH}" in
+    ubuntu:x86_64|macos:arm64)
+      ;;
+
+    ubuntu:aarch64)
+      echo "warning: some build settings currently assume x86_64 and" \
+          "may not work correctly on ubuntu/aarch64." >&2
+      ;;
```
OuChiaYun, line 971: Ubuntu aarch64 now warns and continues to the shared build confirmation without a separate platform-specific prompt.

chestercheng:
This script treats all Linux x86_64 as `ubuntu:x86_64` because Linux is mapped to `SCDV_OS=ubuntu` earlier. But the later recipe has many Ubuntu-only assumptions, including apt package names and hardcoded paths.
So non-Ubuntu Linux, such as arch x86_64, still enters the supported branch.

If we want to fence real supported platforms, this still seems too loose.

contrib/dependency/build-scdv.sh
```diff
@@ -901,6 +904,9 @@ while [ $# -gt 0 ] ; do
     --allow-write-to-active-env)
       SCDV_ALLOW_WRITE_TO_ACTIVE_ENV=1
       ;;
+    --allow-unsupported-platform)
+      SCDV_ALLOW_UNSUPPORTED_PLATFORM=1
+      ;;
```
OuChiaYun, line 909: Added --allow-unsupported-platform and the corresponding environment override so unsupported platform combinations can still be tested explicitly.

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,53 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
 fi
 
+# The per-OS blocks assume x86_64 on Ubuntu and arm64 on macOS.
+SCDV_ARCH=${SCDV_ARCH:-$(uname -m)}
+
+if [ "${SCDV_PRINT_PREFIX_ONLY}" = "1" ] ; then
+  :
+elif [ "${SCDV_ALLOW_UNSUPPORTED_PLATFORM}" = "1" ] ; then
+  echo "warning: platform support check overridden for" \
+       "'${SCDV_OS}/${SCDV_ARCH}'; continuing anyway." >&2
```
OuChiaYun, line 947: --print-prefix is exempted from the platform fence so the documented side-effect-free query still works on unsupported platforms.

contrib/dependency/build-scdv.sh
```diff
@@ -931,6 +937,53 @@ if [ "${SCDV_PRINT_DEPS_ONLY}" = "1" ] ; then
...
+else
+  case "${SCDV_OS}:${SCDV_ARCH}" in
+    macos:arm64)
+      ;;
+
+    ubuntu:x86_64)
```
OuChiaYun, line 953: Linux is mapped to SCDV_OS=ubuntu earlier, so check /etc/os-release here to distinguish actual Ubuntu hosts from other Linux distributions. Non-Ubuntu x86_64 hosts now get an explicit warning before continuing.

contrib/dependency/build-scdv.sh
```diff
@@ -1519,4 +1572,4 @@ else
 
 echo "Set \${SCDVBUILD_ALL} or \${SCDVBUILD_QT} to build QT section"
 
-fi
+fi
```
yungyuc: This is EOF (end of file) or EOL (end of line) change, which is not necessary for the PR.  It is good to avoid this.

yungyuc (approved):
LGTM.  Let me know if you want this to be merged right away.  If you want to fix the EOF/EOL issue, update the PR and request for the next review.

When you are ready for the next review, do leave a global comment along with inline annotations.

OuChiaYun:
I've fixed the EOF/EOL issue and pushed the update. The PR is ready for another review.

yungyuc (approved):
> I've fixed the EOF/EOL issue and pushed the update. The PR is ready for another review.

Beautiful.

PR #1248, opened 2026-07-31
Add a CMake user preset template for scdv

yungyuc:
`CMakeUserPresets.json` names three directories that exist on one machine only, and the templates for it asked every checkout to type all three by hand. `contrib/dependency/build-scdv.sh` already knows them.

The PR adds `CMakeUserPresets.scdv.json` and `CMakeUserPresets.win-scdv.json` beside the example pair, which keeps its literal paths for a prefix built by hand.

Add a new skill `cmake-user-presets` to install the user presets.  Update the `worktree` skill to call it right after creating the worktree.

PR #1276, opened 2026-08-04
Name the user presets `ide-*` and keep agents on `make`

yungyuc:
PR #1248 updated the worktree skill to install the cmake user presets, and it interfered agent auto building.

Remove the `worktree` step that seeded the file before every build.  Also rename the `cmake-user-presets` skill to `ide-user-presets` with an IDE-scoped trigger, and state the boundary in `CLAUDE.md`, so that agents do not think the configuration is for them.

### DIFF
--- a/.claude/skills/worktree/SKILL.md
+++ b/.claude/skills/worktree/SKILL.md
@@ -17,11 +17,7 @@ Based on the agent type, do one of the following:
 - **Cursor:**  use Cursors's builtin `/worktree` command instead. If the built-in command is unavailable, fall back like Codex below.
 - **Codex, otherwise:** run `git worktree add -b worktree-<name> .claude/worktrees/<name> origin/master`, then run all commands from it.
 
-## 2. Seed the local CMake presets
-
-A new worktree is a fresh checkout, so it has no `CMakeUserPresets.json` and no `scdv-rel` preset. Install one with the `cmake-user-presets` skill before the first configure. Skip it when the task touches no build.
-
-## 3. Do the task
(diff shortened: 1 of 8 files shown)


.claude/skills/worktree/SKILL.md
```diff
@@ -17,11 +17,7 @@ Based on the agent type, do one of the following:
 - **Cursor:**  use Cursors's builtin `/worktree` command instead. If the built-in command is unavailable, fall back like Codex below.
 - **Codex, otherwise:** run `git worktree add -b worktree-<name> .claude/worktrees/<name> origin/master`, then run all commands from it.
 
-## 2. Seed the local CMake presets
-
-A new worktree is a fresh checkout, so it has no `CMakeUserPresets.json` and no `scdv-rel` preset. Install one with the `cmake-user-presets` skill before the first configure. Skip it when the task touches no build.
-
-## 3. Do the task
+## 2. Do the task
```
yungyuc, line 20: Drop user presets installation.

contrib/cmake/CMakeUserPresets.example.json
```diff
@@ -4,9 +4,9 @@
   "cmakeMinimumRequired": { "major": 4, "minor": 0, "patch": 1 },
   "configurePresets": [
     {
-      "name": "local-rel",
+      "name": "ide-local-rel",
```
yungyuc, line 7: Prefix the names with `ide-` and say the same in the `displayName` that `cmake --list-presets` prints.

Issue #1267, opened 2026-08-03
Polish the Oblique-Shock Reflection Application


yungyuc:
The oblique-shock reflection[^ywh1985] is solved by the CESE method[^chang1995][^wang1999].  Both the analytic solution of the Mach 3, ten-degree case[^anderson], and the pilot simulation should be validated and polished.

## Goal

1. **Easy-to-use GUI.**  Open the panel, press Start, and watch the run converge onto the analytic reflection.  The panel gains mesh-resolution controls, Stop and Reset, a run-to-steady mode that stops itself at convergence, a residual readout, a colorbar legend, and the analytic zone errors in percent.
2. **Convenient analysis.**  The judging machinery of the verification campaign becomes library code: residual history, zone-window sampling against the analytic states, the incident-shock-angle fit, and line profiles across the shocks.  Everything works headlessly from Python and feeds the GUI unchanged.
3. **Exemplar organization.**  The generic Euler solver separates from the problem built on it.  `solvcon.multidim.euler` keeps only GUI-free, problem-free solver support, and the reflection moves into `solvcon.pilot.apps.obsrefl`, the first entry in an application namespace where a problem pairs its non-GUI driver with its GUI.  A documentation page states the convention other applications copy.

The ability to provide GUI and analysis is key to validate the implementation.  Python and the GUI carry the work.  C++ changes only if a binding gap appears; none is expected, since `EulerCore` already exposes every state array the analysis needs (`so0c`, `so0n`, `cflc`, `gamma`).

## Sub-issues

- [x] Sub-issue 1: #1268
- [x] Sub-issue 2: #1269
- [x] Sub-issue 3: #1270
- [x] Sub-issue 4: #1271
- [x] Sub-issue 5: #1272
- [x] Sub-issue 6: #1273
- [x] Sub-issue 7: #1274

## Target shape

Reorganize the code into two Python packages: (1) the generic Euler solver knows nothing about the reflection, and (2) the reflection application owns everything specific to the problem, GUI and non-GUI alike.

```
solvcon/multidim/euler/        # generic Euler solver support: no GUI,
    __init__.py                # no problem-specific code
    field.py                   # body-cell views, mass, residual (grown
                               # from the existing module)
solvcon/pilot/apps/            # NEW: the application namespace
    __init__.py
    obsrefl/                   # the oblique-shock reflection problem
        __init__.py            # driver/analysis/session exported always;
                               # Qt modules guarded by the pilot toggle
[...]

PR #1280, opened 2026-08-06
Separate the generic Euler solver from the oblique-shock reflection

### LINKED ISSUE #1268: Separate the Generic Euler Solver from the Reflection (polish issue 1)
The goal is the two-package boundary drawn in the issue #1267 master plan's target shape.  The Euler solver is the reusable asset and the reflection is one problem built on it.

- `solvcon.multidim.euler` keeps only generic, GUI-free solver support: the body-cell views, joined by residual and domain-mass helpers (RMS density change per unit time over the body cells, with the definition documented).
- `solvcon.pilot.apps.obsrefl` receives everything specific to the oblique-shock reflection problem:
  - the mesher, relation, and driver from `multidim/euler/oblique.py` as `_driver.py`
  - the `_euler` GUI modules (`_oblique.py` as `_mesh_sample.py`, `_field_render.py` unchanged),
  - `_solution_info.py` split into `_panel.py` and `_app.py` with the feature class renamed from `SolutionInfo` to `ObliqueShockApp`

The move changes no behavior; the existing suites pass with only import paths updated, which is what guards the motion.

- PR #1280

yungyuc:
For issue #1268.

Made `solvcon.multidim.euler` a reusable asset and the oblique-shock reflection be one problem built on it.

No behavior changes. The existing suites pass with only import paths and the feature name updated, which is what guards the motion, and the field module gains the test it never had, over three triangles small enough that every expected value is known by hand.


solvcon/multidim/euler/_field.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+import numpy as np
+
+__all__ = [
+    'EulerField',
+]
+
+
+class EulerField(object):
```
yungyuc, line 16: Represent the field data for the Euler solver.

solvcon/pilot/apps/obsrefl/_app.py
```diff
@@ -0,0 +1,259 @@
...
+        super().__init__(parent)
+        self._on_close = on_close
+
+    def eventFilter(self, obj, event):
+        if event.type() == QEvent.Type.Close:
+            self._on_close()
+        return False
+
+
+class ObliqueShockApp(_gui_common.PilotFeature):
```
yungyuc, line 47: Control the simulation of the oblique shock reflection problem.

yungyuc (commented):
More polishing awaits the master issue #1267.

Issue #1277, opened 2026-08-05
`SimpleArray` and `SimpleCollector` provide `tolist()` helper


yungyuc:
Converting an array to a Python `list` is a common operation.   Currently it requires to use `.ndarray.tolist()` because `SimpleArray` and `SimpleCollector` do not provide `tolist()`:

```python
facn.ndarray[:, 0].tolist()
```

We should implement `SimpleArray.tolist()` (both typed and type-erased classes) and `SimpleCollector.tolist()`.

Chronos193:
I would like to contribute on this issue. Can you please assign it to me?

tigercosmos:
@Chronos193  Sure, please go ahead. Feel free to ask questions!

c1ydehhx:
I can help. Please assign this issue to me.

PR #1281, opened 2026-08-06
Issue#1277 : added tolist method to SimpleArrayPlex and SimpleCollector

Chronos193:
For #1277

### Changes
- Exposed `tolist()` binding on `WrapSimpleArrayPlex` inside `wrap_modifiers()`.
- Exposed `tolist()` binding on `WrapSimpleCollector`.
- Added unit tests in `tests/test_buffer.py`.

### DIFF
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -148,6 +148,13 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "to_column_major",
                 [](wrapped_type const & self)
                 { return self.to_column_major(); })
+            .def(
+                "tolist",
+                [](wrapped_type & self)
+                { return to_ndarray(self).attr("tolist")(); },
+                "Returns the array as a Python list."
+            )
+            
             .def_property_readonly(
                 "is_c_contiguous",
                 [](wrapped_type const & self)
@@ -885,6 +892,12 @@ WrapSimpleCollector<T>::WrapSimpleCollector(pybind11::module & mod, char const *
             "push_back",
             [](wrapped_type & self, value_type value)
             { self.push_back(value); })
+        .def(
+            "tolist",
+            [](pybind11::object const & self)
+            { return self.attr("as_array")().attr("tolist")(); },
+            "Converts SimpleCollector to Python list.")
+            
         .def_timed("as_array", &wrapped_type::as_array)
         //
         ;
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
@@ -315,6 +315,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
                        {
                         const auto shape = make_shape(py_shape);
                         return array.reshape(shape); }); })
+
             .def_property_readonly("has_ghost", DECL_MM_EXECUTE_TYPED_ARRAY_METHOD(has_ghost))
             .def_property(
                 "nghost",
@@ -356,6 +357,19 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
                             array.fill(value);
                         }); },
                 py::arg("value"))
+
+            .def(
+                "tolist",
+                [](wrapped_type & self)
+                {
+                    return execute_callback_with_typed_array(
+                        self,
+                        [](auto & array)
+                        {
+                            return to_ndarray(array).attr("tolist")();
+                        });
+                },
+                "Returns the array as a Python list.")
             //
             ;
 
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -282,6 +282,27 @@ class SimpleArrayBasicTC(unittest.TestCase):
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
         self.assertEqual((2, 2, 2, 3), sarr.reshape((2, 2, 2, 3)).shape)
 
+    def test_tolist(self):
+        # 1D array
+        input_size = 10
+        arr = solvcon.SimpleArray(input_size, dtype='float64')
+        for i in range(input_size):
+            arr[i] = float(i)
+
+        res_list = arr.tolist()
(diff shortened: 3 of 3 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -148,6 +148,13 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "to_column_major",
                 [](wrapped_type const & self)
                 { return self.to_column_major(); })
+            .def(
```
yungyuc, line 151: Use native implementation, depending on only Python.  Do not use numpy implementation.

cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -315,6 +315,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
                        {
                         const auto shape = make_shape(py_shape);
                         return array.reshape(shape); }); })
+
```
yungyuc, line 318: Do not add non-related cosmetic change.

tests/test_buffer.py
```diff
@@ -5296,6 +5317,23 @@ def test_push_back(self):
...
+    
+    def test_tolist_all_types(self):
+        # Test across various collector types
+        test_cases = [
+            (solvcon.SimpleCollectorInt32, [1, 2, 3]),
+            (solvcon.SimpleCollectorFloat64, [1.0, 2.0, 3.0]),
+            (solvcon.SimpleCollectorBool, [True, False, True]),
+        ]
+
+        for collector_cls, expected_vals in test_cases:
```
yungyuc, line 5329: `test_cases` is not a good variable name because it conflicts with "test cases".  Here you do not need to name it:

```suggestion
        for collector_cls, expected_vals in [
            (solvcon.SimpleCollectorInt32, [1, 2, 3]),
            (solvcon.SimpleCollectorFloat64, [1.0, 2.0, 3.0]),
            (solvcon.SimpleCollectorBool, [True, False, True]),
        ]:
```

It will be even better to use sub tests.  Loops make it harder to understand the failures when they happen.

yungyuc (changes requested):
This is not a one-liner change.  Leave inline annotation for your PR.

- [ ] Use native implementation for `tolist()`, depending on only Python.  Do not use numpy implementation.
- [ ] Do not add non-related cosmetic change.  Cosmetic changes should use their own PRs.

yungyuc:
> Resolve #xxx

Never write anything like this.  See https://doc.solvcon.net/devguide/contributing.html .

yungyuc:
@ThreeMonth03 please take a look.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -885,6 +892,12 @@ WrapSimpleCollector<T>::WrapSimpleCollector(pybind11::module & mod, char const *
             "push_back",
             [](wrapped_type & self, value_type value)
             { self.push_back(value); })
+        .def(
+            "tolist",
+            [](pybind11::object const & self)
```
ThreeMonth03, line 897: [P1] This collector path still crosses the Python-bound `as_array()` and `tolist()` methods and therefore ends at the NumPy-backed array conversion. Fixing only the array binding would leave `SimpleCollector.tolist()` dependent on NumPy. Please use the typed collector/array storage with the same native pybind list-conversion helper directly.

tests/test_buffer.py
```diff
@@ -282,6 +282,27 @@ def test_SimpleArray(self):
         self.assertEqual((12, 2), sarr.reshape((12, 2)).shape)
         self.assertEqual((2, 2, 2, 3), sarr.reshape((2, 2, 2, 3)).shape)
 
+    def test_tolist(self):
+        # 1D array
+        input_size = 10
+        arr = solvcon.SimpleArray(input_size, dtype='float64')
```
ThreeMonth03, line 288: [P1] `solvcon.SimpleArray` is the type-erased `SimpleArrayPlex`, so these tests never exercise the typed `WrapSimpleArray<T>` binding added in this PR. Please add a typed case such as `SimpleArrayFloat64` from a non-contiguous or negative-stride NumPy view; otherwise a native implementation that walks storage instead of logical strides can still pass every added test.

tests/test_buffer.py
```diff
@@ -5296,6 +5317,23 @@ def test_push_back(self):
         self.assertEqual(20, ct.capacity)  # double capacity but not power of 2
         self.assertEqual(11, len(ct))
         self.assertEqual(ct[10], 3.14159 * 4)
+    
+    def test_tolist_all_types(self):
+        # Test across various collector types
+        test_cases = [
+            (solvcon.SimpleCollectorInt32, [1, 2, 3]),
+            (solvcon.SimpleCollectorFloat64, [1.0, 2.0, 3.0]),
+            (solvcon.SimpleCollectorBool, [True, False, True]),
```
ThreeMonth03, line 5326: [P2] Please cover at least one complex collector or array. The native implementation needs to preserve the existing `ndarray.tolist()` behavior of returning Python `complex` values, and the current int/float/bool cases cannot detect an incorrect `Complex<T>` conversion.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -148,6 +148,13 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "to_column_major",
                 [](wrapped_type const & self)
                 { return self.to_column_major(); })
+            .def(
+                "tolist",
+                [](wrapped_type & self)
+                { return to_ndarray(self).attr("tolist")(); },
+                "Returns the array as a Python list."
```
ThreeMonth03, line 155: [P1] Please run the repository formatting and whitespace checks before the next revision. `make cformat` rejects this `.def` layout and the collector binding; `git diff --check` also reports trailing whitespace here, in the collector hunk, and at `tests/test_buffer.py:5320` (where flake8 reports W293).

ThreeMonth03 (changes requested):
Requesting changes for the inline findings. A fresh detached build and the two added tests pass, but `make cformat`, `git diff --check`, and flake8 fail. Please also update the branch to resolve its conflict with master.

tigercosmos:
@Chronos193  Would you update the PR?

yungyuc:
Close due to inactivity.

PR #1282, opened 2026-08-07
Reduce matrix profiling repetitions

ThreeMonth03:
Repeated samples reduce benchmark noise, but the matrix profiler ran each planned implementation 75 times per case and caused scheduled CI to time out. This change keeps every case while reducing the default to one sample over three rounds, with one warmup per round.

For more stable local measurements, increase `--samples` and `--rounds`, for example `python profiling/profile_matrix_ops.py --warmups 2 --samples 15 --rounds 5`.

Related to #1172.

### DIFF
--- a/profiling/profile_matrix_ops.py
+++ b/profiling/profile_matrix_ops.py
@@ -1,10 +1,13 @@
 # Copyright (c) 2026, solvcon team <contact@solvcon.net>
 # BSD 3-Clause License, see COPYING
 
+import argparse
 import functools
 import itertools
 import statistics
+
 import numpy as np
+
 import solvcon
 
 
@@ -91,7 +94,7 @@ def profile_one_call(func, *args):
     return result[0]["total_time"]
 
 
-def profile_unbatched_gemm(dtype, sides, samples=10):
+def profile_unbatched_gemm(dtype, sides, samples=1):
     tile_configs = (
         (16, 16, 16),
         (32, 32, 32),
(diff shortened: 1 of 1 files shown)


profiling/profile_matrix_ops.py
```diff
@@ -191,42 +194,75 @@ def iter_planned_cases(dtype):
...
+    parser.add_argument(
+        "--warmups", type=parse_nonnegative_count, default=0,
+        help="untimed planned calls before each round")
+    parser.add_argument(
+        "--samples", type=parse_positive_count, default=1,
+        help="timed calls per method")
+    parser.add_argument(
+        "--rounds", type=parse_positive_count, default=1,
+        help="planned profiling rounds")
+    return parser.parse_args(argv)
```
ThreeMonth03, line 245: Minimize profiling count in CI. If we want a precise result, it is necessary to adjust argument and profile targets repetitvely in local.

ThreeMonth03 (commented):
@yungyuc Please take a look.

profiling/profile_matrix_ops.py
```diff
@@ -1,10 +1,13 @@
 # Copyright (c) 2026, solvcon team <contact@solvcon.net>
 # BSD 3-Clause License, see COPYING
 
+import argparse
 import functools
 import itertools
 import statistics
+
```
yungyuc, line 8: Good blank lines.

profiling/profile_matrix_ops.py
```diff
@@ -191,42 +194,75 @@ def iter_planned_cases(dtype):
...
+        description="Profile matrix multiplication operations.",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        "--warmups", type=parse_nonnegative_count, default=0,
+        help="untimed planned calls before each round")
+    parser.add_argument(
+        "--samples", type=parse_positive_count, default=1,
+        help="timed calls per method")
+    parser.add_argument(
+        "--rounds", type=parse_positive_count, default=1,
```
yungyuc: 1 round is too few.  Use 3 or 5 as the default.

Also add command-line help message to suggest a more accurate round value when profiling.  Nightly can use few (like 3).

profiling/profile_matrix_ops.py
```diff
@@ -191,42 +194,75 @@ def iter_planned_cases(dtype):
...
+        raise argparse.ArgumentTypeError("must be nonnegative")
+    return count
+
+
+def parse_arguments(argv=None):
+    parser = argparse.ArgumentParser(
+        description="Profile matrix multiplication operations.",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        "--warmups", type=parse_nonnegative_count, default=0,
```
yungyuc: 0 warm up is too few.  At least 1.

yungyuc (changes requested):
Thanks for the quick fix.  Need to address some points:

- [x] More default round value.
- [x] Add command-line help message to suggest a more accurate round value when profiling.  Nightly can use few (like 3).
- [x] Use at least 1 warm-up.  Inaccurate timing is worse than no timing.

profiling/profile_matrix_ops.py
```diff
@@ -191,42 +194,68 @@ def iter_planned_cases(dtype):
...
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        "--warmups", type=parse_positive_count, default=1,
+        help="untimed planned calls before each round")
+    parser.add_argument(
+        "--samples", type=parse_positive_count, default=1,
+        help="timed calls per method in each round")
+    parser.add_argument(
+        "--rounds", type=parse_positive_count, default=3,
+        help="planned profiling rounds; use 5 or more for stable results")
```
ThreeMonth03, line 244: Increase the number of warmups and rounds. 
Additionally, remind increasing number for better profiling result.

ThreeMonth03 (commented):
- [x] More default round value.
 - [x] Add command-line help message to suggest a more accurate round value when profiling. Nightly can use few (like 3).
 - [x] Use at least 1 warm-up. Inaccurate timing is worse than no timing.

yungyuc (approved):
LGTM.  Thanks for the swift fix.

PR #1283, opened 2026-08-07
Optimize strided matmul execution

ThreeMonth03:
## Motivation

Some signed-stride cases are much slower than NumPy:

```text
Small GEMM
lhs: shape (S, S), strides (S, -1)
rhs: shape (S, S), strides (S, 1)
cause: CBLAS cannot express the LHS layout; generic IJK reloads each LHS value for every output column.

Unique batch
lhs: shape (B, M, K), strides (M*K, K, -1)
rhs: shape (B, K, N), strides (K*N, -N, 1)
cause: neither operand is reusable, so the broadcast-only packing route leaves the whole batch generic.

Batched vectors
(16,) @ (B, 16, 16)
(B, 16, 16) @ (16,)
cause: BLAS wrapper overhead kept 256-element matrices on the generic route.
```

## Approach

A small signed-stride GEMM now computes adjacent output columns together:

```text
Before
for column:
    for inner:
        output[column] += lhs[inner] * rhs[inner, column]

After
for inner:
    value = lhs[inner]
    for lane in 0..7:
        output[column + lane] += value * rhs[inner, column + lane]
```

One LHS load is reused across eight columns. A four-column block and a scalar tail handle the remainder. This improves the existing generic route without adding shape-specific kernels.

A large incompatible batch materializes only what each BLAS call needs:

```text
lhs: shape (1, M, K), strides (M*K, K, -1)  -> pack once

for each batch:
    rhs: shape (K, N), strides (-N, 1)       -> pack into reusable scratch
    packed lhs @ packed rhs                  -> existing GEMM call
```

The same scratch storage is overwritten for the next unique matrix. If the broadcast operand is on the RHS, the roles are reversed.

The batched-vector crossover follows the measured matrix size:

```text
S = 16 -> 16 * 16 = 256 matrix elements

Before: 256 < 512  -> generic
After:  256 >= 256 -> BLAS
```

BLAS integer conversion now casts valid dimensions directly; invalid dimensions use the range-error helper. The packed-GEMM crossover is 16 for real values and 8 for complex values. Direct BLAS and Winograd are unchanged.

## Profiling

The plots show median execution time per call for Before PR, After PR, and NumPy 2.5.1. All 152 GEMM points improve over Before PR, with family geometric-mean speedups of 4.76x for equal batches and 4.68x when one operand is broadcast. Batched GEVM/GEMV improve by 1.18x; the seven non-improving controls are below 3 us and differ by at most 0.19 us.

Relative to NumPy, equal-batch GEMM, one-broadcast GEMM, and batched GEVM/GEMV use 85%, 78%, and 97% of NumPy's execution time on their family geometric means.

<details>
<summary>float32 profiles</summary>

Batched matrix-vector multiplication - float32

Equal-batch matrix multiplication - float32

Matrix multiplication with one broadcast operand - float32

</details>

<details>
<summary>float64 profiles</summary>

Batched matrix-vector multiplication - float64

Equal-batch matrix multiplication - float64

Matrix multiplication with one broadcast operand - float64

</details>

<details>
<summary>Profiling method</summary>

- Environment: native macOS 26.5.1 arm64 on an 8 GB Apple M1 MacBook Air, Python 3.14.6, NumPy 2.5.1 with Apple Accelerate, and one BLAS thread.
- Before PR is `11dbb05b`; After PR is `20ee8a56`.
- Before PR, After PR, and NumPy run in separate persistent workers. Every point repeats all six method orders twice.
- Each method receives two warmups and five timed calls in each of 12 rounds. The plotted value is the untrimmed median of 60 calls, including result allocation.
- All 432 Before/After differential checks passed.

</details>

<details>
<summary>Raw data</summary>

- [Complete benchmark output](https://github.com/user-attachments/files/31065611/pr1283-macos-arm64-20ee8a56.jsonl.txt) (SHA-256: `8417c3a69a8242c22030cdc3b858ce4e6336574c8e7553abf3eb88b31774ee1a`)
- [Audit summary](https://github.com/user-attachments/files/31065612/pr1283-macos-arm64-audit-20ee8a56.json.txt) (SHA-256: `fed22e15bad0acf0905cba2e4d8558e81f353ac3554647211ba2cf3b3e37d9c5`)

</details>

Related to #1172 and #1337.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -12,9 +12,11 @@
 #include <solvcon/math/math.hpp>
 
 #include <algorithm>
+#include <array>
 #include <cstddef>
 #include <cstdint>
 #include <format>
+#include <limits>
 #include <optional>
 #include <ranges>
 #include <stdexcept>
@@ -87,7 +89,7 @@ struct MatmulTuning
     }; /* end struct DirectBlas */
 
     /**
-     * @brief Pack reused matrices that cannot be described directly to BLAS.
+     * @brief Set the GEMM crossover for packing BLAS-incompatible matrices.
      */
     struct MatrixPacking
     {
@@ -276,10 +278,10 @@ private:
  * evaluates one `(3,4) @ (4,6)` contraction at each offset. The results are
  * written into the allocated `(2,5,3,6)` output.
  *
- * @note The current implementation selects generic, direct BLAS, or Winograd
- * kernels using measured thresholds. Packing materializes a reused broadcast
- * matrix or vector once in row-major storage before traversal. Unsupported
- * equal-batch matrix layouts remain generic.
+ * @note The current implementation selects generic, BLAS, or Winograd kernels
+ * using measured thresholds. Packing materializes reused broadcast operands
+ * once and streams the remaining operands one matrix at a time through
+ * bounded row-major scratch.
  */
 template <typename Array>
 class MatmulExecutor
@@ -315,10 +317,10 @@ private:
             .gemm_min_dimension = 8,
         },
         .matrix_packing = {
-            .gemm_min_dimension = 16,
+            .gemm_min_dimension = is_complex_v<value_type> ? 8 : 16,
         },
         .batched_vector = {
-            .direct_min_matrix_elements = 512,
+            .direct_min_matrix_elements = 256,
             .always_pack_min_matrix_elements = 4096,
             .pack_min_matrix_elements = 1024,
             .pack_min_batch_size = 4,
@@ -327,6 +329,9 @@ private:
         },
         .winograd = WINOGRAD_TUNING,
     };
+    static constexpr size_t MAX_SCRATCH_ELEMENTS = std::min(
+        std::numeric_limits<size_t>::max() / sizeof(value_type),
+        static_cast<size_t>(std::numeric_limits<ssize_t>::max()));
 
     MatmulSelection select_execution() const;
     MatmulSelection select_dot() const;
@@ -347,7 +352,18 @@ private:
     void execute_dot_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
     void execute_gevm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
     void execute_gemv_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+    void prepare_gemm_scratch();
     void execute_gemm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
+
+    static size_t checked_scratch_elements(ssize_t rows, ssize_t columns);
+    static matrix_view_type pack_matrix_to_scratch(
+        value_type const * source,
+        value_type * scratch,
+        value_type const *& cached_source,
+        ssize_t rows,
+        ssize_t columns,
+        ssize_t row_stride,
+        ssize_t column_stride);
 
     void execute_winograd();
 
@@ -360,6 +376,8 @@ private:
         ssize_t column_stride,
         ssize_t rows,
         ssize_t columns);
+    template <size_t ColumnBlock>
+    void multiply_generic_column_block(value_type * output, value_type const * lhs, value_type const * rhs);
     void execute_generic(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base);
 
     MatmulPlan m_plan;
@@ -367,6 +385,10 @@ private:
     Array const & m_rhs;
     std::optional<Array> m_packed_lhs;
     std::optional<Array> m_packed_rhs;
+    std::optional<Array> m_gemm_scratch;
+    size_t m_lhs_scratch_elements = 0;
+    value_type const * m_cached_lhs_source = nullptr;
+    value_type const * m_cached_rhs_source = nullptr;
     value_type * m_output_data;
     value_type const * m_lhs_data;
     value_type const * m_rhs_data;
@@ -632,6 +654,11 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan plan, Array & output, Array con
 template <typename Array>
 void MatmulExecutor<Array>::execute()
 {
+    if (m_plan.batch_size() == 0)
+    {
+        return;
+    }
+
     MatmulSelection const selection = select_execution();
 
     if (selection.packing)
@@ -658,6 +685,7 @@ void MatmulExecutor<Array>::execute()
             execute_contractions<MatmulKernel::BlasGemv>();
             return;
         case MatmulKernel::BlasGemm:
+            prepare_gemm_scratch();
             execute_contractions<MatmulKernel::BlasGemm>();
             return;
         case MatmulKernel::Winograd:
@@ -805,11 +833,10 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
             .lhs = !lhs_blas_compatible,
             .rhs = !rhs_blas_compatible,
         };
-        PackingState const packing = select_matrix_packing(required);
-        if (packing)
-        {
-            return MatmulSelection{.kernel = MatmulKernel::BlasGemm, .packing = packing};
-        }
+        return MatmulSelection{
+            .kernel = MatmulKernel::BlasGemm,
+            .packing = select_matrix_packing(required),
+        };
     }
     return MatmulSelection{};
 }
@@ -817,15 +844,10 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
 template <typename Array>
 PackingState MatmulExecutor<Array>::select_matrix_packing(PackingState required) const
 {
-    bool const lhs_supported =
-        !required.lhs || (m_plan.lhs_is_broadcast() && !m_plan.lhs_has_zero_batch_stride());
-    bool const rhs_supported =
-        !required.rhs || (m_plan.rhs_is_broadcast() && !m_plan.rhs_has_zero_batch_stride());
-    if (!required || !lhs_supported || !rhs_supported)
-    {
-        return PackingState{};
-    }
-    return required;
+    return PackingState{
+        .lhs = required.lhs && m_plan.lhs_is_broadcast() && !m_plan.lhs_has_zero_batch_stride(),
+        .rhs = required.rhs && m_plan.rhs_is_broadcast() && !m_plan.rhs_has_zero_batch_stride(),
+    };
 }
 
 template <typename Array>
@@ -973,14 +995,66 @@ void MatmulExecutor<Array>::execute_gemv_blas(value_type * output, value_type co
 }
 
 template <typename Array>
+void MatmulExecutor<Array>::prepare_gemm_scratch()
+{
+    if (m_gemm_scratch)
+    {
+        m_cached_lhs_source = nullptr;
+        m_cached_rhs_source = nullptr;
+        return;
+    }
+
+    std::optional<matrix_view_type> const lhs_view = lhs_matrix_view(m_lhs_data);
+    std::optional<matrix_view_type> const rhs_view = rhs_matrix_view(m_rhs_data);
+    if (lhs_view && rhs_view)
+    {
+        return;
+    }
+
+    m_lhs_scratch_elements = lhs_view ? 0 : checked_scratch_elements(m_plan.rows(), m_plan.inner_size());
+    size_t const rhs_elements = rhs_view ? 0 : checked_scratch_elements(m_plan.inner_size(), m_plan.columns());
+    if (m_lhs_scratch_elements > MAX_SCRATCH_ELEMENTS - rhs_elements)
+    {
+        throw std::length_error("MatmulExecutor::prepare_gemm_scratch(): scratch size overflows");
+    }
+    typename Array::shape_type const scratch_shape{static_cast<ssize_t>(m_lhs_scratch_elements + rhs_elements)};
+    m_gemm_scratch.emplace(scratch_shape);
+}
+
+template <typename Array>
 void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data)
 {
+    std::optional<matrix_view_type> lhs_view = lhs_matrix_view(lhs_data);
+    std::optional<matrix_view_type> rhs_view = rhs_matrix_view(rhs_data);
+    if (!lhs_view)
+    {
+        lhs_view = pack_matrix_to_scratch(
+            lhs_data,
+            m_gemm_scratch.value().data(),
+            m_cached_lhs_source,
+            m_plan.rows(),
+            m_plan.inner_size(),
+            m_plan.lhs_row_stride(),
+            m_plan.lhs_inner_stride());
+    }
+    if (!rhs_view)
+    {
+        rhs_view = pack_matrix_to_scratch(
+            rhs_data,
+            m_gemm_scratch.value().data() + m_lhs_scratch_elements,
+            m_cached_rhs_source,
+            m_plan.inner_size(),
+            m_plan.columns(),
+            m_plan.rhs_inner_stride(),
+            m_plan.rhs_column_stride());
+    }
+
     BlasGemmOperation<value_type> const operation{
         .rows = m_plan.rows(),
         .columns = m_plan.columns(),
         .inner_size = m_plan.inner_size(),
-        .lhs = require_matrix_view(lhs_matrix_view(lhs_data)),
-        .rhs = require_matrix_view(rhs_matrix_view(rhs_data)),
+        .lhs = require_matrix_view(lhs_view),
+        .rhs = require_matrix_view(rhs_view),
         .output = {
             .m_data = output,
             .m_leading_dimension = m_plan.columns(),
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -101,15 +115,40 @@ struct MatmulTuning
 
     DirectBlas direct_blas;
     MatrixPacking matrix_packing;
+    FixedMatrix fixed_matrix;
     BatchedVector batched_vector;
 }; /* end struct MatmulTuning */
 
+template <typename T>
```
ThreeMonth03: Complex operands use complete-operand packing from side 8, while real operands begin at side 16. Fixed kernels are separately enabled for `Complex<float>`; `Complex<double>` continues through direct BLAS, packing, or generic IJK.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -734,39 +802,71 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
...
     ssize_t const minimum_dimension =
         std::min({m_plan.rows(), m_plan.columns(), m_plan.inner_size()});
-    bool const lhs_blas_compatible = bool(lhs_matrix_view(m_lhs_data));
-    bool const rhs_blas_compatible = bool(rhs_matrix_view(m_rhs_data));
-    if (minimum_dimension >= TUNING.direct_blas.gemm_min_dimension && lhs_blas_compatible && rhs_blas_compatible)
-    {
-        return MatmulSelection{.kernel = MatmulKernel::BlasGemm, .packing = {}};
-    }
-    if (minimum_dimension >= TUNING.matrix_packing.gemm_min_dimension)
+    PackingState required_packing;
```
ThreeMonth03: This function defines the GEMM priority: direct BLAS, fixed IKJ/JKI, materialized BLAS, then generic IJK. Selection happens once before batch traversal, so the hot batch loop does not repeat layout checks.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -734,39 +802,71 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
...
+            return MatmulSelection{.kernel = MatmulKernel::BlasGemm, .packing = required_packing};
+        }
+    }
+#endif
     return MatmulSelection{};
 }
 
 template <typename Array>
-PackingState MatmulExecutor<Array>::select_matrix_packing(PackingState required) const
+std::optional<MatmulSelection> MatmulExecutor<Array>::select_fixed_gemm() const
```
ThreeMonth03: RHS column-contiguous layouts select IKJ, while RHS inner-contiguous layouts select JKI. If neither layout is supplied but one physical RHS is reused by broadcasting, the selector packs that RHS once and uses fixed IKJ.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -734,39 +802,71 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
...
         {
-            return MatmulSelection{.kernel = MatmulKernel::BlasGemm, .packing = packing};
+            return *fixed;
         }
     }
+
+#if (defined(__APPLE__) && defined(__arm64__)) || defined(SC_HAS_CBLAS)
+    if constexpr (can_matmul_blas_v<value_type>)
+    {
+        if (minimum_dimension >= TUNING.matrix_packing.gemm_min_dimension && required_packing)
```
ThreeMonth03: For larger BLAS-incompatible inputs, the selected packing state materializes each complete supplied operand once. The existing `pack()` method then rebuilds `MatmulPlan` for the new row-major storage before traversal.

cpp/solvcon/buffer/matmul.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+    else
+    {
+        static_assert(Kernel == MatmulKernel::FixedJki);
+        traverse_fixed_batch<Kernel, T, Side>(output, lhs, rhs, plan);
+    }
+}
+
+template <MatmulKernel Kernel, typename T, size_t Side>
+static void dispatch_fixed_side(
```
ThreeMonth03: The runtime side search occurs once per matmul call and lands on a compile-time `Side` and kernel. That specialization then traverses every planned batch offset without route or size branches inside each contraction.

tests/test_matrix.py
```diff
@@ -417,6 +417,30 @@ def test_batch_strides(self):
                         shape=lhs_shape, lhs=lhs_case, rhs=rhs_case):
                     self.assert_matmul_planned(lhs, rhs, expected)
 
+    def test_fixed_sides(self):
```
ThreeMonth03: This test exhausts the configured float32 and float64 side ranges. The side-9 cases in `test_matrix_strides()` and `test_broadcast_matrix_strides()` cover BLAS-incompatible complex layouts: complex64 exercises fixed dispatch, while complex128 exercises the earlier packing route.

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -266,23 +330,15 @@ class MatmulExecutor
...
         .direct_blas = {
             .dot_min_length = 128,
             .compact_gevm_min_elements = 729,
             .gemv_min_dimension = 32,
             .gemm_min_dimension = 8,
         },
-        .matrix_packing = {
-            .gemm_min_dimension = 16,
-        },
+        .matrix_packing = MATRIX_PACKING_TUNING<value_type>,
```
yungyuc: Having template expressions like `MATRIX_PACKING_TUNING<value_type>` in a long `constexpr` here is confusing.  I cannot tell what it means.

Please explain what it means in the PR.  After clarify, we decide how it should be implemented.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -229,23 +278,38 @@ class MatmulPlan
...
+
 /**
  * @brief Select and execute one contraction kernel for a MatmulPlan.
  *
  * MatmulExecutor combines plan metadata with MatmulTuning to select a
  * MatmulSelection. It applies the selected operand preparation, rebuilds the
  * plan when a physical layout changes, and traverses every batch offset with
- * the same kernel. Generic kernels read signed strides directly, while BLAS
- * kernels consume compatible vector and matrix descriptors.
+ * the same kernel. The generic IJK kernel reads arbitrary signed strides.
```
yungyuc: Define clearly what "IJK", "IKJ", "JKI" kernels mean.

yungyuc (changes requested):
- [ ] ~~Clarify the use of the template expressions in the `constexpr`.~~
      ~~Be specific to how to determine the machine-specific numbers~~
- [ ] ~~Define clearly what "IJK", "IKJ", "JKI" kernels mean.~~
      ~~Clearly and concisely describe "IJK", "IKJ", "JKI" in code comments.~~

cpp/solvcon/buffer/matmul.hpp
ThreeMonth03: `MATRIX_PACKING_TUNING<T>` is a compile-time table for element type `T`. It is not runtime dispatch. Packing starts at dimension 16 for real types and 8 for complex types. `FIXED_MATRIX_TUNING<T>` likewise stores the measured size limits for the fixed kernels.

cpp/solvcon/buffer/matmul.hpp
ThreeMonth03: `i`, `j`, and `k` are the output row, output column, and reduction index. IJK is row -> column -> reduction. IKJ is row -> reduction -> column. JKI is column -> reduction -> row. Generic IJK supports arbitrary strides. Fixed IKJ is selected for contiguous RHS rows. Fixed JKI is selected for a contiguous RHS reduction dimension. "Fixed" means the square side is a compile-time constant.

cpp/solvcon/buffer/matmul.hpp
yungyuc: I dislike the hierarchy here, reading: `TUNING.MATRIX_PACKING_TUNING` and `TUNING.FIXED_MATRIX_TUNING`.  Even after you explain, I still cannot get it.

What prevented us from using the clear constant definition like the past?

```cpp
.matrix_packing = {
    .gemm_min_dimension = 16,
},
```

cpp/solvcon/buffer/matmul.hpp
yungyuc: It's still unclear to me.

yungyuc (changes requested):

cpp/solvcon/buffer/matmul.hpp
ThreeMonth03: The thresholds differ by value type because each type was benchmarked separately on a MacBook Air M1.

```cpp
// float
.matrix_packing = {
    .gemm_min_dimension = 16,
},
.fixed_matrix = {
    .min_side = 8,
    .ikj_max_side = 20,
    .jki_max_side = 15,
},

// Complex<float>
.matrix_packing = {
    .gemm_min_dimension = 8,
},
.fixed_matrix = {
    .min_side = 8,
    .ikj_max_side = 15,
    .jki_max_side = 15,
},
```

The current template constants hide these measured values. I will remove that indirection and keep the type-dependent thresholds visible in `TUNING`.

cpp/solvcon/buffer/matmul.hpp
ThreeMonth03: For `A(M, K) @ B(K, N) = C(M, N)`, the names specify the loop order:

```cpp
// IJK
for (i)
    for (j)
        for (k)
            C[i][j] += A[i][k] * B[k][j];

// IKJ
for (i)
    for (k)
        for (j)
            C[i][j] += A[i][k] * B[k][j];

// JKI
for (j)
    for (k)
        for (i)
            C[i][j] += A[i][k] * B[k][j];
```

yungyuc: Now it's clear.  Is this documented anywhere in the code comments?  If not, please add it at a proper location.  And I suggest to write in a compact form like:

```cpp
// IJK
for (i) { for (j) { for (k) { C[i][j] += A[i][k] * B[k][j]; } } }
// IKJ
for (i) { for (k) { for (j) { C[i][j] += A[i][k] * B[k][j]; } } }
// JKI
for (j) { for (k) { for (i) { C[i][j] += A[i][k] * B[k][j]; } } }
```

cpp/solvcon/buffer/matmul.hpp
yungyuc: If the numbers are machine/chip dependent, they should be filled in during runtime, not compile time.

If filling them in during runtime hurts performance (I have no idea if it may be true or false; it needs to be clarified with experiments), we will need to build multiple copies of the optimized helpers.

This PR is already long and experimenting with the runtime filling should use another PR.  

In this PR, clearly describe how the numbers are determined in code comments.  Please try to use literals like what you wrote above:

```cpp
// float
.matrix_packing = {
    .gemm_min_dimension = 16,
},
.fixed_matrix = {
    .min_side = 8,
    .ikj_max_side = 20,
    .jki_max_side = 15,
},
```

Perhaps you can use something like `is_same_v<T>` to distinguish `float` from `Complex<float>`?  If that is too cumbersome, try another approach to rename `MATRIX_PACKING_TUNING` and `FIXED_MATRIX_TUNING` to something meaningful.

yungyuc (changes requested):

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -805,27 +833,21 @@ MatmulSelection MatmulExecutor<Array>::select_gemm() const
...
             .rhs = !rhs_blas_compatible,
         };
-        PackingState const packing = select_matrix_packing(required);
-        if (packing)
-        {
-            return MatmulSelection{.kernel = MatmulKernel::BlasGemm, .packing = packing};
-        }
+        return MatmulSelection{
+            .kernel = MatmulKernel::BlasGemm,
+            .packing = select_matrix_packing(required),
```
ThreeMonth03, line 838: Packing happens at two points:

```text
before batch traversal:
    if an incompatible operand reuses one matrix:
        operand = pack_once(operand)

for each contraction:
    if lhs is still incompatible:
        lhs = pack_current_matrix(lhs, lhs_scratch)
    if rhs is still incompatible:
        rhs = pack_current_matrix(rhs, rhs_scratch)
    gemm(lhs, rhs, output)
```

`PackingState` controls the first block. `execute_gemm_blas()` performs the second.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -1060,14 +1194,65 @@ std::optional<typename MatmulExecutor<Array>::matrix_view_type> MatmulExecutor<A
...
+    }
+    std::ranges::copy(totals, output);
+}
+
 template <typename Array>
 void MatmulExecutor<Array>::execute_generic(ssize_t output_base, ssize_t lhs_base, ssize_t rhs_base)
 {
+    constexpr ssize_t LARGE_COLUMN_BLOCK = 8;
+    constexpr ssize_t SMALL_COLUMN_BLOCK = 4;
+    bool const block_columns = m_plan.columns() >= SMALL_COLUMN_BLOCK &&
```
ThreeMonth03, line 1227: This stays inside generic dispatch:

```text
8-column block -> 4-column block -> scalar tail
```

An RHS column stride of `1` or `-1` lets one LHS load update adjacent outputs.

cpp/solvcon/math/blas_compat.cpp
```diff
@@ -44,17 +44,22 @@ static blas_int_type to_blas_int(ssize_t value, char const * name)
...
-        return static_cast<blas_int_type>(value);
-    }
 
     throw std::out_of_range(
         std::format("solvcon BLAS wrapper: {}={} exceeds BLAS integer range",
                     name,
                     value));
 }
 
+static blas_int_type to_blas_int(ssize_t value, char const * name)
```
ThreeMonth03, line 54:
```text
valid   -> range check -> cast -> BLAS
invalid -> range check -> [[noreturn]] helper -> format + throw
```

The split makes successful conversion inlineable.

tests/test_matrix.py
```diff
@@ -465,6 +465,43 @@ def test_broadcast_matrix_strides(self):
                         shape=(m, k, n), lhs=lhs_case, rhs=rhs_case):
                     self.assert_matmul_planned(lhs, rhs, expected)
 
+    def test_large_batched_matrix_strides(self):
+        """Batched matrices preserve unique, broadcast, and zero strides."""
+        dtype = np.dtype(self.dtype).name
+        batch, m, k, n = 3, 16, 17, 18
```
ThreeMonth03, line 471: This non-square case exercises both packing timings:

```text
M, K, N = 16, 17, 18
cases = [
    (unique_lhs, unique_rhs),       # pack each current matrix
    (broadcast_lhs, unique_rhs),    # pack lhs once; pack current rhs
    (unique_lhs, broadcast_rhs),    # pack current lhs; pack rhs once
    (zero_stride_lhs, unique_rhs),  # reuse lhs scratch
    (unique_lhs, zero_stride_rhs),  # reuse rhs scratch
]

for lhs, rhs in cases:
    assert planned_matmul(lhs, rhs) == numpy.matmul(lhs, rhs)
```

The shared test runs float32, float64, complex64, and complex128.

ThreeMonth03 (commented):
I afterwards notice that's not a good idea to directly add more kernels in the branch, so I rewrite pull request and fix the bottleneck of generic matrix multiplication.

cpp/solvcon/buffer/matmul.hpp
yungyuc: This is complex.  But my guts feeling tells me it's necessary, at least for now.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -972,15 +994,67 @@ void MatmulExecutor<Array>::execute_gemv_blas(value_type * output, value_type co
...
+        throw std::length_error("MatmulExecutor::prepare_gemm_scratch(): scratch size overflows");
+    }
+    typename Array::shape_type const scratch_shape{static_cast<ssize_t>(m_lhs_scratch_elements + rhs_elements)};
+    m_gemm_scratch.emplace(scratch_shape);
+}
+
 template <typename Array>
 void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data)
 {
+    auto lhs_view = lhs_matrix_view(lhs_data);
```
yungyuc: Don't use `auto` when the type name is not in the expression.

There are multiple issues like this in the PR.  Please correct them all.

yungyuc (changes requested):
LGTM.  The implementation is hard to follow but I feel we should go with it.  Please fix a minor point:

- [x] Don't use `auto` when the type name is not in the expression.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -972,15 +994,67 @@ void MatmulExecutor<Array>::execute_gemv_blas(value_type * output, value_type co
...
+    }
+    typename Array::shape_type const scratch_shape{static_cast<ssize_t>(m_lhs_scratch_elements + rhs_elements)};
+    m_gemm_scratch.emplace(scratch_shape);
+}
+
 template <typename Array>
 void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data)
 {
+    std::optional<matrix_view_type> lhs_view = lhs_matrix_view(lhs_data);
+    std::optional<matrix_view_type> rhs_view = rhs_matrix_view(rhs_data);
```
ThreeMonth03, line 1028: Replace `auto` with explicit type like `std::optional<matrix_view_type>`.

ThreeMonth03 (commented):

cpp/solvcon/buffer/matmul.hpp
yungyuc: Thank you.  Beautiful.

yungyuc (approved):

yungyuc:
@ThreeMonth03 Please help fix https://github.com/solvcon/solvcon/actions/runs/31809605955 .

PR #1288, opened 2026-08-07
Drive an agent turn as a bounded loop

tigercosmos:
`Turn` in `solvcon/agent/_turn.py` drives one user request to a stop under a step budget. A turn can now take more than one exchange, so the model sees the result of its own commands.

A turn was a single exchange: ask once, apply the reply, stop. A command that failed ended the turn instead of becoming something the model could correct. A GUI could not pump the exchange without blocking its own thread.

`next_request` composes the request on the caller's thread and freezes the scene and a state token into a `TurnRequest`. `feed` re-checks that token, applies the reply, and decides whether the turn continues. The split keeps the slow backend call on a worker, while every state read and every command stays on the thread that owns the state. `run_turn` loops the pair for a caller that can block. The default budget of two buys one action plus one repair; a budget of one is a single shot.

The stop rules read the statuses that the previous step added. A non-`ok` outcome, an empty batch, and prose each end the turn. A malformed batch costs a step and returns to the model with the parse error. A command that fails is an observation the model can act on. Exhausted steps get a marker, not silence.

The state token is the seam the pilot will use. Headless it holds the identity of the bound world, the world object, and a digest of the scene. It keeps the object because CPython gives a freed address to the next allocation. If the token changes between request and reply, the loop drops the commands and ends the turn. A reply cannot land on a canvas the user changed.

Three smaller changes come with the loop. `BackendResponse` resolves an error-only reply to `transport` instead of `empty`, which a driver could not tell from the model saying it is done. `history` gains `skip`, so the loop leaves out the prompt it recorded once that prompt is no longer trailing. `mark` is public and guards an empty transcript, where it raised `IndexError` before. `AgentSession.run_turn` is removed rather than kept as a second way to drive a turn; nothing called it.

The pilot GUI does not use `Turn` yet. Step 4 of #1206 moves it on: the panel calls `next_request` and `feed` on the Qt main thread and keeps `send` in the worker. That step also replaces the token with the active window, the world, and the view transform, and wires a Stop control to `stop()`. Step 5 adds the headless entry point and the scripted smoke tests. Both call `run_turn`, and the error-recovery test needs the step budget.

The scripted backend in the tests is a private double, not a shipped class. Of about 620 diff lines, `tests/test_agent_turn.py` is close to half.

`make pytest` passes: 2142 passed, 14 skipped, 3 xfailed. `make lint` passes.

Related to #1136, #966.

### DIFF
--- a/solvcon/agent/_turn.py
+++ b/solvcon/agent/_turn.py
@@ -1,0 +1,197 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+The bounded turn loop of the Agent.
+
+:class:`Turn` drives one user request to a stop under a step budget.
+:meth:`Turn.next_request` freezes scene and token on the caller's thread;
+:meth:`Turn.feed` applies the reply and decides whether to continue.  Splitting
+the two lets a GUI keep the slow backend call on a worker while state reads and
+commands stay on the owning thread.  :func:`run_turn` loops the pair on one
+thread; a budget of one is a single shot.  No Qt is imported.
+"""
+
+import enum
+import hashlib
+
+from . import _backend
+
+
+class StopReason(enum.Enum):
+    """Why a turn ended.
+
+    ``COMPLETED`` is an empty batch, ``PROSE`` is words instead of commands,
+    ``BUDGET`` is steps exhausted with work still proposed, ``STATE`` is a
+    token mismatch, and ``STOPPED`` is an external halt between steps.  The
+    four transport reasons share names with
+    :class:`~solvcon.agent.TransportOutcome`, so a non-``ok`` outcome maps
+    straight onto one.
+    """
+
+    COMPLETED = "completed"
+    PROSE = "prose"
+    BUDGET = "budget"
+    STATE = "state"
+    STOPPED = "stopped"
+    NO_BACKEND = "no backend"
+    TRANSPORT = _backend.TransportOutcome.TRANSPORT.value
+    TIMEOUT = _backend.TransportOutcome.TIMEOUT.value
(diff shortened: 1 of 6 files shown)


solvcon/agent/_turn.py
```diff
@@ -0,0 +1,197 @@
...
+    apart.  The world itself rides along so that identity stays valid: an id
+    outlives the object it named, and CPython hands the address straight back
+    to the next allocation.  The digest sees only what the scene summary says,
+    so a change the summary does not show does not trip it.
+    """
+    digest = hashlib.sha256(scene.encode("utf-8")).hexdigest()
+    return (id(session.world), session.world, digest)
+
+
+class Turn:
```
tigercosmos, line 65: `Turn` is not used yet by the agent in this PR.

solvcon/agent/_turn.py
```diff
@@ -0,0 +1,197 @@
...
+    def next_request(self):
+        """The next :class:`~solvcon.agent.TurnRequest`, or ``None`` when done.
+
+        Exhausting a multi-step budget records a marker so the transcript does
+        not look like the model went silent; a one-shot budget of one records
+        nothing, because a single step is the whole turn.
+        """
+        if self.done:
+            return None
+        if self._steps >= self._budget:
```
tigercosmos, line 107: Too many steps, stop the turn.

solvcon/agent/_turn.py
```diff
@@ -0,0 +1,197 @@
...
+        self._pending = _NO_REQUEST
+        self._finish(reason, None)
+
+    def _finish(self, reason, note):
+        self._stop = reason
+        if note:
+            self._session.mark(note)
+
+
+def run_turn(session, prompt, budget=2, scene=None, token=None):
```
tigercosmos, line 171: The implementation of executing a turn.

PR #1290, opened 2026-08-08
Reorganize code that conducts the run of obsrefl

### LINKED ISSUE #1269: Make the Analysis Library (polish issue 2)
Develop `_analysis.py` and `_session.py` in the application package.  The code will provide analysis capabilities.

The analysis carries the geometric zone masks cut from `ObliqueShock.shock_path()` with a documented margin, the per-zone report against `zone_states()`, the incident-shock-angle fit (the mid-density crossing per geometric column bin, then a line fit), the reflection-point readout, and horizontal line profiles paired with the analytic three-zone step profile.  (PR #1289)

The session, `ReflectionSession`, owns the driver and the analysis, advances the march in chunks, keeps a bounded history, detects the steady state (a relative drop plus a flat-plateau patience), and stops on steady, on a step cap, or on request.  Fast-lane tests pin the behavior to the prototype numbers recorded in the master plan's study section.  (PR #1290 #1296)

yungyuc:
For issue #1269.

The oblique-shock reflection problem has steady-state solution (that can be analytically computed).  The time-accurate CESE method (which is always time-accurate) will use many time steps to reach the steady state.

This PR makes `ReflectionSession` to hold the driver, the analysis, and a bounded history of the residual and the mass.  It stops on the steady state, on a step cap, or on request.  It allows to drive the run from the GUI timer, a script loop, and a test in the same way.

A unit test case is added to reach the steady state (numerical) solution quickly.  It takes about 0.1 second on Apple M1.

### DIFF
--- a/solvcon/pilot/apps/obsrefl/_session.py
+++ b/solvcon/pilot/apps/obsrefl/_session.py
@@ -1,0 +1,225 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Run the oblique-shock reflection to a steady state.
+
+A reflection run is one long march that has to be watched: it is driven a
+chunk at a time so a caller stays responsive between chunks, it has to keep
+what each chunk measured, and it has to know when marching further buys
+nothing.  :class:`ReflectionSession` holds all three, so the GUI timer, a
+script loop, and a test drive the same run the same way::
+
+    sess = ReflectionSession(mach=3.0, angle=10.0)
+    while not sess.finished:
+        sess.advance()
+    sess.zone_info()
+"""
+
+import collections
+
+from ._analytic import Reflection
+from ._driver import ObliqueShock
+
+__all__ = [
+    'ReflectionSession',
+    'RunHistory',
+    'RunRecord',
+    'SteadyDetector',
+]
+
+
+#: What one marched chunk leaves behind: the step it ended on, the density
+#: residual there, and the total mass, which flattens as the run settles.
+RunRecord = collections.namedtuple('RunRecord', ['step', 'residual', 'mass'])
+
+
+class RunHistory(object):
+    """Keep the newest ``length`` records of a run.
+
+    A run marches for as long as the user lets it, so the history is bounded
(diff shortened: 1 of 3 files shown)


solvcon/pilot/apps/obsrefl/_session.py
```diff
@@ -0,0 +1,225 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+
+"""
+Run the oblique-shock reflection to a steady state.
```
yungyuc, line 6: Even though the docstrings say the run to a steady state, the run may reach it or not.  But we do not need to be too critical to the docstrings now.

solvcon/pilot/apps/obsrefl/_session.py
```diff
@@ -0,0 +1,225 @@
...
+    :ivar step: Steps marched so far.
+    :ivar stop_reason: What ended the run, or None while it runs.
+    """
+
+    #: Default :attr:`steps_per_chunk`, one frame of the GUI timer.
+    STEPS_PER_CHUNK = 5
+    #: Default :attr:`max_steps`.
+    MAX_STEPS = 2000
+
+    def __init__(self, gamma=1.4, density=1.0, pressure=1.0, mach=3.0,
```
yungyuc, line 141: The arguments and member data may need frequent change as we develop the problem.  The numerical solver is usually static (if there is not a bug).  But the controller for the applications (of the solver) changes to meet new needs.

tests/apps/obsrefl/test_session.py
```diff
@@ -0,0 +1,167 @@
...
+    magnitude from the peak to count, and waiting out twenty flat chunks
+    only confirms what five already showed.
+
+    The errors are percents on a mesh this coarse.  They shrink with
+    refinement, which is what a resolution control is for, and the bounds
+    below are loose enough to leave that to the panel rather than pin it.
+    """
+
+    def test_a_coarse_run_settles_onto_the_analytic_answer(self):
+        sess = ReflectionSession(nx=12, ny=4, time_increment=4.e-2,
```
yungyuc, line 149: Use a small mesh to get convergence quickly.

PR #1297, opened 2026-08-08
Detect `ccache` and use it as the compiler launcher

tigercosmos:
A local build never used a compiler cache. CI reaches ccache through the symlinks `hendrikmuhs/ccache-action` drops in `/usr/local/bin` under `create-symlink: true`, which a developer's `PATH` does not carry, so `make` after `make cmakeclean` recompiled every translation unit from scratch.

`CMakeLists.txt` now looks for `ccache` and sets `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER` when it finds one. Detection lives there rather than in `CMakePresets.json` so a host without ccache configures and builds exactly as before, and so a direct `cmake --preset` or IDE build gets it too. `USE_CCACHE` carries its default in the `base` preset alongside the other knobs, and `make USE_CCACHE=OFF` opts out.

The guard skips the cases where the wrap cannot pay. An explicit launcher wins, which is what keeps the `sccache` of `ci-win-mkl`. A compiler that is already ccache wins too, which is the shape of the Linux and macOS CI jobs: wrapping one makes the outer cache hash the ccache binary in place of the real compiler and report the compiler type as `other` rather than `clang`. MSVC is skipped because ccache refuses its default debug information format.

On a 14-core M-series macOS host, `make cmakeclean && make` took 1:09.68 wall against an empty cache and 7.86s against a warm one, with ccache reporting 114 of 115 compiles as direct hits. Both are single runs on a machine that was not idle, so they are orders of magnitude rather than measurements. The one miss is the Qt resource unit `qrc_solvcon_pilot_shaders.cpp`, which `qsb` regenerates on every configure. `make lint`, `make pytest`, and `make gtest` are clean.


CMakeLists.txt
```diff
@@ -33,6 +33,40 @@ if(MSVC)
...
+# without it configures and builds unchanged. The guard skips every case where
+# the wrap cannot pay: an explicit launcher, such as the sccache of the
+# ci-win-mkl preset; a compiler that is already ccache, which is how the Linux
+# and macOS CI jobs install it and where wrapping would hash ccache in place of
+# the real compiler; and MSVC, whose default debug information format ccache
+# refuses, leaving every unit to pay the wrap and return no hit.
+option(USE_CCACHE "use ccache as the compiler launcher when it is available" ON)
+message(STATUS "USE_CCACHE: ${USE_CCACHE}")
+file(REAL_PATH "${CMAKE_CXX_COMPILER}" CCACHE_CXX_COMPILER_REALPATH)
+if(USE_CCACHE
```
tigercosmos, line 46: Use ccache is it is in the system (only for macOS and Unix for now). We may need to modify the CI files in the following PRs.

tigercosmos (commented):
@chestercheng  @yungyuc  Please take a look, thanks!

PR #1298, opened 2026-08-08
Rename the `MM_` macros and the `MMGH_` repository variables

### LINKED ISSUE #1295: Rename remaining mm/modmesh to sc/solvcon
The modmesh to solvcon rename already covered the C++ namespace, the source tree, the include paths, the `MODMESH_*` macros, and the CMake targets, through the two scripts in `contrib/lint/`. This issue tracks what those scripts could not reach.

## Part 1: the `MM_` macros and the `MMGH_` repository variables

Rename the `MM_` preprocessor macros to a single `SC_` prefix. They come in two shapes today, `MM_DECL_X` and `DECL_MM_X`, and both should fold into `SC_DECL_X`. `STYLE.md` states the old prefix as the rule for new macros, so it changes with them. Some of the names are CMake compile definitions rather than in-source macros and have to move together with the guards that test them.

Rename the `MMGH_*` GitHub Actions repository variables to `SCGH_*`. A repository variable lives in the GitHub settings, so no commit can rename one that is already configured: each expression reads the new name and falls back to the old. Once the repositories carry the new variables, a follow-up drops the fallback.

PR list:
- PR #1298

## Part 2: other `mm`/`modmesh` code

- Lowercase `mm_` names in the Python tests, each paired with an `np_` counterpart, so `sc_` is the spelling to use.
- The `mm` namespace alias in the gtest sources.
- `MM`-spelled macro parameters that a macro rename passes by.
- `Modmesh`-prefixed identifiers that a caller can reach, in the Python package and across the binding layer. These want a compatibility decision, not a bare rename.
- Stale `modmesh` prose in comments. A genuine cross-project citation stays as it is.

PR list:
- PR #1313

yungyuc:
Rename the `MM_` C++ preprocessor macros to the `SC_` prefix.  Some macros used an unconventional name like `DECL_MM_X`, and are renamed to `SC_DECL_X` too.

Update `STYLE.md` for the new convention (of using `SC`).

Rename the `MMGH_*` GitHub Actions repository variables to `SCGH_*`. The old `MMGH_*` will stay in the configuration files as fallback for a while.

For the part 1 of issue #1295.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -21,9 +21,9 @@ concurrency:
 jobs:
 
   check_skip:
-    # A fork inherits no MMGH_NIGHTLY, so this gate stops the nightly cron
+    # A fork inherits no SCGH_NIGHTLY, so this gate stops the nightly cron
     # from allocating a runner there.
-    if: github.event_name != 'schedule' || vars.MMGH_NIGHTLY == 'enable'
+    if: github.event_name != 'schedule' || (vars.SCGH_NIGHTLY || vars.MMGH_NIGHTLY) == 'enable'
     uses: ./.github/workflows/check_skip_ci.yml
     permissions:
       contents: read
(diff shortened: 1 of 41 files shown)


STYLE.md
```diff
@@ -625,7 +625,7 @@ namespace detail
 
 ## C Pre-Processor Macro
 
-Prefix macros with `MM_DECL_`. If they are not supposed to be used as a
+Prefix macros with `SC_DECL_`. If they are not supposed to be used as a
```
yungyuc, line 628: New macro name convention to use `SC_*` instead of `MM_*`.

doc/source/devguide/testing.md
```diff
@@ -97,40 +97,40 @@ These variables tune the workflows. Set them as repository variables (under
 Settings, then Secrets and variables, then Actions, then the Variables tab).
 Each is read with a default, so an unset variable keeps the default behavior.
 
-- `MMGH_NIGHTLY`: set to `enable` to let the nightly schedule run its jobs.
+- `SCGH_NIGHTLY`: set to `enable` to let the nightly schedule run its jobs.
```
yungyuc, line 100: Update the document for the github variable names.  The old `MMGH_*` are kept as fallback in the yaml files. They are not mentioned here to encourage to use the new names.

.github/workflows/devbuild.yml
```diff
@@ -21,9 +21,9 @@ concurrency:
 jobs:
 
   check_skip:
-    # A fork inherits no MMGH_NIGHTLY, so this gate stops the nightly cron
+    # A fork inherits no SCGH_NIGHTLY, so this gate stops the nightly cron
```
yungyuc, line 24: Comments do not mention `MMGH_*` from now on.  After the new names stabilize, the old names will be removed from the fallback for github.

yungyuc (commented):
The PR touches many files, but the change is simple.

@ThreeMonth03 this might disrupt your code change in C++.

PR #1300, opened 2026-08-08
Default build to `RelWithDebInfo` and remove `DEBUG_SYMBOL`

tigercosmos:
`DEBUG_SYMBOL` defaulted to `ON` in the `base` preset, so a `Release` build compiled `-O3 -g -DNDEBUG`: what this project called Release was RelWithDebInfo under another name, and CMake's actual `Release` had no way to be spelled. This deletes the option and lets `CMAKE_BUILD_TYPE` carry the axis it already owns.

| `CMAKE_BUILD_TYPE` | Build tree | GCC/Clang flags |
|:--|:--|:--|
| `RelWithDebInfo` (new default) | `build/reldbg<pyvminor>` | `-O3 -g -DNDEBUG` |
| `Release` | `build/rel<pyvminor>` | `-O3 -DNDEBUG` |
| `Debug` | `build/dbg<pyvminor>` | `-g`, no `-O` |

The default compiles what the old default compiled, so no build gets slower. Both preset bases pin the RelWithDebInfo flags to their Release counterparts (`-O3` for GCC and Clang, `/O2 /Ob2` for MSVC) because CMake's own values optimize a step lower and a rename should not cost an optimization level. `dev-reldbg` and `win-reldbg` configure the default and carry the build and test presets that `dev-rel` and `win-rel` used to; those two now configure a true `Release`. The `contrib/cmake` templates and `build.ps1` follow the default so IDE and Windows builds keep their symbols.

Two things change rather than staying put. pybind11 strips the extension module for every build type except Debug, RelWithDebInfo, and None, so the old default shipped a stripped `_solvcon` that a debugger could not see into; under the new default it is left alone, and `Release` still strips it. And the deleted option appended `/DEBUG` to `CMAKE_CXX_FLAGS` on MSVC, where `/DEBUG` is a linker option and `/D` defines a macro, so every Windows build has been defining `EBUG` and getting no debug information for it.

The sanitizer jobs move to `RelWithDebInfo`, since they asked for `Release`, were handed symbols anyway, and a sanitizer report names a file and a line only when it has them. The other CI jobs keep asking for `Release` and now get what they asked for. A `CMAKE_BUILD_TYPE` that is none of the three stops the build instead of quietly configuring the default.

Verified on macOS with Python 3.14: `make lint` clean; `make pytest` 2082 passed, 13 skipped, 3 xfailed; `make gtest` 238 passed. All three build types configure and build, and the flags land as tabled above. The pilot carries 113 debug-map entries under the default and none under `Release`; the extension module carries 112 and none respectively. Preset cross-references all resolve, including the Windows ones this host cannot configure.

Everyone gets one full rebuild when the default build type flips the cache, and `build/rel<pyvminor>` changes meaning for anyone who already has one. A `setup.mk` still setting `DEBUG_SYMBOL` is ignored rather than refused.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -48,6 +48,10 @@ jobs:
 
     timeout-minutes: ${{ fromJSON(vars.SCGH_TIMEOUT_STANDALONE_BUFFER || vars.MMGH_TIMEOUT_STANDALONE_BUFFER || '10') }}
 
+    env:
+      # The build type names this job's scratch directory and its cache key.
+      CMAKE_BUILD_TYPE: ${{ matrix.cmake_build_type }}
+
     strategy:
       matrix:
         os: [ubuntu-24.04]
@@ -120,6 +124,12 @@ jobs:
       # macOS runner that swaps until the job hits its timeout. -j2 is the
       # safe bet on both.
       MAKE_PARALLEL: -j2
+      # Every step in this job builds one build type, and the Makefile derives
+      # the build tree from it. Setting it here rather than per step keeps a
+      # bare `make buildext` on the tree the preceding `make cmake`
+      # configured; naming it in only some of them splits the job across two
+      # trees, and the second one configures from the preset defaults.
+      CMAKE_BUILD_TYPE: ${{ matrix.cmake_build_type }}
       QT_DEBUG_PLUGINS: 1
       # QT_QPA_PLATFORM is set per-OS by the setup actions (xcb on Linux via
       # Xvfb, offscreen on macOS). Do not set it at the job level.
@@ -130,8 +140,15 @@ jobs:
 
     strategy:
         matrix:
-          os: [ubuntu-24.04, macos-26]
-          cmake_build_type: [Release]
+          # One build type per host rather than the same one twice. macOS
+          # takes RelWithDebInfo, which is what a bare `make` configures and
+          # what no other job on a pull request builds; ubuntu and the
+          # Windows job cover Release.
+          include:
+            - os: ubuntu-24.04
+              cmake_build_type: Release
+            - os: macos-26
+              cmake_build_type: RelWithDebInfo
 
         fail-fast: false
 
@@ -517,7 +534,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04]
-        cmake_build_type: [Release]
+        # Sanitizer reports name a file and a line only when the binary
+        # carries debug symbols, so this job builds RelWithDebInfo.
+        cmake_build_type: [RelWithDebInfo]
 
       fail-fast: false
 
@@ -587,8 +606,9 @@ jobs:
 
     strategy:
       matrix:
+        # No build type here: this job configures through the ci-win-asan
+        # preset, which states its own.
         os: [windows-2022]
-        cmake_build_type: [Release]
 
       fail-fast: false
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,16 @@ project(solvcon)
 
 enable_testing()
 
+# The presets and the Makefile always state a build type. A configure that
+# names neither, which is a bare `cmake -S . -B <dir>`, would otherwise get an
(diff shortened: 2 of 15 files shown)


.github/workflows/devbuild.yml
```diff
@@ -516,7 +516,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04]
-        cmake_build_type: [Release]
+        # Sanitizer reports name a file and a line only when the binary
+        # carries debug symbols, so this job builds RelWithDebInfo.
+        cmake_build_type: [RelWithDebInfo]
```
tigercosmos, line 539: use `RelWithDebInfo` for `devbuild`.

contrib/cmake/CMakeUserPresets.example.json
```diff
@@ -5,7 +5,7 @@
   "configurePresets": [
     {
       "name": "ide-local-rel",
-      "inherits": "dev-rel",
+      "inherits": "dev-reldbg",
```
tigercosmos, line 8: Default to `RelWithDebInfo`, so change the name.

doc/source/start/build_solvcon.md
```diff
@@ -8,30 +8,59 @@ up without installation.
...
+## Build Types
+
+`CMAKE_BUILD_TYPE` selects one of CMake's three standard build types.  Each
+gets its own build tree, so switching between them never invalidates another
+one's cache.  `<pyvminor>` is the active Python major and minor version, e.g.
+`314`.
+
+| `CMAKE_BUILD_TYPE` | Build tree               | Flags             |
+|:-------------------|:-------------------------|:------------------|
+| `RelWithDebInfo`   | `build/reldbg<pyvminor>` | `-O3 -g -DNDEBUG` |
```
tigercosmos, line 20: New added `RelWithDebInfo`

CMakeLists.txt
```diff
@@ -247,16 +247,6 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE BOOL "Export compile commands")
 option(HIDE_SYMBOL "hide the symbols of python wrapper" OFF)
 message(STATUS "HIDE_SYMBOL: ${HIDE_SYMBOL}")
 
-option(DEBUG_SYMBOL "add debug information" ON)
-message(STATUS "DEBUG_SYMBOL: ${DEBUG_SYMBOL}")
```
tigercosmos, line 251: No need `DEBUG_SYMBOL` anymore.

Makefile
```diff
@@ -27,22 +29,28 @@ RUNENV += PYTHONPATH=$(SOLVCON_ROOT)
...
 	USE_SANITIZER USE_CCACHE CMAKE_INSTALL_PREFIX \
 	CMAKE_LIBRARY_OUTPUT_DIRECTORY CMAKE_PREFIX_PATH
 CMAKE_OVERRIDES = $(strip $(foreach knob,$(CMAKE_KNOBS), \
 	$(if $(filter-out undefined default,$(origin $(knob))),\
 	-D$(knob)=$($(knob)))))
 
-CMAKE_BUILD_TYPE ?= Release
+# Debugging is the default story. Release drops the symbols for a session
+# that is not going to open a debugger.
+CMAKE_BUILD_TYPE ?= RelWithDebInfo
```
tigercosmos, line 42: Default to RelWithDebInfo

.github/workflows/devbuild.yml
```diff
@@ -130,8 +138,15 @@ jobs:
...
-          cmake_build_type: [Release]
+          # One build type per host rather than the same one twice. macOS
+          # takes RelWithDebInfo, which is what a bare `make` configures and
+          # what no other job on a pull request builds; ubuntu and the
+          # Windows job cover Release.
+          include:
+            - os: ubuntu-24.04
+              cmake_build_type: Release
+            - os: macos-26
+              cmake_build_type: RelWithDebInfo
```
tigercosmos, line 151: Let macos uses `RelWithDebInfo` and let ubuntu keep `Release`

yungyuc (commented):
Pretty good.  Have you used the cmake presets with vscode?  How well does it work?

PR #1302, opened 2026-08-08
Let each build system find ccache for itself

tigercosmos:
Since #1297 `CMakeLists.txt` sets `CMAKE_CXX_COMPILER_LAUNCHER` when it finds `ccache`, so `create-symlink: true` on `ccache-action` left two ways to the same cache. CI took the symlink one, where the compiler CMake found was ccache itself, so the jobs never exercised the path a developer builds through. Dropping the symlinks leaves each build system to find the cache on its own and stops CI manipulating `PATH` for it.

`standalone_buffer` keeps them, because `contrib/standalone_buffer/Makefile` compiles with a bare `c++` and has no discovery step of its own. The `CMakeLists.txt` comment from #1297 is corrected here too, since this change is what made its claim about the CI jobs false.

A second commit gives `nouse_install` its own cache key. It shared the plain `<os>-Release` key with the `build` job while building `BUILD_QT=OFF` under the `setup.py` build directory, so neither its flags nor its paths matched what was stored there and every object missed.

Verified on a fork over two successive push runs, with the nightly-only jobs forced on. Everything passed both rounds. Round two reported 227/229 direct hits on the ubuntu-24.04 build, 114/115 on `profiling` for both ubuntu-24.04 and macos-26, and 115/117 on ubuntu `lint`, with `CMAKE_CXX_COMPILER` now `/usr/bin/c++`, so the launcher wraps the real compiler. Expect the first master push after merge to run cold and repopulate.

Related to #1297.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -75,6 +75,8 @@ jobs:
         with:
           key: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           restore-keys: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
+          # contrib/standalone_buffer/Makefile compiles with a bare `c++`
+          # outside CMake, so the symlinks are this job's only route to ccache.
           create-symlink: true
           save: ${{ github.event_name != 'pull_request' }}
           # Bound the cache so it does not grow without limit across the rolling
@@ -160,7 +162,6 @@ jobs:
       with:
         key: ${{ runner.os }}-${{ matrix.cmake_build_type }}
         restore-keys: ${{ runner.os }}-${{ matrix.cmake_build_type }}
-        create-symlink: true
         save: ${{ github.event_name != 'pull_request' }}
         # Bound the cache so it does not grow without limit across the rolling
         # key. A master push or nightly run saves this cache; pull requests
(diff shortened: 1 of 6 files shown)


.github/workflows/devbuild.yml
```diff
@@ -75,6 +75,8 @@ jobs:
         with:
           key: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
           restore-keys: ${{ runner.os }}-standalone-buffer-${{ matrix.cmake_build_type }}
+          # contrib/standalone_buffer/Makefile compiles with a bare `c++`
+          # outside CMake, so the symlinks are this job's only route to ccache.
```
tigercosmos, line 79: Caveat.

.github/workflows/lint.yml
```diff
@@ -102,7 +102,6 @@ jobs:
       with:
         key: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
         restore-keys: ${{ runner.os }}-tidy-${{ matrix.cmake_build_type }}
-        create-symlink: true
```
tigercosmos, line 105: No need to create symlink after #1297

.github/workflows/nouse_install.yml
```diff
@@ -86,9 +86,12 @@ jobs:
...
         with:
-          key: ${{ runner.os }}-Release
-          restore-keys: ${{ runner.os }}-Release
-          create-symlink: true
+          # This job builds `BUILD_QT=OFF` under the `setup.py` build
+          # directory, so its objects match neither the flags nor the paths
+          # that the `build` job stores under the plain `<os>-Release` key.
+          # Sharing that key only restores a cache it cannot hit.
+          key: ${{ runner.os }}-nouse-Release
+          restore-keys: ${{ runner.os }}-nouse-Release
```
tigercosmos, line 94: `nouse_install` is missing for ccache.

PR #1305, opened 2026-08-08
Keep Qt out of the core library

tigercosmos:
`solvcon_primary` was one library under both settings of `BUILD_QT`, and it linked Qt `PUBLIC`, so Qt's include directories and `QT_*_LIB` definitions landed on every unit in the core. The `build` job of `devbuild` compiles `BUILD_QT=OFF` and then `BUILD_QT=ON` in the same build directory, so the second pass found a changed command line on every core object and rebuilt the whole core.

Only the pilot needs Qt. Every `#include <Q...>` and every `Q_OBJECT` under `cpp/solvcon` sits in `pilot/`, and the one unit outside it that varies with the setting is `python/module.cpp`, which registers the pilot submodule behind `QT_CORE_LIB`.

So the one library becomes three:

- `solvcon_primary`, the core, never sees Qt.
- `solvcon_graphic`, the pilot, is the only target that links Qt. It carries moc, rcc, and uic as target properties instead of directory properties, which used to hand the core an autogen include directory too.
- `solvcon_entry` is `module.cpp` alone. The extension module and the pilot binary link it and reach the rest through it, and `QT_CORE_LIB` still decides what it compiles.

```
    _solvcon (Python ext)      pilot (binary)
              \                  /
               solvcon_entry              module.cpp, knows whether Qt is here
                /          \
    solvcon_graphic   solvcon_primary     the pilot (Qt)   the core (no Qt)
```

Units compiled, replaying the job's step sequence:

| step | master | this PR |
| --- | --- | --- |
| `make buildext BUILD_QT=OFF` | 72 | 72 |
| `make buildext BUILD_QT=ON` | 115 | 45 |
| `make pilot` | 3 | 3 |

Of the units the two passes share, 57 changed command line before and 2 change now.

`make buildext BUILD_QT=ON` on a runner, this branch's run 31270521406 against master's run 31270363328:

| runner | master | this PR |
| --- | --- | --- |
| ubuntu-24.04 Release | 9m38s | 2m25s |
| macos-26 RelWithDebInfo | 1m58s | 1m32s |

The two runs held different ccache contents. The untouched `BUILD_QT=OFF` step moved by 85s on ubuntu and 54s on macos, so the ubuntu gain is real and the macos gain is below that floor. Against an isolated ccache the pass takes 51s before and 27s after when the cache holds only the `BUILD_QT=OFF` pass, and 5s either way once both settings are cached.

Two notes for review. The per-target compile options now cover all three libraries, because the pilot sources and `module.cpp` need the pybind11 and Accelerate suppressions and `wrap_pilot.cpp` may want `/bigobj`. And `pilot_clang_tidy_diff` collects its file list from all three, since the pilot sources it lints no longer live in `solvcon_primary`.

### DIFF
--- a/cpp/solvcon/CMakeLists.txt
+++ b/cpp/solvcon/CMakeLists.txt
@@ -4,10 +4,6 @@
 cmake_minimum_required(VERSION 4.0.1)
 
 if (BUILD_QT)
-    set(CMAKE_AUTOMOC ON)
-    set(CMAKE_AUTORCC ON)
-    set(CMAKE_AUTOUIC ON)
-
     find_package(Qt6 QUIET)
     if(NOT Qt6_FOUND)
         if(MSVC)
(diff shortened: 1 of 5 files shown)


cpp/solvcon/CMakeLists.txt
```diff
@@ -149,40 +176,59 @@ if (BUILD_QT)
 
     # Bake the pilot GLSL shaders to .qsb and embed them in the Qt resource
     # system under :/solvcon/pilot/shaders/. Loaded by RMaterial at runtime.
-    qt6_add_shaders(solvcon_primary "solvcon_pilot_shaders"
+    qt6_add_shaders(solvcon_graphic "solvcon_pilot_shaders"
```
tigercosmos, line 179: Split QT related to a new `solvcon_graphic`.

cpp/solvcon/CMakeLists.txt
```diff
@@ -109,16 +105,47 @@ set(SOLVCON_GRAPHIC_FILES
...
+# that differed by the pilot but a command line that differed everywhere:
+# linking Qt PUBLIC put its include directories and its QT_*_LIB definitions
+# on every unit in the core. A build that compiles both settings, which the
+# devbuild workflow does, therefore compiled the core twice over and shared
+# nothing between the passes, not even through ccache.
+#
+# Three libraries instead of one. `solvcon_primary` is the core and never sees
+# Qt. `solvcon_graphic` is the pilot and is the only target that does.
+# `solvcon_entry` is the single unit that has to know which of the two it was
+# built with. Only the last two are rebuilt when BUILD_QT changes.
```
tigercosmos, line 119: Meat.

cpp/solvcon/CMakeLists.txt
```diff
@@ -109,16 +105,47 @@ set(SOLVCON_GRAPHIC_FILES
...
+    STATIC
+    ${SOLVCON_TERMINAL_FILES}
+)
+
+set(SOLVCON_LIBRARIES solvcon_primary)
+
 if (BUILD_QT)
     qt_add_library(
-        solvcon_primary
+        solvcon_graphic
```
yungyuc, line 130: I like the separation.  Nit: I do not particularly like `graphic` for the name, but do not have a better idea either.  Let's keep it.

yungyuc (commented):
LGTM.

Another possible separation is Python.  We have a very clean design keeping C++ and Python (binding) code separate, but they are built together.  Separation could make the wrapping code too far away from the lib code, so maybe we should not do it.  It's just a topic that we can think about.  Definitely we should not do it without thorough and lengthy discussions.

PR #1307, opened 2026-08-08
Add the Strassen GEMM kernel

### LINKED ISSUE #1303: Add Strassen matmul dispatch
## Motivation

Planned matmul selects one execution kernel before traversal, but it cannot select Strassen. Calls therefore use generic, BLAS, or packing routes even for shapes where Strassen is faster.

Apple M1 measurements against NumPy 2.5.1 show depth 1 at 1.118x for float32 `3072 x 24576 x 3072` and depth 2 at 1.164x for float64 `6144 x 6144 x 6144`. Float32 `6144 x 6144 x 6144` and low-`K` shapes favor direct Accelerate. Selection therefore needs dtype and `M x K x N` entries instead of a size threshold.

## Proposed design

Extend the execution selection from [solvcon/solvcon#1259](https://github.com/solvcon/solvcon/pull/1259):

```text
MatmulPlan + MatmulTuning
          |
          v
MatmulSelection
    +-- StrassenGemm1
    +-- StrassenGemm2
    `-- existing kernel
          |
          v
MatmulExecutor -> reusable workspace -> BLAS GEMM leaves
```

`MatmulTuning` maps dtype and `M x K x N` entries to `StrassenGemm1` or `StrassenGemm2`. `MatmulSelection` records the kernel once. `MatmulExecutor` runs it with thread-local workspace and BLAS leaves. `MatmulPlan` remains a description of shapes and strides. The Apple arm64 table covers compact, non-batched float32 and float64 GEMM; calls outside the table use the existing selection.

The initial `StrassenGemm1` and `StrassenGemm2` candidates require `M`, `K`, and `N` to be divisible by `2^depth`. This issue does not materialize whole padded matrices, so non-divisible calls retain the existing route. A future generated policy may introduce padded Strassen as a separate candidate with its own measured copy, memory, and padded-work features.

## Implementation outline

- [x] **Task 1: Add the Strassen kernel and workspace.** Implement depth-1 and depth-2 rectangular GEMM with BLAS leaves. Test float32 and float64 correctness, both depths, rectangular dimensions, invalid divisibility, and workspace reuse. (Solved by #1307, #1317)
- [x] **Task 2: Add and measure Strassen dispatch.** Add `StrassenGemm1` and `StrassenGemm2` to `MatmulTuning` and `MatmulSelection`, then dispatch them from `MatmulExecutor` with its workspace. Profile both depths through planned matmul against NumPy and direct Accelerate, and enable only entries that pass the residual bound and beat both. Test selected and existing routes, then run the test, lint, and Apple profiling suites. (Solved by #1317)

## Global acceptance

- Planned matmul selects `StrassenGemm1` or `StrassenGemm2` without changing `MatmulPlan` semantics.
- `MatmulExecutor` reuses workspace and executes the recorded kernel.
- Float32 and float64 results remain within the documented scaled residual bound.
- Every table entry outperforms NumPy and direct Accelerate under the Apple protocol; calls outside the table use the existing selection.
- Initial Strassen candidates check divisibility before selection and do not allocate whole padded matrices.

## References

- [Matmul dispatch prototype, ThreeMonth03/solvcon#30](https://github.com/ThreeMonth03/solvcon/pull/30)
- [Apple M1 measurements with NumPy 2.5.1](https://github.com/ThreeMonth03/solvcon/issues/31#issuecomment-5226996668)
- [Whole-matrix padding control](https://github.com/ThreeMonth03/solvcon/blob/codex/strassen-padding-prototype/doc/source/devplan/strassen-dispatch-prototype/padding-results.md)
- [Batched matmul design, solvcon/solvcon#1172](https://github.com/solvcon/solvcon/issues/1172)
- [Matmul selection architecture, solvcon/solvcon#1259](https://github.com/solvcon/solvcon/pull/1259)

ThreeMonth03:
Some large GEMMs benefit from one or two levels of Strassen recursion, while direct BLAS remains preferable for other shapes. Matmul dispatch therefore needs a fixed-depth Strassen kernel that it can select by shape and dtype.

Add the internal kernel for rectangular row-major operands. Each recursion level evaluates the seven Strassen products with reusable workspace and delegates leaf contractions to BLAS. The GEMM descriptor, one-level product evaluation, and recursion driver have separate responsibilities.

Related to task 1 of issue #1303.

### DIFF
--- a/cpp/solvcon/math/Strassen.hpp
+++ b/cpp/solvcon/math/Strassen.hpp
@@ -1,0 +1,449 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Implement a fixed-depth rectangular Strassen GEMM kernel.
+ *
+ * For C = A B, each recursion splits the operands into four blocks and
+ * evaluates Strassen's seven products:
+ *
+ * P1 = (A11 + A22)(B11 + B22), P2 = (A21 + A22)B11
+ * P3 = A11(B12 - B22), P4 = A22(B21 - B11)
+ * P5 = (A11 + A12)B22, P6 = (A21 - A11)(B11 + B12)
+ * P7 = (A12 - A22)(B21 + B22)
+ *
+ * C11 = P1 + P4 - P5 + P7, C12 = P3 + P5
+ * C21 = P2 + P4, C22 = P1 - P2 + P3 + P6
+ *
+ * Inputs are row-major non-transposed views. The output is compact row-major
+ * storage and must not overlap either input.
+ *
+ * Reference:
+ * Volker Strassen, "Gaussian elimination is not optimal," Numerische
+ * Mathematik 13(4), 354-356 (1969).
+ *
+ * @see https://doi.org/10.1007/BF02165411
+ * @ingroup group_core
+ */
+
+#include <solvcon/base.hpp>
+#include <solvcon/math/blas_compat.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <memory>
+#include <stdexcept>
+
+namespace solvcon
+{
+
+namespace detail
+{
+
+namespace strassen
+{
+
+template <typename T>
+struct Gemm
+{
+    ssize_t rows;
+    ssize_t columns;
+    ssize_t inner_size;
+    BlasMatrixView<T> lhs;
+    BlasMatrixView<T> rhs;
+    T * output;
+}; /* end struct Gemm */
+
+/**
+ * @brief Reuse scratch storage along one Strassen recursion path.
+ *
+ * Workspace owns a grow-only buffer and provides stack-like allocation.
+ * prepare() resets the cursor for one multiplication, while mark() and
+ * rewind() bound the storage used by each recursion level. Because each Step
+ * evaluates P1-P7 sequentially, sibling products reuse the same child storage
+ * instead of reserving space for the full recursion tree.
+ *
+ * At depth 2, the buffer holds scratch blocks for two simultaneously active
+ * Steps, rather than all eight non-leaf Steps or 49 leaf products. Its
+ * capacity remains available for later multiplications.
+ */
+template <typename T>
+class Workspace
+{
+public:
+    void prepare(size_t required_size);
+    size_t mark() const noexcept { return m_offset; }
+    T * allocate(size_t count);
+    void rewind(size_t mark) noexcept { m_offset = mark; }
+    size_t capacity() const noexcept { return m_capacity; }
+
+private:
+    std::unique_ptr<T[]> m_storage; // NOLINT(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    size_t m_capacity = 0;
+    size_t m_limit = 0;
+    size_t m_offset = 0;
+}; /* end class Workspace */
+
+template <typename T>
+void Workspace<T>::prepare(size_t required_size)
+{
+    if (required_size > m_capacity)
+    {
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+        m_storage = std::make_unique_for_overwrite<T[]>(required_size);
+        m_capacity = required_size;
+    }
+    m_limit = required_size;
+    m_offset = 0;
+}
+
+template <typename T>
+T * Workspace<T>::allocate(size_t count)
+{
+    if (count > m_limit - m_offset)
+    {
+        throw std::runtime_error("Strassen workspace is too small");
+    }
+    T * block = m_storage.get() + m_offset;
+    m_offset += count;
+    return block;
+}
+
+template <size_t Depth>
+size_t workspace_size(ssize_t rows, ssize_t columns, ssize_t inner_size)
+{
+    if constexpr (Depth == 0)
+    {
+        return 0;
+    }
+    else
+    {
+        rows /= 2;
+        columns /= 2;
+        inner_size /= 2;
+        auto const block_rows = static_cast<size_t>(rows);
+        auto const block_columns = static_cast<size_t>(columns);
+        auto const block_inner_size = static_cast<size_t>(inner_size);
+        size_t const current_size = block_rows * block_inner_size + block_inner_size * block_columns +
+                                    block_rows * block_columns;
+        return current_size + workspace_size<Depth - 1>(rows, columns, inner_size);
+    }
+}
+
+template <size_t Depth, typename T>
+void validate(Gemm<T> const & gemm)
+{
+    static_assert(Depth <= 2, "Strassen GEMM supports recursion depths 0, 1, and 2");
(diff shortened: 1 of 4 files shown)


cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,433 @@
...
+{
+
+namespace detail
+{
+
+namespace strassen
+{
+
+template <typename T>
+struct Gemm
```
ThreeMonth03, line 52: `Gemm` is a POD call descriptor. Its public fields carry one contraction through recursion; validation stays at the kernel entry, while stateful behavior belongs to `Workspace`, `Step`, and `Kernel`.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,433 @@
...
+
+template <typename T>
+void Step<T>::add_product(T * output, T scale) const
+{
+    add_block(output, m_output_stride, m_product.output, m_product.rows, m_product.columns, scale);
+}
+
+template <typename T>
+template <typename Multiply>
+void Step<T>::evaluate(Multiply const & multiply) const
```
ThreeMonth03, line 325: `initialize_output()` writes the first contribution to each uninitialized output quadrant. Later products use `accumulate_output()` with coefficient `1` or `-1`. The order is intentional: P1 initializes C11 and C22, P2 initializes C21, and P3 initializes C12 before subsequent products accumulate into them.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
...
+    if (count > m_limit - m_offset)
+    {
+        throw std::runtime_error("Strassen workspace is too small");
+    }
+    T * block = m_storage.get() + m_offset;
+    m_offset += count;
+    return block;
+}
+
+template <size_t Depth>
```
ThreeMonth03, line 117: `Depth` is a template argument so each supported recursion depth has a fixed call graph. `workspace_size<Depth>()` mirrors that recursion and reserves only the largest live path, rather than storage for all seven products at every level.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
...
+template <size_t Depth, typename T, typename Leaf>
+void multiply(Gemm<T> const & gemm, Workspace<T> & workspace, Leaf const & leaf)
+{
+    Kernel<T, Leaf> kernel(workspace, leaf);
+    kernel.template multiply<Depth>(gemm);
+}
+
+} /* end namespace strassen */
+
+template <size_t Depth, typename T>
```
ThreeMonth03, line 429: This is the integration boundary for future matmul dispatch. Dispatch will select `Depth`, while this wrapper keeps the leaf policy explicit: every terminal contraction is delegated to the existing BLAS GEMM wrapper.

gtests/test_nopython_strassen.cpp
```diff
@@ -0,0 +1,190 @@
...
+void check_depth(size_t expected_leaf_calls)
+{
+    strassen::Workspace<T> workspace;
+    size_t const leaf_calls = run_strassen<T, Depth>(8, 12, 16, workspace);
+    EXPECT_EQ(leaf_calls, expected_leaf_calls);
+}
+
+} /* end namespace */
+
+TEST(StrassenKernel, matches_reference_at_each_depth)
```
ThreeMonth03, line 117: The expected leaf counts verify the recursion topology as well as numerical output: depth `0` calls one leaf, depth `1` calls seven, and depth `2` calls 49. The rectangular `8 x 12 x 16` case also prevents the implementation from assuming square matrices.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+private:
+    std::unique_ptr<T[]> m_storage; // NOLINT(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    size_t m_capacity = 0;
+    size_t m_limit = 0;
+    size_t m_offset = 0;
+}; /* end class Workspace */
+
+template <typename T>
+void Workspace<T>::prepare(size_t required_size)
```
ThreeMonth03, line 93: `m_capacity` records storage retained across calls, while `m_limit` records the exact workspace required by the current multiplication. `prepare()` may reuse a larger allocation, but `allocate()` still checks the current limit so an underestimated recursion plan cannot silently consume stale excess capacity.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
...
+        auto const block_rows = static_cast<size_t>(rows);
+        auto const block_columns = static_cast<size_t>(columns);
+        auto const block_inner_size = static_cast<size_t>(inner_size);
+        size_t const current_size = block_rows * block_inner_size + block_inner_size * block_columns +
+                                    block_rows * block_columns;
+        return current_size + workspace_size<Depth - 1>(rows, columns, inner_size);
+    }
+}
+
+template <size_t Depth, typename T>
```
ThreeMonth03, line 138: This validation defines the narrow kernel contract. Future matmul dispatch should select Strassen only for eligible row-major operands and use another route otherwise; direct internal calls fail here instead of entering recursion with invalid block geometry.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Implement a fixed-depth rectangular Strassen GEMM kernel.
```
ThreeMonth03, line 10: The call flow is `gemm_strassen<Depth>() -> strassen::multiply<Depth>() -> Kernel::recurse<Depth>()`. Each non-leaf level creates one `Step` for the seven products; depth `0` invokes the BLAS leaf. Matmul dispatch remains outside this kernel and chooses `Depth`.

cpp/solvcon/math/Strassen.hpp
```diff
@@ -0,0 +1,423 @@
...
+template <size_t Depth>
+void Kernel<T, Leaf>::recurse(Gemm<T> const & gemm)
+{
+    if constexpr (Depth == 0)
+    {
+        m_leaf(gemm);
+    }
+    else
+    {
+        size_t const mark = m_workspace.mark();
```
ThreeMonth03, line 411: This block executes one Strassen recursion level:

1. `mark()` records the workspace offset before this level allocates scratch storage.
2. `Step` reserves three blocks for the combined lhs, combined rhs, and one product.
3. `evaluate()` computes P1-P7 sequentially and calls `recurse<Depth - 1>()` for each product.
4. Each child recursion rewinds its own allocations, so the seven sibling products reuse the same child storage.
5. The final `rewind(mark)` releases this level logically while retaining the allocated capacity.

For `Depth == 2`:

```text
recurse<2>
  -> 7 x recurse<1>
       -> 49 x recurse<0> (BLAS leaves)
```

ThreeMonth03 (commented):
@yungyuc Please take a look.

yungyuc (approved):
LGTM

> Related to #1303.

Please also add the task number this PR is associated with, e.g.,

> Related to task 1 of issue #1303.

PR #1308, opened 2026-08-08
Document sharing `ccache` between checkouts

tigercosmos:
A second checkout, such as a `git worktree` tree, reuses nothing from the first even though ccache keeps one cache per user. The compile line carries absolute paths, and ccache also hashes the working directory when the build carries debug information, so identical sources in another directory miss.

This adds a short section to `doc/source/start/build_solvcon.md` giving the two settings that remove both, `base_dir` and a `hash_dir = false` written into a build tree, together with what they cost. The directory-specific configuration file needs ccache 4.13 or newer, `make cmakeclean` takes it away with the build tree, and an object reused elsewhere records the source paths of the checkout that compiled it first, which is why a debug tree is left without it.

Documentation only. No build or code behavior changes.

Related to PR #1297

### DIFF
--- a/doc/source/start/build_solvcon.md
+++ b/doc/source/start/build_solvcon.md
@@ -44,6 +44,24 @@ CMake itself is configured through `CMakePresets.json`; see
 {doc}`/devguide/cmake` for the preset a Windows build or an IDE selects, and
 for where the paths of a local dependency prefix belong.
 
+## Sharing ccache Between Checkouts
+
+A second checkout, such as a `git worktree` tree, reuses nothing from the
+first.  The compile line carries absolute paths, and ccache also hashes the
+working directory when the build carries debug information.  Two settings
+remove both:
+
+```sh
+ccache --set-config base_dir=$HOME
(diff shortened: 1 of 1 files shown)


doc/source/start/build_solvcon.md
```diff
@@ -75,6 +44,24 @@ CMake itself is configured through `CMakePresets.json`; see
 {doc}`/devguide/cmake` for the preset a Windows build or an IDE selects, and
 for where the paths of a local dependency prefix belong.
 
+## Sharing ccache Between Checkouts
```
tigercosmos, line 47: Update the documentent for ccache.

doc/source/start/build_solvcon.md
```diff
@@ -75,6 +44,24 @@ CMake itself is configured through `CMakePresets.json`; see
...
+
+A second checkout, such as a `git worktree` tree, reuses nothing from the
```
yungyuc, line 49: The wording can improve.  It's not touching, and reads too lack of purpose.

It's OK for now.  We improve in the future.  No hurry.

PR #1309, opened 2026-08-09
Prune superseded CI caches and drop the unused `lint` restore

tigercosmos:
GitHub cache entries are immutable and `hendrikmuhs/ccache-action` stamps every save with a timestamp, so a lineage such as `ccache-Linux-Release` gains one entry per save and only the newest can ever be restored. The rest hold the repository over its 10 GB quota, where GitHub evicts by last use across every lineage at once, so a run of Linux saves can drop the macOS cache. The sweep keeps the newest generation per lineage and per ref and deletes the rest. It lives in `contrib/ci/prune-caches.sh` rather than inline in the workflow so it can be run from a terminal against any repository, which is how it was checked here, and `.github/workflows/cache_cleanup.yml` runs it daily or by hand with the `keep` and `dry_run` inputs.

The second commit stops `lint` from restoring a ccache it cannot use. A pull request skips `make pilot` and runs `pilot_clang_tidy_diff`, which drives clang-tidy over the diff and never invokes the compiler, so nothing reaches ccache. The old `save:` condition suppressed the save but left the restore in place, so every pull request pulled `ccache-Linux-tidy-Debug` (1102 MB) down for no hit at all.

The workflow was exercised end to end on a fork by manual dispatch. A dry run predicted 15 superseded generations holding 2062 MB, the real run deleted exactly those, and the repository went from 57 entries and 10722 MB, over the quota, down to 42 entries and 8660 MB, with a second pass afterwards reporting nothing left to delete. Run against `solvcon/solvcon` this evening it reports 5 superseded generations holding 4646 MB, every one of them created by today's own runs.

### DIFF
--- a/.github/workflows/cache_cleanup.yml
+++ b/.github/workflows/cache_cleanup.yml
@@ -1,0 +1,54 @@
+name: cache cleanup
+
+on:
+  schedule:
+    # Run daily, after the nightly jobs have saved the generation to keep.
+    - cron: '41 5 * * *'
+  workflow_dispatch:
+    inputs:
+      keep:
+        description: 'Generations to keep per cache lineage'
+        default: '1'
+      dry_run:
+        description: 'List what would be deleted without deleting it'
+        type: boolean
+        default: false
+
+concurrency:
+  group: cache-cleanup
+  cancel-in-progress: false
+
+permissions:
+  # An explicit block sets every unlisted scope to none, so the checkout
+  # needs contents spelled out alongside the cache deletion.
+  actions: write
+  contents: read
+
+jobs:
+
+  prune:
+
+    # A fork carries its own cache quota but nothing there needs a scheduled
+    # sweep, so leave the cron to the upstream repository. A fork that wants
+    # one runs this by hand.
+    if: github.event_name == 'workflow_dispatch' || github.event.repository.fork == false
+
+    runs-on: ubuntu-24.04
+
+    timeout-minutes: 15
+
+    steps:
+
+      # A full tree costs more than the sweep itself, so take the one script
+      # and nothing else. Cone mode matches directories, not files, hence off.
+      - uses: actions/checkout@v6
+        with:
+          sparse-checkout: contrib/ci/prune-caches.sh
+          sparse-checkout-cone-mode: false
+
+      - name: Delete superseded cache generations
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -97,16 +97,18 @@ jobs:
       with:
         workflow: 'lint'
 
+    # Only a push or a nightly run reaches this cache. A pull request builds
+    # no pilot and runs pilot_clang_tidy_diff instead, which drives clang-tidy
+    # and never the compiler, so it compiles nothing and would pull close to a
+    # gigabyte down for no hit at all.
     - name: ccache
+      if: github.event_name != 'pull_request'
```
tigercosmos, line 105: Avoid unneccessary cache loading for PRs.

.github/workflows/cache_cleanup.yml
```diff
@@ -0,0 +1,110 @@
+name: cache cleanup
+
+on:
+  schedule:
+    # Run daily, after the nightly jobs have saved the generation to keep.
+    - cron: '41 5 * * *'
```
tigercosmos, line 6: Everyday 5:41.

contrib/ci/prune-caches.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
```
tigercosmos, line 1: Add a new script to clean up caches.

contrib/ci/prune-caches.sh
```diff
@@ -0,0 +1,117 @@
...
+# deletes the master generation that every other run restores from. Given
+#
+#   ccache-Linux-Release   master    10:00
+#   ccache-Linux-Release   master    11:00
+#   ccache-Linux-Release   feature   12:00
+#
+# grouping by lineage and ref drops only the superseded master 10:00. Grouping
+# by lineage alone ranks all three together, so feature 12:00 takes first place
+# and master 11:00 is dropped as merely second newest, leaving every run that
+# reads master cold.
```
tigercosmos, line 80: Explanation of the machenism.

.github/workflows/cache_cleanup.yml
yungyuc: Please add timezone in the code comment.

PR #1313, opened 2026-08-09
Rename reminaing `mm`/`modmesh` code

### LINKED ISSUE #1295: Rename remaining mm/modmesh to sc/solvcon

yungyuc:
For part 2 of issue #1295.

This PR clears the `mm` and `modmesh` names that the `contrib/lint/` rename scripts and the later `SC_` macro pass left behind. Seven commits, one per kind:

- `mm_x` in the Python tests becomes `sc_x`, matching the `sc` import every test already uses.
- The gtest namespace alias `mm` becomes `sc`.
- The `MMTPN` and `MMCL` parameters of `SC_DECL_SWITCH_ELEMENT_TYPE` become `SCTPN` and `SCCL`.
- `MMROOT` becomes `SCROOT` in the standalone buffer make file.
- The four `Modmesh`-prefixed identifiers become `Solvcon`-prefixed.
- The `mmpath` local in the pytest runner becomes `testpath`.
- Two stale comments in the dependency scripts drop the old project name.

The `STYLE.md` is also updated to write solvcon/sc instead of modmesh/mm. The `Modmesh` identifiers get no compatibility alias, since renaming the package already broke every import path an out-of-tree caller could have held.

The `solvcon/modmesh#814` citation in `check_skip_ci.yml` stays, and after this PR it is the only `modmesh` string outside the two rename scripts.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -593,15 +593,15 @@ solvcon::real_type local_function(solvcon::int_type value);
 } /* end namespace */
 ```
 
-The namespace `solvcon` may be aliased to `mm` in a local scope. No alias
+The namespace `solvcon` may be aliased to `sc` in a local scope. No alias
 should be use outside a local scope.
 
 ```cpp
 solvcon::real_type local_function(solvcon::int_type value)
 {
-    // Alias the solvcon namespace to mm.
-    namespace mm = solvcon;
-    return mm::real_type(value); // Same as solvcon::real_type(value);
+    // Alias the solvcon namespace to sc.
+    namespace sc = solvcon;
+    return sc::real_type(value); // Same as solvcon::real_type(value);
 }
 ```
 
--- a/STYLE.md
+++ b/STYLE.md
@@ -286,7 +286,7 @@ from matplotlib import figure
 from PySide6 import QtCore, QtWidgets, QtGui
 
 # Modules in the current project.
-import solvcon as mm
+import solvcon as sc
 from solvcon import onedim
 from solvcon.plot import svg
 
@@ -297,7 +297,7 @@ from . import _base_app
 
 > **Note:**
 >
-> `solvcon` can be shorthanded as `mm`.
+> `solvcon` can be shorthanded as `sc`.
 
 Do not import module content (classes, functions, or constants) directly.
 Always use the `foo.bar` pattern to access classes, functions, or constants:
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -27,10 +27,10 @@ TEST(ConcreteBuffer, iterator)
 
 TEST(SimpleArray, construction)
 {
-    namespace mm = solvcon;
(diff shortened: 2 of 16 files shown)


gtests/test_nopython_buffer.cpp
```diff
@@ -27,10 +27,10 @@ TEST(ConcreteBuffer, iterator)
 
 TEST(SimpleArray, construction)
 {
-    namespace mm = solvcon;
-    mm::SimpleArray<double> arr_double(10);
+    namespace sc = solvcon;
```
yungyuc, line 30: Alias should be named as `sc`, not `mm`.

solvcon/system.py
```diff
@@ -81,14 +81,14 @@ def _run_pytest(extra_args=None):
     # solvcon.
     import pytest
     import shlex
-    mmpath = os.path.join(os.path.dirname(solvcon.__file__), '..', 'tests')
-    mmpath = os.path.abspath(mmpath)
+    testpath = os.path.join(os.path.dirname(solvcon.__file__), '..', 'tests')
```
yungyuc, line 84: No need to say `mmpath`.

STYLE.md
```diff
@@ -286,7 +286,7 @@ from matplotlib import figure
 from PySide6 import QtCore, QtWidgets, QtGui
 
 # Modules in the current project.
-import solvcon as mm
+import solvcon as sc
```
yungyuc, line 289: Update `STYLE.md`

PR #1315, opened 2026-08-09
Enable slicing for geometry pads

### LINKED ISSUE #1292: Finish the container protocol on the geometry pads
`PointPad`, `SegmentPad`, and `CurvePad` take an integer index and nothing else. Give them the rest of the sequence protocol, a slice and a negative index, so a pad reads as the container it is.

Checked against the extension built at master (`1a504278`), on all three pads:

- A slice raises `TypeError`. `pad[0:2]` cannot bind, because the bound getter takes a `size_t`. A slice is what hands a plot the segments it asks for without a loop.
- A negative index raises `TypeError` for the same reason. `pad[-1]` should read the last element.
- An integer index and iteration already work, so this is the missing part rather than a new facility.

Scope:

- Accept a slice in `__getitem__` on the three pads, returning a new pad of the same type and `ndim` that holds copied data.
- Accept a negative index, normalized against `size()`, and keep `IndexError` for a genuine out-of-range.
- Leave `get_at` and `set_at` as they are. They are the explicit spelling, and the C++ side is written against them.
- Test in `tests/test_universe_shape0d.py` and `tests/test_universe_shape1d.py`: the empty pad, a full slice, a strided slice, a negative index, and an out-of-range index.

A slice copies the data rather than viewing it. A pad owns one `SimpleArray` per axis, so a view would need the offset and stride the pads do not carry today, while a copy matches `clone()` and `pack_array()`, which already copy. State the copy in the docstring, so no caller expects a write to a slice to reach the pad it came from.

PR list:

- PR #1315

## Follow-up: the remaining pads

`PointPad`, `SegmentPad`, and `CurvePad` now accept a negative index and a slice. Three pads in `cpp/solvcon/universe/` still do not, and they do not all need the same amount of work. Line references below are pinned to `master` at [`e42ceefb`](https://github.com/solvcon/solvcon/commit/e42ceefbcc7028979e13c81725705a612f474c1b).

1. **`TrianglePad`** ([`polygon.hpp:1513`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1513), bound at [`wrap_shape2d.cpp:319-324`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/pymod/wrap_shape2d.cpp#L319-L324)) is the closest match to what was just done. It already binds `__len__` and an integer `__getitem__`, but that getter takes a `size_t`, so `tp[-1]` and `tp[0:2]` raise `TypeError` exactly as the other three did. It has `construct(ndim, nelem)`, [`size()`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1697), [`ndim()`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1695), [`get_at`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1806), and [`set_at`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1817), so the index and slice helpers apply unchanged. This is a near-copy of the `SegmentPad` change.

2. **`TrapezoidPad`** ([`polygon.hpp:291`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L291), bound at [`wrap_shape2d.cpp:712-738`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/pymod/wrap_shape2d.cpp#L712-L738)) is a larger gap. It binds no dunder at all: no `__len__`, no `__getitem__`, no `__setitem__`. It exposes [`size` as a read-only property](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/pymod/wrap_shape2d.cpp#L726) rather than `__len__`, which is inconsistent with every other pad. Python cannot read a whole trapezoid element today, only loose coordinates through `x0(i)` .. `z3(i)`, even though C++ has [`get_at`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L632) and [`set_at`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L649) carrying a `trapezoid_type`. The C++ side already has what the helpers need, so the work is writing the element accessor that was never written, not just fixing an index type.

3. **`PolygonPad`** ([`polygon.hpp:2204`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L2204), bound at [`wrap_shape2d.cpp:611-643`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/pymod/wrap_shape2d.cpp#L611-L643)) is the one that does not fit. It is a ragged container: polygons of varying node counts stored behind `m_begins` offsets. It has [`num_polygons()`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L2253) and [`num_points()`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L2255) but no `size()`, no `get_at` or `set_at`, and only a `construct(ndim)` with no `nelem` overload, so the slice helper cannot be reused as written. It reads through [`get_polygon(polygon_id)`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L2307) instead.

### Current state

| Pad | `__len__` | `__getitem__` int | negative index | slice | `__setitem__` | Helper reuse | Tests |
| --- | --- | --- | --- | --- | --- | --- | --- |
| [`PointPad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/coord.hpp#L280) | yes | yes | yes | yes | no, `set_at` only | done | yes |
| [`SegmentPad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/coord.hpp#L916) | yes | yes | yes | yes | no, `set_at` only | done | yes |
| [`CurvePad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/bezier.hpp#L228) | yes | yes | yes | yes | yes | done | yes |
| [`TrianglePad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L1513) | yes | `size_t` | no | no | no | as-is | `TrianglePadTB` |
| [`TrapezoidPad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L291) | no, `size` property | no | no | no | no | as-is | none direct |
| [`PolygonPad`](https://github.com/solvcon/solvcon/blob/e42ceefbcc7028979e13c81725705a612f474c1b/cpp/solvcon/universe/polygon.hpp#L2204) | no | no | no | no | no | needs a new design | via `Polygon3dTB` |

## Issue background

The oblique-shock reflection work (issue #1269, PR #1289) had a TODO note for geometry pad indexing at [`_analytic.py:L181-L183`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L181-L183):

```python
# TODO: SegmentPad has no __getitem__, so an arm is taken with get_at.
# Indexing would read as the container it is, and a slice would hand a
# plot the arms it wants without a loop.
```

Half of that is already in the tree, which is why the scope above is the slice and the negative index alone. `SegmentPad.__getitem__` is bound at [`wrap_shape1d.cpp:L312-L316`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape1d.cpp#L312-L316), `PointPad.__getitem__` at [`wrap_shape0d.cpp:L303-L307`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape0d.cpp#L303-L307), and `CurvePad.__getitem__` at [`wrap_shape1d.cpp:L735`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape1d.cpp#L735). [`test_universe_shape1d.py:L333`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/tests/test_universe_shape1d.py#L333) reads `sp[0]` today.

The first sentence of the TODO was not correct.  The code can be corrected to read `self.arms[0]` instead of `get_at(0)` right now.

yungyuc:
`PointPad`, `SegmentPad`, and `CurvePad` bind `__getitem__` to a `size_t`, so `pad[-1]` and `pad[0:2]` both raise `TypeError` and a pad reads as less of a container than it is. This branch completes the sequence protocol on the three pads.

Each pad gains a signed integer `__getitem__` that normalizes against `size()` and raises `IndexError` when out of range, plus a slice overload that returns a new pad of the same type and `ndim`. A slice copies rather than views, because a pad owns one `SimpleArray` per axis and carries neither an offset nor a stride. `CurvePad.__setitem__` takes the negative index too. Slice assignment is out of scope, and `get_at` and `set_at` are untouched.

### DIFF
--- a/cpp/solvcon/universe/pymod/universe_pymod.hpp
+++ b/cpp/solvcon/universe/pymod/universe_pymod.hpp
@@ -11,11 +11,66 @@
 #include <solvcon/python/common.hpp>
 #include <solvcon/universe/universe.hpp>
 
+#include <format>
+#include <stdexcept>
+
 namespace solvcon
 {
 
 namespace python
 {
+
+/**
+ * Turn a Python index into a pad offset, counting a negative index from the
+ * end of the pad.
+ *
+ * @param index the index as Python spells it
+ * @param size the number of elements in the pad
+ * @param name the pad type name the error message carries
+ *
+ * @throw std::out_of_range if the normalized index falls outside the pad
+ */
+inline size_t normalize_pad_index(ssize_t index, size_t size, char const * name)
+{
+    auto const nelem = static_cast<ssize_t>(size);
+    ssize_t const it = index < 0 ? index + nelem : index;
+    if (it < 0 || it >= nelem)
+    {
+        throw std::out_of_range(
+            std::format("{}: index {} is out of bounds with size {}", name, index, size));
+    }
+    return static_cast<size_t>(it);
+}
+
+/**
+ * Copy the elements a Python slice selects into a new pad of the same type
+ * and dimensionality.
+ *
+ * A pad owns one array per axis and carries neither an offset nor a stride,
+ * so the result copies the selected elements instead of viewing them. The
+ * result is unaligned, because an alignment the slice length does not divide
+ * cannot be honored.
+ *
+ * @param self the pad to read
+ * @param key the Python slice to apply
+ */
+template <typename Pad>
+std::shared_ptr<Pad> copy_pad_slice(Pad const & self, pybind11::slice const & key)
+{
+    ssize_t start = 0, stop = 0, step = 0, slicelength = 0;
+    if (!key.compute(static_cast<ssize_t>(self.size()), &start, &stop, &step, &slicelength))
+    {
+        throw pybind11::error_already_set();
+    }
+
+    std::shared_ptr<Pad> ret = Pad::construct(self.ndim(), static_cast<size_t>(slicelength));
+    for (ssize_t i = 0; i < slicelength; ++i)
+    {
+        ret->set_at(static_cast<size_t>(i), self.get_at(static_cast<size_t>(start)));
+        start += step;
+    }
+    return ret;
+}
 
 void initialize_universe(pybind11::module & mod);
 void wrap_shape0d(pybind11::module & mod);
--- a/solvcon/pilot/apps/obsrefl/_analytic.py
+++ b/solvcon/pilot/apps/obsrefl/_analytic.py
@@ -178,19 +178,15 @@ class Reflection(object):
                 core.Point3dFp64(head[0], head[1], 0.0),
                 core.Point3dFp64(tail[0], tail[1], 0.0)))
 
-    # TODO: SegmentPad has no __getitem__, so an arm is taken with get_at.
-    # Indexing would read as the container it is, and a slice would hand a
-    # plot the arms it wants without a loop.
-
     @property
     def incident(self):
         """The segment the incident shock runs along."""
-        return self.arms.get_at(0)
+        return self.arms[0]
 
     @property
     def reflected(self):
         """The segment the reflected shock runs along, or None."""
-        return self.arms.get_at(1) if self.has_reflection else None
+        return self.arms[1] if self.has_reflection else None
 
     @staticmethod
     def _relative(value, target):
--- a/tests/test_universe_shape0d.py
+++ b/tests/test_universe_shape0d.py
@@ -454,6 +454,71 @@ class PointPadTB(testing.TestBase):
         self.assert_allclose(pp.z_at(1), 8.23)
         self.assert_allclose(pp.z_at(2), 3.3 * 5.1)
 
+    def test_getitem_index(self):
+        pp = self.PointPad(ndim=2)
+        for it in range(4):
+            pp.append(float(it), float(it) * 10)
+
+        self.assert_allclose(list(pp[0]), (0, 0, 0))
+        self.assert_allclose(list(pp[-1]), (3, 30, 0))
+        self.assert_allclose(list(pp[-4]), (0, 0, 0))
+
+        with self.assertRaisesRegex(
+                IndexError, "PointPad: index 4 is out of bounds with size 4"):
+            pp[4]
+        with self.assertRaisesRegex(
+                IndexError, "PointPad: index -5 is out of bounds with size 4"):
+            pp[-5]
+
+        empty = self.PointPad(ndim=3)
+        with self.assertRaisesRegex(
+                IndexError, "PointPad: index 0 is out of bounds with size 0"):
+            empty[0]
+        with self.assertRaisesRegex(
+                IndexError, "PointPad: index -1 is out of bounds with size 0"):
+            empty[-1]
+
+    def test_getitem_slice(self):
+        pp = self.PointPad(ndim=3)
+        for it in range(4):
+            pp.append(float(it), float(it) * 10, float(it) * 100)
+
+        full = pp[:]
+        self.assertIsInstance(full, type(pp))
+        self.assertEqual(full.ndim, 3)
+        self.assertEqual(len(full), 4)
+        self.assert_allclose(full.x, [0, 1, 2, 3])
+        self.assert_allclose(full.y, [0, 10, 20, 30])
+        self.assert_allclose(full.z, [0, 100, 200, 300])
+
+        strided = pp[1::2]
+        self.assertEqual(strided.ndim, 3)
+        self.assert_allclose(strided.x, [1, 3])
+        self.assert_allclose(strided.y, [10, 30])
+        self.assert_allclose(strided.z, [100, 300])
+
+        flipped = pp[::-1]
+        self.assert_allclose(flipped.x, [3, 2, 1, 0])
+
+        # A slice copies, so writing to it leaves the source alone.
+        part = pp[1:3]
+        part.set_at(0, -1.0, -2.0, -3.0)
+        self.assert_allclose(list(part[0]), (-1, -2, -3))
+        self.assert_allclose(list(pp[1]), (1, 10, 100))
+
+        # A 2D pad keeps its dimensionality and its empty z array.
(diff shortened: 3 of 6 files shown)


cpp/solvcon/universe/pymod/universe_pymod.hpp
```diff
@@ -11,12 +11,67 @@
...
+ * @param index the index as Python spells it
+ * @param size the number of elements in the pad
+ * @param name the pad type name the error message carries
+ *
+ * @throw std::out_of_range if the normalized index falls outside the pad
+ */
+inline size_t normalize_pad_index(ssize_t index, size_t size, char const * name)
+{
+    auto const nelem = static_cast<ssize_t>(size);
+    ssize_t const it = index < 0 ? index + nelem : index;
```
yungyuc, line 36: Allow negative index.

cpp/solvcon/universe/pymod/universe_pymod.hpp
```diff
@@ -11,12 +11,67 @@
...
+ * A pad owns one array per axis and carries neither an offset nor a stride,
+ * so the result copies the selected elements instead of viewing them. The
+ * result is unaligned, because an alignment the slice length does not divide
+ * cannot be honored.
+ *
+ * @param self the pad to read
+ * @param key the Python slice to apply
+ */
+template <typename Pad>
+std::shared_ptr<Pad> copy_pad_slice(Pad const & self, pybind11::slice const & key)
```
yungyuc, line 58: Support slicing.

solvcon/pilot/apps/obsrefl/_analytic.py
```diff
@@ -178,19 +178,15 @@ def __init__(self, shock, margin=MARGIN_FRACTION):
...
 
-    # TODO: SegmentPad has no __getitem__, so an arm is taken with get_at.
-    # Indexing would read as the container it is, and a slice would hand a
-    # plot the arms it wants without a loop.
-
     @property
     def incident(self):
         """The segment the incident shock runs along."""
-        return self.arms.get_at(0)
+        return self.arms[0]
```
yungyuc, line 184: Use container protocol instead of ad hoc `get_at()`.

tests/test_universe_shape0d.py
```diff
@@ -454,6 +454,71 @@ def test_append_3d(self):
         self.assert_allclose(pp.z_at(1), 8.23)
         self.assert_allclose(pp.z_at(2), 3.3 * 5.1)
 
+    def test_getitem_index(self):
+        pp = self.PointPad(ndim=2)
```
yungyuc, line 458: Test `PointPad`.

tests/test_universe_shape1d.py
```diff
@@ -580,6 +580,74 @@ def test_append_3d(self):
         self.assert_allclose(list(sp.y1), list(sp.p1.y))
         self.assert_allclose(list(sp.z1), list(sp.p1.z))
 
+    def test_getitem_index(self):
```
yungyuc, line 583: Test `SegmentPad`.

tests/test_universe_shape1d.py
```diff
@@ -911,6 +979,85 @@ def test_sample_2d(self):
         self.assert_allclose(list(sp[9].p1),
                              [7.0, 0.0, 0.0])
 
+    def _make_curve_pad(self, ndim, ncurve):
+        cp = self.CurvePad(ndim=ndim)
```
yungyuc, line 983: Test `CurvePad`.

tigercosmos (commented):
LGTM

PR #1316, opened 2026-08-09
Check matmul BLAS matrix views

ThreeMonth03:
The full clang-tidy build rejects four `std::optional::value()` calls in `MatmulExecutor` as unchecked matrix-view access.

Add one shared `require_matrix_view()` guard and route GEVM, GEMV, and GEMM descriptors through it. This preserves the selector and executor invariant while reporting an internal error if a selected BLAS kernel lacks a compatible matrix view.

Related to #1172.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -315,6 +315,7 @@ private:
     void execute_gemm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data);
     std::optional<matrix_view_type> lhs_matrix_view(value_type const * data) const;
     std::optional<matrix_view_type> rhs_matrix_view(value_type const * data) const;
+    static matrix_view_type require_matrix_view(std::optional<matrix_view_type> view);
     static std::optional<matrix_view_type> make_matrix_view(
         value_type const * data,
         ssize_t row_stride,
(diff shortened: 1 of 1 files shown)


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -904,24 +905,24 @@ void MatmulExecutor<Array>::execute_dot_blas(value_type * output, value_type con
 template <typename Array>
 void MatmulExecutor<Array>::execute_gevm_blas(value_type * output, value_type const * lhs_data, value_type const * rhs_data)
 {
-    matrix_view_type const matrix = rhs_matrix_view(rhs_data).value();
+    matrix_view_type const matrix = require_matrix_view(rhs_matrix_view(rhs_data));
```
ThreeMonth03, line 908: GEVM, GEMV, and GEMM now use the same checked conversion. Kernel selection and BLAS descriptor construction remain unchanged.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -941,6 +942,17 @@ std::optional<typename MatmulExecutor<Array>::matrix_view_type> MatmulExecutor<A
         data, m_plan.rhs_inner_stride(), m_plan.rhs_column_stride(), m_plan.inner_size(), m_plan.columns());
 }
 
+template <typename Array>
+typename MatmulExecutor<Array>::matrix_view_type MatmulExecutor<Array>::require_matrix_view(
+    std::optional<matrix_view_type> view)
+{
+    if (!view)
```
ThreeMonth03, line 949: The selector already rejects incompatible layouts. This guard makes that invariant explicit at the executor boundary, so full clang-tidy builds can prove the optional access is checked.

ThreeMonth03 (commented):
The inline comments explain the selector and executor invariant behind the checked optional access.

yungyuc (approved):
LGTM.

CC gave me the same code.  Please also check that it does not hurt runtime.  I guess it's benign.

PR #1319, opened 2026-08-09
Use boxed sub-widgets in the reflection panel

### LINKED ISSUE #1270: Drive the Panel with the Run Session (polish issue 3)
Develop a GUI panel for the run logic of the session.  `_app.py` drops its inline march-and-count loop and drives `ReflectionSession` from the frame timer, and the status tree gains the residual and the per-zone errors in percent, updated as the session advances.  The behavior the panel already has (start, pause, step, the viewer lifecycle) is unchanged, and the window-lane tests extend to the new rows.

Input from https://github.com/solvcon/solvcon/pull/1299#issuecomment-5226941869 (@j8xixo12): Opening the euler solver panel does not open the Domain viewer alongside it. The Domain viewer should open automatically when panel launch and render the initial state. Additionally, there is an issue with the run state indicator at the bottom of the panel: it keep showing **running**, even after pressing pause.

We may need some design iterations to polish the UI.

- PR #1299
- PR #1318
- PR #1319
- PR #1320

yungyuc:
For issue #1270 

Rebuild `SolutionPanel` from one flat widget into five boxes stacked in the order a run is used:

- `FreeStreamBox` and `NumericsBox` hold the inputs consumed at Start
- `RunBox` pairs the march controls with the march readout
- `FieldBox` picks what the viewer colors and shows the drawn range
- `ZoneBox` judges the field against the analytic states

The readout now shows progress against the known step cap, what ended the run, and the recorded overall mass, in right-aligned fixed-width cells that update without jitter (I think; if you find jittering, we fix again).

The shock angle carries its unit as a spin-box suffix, and Pause and Step are disabled until a run exists. The panel forwards the boxes' callbacks and accessors as one flat surface, so the controller is untouched.

### DIFF
--- a/solvcon/pilot/apps/obsrefl/_panel.py
+++ b/solvcon/pilot/apps/obsrefl/_panel.py
@@ -4,13 +4,14 @@
 
 """Control widget for the oblique-shock reflection run.
 
-The controls set the free stream and the mesh, open or close the domain
-viewer sub-window, start / pause / step the march, and pick which derived
-field (density, velocity, pressure, Mach, ...) the viewer colors.  Below
-them the status tree reads out the run: how far it has come, and how far
-each zone stands from the analytic state it has to hold.  The widget only
-reports what its controls hold and calls back into its owner; the solver
-and the viewer belong to the feature in :mod:`._app`.
+The input boxes hold what a run consumes when it starts: the free stream
+and the incident shock in one, the discretization in the other.  The run
+box holds the march controls.  Below them the field selector picks what
+the viewer colors, and the status tree reads out the run.  Each box is a
+passive widget that only reports what its controls hold and calls back
+into its owner; the panel composes the boxes and forwards their surface,
+and the solver and the viewer belong to the controller in
+:mod:`._control`.
 """
 
 import math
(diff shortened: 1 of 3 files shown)


solvcon/pilot/apps/obsrefl/_panel.py
```diff
@@ -27,56 +29,106 @@
...
+def _value_label():
+    """A live readout cell: right-aligned in a fixed-width font, so a value
+    can update every frame without jittering the column it stands in."""
+    label = QLabel("")
+    label.setAlignment(Qt.AlignRight | Qt.AlignVCenter)
+    label.setFont(QFontDatabase.systemFont(QFontDatabase.SystemFont.FixedFont))
+    return label
+
+
+class FreeStreamBox(QGroupBox):
```
yungyuc, line 59: Group information in boxes.

tests/apps/obsrefl/test_panel.py
yungyuc, line 1: Testing code to keep the panel working as expected.

tests/gui/test_obsrefl.py
yungyuc, line 1: Update the accessing path according to the new boxes.

PR #1322, opened 2026-08-09
Delete the caches stranded on refs that are gone

tigercosmos:
A cache is scoped to the ref that saved it, so the newest generation on a deleted branch or a closed pull request survives the sweep even though nothing can ever restore it. Only GitHub's own eviction takes those, and it runs on last use across the whole repository, which is the behaviour the sweep exists to stay ahead of. This asks about each ref that holds a cache and deletes every generation on the ones that are gone.

Deleting on liveness fails differently from deleting a superseded generation. The latter is provably safe: a newer entry of the same key already exists, so nothing restorable is lost. Liveness is read from outside, so the evidence has to be positive rather than an absence, and what counts as positive differs by ref. A branch really can be deleted, so a 404 is its death, but `gh` says 404 just as readily for a repository it cannot read, which would answer for every branch at once; a 404 is therefore trusted only while the default branch, which cannot be missing, still answers for itself. A pull request cannot be deleted at all, and the merge ref is itself proof the number existed, so a 404 there means the probe cannot see pull requests, which is also what a token missing that scope returns; only a closed state counts. A branch name carrying `#` or `%` is left unprobed, since neither survives the URL path and the answer would describe some other branch.

A ref the probe cannot settle stays live, and the sweep carries on to the superseded generations, which need no probe at all. One such ref is a warning, since it costs only a little space and the run still did its main job. Every probed ref going unsettled is a different thing, a repository the run could not read, and that ends the run non-zero.

Validated against a fork. A real run deleted 34 entries, 28 superseded and 6 on two closed pull requests, taking the repository from 64 entries and 10782 MB, over the 10 GB quota, down to 30 entries and 6004 MB, with every live ref keeping its caches and a second pass finding nothing left to do. That run predates the probe rework, so the current revision was dispatched again on the same fork: six refs probed, no warnings, and the count correctly unchanged because nothing there is stale any more. The schedule and the fork guard were checked the same way, by moving the cron a few minutes ahead on a throwaway commit, and the run fired and skipped the job as a fork should.

The failure paths were stubbed and checked one at a time: a 500 on a branch probe leaves that branch live, a 404 on a pull request probe marks nothing dead, an unreadable repository aborts, one unsettled ref out of six warns and exits zero, and the only unsettled ref of one exits non-zero after still sweeping the superseded generations.

### DIFF
--- a/.github/workflows/cache_cleanup.yml
+++ b/.github/workflows/cache_cleanup.yml
@@ -2,7 +2,7 @@ name: cache cleanup
 
 on:
   schedule:
-    # Run daily, after the nightly jobs have saved the generation to keep.
+    # Run daily, after the nightly jobs have saved the generation to keep at 5:41 UTC.
     - cron: '41 5 * * *'
   workflow_dispatch:
     inputs:
@@ -19,10 +19,12 @@ concurrency:
   cancel-in-progress: false
 
 permissions:
-  # An explicit block sets every unlisted scope to none, so the checkout
-  # needs contents spelled out alongside the cache deletion.
+  # An explicit block sets every unlisted scope to none. The checkout and the
+  # branch probe read contents, the pull request probe reads pull-requests,
+  # and the sweep itself writes actions.
   actions: write
   contents: read
+  pull-requests: read
 
 jobs:
 
@@ -46,7 +48,7 @@ jobs:
           sparse-checkout: contrib/ci/prune-caches.sh
           sparse-checkout-cone-mode: false
 
-      - name: Delete superseded cache generations
+      - name: Delete unrestorable caches
         env:
           GH_TOKEN: ${{ github.token }}
           KEEP: ${{ inputs.keep || '1' }}
--- a/contrib/ci/prune-caches.sh
+++ b/contrib/ci/prune-caches.sh
@@ -3,9 +3,10 @@
 # Copyright (c) 2026, solvcon team <contact@solvcon.net>
 # BSD 3-Clause License, see COPYING
 #
-# Delete the cache generations that GitHub keeps but no run can ever restore.
-# Driven by .github/workflows/cache_cleanup.yml, and runnable by hand against
-# any repository the local `gh` is authenticated for:
+# Delete the caches that GitHub keeps but no run can ever restore: the
+# generations a newer save superseded, and everything saved on a ref that is
+# gone. Driven by .github/workflows/cache_cleanup.yml, and runnable by hand
+# against any repository the local `gh` is authenticated for:
 #
 #   GITHUB_REPOSITORY=solvcon/solvcon DRY_RUN=true contrib/ci/prune-caches.sh
 #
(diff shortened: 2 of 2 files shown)


.github/workflows/cache_cleanup.yml
```diff
@@ -19,10 +19,12 @@ concurrency:
...
 
 permissions:
-  # An explicit block sets every unlisted scope to none, so the checkout
-  # needs contents spelled out alongside the cache deletion.
+  # An explicit block sets every unlisted scope to none. The checkout and the
+  # branch probe read contents, the pull request probe reads pull-requests,
+  # and the sweep itself writes actions.
   actions: write
   contents: read
+  pull-requests: read
```
tigercosmos, line 27: Check if the PR is still opened?

contrib/ci/prune-caches.sh
```diff
@@ -65,6 +66,109 @@ if [ -s "$workdir/caches.tsv" ] \
...
+# the sweep goes on to the generations that need no probe at all.
+: > "$workdir/dead-refs.txt"
+probed=0
+unjudged=0
+while read -r ref; do
+  case "$ref" in
+    refs/heads/*)
+      branch="${ref#refs/heads/}"
+      case "$branch" in
+        *['#%']*)
```
tigercosmos, line 112: branch guard for #/% name.

contrib/ci/prune-caches.sh
```diff
@@ -65,6 +66,109 @@ if [ -s "$workdir/caches.tsv" ] \
...
+      number="${ref#refs/pull/}"
+      number="${number%/merge}"
+      # A closed pull request is the only death here. GitHub does not delete
+      # pull requests, and this ref is proof number existed, so a 404 says the
+      # probe cannot see them, which a token without pull-requests read also
+      # answers. Reading that as death would sweep every merge ref at once.
+      probed=$((probed + 1))
+      if state="$(gh api "repos/$GITHUB_REPOSITORY/pulls/$number" --jq '.state' \
+          2>"$workdir/probe.err")"; then
+        if [ "$state" = 'closed' ]; then
```
tigercosmos, line 144: If a PR is closed, then it's dead—no need to keep the cache.

contrib/ci/prune-caches.sh
```diff
@@ -65,6 +66,109 @@ if [ -s "$workdir/caches.tsv" ] \
...
+  cat "$workdir/probe.err" >&2
+  exit 1
+fi
+
+# Checking that once would only cover the state before the loop, and access
+# can be lost while it runs: a repository made private, renamed, or a token
+# rotated away turns every later probe into a 404 and would condemn every
+# branch in one pass. So a 404 is read as death only while the default
+# branch, which cannot be missing, still answers for itself.
+probe_branch() {
```
tigercosmos, line 85: Check if the branch is still alive. No need to keep the dead branch's cache.

Issue #1310, opened 2026-08-09
The core libraries are built with no warning flags


tigercosmos:
`COMMON_COMPILER_OPTIONS` is set in the top-level `CMakeLists.txt` at lines 381 to 388, but `add_subdirectory(cpp/solvcon)` runs at line 374, before it. CMake expands a variable at the point of use, so the two `${COMMON_COMPILER_OPTIONS}` references in `cpp/solvcon/CMakeLists.txt` (lines 341 and 347) expand to nothing. The core libraries are built with no warning flags at all: no `-Wall -Wextra -Werror` on Clang and GCC, and no `/W4 /WX` on MSVC.

Only the `-Wno-*` suppressions and `/bigobj`, which are written as literals next to the expansion, actually reach the core.

Evidence from a configured tree, counting entries in `compile_commands.json` that carry `-Wall`:

```
total TUs: 157   with -Wall: 4
    cpp/binary/pymod_solvcon/module.cpp
    cpp/binary/pilot/pilot.cpp
    build/rel314/cpp/binary/pilot/pilot_autogen/mocs_compilation.cpp
    build/rel314/cpp/binary/pilot/.qt/rcc/qrc_app_icon.cpp
```

All four come from directories added after line 381, so they pick the variable up. Nothing under `cpp/solvcon` does. The generated flags for a core object are:

```
-O3 -DNDEBUG -std=c++23 -arch arm64 -fPIC
-Wno-unused-value -Wno-noexcept-type -Wno-elaborated-enum-base
```

The fix is to move the `COMMON_COMPILER_OPTIONS` block above `add_subdirectory(cpp/solvcon)`.

That is a one-line move, but it is not a small change in effect: it turns on `-Wall -Wextra -Werror` for roughly 72 core translation units that have never been compiled with them, and `LINT_AS_ERRORS` is on by default. Expect a backlog of warnings to work through, so the move and the cleanup probably want to be separate steps, or the move wants to land together with whatever suppressions the first build turns up.

This is long-standing rather than a regression. It surfaced during review of #1305, which splits the core into three libraries and applies the same (empty) expansion to each of them.

yungyuc:
[...]

PR #1323, opened 2026-08-09
Build the core libraries with the common warning flags

tigercosmos:
Related to #1310.

`COMMON_COMPILER_OPTIONS` was set after `add_subdirectory(cpp/solvcon)`. `add_subdirectory()` copies the calling scope's variables into the subdirectory as it is called, so a variable set afterwards is not visible there at all: the two expansions in `cpp/solvcon/CMakeLists.txt` produced nothing, and the core libraries were built with no warning flags. No `-Wall -Wextra -Werror` on Clang and GCC, and no `/W4 /WX` on MSVC. Only the literal `-Wno-*` suppressions and `/bigobj` written next to the expansion reached the core. The first commit moves the block above the core so both expansions carry the flags.

Turning the flags on exposed 30 sites across 19 files on Clang and GCC. The second commit fixes each one rather than suppressing it. Most are `-Wsign-compare` from a loop counter typed `size_t` against `ndim()`, `nbody()`, or `shape(0)`, which all return `ssize_t`, and against the `uint_type` shapes in the Plot3d reader. The rest are `-Wpessimizing-move` on a returned temporary, `-Wmissing-braces` on the union aggregate initializers in `bezier.hpp`, `coord.hpp`, and `polygon.hpp`, `-Wreorder-ctor` in the `SimpleArrayPlex` constructors, `-Wmissing-field-initializers` on `ShapeRecord`, and a few dead variables, one dead private field, and one unused parameter.

MSVC reported a separate and larger set that reproduces on no other compiler, and it came in layers, because MSVC stops a translation unit at its first error. Three of those turned out to be worth fixing rather than silencing:

- `bit_reverse()` in `transform/fourier.cpp` shifted an `int` literal (`1 << i`) while taking and returning `size_t`, so the shift was undefined for more than 32 bits (C4334). It now shifts `size_t{1}`.
- Two loops in `serialization/SerializableItem.cpp` declared a `c` that shadowed the enclosing `const char c = json[index]` (C4456). The inner variable is now `digit`.
- `include_directories()` added the numpy headers as a normal include, so `/W4 /WX` applied to numpy's own code (C4127). They are a `SYSTEM` include now, which also makes them `-isystem` on Clang and GCC, and the core passes `/external:W0` so MSVC honors it.

What remains is C4804, C4244, C4267, C4458, and C4702, which come from template bodies instantiated over every value type, including `SimpleArray<bool>`. Those are suppressed in the core for now and #1324 tracks fixing the code and removing them one code at a time.

Two changes go past the diagnostic that reported them. `OasisRecordPoly::to_bytes()` compared against `m_vertices.size() - 1`, which wraps on an empty vector, so the guard is now `i + 1 < m_vertices.size()`. `Trapezoid3d` has the same brace defect as the three that warned but is not instantiated in this build, and is fixed alongside them.

The commits are split so the warning fixes are reviewable on their own, but only the pair builds: the first commit alone turns the flags on before the sources are clean.

This restores the flags but keeps the shape of the original trap, since an empty expansion is still silent. #1325 proposes carrying the flags on an `INTERFACE` target instead, which turns the same ordering mistake into a configure-time error, and covers the `gtests` target, which never referenced the variable and still compiles roughly thirty core sources with no warning flags under a different preprocessor configuration.

Reviewing this branch also turned up three defects that are not warnings and that these flags do not catch, including a Python-reachable segfault. They are unrelated to the flag work and are filed separately as #1326.

Verified locally on Clang 20, arm64 macOS: `make` builds clean with `-Werror` active on the core, `make gtest` passes 242 tests, `make pytest` passes 2107 tests, and `make lint` is clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,12 +371,9 @@ if(BUILD_QT AND "${PYSIDE6_LIBFILE}" STREQUAL "")
         "configure with BUILD_QT=OFF.")
 endif()
 
-add_subdirectory(cpp/solvcon)
-
-include(GNUInstallDirs)
-
-set(SOLVCON_PY_DIR "${PROJECT_SOURCE_DIR}/solvcon")
-
+# add_subdirectory() copies this scope's variables into the subdirectory as it
+# is called, so this must stay above every add_subdirectory() that consumes it.
+# Set it below one and the subdirectory sees nothing, with no diagnostic.
 if(MSVC)
     set(COMMON_COMPILER_OPTIONS /W4 /bigobj /WX)
 else()
@@ -388,6 +385,12 @@ else()
         message(STATUS "Enable -Werror because clang-tidy (with lint as errors) is disabled")
     endif()
 endif()
+
+add_subdirectory(cpp/solvcon)
+
+include(GNUInstallDirs)
+
+set(SOLVCON_PY_DIR "${PROJECT_SOURCE_DIR}/solvcon")
 
 add_subdirectory(cpp/binary/pymod_solvcon)
 if(BUILD_QT)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,7 +282,9 @@ include_directories(${pybind11_INCLUDE_DIRS})
 
 set(NUMPY_INCLUDE_DIR "${Python_NumPy_INCLUDE_DIRS}")
 message(STATUS "NUMPY_INCLUDE_DIR: ${NUMPY_INCLUDE_DIR}")
-include_directories(${NUMPY_INCLUDE_DIR})
+# SYSTEM keeps the core's warning flags off numpy's own headers, which do not
+# build clean under `/W4` and are not ours to fix.
+include_directories(SYSTEM ${NUMPY_INCLUDE_DIR})
 
 set(SOLVCON_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp" CACHE INTERNAL "")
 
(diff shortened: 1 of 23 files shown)


cpp/solvcon/CMakeLists.txt
```diff
@@ -340,6 +340,13 @@ foreach (SOLVCON_LIBRARY IN LISTS SOLVCON_LIBRARIES)
             ${SOLVCON_LIBRARY} PRIVATE
             ${COMMON_COMPILER_OPTIONS}
             /bigobj # C1128: number of sections exceeded object file format limit
+            # TODO: fix the code and drop each of these; see issue #1324. They
```
tigercosmos, line 343: TODO

CMakeLists.txt
```diff
@@ -389,6 +386,12 @@ else()
     endif()
 endif()
 
+add_subdirectory(cpp/solvcon)
+
+include(GNUInstallDirs)
+
+set(SOLVCON_PY_DIR "${PROJECT_SOURCE_DIR}/solvcon")
```
tigercosmos, line 395: Meat.

cpp/solvcon/oasis/oasis_device.cpp
```diff
@@ -140,7 +140,7 @@ std::vector<uint8_t> OasisRecordPoly::to_bytes() const
     // Please refer Point-list in OASIS draft 7.7.
     std::vector<uint8_t> point_list;
 
-    for (int i = 0; i < m_vertices.size() - 1; i++)
+    for (size_t i = 0; i + 1 < m_vertices.size(); i++)
```
yungyuc, line 143: TIL

cpp/solvcon/CMakeLists.txt
yungyuc: If it's not mentioned in an existing issue, create one.

tigercosmos: It's #1324 as mentioned in the comment.

PR #1327, opened 2026-08-10
Fix plot3d multiblock issues

### LINKED ISSUE #1326: Plot3d multi-block cells are degenerate, and OasisRecordPoly crashes on short input
Three defects found while reviewing the warning-flag work in #1310. None of them is a compiler warning, so turning `-Wall -Wextra -Werror` on for the core does not catch any of them. All three pre-date that work and are on master today. Each claim below was checked against the code, and the crash was reproduced.

## 1. `Plot3d` builds garbage cells for every block after the first

`cpp/solvcon/inout/plot3d.hpp:66` reads:

```cpp
return blk > 0 ? m_blk_sizes(blk - 1) : 0 + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk);
```

The conditional operator binds looser than `+`, so this parses as `blk > 0 ? m_blk_sizes(blk - 1) : (0 + x + y * ... + z * ...)`. When `blk > 0` the function returns `m_blk_sizes(blk - 1)` and throws `x`, `y`, and `z` away.

The only caller is `buildHexahedronElements`, which fills all eight node ids of every hexahedron through this helper. For every cell in every block after the first, all eight ids collapse to one constant, so the cells are fully degenerate and `build_interior` writes them into `clnds` as they are. A multi-block Plot3d file therefore produces a `StaticMesh` that is wrong everywhere past block 0, silently.

Parenthesizing as `(blk > 0 ? m_blk_sizes(blk - 1) : 0) + x + ...` fixes the parse but is still wrong for `blk >= 2`: `m_blk_sizes(i)` holds one block's size, not a running total (`plot3d.cpp:39`). The base offset needs to be the cumulative size of all preceding blocks, so either store a prefix sum or accumulate it in the helper.

`tests/test_plot3d.py` covers a single block only (its data begins with `1`), and there is no gtest for `Plot3d`, which is why this has gone unnoticed.

## 2. `solvcon.OasisRecordPoly([]).to_bytes()` segfaults

`OasisRecordPoly::to_bytes()` in `cpp/solvcon/oasis/oasis_device.cpp` indexes `m_vertices[0]` and `m_vertices[1]` at lines 127 and 131, and `m_vertices[0]` again at lines 157 and 160, with no size check. The constructor accepts any vector, including an empty one, and `wrap_oasis.cpp:65` binds it straight through as `py::init<std::vector<std::pair<int, int>>>()` with no validation.

```python
import solvcon
solvcon.OasisRecordPoly([]).to_bytes()        # segfault
solvcon.OasisRecordPoly([[0, 0]]).to_bytes()  # reads m_vertices[1] out of bounds
```

The first line kills the interpreter (verified: the process dies on SIGSEGV). The second reads past the end and returns bytes built from the garbage it found.

Line 137 has a related defect that a size guard would also cover. `segment.push_back(m_vertices.size() - 1)` wraps to `SIZE_MAX` on an empty vector, and it truncates in any case, because `segment` is a `std::vector<uint8_t>`: a polygon with 258 vertices writes a vertex count of 1. Beyond the truncation, the count is written as one raw byte, but OASIS encodes it as a variable-length unsigned integer, so any value from 128 up is misparsed by a conforming reader even before the byte overflows. `OasisDevice::append_unsigned_integer` is the routine the rest of this function already uses for exactly this.

Suggested fix: reject fewer than two vertices in `to_bytes()` (or in the constructor) with a clear exception, and write the count through `append_unsigned_integer`.

## 3. Self-comparison hides the 1-delta type classification

`cpp/solvcon/oasis/oasis_device.cpp:131`:

```cpp
else if (m_vertices[0].first == m_vertices[0].first)
```

This compares an element to itself and is always true, so the branch is an unconditional `else`. Per the Type 0 and Type 1 comments just above, the intended test is `m_vertices[0].first == m_vertices[1].first`, that is, the first segment is vertical because x is constant.

The practical impact is small and worth stating honestly: for any valid Manhattan polygon the emitted bytes are identical either way. The two versions diverge only when the first segment is diagonal, and the 1-delta encoding at line 151 cannot represent a diagonal regardless. So this is a latent correctness bug rather than a live data corruption. It is still worth fixing, and the `else` is a good place to reject non-Manhattan input rather than emit no type byte at all.

Note that GCC's `-Wtautological-compare` does not fire here, because `std::vector::operator[]` is a function call rather than a plain operand.

- PR #1361

j8xixo12:
`coordinate_to_node_id` parsed as `blk > 0 ? m_blk_sizes(blk - 1) : (0 + ...)` because the conditional operator precedence looser than `+`, so every cell in every block after the first collapsed its eight node ids into one constant and the mesh silently degenerated. Parenthesizing alone would still be wrong from the third block on, because `m_blk_sizes` holds per-block sizes, so the helper now reads the base from `m_blk_offsets`, the exclusive prefix sum the constructor accumulates.  A test over three literal Plot3d blocks of unequal shapes pins every node coordinate and every cell's connectivity; on the unfixed reader it fails with the eight node ids of a later block's cell collapsed into one.

Related to #1326 plot3d issue.

### DIFF
--- a/cpp/solvcon/inout/plot3d.cpp
+++ b/cpp/solvcon/inout/plot3d.cpp
@@ -27,6 +27,7 @@ Plot3d::Plot3d(const std::string & data)
     m_y_shape.remake(small_vector<ssize_t>{static_cast<ssize_t>(nblocks)}, 0);
     m_z_shape.remake(small_vector<ssize_t>{static_cast<ssize_t>(nblocks)}, 0);
     m_blk_sizes.remake(small_vector<ssize_t>{static_cast<ssize_t>(nblocks)}, 0);
+    m_blk_offsets.remake(small_vector<ssize_t>{static_cast<ssize_t>(nblocks)}, 0);
 
     // parsing xyz dimension of each block
     for (auto i = 0; i < nblocks; ++i)
@@ -37,6 +38,7 @@ Plot3d::Plot3d(const std::string & data)
         m_y_shape(i) = std::stoul(tokens[1]);
         m_z_shape(i) = std::stoul(tokens[2]);
         m_blk_sizes(i) = m_x_shape(i) * m_y_shape(i) * m_z_shape(i);
+        m_blk_offsets(i) = static_cast<uint_type>(total_blk_size);
         total_blk_size += m_blk_sizes(i);
     }
 
@@ -48,7 +50,6 @@ Plot3d::Plot3d(const std::string & data)
 
 void Plot3d::parseCoordinates(const uint_type nblocks)
 {
-    uint_type base_idx = 0;
     std::string line;
     std::queue<real_type> parsing_q;
     const std::regex plot3d_coord_delim(R"(-?\d+(\.\d+)?([eE][+-]?\d+)?|\b\d+\b)");
@@ -56,6 +57,7 @@ void Plot3d::parseCoordinates(const uint_type nblocks)
     // TODO: The nested for-loop needs to be enhanced
     for (auto blk = 0; blk < nblocks; ++blk)
     {
+        const uint_type base_idx = m_blk_offsets(blk);
         for (auto k = 0; k < 3; ++k)
         {
             for (auto j = 0; j < m_blk_sizes(blk); ++j)
@@ -72,7 +74,6 @@ void Plot3d::parseCoordinates(const uint_type nblocks)
                 parsing_q.pop();
             }
         }
-        base_idx += m_blk_sizes(blk);
     }
 }
 
--- a/cpp/solvcon/inout/plot3d.hpp
+++ b/cpp/solvcon/inout/plot3d.hpp
@@ -63,7 +63,7 @@ public:
 private:
     inline uint_type coordinate_to_node_id(uint_type blk, uint_type x, uint_type y, uint_type z)
     {
-        return blk > 0 ? m_blk_sizes(blk - 1) : 0 + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk);
+        return m_blk_offsets(blk) + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk);
     }
 
     void parseCoordinates(uint_type nblocks);
@@ -76,6 +76,8 @@ private:
     SimpleArray<uint_type> m_y_shape;
     SimpleArray<uint_type> m_z_shape;
     SimpleArray<uint_type> m_blk_sizes;
+    /// Exclusive prefix sum of m_blk_sizes: the node-id base of each block.
+    SimpleArray<uint_type> m_blk_offsets;
 
     std::unordered_map<uint_type, small_vector<uint_type>> m_elems;
 }; /* end class Plot3d */
--- a/tests/test_plot3d.py
+++ b/tests/test_plot3d.py
@@ -41,4 +41,54 @@ class Plot3dTC(unittest.TestCase):
         self.assertEqual(blk.clnds.ndarray[6:, :].tolist(),
                          [[8, 0, 2, 6, 4, 1, 3, 7, 5]])
 
+    def test_multi_block_cells_stay_in_their_block(self):
+        # Three disjoint blocks along x; the first has a different shape
+        # so a per-block size cannot be treated as a cumulative offset.
+        # Per block the file lists x, then y, then z of every node, with
+        # the x grid index running fastest; the node at grid (i, j, k)
+        # sits at that coordinate plus the block's x offset (0, 10, 20).
+        data = """3
+3 2 2
+2 2 2
+2 2 2
+0 1 2 0 1 2 0 1 2 0 1 2
+0 0 0 1 1 1 0 0 0 1 1 1
+0 0 0 0 0 0 1 1 1 1 1 1
+10 11 10 11 10 11 10 11
+0 0 1 1 0 0 1 1
+0 0 0 0 1 1 1 1
+20 21 20 21 20 21 20 21
+0 0 1 1 0 0 1 1
+0 0 0 0 1 1 1 1
+"""
+        blk = solvcon.core.Plot3d(data.encode('utf-8')).to_block()
+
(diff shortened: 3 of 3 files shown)


cpp/solvcon/inout/plot3d.hpp
```diff
@@ -63,7 +63,7 @@ class Plot3d
 private:
     inline uint_type coordinate_to_node_id(uint_type blk, uint_type x, uint_type y, uint_type z)
     {
-        return blk > 0 ? m_blk_sizes(blk - 1) : 0 + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk);
```
j8xixo12, line 66: Because the ternary conditional operator has a lower precedence than the addition operator, the original expression is parsed as `blk > 0 ? m_blk_sizes(blk - 1) : (0 + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk))`. However, the intended expression is `(blk > 0 ? m_blk_sizes(blk - 1) : 0) + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk)`."

tests/test_plot3d.py
```diff
@@ -41,4 +41,54 @@ def test_plot3d_parsing(self):
         self.assertEqual(blk.clnds.ndarray[6:, :].tolist(),
                          [[8, 0, 2, 6, 4, 1, 3, 7, 5]])
 
+    def test_multi_block_cells_stay_in_their_block(self):
```
j8xixo12, line 44: I also tested it on solvcon console: 

Image

tests/test_plot3d.py
```diff
@@ -41,4 +41,54 @@ def test_plot3d_parsing(self):
...
+        # Per block the file lists x, then y, then z of every node, with
+        # the x grid index running fastest; the node at grid (i, j, k)
+        # sits at that coordinate plus the block's x offset (0, 10, 20).
+        data = """3
+3 2 2
+2 2 2
+2 2 2
+0 1 2 0 1 2 0 1 2 0 1 2
+0 0 0 1 1 1 0 0 0 1 1 1
+0 0 0 0 0 0 1 1 1 1 1 1
```
j8xixo12, line 56: Intentionally using different block sizes here to verify that the block offset logic is correct.

cpp/solvcon/inout/plot3d.hpp
```diff
@@ -63,7 +63,7 @@ class Plot3d
...
+        return m_blk_offsets(blk) + x + y * m_x_shape(blk) + z * m_x_shape(blk) * m_y_shape(blk);
```
j8xixo12, line 66: The id of downstream block nodes rely on the cumulative count of all preceding block nodes as an offset. However, `m_blk_sizes` originally calculated only individual block sizes. To address this, `m_blk_offsets` is introduced to track the cumulative offset for each block.

`m_blk_offsets(blk)` is block 0 ~ (blk-1) total node counts.

cpp/solvcon/inout/plot3d.cpp
```diff
@@ -37,6 +38,7 @@ Plot3d::Plot3d(const std::string & data)
         m_y_shape(i) = std::stoul(tokens[1]);
         m_z_shape(i) = std::stoul(tokens[2]);
         m_blk_sizes(i) = m_x_shape(i) * m_y_shape(i) * m_z_shape(i);
+        m_blk_offsets(i) = static_cast<uint_type>(total_blk_size);
         total_blk_size += m_blk_sizes(i);
```
j8xixo12, line 42: When block size is 1, the `m_blk_offsets(0)` will be 0 naturally, because this loop does not execute.

tests/test_plot3d.py
```diff
@@ -41,4 +41,54 @@ def test_plot3d_parsing(self):
         self.assertEqual(blk.clnds.ndarray[6:, :].tolist(),
                          [[8, 0, 2, 6, 4, 1, 3, 7, 5]])
 
+    def test_multi_block_cells_stay_in_their_block(self):
+        # Three disjoint blocks along x; the first has a different shape
+        # so a per-block size cannot be treaded as a cumulative offset.
```
tigercosmos: `treated` typo.

tests/test_plot3d.py
```diff
@@ -41,4 +41,54 @@ def test_plot3d_parsing(self):
...
+10 11 10 11 10 11 10 11
+0 0 1 1 0 0 1 1
+0 0 0 0 1 1 1 1
+20 21 20 21 20 21 20 21
+0 0 1 1 0 0 1 1
+0 0 0 0 1 1 1 1
+"""
+        blk = solvcon.core.Plot3d(data.encode('utf-8')).to_block()
+
+        self.assertEqual(28, blk.nnode)
```
tigercosmos: nit: put actual first, and expectation second.

cpp/solvcon/inout/plot3d.cpp
tigercosmos, line 51: Not a bug, just a cleanup: after this fix there are still two places computing the block's node-id base.  Please check `base_idx` in `parseCoordinates`.

tigercosmos (changes requested):

tests/test_plot3d.py
j8xixo12: Thanks for the suggestion!

cpp/solvcon/inout/plot3d.cpp
j8xixo12: I have cleanup the base index computing code in latest commit.

tigercosmos (approved):

cpp/solvcon/inout/plot3d.hpp
```diff
@@ -76,6 +76,8 @@ class Plot3d
     SimpleArray<uint_type> m_y_shape;
     SimpleArray<uint_type> m_z_shape;
     SimpleArray<uint_type> m_blk_sizes;
+    /// Exclusive prefix sum of m_blk_sizes: the node-id base of each block.
```
yungyuc, line 79: This line of comment is redundant.  Please remove.

Issue #1285, opened 2026-08-07
MCAP signal processing step 1: Add time-series kernels: SimpleArray searchsorted/diff/cumsum and a new timeseries/ subsystem


tigercosmos:
This is the first of three issues adding signal-processing support to solvcon for analyzing recorded time-series logs, such as robotics or vehicle recordings where multiple signals are sampled at different rates and must be evaluated against criteria over time windows. This issue delivers the compute kernels; a follow-up issue adds a columnar MCAP reader, and a third grows `solvcon/track/DataFrame` into the signal-log container.

The kernels replace the per-sample Python loops such analysis otherwise needs: bisect-based sample lookup point by point, building union time grids with a set plus `sorted()`, and per-point predicate sweeps. Each kernel below turns one of those loops into a single array operation on `SimpleArray`.

### Semantics contract (binding for every kernel)

- Timestamps are integer nanoseconds (`SimpleArrayUint64`) end to end; convert to float only where the math requires it (the division in `deriv`).
- Kernels take (times, values) pairs of sorted 1-D contiguous arrays and return new arrays; inputs are never mutated.
- Half-open everywhere: windows `[start, end)`, trailing intervals `(t - T, t]`.
- Sample lookup is zero-order hold (ZOH) and strictly causal: the value at t is the last sample at or before t, never interpolated, never looking ahead. No sample at or before the query time means no value there.
- Timestamps are non-decreasing, so duplicates are allowed. A duplicate group resolves to its last sample, which is what an upper-bound `searchsorted` gives and what a recorded log needs when one publisher stamps two messages in the same nanosecond. `deriv` is the single exception: a zero time step has no derivative, so it requires strictly increasing timestamps and raises otherwise rather than dividing by zero and returning an infinity that later kernels would carry silently. `dedup_last` collapses a duplicate group to its last sample, so differentiating a log that has duplicates is one call rather than a Python loop.
- Window membership counts samples, it does not weight by time. A burst of samples sharing one timestamp therefore pulls a `movavg` window by how many of them there are.
- Transforms compute over the whole log first; windowing happens afterward, so there is no truncation artifact at window start.
[...]

PR #1328, opened 2026-08-10
Order `sort()` and `argsort()` the way numpy does

tigercosmos:
`sort()` and `argsort()` ordered through the built-in `<`, which is not a strict weak ordering once a NaN is present, so `std::sort()` was undefined over such an array. It showed: `sort()` of `[3, nan, 1, nan, 2]` answered `[1, nan, nan, 2, 3]` and `argsort()` answered the identity, where numpy answers `[1, 2, 3, nan, nan]` and `[2, 4, 0, 1, 3]`.

`nan_aware_less()` puts a NaN after every number and holds it equal to another NaN, which is numpy's rule and the one `<` cannot express because it answers false in both directions. A complex value carrying a NaN in either component goes past all the others as one group, ordered lexicographically within it.

`ComplexImpl::operator<()` answered true for equal values, which is the `<=` relation rather than the `<` that the lexicographic ordering documented above it calls for. A comparison that holds `a < a` is not a strict weak ordering either, so `std::sort()` was undefined over repeated complex values. The operator becomes strict only where no NaN is present, which is why the ordering algorithms use `nan_aware_less()` rather than these operators.

Two Python-visible changes follow from that operator, both tested here. `<` and `>` are bound on the complex scalar, so `complex128(1, 2) < complex128(1, 2)` now answers false where it answered true. Complex `argmin()` and `argmax()` resolve a tie to the first occurrence rather than the last, which is what numpy does.

`argsort()` now breaks ties by position, so equal values keep a deterministic order instead of the arbitrary one `std::sort()` left. While adding it I found that `argsort()` indexed from `body()` while numbering the whole array, so an array with ghost read past the end of its buffer and returned a different order on every run. It now indexes from `begin()`, the range that `sort()` and `take_along_axis()` already work on, and a test pins the order for a ghosted array.

The repeated one-dimension check in the sort mixin moves into a `validate_1d()` helper, which also fixes `take_along_axis_simd()` reporting the wrong function name in its error message.

## The NaN order is free on the real types

Apple M-series (arm64), macOS, Apple clang `-O3`, libc++. Figures come from a standalone microbenchmark that reproduces the code path in isolation; best of N runs, variants interleaved.

Passing any comparator to `std::sort()` costs 2.5x to 3.1x on every dtype. libc++ gates its branchless sorting-network path on `__desugars_to_v<__less_tag, _Compare>`, which only `std::less` and the no-comparator form satisfy; a lambda never does, however trivially it inlines. Partitioning NaN to the tail and sorting the rest comparator-free keeps that path, and costs nothing because every NaN is equal to every other, so the tail needs no ordering.

| 5e6 elements, shuffled | `int32` | `float64` |
| --- | --- | --- |
| no comparator (before this branch) | 98.6 ms | 99.8 ms |
| NaN-aware comparator passed to `std::sort` | 244.5 ms | 305.8 ms |
| partition NaN to tail, then comparator-free sort (this branch) | n/a | 101.0 ms |

Through the binding on 5e6 elements this branch measures 97 ms for `int32` and 106 ms for `float64`. This is a libc++ property; libstdc++ has no equivalent specialization, so the gap on a Linux build is expected to be smaller.

## Why `argsort()` does not use `std::stable_sort`

The positional tie-break calls the comparator up to twice per comparison where `std::stable_sort` would call it once, which argues for the latter. Measured across six data patterns it goes the other way: `std::sort` has adaptive paths for already-sorted and reversed input that `stable_sort` does not, and `stable_sort` still pays its merge. Both forms produced identical permutations in all 36 configurations tested, so this is purely about cost.

| n = 2e7 | tie-break `int32` | `stable_sort` `int32` | tie-break `float64` | `stable_sort` `float64` |
| --- | --- | --- | --- | --- |
| already sorted | **16.7 ms** | 303.2 ms | **22.8 ms** | 142.0 ms |
| reversed | **23.9 ms** | 598.6 ms | **30.2 ms** | 671.4 ms |
| all equal | **17.6 ms** | 141.6 ms | **23.0 ms** | 141.8 ms |
| random, all distinct | **1919.9 ms** | 5764.2 ms | **2203.7 ms** | 2351.6 ms |
| ties, n/100 distinct | **3134.6 ms** | 6096.8 ms | 2341.0 ms | **2009.3 ms** |
| ties, 8 distinct | 1375.6 ms | **797.5 ms** | 1832.8 ms | **656.1 ms** |

`stable_sort` wins only where a huge array holds a handful of distinct values. An already-sorted time index is the input `argsort()` sees in the case this series is for, and that is where it loses by 6x to 18x.

This is the first of two stacked pull requests. `SimpleArray::searchsorted()` builds on this ordering and follows in #1252.

Related to #1250 and #1285.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -25,8 +25,8 @@
 
 #include <algorithm>
 #include <array>
-#include <concepts>
 #include <cmath>
+#include <concepts>
 #include <format>
 #include <functional>
 #include <limits>
@@ -1323,6 +1323,68 @@ void SimpleArrayMixinCalculators<A, T>::find_two_bins(const uint32_t * freq, siz
     }
 }
 
+/**
+ * Order one value the way numpy sorts and searches it: a NaN goes after every
+ * number and counts equal to another NaN, which the built-in comparison cannot
+ * express because it answers false in both directions for a NaN. A complex
+ * value carrying a NaN in either component goes past all the others as one
+ * group, whatever its other component holds, and orders within that group
+ * lexicographically like any other pair.
+ */
+template <typename V>
+bool nan_aware_less(V const & lhs, V const & rhs)
+{
+    if constexpr (is_complex_v<V>)
+    {
+        bool const lhs_nan = std::isnan(lhs.real()) || std::isnan(lhs.imag());
+        bool const rhs_nan = std::isnan(rhs.real()) || std::isnan(rhs.imag());
+        if (lhs_nan != rhs_nan)
+        {
+            return rhs_nan;
+        }
+        if (!lhs_nan)
+        {
+            return lhs < rhs;
+        }
+
+        if (nan_aware_less(lhs.real(), rhs.real()))
+        {
+            return true;
+        }
+        if (nan_aware_less(rhs.real(), lhs.real()))
+        {
+            return false;
+        }
+        return nan_aware_less(lhs.imag(), rhs.imag());
+    }
+    else if constexpr (std::is_floating_point_v<V>)
+    {
+        if (std::isnan(rhs))
+        {
+            return !std::isnan(lhs);
+        }
+        if (std::isnan(lhs))
+        {
+            return false;
+        }
+        return lhs < rhs;
+    }
+    else
+    {
+        return lhs < rhs;
+    }
+}
+
+struct NanAwareLess
+{
+    template <typename V>
+    bool operator()(V const & lhs, V const & rhs) const { return nan_aware_less(lhs, rhs); }
+}; /* end struct NanAwareLess */
+
+// TODO: every member here walks begin() to end() and so reads the buffer in
+// storage order, ignoring stride(). A strided one-dimensional view is accepted
+// and then sorted as the wrong span. Reject a non-contiguous receiver the way
+// argmin() already does, in a change that covers the whole mixin at once.
 template <typename A, typename T>
 class SimpleArrayMixinSort
 {
@@ -1342,20 +1404,50 @@ public:
     template <IntegralType I>
     A take_along_axis_simd(SimpleArray<I> const & indices);
 
+private:
+
+    void validate_1d(char const * op, char const * what = "array") const;
+
 }; /* end class SimpleArrayMixinSort */
+
+template <typename A, typename T>
+void SimpleArrayMixinSort<A, T>::validate_1d(char const * op, char const * what) const
+{
+    auto const * athis = static_cast<A const *>(this);
+    if (athis->ndim() != 1)
+    {
+        throw std::runtime_error(
+            std::format("SimpleArray::{}(): "
+                        "currently only support 1D array but the {} is {} dimension",
+                        op,
+                        what,
+                        athis->ndim()));
+    }
+}
 
 template <typename A, typename T>
 void SimpleArrayMixinSort<A, T>::sort()
 {
+    validate_1d("sort");
     auto athis = static_cast<A *>(this);
-    if (athis->ndim() != 1)
-    {
-        throw std::runtime_error(
-            std::format(
-                "SimpleArray::sort(): currently only support 1D array but the array is {} dimension", athis->ndim()));
-    }
 
-    std::sort(athis->begin(), athis->end());
+    if constexpr (is_complex_v<value_type>)
+    {
+        // Complex numbers are sorted lexicographically by real and then imaginary parts.
+        std::sort(athis->begin(), athis->end(), NanAwareLess{});
+    }
+    else if constexpr (std::is_floating_point_v<value_type>)
+    {
+        // Partition the array into two parts: non-NaN values and NaN values.
+        auto const mid = std::partition(athis->begin(), athis->end(), [](value_type const & v)
+                                        { return !std::isnan(v); });
+        std::sort(athis->begin(), mid);
+    }
+    else
+    {
+        // For integral types, we can use the default sort.
+        std::sort(athis->begin(), athis->end());
+    }
 }
 
 template <typename T, IntegralType I>
@@ -2913,14 +3005,8 @@ void SimpleArray<T>::copy_logical_into(SimpleArray & out) const
 template <typename A, typename T>
 SimpleArray<uint64_t> detail::SimpleArrayMixinSort<A, T>::argsort()
 {
+    validate_1d("argsort");
     auto athis = static_cast<A *>(this);
-    if (athis->ndim() != 1)
-    {
-        throw std::runtime_error(
-            std::format("SimpleArray::argsort(): "
-                        "currently only support 1D array but the array is {} dimension",
-                        athis->ndim()));
-    }
 
     SimpleArray<uint64_t> ret(athis->shape());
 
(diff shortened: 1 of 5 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1323,6 +1323,68 @@ void SimpleArrayMixinCalculators<A, T>::find_two_bins(const uint32_t * freq, siz
...
+        }
+        return lhs < rhs;
+    }
+    else
+    {
+        return lhs < rhs;
+    }
+}
+
+struct NanAwareLess
```
tigercosmos, line 1378: New `NanAwareLess` for numpy-like sorting.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1323,6 +1323,68 @@ void SimpleArrayMixinCalculators<A, T>::find_two_bins(const uint32_t * freq, siz
...
+{
+    template <typename V>
+    bool operator()(V const & lhs, V const & rhs) const { return nan_aware_less(lhs, rhs); }
+}; /* end struct NanAwareLess */
+
+// TODO: every member here walks begin() to end() and so reads the buffer in
```
tigercosmos, line 1384: TODO for stride.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1342,20 +1404,50 @@ class SimpleArrayMixinSort
...
+void SimpleArrayMixinSort<A, T>::sort()
+{
+    validate_1d("sort");
+    auto athis = static_cast<A *>(this);
 
-    std::sort(athis->begin(), athis->end());
+    if constexpr (is_complex_v<value_type>)
+    {
+        // Complex numbers are sorted lexicographically by real and then imaginary parts.
+        std::sort(athis->begin(), athis->end(), NanAwareLess{});
```
tigercosmos, line 1437: Handle complex with the nan-aware method.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1342,20 +1404,50 @@ class SimpleArrayMixinSort
...
+    }
+    else if constexpr (std::is_floating_point_v<value_type>)
+    {
+        // Partition the array into two parts: non-NaN values and NaN values.
+        auto const mid = std::partition(athis->begin(), athis->end(), [](value_type const & v)
```
tigercosmos, line 1442: Sort floating point numbers with non-nana and nan values.

doc/source/compute/buffer/reduce.md
```diff
@@ -164,6 +164,20 @@ sarr.sort()
 assert sarr.ndarray.tolist() == [1.0, 2.0, 3.0]
 ```
 
+The order is the numpy one rather than the one the built-in `<` gives. A NaN
```
tigercosmos: Update the doc.

doc/source/compute/buffer/reduce.md
```diff
@@ -164,6 +164,12 @@ sarr.sort()
 assert sarr.ndarray.tolist() == [1.0, 2.0, 3.0]
 ```
 
+NaN sorts last, as numpy sorts it: after every number, and equal to another
+NaN. A complex value carrying a NaN in either component sorts after all the
```
yungyuc: Write explicitly how complex numbers are sorted.  Order lexicographically is not clear enough.  Use examples.

@j8xixo12 FYI.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1323,6 +1323,68 @@ void SimpleArrayMixinCalculators<A, T>::find_two_bins(const uint32_t * freq, siz
...
+    {
+        bool const lhs_nan = std::isnan(lhs.real()) || std::isnan(lhs.imag());
+        bool const rhs_nan = std::isnan(rhs.real()) || std::isnan(rhs.imag());
+        if (lhs_nan != rhs_nan)
+        {
+            return rhs_nan;
+        }
+        if (!lhs_nan)
+        {
+            return lhs < rhs;
```
yungyuc, line 1347: This gonna be tricky, because [`Complex.hpp:146-148`](https://github.com/solvcon/solvcon/blob/3cadef6cfa818c5836280f8d848b24bad7ce7dd4/cpp/solvcon/math/Complex.hpp#L146-L148) writes:

```cpp
        return lhs.imag_v <= rhs.imag_v;
    }
    return lhs.real_v <= rhs.real_v;
```

@j8xixo12 Please clarify why it was written this way.

yungyuc (changes requested):
- [x] Clarify if there is a bug in the complex number comparison.  @j8xixo12 .
- [x] Write explicitly how complex numbers are sorted in the document.

@Yuli-ytd Please take a look.

doc/source/compute/buffer/reduce.md
tigercosmos: Sure, add an example in the doc.

cpp/solvcon/buffer/SimpleArray.hpp
tigercosmos: It looks like a slip rather than a choice. `b278719f` converted both operators from `norm()` to lexicographic in a single commit, and `operator>` came out strict while `operator<` got `<=`. The `norm()` form it replaced was strict on both sides as well.

`tests/test_math.py::test_operator_comparison` looks like it pins the `<=`, but it recomputes the same expression the implementation uses, and both fixtures have `real1 != real2` (0.7 vs 2.5, 4.3 vs 6.1). That path returns `real1 <= real2`, which agrees with `<` whenever the reals differ, so the test cannot tell the two apart and passed unchanged after the fix.

  The effect is not cosmetic: a comparison for which `a < a` holds is not a strict weak ordering, so `std::sort` behind `sort()` and `argsort()` was undefined over repeated complex values. It is also Python-visible, since `<` is bound on the complex scalar: `complex128(1, 2) < complex128(1, 2)` answered `True` before and answers `False` now.

This branch rewrites that test to assert the relations directly and cover the equal case. I checked it is meaningful by reverting `operator<` to `<=` and rebuilding: the updated test fails on both `ComplexFp32TC` and `ComplexFp64TC`, where the old one passed.

One note: the `<=` came from @ThreeMonth03 in `b278719f`; @j8xixo12 appears around the earlier `norm()`-based version, in case the question should go elsewhere.

doc/source/compute/buffer/reduce.md
```diff
@@ -164,6 +164,25 @@ sarr.sort()
...
+
+A complex value compares by its real part, and by its imaginary part only when
+the real parts are equal. `2-1j` therefore sorts before `2+1j`, and both sort
+after every value with a smaller real part:
+
+```python
+narr = np.array([2 + 1j, 1 + 5j, 2 - 1j, 0 + 0j], dtype='complex128')
+sarr = solvcon.SimpleArrayComplex128(array=narr)
+sarr.sort()
+assert sarr.ndarray.tolist() == [0j, 1 + 5j, 2 - 1j, 2 + 1j]
```
tigercosmos, line 179: Added a new example in the doc.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: I created issue #1336 to track this.

PR #1329, opened 2026-08-10
Drop the unused extra pack from `WrapBase::commit`

yungyuc:
`WrapBase::commit()` takes a variadic `Extra` pack and forwards it into the derived wrapper's constructor, but every one of the 116 call sites in the tree passes exactly the three arguments `mod`, `pyname` and `pydoc`. The pack has never been deduced as non-empty.

### DIFF
--- a/cpp/solvcon/python/common.hpp
+++ b/cpp/solvcon/python/common.hpp
@@ -277,14 +277,12 @@ public:
     >;
     // clang-format on
 
-    template <class... Extra>
-    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc, Extra &&... extra)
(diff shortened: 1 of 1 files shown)


cpp/solvcon/python/common.hpp
```diff
@@ -277,14 +277,12 @@ WrapBase
     >;
     // clang-format on
 
-    template <class... Extra>
-    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc, Extra &&... extra)
+    static wrapper_type & commit(pybind11::module & mod, char const * pyname, char const * pydoc)
```
yungyuc, line 280: The variadic pack is never used.  Remove it.

PR #1330, opened 2026-08-10
fix: remove redundant comment from plot3d.hpp

j8xixo12:
This PR is to follow-up the #1327  to remove the redundant comment from `plot3d.hpp`.

Related to #1326


tigercosmos (approved):
LGTM

PR #1331, opened 2026-08-10
Add the plot model

### LINKED ISSUE #1049: Use native Qt to draw XY plot for 1D Euler solver

j8xixo12:
`RPlotModel` completes the series-and-model half of the native xy plot core: it keeps the series list in shared ownership with the Python side, colors an uncolored series from the `matplotlib` cycle without spending a slot on an explicitly colored one, unions the per-series NaN-safe data limits, and autoscales the view limits by opening a singular span before applying the margin.  `view()` maps the view limits onto a screen as a `ViewTransform2dFp64`, reusing the existing affine instead of adding one; the single zoom fits the view box inside the screen and centers it, leaving a per-axis stretch to the widget.

Related to #1049 step 1

### DIFF
--- a/cpp/solvcon/pilot/plot/RPlotModel.cpp
+++ b/cpp/solvcon/pilot/plot/RPlotModel.cpp
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/pilot/plot/RPlotModel.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <format>
+#include <stdexcept>
+#include <utility>
+
+namespace solvcon
+{
+
+namespace
+{
+
+/**
+ * The nonsingular guard, then the autoscale margin: a degenerate span is
+ * first opened around its value so the margin has a span to scale.
+ */
+std::pair<double, double> expand_axis(double lo, double hi, double margin)
+{
+    if (lo == hi)
+    {
+        double const half = (lo == 0.0) ? 0.5 : std::abs(lo) * 0.05;
+        lo -= half;
+        hi += half;
+    }
+
+    double const pad = (hi - lo) * margin;
+    return {lo - pad, hi + pad};
+}
+
+void validate_axis_limits(double lo, double hi, char const * axis)
+{
+    if (!std::isfinite(lo) || !std::isfinite(hi) || !(lo < hi))
+    {
+        throw std::invalid_argument(
+            std::format(
+                "RPlotModel::set_view_limits: {} limits must be finite and increasing, but they are {} and {}",
+                axis,
+                lo,
+                hi));
+    }
+}
+
+} /* end namespace */
+
+std::shared_ptr<RPlotSeries> RPlotModel::add_series(std::shared_ptr<RPlotSeries> const & series)
+{
+    if (!series)
+    {
+        throw std::invalid_argument("RPlotModel::add_series: series must not be None");
+    }
+
+    if (!series->color_is_set())
+    {
+        series->set_color(plot_cycle_color(m_cycle_index++));
+    }
+    m_series.push_back(series);
+    return m_series.back();
+}
+
+std::shared_ptr<RPlotSeries> const & RPlotModel::series(std::size_t it) const
+{
+    if (it >= m_series.size())
+    {
+        throw std::out_of_range(
+            std::format("RPlotModel::series: index {} is out of bounds with size {}", it, m_series.size()));
+    }
+    return m_series[it];
+}
+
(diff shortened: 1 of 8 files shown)


tests/test_pilot_plot.py
```diff
@@ -189,4 +189,104 @@ def test_bad_line_width_is_rejected(self):
...
+                model.series(index)
+
+    def test_data_limits_union_all_series(self):
+        model = pilot.RPlotModel()
+        self.assertIsNone(model.data_limits())
+        model.add_series().set_data(_array([0.0, 1.0]), _array([5.0, 6.0]))
+        model.add_series()
+        self.assertEqual((0.0, 1.0, 5.0, 6.0), model.data_limits())
+        model.add_series().set_data(_array([-3.0, 0.5]), _array([7.0, 8.0]))
+        self.assertEqual((-3.0, 1.0, 5.0, 8.0), model.data_limits())
```
j8xixo12, line 238: Using 3 different series to test whether `data_limits()` union all series limits. And the 2nd series is an empty series, it should not affect the `data_limits()` results.

cpp/solvcon/pilot/plot/plot_style.hpp
```diff
@@ -48,6 +48,12 @@ struct PlotColor
 /// Default stroke width in screen pixels; matplotlib's lines.linewidth.
 inline constexpr double PLOT_DEFAULT_LINE_WIDTH = 1.5;
 
+/**
+ * Default autoscale margin as a fraction of the data span on each side;
+ * matplotlib's axes.xmargin and axes.ymargin.
+ */
+inline constexpr double PLOT_DEFAULT_MARGIN = 0.05;
```
yungyuc, line 55: These default values should not be made as compile time constant.  They should be populated during runtime.  Please create an issue to track the work to design a default value container for the plotting sub-system.

cpp/solvcon/pilot/plot/RPlotModel.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    if (it >= m_series.size())
+    {
+        throw std::out_of_range(
+            std::format("RPlotModel::series: index {} is out of bounds with size {}", it, m_series.size()));
+    }
+    return m_series[it];
+}
+
+std::optional<std::array<double, 4>> RPlotModel::data_limits() const
```
yungyuc, line 77: Returning a large static array suggests that the data are not properly structured and organized using a class.  Please fix it in subsequent PRs.

cpp/solvcon/pilot/plot/RPlotModel.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include <solvcon/universe/ViewTransform2d.hpp>
+
+#include <solvcon/pilot/plot/plot_style.hpp>
+#include <solvcon/pilot/plot/RPlotSeries.hpp>
+
+namespace solvcon
+{
+
+class RPlotModel
```
yungyuc, line 37: Use a doxygen comment block to summarize that it does.  No need to be long.

yungyuc (changes requested):
- [x] Use a doxygen comment block to summarize that `RPlotModel` does.  No need to be long.

cpp/solvcon/pilot/plot/plot_style.hpp
yungyuc: Issue #1344.

cpp/solvcon/pilot/plot/RPlotModel.hpp
j8xixo12: I added a brief summary describing what the `RPlotModel` does.

cpp/solvcon/pilot/plot/RPlotModel.cpp
j8xixo12: I will use a follow-up PR to improve this.

cpp/solvcon/pilot/plot/RPlotModel.hpp
yungyuc: Autoscaling usually induces unwanted effect.  Let's observe.

yungyuc (approved):
LGTM

Issue #1325, opened 2026-08-09
Make the common warning flags impossible to lose, and cover gtests


tigercosmos:
#1310 found that `COMMON_COMPILER_OPTIONS` was set after `add_subdirectory(cpp/solvcon)`, so the core libraries were built with no warning flags for a long time without anyone noticing. The fix moves the block above the core. That restores the flags, but it leaves the failure mode in place: the variable is expanded where it is used, so an empty or not yet defined variable expands to nothing and `target_compile_options(... ${COMMON_COMPILER_OPTIONS})` silently becomes a no-op. The build stays green and nothing reports that the flags went missing.

Two changes would turn that silent failure into a loud one.

First, carry the flags on an `INTERFACE` target instead of a variable. An `add_library(solvcon_warnings INTERFACE)` with `target_compile_options(solvcon_warnings INTERFACE ...)`, linked by each consumer, makes the same ordering mistake a hard error at generate time ("links to target `solvcon_warnings` but the target was not found") rather than a quiet loss of coverage. The option set is currently written out by hand in three places, `cpp/solvcon/CMakeLists.txt`, `cpp/binary/pymod_solvcon/CMakeLists.txt`, and `cpp/binary/pilot/CMakeLists.txt`, and those three copies have already drifted apart in which suppressions they carry. Folding them into one or two interface targets would make the drift visible.

Second, cover `gtests`. `gtests/CMakeLists.txt` never referenced `COMMON_COMPILER_OPTIONS` at all, so the `test_nopython` target still compiles roughly thirty core sources with no warning flags. This is not merely a duplicate compilation of what `solvcon_primary` already covers: `test_nopython` does not link `solvcon_primary`, so it never receives the `PUBLIC` `SC_HAS_CBLAS`, `SC_HAS_MKL`, and `SC_HAS_VENDOR_LAPACK` definitions. Those translation units compile under a different preprocessor configuration and take branches that `solvcon_primary` never builds. The non-vendor fallback paths in the matmul and eigen code are therefore the only core code with no warning coverage even after #1310 is fixed.

A related detail worth folding in: `-Wno-unused-value` is applied to the whole of `solvcon_primary`, justified by `PYBIND11_EXPAND_SIDE_EFFECTS` in a pybind11 header. That suppression was inert while the option set expanded to nothing. Now that the flags are live it is a real hole across the core, most of which contains no pybind11 at all. Scoping it to the `pymod` sources would keep `-Wunused-value` working where it can still catch a discarded expression statement.

PR #1333, opened 2026-08-10
Stop a gcc false positive in pybind11 from failing the default build

tigercosmos:
A bare `make` configures RelWithDebInfo, which this project compiles at `-O3 -g`. At that flag pair gcc 16 reports `may be used uninitialized` against `tuple result(size)` in pybind11's `cast.h`, and `-Werror` turns it into an error, so every default local build on Linux stops at `cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp`. The report is wrong, because pybind11 fills that tuple.

gcc computes `-Wmaybe-uninitialized` after inlining, so what it sees depends on how the code was optimized, and only `-O3 -g` produces the shape it misreads. `-O3` alone, `-O2 -g`, and `-O1 -g` all compile the same file clean. That is why no CI job reports it: the ubuntu build job configures Release, and the RelWithDebInfo job runs on macOS, where clang has no such warning. Every ingredient is in CI, but that combination is not.

The change adds `-Wno-error=maybe-uninitialized` to `COMMON_COMPILER_OPTIONS` for gcc only. The `-Wno-error=` form rather than `-Wno-` keeps a real uninitialized use in the log and drops only the escalation to an error. The guard on the compiler id is required rather than defensive: clang has no `-Wmaybe-uninitialized` group and rejects the flag under `-Werror`, so an unguarded option would break every clang and AppleClang build on its first file.

Marking the pybind11 headers SYSTEM was tried first and does not work. The diagnostic does point into `cast.h`, but gcc's late warnings are emitted past the point where system-header suppression applies, and `-isystem` leaves the error unchanged.

One caveat for the reviewer, which is why this references #1325: the suppression is tree-wide, and #1325 argues for scoping such suppressions down rather than widening them. Wrapping the pybind11 include in a `#pragma GCC diagnostic` region confines the relaxation to that header and was verified to work, but doing it robustly wants a choke-point header for the 55 pybind11 includes spread across 32 files, so it belongs with that issue rather than here.

With the change, `make pilot` completes and prints the warning once, and `make lint` passes.

Related to #1325.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,6 +380,13 @@ if(MSVC)
     set(COMMON_COMPILER_OPTIONS /W4 /bigobj /WX)
 else()
     set(COMMON_COMPILER_OPTIONS -Wall -Wextra)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        # gcc computes -Wmaybe-uninitialized after inlining, so at -O3 -g it
+        # fires inside pybind11 on a tuple that is in fact initialized, and
+        # -O3 -g is what a bare `make` configures. Demote rather than disable,
+        # so a real uninitialized use still reaches the log.
+        set(COMMON_COMPILER_OPTIONS ${COMMON_COMPILER_OPTIONS} -Wno-error=maybe-uninitialized)
+    endif()
     if(USE_CLANG_TIDY AND LINT_AS_ERRORS)
         message(STATUS "Disable -Werror because clang-tidy (with lint as errors) is enabled")
     else()


CMakeLists.txt
```diff
@@ -380,6 +380,13 @@ if(MSVC)
     set(COMMON_COMPILER_OPTIONS /W4 /bigobj /WX)
 else()
     set(COMMON_COMPILER_OPTIONS -Wall -Wextra)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        # gcc computes -Wmaybe-uninitialized after inlining, so at -O3 -g it
+        # fires inside pybind11 on a tuple that is in fact initialized, and
+        # -O3 -g is what a bare `make` configures. Demote rather than disable,
+        # so a real uninitialized use still reaches the log.
```
yungyuc, line 387: Good call to keep it a warning.

Issue #1332, opened 2026-08-10
Give SimpleArray<bool> reductions the result types numpy uses


tigercosmos:
`SimpleArray<T>` is one template instantiated for every value type, and its reductions and elementwise arithmetic are written for numeric types. Instantiated for `bool`, the arithmetic still compiles because `bool` promotes to `int` and truncates back, so an accumulator declared `bool` can never hold a count, a mean, or a variance. The results that fall out of that are accidental rather than chosen.

MSVC names the problem directly. C4804 ("unsafe use of type `bool` in an arithmetic operation") fires at fifteen sites in `buffer/SimpleArray.hpp`, all of them inside `mean`, `average`, `var`, and `symmetrize`. #1324 clears the rest of the MSVC backlog but leaves `/wd4804` in `cpp/solvcon/CMakeLists.txt`, because removing it means deciding what these operations mean for `bool`, which is an API question and not a warning fix.

## What the code does today

Measured on the current `master` build, comparing against numpy 1.26.4 for the same input.

| input | method | solvcon | numpy | numpy dtype |
| --- | --- | --- | --- | --- |
| `[T,F,T,F]` | `sum` | `True` | `2` | int64 |
| `[T,F,T,F]` | `mean` | `True` | `0.5` | float64 |
| `[T,F,T,F]` | `var` | `False` | `0.25` | float64 |
| `[T,F,T,F]` | `std` | `False` | `0.5` | float64 |
| `[T,F,T,F]` | `median` | `True` | `0.5` | float64 |
| `[T,F,T,F]` | `average(w)` | `True` | `0.5` | float64 |
| `[T,F,F,F]` | `mean` | `True` | `0.25` | float64 |
| `[T,F,F,F]` | `var` | `False` | `0.1875` | float64 |
| `[T,T,T,T]` | `var` | `False` | `0.0` | float64 |

`mean` and `average` come out as "any element is true", which is what `bool(np.mean(x))` also gives, so those two are defensible as they stand. `var` and `std` are not: they return `False` for every boolean array whatever the data, because the squared differences collapse into a `bool` before the division. That is a wrong answer, not a convention.

## What numpy does

numpy chooses the output dtype that can hold the answer, and keeps `bool` only where the answer is genuinely boolean.

| expression | numpy result | out dtype |
| --- | --- | --- |
| `a + b` | disjunction | bool |
| `a * b` | conjunction | bool |
[...]

Issue #1324, opened 2026-08-09
Fix the MSVC warning backlog the core libraries now expose


tigercosmos:
`COMMON_COMPILER_OPTIONS` was set after `add_subdirectory(cpp/solvcon)`, so the two expansions in `cpp/solvcon/CMakeLists.txt` produced nothing and the core libraries were built with no warning flags. Moving the block above the core (see the pull request related to #1310) turns `/W4 /WX` on for the core on MSVC for the first time, and that exposes a backlog the Clang and GCC cleanup did not cover.

To keep the move landable, the core suppresses four MSVC codes for now. This issue tracks fixing the underlying code and removing the suppressions one code at a time.

The first Windows build reported at least 33 distinct sites. MSVC stops a translation unit at its first error, so that is a lower bound rather than a count.

| Code | Meaning | Where it concentrates |
| --- | --- | --- |
| C4804 | Unsafe use of type `bool` in an arithmetic operation | `buffer/SimpleArray.hpp` reductions, `buffer/matmul.hpp` accumulation loops |
| C4244 | Conversion with possible loss of data | `buffer/SimpleArray.hpp`, `buffer/pymod/wrap_SimpleArray.hpp` |
| C4267 | `size_t` conversion with possible loss of data | `buffer/SimpleArray.hpp` |
| C4458 | Declaration hides a class member | `buffer/pymod/array_common.hpp` |

Nearly all of these come from template bodies instantiated across every supported value type, so one source line produces one warning per instantiation. C4804 in particular fires where a reduction written for numeric types is instantiated for `bool`; the arithmetic is intended there, so the fix is to make the accumulator type explicit rather than to silence the diagnostic.

Suggested order, smallest and most contained first: C4458 (rename the shadowing locals), then C4267 and C4244 (make the narrowing explicit where it is intended and correct it where it is not), then C4804 (give the reductions an explicit accumulator type). Remove each `/wd` entry from `cpp/solvcon/CMakeLists.txt` as its code reaches zero, so the build never silently regains a suppression it no longer needs.

Note that a Windows build is the only way to see these. They cannot be reproduced with Clang or GCC.

tigercosmos:
[...]

PR #1335, opened 2026-08-10
Clear the MSVC C4458, C4244, C4267 and C4702 backlog in the core libraries

tigercosmos:
Moving `COMMON_COMPILER_OPTIONS` above the core turned `/W4 /WX` on for the core libraries for the first time, and to keep that move landable the core suppressed five MSVC codes. This clears four of them and drops their `/wd` entries, leaving only `/wd4804`.

C4458 was two parameters that happen to share a name with a member in scope: `direct_converter()`'s `value` against the `value` of the enclosing `npy_format_descriptor`, and `writeToHistory()`'s `data` against `QWidget::data`. They are renamed to `storage` and `text`. C4244 and C4267 are implicit narrowings, mostly in template bodies instantiated over every value type, so a single line reports once per instantiation; each is now spelled out. Two of them were real rather than intended: `CubicBezierSampler` now takes the `uint8_t ndim` that `SegmentPad` wants instead of narrowing a `size_t` at the call, and `Plot3d::build_interior()` binds the element it already looked up instead of rebuilding the map key three times per iteration. C4702 came from two shapes that leave a statement no execution can reach once instantiated: `isub()` and `idiv()` threw from the `bool` branch of an `if constexpr` and then fell through to a `return` shared with the numeric branch, and `MatmulExecutor::execute_at()` guarded its BLAS dispatch with `#if` around an `if constexpr` that returns. The throw now lives inside the branch, and a `has_blas_backend` constant lets `if constexpr` discard the dispatch rather than the preprocessor delete it.

This is meant to be a pure warning fix, so every cast wraps the expression it makes explicit rather than retyping an operand. That distinction is not cosmetic: an earlier attempt wrote `T{1} - t` for `1.0 - t` in `bernstein.hpp`, which moved a float32 intermediate from `double` to `float` and broke `CurvePadFp32TC::test_sample_2d`. For the same reason `var()` keeps its product in the `size_t` the count promotes to, where overflow is defined to wrap, and the `dft()` angle is deduced with `auto` so a `long double` instantiation keeps its precision. No `SimpleArray<bool>` result changes: `sum`, `mean`, `median`, `var`, `std`, `average`, `symmetrize`, `trace`, `matmul`, the elementwise operations and the `*_simd` variants were compared against master and are byte-identical, and the exception type and message text for `sub`/`div`/`isub`/`idiv` are unchanged.

`/wd4702` could not be removed outright. Both remaining reports name Qt's own `qnumeric.h`, where there is no solvcon line to fix, and `/external:W0` cannot reach them because MSVC raises C4702 from the code generator rather than the front end. A full rebuild with every suppression stripped shows exactly one translation unit affected, so the suppression is applied to `RDrawable.cpp` alone. The rest of the pilot keeps C4702 in force, and a second unit reaching the same inlining will fail the build and name itself.

C4804 stays suppressed. It fires only in the `SimpleArray<bool>` reductions, where clearing it means deciding what `mean`, `average` and variance mean for `bool` rather than rewriting an expression, which is an API question rather than a warning fix. Issue #1332 records the measured solvcon-versus-numpy behaviour and tracks that work; the comment on `/wd4804` now points at it.

Verified on a Windows VM with MSVC 19.44: a `/W4 /WX` build of the module, the pilot and the gtest binary compiles with no warnings and no errors, gtest passes 242 of 242, and the pilot's in-binary pytest passes 1800 with none failing. A survey build with `/WX` and every `/wd` removed reports C4458, C4244, C4267 and C4702 at zero across 189 translation units, down from 14, 70, 26 and 32. On Linux the suite is unchanged against master: 1618 passed, with `test_average_with_axis` failing identically before and after in this environment. `make lint` is clean.

Related to #1324.

### DIFF
--- a/cpp/solvcon/CMakeLists.txt
+++ b/cpp/solvcon/CMakeLists.txt
@@ -340,14 +340,11 @@ foreach (SOLVCON_LIBRARY IN LISTS SOLVCON_LIBRARIES)
             ${SOLVCON_LIBRARY} PRIVATE
             ${COMMON_COMPILER_OPTIONS}
             /bigobj # C1128: number of sections exceeded object file format limit
-            # TODO: fix the code and drop each of these; see issue #1324. They
-            # come from templates instantiated over every value type, so one
-            # line warns once per instantiation.
+            # This is the last suppression left from issue #1324. It fires in
+            # the SimpleArray<bool> reductions, where clearing it means
+            # deciding what mean, average and variance mean for bool rather
+            # than rewriting an expression. Issue #1332 tracks that.
             /wd4804 # bool in an arithmetic operation, from reductions over SimpleArray<bool>
-            /wd4244 # narrowing conversion
-            /wd4267 # narrowing conversion from size_t
-            /wd4458 # a declaration hides a class member
-            /wd4702 # unreachable code, after the `if constexpr` bool branches that throw
             /external:W0 # do not apply /W4 to a SYSTEM include such as numpy
         )
     else () # MSVC
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -230,6 +230,11 @@ public:
 
 }; /* end class SimpleArrayMixinModifiers */
 
+// SimpleArray<bool> reaches these mixins like any other value type, and the
+// operations that have no boolean meaning have to say so one by one.
+template <typename U>
+inline constexpr bool is_bool_v = std::is_same_v<bool, std::remove_const_t<U>>;
+
 template <typename U>
 struct select_real_t
 {
@@ -551,7 +556,7 @@ public:
         {
             sum += v;
         }
-        return sum / static_cast<value_type>(n);
+        return sum / convert_count(n);
     }
 
     A mean(const shape_type & axis) const
@@ -567,7 +572,7 @@ public:
         {
             throw std::runtime_error("SimpleArray::mean(): empty array");
         }
-        return athis->sum() / static_cast<value_type>(n);
+        return athis->sum() / convert_count(n);
     }
 
     real_type var_op(small_vector<value_type> & sv, size_t ddof) const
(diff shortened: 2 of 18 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -738,16 +761,49 @@ class SimpleArrayMixinCalculators
...
 
 private:
 
+    // Subtraction and division have no meaning on bool. Throwing from inside
+    // the `if constexpr` branch, instead of falling through to a shared
+    // `return`, keeps the bool instantiation free of a statement no execution
+    // can reach. The out-of-place `sub()` and `div()` pass the name of the
+    // in-place operation they delegate to, so the thrown message stays the one
+    // callers already see.
+    [[noreturn]] static void reject_bool_operation(char const * op)
```
tigercosmos, line 794: Reject bool operation.

cpp/solvcon/math/blas_compat.hpp
```diff
@@ -14,6 +14,20 @@
...
+/**
+ * Whether this build has a BLAS backend behind the wrappers below.
+ *
+ * A constant rather than an `#if` at each call site, so a caller can discard
+ * its BLAS branch through `if constexpr` instead of having the preprocessor
+ * delete it. That leaves no unreachable statement behind the branch, which
+ * MSVC reports as C4702 from its code generator.
+ */
+#if (defined(__APPLE__) && defined(__arm64__)) || defined(SC_HAS_CBLAS)
+inline constexpr bool has_blas_backend = true;
```
tigercosmos, line 26: A constant to prevent MSVC error.

cpp/solvcon/pilot/visual/RDrawable.cpp
```diff
@@ -3,7 +3,20 @@
  * BSD 3-Clause License, see COPYING
  */
 
+// Qt's `qnumeric.h` is the only place C4702 still fires, from a dead branch
```
tigercosmos, line 6: QT error workaround for MSVC.

cpp/solvcon/CMakeLists.txt
```diff
@@ -340,14 +340,11 @@ foreach (SOLVCON_LIBRARY IN LISTS SOLVCON_LIBRARIES)
             ${SOLVCON_LIBRARY} PRIVATE
             ${COMMON_COMPILER_OPTIONS}
             /bigobj # C1128: number of sections exceeded object file format limit
-            # TODO: fix the code and drop each of these; see issue #1324. They
-            # come from templates instantiated over every value type, so one
-            # line warns once per instantiation.
+            # This is the last suppression left from issue #1324. It fires in
+            # the SimpleArray<bool> reductions, where clearing it means
+            # deciding what mean, average and variance mean for bool rather
+            # than rewriting an expression. Issue #1332 tracks that.
```
tigercosmos, line 346: TODO: #1332 for `/wd4804 `

PR #1347, opened 2026-08-12
Add `SimpleArray::diff()` and `cumsum()`

tigercosmos:
`diff()` subtracts each element of a one-dimensional array from the element after it, and `cumsum()` accumulates the running total. Both return a new array of the receiver's class. Issue #1285 needs them for its time-series kernels. `validate_1d()` becomes a free function template, so both mixins share one check.

Both compute in the element type and wrap at the numeric boundary, through the unsigned counterpart where C++ defines the wrap. `numpy.diff` wraps the same way, but `numpy.cumsum` promotes a narrow integer array: `uint8` `[200, 100]` gives `[200, 300]` there and `[200, 44]` here. `SimpleArrayBool::diff()` refuses as `sub()` refuses, and `cumsum()` there accumulates with logical or, as `sum()` does.

The tests compare twelve dtypes against numpy and cover the length edges, both wraps, the bool cases, a strided view, and a ghosted receiver. `make lint` is clean, `make pytest` passes with 2197 tests, and `make gtest` with 244.

Related to #1285.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -222,6 +222,24 @@ private:
 std::string format_shape(shape_type const & shape);
 std::string format_flat_index(shape_type const & shape, ssize_t offset);
 
+/**
+ * Reject an array that is not one-dimensional. `op` names the calling
+ * operation and `what` the role the array plays in it, both for the message.
+ */
+template <typename A>
+void validate_1d(A const & array, char const * op, char const * what = "array")
+{
+    if (array.ndim() != 1)
+    {
+        throw std::runtime_error(
+            std::format("SimpleArray::{}(): "
+                        "currently only support 1D array but the {} is {} dimension",
+                        op,
+                        what,
+                        array.ndim()));
+    }
+}
+
 template <typename T>
 struct SimpleArrayInternalTypes
 {
@@ -732,6 +750,12 @@ public:
         return ret;
     }
 
+    /// Subtract every element of a one-dimensional receiver from the element after it.
+    A diff() const;
+
+    /// Accumulate a one-dimensional receiver into its running total.
+    A cumsum() const;
+
     A add(A const & other) const
     {
         validate_same_shape(other, "add");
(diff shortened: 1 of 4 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -222,6 +222,24 @@ class IndexRange
 std::string format_shape(shape_type const & shape);
 std::string format_flat_index(shape_type const & shape, ssize_t offset);
 
+/**
+ * Reject an array that is not one-dimensional. `op` names the calling
+ * operation and `what` the role the array plays in it, both for the message.
+ */
+template <typename A>
+void validate_1d(A const & array, char const * op, char const * what = "array")
```
tigercosmos, line 230: Move `validate_1d` to `detail`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1164,6 +1194,88 @@ class SimpleArrayMixinCalculators
...
+ * counterpart, whose overflow wraps. Both conversions are modular, not magnitude:
+ * -1 converts to 2^N - 1, subtracting there carries the borrow the same way, and a
+ * result that fits the element type converts back unchanged. Only a boundary
+ * crossing wraps. The conversion back is defined since C++20; before that it was
+ * implementation-defined.
+ */
+template <typename A, typename T>
+template <typename Op>
+typename detail::SimpleArrayMixinCalculators<A, T>::value_type
+detail::SimpleArrayMixinCalculators<A, T>::wrapping_op(value_type lhs, value_type rhs, Op op)
```
tigercosmos, line 1208: wrapping the operations.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1164,6 +1194,88 @@ class SimpleArrayMixinCalculators
...
+        return static_cast<value_type>(op(static_cast<unsigned_type>(lhs), static_cast<unsigned_type>(rhs)));
+    }
+    else
+    {
+        return static_cast<value_type>(op(lhs, rhs));
+    }
+}
+
+template <typename A, typename T>
+A detail::SimpleArrayMixinCalculators<A, T>::diff() const
```
tigercosmos, line 1222: Implementation for `diff`.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1164,6 +1194,88 @@ class SimpleArrayMixinCalculators
...
+        {
+            dst[i - 1] = wrapping_sub(src[i * step], src[(i - 1) * step]);
+        }
+
+        return ret;
+    }
+}
+
+template <typename A, typename T>
+A detail::SimpleArrayMixinCalculators<A, T>::cumsum() const
```
tigercosmos, line 1250: Implementation for `cumsum`.

Issue #1234, opened 2026-07-30
Complete the Polygon as a Real Canvas Shape


tigercosmos:
## Background

The pilot 2D canvas already has a single-ring polygon, landed in #1140: `ShapeType::POLYGON`, `World::add_polygon(vertices)` (a flat vertex list forming one closed ring), Python bindings, and the Agent Draw `add_polygon` command. What is still missing is everything that makes the polygon feel like a real drawing-app shape:

- No holes. A polygon is exactly one ring; a region with a hole is not representable.
- Selection is approximate for every shape. `pick_shape` combines outline distance with an inside-axis-aligned-bounding-box shortcut, so clicking empty space near a rotated or concave shape can select it, and there is no true point-in-polygon test.
- No GUI tool. A user cannot draw a polygon by hand; polygons only come from the API.
- Agent read-back is a segment dump. `get_shape` and `describe_state` report a polygon as anonymous segments, so an agent cannot see its ring structure.
- Degenerate polygons are underdiagnosed. There is no polygon branch in the shape-degeneracy checks: zero area and fewer-than-3-distinct-vertices pass silently, and there is no hole-containment check yet.

This MVP closes those gaps. The landed `add_polygon(vertices)` API and its callers keep working unchanged; the exact compatibility promise is under Contracts.

## Goal

Upgrade the landed single-ring polygon to a complete canvas shape:

- Multi-ring polygons: one outer ring plus zero or more hole rings, with full CRUD parity (create, select, move, rotate, delete, undo/redo) shared with the other primitives.
- Correct interior selection for all closed shapes: clicking inside a polygon selects it; clicking inside a hole or in an empty corner of the bounding box does not.
- A GUI tool to draw polygons by hand with a multi-click gesture.
- Agent-legible read-back: a polygon with a hole reads back through Agent Draw as two labeled rings with their vertices, not as an anonymous pile of segments.
- Degenerate polygon input surfaces as named diagnostics instead of being silently accepted.

Definition of done: a user click-draws a concave polygon in the pilot canvas, selects it by clicking its interior, moves it and undoes the move like any primitive; an agent creates a polygon with a hole and reads it back as outer + hole rings; the ICCAD-2013 sample in the Canvas menu loads as real, selectable polygon shapes instead of the raw segment dump it is today.

## Non-Goals

Out of scope for this MVP, planned as follow-up work:

[...]

PR #1348, opened 2026-08-12
Store multi-ring polygons in World and expose them as shape_rings

tychuang1211:
This teaches solvcon::World that a polygon is a region rather than a single closed loop: it can now have one outer ring plus zero or more hole rings. A new RingEntry table, parallel to World's existing segment and curve pads, is referenced from ShapeRecord by ring_offset and ring_count, following the same offset-and-count convention the rest of the class already uses.

add_polygon_rings(rings) is the new primitive; the landed add_polygon(vertices) becomes its one-ring special case, with its exact error message and observable behavior preserved. A new shape_rings(shape_id) accessor reconstructs each ring's role and vertices from the segment pad, and is exposed to Python through a new pybind11 binding alongside add_polygon_rings. translate_shape, rotate_shape, undo/redo, and shape removal needed no changes, since every ring of a shape lives in that shape's existing contiguous segment range.

This is Step 1 of the "Shape MVP: Complete the Polygon as a Real Canvas Shape" plan; interior-aware picking, a GUI polygon tool, agent read-back, and degeneracy diagnostics are later steps and are not part of this PR.

Related to #1234.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ profiling/results/
 thirdparty/install/
 thirdparty/archive/
 tmp/
+graphify-out/
 setup.mk
 CMakeUserPresets.json
 *.egg-info/
--- a/cpp/solvcon/universe/World.hpp
+++ b/cpp/solvcon/universe/World.hpp
@@ -263,6 +263,7 @@ struct ShapeRecord
     size_t curve_count; ///< Number of cubic Beziers this shape occupies.
     size_t ring_offset; ///< First index in the ring registry. Only POLYGON shapes use this.
     size_t ring_count; ///< Number of rings this shape owns. Zero for non-polygon shapes.
+    uint8_t ring_ndim; ///< Dimensionality (2 or 3) of this shape's rings. Zero for non-polygon shapes.
 
     /// OBB corner x's in world coordinates, ordered TL, TR, BR, BL.
     bbox_array_type obb_x;
@@ -323,6 +324,7 @@ public:
     using point_pad_type = PointPad<T>;
     using segment_pad_type = SegmentPad<T>;
     using curve_pad_type = CurvePad<T>;
+    using ring_list_type = RingList<T>;
     using bbox_type = BoundBox3d<T>;
     using rtree_type = RTree<ShapeEntry<T>, bbox_type>;
 
@@ -461,9 +463,10 @@ public:
      * Add a polygon with one outer ring and zero or more hole rings. The
      * first ring in `rings` is the outer boundary; the rest are holes.
      * Each ring needs at least 3 vertices. Rings are stored as one
-     * contiguous segment run, outer ring first.
+     * contiguous segment run, outer ring first. `rings` may be 2- or
+     * 3-dimensional; every ring in one call shares `rings`'s ndim.
      */
-    int32_t add_polygon_rings(std::vector<std::vector<std::array<T, 2>>> const & rings);
+    int32_t add_polygon_rings(ring_list_type const & rings);
 
     /**
      * Add a closed chain of straight segments through the given vertices, with
@@ -507,13 +510,17 @@ public:
         return m_curves->get(rec.curve_offset + i);
     }
 
-    /// One ring of a live shape (0-based within the shape), with vertices
-    /// reconstructed from the segment pad in input order (no repeated
-    /// closing vertex) and its role (outer for ring 0, hole otherwise).
+    /**
+     * One ring of a live shape (0-based within the shape), with vertices
+     * reconstructed from the segment pad in input order (no repeated
+     * closing vertex) and its role (outer for ring 0, hole otherwise).
+     * Each vertex has 2 elements for a 2-dimensional ring, 3 for a
+     * 3-dimensional one.
+     */
     struct RingInfo
     {
         RingRole role;
-        std::vector<std::array<T, 2>> vertices;
+        std::vector<std::vector<T>> vertices;
     }; /* end struct RingInfo */
 
     std::vector<RingInfo> shape_rings(int32_t shape_id) const;
@@ -726,7 +733,8 @@ private:
                            size_t curve_offset = 0,
                            size_t curve_count = 0,
                            size_t ring_offset = 0,
-                           size_t ring_count = 0);
+                           size_t ring_count = 0,
+                           uint8_t ring_ndim = 0);
 
     /**
      * Check if shape_id is valid and not DEAD.
@@ -824,10 +832,11 @@ int32_t World<T>::register_shape(ShapeType type,
                                  size_t curve_offset,
                                  size_t curve_count,
                                  size_t ring_offset,
-                                 size_t ring_count)
+                                 size_t ring_count,
+                                 uint8_t ring_ndim)
 {
     auto shape_id = static_cast<int32_t>(m_shape_registry.size());
-    m_shape_registry.push_back(ShapeRecord{type, segment_offset, segment_count, curve_offset, curve_count, ring_offset, ring_count, {}, {}}); // NOLINT(modernize-use-designated-initializers)
+    m_shape_registry.push_back(ShapeRecord{type, segment_offset, segment_count, curve_offset, curve_count, ring_offset, ring_count, ring_ndim, {}, {}}); // NOLINT(modernize-use-designated-initializers)
     ++m_nshape;
     bbox_type const bb = compute_shape_bbox(m_shape_registry[shape_id]);
 
@@ -973,15 +982,15 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
 }
 
 template <typename T>
-int32_t World<T>::add_polygon_rings(std::vector<std::vector<std::array<T, 2>>> const & rings)
+int32_t World<T>::add_polygon_rings(ring_list_type const & rings)
 {
-    if (rings.empty())
+    if (rings.nring() == 0)
     {
         throw std::invalid_argument("World: add_polygon_rings needs at least one ring");
     }
-    for (auto const & ring : rings)
+    for (size_t r = 0; r < rings.nring(); ++r)
     {
-        if (ring.size() < 3)
+        if (rings.ring_size(r) < 3)
         {
             throw std::invalid_argument("World: add_polygon_rings needs at least 3 vertices per ring");
         }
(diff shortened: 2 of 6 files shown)


docs/superpowers/specs/2026-08-12-polygon-ring-storage-design.md
```diff
@@ -0,0 +1,191 @@
+# Multi-Ring Polygon Storage in the World Data Model
```
tigercosmos: Why do you put these MD files in the PR?

tychuang1211: Those were the design spec and implementation plan generated during development, internal workflow artifacts that were not meant to ship in the repo. Dropped both files in 83fce4ed.

cpp/solvcon/universe/World.hpp
```diff
@@ -219,6 +219,32 @@ class WorldState : public SerializableItem
...
+
+/**
+ * One ring (outer boundary or hole) of a polygon shape, as a range into
+ * the shape's own segment run.
+ *
+ * @ingroup group_geometry
+ */
+struct RingEntry
+{
+    size_t segment_offset; ///< Absolute first index of this ring's run in SegmentPad.
```
tychuang1211, line 243: Note this is absolute into the shared SegmentPad, not relative to the shape's own `segment_offset` in `ShapeRecord` right below.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +972,73 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
...
+        for (size_t i = 0; i < ring.size(); ++i)
+        {
+            size_t const j = (i + 1) % ring.size();
+            m_segments->append(
+                point_type(ring[i][0], ring[i][1], 0),
+                point_type(ring[j][0], ring[j][1], 0));
+        }
+        m_ring_registry.push_back(RingEntry{ring_seg_offset, ring.size(), r == 0 ? RingRole::OUTER : RingRole::HOLE});
+    }
+    return register_shape(ShapeType::POLYGON, segment_offset, m_segments->size() - segment_offset,
```
tychuang1211, line 1040: Two offsets at different granularities here: `segment_offset` is the shape's whole contiguous run across every ring (what `ShapeRecord` stores), while `ring_seg_offset` inside the loop above is the sub-range for just one ring (what each `RingEntry` stores). All rings of a shape land in one contiguous SegmentPad run; only the ring registry knows where each ring starts within it.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +972,73 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
...
+        auto const & ring = rings[r];
+        size_t const ring_seg_offset = m_segments->size();
+        for (size_t i = 0; i < ring.size(); ++i)
+        {
+            size_t const j = (i + 1) % ring.size();
+            m_segments->append(
+                point_type(ring[i][0], ring[i][1], 0),
+                point_type(ring[j][0], ring[j][1], 0));
+        }
+        m_ring_registry.push_back(RingEntry{ring_seg_offset, ring.size(), r == 0 ? RingRole::OUTER : RingRole::HOLE});
```
tychuang1211: Role is purely positional: ring 0 is always outer, everything after is a hole. No winding-direction check, so passing rings in the wrong order produces silently wrong geometry rather than an error. A winding check could be added in a later step (e.g. interior-aware picking) if that turns out to matter.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +972,73 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
...
         throw std::invalid_argument("World: add_polygon needs at least 3 vertices");
     }
-    size_t const offset = m_segments->size();
-    for (size_t i = 0; i < vertices.size(); ++i)
-    {
-        size_t const j = (i + 1) % vertices.size();
-        m_segments->append(
-            point_type(vertices[i][0], vertices[i][1], 0),
-            point_type(vertices[j][0], vertices[j][1], 0));
+    return add_polygon_rings({vertices});
```
tychuang1211: The `vertices.size() < 3` check above is duplicated on purpose: `add_polygon_rings` has its own check with a different message ("...at least 3 vertices per ring"), and this keeps `add_polygon`'s original error message unchanged for existing callers. Covered by `test_polygon_needs_three_vertices`.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +972,73 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
...
+    for (size_t r = 0; r < rec.ring_count; ++r)
+    {
+        RingEntry const & ring = m_ring_registry[rec.ring_offset + r];
+        RingInfo info;
+        info.role = ring.role;
+        info.vertices.reserve(ring.segment_count);
+        for (size_t i = 0; i < ring.segment_count; ++i)
+        {
+            segment_type const seg = m_segments->get(ring.segment_offset + i);
+            info.vertices.push_back({seg.p0().x(), seg.p0().y()});
```
tychuang1211: Only `p0` of each segment is collected here, not `p1`. That works because a ring's segments already form a closed loop (the last segment's `p1` wraps back to the first segment's `p0`), so walking `p0` in order recovers exactly the original vertex list with no duplicated closing vertex, mirroring how `add_polygon_rings` built the segments in the first place.

cpp/solvcon/universe/pymod/wrap_World.cpp
```diff
@@ -145,6 +145,21 @@ WrapWorld<T> & WrapWorld<T>::wrap_management()
...
             py::arg("shape_id"))
+        .def(
+            "shape_rings",
+            [](wrapped_type const & self, int32_t shape_id)
+            {
+                py::list result;
+                for (auto const & ring : self.shape_rings(shape_id))
+                {
+                    py::dict d;
+                    d["role"] = (ring.role == RingRole::OUTER) ? "outer" : "hole";
```
tychuang1211, line 164: `RingRole` is exposed as a plain Python string here rather than as a bound `py::enum_`, since it has no other Python-facing use yet. Worth a `py::enum_<RingRole>` if a second consumer shows up.

tests/test_universe.py
```diff
@@ -360,6 +360,85 @@ def test_polygon_needs_three_vertices(self):
...
+        sid = self.w.add_polygon_rings([outer, hole])
+        # A 90-degree rotation about the origin maps (x, y) -> (-y, x).
+        self.w.rotate_shape(sid, math.pi / 2, 0, 0)
+        rings = self.w.shape_rings(sid)
+        for ring, expected in ((rings[0], outer), (rings[1], hole)):
+            for (x, y), (ex, ey) in zip(ring["vertices"], expected):
+                self.assertAlmostEqual(x, -ey)
+                self.assertAlmostEqual(y, ex)
+
+    def test_multi_ring_polygon_undo_redo(self):
```
tychuang1211, line 446: This confirms rings need no special-casing in undo/redo: `kill_shape`/`revive_shape` only flip the shape's DEAD/live type, never touch `m_ring_registry` (same pattern already used for segments and curves), so ring data survives a dead shape and comes back untouched on redo.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +972,73 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
     return register_shape(ShapeType::POLYLINE, offset, vertices.size() - 1);
 }
 
+template <typename T>
+int32_t World<T>::add_polygon_rings(std::vector<std::vector<std::array<T, 2>>> const & rings)
```
yungyuc: The input argument type does not clearly state what it does.   It is not clear what does `std::array<T, 2>` mean.  It is also inefficient to use a vector of vectors to pass data because it incurs a lot of memory allocations.

```suggestion
int32_t World<T>::add_polygon_rings(RingList const & rings)
```

The rings should support both 2 and 3 dimensional points.

yungyuc (changes requested):
- [x] Change `World::add_polygon_rings()` to a new `RingList` class that does not incur multiple memory allocations.
- [x] `RingList` and `add_polygon_rings()` support both 2 and 3 dimensional points (node lists).
- [x] The rings in the new class `RingList` should follow right-hand rule.  That is, for positive (area) polygons described by a ring, the nodes order counter-clock-wise.  For negative (area) polygons (holes), the nodes order clock-wise.
  - [x] Add polygon helpers and tests for the area calculation.  (If you want to use a follow-up PR for this, say so in PR global comment).

cpp/solvcon/universe/World.hpp
tychuang1211: Will implement RingList and add_polygon_rings(RingList const &) with 2D/3D support in this PR.

World's shape layer is 2D-only for every shape today, not just polygons, so bbox/OBB/picking will keep operating on x/y for now. Making the whole shape layer 3D-aware is separate follow-up work.

cpp/solvcon/universe/coord.hpp
```diff
@@ -739,6 +731,137 @@ class PointPad
...
+        }
+        return m_points->get(m_ring_offset[r] + i);
+    }
+
+    /// Result of signed_area(): the area itself, and a scale to compare it
+    /// against when judging whether it is degenerate (near zero relative
+    /// to the ring's own coordinate magnitude, not to an absolute bound).
+    struct SignedArea
+    {
+        T value; ///< Shoelace signed area: positive CCW, negative CW.
```
yungyuc, line 818: Good note.

cpp/solvcon/universe/World.hpp
```diff
@@ -919,22 +984,110 @@ int32_t World<T>::add_polyline(std::vector<std::array<T, 2>> const & vertices)
...
+template <typename T>
+int32_t World<T>::add_polygon_rings(ring_list_type const & rings)
+{
+    if (rings.nring() == 0)
+    {
+        throw std::invalid_argument("World: add_polygon_rings needs at least one ring");
+    }
+    for (size_t r = 0; r < rings.nring(); ++r)
+    {
+        if (rings.ring_size(r) < 3)
```
yungyuc, line 996: In the future, we may relax the constraint to be 2 to allow zero-area ring (two line segments).  But we keep 3 for now.

.gitignore
```diff
@@ -11,6 +11,7 @@ profiling/results/
 thirdparty/install/
 thirdparty/archive/
 tmp/
+graphify-out/
```
yungyuc, line 14: It should use alphabetical order.  Please correct it in a future PR.

yungyuc (approved):
LGTM.  Sorry for taking longer time to review.

PR #1354, opened 2026-08-14
Route public matmul through selected kernels

ThreeMonth03:
## Motivation

Public `matmul` still used older implementations even though layout-aware kernel selection now lives in `MatmulPlan` and `MatmulExecutor`. Keeping both paths duplicates APIs and prevents the public operation from using the optimized routes.

## Approach

```text
lhs.matmul(rhs), lhs @ rhs, lhs @= rhs
    -> MatmulPlan
    -> generic | BLAS | packed BLAS | Winograd
    -> MatmulExecutor
```

Route the public entry points through the executor and remove the superseded alternate entry points from C++, Python bindings, tests, and profiling.

## Validation

`make lint`, 1,660 Python tests, 244 C++ tests, and `make pyprof` pass.

Related to #1172.


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1269,18 +1258,10 @@ detail::SimpleArrayMixinCalculators<A, T>::median_freq(small_vector<value_type>
...
-    auto const * athis = static_cast<A const *>(this);
-    SimpleArrayMatmulHelper<A, T> helper(*athis, other);
-    return helper.matmul();
-}
-
-template <typename A, typename T>
-A SimpleArrayMixinCalculators<A, T>::matmul_planned(A const & other) const
 {
     auto const * athis = static_cast<A const *>(this);
     MatmulPlan plan = MatmulPlan::make(*athis, other);
```
ThreeMonth03, line 1267: This is the shared public call path:

```text
matmul / @ / imatmul / @=
    -> MatmulPlan (validate shapes, broadcast batches, build strides)
    -> MatmulExecutor (generic | BLAS | packed BLAS | Winograd)
```

Every public entry point now uses the same layout-aware selection.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -427,20 +427,6 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
 
         (*this)
             .def("matmul", &wrapped_type::matmul)
-            .def("matmul_planned", &wrapped_type::matmul_planned)
```
ThreeMonth03, line 430: The Python surface now exposes one policy-free route:

```text
matmul / @   -> wrapped_type::matmul
imatmul / @= -> wrapped_type::imatmul
```

The removed `planned`, `blas`, and manual-tile methods selected implementation details that `MatmulExecutor` now chooses internally.

ThreeMonth03 (commented):

tests/test_blas_lapack.py
```diff
@@ -26,24 +26,25 @@
 }
 
 
-class MatmulBlasTC(unittest.TestCase):
-    """SimpleArray.matmul_blas() (CBLAS ?gemm) must match NumPy."""
+class MatmulBackendTC(unittest.TestCase):
```
yungyuc, line 31: It's good to remove duplicate tests.  But how does the new tests ensure coverage of all backend we support?  Please annotate at the key lines.

yungyuc (approved):
LGTM.  Could you please elaborate how to ensure the test coverage for all supported backends?

tests/test_blas_lapack.py
ThreeMonth03: The CI matrix covers the supported BLAS backends with the same public test: OpenBLAS on Ubuntu, Accelerate on macOS arm64, and MKL through the Windows pilot pytest suite. The four dtypes exercise s/d/c/zgemm. I will annotate those facts at the dtype table and the public `matmul()` call.

There is a separate dispatch-coverage gap. The current public matmul tests exercise the generic and BLAS routes, but they do not force each route independently. The Winograd dispatch is not covered. Before adding more dispatch-performance visualization, I think we should refactor the test and profiling seam so a caller can explicitly select a dispatch route. I would do that in a follow-up instead of expanding this API-removal PR.

tests/test_blas_lapack.py
```diff
@@ -17,6 +17,8 @@
 import solvcon as sc
 
 
+# CI runs this table with OpenBLAS on Ubuntu, Accelerate on macOS arm64,
```
ThreeMonth03, line 20: `MatmulBackendTC.test_matches_numpy` verifies the direct BLAS dispatch. Its contiguous `17x18 @ 18x19` operands exceed the direct CBLAS cutoff. The CI matrix runs its four dtype cases against OpenBLAS, Accelerate, and MKL.

tests/test_matrix.py
```diff
@@ -152,11 +128,7 @@ def test_square(self):
         # Expected result: [[19, 22], [43, 50]]
         expected = np.array([[19.0, 22.0], [43.0, 50.0]], dtype=self.dtype)
 
-        # Test matrix multiplication
-        result = self.assert_matmul(a, b, expected)
-        self.assert_matmul_fast(a, b, expected, result)
-        self.assert_matmul_blas(a, b, expected, result)
-        self.assert_matmul_planned(a, b, expected)
+        self.assert_matmul(a, b, expected)
```
ThreeMonth03, line 131: `MatmulTestBase.test_square` verifies the Generic dispatch. Its `2x2 @ 2x2` operands are below the direct CBLAS cutoff of `8`, so public `matmul()` selects Generic.

PR #1355, opened 2026-08-14
Discard the Winograd fallback with `if constexpr`

### LINKED ISSUE #1353: Nightly devbuild: Windows Debug and sanitizer jobs fail on C4702 in matmul.hpp
## Summary

The scheduled `devbuild` workflow fails on the two Windows Debug jobs. MSVC reports C4702 (unreachable code) at the end of `MatmulExecutor::execute_winograd()` in `cpp/solvcon/buffer/matmul.hpp`, and `/WX` turns it into an error, so the build stops.

## Failing runs

| Nightly | Head | Result |
| --- | --- | --- |
| 2026-08-11 | 95a68122 | success |
| 2026-08-12 | ca78ff71 | [failure](https://github.com/solvcon/solvcon/actions/runs/31627455665) |
| 2026-08-13 | d662a740 | [failure](https://github.com/solvcon/solvcon/actions/runs/31730628922) |

Both failures break the same two jobs: `build_windows-2022_Debug` and `sanitizer_windows-2022`. Every other job in the run passes, including `build_windows-2022_Release`. Nothing in the workflow changed between the passing and failing nightlies.

## Error

The compile fails on the two translation units that instantiate `MatmulExecutor` for a BLAS value type, `wrap_SimpleArray_float.cpp` and `wrap_SimpleArray_complex.cpp` ([job log](https://github.com/solvcon/solvcon/actions/runs/31730628922/job/94550411527)):

```
FAILED: [code=2] cpp/solvcon/CMakeFiles/solvcon_primary.dir/buffer/pymod/wrap_SimpleArray_complex.cpp.obj
cl.exe ... /Od /Ob0 -std:c++latest /W4 /WX ... -c cpp\solvcon\buffer\pymod\wrap_SimpleArray_complex.cpp
  matmul.hpp(1014) : error C2220: the following warning is treated as an error
  matmul.hpp(1014) : warning C4702: unreachable code
ninja: build stopped: subcommand failed.
```

## Cause

`execute_winograd()` guards its BLAS-backed kernel with an `#if` on the backend macros and leaves the `throw` after it as the fallback:

```cpp
#if (defined(__APPLE__) && defined(__arm64__)) || defined(SC_HAS_CBLAS)
    if constexpr (can_matmul_blas_v<value_type>)
    {
        ...
        return;
    }
#endif
    throw std::logic_error("MatmulExecutor::execute_winograd(): unavailable Winograd kernel");
```

When the build has a BLAS backend and the value type is one BLAS takes, the preprocessor keeps both parts and the `if constexpr` branch always returns, so the `throw` on line 1014 is dead code. `int` and `uint` instantiate the fallback and compile fine; `float` and `complex` do not.

The function arrived with 78872eb2 in #1317, which merged into the 2026-08-12 nightly. That matches the first failing run.

## Expected

The nightly Windows Debug and sanitizer jobs build clean under `/W4 /WX`.

## Status

#1355 replaced the `#if` plus `if constexpr` pair with a single `if constexpr (use_matmul_blas_v<value_type>)` and moved the `throw` into the `else`, so the untaken branch is discarded instead of standing as dead code. It merged as 14f2b095 on 2026-08-14. Runs after that merge are green, but no scheduled nightly has run since, so this issue stays open until one does.

yungyuc:
For issue #1353

The issue is that the Windows compiler encountered unreachable code and issued error code [C4702](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4702) in `MatmulExecutor::execute_winograd()`.

Use `use_matmul_blas_v` and put the `throw` in the `else` to fix it.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -994,8 +994,7 @@ void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type co
 template <typename Array>
 void MatmulExecutor<Array>::execute_winograd()
 {
-#if (defined(__APPLE__) && defined(__arm64__)) || defined(SC_HAS_CBLAS)
-    if constexpr (can_matmul_blas_v<value_type>)
+    if constexpr (use_matmul_blas_v<value_type>)
     {
         BlasOutputView<value_type> const output{
             .m_data = m_output_data,
@@ -1008,10 +1007,11 @@ void MatmulExecutor<Array>::execute_winograd()
             require_matrix_view(lhs_matrix_view(m_lhs_data)),
             require_matrix_view(rhs_matrix_view(m_rhs_data)),
             output);
-        return;
     }
-#endif
-    throw std::logic_error("MatmulExecutor::execute_winograd(): unavailable Winograd kernel");
+    else
+    {
+        throw std::logic_error("MatmulExecutor::execute_winograd(): unavailable Winograd kernel");
+    }
 }
 
 template <typename Array>


cpp/solvcon/buffer/matmul.hpp
```diff
@@ -994,8 +994,7 @@ void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type co
 template <typename Array>
 void MatmulExecutor<Array>::execute_winograd()
 {
-#if (defined(__APPLE__) && defined(__arm64__)) || defined(SC_HAS_CBLAS)
-    if constexpr (can_matmul_blas_v<value_type>)
+    if constexpr (use_matmul_blas_v<value_type>)
```
yungyuc, line 997: `use_matmul_blas_v` should be sufficient to fence for BLAS availability.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -1008,10 +1007,11 @@ void MatmulExecutor<Array>::execute_winograd()
             require_matrix_view(lhs_matrix_view(m_lhs_data)),
             require_matrix_view(rhs_matrix_view(m_rhs_data)),
             output);
-        return;
     }
-#endif
-    throw std::logic_error("MatmulExecutor::execute_winograd(): unavailable Winograd kernel");
+    else
+    {
+        throw std::logic_error("MatmulExecutor::execute_winograd(): unavailable Winograd kernel");
```
yungyuc, line 1013: Throw in `else`.

PR #1358, opened 2026-08-14
Cache GEMM scratch data pointer

ThreeMonth03:
## Motivation

The scheduled `clang-tidy` job reports `bugprone-unchecked-optional-access` for GEMM scratch storage. The optional is prepared before batched GEMM execution, but that invariant is not visible at each access.

## Approach

Cache the scratch data pointer when storage is prepared and use it while packing incompatible operands. This keeps setup outside the batch loop and removes repeated optional checks.

## Validation

`make lint`, 244 C++ tests, and 106 matrix tests pass.

Related to #1283.

### DIFF
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -386,6 +386,7 @@ private:
     std::optional<Array> m_packed_lhs;
     std::optional<Array> m_packed_rhs;
     std::optional<Array> m_gemm_scratch;
+    value_type * m_gemm_scratch_data = nullptr;
     size_t m_lhs_scratch_elements = 0;
     value_type const * m_cached_lhs_source = nullptr;
     value_type const * m_cached_rhs_source = nullptr;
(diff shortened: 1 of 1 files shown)


ThreeMonth03 (commented):

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -386,6 +386,7 @@ class MatmulExecutor
     std::optional<Array> m_packed_lhs;
     std::optional<Array> m_packed_rhs;
     std::optional<Array> m_gemm_scratch;
+    value_type * m_gemm_scratch_data = nullptr;
```
yungyuc, line 389: This line `value_type * m_gemm_scratch_data = nullptr;` duplicates with the previous `std::optional<Array> m_gemm_scratch;` and it does not resolve the semantic of the dangling pointer.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -1041,7 +1043,7 @@ void MatmulExecutor<Array>::execute_gemm_blas(value_type * output, value_type co
     {
         rhs_view = pack_matrix_to_scratch(
             rhs_data,
-            m_gemm_scratch.value().data() + m_lhs_scratch_elements,
```
yungyuc, line 1044: Please elaborate why the validity of the optional `m_gemm_scratch` is not checked.

yungyuc (changes requested):
- [ ] The implementation does not resolve the semantic of dangling pointer/object.

PR #1360, opened 2026-08-15
Qualify detail:: calls instead of using namespace detail in simd.hpp

### LINKED ISSUE #631: Do not `using namespace` even in function local

rockleona:
`cpp/solvcon/simd/simd.hpp` defined add, sub, mul, and div, each opening a function-local using namespace detail; to reach `detect_simd()` and `SIMD_NEON` from the `solvcon::simd::detail namespace`.

A local using-directive hides which symbols are pulled in from detail, making it harder for a maintainer to tell at a glance which names are implementation details versus local declarations, which is why each occurrence carried a FIXME comment.

This PR removes the four using namespace detail; directives and qualifies the calls explicitly as `detail::detect_simd()` and `detail::SIMD_NEON`, matching the pattern check_between in the same file already used.

Related to #631.

### DIFF
--- a/cpp/solvcon/simd/simd.hpp
+++ b/cpp/solvcon/simd/simd.hpp
@@ -34,10 +34,9 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
 template <typename T>
 void add(T * dest, T const * dest_end, T const * src1, T const * src2)
 {
-    using namespace detail; // FIXME: NOLINT(google-build-using-namespace)
-    switch (detect_simd())
+    switch (detail::detect_simd())
     {
-    case SIMD_NEON:
+    case detail::SIMD_NEON:
         return neon::add<T>(dest, dest_end, src1, src2);
         break;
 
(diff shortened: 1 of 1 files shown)


cpp/solvcon/simd/simd.hpp
```diff
@@ -34,10 +34,9 @@ const T * check_between(T const * start, T const * end, T const & min_val, T con
 template <typename T>
 void add(T * dest, T const * dest_end, T const * src1, T const * src2)
 {
-    using namespace detail; // FIXME: NOLINT(google-build-using-namespace)
-    switch (detect_simd())
+    switch (detail::detect_simd())
     {
-    case SIMD_NEON:
```
rockleona, line 40: `using namespace detail;` had been removed, the functions and variables under `detail` will be point out inline.

cpp/solvcon/simd/simd.hpp
```diff
@@ -100,3 +96,5 @@ T max(T const * start, T const * end)
 } /* end namespace simd */
 
 } /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
```
yungyuc, line 100: Good catch!

yungyuc (approved):
LGTM.  Thanks.

PR #1362, opened 2026-08-15
Implement chained function call (fluent API) for SimpleArray in-place arithmetic operation

KHLee529:
This PR enables chaining function call (fluent API) for SimpleArray in-place arithmetic operations in Python. For example `sarr1.iadd(sarr2).isub(sarr3)` is now valid function call syntax in Python.

Related to https://github.com/solvcon/solvcon/issues/514.

### DIFF
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -454,7 +454,8 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 [](wrapped_type const & self, wrapped_type const & other)
                 { return self.matmul(other); },
                 py::arg("other"))
-            // TODO: In-place operation should return reference to self to support function chaining
+            // In-place operation return reference to self by std::move and casting self reference
+            // to keep python object ID the same
             /*
              * Regular in-place methods (iadd, imul, etc.) are procedural calls and do
              * NOT need to return self. However, special __i*__ methods (__iadd__,
@@ -466,35 +467,72 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def(
                 "iadd",
                 [](wrapped_type & self, wrapped_type const & other)
-                { self.iadd(other); })
+                { return py::cast(std::move(self.iadd(other))); })
             .def(
                 "iadd",
                 [](wrapped_type & self, value_type scalar)
-                { self.iadd(scalar); })
+                { return py::cast(std::move(self.iadd(scalar))); })
             .def(
                 "isub",
                 [](wrapped_type & self, wrapped_type const & other)
-                { self.isub(other); })
+                {
+                    if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+                    {
+                        return py::cast(std::move(self.isub(other)));
+                    }
+                    else
+                    {
+                        throw std::runtime_error(
+                            "SimpleArray<bool>::isub(): boolean value doesn't support this operation");
+                    }
+                })
             .def(
                 "isub",
                 [](wrapped_type & self, value_type scalar)
-                { self.isub(scalar); })
+                {
+                    if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+                    {
+                        return py::cast(std::move(self.isub(scalar)));
+                    }
+                    else
+                    {
+                        throw std::runtime_error(
+                            "SimpleArray<bool>::isub(): boolean value doesn't support this operation");
+                    }
+                })
             .def(
                 "imul",
                 [](wrapped_type & self, wrapped_type const & other)
-                { self.imul(other); })
+                { return py::cast(std::move(self.imul(other))); })
             .def(
                 "imul",
                 [](wrapped_type & self, value_type scalar)
-                { self.imul(scalar); })
+                { return py::cast(std::move(self.imul(scalar))); })
             .def("idiv", [](wrapped_type & self, wrapped_type const & other)
-                 { self.idiv(other); })
+                 {
+                    if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+                    {
+                        return py::cast(std::move(self.idiv(other)));
+                    }
+                    else
+                    {
+                        throw std::runtime_error(
+                            "SimpleArray<bool>::idiv(): boolean value doesn't support this operation");
+                    } })
             .def(
-                "idiv",
-                [](wrapped_type & self, value_type scalar)
-                { self.idiv(scalar); })
+                "idiv", [](wrapped_type & self, value_type scalar)
+                {
+                    if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
+                    {
+                        return py::cast(std::move(self.idiv(scalar)));
+                    }
+                    else
+                    {
+                        throw std::runtime_error(
+                            "SimpleArray<bool>::idiv(): boolean value doesn't support this operation");
+                    } })
             .def("imatmul", [](wrapped_type & self, wrapped_type const & other)
-                 { self.imatmul(other); })
+                 { return py::cast(std::move(self.imatmul(other))); })
             .def("__imatmul__", [](wrapped_type & self, wrapped_type const & other)
                  {
                      self.imatmul(other);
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -454,45 +454,78 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
             .def(
                 "iadd",
                 [](wrapped_type & self, wrapped_type const & other)
-                { self.iadd(other); })
+                { return py::cast(std::move(self.iadd(other))); })
```
KHLee529, line 443: Return the `self` object in python wrapper

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -454,45 +454,78 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
                 "iadd",
                 [](wrapped_type & self, value_type scalar)
-                { self.iadd(scalar); })
+                { return py::cast(std::move(self.iadd(scalar))); })
             .def(
                 "isub",
                 [](wrapped_type & self, wrapped_type const & other)
-                { self.isub(other); })
+                {
+                    if constexpr (!std::is_same_v<bool, std::remove_const_t<value_type>>)
```
KHLee529, line 452: Skip `isub` and `idiv` for `bool` due to `SimpleArray<bool>::isub` and `SimpleArray<bool>::idiv` [returns no value](https://github.com/solvcon/solvcon/blob/7118fd6996e4af5e3e57ff46e97fa1b61b1a28e7/cpp/solvcon/buffer/SimpleArray.hpp#L930) and cause [compilation error in Windows](https://github.com/solvcon/solvcon/actions/runs/31867319422/job/94970652684).

```bash
D:\a\solvcon\solvcon\cpp\solvcon\buffer\pymod\wrap_SimpleArray.hpp(469) : error C2220: the following warning is treated as an error
D:\a\solvcon\solvcon\cpp\solvcon\buffer\pymod\wrap_SimpleArray.hpp(469) : warning C4702: unreachable code
```

---
 
@tigercosmos I wonder if you have better idea to address this error.

tests/test_buffer.py
```diff
@@ -2939,6 +2939,62 @@ def test_binary_arithmetic_shape_mismatch(self):
                         getattr(lhs, operation)(rhs)
                     np.testing.assert_array_equal(expected, lhs.ndarray)
 
+    def test_inplace_arithmetic_returns_self(self):
```
KHLee529, line 2942: Unit test checking if the in-place arithmetic function return caller object in python.

tests/test_buffer.py
```diff
@@ -2939,6 +2939,62 @@ def test_binary_arithmetic_shape_mismatch(self):
...
+        sarr1 = solvcon.SimpleArrayFloat64((3,), value=4)
+        self.assertIs(sarr1.isub(2.0), sarr1)
+
+        sarr1 = solvcon.SimpleArrayFloat64((3,), value=4)
+        self.assertIs(sarr1.imul(2.0), sarr1)
+
+        sarr1 = solvcon.SimpleArrayFloat64((3,), value=4)
+        self.assertIs(sarr1.idiv(2.0), sarr1)
+
+    def test_inplace_arithmetic_fluent_chaining(self):
```
KHLee529, line 2973: Unit test checking calculation result correctness

tests/test_buffer_simd.py
```diff
@@ -73,4 +73,55 @@ def test_int64_mul_falls_back_to_generic(self):
         for i, want in enumerate(expected):
             self.assertEqual(out[i], want)
 
+
+class SimdTransformInplaceBinaryTC(unittest.TestCase):
```
KHLee529, line 77: Unit test counterpart for in-place SIMD arithmetic operation.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc: Please elaborate why you use move and cast instead of a [non-copying return value policy](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies).

yungyuc (changes requested):
- [x] Elaborate the reason not using return value policy.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
KHLee529: @yungyuc This is implemented following the coding convention like [`take_along_axis`](https://github.com/KHLee529/solvcon/blob/e8b5c8f432a6c81849ce2f753e5f407f21241508/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp#L639).

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
tigercosmos: I think skipping it is okay for now. Please add a inline comment in the code to explain the intention.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -440,35 +440,68 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
+                    {
+                        return py::cast(std::move(self.isub(other)));
+                    }
+                    else
+                    {
+                        self.isub(other);
+                    }
```
tigercosmos, line 460: The code will reach here if it is `SimpleArray<bool>::isub`, which results an exception in `isub`. I think it's more straightforward to throw here instead of going into `isub`.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
KHLee529: After trying using `return_value_policy::move` and `return_value_policy::take_ownership`, I found the python object will not be the same if I don't use `py::cast(std::move(<result>))`.  This would cause failure on the newly added unit-test testcases.

``` python
self = <test_buffer.SimpleArrayCalculatorsTC testMethod=test_inplace_arithmetic_returns_self>

    def test_inplace_arithmetic_returns_self(self):
        # In-place methods must return the same Python object as the
        # receiver (not a copy) so that calls can be chained fluently.
        sarr1 = solvcon.SimpleArrayFloat64((3,), value=4)
        sarr2 = solvcon.SimpleArrayFloat64((3,), value=2)
>       self.assertIs(sarr1.iadd(sarr2), sarr1)
E       AssertionError: <_solvcon.SimpleArrayFloat64 object at 0x11f7c36f0> is not <_solvcon.SimpleArrayFloat64 object at 0x11f3c1370>

tests/test_buffer.py:2947: AssertionError
__________________________________________________ SimdTransformInplaceBinaryTC.test_inplace_simd_fluent_chaining ___________________________________________________
```

yungyuc: > After trying using `return_value_policy::move` and `return_value_policy::take_ownership`, I found the python object will not be the same if I don't use `py::cast(std::move(<result>))`. This would cause failure on the newly added unit-test testcases.

Try `return_value_policy::reference_internal` and return `self`:

```cpp
                {
                    self.iadd(other);
                    return self;
                })
```

In this way the code matches your intention: chaining the operations.

yungyuc (changes requested):
- [x] Use `return_value_policy::reference_internal`.  If it does not work, clarify why not.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -436,25 +426,10 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-                py::arg("kernel") = py::none())
-            .def(
-                "__matmul__",
-                [](wrapped_type const & self, wrapped_type const & other)
-                { return self.matmul(other); },
-                py::arg("other"))
-            // TODO: In-place operation should return reference to self to support function chaining
+            .def("matmul", &wrapped_type::matmul)
+            .def("__matmul__", &wrapped_type::matmul)
+            // In-place operation return reference to self by std::move and casting self reference
```
KHLee529, line 457: Comment indicating the reason using `py::cast(std::move(self))`

yungyuc (approved):
LGTM

PR #1364, opened 2026-08-15
Implement geometry helper functions for segment type

### LINKED ISSUE #1294: Give the segment types their plane geometry
The oblique-shock reflection work (issue #1269, PR #1289) left a TODO note at [`_analytic.py:L200-L203`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L200-L203) for what the segment types do not compute:

```python
# TODO: the three helpers below are the plane geometry of a segment, not
# of this problem, and every user of Segment3d has to write them again.
# Slope, direction, and normal belong on the class, and the offset wants
# to stay vectorised: a point-to-line distance taking arrays.
```

The helpers are `_slope`, `_offset`, and `_abscissa`, at [`_analytic.py:L205-L220`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L205-L220). None of them knows anything about a shock. They exist because the segment carries no geometry: the whole Python surface of `Segment3d` is `mirror`, `p0`, `p1`, and the six coordinates. `Point3d` is thinner, `fill`, `mirror`, `x`, `y`, and `z`. `calc_length()` and `calc_length2()` are written at [`coord.hpp:L169-L170`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/coord.hpp#L169-L170) but never bound, and only the compound assignment operators reach Python, so `p1 - p0` raises `TypeError` and a direction vector costs four scalar reads.

Proposed on `Segment3d`:

- `calc_length()` and `calc_length2()`, along with the `Point3d` pair that is already written and waiting for a binding.
- `direction()`, the unit vector from p0 to p1.
- `normal()`, the unit normal. A segment in space has no unique normal, so either define it for the plane case (the direction turned a quarter turn, z left at zero) and raise on a segment with a z extent, or take the reference axis as an argument.
- `offset(point)`, the signed perpendicular distance, positive on a named side.
- `point_at(axis, value)`, the point on the line where the named axis reaches `value`, returned as a `Point3d`. It names the axis the way `mirror()` already does, and it raises when the direction has no component along that axis, since the line then reaches the value nowhere or everywhere. `_abscissa` is the y case of it, read back through `.x`.

Proposed on `SegmentPad`, the vectorised form the note asks for: an offset that takes numpy arrays of x and y and returns an array of distances from one arm. The pad already publishes `x0`, `y0`, `x1`, and `y1` as `SimpleArray` at [`wrap_shape1d.cpp:L432-L448`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape1d.cpp#L432-L448), so the loop stays in C++ while the caller keeps its mask arithmetic in numpy.

The library version should not copy the shape of the private helpers. `_slope` divides by `x1 - x0`, so a vertical segment gives an infinite slope and `_offset` then gives nan; `_abscissa` divides by that slope, so a horizontal segment does the same. Normalizing the direction first and crossing against it is finite for every segment that has a length, and leaves one honest error to raise, the segment of zero length.

The reflection code is the first caller but not the only one. Any consumer that has to decide which side of a line a cell sits on writes the same three functions today.

Test in `tests/test_universe_shape0d.py` and `tests/test_universe_shape1d.py`, with the vertical, horizontal, and zero-length segments carried explicitly. When this lands, `_analytic.py` drops the three helpers and the TODO with them.

Related to #1269.

j8xixo12:
Slope, direction, and normal were the plane geometry of a segment, yet they lived as private helpers of the reflection analysis, so every consumer of Segment3d had to write them again.  Segment3d now carries `calc_length` and `calc_length2` (with the Point3d pair finally bound), `direction`, `normal`, `offset`, and `point_at`, and SegmentPad answers the vectorised form: `offset` of one arm against arrays of coordinates, with the loop in C++ and the mask arithmetic left to numpy.

`normal` takes the reference axis (default z) and returns the reference crossed with the direction, normalized, so the only segment without one is a segment parallel to the reference.  The implementation normalizes the direction instead of dividing a slope, so a vertical or a horizontal segment stays finite and the remaining honest error is the zero-length segment.  The reflection analysis drops its three private helpers and reads the geometry from the segment types.

Related to #1294

### DIFF
--- a/cpp/solvcon/universe/coord.hpp
+++ b/cpp/solvcon/universe/coord.hpp
@@ -16,6 +16,9 @@
 #include <solvcon/base.hpp>
 #include <solvcon/buffer/buffer.hpp>
 
+#include <cmath>
+#include <limits>
+
 namespace solvcon
 {
 
@@ -871,6 +874,50 @@ public:
 
     size_t size() const { return 2; }
 
+    value_type calc_length2() const;
+    value_type calc_length() const { return std::sqrt(calc_length2()); }
+
+    /**
+     * The unit vector from p0 to p1 for every non-zero segment, and the
+     * zero vector for a zero-length segment, without a branch that would
+     * diverge under SIMD or SIMT execution. Scaling by the largest
+     * coordinate difference first keeps the result a unit vector at any
+     * coordinate magnitude, from the subnormal range up to where the
+     * squared length would overflow.
+     */
+    point_type direction() const;
+
+    /**
+     * The unit normal perpendicular to both the segment and the reference
+     * axis: the direction crossed with the reference, normalized, so the
+     * normal, the direction, and the reference form a right-handed frame
+     * and the normal points outward on a counterclockwise polygon. With
+     * the default z reference a plane segment gets the direction turned a
+     * quarter turn clockwise. A segment parallel to the reference axis has
+     * no such normal and throws.
+     */
+    point_type normal_by_axis(Axis axis = Axis::Z) const;
+
+    /**
+     * The signed perpendicular distance from the line to the point,
+     * positive on the side the axis-referenced normal points to, the right
+     * of the direction. Normalizing the direction instead of dividing a
+     * slope keeps every segment with a length finite, vertical and
+     * horizontal alike.
+     */
+    value_type offset_by_axis(point_type const & p, Axis axis = Axis::Z) const
+    {
+        point_type const nml = normal_by_axis(axis);
+        return (p.x() - m_data.f.x0) * nml.x() + (p.y() - m_data.f.y0) * nml.y() + (p.z() - m_data.f.z0) * nml.z();
+    }
+
+    /**
+     * The point on the segment's line where the named axis reaches the
+     * value. Throws when the direction has no component along that axis,
+     * since the line then reaches the value nowhere or everywhere.
+     */
+    point_type point_along_axis(value_type value, Axis axis = Axis::Z) const;
+
     void mirror_x()
     {
         m_data.f.y0 = -m_data.f.y0;
@@ -919,6 +966,108 @@ private:
     detail::Segment3dData<T> m_data;
 
 }; /* end class Segment3d */
+
+template <typename T>
+typename Segment3d<T>::value_type Segment3d<T>::calc_length2() const
+{
+    value_type const dx = m_data.f.x1 - m_data.f.x0;
+    value_type const dy = m_data.f.y1 - m_data.f.y0;
+    value_type const dz = m_data.f.z1 - m_data.f.z0;
+    return dx * dx + dy * dy + dz * dz;
+}
+
+template <typename T>
+typename Segment3d<T>::point_type Segment3d<T>::direction() const
+{
+    value_type const tiny = std::numeric_limits<value_type>::min();
+    value_type const dx = m_data.f.x1 - m_data.f.x0;
+    value_type const dy = m_data.f.y1 - m_data.f.y0;
+    value_type const dz = m_data.f.z1 - m_data.f.z0;
+    // Scaling by the largest difference bounds the squared sum away from
+    // overflow and underflow at any coordinate magnitude; the smallest
+    // normal keeps the divisor non-zero and perturbs no physical length.
+    value_type const amax = std::max(std::abs(dx), std::max(std::abs(dy), std::abs(dz)));
+    value_type const scale = 1 / (amax + tiny);
+    value_type const ux = dx * scale;
+    value_type const uy = dy * scale;
+    value_type const uz = dz * scale;
+    // The scaled length sits far above the regularizer for every non-zero
+    // segment, so the factor is 1 / len there and exactly 0 at zero length.
+    value_type const len = std::sqrt(ux * ux + uy * uy + uz * uz);
+    value_type const fac = len / (len * len + tiny);
+    return point_type(ux * fac, uy * fac, uz * fac);
+}
+
+template <typename T>
+typename Segment3d<T>::point_type Segment3d<T>::normal_by_axis(Axis axis) const
+{
+    point_type const dir = direction();
+    value_type nx;
+    value_type ny;
+    value_type nz;
+    switch (axis)
+    {
+    case Axis::X:
+        nx = 0;
+        ny = dir.z();
+        nz = -dir.y();
+        break;
+    case Axis::Y:
+        nx = -dir.z();
+        ny = 0;
+        nz = dir.x();
+        break;
+    case Axis::Z:
+        nx = dir.y();
+        ny = -dir.x();
+        nz = 0;
+        break;
+    default: throw std::invalid_argument("Segment3d::normal_by_axis: invalid axis");
+    }
+
+    value_type const len = std::sqrt(nx * nx + ny * ny + nz * nz);
+    if (len == 0)
+    {
+        throw std::invalid_argument("Segment3d::normal_by_axis: segment is parallel to the reference axis");
+    }
+    return point_type(nx / len, ny / len, nz / len);
+}
+
+template <typename T>
+typename Segment3d<T>::point_type Segment3d<T>::point_along_axis(value_type value, Axis axis) const
+{
+    // The raw differences walk the same line as the unit direction with
+    // two fewer roundings, so an exactly representable crossing stays
+    // exact, and a component keeps its sign bit even where normalizing
+    // would flush it to zero.
+    value_type const dx = m_data.f.x1 - m_data.f.x0;
+    value_type const dy = m_data.f.y1 - m_data.f.y0;
+    value_type const dz = m_data.f.z1 - m_data.f.z0;
+    value_type origin;
+    value_type along;
+    switch (axis)
+    {
+    case Axis::X:
+        origin = m_data.f.x0;
+        along = dx;
+        break;
+    case Axis::Y:
+        origin = m_data.f.y0;
+        along = dy;
+        break;
+    case Axis::Z:
+        origin = m_data.f.z0;
+        along = dz;
+        break;
+    default: throw std::invalid_argument("Segment3d::point_along_axis: invalid axis");
+    }
+    if (along == 0)
+    {
+        throw std::invalid_argument("Segment3d::point_along_axis: the direction has no component along the axis");
+    }
+    value_type const t = (value - origin) / along;
+    return point_type(m_data.f.x0 + t * dx, m_data.f.y0 + t * dy, m_data.f.z0 + t * dz);
+}
 
 using Segment3dFp32 = Segment3d<float>;
 using Segment3dFp64 = Segment3d<double>;
@@ -1371,6 +1520,14 @@ public:
         }
     }
 
+    /**
+     * The signed perpendicular distances from the line of the indexed
+     * segment to every point (xs[i], ys[i], 0), positive on the side the
+     * axis-referenced normal points to. The loop stays here so the caller
+     * can keep its mask arithmetic in numpy.
+     */
+    SimpleArray<T> offset_by_axis(size_t index, SimpleArray<T> const & xs, SimpleArray<T> const & ys, Axis axis = Axis::Z) const;
+
 private:
 
     void check_constructor_point_size(point_pad_type const & p0, point_pad_type const & p1)
@@ -1388,6 +1545,40 @@ private:
     std::shared_ptr<point_pad_type> m_p1;
 
 }; /* end class SegmentPad */
+
+template <typename T>
+SimpleArray<T> SegmentPad<T>::offset_by_axis(size_t index, SimpleArray<T> const & xs, SimpleArray<T> const & ys, Axis axis) const
+{
+    if (index >= size())
+    {
+        throw std::out_of_range(
+            std::format("SegmentPad::offset_by_axis: index {} is out of bounds with size {}", index, size()));
+    }
+    if (xs.ndim() != 1 || ys.ndim() != 1 || xs.shape(0) != ys.shape(0))
+    {
+        throw std::invalid_argument(
+            std::format("SegmentPad::offset_by_axis: xs and ys must be 1-dimensional and equally long, "
+                        "but their ndims are {} and {} and shapes {} and {}",
+                        xs.ndim(),
+                        ys.ndim(),
+                        xs.shape(0),
+                        ys.shape(0)));
+    }
+
+    segment_type const seg = get_at(index);
+    point_type const nml = seg.normal_by_axis(axis);
+    value_type const x0 = seg.x0();
+    value_type const y0 = seg.y0();
+    // The query points sit at z = 0, the plane of a two-dimensional pad.
+    value_type const zterm = (0 - seg.z0()) * nml.z();
+    ssize_t const npoint = xs.shape(0);
+    SimpleArray<T> result(small_vector<ssize_t>{npoint});
+    for (ssize_t i = 0; i < npoint; ++i)
+    {
+        result(i) = (xs(i) - x0) * nml.x() + (ys(i) - y0) * nml.y() + zterm;
+    }
+    return result;
+}
 
 using SegmentPadFp32 = SegmentPad<float>;
 using SegmentPadFp64 = SegmentPad<double>;
--- a/solvcon/pilot/apps/obsrefl/_analytic.py
+++ b/solvcon/pilot/apps/obsrefl/_analytic.py
@@ -185,28 +185,6 @@ class Reflection(object):
         """Return the error of ``value`` relative to ``target``."""
         return (value - target) / target if target else float('nan')
 
-    # TODO: the three helpers below are the plane geometry of a segment, not
-    # of this problem, and every user of Segment3d has to write them again.
-    # Slope, direction, and normal belong on the class, and the offset wants
-    # to stay vectorised: a point-to-line distance taking arrays.
-
-    @staticmethod
-    def _slope(seg):
-        """Return the slope of the line ``seg`` lies on."""
-        return (seg.y1 - seg.y0) / (seg.x1 - seg.x0)
-
-    @classmethod
-    def _offset(cls, seg, xs, ys):
-        """Return the perpendicular distance from ``seg``, positive above
-        it."""
-        slope = cls._slope(seg)
-        return (ys - seg.y0 - slope * (xs - seg.x0)) / math.hypot(1.0, slope)
-
-    @classmethod
-    def _abscissa(cls, seg, height):
-        """Return where the line of ``seg`` reaches ``height``."""
-        return seg.x0 + (height - seg.y0) / cls._slope(seg)
-
     @property
     def has_reflection(self):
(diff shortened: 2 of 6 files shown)


cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,117 @@ class Segment3d
...
+    }
+
+    /**
+     * The unit normal perpendicular to both the segment and the reference
+     * axis: the reference crossed with the direction, normalized. With the
+     * default z reference a plane segment gets the direction turned a
+     * quarter turn counterclockwise. A segment parallel to the reference
+     * axis has no such normal and throws.
+     */
+    point_type normal(Axis axis = Axis::Z) const
```
yungyuc: It is ambiguous to name it just `normal()`.  `normal_by_axis()` would work better:

```suggestion
    point_type normal_by_axis(Axis axis = Axis::Z) const
```

cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,117 @@ class Segment3d
 
     size_t size() const { return 2; }
 
+    value_type calc_length2() const
```
yungyuc: Implement long (more than 2 lines) functions outside the class declaration.

cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,117 @@ class Segment3d
...
+    }
+
+    /**
+     * The signed perpendicular distance from the line to the point,
+     * positive on the side the (z-referenced) normal points to, the left
+     * of the direction. Normalizing the direction instead of dividing a
+     * slope keeps every segment with a length finite, vertical and
+     * horizontal alike.
+     */
+    value_type offset(point_type const & p) const
```
yungyuc:
```suggestion
    value_type offset_by_axis(point_type const & p, Axis axis = Axis::Z) const
```

For this is an ad hoc implementation, `offset()` is too general a name.  A general name should be reserved for a really general definition.

cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,117 @@ class Segment3d
...
+    {
+        point_type const nml = normal();
```
yungyuc: It is wrong to hard-code the normal direction by the Z axis.

cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,117 @@ class Segment3d
...
+        return (p.x() - m_data.f.x0) * nml.x() + (p.y() - m_data.f.y0) * nml.y() + (p.z() - m_data.f.z0) * nml.z();
+    }
+
+    /**
+     * The point on the segment's line where the named axis reaches the
+     * value. Throws when the direction has no component along that axis,
+     * since the line then reaches the value nowhere or everywhere.
+     */
+    point_type point_at(Axis axis, value_type value) const
```
yungyuc:
```suggestion
    point_type point_along_axis(Axis axis, value_type value) const
```

This is an ad hoc implementation, use a clearly defined function name.

yungyuc (changes requested):
- [x] Keep the implementation of long (more than 2 lines) functions outside the class declaration.
- [x] Use `normal_by_axis()` instead of `normal()`.
- [x] Use `offset_by_axis()` instead of `offset()`.
- [x] Use `point_along_axis()` instead of `point_at()`.

cpp/solvcon/universe/coord.hpp
```diff
@@ -849,6 +849,44 @@ class Segment3d
...
+        return (p.x() - m_data.f.x0) * nml.x() + (p.y() - m_data.f.y0) * nml.y() + (p.z() - m_data.f.z0) * nml.z();
+    }
+
+    /**
+     * The point on the segment's line where the named axis reaches the
+     * value. Throws when the direction has no component along that axis,
+     * since the line then reaches the value nowhere or everywhere.
+     */
+    point_type point_along_axis(value_type value, Axis axis = Axis::Z) const;
+
```
j8xixo12, line 920: These functions have been renamed by adding `_by_axis` postfix for avoiding ambiguity.

cpp/solvcon/universe/coord.hpp
```diff
@@ -1363,6 +1494,40 @@ class SegmentPad
...
+            std::format("SegmentPad::offset_by_axis: xs and ys must be 1-dimensional and equally long, "
+                        "but their ndims are {} and {} and shapes {} and {}",
+                        xs.ndim(),
+                        ys.ndim(),
+                        xs.shape(0),
+                        ys.shape(0)));
+    }
+
+    segment_type const seg = get_at(index);
+    point_type const nml = seg.normal_by_axis(axis);
```
j8xixo12, line 1569: The reference axis is input by function argument.

cpp/solvcon/universe/coord.hpp
yungyuc: `_by_axis` and/or `_along_axis`.

cpp/solvcon/universe/coord.hpp
```diff
@@ -898,6 +936,91 @@ class Segment3d
...
+    return dx * dx + dy * dy + dz * dz;
+}
+
+template <typename T>
+typename Segment3d<T>::point_type Segment3d<T>::direction() const
+{
+    value_type const len = calc_length();
+    if (len == 0)
+    {
+        throw std::invalid_argument("Segment3d::direction: segment has zero length");
```
yungyuc: Don't throw on 0 length.  Return `(0, 0, 0)` instead.

cpp/solvcon/universe/coord.hpp
```diff
@@ -898,6 +936,91 @@ class Segment3d
...
+    }
+    return point_type((m_data.f.x1 - m_data.f.x0) / len, (m_data.f.y1 - m_data.f.y0) / len, (m_data.f.z1 - m_data.f.z0) / len);
```
yungyuc: Don't throw on 0 length.  Return `(0, 0, 0)` instead.

```suggestion
    value_type const fac = len / (len*len + std::numeric_limits<value_type>::epsilon());
    return point_type((m_data.f.x1 - m_data.f.x0) * fac, (m_data.f.y1 - m_data.f.y0) * fac, (m_data.f.z1 - m_data.f.z0) * fac);
```

The first line is numerically equivalent to

```cpp
fac = (len == 0) ? 0 : (1.0/len);
```

Dropping the `if` (or ternary) avoids branching instruction and will make it faster on CPU and easier to be ported to GPU (which hates branching).  The additional multiplications take negligible runtime overhead on modern silicons (latency is dominated by memory access).

cpp/solvcon/universe/coord.hpp
```diff
@@ -898,6 +936,91 @@ class Segment3d
...
+        break;
+    case Axis::Y:
+        nx = dir.z();
+        ny = 0;
+        nz = -dir.x();
+        break;
+    case Axis::Z:
+        nx = -dir.y();
+        ny = dir.x();
+        nz = 0;
```
yungyuc, line 1023: Let the normal vector be $$\vec{n}$$, segment unit vector $$\vec{t}$$ (tangential), axis unit vector $$\vec{a}$$.  The 3 vectors $$\vec{n}, \vec{t}, \vec{a}$$ must form a Cartesian coordinate system and $$\vec{n}\times\vec{t} = \vec{a}$$ (" $$\times$$ " is the vector cross product), $$\vec{t}\times\vec{a} = \vec{n}$$, and $$\vec{a}\times\vec{n} = \vec{t}$$.  That is, $$\vec{n}, \vec{t}, \vec{a}$$ are the basis (unit vectors along the axes) of the new coordinate system $$(\xi_1, \xi_2, \xi_3)$$ and follow right-hand rule.

In this way, $$\vec{n}$$ is pointing outward when segments are connected to form a polygon.

Code under `Axis::X`, `Axis::Y`, `Axis::Z` should be changed to fulfill the requirement (make sure all axes are consistent):

```suggestion
    case Axis::Z:
        nx = dir.y();
        ny = -dir.x();
        nz = 0;
```

Add unit tests for the cross-product results.

cpp/solvcon/universe/coord.hpp
```diff
@@ -898,6 +936,91 @@ class Segment3d
...
+        break;
+    case Axis::Z:
+        origin = m_data.f.z0;
+        along = dir.z();
+        break;
+    default: throw std::invalid_argument("Segment3d::point_along_axis: invalid axis");
+    }
+    if (along == 0)
+    {
+        throw std::invalid_argument("Segment3d::point_along_axis: the direction has no component along the axis");
```
yungyuc, line 1066: Keep the `point_along_axis()` exception for now.

tests/test_universe_shape1d.py
```diff
@@ -63,6 +63,118 @@ def test_mirror(self):
...
+
+        s = Segment(Point(1, 2, 0), Point(4, 6, 0))
+        self.assert_allclose(tuple(s.direction()), (0.6, 0.8, 0.0))
+        # A vertical and a horizontal segment stay finite: the direction
+        # is normalized instead of dividing a slope.
+        vertical = Segment(Point(2, -1, 0), Point(2, 9, 0))
+        self.assert_allclose(tuple(vertical.direction()), (0.0, 1.0, 0.0))
+        horizontal = Segment(Point(3, 5, 0), Point(-7, 5, 0))
+        self.assert_allclose(tuple(horizontal.direction()), (-1.0, 0.0, 0.0))
+        with self.assertRaisesRegex(
```
yungyuc: Replace to test for zero output.

tests/test_universe_shape1d.py
```diff
@@ -63,6 +63,118 @@ def test_mirror(self):
...
+                ValueError, "Segment3d::direction: segment has zero length"):
+            Segment(Point(1, 2, 3), Point(1, 2, 3)).direction()
+
+    def test_normal_by_axis(self):
+        Point = self.Point
+        Segment = self.Segment
+
+        s = Segment(Point(0, 0, 0), Point(10, 0, 0))
+        self.assert_allclose(tuple(s.normal_by_axis()), (0.0, 1.0, 0.0))
```
yungyuc: The results should be exact.  Avoid allclose when you can get exact results.

yungyuc (changes requested):
- [x] Don't throw on 0 length in `direction()`.  Return `(0, 0, 0)` instead.
- [x] Change `normal_by_axis()` to follow the right-hand rule $$\vec{n}\times\vec{t} = \vec{a}$$ to make the normal vector $$\vec{n}$$ pointing outward.  Add unit tests for the cross product.
- [x] Avoid allclose in testing when the results can be exact.

cpp/solvcon/universe/coord.hpp
```diff
@@ -898,6 +943,88 @@ class Segment3d
...
+    value_type const dz = m_data.f.z1 - m_data.f.z0;
+    return dx * dx + dy * dy + dz * dz;
+}
+
+template <typename T>
+typename Segment3d<T>::point_type Segment3d<T>::direction() const
+{
+    value_type const len = calc_length();
+    value_type const fac = len / (len * len + std::numeric_limits<value_type>::epsilon());
+    return point_type((m_data.f.x1 - m_data.f.x0) * fac, (m_data.f.y1 - m_data.f.y0) * fac, (m_data.f.z1 - m_data.f.z0) * fac);
```
j8xixo12: The direction vector will not be a unit vector when the segment length is small.

```
... import numpy as np, solvcon as sc
... for S, P in ((sc.Segment3dFp32, sc.Point3dFp32),
...         (sc.Segment3dFp64, sc.Point3dFp64)):
...     for L in (1.0, 1e-2, 1e-3, 1e-4, 1e-6, 1e-8):
...         d = S(P(0, 0, 0), P(L, 0, 0)).direction()
...         print(S.__name__, L, np.hypot(d.x, d.y))
...
Segment3dFp32 1.0 0.9999998807907104
Segment3dFp32 0.01 0.9988093376159668
Segment3dFp32 0.001 0.8934879302978516
Segment3dFp32 0.0001 0.07739381492137909
Segment3dFp32 1e-06 8.38853793538874e-06
Segment3dFp32 1e-08 8.388608141451925e-10
Segment3dFp64 1.0 0.9999999999999998
Segment3dFp64 0.01 0.9999999999977796
Segment3dFp64 0.001 0.9999999997779555
Segment3dFp64 0.0001 0.9999999777955401
Segment3dFp64 1e-06 0.9997780046879362
Segment3dFp64 1e-08 0.3105159921038857
```

tests/test_universe_shape1d.py
```diff
@@ -63,6 +63,154 @@ def test_mirror(self):
...
+                ValueError,
+                "Segment3d::normal_by_axis: "
+                "segment is parallel to the reference axis"):
+            Segment(Point(1, 2, 3), Point(1, 2, 3)).normal_by_axis()
+        with self.assertRaisesRegex(
+                ValueError,
+                "Segment3d::normal_by_axis: axis must be 'x', 'y', or 'z'"):
+            s.normal_by_axis('Z')
+
+    def test_normal_by_axis_right_hand_rule(self):
```
j8xixo12, line 164: Make sure the `normal_by_axis` follows the right-hand rule.

tests/test_universe_shape1d.py
```diff
@@ -63,6 +63,154 @@ def test_mirror(self):
...
+        self.assert_allclose(tuple(s.direction()), (0.6, 0.8, 0.0))
+        # A vertical and a horizontal segment stay finite: the direction
+        # is normalized instead of dividing a slope.
+        vertical = Segment(Point(2, -1, 0), Point(2, 9, 0))
+        self.assertEqual(tuple(vertical.direction()), (0.0, 1.0, 0.0))
+        horizontal = Segment(Point(3, 5, 0), Point(-7, 5, 0))
+        self.assertEqual(tuple(horizontal.direction()), (-1.0, 0.0, 0.0))
+        # A zero-length segment regularizes to the zero vector.
+        zero = Segment(Point(1, 2, 3), Point(1, 2, 3))
+        self.assertEqual(tuple(zero.direction()), (0.0, 0.0, 0.0))
```
j8xixo12, line 104: The `direction()` returns `(0, 0, 0)` instead throw an error.

solvcon/pilot/apps/obsrefl/_analytic.py
```diff
@@ -239,16 +217,18 @@ def zone_masks(self, margin=None):
...
+        reflected = (self.arms.offset_by_axis(1, xs_arr, ys_arr).ndarray
                      if self.has_reflection
-                     else np.full(xs.shape, np.inf, dtype='float64'))
-        return [incident < -gap,
-                (incident > gap) & (reflected > gap),
-                reflected < -gap]
+                     else np.full(xs.shape, -np.inf, dtype='float64'))
+        return [incident > gap,
+                (incident < -gap) & (reflected < -gap),
+                reflected > gap]
```
j8xixo12, line 223: Since `normal_by_axis()` follows the right-hand rule, zone identification for the oblique shock must be updated.

cpp/solvcon/universe/coord.hpp
yungyuc: Good catch.  It is a good test and revealed a severe flaw in the implemented I provided you.  `len / (len * len)` lost precision when `len` is close to the floating-point precision limit.  Please help enhance it while keeping the implementation branchless.

Perhaps you can start with trying this tweak:

```cpp
    value_type const tiny = std::numeric_limits<value_type>::min();
    value_type const dx = m_data.f.x1 - m_data.f.x0;
    value_type const dy = m_data.f.y1 - m_data.f.y0;
    value_type const dz = m_data.f.z1 - m_data.f.z0;
    // Scaling by the largest difference bounds the squared sum away from
    // overflow and underflow at any coordinate magnitude; the smallest
    // normal keeps the divisor non-zero and perturbs no physical length.
    value_type const amax = std::max(std::abs(dx), std::max(std::abs(dy), std::abs(dz)));
    value_type const scale = amax + tiny;
    value_type const ux = dx / scale;
    value_type const uy = dy / scale;
    value_type const uz = dz / scale;
    // The scaled length sits far above the regularizer for every non-zero
    // segment, so the factor is 1 / len there and exactly 0 at zero length.
    value_type const len = std::sqrt(ux * ux + uy * uy + uz * uz);
    value_type const fac = len / (len * len + tiny);
    return point_type(ux * fac, uy * fac, uz * fac);
```

I did not test it myself and not sure about how good it works.

yungyuc (changes requested):
Thank you for the swift update and catching the defect in my suggestion.  Please help polish it.

- [x] Improve the branchless `direction()` implementation.

tests/test_universe_shape1d.py
```diff
@@ -77,6 +77,171 @@ def test_mirror(self):
...
+        # around 1e-4 down in single precision and overflowed to nan for
+        # a length beyond around 2e19.
+        for length in (1e-8, 1e-6, 1e-4, 1.0, 1e4, 1e20):
+            with self.subTest(length=length):
+                s = Segment(Point(0, 0, 0), Point(length, 0, 0))
+                self.assertEqual(tuple(s.direction()), (1.0, 0.0, 0.0))
+        # A mixed huge segment must not overflow the squared sum.
+        wide = Segment(Point(0, 0, 0), Point(1e20, 1e20, 0))
+        self.assert_allclose(tuple(wide.direction()),
+                             (2.0 ** -0.5, 2.0 ** -0.5, 0.0))
```
j8xixo12, line 121: Make sure the `direction()` is working properly in any magnitude.

cpp/solvcon/universe/coord.hpp
```diff
@@ -920,6 +967,108 @@ class Segment3d
...
+    value_type const dy = m_data.f.y1 - m_data.f.y0;
+    value_type const dz = m_data.f.z1 - m_data.f.z0;
+    // Scaling by the largest difference bounds the squared sum away from
+    // overflow and underflow at any coordinate magnitude; the smallest
+    // normal keeps the divisor non-zero and perturbs no physical length.
+    value_type const amax = std::max(std::abs(dx), std::max(std::abs(dy), std::abs(dz)));
+    value_type const scale = 1 / (amax + tiny);
+    value_type const ux = dx * scale;
+    value_type const uy = dy * scale;
+    value_type const uz = dz * scale;
```
j8xixo12, line 993: Division is more expensive than multiplication, so calculate the division once and use three multiplications instead of performing multiple divisions.

tests/test_universe_shape1d.py
yungyuc: Nit: Test code will looks clearer if the tolerance is spelled out here.  Consider to improve this in a followup PR.

yungyuc (approved):
LGTM

PR #1365, opened 2026-08-16
Add scalar constructors for Segment3d and document 2D convention

### LINKED ISSUE #1293: Give `Segment3d` scalar constructors and document the 2D pad convention
Plane code carries a coordinate it never uses. The geometry types stay 3D, as issue #754 records they are meant to, so close the cost where it is paid instead of adding a plane family: let a segment be built from loose coordinates, and write down what the z of a 2D pad means.

What plane code pays today, checked against the extension built at master (1a504278):

- `Segment3d` has no scalar constructor. `Segment3dFp64(0., 0., 1., 1.)` raises `TypeError`, because the only bound form is `Segment3dFp64(p0, p1)` at [`wrap_shape1d.cpp:L62-L64`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape1d.cpp#L62-L64), matching the single C++ constructor at [`coord.hpp:L771-L774`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/coord.hpp#L771-L774). Building a segment outside a pad costs two point constructions.
- `pad[i]` on a 2D pad returns a `Segment3dFp64` whose `p0` and `p1` carry a z that is always zero. Nothing says so, so a reader has to find it out.
- The habit shows across the plane code. [`_naca.py:L211-L214`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/airfoil/_naca.py#L211-L214) rebuilds both endpoints only to drop the z it has just read, `world.add_segment(Point(p0.x, p0.y, 0), Point(p1.x, p1.y, 0))`, and 28 of the 45 `Point3dFp64(...)` construction sites under `solvcon/` pass an explicit zero.

The work:

- Add `Segment3d(x0, y0, x1, y1)` and `Segment3d(x0, y0, z0, x1, y1, z1)` in `coord.hpp`, and bind both. The four-scalar form leaves z at zero, so it reads as the plane constructor without being a separate type. The pair matches the append overloads `SegmentPad` already carries.
- State the convention in the class docstrings of `Point3d`, `Segment3d`, and the pads: a pad of ndim 2 stores and reports z as zero, and a segment or point read back from it carries that zero. Plane code is meant to ignore it, not to work around it.
- Test in `tests/test_universe_shape1d.py`: construct from four scalars and from six, and check each against the same segment built from a point pair.

A plane family, `Point2d` and `Segment2d`, was the other direction, and this issue turns it down. It doubles the wrapped surface (points, segments, beziers, the three pads, `World::add_segment`, the canvas) and puts a branch on ndim in every consumer, for a cost that is ergonomic rather than structural. It is worth reopening only together with #754, which asks the same question one layer up, at the canvas.

## Issue background

The oblique-shock reflection work (issue #1269, PR #1289) had two TODO notes for handling 3D types on a 2D pad.

[`_analytic.py:L168-L170`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L168-L170):

```python
# TODO: SegmentPad takes a Segment3d built from two Point3d, unlike
# PointPad, which appends loose coordinates.  An append(x0, y0, x1,
# y1) overload would drop the two constructions per arm.
```

[`_analytic.py:L174-L176`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L174-L176):

```python
# TODO: the endpoints are 3D even on a 2D pad, so a plane user
# pins z twice per segment.  A Segment2d, or a pad that carries
# only ndim coordinates per point, would take the zeros away.
```

Both asks are already served on the write side of the pad, which is why the work above is the segment constructor and the convention alone. `SegmentPad.append(x0, y0, x1, y1)` is bound at [`wrap_shape1d.cpp:L337-L346`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape1d.cpp#L337-L346), and `Point3d` takes two coordinates with z defaulting to zero at [`wrap_shape0d.cpp:L61`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/cpp/solvcon/universe/pymod/wrap_shape0d.cpp#L61). The loop body at [`_analytic.py:L171-L179`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L171-L179) can be corrected to one line right now, with no point and no segment constructed:

```python
self.arms.append(head[0], head[1], tail[0], tail[1])
```

The two TODO notes go away with this issue.

YenPeiChen07:
Add two scalar constructors (four-scalar and six-scalar) for `Segment3d` and bind them into Python. The two constructors enable `Segment3d` to be constructed without costing two point constructors outside a pad. They are tested against the same segment built from a point pair in `tests/test_universe_shape1d.py`. 

Also, the plane code convention is added in the class docstrings of `Point3d`, `PointPad`, `Segment3d`, and `SegmentPad`.

Related to #1293.

### DIFF
--- a/cpp/solvcon/universe/coord.hpp
+++ b/cpp/solvcon/universe/coord.hpp
@@ -34,6 +34,10 @@ enum class Axis : uint8_t
 /**
  * Point in three-dimensional space.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. 2-dimensional code is meant to ignore it, not
+ * to work around it.
+ *
  * @tparam T floating-point type
  *
  * @ingroup group_geometry
@@ -271,6 +275,10 @@ using Point3dFp64 = Point3d<double>;
  * Coordinates are stored as separate per-axis arrays (x, y, and, for
  * three dimensions, z), so the layout is structure-of-arrays. The
  * dimensionality (2 or 3) is fixed at construction and cannot change.
+ *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. 2-dimensional code is meant to ignore it, not
+ * to work around it.
  *
  * @tparam T floating-point type
  *
@@ -752,6 +760,10 @@ union Segment3dData
 /**
  * Segment in three-dimensional space.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. 2-dimensional code is meant to ignore it, not
+ * to work around it.
+ *
  * @tparam T floating-point type
  *
  * @ingroup group_geometry
@@ -770,6 +782,16 @@ public:
 
     Segment3d(point_type const & p0, point_type const & p1)
         : m_data{{p0.x(), p1.x(), p0.y(), p1.y(), p0.z(), p1.z()}}
+    {
+    }
+
+    Segment3d(T x0, T y0, T x1, T y1)
+        : m_data{{x0, x1, y0, y1, /*z0*/ 0.0, /*z1*/ 0.0}}
+    {
+    }
+
+    Segment3d(T x0, T y0, T z0, T x1, T y1, T z1)
+        : m_data{{x0, x1, y0, y1, z0, z1}}
     {
     }
 
@@ -907,6 +929,10 @@ using Segment3dFp64 = Segment3d<double>;
  * Each segment is stored as a pair of endpoints, held in two PointPad
  * objects (one for each endpoint). The dimensionality (2 or 3) comes
  * from the underlying point pads.
+ *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. 2-dimensional code is meant to ignore it, not
+ * to work around it.
  *
  * @tparam T floating-point type
  *
--- a/cpp/solvcon/universe/pymod/wrap_shape1d.cpp
+++ b/cpp/solvcon/universe/pymod/wrap_shape1d.cpp
@@ -62,6 +62,18 @@ WrapSegment3d<T> & WrapSegment3d<T>::wrap_management()
         .def(py::init<point_type const &, point_type const &>(),
              py::arg("p0"),
              py::arg("p1"))
+        .def(py::init<value_type, value_type, value_type, value_type>(),
+             py::arg("x0"),
+             py::arg("y0"),
+             py::arg("x1"),
+             py::arg("y1"))
+        .def(py::init<value_type, value_type, value_type, value_type, value_type, value_type>(),
+             py::arg("x0"),
+             py::arg("y0"),
+             py::arg("z0"),
+             py::arg("x1"),
+             py::arg("y1"),
+             py::arg("z1"))
         //
         ;
 
--- a/solvcon/pilot/apps/obsrefl/_analytic.py
+++ b/solvcon/pilot/apps/obsrefl/_analytic.py
@@ -165,18 +165,10 @@ class Reflection(object):
         self.margin = margin
         # The analytic path is fixed once the mesh is built, so its corners
         # are walked once into the segments the measurements read.
-        # TODO: SegmentPad takes a Segment3d built from two Point3d, unlike
-        # PointPad, which appends loose coordinates.  An append(x0, y0, x1,
-        # y1) overload would drop the two constructions per arm.
         self.arms = core.SegmentPadFp64(ndim=2)
         path = shock.shock_path()
         for head, tail in zip(path[:-1], path[1:]):
-            # TODO: the endpoints are 3D even on a 2D pad, so a plane user
-            # pins z twice per segment.  A Segment2d, or a pad that carries
-            # only ndim coordinates per point, would take the zeros away.
-            self.arms.append(core.Segment3dFp64(
-                core.Point3dFp64(head[0], head[1], 0.0),
-                core.Point3dFp64(tail[0], tail[1], 0.0)))
+            self.arms.append(head[0], head[1], tail[0], tail[1])
 
     @property
     def incident(self):
--- a/tests/test_universe_shape1d.py
+++ b/tests/test_universe_shape1d.py
@@ -21,6 +21,20 @@ class Segment3dTB(testing.TestBase):
         self.assertEqual(tuple(s.p0), (0.0, 0.0, 0.0))
         self.assertEqual(tuple(s.p1), (1.0, 1.0, 1.0))
 
+        # Test six-scalar construcotr
+        s = Segment(p0=Point(x=1, y=2, z=3), p1=Point(x=4, y=5, z=6))
+        s_6scalar = Segment(x0=1, y0=2, z0=3, x1=4, y1=5, z1=6)
+        self.assertEqual(s_6scalar, s)
+        s_6scalar = Segment(1, 2, 3, 4, 5, 6)
+        self.assertEqual(s_6scalar, s)
+
+        # Test four-scalar construcotr
+        s = Segment(p0=Point(x=1, y=2), p1=Point(x=4, y=5))
+        s_4scalar = Segment(x0=1, y0=2, x1=4, y1=5)
+        self.assertEqual(s_4scalar, s)
+        s_4scalar = Segment(1, 2, 4, 5)
+        self.assertEqual(s_4scalar, s)
+
         s.p0 = Point(x=3, y=7, z=0)
         s.p1 = Point(x=-1, y=-4, z=9)
         self.assertEqual(s.x0, 3)


cpp/solvcon/universe/pymod/wrap_shape1d.cpp
```diff
@@ -62,6 +62,18 @@ WrapSegment3d<T> & WrapSegment3d<T>::wrap_management()
         .def(py::init<point_type const &, point_type const &>(),
              py::arg("p0"),
              py::arg("p1"))
+        .def(py::init<value_type, value_type, value_type, value_type>(),
```
YenPeiChen07, line 65: Bind `Segment3d` four-scalar constructor.

cpp/solvcon/universe/pymod/wrap_shape1d.cpp
```diff
@@ -62,6 +62,18 @@ WrapSegment3d<T> & WrapSegment3d<T>::wrap_management()
...
+             py::arg("x0"),
+             py::arg("y0"),
+             py::arg("x1"),
+             py::arg("y1"))
+        .def(py::init<value_type, value_type, value_type, value_type, value_type, value_type>(),
```
YenPeiChen07, line 70: Bind `Segment3d` six-scalar constructor.

cpp/solvcon/universe/coord.hpp
```diff
@@ -34,6 +34,10 @@ enum class Axis : uint8_t
 /**
  * Point in three-dimensional space.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
```
YenPeiChen07, line 37: State 2D pad convention.

cpp/solvcon/universe/coord.hpp
```diff
@@ -773,6 +785,16 @@ class Segment3d
     {
     }
 
+    Segment3d(T x0, T y0, T x1, T y1)
```
YenPeiChen07, line 788: Add four-scalar constructor.

cpp/solvcon/universe/coord.hpp
```diff
@@ -773,6 +785,16 @@ class Segment3d
...
+        : m_data{{x0, x1, y0, y1, 0.0, 0.0}}
+    {
+    }
+
+    Segment3d(T x0, T y0, T z0, T x1, T y1, T z1)
```
YenPeiChen07, line 793: Add six-scalar constructor.

tests/test_universe_shape1d.py
```diff
@@ -21,6 +21,13 @@ def test_construct(self):
         self.assertEqual(tuple(s.p0), (0.0, 0.0, 0.0))
         self.assertEqual(tuple(s.p1), (1.0, 1.0, 1.0))
 
+        s_6scalar = Segment(x0=0, y0=0, z0=0, x1=1, y1=1, z1=1)
```
YenPeiChen07: Test segment constructed with scalar against with the segment built from a point pair.

cpp/solvcon/universe/coord.hpp
```diff
@@ -904,9 +926,9 @@ using Segment3dFp64 = Segment3d<double>;
 /**
  * Container of line segments in two- or three-dimensional space.
  *
- * Each segment is stored as a pair of endpoints, held in two PointPad
```
yungyuc, line 907: The original comments are not wrong.  Please elaborate why do you drop it.

tests/test_universe_shape1d.py
yungyuc: Also test with positional arguments: 

```python
Segment(0, 0, 0, 1, 1, 1)
Segment(0, 0, 1, 1)
```

yungyuc (changes requested):
- [x] The original comments for `SegmentPad` are not wrong.  Elaborate why do you drop it.
- [x] The related TODO's in [`_analytic.py`](https://github.com/solvcon/solvcon/blob/1a5042788e1795014eba6c46981a0de4c067bbe8/solvcon/pilot/apps/obsrefl/_analytic.py#L168-L170) have not been updated to use the new API.
- [x] Test both positional and keyword arguments to the Python wrapper of `Segment`.

cpp/solvcon/universe/coord.hpp
YenPeiChen07: Sorry, the original comments were deleted by mistake. I will add them back.

cpp/solvcon/universe/coord.hpp
```diff
@@ -272,6 +276,10 @@ using Point3dFp64 = Point3d<double>;
  * three dimensions, z), so the layout is structure-of-arrays. The
  * dimensionality (2 or 3) is fixed at construction and cannot change.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
```
tigercosmos, line 279: The sentence "A pad of ndim 2 stores and reports z as zero" is not accurate. A 2D `PointPad` does not store z: `append(x, y)` and `expand()` leave `m_z` empty, and `tests/test_universe_shape0d.py` asserts `len(pp.z) == 0` for a 2D pad. Only `get()` and `get_at()` synthesize `0.0` when they build a `Point3d` or a `Segment3d`. `z_at(i)` throws and `z()` returns an empty array.

Please reword, for example: a 2D pad has no z storage; `get` and `get_at` report z as 0. The same sentence appears at lines 37, 763, and 929.

cpp/solvcon/universe/coord.hpp
tigercosmos, line 37: This paragraph is the same text as `PointPad` at line 279, `Segment3d` at line 763, and `SegmentPad` at line 929. Here "a pad" is a concept that the header declares 240 lines later, and there is no `@ref`. The two new constructors at lines 788-796 have no brief.

Consider a one-line brief on each new constructor (z0 and z1 are zero) and `@ref PointPad` in the value types. That replaces the fourth copy of the paragraph. Four copies drift on the next wording fix; the correction in the thread at line 279 already applies to all four.

tests/test_universe_shape1d.py
tigercosmos: The scalar-constructor checks use axis-symmetric values: `(0, 0, 0)` to `(1, 1, 1)` and `(0, 0)` to `(1, 1)`. A swap of x and y (or x, y, z) in the constructor's mapping into the `{x0, x1, y0, y1, z0, z1}` union produces the same `m_data`. A `py::arg` bound to the wrong position produces the same result. Both still pass.

Please use distinct coordinates as `test_mirror` does, for example `Segment(x0=1, y0=2, z0=3, x1=4, y1=5, z1=6)` against `Segment(Point(1, 2, 3), Point(4, 5, 6))`.

tests/test_universe_shape1d.py
```diff
@@ -21,6 +21,13 @@ def test_construct(self):
...
+        self.assertEqual(s_6scalar, s)
+
```
tigercosmos, line 30: `s.z1 = 0.0` mutates the fixture to produce the expected value. The four-scalar check then depends on the `z1` setter, and it covers the z0 default only because `s.z0` is already 0.

Build the expectation directly, for example `self.assertEqual(Segment(x0=1, y0=2, x1=4, y1=5), Segment(Point(1, 2, 0), Point(4, 5, 0)))`. The positional forms from the thread above, `Segment(1, 2, 3, 4, 5, 6)` and `Segment(1, 2, 4, 5)`, fit here as well.

cpp/solvcon/universe/coord.hpp
tigercosmos, line 788: The four-scalar constructor spells out the interleaved storage order `{x0, x1, y0, y1, 0.0, 0.0}` a third time. `Point3d(T x, T y)` at line 68 delegates instead: `Point3d(x, y, /*z*/ 0.0)`.

Suggest `Segment3d(T x0, T y0, T x1, T y1) : Segment3d(x0, y0, /*z0*/ 0.0, x1, y1, /*z1*/ 0.0) {}` so that the slot order lives in one place.

cpp/solvcon/universe/coord.hpp
```diff
@@ -773,6 +785,16 @@ class Segment3d
     {
     }
 
+    Segment3d(T x0, T y0, T x1, T y1)
+        : m_data{{x0, x1, y0, y1, 0.0, 0.0}}
```
tigercosmos: `m_data{{x0, x1, y0, y1, 0.0, 0.0}}` brace-initializes `T v[6]` with `double` literals. For an integral `T`, which `static_assert(std::is_arithmetic_v<T>)` admits, that is a narrowing conversion and the constructor fails to compile. This is latent today because only Fp32 and Fp64 exist. `T{}` avoids it, and the delegating form suggested above avoids it as well.

tigercosmos (commented):
This is an AI-generated review (Claude Code) that I ran and read through before posting. The inline comments cover points that the existing threads do not raise. They concern the accuracy of the new docstring, the values in the new test, and the four-scalar constructor.

cpp/solvcon/universe/coord.hpp
```diff
@@ -773,6 +785,16 @@ class Segment3d
     {
     }
 
+    Segment3d(T x0, T y0, T x1, T y1)
+        : m_data{{x0, x1, y0, y1, /*z0*/ 0.0, /*z1*/ 0.0}}
```
YenPeiChen07, line 789: Add comments in four-scalar constructor to specify the default values are for `z0` and `z1` respectively.

solvcon/pilot/apps/obsrefl/_analytic.py
```diff
@@ -165,18 +165,10 @@ def __init__(self, shock, margin=MARGIN_FRACTION):
...
         self.arms = core.SegmentPadFp64(ndim=2)
         path = shock.shock_path()
         for head, tail in zip(path[:-1], path[1:]):
-            # TODO: the endpoints are 3D even on a 2D pad, so a plane user
-            # pins z twice per segment.  A Segment2d, or a pad that carries
-            # only ndim coordinates per point, would take the zeros away.
-            self.arms.append(core.Segment3dFp64(
-                core.Point3dFp64(head[0], head[1], 0.0),
-                core.Point3dFp64(tail[0], tail[1], 0.0)))
+            self.arms.append(head[0], head[1], tail[0], tail[1])
```
YenPeiChen07, line 171: Update to use the new API to append a segment to pad by scalars.

tests/test_universe_shape1d.py
```diff
@@ -21,6 +21,20 @@ def test_construct(self):
         self.assertEqual(tuple(s.p0), (0.0, 0.0, 0.0))
         self.assertEqual(tuple(s.p1), (1.0, 1.0, 1.0))
 
+        # Test six-scalar construcotr
+        s = Segment(p0=Point(x=1, y=2, z=3), p1=Point(x=4, y=5, z=6))
+        s_6scalar = Segment(x0=1, y0=2, z0=3, x1=4, y1=5, z1=6)
+        self.assertEqual(s_6scalar, s)
+        s_6scalar = Segment(1, 2, 3, 4, 5, 6)
+        self.assertEqual(s_6scalar, s)
```
YenPeiChen07, line 29: Test six-scalar contractor with both positional and keyword arguments.

tests/test_universe_shape1d.py
```diff
@@ -21,6 +21,20 @@ def test_construct(self):
...
+
+        # Test four-scalar construcotr
+        s = Segment(p0=Point(x=1, y=2), p1=Point(x=4, y=5))
+        s_4scalar = Segment(x0=1, y0=2, x1=4, y1=5)
+        self.assertEqual(s_4scalar, s)
+        s_4scalar = Segment(1, 2, 4, 5)
+        self.assertEqual(s_4scalar, s)
```
YenPeiChen07, line 36: Test four-scalar contractor with both positional and keyword arguments.

cpp/solvcon/universe/coord.hpp
```diff
@@ -908,6 +930,10 @@ using Segment3dFp64 = Segment3d<double>;
  * objects (one for each endpoint). The dimensionality (2 or 3) comes
  * from the underlying point pads.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. Plane code is meant to ignore it, not to
```
yungyuc: It is not entirely accurate to say "plane code".  I suppose you meant "2-dimensional code".  There is plane in 3-dimensional space too.  Please correct the wording.

yungyuc (changes requested):
LGTM.  One wording defect remains:

- [x] Write 2-dimensional when you meant 2-dimensional.  Do not say "plane".

cpp/solvcon/universe/coord.hpp
```diff
@@ -272,6 +276,10 @@ using Point3dFp64 = Point3d<double>;
  * three dimensions, z), so the layout is structure-of-arrays. The
  * dimensionality (2 or 3) is fixed at construction and cannot change.
  *
+ * A pad of ndim 2 stores and reports z as zero, and a segment or point read
+ * back from it carries that zero. 2-dimensional code is meant to ignore it, not
```
YenPeiChen07, line 280: Use "2-dimensional code" in 2D pad convention for accuracy instead of using "plan code".

yungyuc (approved):
LGTM

PR #1367, opened 2026-08-16
Use shared pi constant in RDomainWidget

### LINKED ISSUE #1001: Organize mathematical constants
PR #999 introduced [a local constant for PI in `RDomainWidget.cpp`](https://github.com/solvcon/solvcon/pull/999/changes#diff-d0469524b64b4f527d55b8915d5db9b12066fce6b541567418af68a408b728cfR34).  This issue is to track the follow-up work.  We should organize mathematic constant in a single location.  If local constants are needed, there should be a comment stating the reason.

OuChiaYun:
Use pi<float> from solvcon/math/math.hpp instead of defining a local PI in RDomainWidget.cpp, keeping the existing float-based angle calculations unchanged.

Related to #1001.


yungyuc (approved):
LGTM

Issue #1369, opened 2026-08-16
Add a Qt matmul benchmark inspector to Pilot


ThreeMonth03:
## Motivation

The matmul work in #1172 provides several legal kernels. Pilot needs a Qt inspector that compares their runtime and numerical differences from NumPy on the same exact operands before automatic dispatch is tuned.

The Qt page runs benchmarks in an isolated worker, shows the active kernel and elapsed time, and lets the user stop long work without blocking Pilot. This does not change automatic dispatch.

## Tasks

Keep each task and pull request at or below 500 changed lines, including tests.

- [x] **Task 1: Expose selectable kernels.** Add a keyword-only `matmul(..., kernel=...)` without changing Auto. Use Naive for the direct loop. (Resolved by #1370)
- [x] **Task 2: Split the matmul implementation.** Move non-template implementation out of `matmul.hpp` into `matmul.cpp`, keeping required templates in the header. (Resolved by #1387)
- [x] **Task 3: Define exact benchmark specs.** Validate dtype, full shapes, element strides, sampling, and kernel names without allocating arrays or imposing workload limits. (Resolved by #1394)
- [x] **Task 4a: Prepare one comparison.** Generate deterministic operands from a bounded FP64 component stream without a dtype-specific RNG table, preserving the requested shape, dtype, and element strides. Resolve each dtype through its registered pybind11 `SimpleArray<T>` class without copying operand storage. Classify unavailable kernels as `ineligible`, classify wrong shapes, wrong dtypes, and non-finite results as `invalid`, and record each `measured` result's maximum absolute and relative differences from NumPy. (Implemented by #1452)
- [x] **Task 4b: Collect one comparison.** Run every `measured` kernel and NumPy through the same balanced timing schedule, preserving raw repetition-block times and round order. (Implemented by #1433; stacked on #1452)
- [x] **Task 5: Store and execute benchmarks.** Add a strict artifact, atomic JSON I/O, and an isolated JSON-lines worker. (Implemented by #1487)
- [x] **Task 6: Control the worker from Qt.** Use `QProcess`, show the active kernel and elapsed time, and make Stop or worker failure terminate cleanly. (Implemented by #1497)
- [ ] **Task 7a: Build the Qt Route Inspector input form.** Add exact shape/stride controls, live sampling help, worker threads, kernel checkboxes, and input tooltips. Explain NumPy timing and numerical comparison, and provide Run/Stop controls. (Draft PR: #1516)
[...]

PR #1370, opened 2026-08-16
Expose selectable matmul kernels

ThreeMonth03:
`SimpleArray::matmul()` currently exposes only automatic selection, so benchmark work cannot compare a chosen kernel for the same operands. This change adds a keyword-only `kernel` argument directly to `matmul()`, while preserving automatic selection when the argument is omitted and for the `@` operator.

The Python boundary accepts kernel names as strings, while the C++ implementation uses `MatmulKernel` and rejects unknown or ineligible choices before allocating the output. Tests live with the existing matrix suite and cover float and complex dtypes, vector and matrix roles, strided and batched inputs, empty contractions, and keyword validation.

Related to #1369.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -1186,6 +1186,7 @@ public:
     }
 
     A matmul(A const & other) const;
+    A matmul(A const & other, MatmulKernel kernel) const;
     A & imatmul(A const & other);
 
 private:
@@ -1378,6 +1379,17 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
     A output(plan.output_shape());
     MatmulExecutor<A> executor(std::move(plan), output, *athis, other);
     executor.execute();
+    return output;
+}
+
+template <typename A, typename T>
+A SimpleArrayMixinCalculators<A, T>::matmul(A const & other, MatmulKernel kernel) const
+{
+    auto const * athis = static_cast<A const *>(this);
+    MatmulPlan plan = MatmulPlan::make(*athis, other);
+    A output(plan.output_shape());
+    MatmulExecutor<A> executor(std::move(plan), output, *athis, other);
+    executor.execute(kernel);
     return output;
 }
 
--- a/cpp/solvcon/buffer/matmul.hpp
+++ b/cpp/solvcon/buffer/matmul.hpp
@@ -21,6 +21,7 @@
 #include <ranges>
 #include <stdexcept>
 #include <string>
+#include <string_view>
 #include <type_traits>
 #include <utility>
 
@@ -46,7 +47,7 @@ inline constexpr bool use_matmul_blas_v = has_blas_backend && is_blas_element_v<
  */
 enum class MatmulKernel : std::uint8_t
 {
-    Generic,
+    Naive,
     BlasDot,
     BlasGevm,
     BlasGemv,
@@ -68,7 +69,7 @@ struct PackingState
 /**
  * @brief Group measured thresholds by matmul dispatch decision.
  *
- * A tuning table supplies the workload boundaries used to compare generic,
+ * A tuning table supplies the workload boundaries used to compare naive,
  * direct BLAS, and packing routes. Keeping the values separate from dispatch
  * code allows a backend or value type to provide another table.
  *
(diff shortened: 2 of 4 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -1271,6 +1272,17 @@ A SimpleArrayMixinCalculators<A, T>::matmul(A const & other) const
     return output;
 }
 
+template <typename A, typename T>
+A SimpleArrayMixinCalculators<A, T>::matmul(A const & other, MatmulKernel kernel) const
+{
+    auto const * athis = static_cast<A const *>(this);
+    MatmulPlan plan = MatmulPlan::make(*athis, other);
+    A output(plan.output_shape());
```
ThreeMonth03: [P1] Validate the forced kernel before allocating the output. An ineligible choice is currently rejected only by `executor.execute(kernel)`, after `A output(plan.output_shape())` has allocated the result. A large vector/matrix case forced to `blas_gemm`, for example, can allocate a large output before throwing. This also contradicts the PR description's pre-allocation validation claim. Please resolve eligibility before constructing `output`, or split validation from execution.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -426,8 +438,24 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
         }
 
         (*this)
-            .def("matmul", &wrapped_type::matmul)
-            .def("__matmul__", &wrapped_type::matmul)
+            .def(
+                "matmul",
```
ThreeMonth03, line 444: One Python API covers both modes:

```text
a @ b                    -> Auto
a.matmul(b)              -> Auto
a.matmul(b, kernel=name) -> force one kernel
```

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -653,27 +684,45 @@ MatmulExecutor<Array>::MatmulExecutor(MatmulPlan plan, Array & output, Array con
 
 template <typename Array>
 void MatmulExecutor<Array>::execute()
+{
+    run(select());
+}
+
+template <typename Array>
+void MatmulExecutor<Array>::execute(MatmulKernel kernel)
```
ThreeMonth03, line 692: Selection runs once:

```text
Auto                -> select()       -> kernel + packing -> run()
kernel="..."        -> select(kernel) -> kernel + packing -> run()
ineligible kernel   -> select(kernel) -> ValueError
```

cpp/solvcon/buffer/SimpleArray.hpp
ThreeMonth03, line 1390: Reject before reserving the result.

```text
current: plan -> allocate output -> select(kernel)
wanted:  plan -> select(kernel) -> allocate output
```

An ineligible large request should not allocate its result.

tests/test_matrix.py
```diff
@@ -107,14 +107,44 @@ def make_matrix_stride_cases(cls, data, axis):
             ('step_two', cls.make_strided_view(data, axis, 2)),
         )
 
-    def assert_matmul(self, lhs, rhs, expected):
-        result = lhs.matmul(rhs)
+    def assert_matmul(
+            self, lhs, rhs, expected, forced_kernels=()):
+        for kernel in (None, 'naive', *forced_kernels):
```
ThreeMonth03, line 112: The existing matrix cases cover the routes:

```text
every case -> Auto + Naive
matrix     -> BLAS GEMM (+ Winograd when eligible)
vector     -> BLAS DOT / GEVM / GEMV
```

ThreeMonth03 (commented):
@yungyuc Please take a look.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc: This is exactly why we chose to add explicit function API for array operations instead of merely relying on operator overload.

cpp/solvcon/buffer/matmul.hpp
```diff
@@ -46,7 +47,7 @@ inline constexpr bool use_matmul_blas_v = has_blas_backend && is_blas_element_v<
  */
 enum class MatmulKernel : std::uint8_t
 {
-    Generic,
+    Naive,
```
yungyuc, line 50: It's good to change to `Naive`.  It is much more specific.  `Generic` can imply any kind of kernel.

yungyuc (approved):
LGTM

PR #1371, opened 2026-08-17
Move nightly CI jobs into nightly- workflows

tigercosmos:
The nightly cron used to fire in four workflows, and two of them (`devbuild` and `lint`) mixed nightly-only jobs with the fast set that answers a pull request. Reading the Actions list gave no way to tell which run was which, and a job that ran only at night sat next to one that ran on every push.

This change puts every nightly job in a workflow whose name starts with `nightly-`, and leaves the cron out of everything else. `profiling` and `nouse_install` are renames. The sanitizer pair moves out of `devbuild` into `nightly-sanitizer`. The Windows Debug build and the portable artifact move into `nightly-build_windows`, which also builds Release, because the artifact needs a Release tree to package. `devbuild` keeps Windows Release for pull requests and master pushes. Every `nightly-` workflow accepts `workflow_dispatch`, and `nightly-build_windows` has no `SCGH_FORCE_*` variable, so a manual run is the way to exercise it from a fork.

`send_email_on_fail.yml` now has four callers, all nightly, each gated on the schedule event so that a force variable cannot mail from a pull request. `devbuild` and `lint` lose their mail job with the cron. Before, that job also mailed when a master push failed; now a broken master push mails no one, and only the next nightly (or a person reading the Actions list) reports it. That is a deliberate narrowing, so please push back if the master-push mail should stay.

`nouse_install` and the sanitizer jobs no longer consult `check_skip`, since a `SCGH_FORCE_*` variable is a request to run and a `skip-ci` label is a request not to. Only `devbuild` and `lint` read the label now.

The two sets are disjoint as a result, so the nightly is no longer a superset of the fast set. Nothing runs the Linux and macOS build, the pilot tests, or lint on the cron. That gives up the overnight canary against external drift, a new runner image or an updated dependency, which will now surface on the next pull request instead. A `nightly-devbuild` workflow would restore it if we want it back.

`doc/source/devguide/testing.md` describes the two sets, `tests/gui/README.md` follows (the full pytest now runs on a push only), and the README points at the new workflow for the portable Windows binary. The README no longer claims a direct download link on doc.solvcon.net, because the site does not carry one.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -4,8 +4,6 @@ on:
   push:
   pull_request:
     types: [opened, reopened, synchronize]
-  schedule:
-    - cron: '34 17 * * *'
 
 concurrency:
   # Key each master push on its commit SHA so every merge lands in its own
@@ -21,9 +19,6 @@ concurrency:
 jobs:
 
   check_skip:
-    # A fork inherits no SCGH_NIGHTLY, so this gate stops the nightly cron
-    # from allocating a runner there.
-    if: github.event_name != 'schedule' || (vars.SCGH_NIGHTLY || vars.MMGH_NIGHTLY) == 'enable'
     uses: ./.github/workflows/check_skip_ci.yml
     permissions:
       contents: read
@@ -39,8 +34,7 @@ jobs:
       github.event_name == 'pull_request' ||
       (github.event_name == 'push' &&
        ((vars.SCGH_PUSH_RUN_BRANCH || vars.MMGH_PUSH_RUN_BRANCH) == '*' ||
-        contains(vars.SCGH_PUSH_RUN_BRANCH || vars.MMGH_PUSH_RUN_BRANCH, github.ref_name))) ||
-      (github.event_name == 'schedule' && (vars.SCGH_NIGHTLY || vars.MMGH_NIGHTLY) == 'enable'))
+        contains(vars.SCGH_PUSH_RUN_BRANCH || vars.MMGH_PUSH_RUN_BRANCH, github.ref_name))))
 
     name: standalone_buffer_${{ matrix.os }}
 
@@ -84,9 +78,8 @@ jobs:
           create-symlink: true
           save: ${{ github.event_name != 'pull_request' }}
           # Bound the cache so it does not grow without limit across the rolling
-          # key. A master push or nightly run saves this cache; pull requests
-          # restore it (a PR on a non-default base branch cannot, so it runs
-          # cold).
+          # key. A master push saves this cache; pull requests restore it (a PR
+          # on a non-default base branch cannot, so it runs cold).
           max-size: 500M
 
       - name: make standalone_buffer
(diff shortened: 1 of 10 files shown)


.github/workflows/nightly-build_windows.yml
```diff
@@ -0,0 +1,249 @@
+name: nightly-build_windows
```
tigercosmos, line 1: Separate all nightly workflows to new `nightly-*` workflows.

.github/workflows/nightly-profiling.yml
```diff
@@ -1,4 +1,4 @@
-name: profiling
+name: nightly-profiling
```
tigercosmos, line 1: Rename the workflow with the `nightly-` prefix.

.github/workflows/devbuild.yml
```diff
@@ -21,9 +19,6 @@ concurrency:
 jobs:
 
   check_skip:
-    # A fork inherits no SCGH_NIGHTLY, so this gate stops the nightly cron
-    # from allocating a runner there.
-    if: github.event_name != 'schedule' || (vars.SCGH_NIGHTLY || vars.MMGH_NIGHTLY) == 'enable'
```
tigercosmos, line 26: Remove all nightly jobs from workflows for PRs.

.github/workflows/nightly-build_windows.yml
yungyuc: Workflow names should be either `all-hyphen` or `all_underscore`.  It is not good to mix hyphen and underscore.

yungyuc (commented):
It's good to separate nightly workflows to distinct files, but a large change like this should not be merged before being discussed.  @tigercosmos please leave a note stating that you want to merge early next time.

@ExplorerRay @chestercheng please take a look.

.github/workflows/devbuild.yml
```diff
@@ -4,8 +4,6 @@ on:
   push:
   pull_request:
     types: [opened, reopened, synchronize]
-  schedule:
-    - cron: '34 17 * * *'
```
chestercheng, line 7: Since `numpy` and `matplotlib` are not pinned version in CI, a new release could unexpectedly break the build.
Now that the nightly job no longer runs the dev build, we might not notice until the next PR.

It might be worth keeping a nightly dev-build job to preserve that coverage and catch compatibility issues earlier.

tigercosmos: @chestercheng  @yungyuc  I would like to discuss this. Since we are actively creating PRs these days. I think we don't need a nighlty job for dev-build. That is, we can always notice that the CI is failed while crearting PRs. What do you think?

chestercheng: Given the current activity, I think it’s ok to keep the nightly dev-build disabled for now.
We can keep this in mind if PR activity becomes lower in the future.

yungyuc: Having frequent PR differs from regular nightly.  We still should run the dev build nightly.   @tigercosmos please create an issue and add the schedule.

tigercosmos: I see. Let me add it back.

PR #1373, opened 2026-08-18
Use underscores only in the workflow names

tigercosmos:
yungyuc asked in #1371 that a workflow name not mix a hyphen with an underscore. Every workflow name now uses an underscore. The four nightly workflows read `nightly_build_windows`, `nightly_nouse_install`, `nightly_profiling`, and `nightly_sanitizer`. The file name and the `name:` in the Actions list both carry the new spelling.

Three other workflows carried a free-form title that did not match their file name. `cache cleanup`, `Send Email on Failure`, and `Update Contributors` now read `cache_cleanup`, `send_email_on_fail`, and `update_contributors`.

The job names keep their spelling. Each job name is a check-run name that branch protection matches on. A rename would drop a required check until someone updates the repository settings.

The README link, `doc/source/devguide/testing.md`, and the comments in `devbuild.yml` and `nightly_nouse_install.yml` follow the new names. No reusable-workflow path changed, because `check_skip_ci.yml` and `send_email_on_fail.yml` keep their file names.

Every workflow parses as YAML with the expected `name:`, every `uses: ./.github/workflows/...` path resolves, and `make lint` passes. The diff changes no code, so no test suite covers it.

Related to #1371.

### DIFF
--- a/.github/workflows/cache_cleanup.yml
+++ b/.github/workflows/cache_cleanup.yml
@@ -1,4 +1,4 @@
-name: cache cleanup
+name: cache_cleanup
 
 on:
   schedule:
--- a/.github/workflows/nightly_build_windows.yml
+++ b/.github/workflows/nightly_build_windows.yml
@@ -1,4 +1,4 @@
-name: nightly-build_windows
+name: nightly_build_windows
 
 on:
   schedule:
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: nightly-build-windows
+  group: nightly_build_windows
   cancel-in-progress: false
 
 jobs:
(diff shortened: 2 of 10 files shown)


.github/workflows/nightly_build_windows.yml
```diff
@@ -1,4 +1,4 @@
-name: nightly-build_windows
+name: nightly_build_windows
```
tigercosmos, line 1: Fix the naming issue.

.github/workflows/cache_cleanup.yml
```diff
@@ -1,4 +1,4 @@
-name: cache cleanup
+name: cache_cleanup
```
tigercosmos, line 1: Make the naming follow the same underscore pattern.

Issue #1286, opened 2026-08-07
MCAP signal processing step 2: Add a columnar MCAP/CDR reader in a new BUILD_MCAP-gated mcap/ subsystem


tigercosmos:
Second of three issues adding signal-processing support to solvcon (first: time-series kernels; third: `track/DataFrame` composition). This issue adds a reader for ROS 2 MCAP recordings that decodes requested field paths directly into `SimpleArray` columns.

This is the feature that attacks both the CPU cost and the memory cost of working with recorded logs. Typical pure-Python MCAP/CDR decoding unpacks every field of every message through per-field dispatch and `struct.unpack_from`, and retaining a decoded topic as Python lists of `(timestamp, value)` tuples costs on the order of a hundred bytes per sample in boxed objects, which reaches gigabytes for an hour-long multi-topic log. A float64 sample in a `SimpleArray` column costs about 16 contiguous bytes, so the same hour lands in tens of MB, with the per-byte decode work done once in C++.

### Design constraints

- New top-level `cpp/solvcon/mcap/` subsystem, gated by `option(BUILD_MCAP ...)` following the `BUILD_QT`/`BUILD_METAL` precedent. Not under `inout/`: that subsystem is mesh ingestion, and MCAP's lz4/zstd dependencies want their own gate.
- Dependency arrow is `mcap/ -> buffer/` only. No edge to or from `timeseries/` in either direction, so `BUILD_MCAP=OFF` leaves the full signal-processing stack building and passing tests, and the kernels stay usable on non-MCAP data.
- IDL parsing stays in Python; only the compiled flat decode plan (field offsets, alignment ops, primitive types) runs in C++. Porting a full IDL grammar to C++ is explicitly not required for the payoff.
- Columnar extraction is numeric/bool only. Fields holding strings, enums-as-strings, or nested messages are rejected by the plan compiler with a clear error, and callers fall back to the raw `messages()` iterator for those.
- `SimpleCollector` is the accumulation buffer during decode (per-topic message counts are not trusted up front), handing its buffer to `SimpleArray` copy-free via `as_array()`.

### PRs (4, each under 500 changed lines)

[...]

PR #1381, opened 2026-08-19
Support MCAP: Add the BUILD_MCAP gate and its lz4 and zstd dependency

tigercosmos:
Add a BUILD_MCAP option that gates a coming MCAP recording reader, and locate the two compression libraries it needs. MCAP stores its message chunks compressed with lz4 or zstd, and no other part of solvcon needs either library, so the option keeps them out of a build that does not read recordings. It defaults OFF, so an existing build picks up no new hard dependency. Configuring with it ON finds the two libraries, defines SOLVCON_MCAP, and stops with the name of a helper script when either one is missing.

That helper, contrib/dependency/show-mcap-dependency.sh, installs nothing. It prints the package-manager commands for the detected platform and leaves the decision to the reader, because installing a dependency needs review and consent. This follows build-scdv.sh, which also never invokes the package manager.

The change also adds the generator for the MCAP test recordings and its four outputs, together under 32 KB. Nothing reads them in this pull request; they land here so the reader that consumes them stays reviewable on its own. Two more pull requests follow: the container reader, then the per-topic message iterator.

Verified by configuring and building with BUILD_MCAP=ON and with BUILD_MCAP=OFF, both of which succeed and find lz4 and zstd from Homebrew when the option is on; make lint is clean, and make pytest-fast reports 1690 passed, 343 skipped, 3 xfailed.

Related to #1286.


tests/data/make_mcap_fixtures.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""
+Regenerate the tiny MCAP recordings that tests/test_mcap.py reads.
```
tigercosmos, line 5: The script to generate testing mcap files.

CMakeLists.txt
```diff
@@ -108,6 +108,34 @@ if(BUILD_METAL)
     add_compile_options(-DSOLVCON_METAL)
 endif()
 
+# The mcap subsystem reads MCAP recordings, and lz4 and zstd decompress their
+# chunks. No other part of solvcon needs those two libraries, so the option
+# keeps them out of a build that does not read recordings. The option starts
+# OFF, so an existing build picks up no new hard dependency.
+option(BUILD_MCAP "build with MCAP recording support" OFF)
```
tigercosmos, line 115: New added `BUILD_MCAP` option.

CMakeLists.txt
```diff
@@ -108,6 +108,34 @@ if(BUILD_METAL)
...
+    find_path(LZ4_INCLUDE_DIR NAMES lz4frame.h
+        HINTS ${MCAP_PREFIX_HINTS} PATH_SUFFIXES include)
+    find_library(ZSTD_LIB NAMES zstd libzstd
+        HINTS ${MCAP_PREFIX_HINTS} PATH_SUFFIXES lib)
+    find_path(ZSTD_INCLUDE_DIR NAMES zstd.h
+        HINTS ${MCAP_PREFIX_HINTS} PATH_SUFFIXES include)
+    if(LZ4_LIB AND LZ4_INCLUDE_DIR AND ZSTD_LIB AND ZSTD_INCLUDE_DIR)
+        message(STATUS "Found lz4: ${LZ4_LIB}")
+        message(STATUS "Found zstd: ${ZSTD_LIB}")
+        add_compile_definitions(SOLVCON_MCAP)
```
tigercosmos, line 130: lz4 and zstd are required.

tigercosmos (commented):

Issue #1382, opened 2026-08-19
apt-get update stalls on the Canonical mirrors until the CI job times out


tigercosmos:
The `setup dependencies` step can stall inside `apt-get update` until the job runs out of time. Run 32232097750 shows it on `standalone_buffer_ubuntu-24.04`. The step started at 08:43:15 and printed its last line at 08:43:53. The runner then cancelled the job at 08:53:22, under the 10-minute limit set in `.github/workflows/devbuild.yml`.

https://github.com/solvcon/solvcon/actions/runs/32232097750/job/96009723227

## What the log shows

`http://azure.archive.ubuntu.com` returned `Ign` on every index file. apt fell through to `https://archive.ubuntu.com` and began `Get:5 https://archive.ubuntu.com/ubuntu noble-security InRelease`. It printed nothing more for 9 minutes and 29 seconds.

## Expected and actual

Expected: `apt-get update` finishes, or fails, well inside the job time limit.

Actual: `apt-get update` runs for 9.5 minutes with no output, and the runner cancels the job.

## Why the mirror list does not help

The runner image points apt at `/etc/apt/apt-mirrors.txt`. That file lists `azure.archive.ubuntu.com`, `archive.ubuntu.com`, and `security.ubuntu.com`. All three are Canonical hosts, so one Canonical outage removes every mirror at the same time.

## Why the apt timeout does not help

apt on noble already defaults `Acquire::Retries` to 3, and the http and https method timeout to 30 seconds. A stalled fetch should therefore fail within about two minutes and move on. It did not. The connection most likely stayed alive and delivered data slowly. apt applies the timeout per read, not to the whole transfer, so a slow trickle never trips it.

## Possible fixes

1. Add a mirror on infrastructure independent of Canonical. apt then has somewhere to go when Canonical fails. This covers the `Ign` case, but not the slow-trickle case.
2. Wrap each `apt-get update` in `timeout` and retry it. No single command can then consume the whole job. This covers the slow-trickle case.

PR #1383, opened 2026-08-19
Cap the CI apt-get update with a wall-clock timeout

tigercosmos:
apt applies its acquire timeout to a single read, not to the whole transfer. A mirror that stays connected and delivers data slowly therefore never trips it. On 2026-08-19 the Ubuntu mirrors degraded rather than failed, and one job measured 27.6 MB in 20 minutes 51 seconds. That is 22.1 kB/s against a healthy 732 kB/s. An update in that state runs until the job time limit cancels the job, and a cancelled job states no cause.

This adds `contrib/ci/apt-update.sh`, which caps each run with `timeout` and retries once, and routes every update in the Linux setup action through it. A degraded mirror now fails the step in minutes, with a message that names the attempt, instead of consuming the whole job. The retry also recovers a transient slowdown, because it opens fresh connections. `add-apt-repository` takes `-n` so the PPA step no longer runs an unbounded update of its own.

An earlier version of this branch also added a fourth apt mirror. The logs show that apt never reaches it, because apt advances its mirror list only when a mirror fails, and a slow mirror never fails. That commit is dropped.

The cap covers `apt-get update` only. A slow mirror can still stall an `apt-get install`, whose download needs the `-d` split before a timeout can bound it safely.

No automated test covers the script, so I checked it by hand. It returns 0 for a healthy `apt-get`, and 1 for one that always fails. It stops a run that hangs, it recovers when the second attempt succeeds, and `ATTEMPTS` overrides the attempt count.

Related to #1382.

### DIFF
--- a/.github/actions/setup_linux/action.yml
+++ b/.github/actions/setup_linux/action.yml
@@ -54,7 +54,7 @@ runs:
         echo "::group::dependency by apt"
         # To update the cmake version > 4, install the latest cmake by kitware apt repository.
         # Reference: https://apt.kitware.com/
-        sudo apt-get -y update
+        sudo contrib/ci/apt-update.sh -y
         sudo apt-get -qy install ca-certificates gpg wget
         test -f /usr/share/doc/kitware-archive-keyring/copyright || \
             wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | \
@@ -62,7 +62,7 @@ runs:
             sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
         echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ noble main' | \
             sudo tee /etc/apt/sources.list.d/kitware.list >/dev/null
-        sudo apt-get -qqy update
+        sudo contrib/ci/apt-update.sh -qq
         test -f /usr/share/doc/kitware-archive-keyring/copyright || \
             sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
         sudo apt-get -qy install kitware-archive-keyring
@@ -99,7 +99,7 @@ runs:
         # Install clang-tidy-22 from LLVM repository for full C++23 support
         wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
         echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-22 main" | sudo tee /etc/apt/sources.list.d/llvm.list
-        sudo apt-get -qqy update
+        sudo contrib/ci/apt-update.sh -qq
         sudo apt-get -qy install clang-tidy-22
         sudo ln -fs "$(which clang-tidy-22)" "/usr/local/bin/clang-tidy"
         echo "::endgroup::"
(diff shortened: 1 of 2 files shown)


.github/actions/setup_linux/action.yml
```diff
@@ -48,21 +48,59 @@ runs:
...
+        echo "::group::Add a fallback apt mirror"
+        # The runner image points apt at Canonical mirrors only. An outage
+        # there stops every apt-get in this job. The apt mirror method sorts
+        # an entry with no priority tag last. The mirror added here therefore
+        # serves only as a fallback.
+        # https://salsa.debian.org/apt-team/apt/-/blob/main/methods/mirror.cc
+        mirrors=/etc/apt/apt-mirrors.txt
+        if [ -f "$mirrors" ] ; then
+            grep -q mirrors.edge.kernel.org "$mirrors" || \
+                echo "http://mirrors.edge.kernel.org/ubuntu/" | sudo tee -a "$mirrors" >/dev/null
```
tigercosmos: Add the mirrors in the list.

.github/actions/setup_linux/action.yml
```diff
@@ -48,21 +48,59 @@ runs:
...
+        sudo tee /usr/local/bin/solvcon-apt-update >/dev/null <<'EOF'
+        #!/bin/bash
+        for attempt in 1 2 ; do
+            if [ "$attempt" -gt 1 ] ; then sleep 5 ; fi
+            timeout --kill-after=10 120 apt-get "$@" update && exit 0
+            echo "apt-get update attempt $attempt failed or timed out"
+        done
+        exit 1
+        EOF
+        sudo chmod +x /usr/local/bin/solvcon-apt-update
```
tigercosmos: Add a timeout wrapper for apt-update.

yungyuc: Is there a reason to write out a script on the fly instead of checking it in `contrib/`?

yungyuc (changes requested):
- [x] Clarify why write out the script on the fly.

.github/actions/setup_linux/action.yml
tigercosmos: I also think about it. Since it's a tiny script, I didn't make it a file. I can do it, though.

contrib/ci/apt-update.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
tigercosmos, line 1: Add a new `apt-update.sh` file.

.github/actions/setup_linux/action.yml
yungyuc: Explicit is better than implicit.  A standalone script makes it easier to debug.

contrib/ci/apt-update.sh
```diff
@@ -0,0 +1,36 @@
...
+#
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+#
+# Run `apt-get update` under a wall-clock cap and retry it. apt applies its
+# acquire timeout to one read, not to the whole transfer, so a mirror that
+# stays connected and trickles data never trips it, and the update runs until
+# the CI job hits its time limit. A retry opens fresh connections, which a
+# round-robin archive host can answer from a healthy backend. Driven by
+# .github/actions/setup_linux/action.yml, and runnable by hand:
```
yungyuc, line 11: It's good that this can be run from a terminal manually.  It helps testing.

yungyuc (approved):
LGTM

PR #1380, opened 2026-08-19
Add dedup_last and deriv to the timeseries subsystem

tigercosmos:
`dedup_last(times, values)` keeps the last sample of every group of equal timestamps and returns strictly increasing times, which is what `deriv` requires. It takes every typed SimpleArray value class.

`deriv(times, values)` is the backward difference and returns `(times[1:], derivatives)`. It takes the real-number classes only, so `SimpleArrayBool` and the complex classes raise `TypeError`. A repeated timestamp throws instead of dividing by zero. An integer difference is taken unsigned with the sign restored, so it is exact over the full 64-bit range and a falling unsigned signal gives a negative derivative where `SimpleArray::diff()` wraps. A `SimpleArrayFloat32` input gives `SimpleArrayFloat32` derivatives; every other class gives `SimpleArrayFloat64`. The returned timestamps are always `SimpleArrayUint64`.

`timeseries.hpp` gains the shared check for a times and values pair, and `validate_sorted()` gains a `strict` flag so a kernel can reject a repeated timestamp as well as a decreasing one. The bindings instantiate each kernel over a shared type list, so a kernel added later names the types it takes once.

A ghosted series has no settled meaning in this subsystem, so the shared check rejects an array that carries ghost elements, and a TODO records the open question.

This is the second of the two stacked PRs from PR 3 of the plan. The first, #1372, is merged.

`make lint` is clean and `make gtest` passes (244 passed). `make pytest-fast` passes (1707 passed, 343 skipped, 3 xfailed), of which `tests/test_timeseries.py` contributes 26. The three examples on the documentation page execute with their assertions passing. I did not run the `tests/gui/` lane, because the change touches no GUI code.

Related to #1285.

PR #1372, opened 2026-08-18
Add the timeseries subsystem with merge_sorted_unique

tigercosmos:
`cpp/solvcon/timeseries/` holds kernels over recorded time series: integer-nanosecond timestamps as `SimpleArrayUint64` and the values sampled at them. Python reaches them as the `solvcon.timeseries` submodule, so generic kernel names stay out of the top-level namespace. `SimpleArray`'s `validate_1d()` now takes the exception class and the scope name, so a kernel outside `SimpleArray` reuses it.

`merge_sorted_unique(*arrays)` merges sorted timestamp arrays into one array of the distinct timestamps, the union time grid for two signals sampled at different rates.

`make lint` is clean, `make pytest` passes (1690 passed, 542 skipped), and `make gtest` passes (244 passed).

This is the first of two stacked PRs from PR 3 of the plan; the next adds `dedup_last` and `deriv` on this branch.

Related to #1285.

PR #1384, opened 2026-08-19
Add movavg and held to the timeseries subsystem

tigercosmos:
`movavg(times, values, span)` averages a series over the trailing half-open window `(t - span, t]`, and `held(times, values, span)` reports whether a `SimpleArrayBool` series was true over that same window. Both answer at every timestamp of the input and return those timestamps beside their result, so a chain of kernels keeps one time base.

The window covers timestamps, not positions in the array, so samples that share a timestamp all weigh in it. `t - span` wraps in `uint64_t` near the head of the log, so the window test compares instead of subtracting. `held` stays false until a sample exists at or before `t - span`; that boundary sample carries the state into the window.

`movavg` carries one running sum in `double`. The sweep is linear and fast but not exact: rounding error accumulates over the whole series, and one non-finite sample makes every later mean NaN. The docstrings state the tradeoff, a test pins it, and a `TODO` records the exact suffix-sum sweep that confines both. The result-type alias `deriv_value_t` becomes `real_value_t`, because two kernels now return it.

`make lint` is clean, `make gtest` passes (244 passed), and `make pytest` passes (2270 passed, 14 skipped, 3 xfailed), of which `tests/test_timeseries.py` contributes 46. The five examples on the documentation page execute with their assertions passing.

This is PR 4 of the plan. PR 3 landed as #1372 and #1380.

Related to #1285.

### DIFF
--- a/cpp/solvcon/timeseries/timeseries.hpp
+++ b/cpp/solvcon/timeseries/timeseries.hpp
@@ -72,6 +72,21 @@ inline void validate_sorted(
     }
 }
 
+/// Throw `std::invalid_argument` when a trailing window has no length, which would leave it empty everywhere.
+inline void validate_span(uint64_t span, char const * op)
+{
+    if (0 == span)
+    {
+        throw std::invalid_argument(std::format("timeseries::{}(): span must be positive but is 0", op));
+    }
+}
+
+/// Report whether a timestamp is before the trailing half-open window `(t - span, t]`.
+inline bool is_before_window(uint64_t stamp, uint64_t t, uint64_t span)
+{
+    return span <= t && stamp <= t - span;
+}
+
 template <typename R, typename T>
 R subtract_exact(T x1, T x0)
 {
@@ -211,14 +226,15 @@ dedup_last(SimpleArray<uint64_t> const & times, SimpleArray<T> const & values)
 }
 
 /**
- * The element type `deriv()` returns.
+ * The result element type of a kernel that computes in floating point.
  *
- * @tparam T The value type of the input series.
+ * @tparam T The value type of the input series. A floating-point type keeps its own width; every other real
+ *         number type promotes to `double`.
  *
  * @ingroup group_numerics
  */
 template <typename T>
-using deriv_value_t = std::conditional_t<std::is_floating_point_v<T>, T, double>;
+using real_value_t = std::conditional_t<std::is_floating_point_v<T>, T, double>;
 
 /**
  * Differentiate a series by the backward difference `(x_i - x_{i-1}) / (t_i - t_{i-1})`.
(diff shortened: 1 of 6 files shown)


cpp/solvcon/timeseries/timeseries.hpp
```diff
@@ -267,6 +283,125 @@ deriv(SimpleArray<uint64_t> const & times, SimpleArray<T> const & values)
...
+ * @param span The window length in nanoseconds, which must be positive.
+ * @return @p times and one mean per sample. An empty series gives empty arrays.
+ * @throw std::invalid_argument An array that is not one-dimensional or that carries ghost elements, a length
+ *        mismatch, a decreasing timestamp, or a zero @p span.
+ *
+ * @ingroup group_numerics
+ */
+template <typename T>
+std::pair<SimpleArray<uint64_t>, SimpleArray<real_value_t<T>>>
+movavg(SimpleArray<uint64_t> const & times, SimpleArray<T> const & values, uint64_t span)
```
tigercosmos, line 305: Implementation of `movavg`.

cpp/solvcon/timeseries/timeseries.hpp
```diff
@@ -267,6 +283,125 @@ deriv(SimpleArray<uint64_t> const & times, SimpleArray<T> const & values)
...
+
+    ssize_t const nsample = times.shape(0);
+    SimpleArray<uint64_t> otimes = times.to_row_major();
+    SimpleArray<result_type> omean(nsample);
+
+    uint64_t const * const tsrc = times.logical_data();
+    ssize_t const tstep = times.stride(0);
+    T const * const vsrc = values.logical_data();
+    ssize_t const vstep = values.stride(0);
+    // TODO: this running sum is the fastest sweep but not an exact one; the caveat in the function comment
```
tigercosmos, line 322: TODO: Limitation of the floating number error while doing the sweep.

cpp/solvcon/timeseries/timeseries.hpp
```diff
@@ -267,6 +283,125 @@ deriv(SimpleArray<uint64_t> const & times, SimpleArray<T> const & values)
...
+ * @param values The booleans sampled at @p times, of the same length.
+ * @param span The window length in nanoseconds, which must be positive.
+ * @return @p times and one answer per sample. An empty series gives empty arrays.
+ * @throw std::invalid_argument An array that is not one-dimensional or that carries ghost elements, a length
+ *        mismatch, a decreasing timestamp, or a zero @p span.
+ *
+ * @ingroup group_numerics
+ */
+inline std::pair<SimpleArray<uint64_t>, SimpleArray<bool>>
+held(SimpleArray<uint64_t> const & times, SimpleArray<bool> const & values, uint64_t span)
```
tigercosmos, line 363: Implementation of `held`.

doc/source/compute/numerics/timeseries.md
```diff
@@ -63,26 +54,81 @@ assert t_u.ndarray.tolist() == [0, 1, 2]
...
+```
+
+The first sample has no predecessor, so `n` samples give `n - 1` derivatives
+and fewer than two samples give empty arrays. The output timestamps are
+strictly increasing, so a second `deriv` chains directly. The kernel takes
+the real-number classes only; `SimpleArrayBool` and the complex classes raise
+`TypeError`. A `SimpleArrayFloat32` input gives `SimpleArrayFloat32`
+derivatives, and every other class gives `SimpleArrayFloat64`.
+
+## Trailing Windows
```
tigercosmos, line 81: Doc.

PR #1386, opened 2026-08-20
MCAP 2: Read the container layer of an MCAP recording

tigercosmos:
Add an mcap subsystem that reads the container layer of an MCAP recording: the file magic, the footer, and the summary section. Reader reports the topics with their schema names, the schema of one topic, the log time range, and the chunk index count. It reads nothing else, so opening a recording costs the same whatever the length of the log.

The subsystem sits beside inout/ rather than inside it, because inout/ is mesh ingestion and MCAP brings lz4 and zstd along. It depends on buffer/ alone.

The summary is walked one record at a time, so what it costs follows its largest record and not its length, and a record the reader does not keep is stepped over rather than read. Reading a field is what validates it, so a record that stops before a field the specification requires is rejected rather than half-parsed. Fields added after the ones read here stay unread, which keeps this reader working against a newer writer. A file with no summary section, a summary offset at or past the footer, a summary ending inside a record header, and two records that state one id without agreeing are all rejected.

The channel of highest id names a topic that several channels carry, and schema() answers with that same channel, so the two cannot disagree. schema() carries a TODO, because the decode plan is its first caller and what it needs should settle the shape returned; the definition comes back as bytes, since a ROS 2 encoding states it as text while a binary one does not.

Related to #1286.

### DIFF
--- a/.github/actions/setup_linux/action.yml
+++ b/.github/actions/setup_linux/action.yml
@@ -66,9 +66,11 @@ runs:
         test -f /usr/share/doc/kitware-archive-keyring/copyright || \
             sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
         sudo apt-get -qy install kitware-archive-keyring
+        # liblz4-dev and libzstd-dev back BUILD_MCAP, which decompresses the
+        # chunks of an MCAP recording.
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ \
-            libopenblas-dev \
+            libopenblas-dev liblz4-dev libzstd-dev \
             python3 python3-dev python3-venv
         # The path of apt-get cmake is `/usr/bin/cmake`,
         # but the path of built-in cmake of runner image is `/usr/local/bin/cmake`.
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -226,6 +226,25 @@ jobs:
         make pytest ${JOB_MAKE_ARGS}
         echo "::endgroup::"
 
+    - name: make pytest BUILD_MCAP=ON
+      run: |
+        echo "::group::make pytest BUILD_MCAP=ON"
+        # BUILD_MCAP is off in every other lane, so the subsystem gets a lane
+        # of its own rather than riding on one that means something else.
+        # SOLVCON_MCAP reaches module.cpp alone, so this configure recompiles
+        # the subsystem and that file and takes the rest from the cache.
+        rm -f build/*/Makefile
+        make cmake \
+          VERBOSE=1 USE_CLANG_TIDY=OFF \
+          BUILD_QT=OFF BUILD_MCAP=ON \
+          CMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }} \
+          CMAKE_ARGS="-DPYTHON_EXECUTABLE=$(which python3)"
+        make buildext VERBOSE=1
+        # A silent OFF would leave every mcap test skipped and the lane green.
+        python3 -c "import solvcon; assert solvcon.HAS_MCAP == True"
+        make pytest VERBOSE=1 PYTEST_OPTS="-k mcap"
+        echo "::endgroup::"
+
     - name: make buildext BUILD_QT=ON
       run: |
         echo "::group::make buildext BUILD_QT=ON"
--- a/cpp/solvcon/mcap/mcap_reader.hpp
+++ b/cpp/solvcon/mcap/mcap_reader.hpp
@@ -1,0 +1,162 @@
+#pragma once
+
+/*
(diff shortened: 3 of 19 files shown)


.github/actions/setup_linux/action.yml
```diff
@@ -66,9 +66,11 @@ runs:
         test -f /usr/share/doc/kitware-archive-keyring/copyright || \
             sudo rm /usr/share/keyrings/kitware-archive-keyring.gpg
         sudo apt-get -qy install kitware-archive-keyring
+        # liblz4-dev and libzstd-dev back BUILD_MCAP, which decompresses the
+        # chunks of an MCAP recording.
         sudo apt-get -qy install \
             sudo curl git build-essential make cmake libc6-dev gcc g++ \
-            libopenblas-dev \
+            libopenblas-dev liblz4-dev libzstd-dev \
```
tigercosmos, line 73: The libaries for MCAP testing.

.github/workflows/devbuild.yml
```diff
@@ -226,6 +226,25 @@ jobs:
         make pytest ${JOB_MAKE_ARGS}
         echo "::endgroup::"
 
+    - name: make pytest BUILD_MCAP=ON
```
tigercosmos, line 229: During the development, make `BUILD_MCAP=ON` a standalone job. Consider to make `BUILD_MCAP` to default `ON` once all features are done.

cpp/solvcon/mcap/pymod/wrap_McapReader.cpp
```diff
@@ -0,0 +1,99 @@
...
+            //
+            ;
+
+        (*this)
+            .def_property_readonly("path", &wrapped_type::path)
+            .def("topics", &wrapped_type::topics)
+            .def("chunk_count", &wrapped_type::chunk_count)
+            .def("schema", &wrapped_type::schema, py::arg("topic"))
+            .def("time_range", &wrapped_type::time_range)
+            .def("has_time_range", &wrapped_type::has_time_range)
```
tigercosmos, line 82: pybind APIs.

cpp/solvcon/mcap/mcap_reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+ * summary section only: the schema, channel, chunk index, and statistics
+ * records.  No chunk is read, and the message payloads stay where they are.
+ *
+ * The file must carry a summary section.  Every writer that indexes its output
+ * writes one.  Without it, a reader would have to walk the whole recording to
+ * collect the topics.
+ *
+ * @ingroup group_inout
+ */
+class Reader
```
tigercosmos, line 89: The MCAP reader.

cpp/solvcon/mcap/mcap_reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    std::string m_path;
+    std::ifstream m_stream;
+    uint64_t m_file_size = 0;
+    // TODO: Replace the STL containers below once the buffer subsystem offers
+    // an id-keyed map and a collector of records; SimpleCollector holds
+    // fundamental types and these hold records (issue #1286).
+    std::map<uint16_t, SchemaRecord> m_schemas;
+    std::map<uint16_t, ChannelRecord> m_channels;
+    std::vector<ChunkIndexRecord> m_chunk_indices;
```
tigercosmos, line 152: TODO: replace STL.

tests/test_mcap.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    head = (MAGIC + _record(0x01, _text("") + _text("solvcon test"))
+            + _record(0x0F, struct.pack("<I", 0)))
+    summary = b"".join(records) + trailing
+    footer = _record(0x02, struct.pack("<QQI", len(head), 0, 0))
+    return head + summary + footer + MAGIC
+
+
+@unittest.skipUnless(sc.mcap.HAS_MCAP, "built without BUILD_MCAP")
+class McapReaderTB(unittest.TestCase):
```
tigercosmos, line 84: Some basic checks.

solvcon/mcap/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+mcap: read MCAP recordings into solvcon.
+
+The extension carries the subsystem only when it is configured with
+``BUILD_MCAP=ON``, because decompressing MCAP chunks needs lz4 and zstd.
+``HAS_MCAP`` is true only in such a build, and ``Reader`` exists only then.
+"""
+
+from .. import core
+
+HAS_MCAP = core.HAS_MCAP
```
yungyuc, line 14: MCAP support should be always on.  There should not be a switch `HAS_MCAP`.

We will need to correct the architecture.  A productive way is to remove all MCAP code and start over.

PR #1388, opened 2026-08-21
Add true_intervals and benchmarks to the timeseries subsystem

tigercosmos:
`true_intervals(times, values)` run-length encodes a `SimpleArrayBool` series into the intervals where it was true and returns a `SimpleArrayUint64` of shape `(nrun, 3)` holding `(start, end, duration)` per run. Each row spans the half-open `[start, end)`: a run starts at the sample that turns the series true and ends at the sample that turns it false again. The rows stay a `SimpleArray`, so the total duration is one column sum.

Two policies the plan left open are settled and pinned in tests. A run still open at the last sample ends there, where the log ends, and a run that also started there is kept with zero duration. A group of equal timestamps is read at its last sample only, as the other kernels resolve one, so a true that a false overrides in the same nanosecond starts no run. A test in `tests/test_buffer.py` pins that `SimpleArray.std()` defaults to the population form (`ddof=0`).

`profiling/profile_timeseries.py` times every kernel of the subsystem against the pure-Python loop it replaces on an hour-long log at 100 Hz with repeated stamps. On one machine the kernels run 22x to 92x faster; `true_intervals` takes about 0.9 ms for 360k samples.

`make lint` is clean, `make gtest` passes (244 passed), and `make pytest-fast` passes (2067 passed, 12 skipped, 3 xfailed), of which `tests/test_timeseries.py` contributes 54. The documentation example executes with its assertions passing.

This is PR 5 of the plan. PR 4 landed as #1384.

Related to #1285.

### DIFF
--- a/cpp/solvcon/timeseries/timeseries.hpp
+++ b/cpp/solvcon/timeseries/timeseries.hpp
@@ -402,6 +402,72 @@ held(SimpleArray<uint64_t> const & times, SimpleArray<bool> const & values, uint
     return {std::move(otimes), std::move(oheld)};
 }
 
+/**
+ * Run-length encode the true stretches of a boolean series into `(start, end, duration)` rows.
+ *
+ * A run starts at the timestamp of the sample that turns the series true and ends at the timestamp of the sample
+ * that turns it false again, so each row spans the half-open interval `[start, end)` and `duration = end - start`.
+ * A group of equal timestamps is read at its last sample only. A run still open at the last sample ends there,
+ * so it has zero duration when it also started there.
+ *
+ * @param times The sample timestamps in nanoseconds, non-decreasing.
+ * @param values The booleans sampled at @p times, of the same length.
+ * @return An array of shape `(nrun, 3)` whose columns are the start, the end, and the duration in nanoseconds.
+ *         A series that is never true gives shape `(0, 3)`.
+ * @throw std::invalid_argument An array that is not one-dimensional or that carries ghost elements, a length
+ *        mismatch, or a decreasing timestamp.
+ *
+ * @ingroup group_numerics
+ */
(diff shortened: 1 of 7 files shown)


cpp/solvcon/timeseries/timeseries.hpp
```diff
@@ -402,6 +402,72 @@ held(SimpleArray<uint64_t> const & times, SimpleArray<bool> const & values, uint
...
+ * @param times The sample timestamps in nanoseconds, non-decreasing.
+ * @param values The booleans sampled at @p times, of the same length.
+ * @return An array of shape `(nrun, 3)` whose columns are the start, the end, and the duration in nanoseconds.
+ *         A series that is never true gives shape `(0, 3)`.
+ * @throw std::invalid_argument An array that is not one-dimensional or that carries ghost elements, a length
+ *        mismatch, or a decreasing timestamp.
+ *
+ * @ingroup group_numerics
+ */
+inline SimpleArray<uint64_t> true_intervals(SimpleArray<uint64_t> const & times, SimpleArray<bool> const & values)
```
tigercosmos, line 422: Implementation of `true_intervals`.

doc/source/compute/numerics/timeseries.md
```diff
@@ -131,4 +131,31 @@ assert t_h.ndarray.tolist() == [0, 10, 20, 30, 40]
 assert brake_held.ndarray.tolist() == [False, True, False, False, True]
 ```
 
+## The `true_intervals` Function
```
tigercosmos, line 134: Document.

profiling/profile_timeseries.py
```diff
@@ -0,0 +1,198 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
```
tigercosmos, line 1: Profiler for time series functions.

Issue #428, opened 2024-09-24
Add __str__ and __repr__ for simple array


tigercosmos:
Add `__str__()` and `__repr__()` for simple array. Currently, it only prints the memory address.
```
>>> import _modmesh as mm
>>> a = mm.SimpleArray(10, "int32")
>>> repr(a)
'<_modmesh.SimpleArray object at 0x7fb28deae270>'
>>> print(a)
<_modmesh.SimpleArray object at 0x7fb28deae270>
```

yungyuc:
The two string representations of arrays need careful design.  Once made available, the results will be widely used in test cases and go beyond our control.

PR #1390, opened 2026-08-22
Add string representation methods for SimpleArray and SimpleArrayPlex

scottmcmaster:
This PR adds Python `__str__` and `__repr__` methods to the typed `SimpleArray` bindings and the dtype-erased `SimpleArray` interface, providing useful output from `print()`, `str()`, `repr()`, and interactive Python sessions.

The methods delegate to NumPy's established array formatting through the existing zero-copy ndarray view. Tests cover integer, boolean, complex, and dtype-erased arrays to ensure both representations remain consistent with NumPy.

Related to #428.

### DIFF
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -131,6 +131,14 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 })
             .def("__len__", &wrapped_type::size)
             .def(
+                "__str__",
+                [](wrapped_type & self)
+                { return py::str(to_ndarray(self)); })
+            .def(
+                "__repr__",
+                [](wrapped_type & self)
+                { return py::repr(to_ndarray(self)); })
+            .def(
                 "__getitem__",
                 [](wrapped_type const & self, ssize_t key)
                 { return self.at(key); })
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -130,6 +130,14 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                     return ret;
                 })
             .def("__len__", &wrapped_type::size)
+            .def(
+                "__str__",
+                [](wrapped_type & self)
+                { return py::str(to_ndarray(self)); })
```
tigercosmos, line 136: As @yungyuc pointed out [here](https://github.com/solvcon/solvcon/issues/428#issuecomment-2371031600), the string representation should be discussed before the implementation. Current `py::str(to_ndarray(self))` might oversimplify the problem.

tigercosmos (changes requested):
Thanks for the PR! Please re-design the representation and discuss the design with the team (on PR or on [Discord](https://discord.com/channels/730297880140578906/730393342818648204)).

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
yungyuc: It oversimplifies.  `SimpleArray` should support collector and GPU.  The string representation should take that into account.

We can discuss this.  @scottmcmaster why do you think delegating to numpy is a good idea?  Consider to take this to the discord as well.

yungyuc (commented):
This PR needs in-depth discussion.

Issue #1391, opened 2026-08-22
Generate .pyi type stubs for pybind11 extension modules


scottmcmaster:
The Python API includes classes, functions, and submodules exported from the _solvcon pybind11 extension. These work correctly at runtime, but static analysis tools such as Pylance/Pyright/VS Code cannot reliably infer the API exposed by the compiled .so files.

As a result, valid usages can show up as unknown modules, types, or attributes, reducing the usefulness of autocomplete, navigation, and type checking.

This can be fixed by generating .pyi stub files for the pybind11 bindings as part of the development/build workflow, then making sure they are imported as needed by the impl modules. The stubs would describe the Python-visible API of _solvcon and its submodules so editors and type checkers can understand the native bindings without executing them.

yungyuc:
It's great to have IDE recognize the Python extension API.  Does `.pyi` also work for the embedding mode?  Embedding is the primary mode for solvcon.

PR #1392, opened 2026-08-22
Generate pyi stubs for pybind classes for type hinting support

scottmcmaster:
This change adds a `pyi` build target backed by `pybind11-stubgen` and makes the default build generate an ignored `_solvcon` stub tree alongside the compiled extension. The extension-only build remains available through `make buildext`, and cleanup removes the generated stubs.

It also exposes extension-backed symbols to static analyzers through explicit or type-checking imports in the core, pilot, spacetime, and timeseries wrappers. Needed to mark Kalman filter constructor arguments keyword-only where necessary so the generated signatures are valid Python. These changes let VSCode and Pyright resolve types such as `Solver`, `RManager`, and `StaticMesh` without changing their runtime objects.

Related to #1391.


Makefile
```diff
@@ -107,7 +109,7 @@ else
 endif
 
 .PHONY: default
-default: buildext
+default: pyi
```
j8xixo12, line 112: In my opinion, we shouldn't modify the default target. Users should be able to decide whether to generate the `pyi` stubs. Since stub generation requires the additional `pybind_stubgen` dependency and is an optional developer aid rather than a mandatory build requirement, the default target should remain as minimal as possible.

cpp/solvcon/linalg/pymod/wrap_kalman_filter.cpp
```diff
@@ -84,6 +84,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapKalmanFilter
                 py::arg("x"),
                 py::arg("f"),
                 py::arg("b") = py::none(),
+                py::kw_only(),
```
j8xixo12, line 87: It seems the Python constructor for `KalmanFilter` does not follow Python's function signature rules: once a parameter has a default value, it cannot be followed by a positional parameter without a default value.
Adding `py::kw_only` seems to change the API by forcing the constructor arguments to be passed as keyword arguments. @ThreeMonth03  Do you have any thoughts on it?

yungyuc: Having the positional argument `h` to follow the keyword argument `b` was an existing defect that should be handled, but it should be handled separately from the stub addition.

Let's keep this PR free from behavior change.  Please help create an issue to keep track of the argument order and leave a comment here.

solvcon/pilot/_pilot_core.py
```diff
@@ -6,8 +6,28 @@
 Import C++ implementation.
 """
 
+from typing import TYPE_CHECKING
```
yungyuc, line 9: This does not seem to be related to the stub generation.  Don't change unrelated code in this PR.

If you have to change this, please explain in inline annotation.

solvcon/spacetime.py
```diff
@@ -7,38 +7,45 @@
...
 """
 
-
-try:
-    from _solvcon import spacetime as _impl  # noqa: F401
-except ImportError:
-    from ._solvcon import spacetime as _impl  # noqa: F401
-
-_toload = [
+from typing import TYPE_CHECKING
```
yungyuc, line 10: `spacetime` is deprecated and will be scheduled to be removed in the future.  Do not change its code.

yungyuc (changes requested):
@scottmcmaster Please leave inline annotation to explain what you want to do with the code.

Points to address:
- [ ] Don't change behaviors in this PR.
- [ ] Don't do TYPE_CHECKING in this PR.
- [ ] Don't change spacetime.

PR #1394, opened 2026-08-22
Define exact `matmul` benchmark specs

ThreeMonth03:
## Motivation

The Route Inspector needs a serializable specification of one exact matmul comparison before the benchmark worker allocates operands.

## Approach

Add immutable specification types with strict dictionary round trips. Validate supported dtypes, complete operand shapes, element strides, representable native and NumPy byte layouts, sampling counts, and selected kernel names. Derive contraction, broadcast, and output shapes without allocating arrays or imposing runtime workload limits.

Related to #1369.

### DIFF
--- a/solvcon/benchmark/__init__.py
+++ b/solvcon/benchmark/__init__.py
@@ -1,0 +1,16 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Expose common and operation-specific benchmark specification types."""
+
+from . import matmul
+from . import spec
+
+
+__all__ = [
+    'matmul',
+    'spec',
+]
+
+
+# vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/solvcon/benchmark/matmul.py
+++ b/solvcon/benchmark/matmul.py
@@ -1,0 +1,143 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Validate exact matmul benchmark specifications without allocating arrays."""
+
+import dataclasses
+
+from . import spec
+
+
+MATMUL_DTYPE_SIZES = {
+    'float32': 4,
+    'float64': 8,
+    'complex64': 8,
+    'complex128': 16,
+}
+MATMUL_DTYPES = tuple(MATMUL_DTYPE_SIZES)
+MATMUL_KERNELS = (
+    'naive',
+    'blas_dot',
+    'blas_gevm',
+    'blas_gemv',
+    'blas_gemm',
+    'winograd',
+)
+
+
+def _broadcast_shape(lhs, rhs):
+    result = []
+    for offset in range(1, max(len(lhs), len(rhs)) + 1):
+        left = lhs[-offset] if offset <= len(lhs) else 1
+        right = rhs[-offset] if offset <= len(rhs) else 1
+        if left == right:
+            extent = left
+        elif left == 1:
+            extent = right
+        elif right == 1:
+            extent = left
+        else:
+            raise spec.SpecError(
+                'matmul batch dimensions do not match')
+        result.append(extent)
+    return tuple(reversed(result))
+
+
+def _batch_shape(shape):
+    return () if len(shape) == 1 else shape[:-2]
+
+
+@dataclasses.dataclass(frozen=True)
+class MatmulSpec:
+    """Describe one exact comparison of matmul kernels."""
+
+    OPERATION = 'matmul'
+
+    lhs: spec.OperandSpec
+    rhs: spec.OperandSpec
+    dtype: str
+    sampling: spec.Sampling
+    kernels: tuple
+
+    def __post_init__(self):
+        if not isinstance(self.dtype, str) or self.dtype not in MATMUL_DTYPES:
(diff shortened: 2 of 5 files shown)


solvcon/benchmark/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Data contracts shared by operation benchmarks."""
+
+from . import matmul
```
ThreeMonth03, line 6: The package exposes common specification types separately from operation-specific rules. `spec` contains operand and sampling types, while each operation module owns its validation and derived data.

```text
benchmark
|-- spec    common operand and sampling types
`-- matmul  matmul-specific specification rules
```

solvcon/benchmark/request.py
```diff
@@ -0,0 +1,93 @@
...
+def _int_tuple(value, name, minimum=None):
+    if not isinstance(value, (list, tuple)):
+        raise RequestError(f'{name} must be an array')
+    return tuple(
+        _require_int(item, f'{name}[{index}]', minimum)
+        for index, item in enumerate(value)
+    )
+
+
+@dataclasses.dataclass(frozen=True)
```
ThreeMonth03, line 83: `OperandSpec` and `Sampling` contain only data shared by operation benchmarks. They omit dtype and kernel rules so future operations can reuse them without inheriting matmul assumptions.

```text
OperandSpec = shape + element strides
Sampling    = warmups + repetitions + rounds
```

solvcon/benchmark/matmul.py
```diff
@@ -0,0 +1,132 @@
...
+        result.append(extent)
+    return tuple(reversed(result))
+
+
+def _batch_shape(shape):
+    return () if len(shape) == 1 else shape[:-2]
+
+
+@dataclasses.dataclass(frozen=True)
+class MatmulRequest:
```
ThreeMonth03: `MatmulSpec` combines the common operand and sampling types with matmul-specific validation and derived shapes. A future operation can define its own spec without modifying these common types.

```text
OperandSpec + Sampling
          |
      MatmulSpec
       |-- contraction / broadcast
       `-- dtype / kernel names
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+        'rhs': {'shape': [3, 2], 'strides': [2, 1]},
+        'dtype': 'float64',
+        'sampling': {'warmups': 2, 'repetitions': 5, 'rounds': 2},
+        'kernels': ['naive', 'blas_gemm'],
+    }
+    data.update(updates)
+    return benchmark.matmul.MatmulRequest.from_dict(data)
+
+
+class OperandSpecTC(unittest.TestCase):
```
ThreeMonth03, line 24: The tests mirror the production boundary: common operand and sampling types are verified separately from matmul-specific rules. This keeps failures localized when more benchmark operations are added.

```text
OperandSpecTC + SamplingTC -> common spec types
MatmulSpecTC                -> matmul-specific rules
```

ThreeMonth03 (commented):

solvcon/benchmark/request.py
```diff
@@ -0,0 +1,136 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Types shared by operation benchmark requests."""
```
yungyuc, line 4: How is a benchmark request defined?  The word "request" sounds generic and I am not sure it is a good name.  I need to know more about what it does.

yungyuc (approved):
The code is OK, but I'd need information about why you name "benchmark request".  Please share the following information:
1. Reference links or documents for how the industry names it.
2. A clear description.  As formal as possible.

I can merge the code as we discuss.  Let me know what you want: (1) merge now and discuss or (2) discuss and then merge.

solvcon/benchmark/request.py
ThreeMonth03: There is no benchmark-specific standard behind the word `request`. Google Benchmark uses benchmark arguments and command-line options, while JMH passes a serializable `Options` object to its `Runner`. I used `request` in the request/response sense because Task 5 will send this JSON value to an isolated worker and receive one comparison.

Formally:

```text
MatmulRequest =
    operation
  + lhs(shape, element strides)
  + rhs(shape, element strides)
  + dtype
  + sampling(warmups, repetitions, rounds)
  + ordered kernels
```

It is an immutable, serializable input that completely specifies one collector invocation before allocation or execution. It contains no allocated arrays, runtime state, environment metadata, or results.

References: [Google Benchmark user guide](https://google.github.io/benchmark/user_guide.html), [JMH `Options`](https://javadoc.io/static/org.openjdk.jmh/jmh-core/1.37/org/openjdk/jmh/runner/options/Options.html), and [JSON-RPC Request object](https://www.jsonrpc.org/specification#request_object).

I prefer option (2), discuss before merge. If the protocol meaning is too indirect, I will rename it to `MatmulSpec` in this PR and update the stacked collector PR.

yungyuc: It is a little bit difficult for me to picture how a request for benchmark is responded.  RPC has request/response but the use case of benchmark here is definitely not an RPC.

It sounds good to rename it to "spec".  I like it.

yungyuc (approved):
LGTM.  I'll merge after fixing CI failure.

Issue #1393, opened 2026-08-22
Add Float16 support to SimpleArray


ThreeMonth03:
## Motivation

`SimpleArray` needs a portable IEEE 754 binary16 dtype for compact array storage, zero-copy interchange with NumPy `float16`, and half-precision input to numerical libraries. C++23 does not guarantee `std::float16_t`, and MSVC STL does not currently provide it.

The implementation keeps one public type and selects its conversion path at compile time:

```cpp
#if defined(__STDCPP_FLOAT16_T__)
using native_type = std::float16_t;
#elif defined(__FLT16_MANT_DIG__) && __FLT16_MANT_DIG__ == 11
using native_type = _Float16;
#else
// MSVC: use portable software conversion.
#endif
```

`solvcon::Float16` keeps the same two-byte binary16 representation on every path.

## Tasks

Keep each task and pull request at or below 500 changed lines, including tests.

- [x] **Task 1: Add the Float16 C++ dtype.** Implement two-byte binary16 storage with native conversion where available and a portable software fallback otherwise. Register `DataType::Float16`, and integrate construction, copying, alignment, destruction, and dispatch through `SimpleArray` and `SimpleArrayPlex`. (#1395)
- [x] **Task 2: Define scalar arithmetic and comparison.** Add Float16-to-Float16 `+`, `-`, `*`, `/`, compound assignments, `==`, and `<=>`. Evaluate arithmetic in FP32, round results to FP16, and use IEEE partial ordering for finite values, infinity, NaN, and signed zero. (#1396)
- [x] **Task 3: Add Python and NumPy integration.** Expose `SimpleArrayFloat16` and `SimpleCollectorFloat16`, add the PEP 3118 `e` buffer format and zero-copy NumPy views, and define separate conversion rules for typed NumPy elements, dynamic Python scalars, Float16, and Complex. (#1428)
- [x] **Task 4: Complete array operations and verification.** Reuse and verify element-wise operations, reductions, sorting, searching, matrix helpers, and matrix multiplication, then cover public C++ and Python behavior across supported toolchains. (#1460)

Accelerator kernels, automatic mixed precision, BLAS dispatch, and device-backed buffers belong in later issues.

## Acceptance

- `solvcon::Float16` provides two-byte IEEE 754 binary16 storage with native conversion where available and software conversion otherwise.
[...]

PR #1395, opened 2026-08-22
Add the `Float16` C++ array type

ThreeMonth03:
`SimpleArray` needs one portable IEEE 754 binary16 type across GCC, Clang, and MSVC. C++23 does not guarantee `std::float16_t`, and MSVC STL does not currently provide it.

The conversion path is selected at compile time:

```cpp
#if defined(__STDCPP_FLOAT16_T__)
using native_type = std::float16_t;
#elif defined(__FLT16_MANT_DIG__) && __FLT16_MANT_DIG__ == 11
using native_type = _Float16;
#else
// MSVC: use portable software conversion.
#endif
```

This change stores the same two-byte binary16 representation behind `solvcon::Float16` on every path, then integrates `DataType::Float16` and `SimpleArrayFloat16` with the existing array lifecycle and buffer management.

Related to #1393.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -3790,6 +3790,7 @@ using SimpleArrayUint8 = SimpleArray<uint8_t>;
 using SimpleArrayUint16 = SimpleArray<uint16_t>;
 using SimpleArrayUint32 = SimpleArray<uint32_t>;
 using SimpleArrayUint64 = SimpleArray<uint64_t>;
+using SimpleArrayFloat16 = SimpleArray<Float16>;
 using SimpleArrayFloat32 = SimpleArray<float>;
 using SimpleArrayFloat64 = SimpleArray<double>;
 using SimpleArrayComplex64 = SimpleArray<Complex<float>>;
@@ -3798,10 +3799,8 @@ using SimpleArrayComplex128 = SimpleArray<Complex<double>>;
 /**
  * Runtime element-type tag mirroring the scalar types a SimpleArray supports.
  *
- * Covers bool, the signed and unsigned 8- to 64-bit integers, the 32- and
- * 64-bit floats, and the 64- and 128-bit complex types. Converts to and from
- * its enum and a type string, and DataType::from<T>() maps a C++ type to its
- * tag.
+ * DataType::from<T>() maps a C++ type to its tag. The string constructor maps
+ * registered public type names to tags.
  *
  * @ingroup group_core
  */
@@ -3820,6 +3819,7 @@ public:
         Uint16,
         Uint32,
         Uint64,
+        Float16,
         Float32,
         Float64,
         Complex64,
--- a/cpp/solvcon/math/Float16.hpp
+++ b/cpp/solvcon/math/Float16.hpp
@@ -1,0 +1,273 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Portable IEEE 754 binary16 storage and conversion.
+ *
+ * @ingroup group_core
+ */
+
+#include <bit>
+#include <concepts>
+#include <cstdint>
+#include <limits>
+#if __has_include(<stdfloat>)
+#include <stdfloat>
+#endif
+#include <type_traits>
+
+#if __has_include(<stdfloat>) && defined(__STDCPP_FLOAT16_T__)
+#define SC_FLOAT16_NATIVE_TYPE std::float16_t
+#elif defined(__FLT16_MANT_DIG__) && __FLT16_MANT_DIG__ == 11
+#define SC_FLOAT16_NATIVE_TYPE _Float16
+#endif
+
+namespace solvcon
+{
+
+/**
+ * Two-byte IEEE 754 binary16 value with stable cross-platform type identity.
+ *
+ * Native compiler conversions follow the current rounding mode. Other
+ * toolchains use a bitwise round-to-nearest-even fallback.
+ *
+ * @ingroup group_core
+ */
+class Float16
+{
+
+public:
+
+    using storage_type = uint16_t;
+
+    Float16() = default;
+    Float16(Float16 const &) = default;
+    Float16(Float16 &&) = default;
+    Float16 & operator=(Float16 const &) = default;
+    Float16 & operator=(Float16 &&) = default;
+    ~Float16() = default;
+
+    /**
+     * Construct from a single-precision value.
+     *
+     * @param value Source value.
+     */
+    constexpr Float16(float value); // NOLINT(google-explicit-constructor)
+
+    /**
+     * Construct from a double-precision value.
+     *
+     * @param value Source value.
+     */
+    constexpr Float16(double value); // NOLINT(google-explicit-constructor)
+
+    /**
+     * Construct from an integral value.
+     *
+     * @tparam T Integral source type.
+     * @param value Source value.
+     */
+    template <std::integral T>
+    constexpr Float16(T value); // NOLINT(google-explicit-constructor)
+
+    /**
+     * Convert to a single-precision value.
+     *
+     * @return The represented value as ``float``.
+     */
+    explicit constexpr operator float() const { return decode(m_bits); }
+
+    /**
+     * Return the raw binary16 representation.
+     *
+     * @return The sign, exponent, and fraction bits.
+     */
+    constexpr storage_type bits() const { return m_bits; }
+
+    /**
+     * Construct directly from a raw binary16 representation.
+     *
+     * @param bits The sign, exponent, and fraction bits.
+     * @return A value containing the supplied representation.
+     */
+    static constexpr Float16 from_bits(storage_type bits);
+
+private:
+
+#ifdef SC_FLOAT16_NATIVE_TYPE
+    using native_type = SC_FLOAT16_NATIVE_TYPE;
+    static_assert(sizeof(native_type) == sizeof(storage_type));
+#endif
+
+    template <typename T>
+    static constexpr storage_type encode(T value);
+    static constexpr float decode(storage_type value_bits);
+    template <typename T>
+    static constexpr storage_type encode_fallback(T value);
+    static constexpr float decode_fallback(storage_type value_bits);
+
+    storage_type m_bits;
+
+}; /* end class Float16 */
+
+constexpr Float16 Float16::from_bits(storage_type bits) { return std::bit_cast<Float16>(bits); }
+
+template <typename T>
+constexpr Float16::storage_type Float16::encode(T value)
+{
+#ifdef SC_FLOAT16_NATIVE_TYPE
+    return std::bit_cast<storage_type>(static_cast<native_type>(value));
+#else
+    return encode_fallback(value);
+#endif
+}
+
+constexpr Float16::Float16(float value)
+    : m_bits(encode(value))
+{
+}
+
+constexpr Float16::Float16(double value)
+    : m_bits(encode(value))
+{
+}
+
+template <std::integral T>
+constexpr Float16::Float16(T value)
+    : Float16(static_cast<double>(value))
+{
+}
+
+constexpr float Float16::decode(storage_type value_bits)
+{
+#ifdef SC_FLOAT16_NATIVE_TYPE
+    return static_cast<float>(std::bit_cast<native_type>(value_bits));
+#else
+    return decode_fallback(value_bits);
+#endif
+}
+
+template <typename T>
+constexpr Float16::storage_type Float16::encode_fallback(T value)
+{
+    static_assert(sizeof(T) == sizeof(uint32_t) || sizeof(T) == sizeof(uint64_t));
+    static_assert(std::numeric_limits<T>::is_iec559);
+
+    using bits_type = std::conditional_t<sizeof(T) == sizeof(uint32_t), uint32_t, uint64_t>;
+    constexpr uint32_t frac_bits = std::numeric_limits<T>::digits - 1;
+    constexpr uint32_t exp_bits = sizeof(T) * 8 - frac_bits - 1;
+    constexpr bits_type exp_mask = (bits_type{1} << exp_bits) - 1;
+    constexpr bits_type frac_mask = (bits_type{1} << frac_bits) - 1;
+    constexpr int32_t exp_bias = (1 << (exp_bits - 1)) - 1;
+
+    auto const value_bits = std::bit_cast<bits_type>(value);
+    uint32_t const sign = static_cast<uint32_t>(value_bits >> (frac_bits + exp_bits)) << 15;
+    bits_type const exponent = (value_bits >> frac_bits) & exp_mask;
+    bits_type const fraction = value_bits & frac_mask;
+
+    if (exponent == exp_mask)
+    {
+        if (fraction == 0)
+        {
+            return static_cast<storage_type>(sign | 0x7c00U);
+        }
+        bits_type const payload = (fraction >> (frac_bits - 10)) | 0x200U;
+        return static_cast<storage_type>(sign | 0x7c00U | payload);
+    }
+
+    int32_t half_exp = static_cast<int32_t>(exponent) - exp_bias + 15;
+    if (half_exp < -10)
+    {
+        return static_cast<storage_type>(sign);
+    }
+
+    if (half_exp <= 0)
+    {
+        bits_type const significand = fraction | (bits_type{1} << frac_bits);
+        auto const shift = static_cast<uint32_t>(static_cast<int32_t>(frac_bits) - 9 - half_exp);
+        bits_type rounded = significand >> shift;
+        bits_type const remainder = significand & ((bits_type{1} << shift) - 1);
+        bits_type const halfway = bits_type{1} << (shift - 1);
+        if (remainder > halfway || (remainder == halfway && (rounded & 1U)))
+        {
+            ++rounded;
+        }
+        return static_cast<storage_type>(sign | rounded);
+    }
+
+    if (half_exp >= 31)
+    {
+        return static_cast<storage_type>(sign | 0x7c00U);
+    }
+
+    constexpr uint32_t shift = frac_bits - 10;
+    bits_type rounded = fraction >> shift;
+    bits_type const remainder = fraction & ((bits_type{1} << shift) - 1);
+    bits_type const halfway = bits_type{1} << (shift - 1);
+    if (remainder > halfway || (remainder == halfway && (rounded & 1U)))
+    {
+        ++rounded;
+        if (rounded == 0x400U)
+        {
+            rounded = 0;
+            ++half_exp;
+        }
+    }
+    if (half_exp >= 31)
+    {
+        return static_cast<storage_type>(sign | 0x7c00U);
+    }
+    return static_cast<storage_type>(sign | (static_cast<uint32_t>(half_exp) << 10) | rounded);
+}
+
+constexpr float Float16::decode_fallback(storage_type value_bits)
+{
+    uint32_t const sign = static_cast<uint32_t>(value_bits & 0x8000U) << 16;
+    uint32_t const exponent = (value_bits >> 10) & 0x1fU;
+    uint32_t fraction = value_bits & 0x3ffU;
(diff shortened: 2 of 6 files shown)


cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include <bit>
+#include <concepts>
+#include <cstdint>
+#include <limits>
+#if __has_include(<stdfloat>)
+#include <stdfloat>
+#endif
+#include <type_traits>
+
+#ifdef __STDCPP_FLOAT16_T__
```
ThreeMonth03: One public type selects the best available conversion path:

```text
solvcon::Float16
    |-- std::float16_t when the library exposes it
    |-- _Float16 on supported compilers
    `-- software conversion otherwise
```

The stored representation stays two-byte IEEE binary16.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+    template <std::integral T>
+    constexpr Float16(T value) // NOLINT(google-explicit-constructor)
+        : Float16(static_cast<double>(value))
+    {
+    }
+
+    explicit constexpr operator float() const { return decode(m_bits); }
+
+    constexpr storage_type bits() const { return m_bits; }
+    static constexpr Float16 from_bits(storage_type bits);
```
ThreeMonth03, line 98: Numeric construction and raw-bit construction stay distinct:

```cpp
Float16(1).bits()            == 0x3c00;
Float16::from_bits(1).bits() == 0x0001;
```

The first encodes the number `1`; the second preserves the supplied bit pattern.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3790,6 +3790,7 @@ using SimpleArrayUint8 = SimpleArray<uint8_t>;
 using SimpleArrayUint16 = SimpleArray<uint16_t>;
 using SimpleArrayUint32 = SimpleArray<uint32_t>;
 using SimpleArrayUint64 = SimpleArray<uint64_t>;
+using SimpleArrayFloat16 = SimpleArray<Float16>;
```
ThreeMonth03, line 3793: `Float16` enters the existing array path without a separate container:

```text
Float16
   -> SimpleArray<Float16>
   -> SimpleArrayFloat16
   -> DataType::Float16
   -> SimpleArrayPlex lifecycle
```

gtests/test_nopython_buffer.cpp
```diff
@@ -25,6 +30,113 @@ TEST(ConcreteBuffer, iterator)
...
+
+    static_assert(sizeof(sc::Float16) == 2);
+    static_assert(std::is_trivially_copyable_v<sc::Float16>);
+    static_assert(std::is_convertible_v<float, sc::Float16>);
+    static_assert(std::is_convertible_v<double, sc::Float16>);
+    static_assert(std::is_convertible_v<int, sc::Float16>);
+    static_assert(!std::is_convertible_v<sc::Float16, float>);
+}
+
+TEST(Float16, binary_layout)
```
ThreeMonth03, line 46: The tests start from one readable field mapping:

```text
float32 1.5: 0 | 01111111 | 10000000000000000000000
float16 1.5: 0 | 01111    | 1000000000
```

Separate cases cover limits, ties-to-even, direct `double`, raw bits, and special values.

ThreeMonth03 (commented):

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+    bits_type const fraction = value_bits & frac_mask;
+
+    if (exponent == exp_mask)
+    {
+        if (fraction == 0)
+        {
+            return static_cast<storage_type>(sign | 0x7c00U);
+        }
+        bits_type const payload = (fraction >> (frac_bits - 10)) | ((fraction >> (frac_bits - 10)) == 0);
+        return static_cast<storage_type>(sign | 0x7c00U | payload);
```
tigercosmos: The fallback does not set the binary16 quiet bit (0x200), so it can return a signaling NaN and it disagrees with the native path.

Float `0x7f802000` encodes to `0x7e01` natively and to `0x7c01` here. `0x7c01` is a signaling NaN. 8388606 of the 16777214 float NaN inputs encode differently between the two paths. IEEE 754 requires a conversion to return a quiet NaN, so this also breaks the guarantee the PR is built on: the same two bytes on every platform.

```suggestion
        return static_cast<storage_type>(sign | 0x7e00U | (fraction >> (frac_bits - 10)));
```

The `== 0` guard is no longer necessary, because 0x200 keeps the payload non-zero. Infinity does not reach this line, because the `fraction == 0` branch above returns first.

With this and the decode fix, the fallback matches the native conversion bit for bit. I checked all 2^32 float inputs, all 2^16 half decodes, and 20M random doubles.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+                --unbiased_exp;
+            }
+            fraction &= 0x3ffU;
+            result |= static_cast<uint32_t>(unbiased_exp + 127) << 23;
+            result |= fraction << 13;
+        }
+    }
+    else if (exponent == 0x1fU)
+    {
+        result |= 0x7f800000U | (fraction << 13);
```
tigercosmos: The same defect in the decode direction. Half `0x7c01` decodes to `0x7fc02000` natively and to `0x7f802000` here. 1022 of the 2046 half NaN patterns decode to a signaling float.

Set the quiet bit only when the fraction is non-zero. This branch also carries infinity, so an unconditional `0x7fc00000` turns both infinities into NaN.

```suggestion
        result |= 0x7f800000U | (fraction << 13);
        if (fraction != 0)
        {
            result |= 0x00400000U;
        }
```

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+    constexpr Float16(float value) // NOLINT(google-explicit-constructor)
+        : m_bits(encode(value))
+    {
+    }
+
+    constexpr Float16(double value) // NOLINT(google-explicit-constructor)
+        : m_bits(encode(value))
+    {
+    }
```
tigercosmos: `constexpr` does not work here under Clang:

```
error: constexpr variable 'one' must be initialized by a constant expression
note: undefined function 'encode<float>' cannot be used in a constant expression
```

Both constructors sit in the class body, where `encode` is only declared. Its definition is at line 95. That places the instantiation of `encode<float>` before its definition. Apple Clang 21 rejects the code and GCC 15 accepts it. No code in this PR builds a `Float16` in a constant expression, so CI stays green and the `constexpr` does nothing on macOS. Task 2 hits this at the first `static_assert` over a comparison.

```suggestion
    constexpr Float16(float value); // NOLINT(google-explicit-constructor)
    constexpr Float16(double value); // NOLINT(google-explicit-constructor)
```

Then define them after `encode`, below line 103:

```cpp
constexpr Float16::Float16(float value)
    : m_bits(encode(value))
{
}

constexpr Float16::Float16(double value)
    : m_bits(encode(value))
{
}
```

The integral constructor needs no change. It delegates to the `double` one, and a template instantiates at the call site. After the change, both compilers evaluate `Float16(1.0F)`, `Float16(5)`, and overflow to infinity at compile time.

Moving the four conversion helpers into `namespace detail`, defined ahead of the class, removes the ordering hazard for good. That is a larger change, so the patch above is the smaller path.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+#if __has_include(<stdfloat>)
+#include <stdfloat>
+#endif
+#include <type_traits>
+
+#ifdef __STDCPP_FLOAT16_T__
+#define SC_FLOAT16_NATIVE_TYPE std::float16_t
+#elif defined(__FLT16_MANT_DIG__) && __FLT16_MANT_DIG__ == 11
+#define SC_FLOAT16_NATIVE_TYPE _Float16
+#endif
```
tigercosmos: Line 19 guards the include with `__has_include(<stdfloat>)`, but this test uses the predefined macro alone. The compiler defines the macro and the library provides the header. A Clang built against a libstdc++ older than 13 takes this branch, finds no header, and fails to compile.

```suggestion
#if defined(__STDCPP_FLOAT16_T__) && __has_include(<stdfloat>)
#define SC_FLOAT16_NATIVE_TYPE std::float16_t
#elif defined(__FLT16_MANT_DIG__) && __FLT16_MANT_DIG__ == 11
#define SC_FLOAT16_NATIVE_TYPE _Float16
#endif
```

For the PR body: Apple Clang 21 and upstream Clang 22.1.6 both leave `__STDCPP_FLOAT16_T__` undefined, and libc++ ships no `<stdfloat>`. The first branch is GCC-only today, the `_Float16` branch carries macOS, and MSVC alone needs the software fallback.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+namespace solvcon
+{
+
+/**
+ * Two-byte IEEE 754 binary16 value with stable cross-platform type identity.
+ *
+ * Native compiler types provide conversion when available. Other toolchains
+ * use a bitwise round-to-nearest-even fallback.
+ */
```
tigercosmos, line 40: The doc comment says the fallback rounds to nearest even. The native path does not: it follows the current rounding mode. The two paths still disagree outside the default environment.

```
FE_TONEAREST   native=3c00  fallback=3c00
FE_UPWARD      native=3c01  fallback=3c00
FE_DOWNWARD    native=3c00  fallback=3c00
FE_TOWARDZERO  native=3c00  fallback=3c00
```

solvcon never calls `fesetround`, so nothing hits this today. Please say so in the comment, because "the same representation on every path" is the property this class exists to provide.

gtests/test_nopython_buffer.cpp
```diff
@@ -25,6 +30,113 @@ TEST(ConcreteBuffer, iterator)
...
+    sc::Float16 const neg_inf(-std::numeric_limits<float>::infinity());
+    EXPECT_EQ(0x7c00U, pos_inf.bits());
+    EXPECT_EQ(0xfc00U, neg_inf.bits());
+    EXPECT_TRUE(std::isinf(static_cast<float>(pos_inf)));
+    EXPECT_TRUE(std::signbit(static_cast<float>(neg_inf)));
+
+    sc::Float16 const quiet_nan(std::numeric_limits<float>::quiet_NaN());
+    EXPECT_EQ(0x7c00U, quiet_nan.bits() & 0x7c00U);
+    EXPECT_NE(0U, quiet_nan.bits() & 0x03ffU);
+    EXPECT_TRUE(std::isnan(static_cast<float>(quiet_nan)));
```
tigercosmos: These assertions accept any non-zero payload, so they pass on both the native result `0x7e00` and the fallback result `0x7c01`. That is why the NaN defect reached green CI on all three platforms.

```suggestion
    sc::Float16 const quiet_nan(std::numeric_limits<float>::quiet_NaN());
    EXPECT_EQ(0x7e00U, quiet_nan.bits());
    EXPECT_TRUE(std::isnan(static_cast<float>(quiet_nan)));

    // A conversion must quiet a signaling NaN on every path.
    sc::Float16 const signaling_nan(std::numeric_limits<float>::signaling_NaN());
    EXPECT_EQ(0x0200U, signaling_nan.bits() & 0x0200U);
    EXPECT_TRUE(std::isnan(static_cast<float>(signaling_nan)));
```

`signaling_NaN()` reaches `encode` intact on this toolchain. The native path returns `0x7f00`, so these assertions pass on GCC and Clang and fail only on the current MSVC path, which returns `0x7d00`.

gtests/test_nopython_buffer.cpp
```diff
@@ -197,10 +331,39 @@ TEST(SimpleArray_DataType, from_string)
...
 }
 
+TEST(SimpleArrayPlex, float16_lifecycle)
+{
+    namespace sc = solvcon;
+
+    sc::small_vector<ssize_t> const shape{16};
+    sc::SimpleArrayPlex plex(shape, sc::DataType::Float16, 32);
+    EXPECT_EQ(sc::DataType::Float16, plex.data_type());
+    EXPECT_EQ(32, plex.alignment());
```
tigercosmos: The test reads back the stored alignment, not the address. It passes even if the buffer comes back unaligned. One more line closes that:

```cpp
    EXPECT_EQ(0U, reinterpret_cast<uintptr_t>(array->data()) % 32U);
```

gtests/test_nopython_buffer.cpp
```diff
@@ -197,10 +331,39 @@ TEST(SimpleArray_DataType, from_string)
     solvcon::DataType dt_bool = solvcon::DataType("bool");
     EXPECT_EQ(dt_bool.type(), solvcon::DataType::Bool);
 
-    EXPECT_THROW(solvcon::DataType("float16"), std::invalid_argument); // float16 does not exist
+    EXPECT_THROW(solvcon::DataType("float16"), std::invalid_argument);
```
tigercosmos, line 340: Minor. The deleted `// float16 does not exist` comment was correct before this PR and is wrong now. A short replacement keeps the expectation self-explanatory next to the `bool8` line:

```suggestion
    EXPECT_THROW(solvcon::DataType("float16"), std::invalid_argument); // float16 has no type string yet
```

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,228 @@
...
+    constexpr Float16(double value) // NOLINT(google-explicit-constructor)
+        : m_bits(encode(value))
+    {
+    }
+
+    template <std::integral T>
+    constexpr Float16(T value) // NOLINT(google-explicit-constructor)
+        : Float16(static_cast<double>(value))
+    {
+    }
```
tigercosmos: Minor, no change needed. `std::integral` accepts `bool` and the character types, so `Float16('a')` is 97.0 on an ASCII platform. Flagging it so the breadth is deliberate.

tigercosmos (changes requested):
**Fix before merge**

- [ ] `Float16.hpp:138` Set the quiet bit in `encode_fallback`. The path returns a signaling NaN today.
- [ ] `Float16.hpp:211` Set the quiet bit in `decode_fallback`, and leave infinity alone.
- [ ] `Float16.hpp:53` Define both converting constructors after `encode`. Clang rejects them in a constant expression.

**Worth fixing**

- [ ] `Float16.hpp:24` Test `__has_include(<stdfloat>)` together with the macro.
- [ ] `Float16.hpp:33` Record in the comment that the native path follows the current rounding mode.
- [ ] `test_nopython_buffer.cpp:129` Assert exact NaN bits, and add a signaling NaN input.
- [ ] `test_nopython_buffer.cpp:345` Check the address of the buffer, not the stored alignment.

**Minor**

- [ ] `test_nopython_buffer.cpp:334` Replace the deleted comment on the `"float16"` expectation.
- [ ] `Float16.hpp:63` No change needed. `std::integral` admits `bool` and the character types.

cpp/solvcon/math/Float16.hpp
ThreeMonth03: `encode_fallback()` now always sets the binary16 quiet bit while preserving the sign and payload. Exact quiet and signaling NaN patterns are covered by the test.

cpp/solvcon/math/Float16.hpp
ThreeMonth03: `decode_fallback()` now adds the float quiet bit only when the half fraction is non-zero, so signaling NaNs are quieted and infinities remain infinities.

cpp/solvcon/math/Float16.hpp
ThreeMonth03: The converting constructors are now declared in the class and defined after `encode()`. A constexpr `static_assert` exercises construction, and the macOS Clang CI passes.

cpp/solvcon/math/Float16.hpp
ThreeMonth03: The `std::float16_t` branch now requires both `__has_include(<stdfloat>)` and `__STDCPP_FLOAT16_T__`. macOS continues through `_Float16`.

cpp/solvcon/math/Float16.hpp
ThreeMonth03: The class comment now distinguishes native current-rounding-mode conversions from fallback round-to-nearest-even conversions.

gtests/test_nopython_buffer.cpp
ThreeMonth03: The test now uses explicit quiet and signaling float NaN bit patterns and asserts the exact half encodings and decoded float bits.

gtests/test_nopython_buffer.cpp
ThreeMonth03: The test now checks `reinterpret_cast<std::uintptr_t>(array->data()) % 32U` instead of reading back the requested alignment value.

gtests/test_nopython_buffer.cpp
ThreeMonth03: The nearby comment now states the current invariant: `Float16` has no registered string name.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3820,6 +3819,7 @@ class DataType
         Uint16,
         Uint32,
         Uint64,
+        Float16,
```
yungyuc, line 3822: `DataType::enum_type` proves to be a good design.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,274 @@
...
+}; /* end class Float16 */
+
+constexpr Float16 Float16::from_bits(storage_type bits) { return std::bit_cast<Float16>(bits); }
+
+template <typename T>
+constexpr Float16::storage_type Float16::encode(T value)
+{
+#ifdef SC_FLOAT16_NATIVE_TYPE
+    auto const native = static_cast<native_type>(value);
+    return std::bit_cast<storage_type>(native);
```
yungyuc: The `decode()` writes:
```cpp
    return static_cast<float>(std::bit_cast<native_type>(value_bits));
```

Making it a symmetrical line reads better:

```suggestion
    return std::bit_cast<storage_type>(static_cast<native_type>(value));
```

I wonder if there is a reason to use two lines?

cpp/solvcon/math/Float16.hpp
yungyuc: @ThreeMonth03 Please create an issue to add all rounding mode in the fallback.  It's of low priority but good to have.

yungyuc (changes requested):
I like the code.  It's what we need.  Polishing can happen with this and followup PRs.

- [ ] Create an issue for adding all rounding modes in fallback.
- [ ] Consider to make encode/decode code reading symmetrical.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,273 @@
...
+#endif
+
+namespace solvcon
+{
+
+/**
+ * Two-byte IEEE 754 binary16 value with stable cross-platform type identity.
+ *
+ * Native compiler conversions follow the current rounding mode. Other
+ * toolchains use a bitwise round-to-nearest-even fallback.
```
ThreeMonth03, line 37: Issue #1404 tracks runtime fallback support for all four rounding modes. Constant evaluation remains round-to-nearest, ties-to-even, and floating-point exception flags stay out of scope.

cpp/solvcon/math/Float16.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+}; /* end class Float16 */
+
+constexpr Float16 Float16::from_bits(storage_type bits) { return std::bit_cast<Float16>(bits); }
+
+template <typename T>
+constexpr Float16::storage_type Float16::encode(T value)
+{
+#ifdef SC_FLOAT16_NATIVE_TYPE
+    return std::bit_cast<storage_type>(static_cast<native_type>(value));
```
ThreeMonth03, line 124: `encode()` now mirrors `decode()` with one conversion expression. The removed temporary did not carry separate semantics.

ThreeMonth03 (commented):
> - [x] Create an issue for adding all rounding modes in fallback (#1404).
> - [x] Make the native encode/decode paths read symmetrically.

tigercosmos (approved):
LGTM, thanks!

PR #1396, opened 2026-08-22
Add `Float16` scalar operations

ThreeMonth03:
`solvcon::Float16` needs scalar arithmetic and comparison before array-level operations can reuse it.

This change adds Float16-to-Float16 arithmetic, compound assignments, equality, and three-way comparison. Arithmetic is evaluated in FP32 and rounded back to FP16. Comparisons use IEEE partial ordering for finite values, infinity, NaN, and signed zero.

The underlying `Float16` array type was added in #1395.

Related to #1393.

### DIFF
--- a/contrib/standalone_buffer/Makefile
+++ b/contrib/standalone_buffer/Makefile
@@ -69,6 +69,9 @@ $(SETUPROOT)/ConcreteBuffer.o: $(SETUPROOT)/solvcon/buffer/ConcreteBuffer.cpp Ma
 $(SETUPROOT)/SimpleArray.o: $(SETUPROOT)/solvcon/buffer/SimpleArray.cpp Makefile
 	c++ $(CFLAGS) -c $< -o $@
 
+$(SETUPROOT)/Float16.o: $(SETUPROOT)/solvcon/math/Float16.cpp Makefile
+	c++ $(CFLAGS) -frounding-math -c $< -o $@
+
 $(SETUPROOT)/matmul.o: $(SETUPROOT)/solvcon/buffer/matmul.cpp Makefile
 	c++ $(CFLAGS) -c $< -o $@
 
@@ -88,6 +91,7 @@ OBJS = \
 	$(SETUPROOT)/BufferExpander.o \
 	$(SETUPROOT)/ConcreteBuffer.o \
 	$(SETUPROOT)/SimpleArray.o \
+	$(SETUPROOT)/Float16.o \
 	$(SETUPROOT)/matmul.o \
 	$(SETUPROOT)/blas_compat.o \
 	$(SETUPROOT)/wrap_ConcreteBuffer.o \
--- a/cpp/solvcon/CMakeLists.txt
+++ b/cpp/solvcon/CMakeLists.txt
@@ -132,6 +132,10 @@ add_library(
     STATIC
     ${SOLVCON_TERMINAL_FILES}
 )
+set_source_files_properties(
+    ${CMAKE_CURRENT_SOURCE_DIR}/math/Float16.cpp
+    PROPERTIES COMPILE_OPTIONS "${SOLVCON_FLOAT16_COMPILE_OPTIONS}"
+)
 
 set(SOLVCON_LIBRARIES solvcon_primary)
 
--- a/cpp/solvcon/math/Float16.cpp
+++ b/cpp/solvcon/math/Float16.cpp
@@ -9,17 +9,50 @@
 // Tell MSVC that calls in this translation unit observe the floating-point
 // environment changed by std::fesetround().
 #pragma fenv_access(on)
+#endif
+
+// Keep fenv-aware arithmetic out of non-strict callers so it observes the rounding mode selected by std::fesetround().
+#ifdef _MSC_VER
+#define SC_DECL_FLOAT16_NOINLINE __declspec(noinline)
+#elif defined(__clang__) || defined(__GNUC__)
+#define SC_DECL_FLOAT16_NOINLINE __attribute__((noinline))
+#else
+#define SC_DECL_FLOAT16_NOINLINE
+#endif
 
 namespace solvcon
 {
 
+#ifdef _MSC_VER
 // Prevent LTCG from inlining this fenv-aware query into a caller.
 __declspec(noinline) int Float16::rounding_mode()
 {
     return std::fegetround();
 }
+#endif
+
+SC_DECL_FLOAT16_NOINLINE Float16 Float16::add_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs + rhs));
+}
+
+SC_DECL_FLOAT16_NOINLINE Float16 Float16::sub_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs - rhs));
+}
+
+SC_DECL_FLOAT16_NOINLINE Float16 Float16::mul_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs * rhs));
+}
+
+SC_DECL_FLOAT16_NOINLINE Float16 Float16::div_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs / rhs));
+}
 
 } /* end namespace solvcon */
-#endif
+
+#undef SC_DECL_FLOAT16_NOINLINE
 
 // vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/math/Float16.hpp
+++ b/cpp/solvcon/math/Float16.hpp
@@ -14,6 +14,7 @@
 
 #include <bit>
 #include <cfenv>
+#include <compare>
 #include <concepts>
 #include <cstdint>
 #include <limits>
@@ -85,6 +86,11 @@ public:
      */
     explicit constexpr operator float() const { return decode(m_bits); }
 
+    constexpr Float16 & operator+=(Float16 other);
+    constexpr Float16 & operator-=(Float16 other);
+    constexpr Float16 & operator*=(Float16 other);
+    constexpr Float16 & operator/=(Float16 other);
+
     /**
      * Return the raw binary16 representation.
      *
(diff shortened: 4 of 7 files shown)


cpp/solvcon/math/Float16.hpp
```diff
@@ -204,6 +221,57 @@ constexpr bool Float16::should_round_away(
...
+    if consteval
+    {
+        return Float16(static_cast<float>(lhs) * static_cast<float>(rhs));
+    }
+    return Float16::mul_runtime(static_cast<float>(lhs), static_cast<float>(rhs));
+}
+
+constexpr Float16 operator/(Float16 lhs, Float16 rhs)
+{
+    if consteval
```
ThreeMonth03: Two paths are required. Constant evaluation cannot query the floating-point environment, so this branch keeps expressions such as `static_assert((Float16(1) / Float16(2)).bits() == 0x3800U)` valid with nearest-even behavior. Runtime uses the strict out-of-line helper so cases such as `Float16(1) / Float16(3)` honor `FE_UPWARD` and the other active modes.

cpp/solvcon/CMakeLists.txt
```diff
@@ -132,6 +132,10 @@ add_library(
     STATIC
     ${SOLVCON_TERMINAL_FILES}
 )
+set_source_files_properties(
+    ${CMAKE_CURRENT_SOURCE_DIR}/math/Float16.cpp
+    PROPERTIES COMPILE_OPTIONS "${SOLVCON_FLOAT16_COMPILE_OPTIONS}"
```
ThreeMonth03: These options are deliberately source-scoped. Only `Float16.cpp` and the fenv test need dynamic-rounding semantics; applying strict FP to the whole target would constrain unrelated numerical code.

cpp/solvcon/math/Float16.cpp
```diff
@@ -9,17 +9,52 @@
...
+}
+
+SC_DECL_FLOAT16_NOIPA Float16 Float16::mul_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs * rhs));
+}
+
+SC_DECL_FLOAT16_NOIPA Float16 Float16::div_runtime(float lhs, float rhs)
+{
+    return from_bits(encode_fallback(lhs / rhs));
```
ThreeMonth03: Returning the final `Float16` here is important. Returning `float` would move the FP16 narrowing into an ordinary caller, where the active rounding mode could be lost. The noinline/noipa boundary keeps the FP32 operation and final encoding inside this strict translation unit.

gtests/test_nopython_buffer.cpp
```diff
@@ -16,6 +17,26 @@
...
+
+template <typename Evaluate, typename Expected>
+void expect_rounding_result(int mode, Evaluate evaluate, Expected const & expected)
+{
+    std::fenv_t original_environment;
+    ASSERT_EQ(0, std::fegetenv(&original_environment));
+
+    int const mode_status = std::fesetround(mode);
+    auto const actual = evaluate();
+    int const restore_status = std::fesetenv(&original_environment);
```
ThreeMonth03: The order is intentional: evaluate under the requested mode, restore the original fenv, then call GTest. For example, `FE_DOWNWARD` must make `1 - 1` produce `0x8000`, but the assertion itself runs after restoration.

contrib/standalone_buffer/Makefile
```diff
@@ -66,6 +66,9 @@ $(SETUPROOT)/BufferExpander.o: $(SETUPROOT)/solvcon/buffer/BufferExpander.cpp Ma
 $(SETUPROOT)/SimpleArray.o: $(SETUPROOT)/solvcon/buffer/SimpleArray.cpp Makefile
 	c++ $(CFLAGS) -c $< -o $@
 
+$(SETUPROOT)/Float16.o: $(SETUPROOT)/solvcon/math/Float16.cpp Makefile
```
ThreeMonth03: The operators now reference out-of-line runtime helpers in `Float16.cpp`, so standalone consumers must compile and link this object. Without it, code using `+`, `-`, `*`, or `/` would fail at link time.

ThreeMonth03 (commented):

cpp/solvcon/math/Float16.hpp
ThreeMonth03, line 253: Two paths are required. Constant evaluation cannot query the floating-point environment, so this branch keeps expressions such as `static_assert((Float16(1) / Float16(2)).bits() == 0x3800U)` valid with nearest-even behavior. Runtime uses the strict out-of-line helper so cases such as `Float16(1) / Float16(3)` honor `FE_UPWARD` and the other active modes.

cpp/solvcon/CMakeLists.txt
ThreeMonth03, line 137: These options are deliberately source-scoped. Only `Float16.cpp` and the fenv test need dynamic-rounding semantics; applying strict FP to the whole target would constrain unrelated numerical code.

cpp/solvcon/math/Float16.cpp
ThreeMonth03, line 51: Returning the final `Float16` here is important. Returning `float` would move the FP16 narrowing into an ordinary caller, where the active rounding mode could be lost. The noinline/noipa boundary keeps the FP32 operation and final encoding inside this strict translation unit.

gtests/test_nopython_buffer.cpp
ThreeMonth03, line 31: The order is intentional: evaluate under the requested mode, restore the original fenv, then call GTest. For example, `FE_DOWNWARD` must make `1 - 1` produce `0x8000`, but the assertion itself runs after restoration.

contrib/standalone_buffer/Makefile
```diff
@@ -69,6 +69,9 @@ $(SETUPROOT)/ConcreteBuffer.o: $(SETUPROOT)/solvcon/buffer/ConcreteBuffer.cpp Ma
...
```
ThreeMonth03, line 72: The operators now reference out-of-line runtime helpers in `Float16.cpp`, so standalone consumers must compile and link this object. Without it, code using `+`, `-`, `*`, or `/` would fail at link time.

ThreeMonth03 (commented):

cpp/solvcon/math/Float16.hpp
```diff
@@ -204,6 +221,57 @@ constexpr bool Float16::should_round_away(
...
+
+constexpr Float16 & Float16::operator*=(Float16 other) { return *this = *this * other; }
+
+constexpr Float16 & Float16::operator/=(Float16 other) { return *this = *this / other; }
+
+constexpr bool operator==(Float16 lhs, Float16 rhs) { return static_cast<float>(lhs) == static_cast<float>(rhs); }
+
+constexpr std::partial_ordering operator<=>(Float16 lhs, Float16 rhs)
+{
+    return static_cast<float>(lhs) <=> static_cast<float>(rhs);
```
tigercosmos, line 272: This is fancy.

gtests/test_nopython_buffer.cpp
```diff
@@ -16,6 +17,26 @@
...
+{
+
+template <typename Evaluate, typename Expected>
+void expect_rounding_result(int mode, Evaluate evaluate, Expected const & expected)
+{
+    std::fenv_t original_environment;
+    ASSERT_EQ(0, std::fegetenv(&original_environment));
+
+    int const mode_status = std::fesetround(mode);
+    auto const actual = evaluate();
```
tigercosmos: Nit: I remember we have a convention that we don't use `auto` for functions without type infomation.

cpp/solvcon/math/Float16.cpp
```diff
@@ -9,17 +9,52 @@
 // Tell MSVC that calls in this translation unit observe the floating-point
 // environment changed by std::fesetround().
 #pragma fenv_access(on)
+#endif
+
+#ifdef _MSC_VER
```
yungyuc, line 15: Add code comment to briefly explain what does "IPA" or "NOIPA" mean (don't make maintainers guess).

If it's already there, leave review annotation to highlight it.

yungyuc (changes requested):
LGTM.  Nit for maintainability:

- [ ] Add code comment to explain "IPA".

gtests/test_nopython_buffer.cpp
yungyuc: Good catch.  Please remove the `auto` here.

cpp/solvcon/math/Float16.cpp
```diff
@@ -9,17 +9,50 @@
 // Tell MSVC that calls in this translation unit observe the floating-point
 // environment changed by std::fesetround().
 #pragma fenv_access(on)
+#endif
+
+// Keep fenv-aware arithmetic out of non-strict callers so it observes the rounding mode selected by std::fesetround().
```
ThreeMonth03, line 14: Renamed `NOIPA` to `NOINLINE` because the code only relies on the cross-compiler no-inline guarantee; GCC's `noipa` disables broader optimizations that are not required. The call boundary keeps fenv-aware arithmetic in the strict-FP translation unit.

gtests/test_nopython_buffer.cpp
```diff
@@ -16,6 +17,26 @@
...
+{
+
+template <typename Evaluate, typename Expected>
+void expect_rounding_result(int mode, Evaluate evaluate, Expected const & expected)
+{
+    std::fenv_t original_environment;
+    ASSERT_EQ(0, std::fegetenv(&original_environment));
+
+    int const mode_status = std::fesetround(mode);
+    Expected const actual = evaluate();
```
ThreeMonth03, line 30: Changed `auto` to `Expected` so the helper result type is explicit, following @tigercosmos's suggestion.

ThreeMonth03 (commented):
@yungyuc Please review the latest revision.

- [x] Replace the IPA/NOIPA terminology with `NOINLINE` and explain why fenv-aware arithmetic must stay out of non-strict callers.
- [x] Replace `auto const actual` with `Expected const actual`, as @tigercosmos suggested.

tigercosmos (approved):

yungyuc (approved):
LGTM

PR #1401, opened 2026-08-24
Walk the messages of one MCAP topic

tigercosmos:
`Reader::messages()` returns an iterator over the log time and the undecoded payload of every message on a topic. The reader stopped at the summary until now, so a caller could learn what a recording holds but could not read any of it.

The summary's chunk indexes say which channels each chunk holds, so a query decompresses only the chunks that carry the topic: on the fixtures the imu topic reads 8 of the 19 chunks, while the status topic reads all 19. An index states its channels through the map of message index offsets, and an empty map means no message indexing is available, so an index without channel ids cannot prune its chunk.

A file carrying chunk indexes should hold every message in a chunk, but the specification only recommends it and a message record is legal in the data section, so the walk continues through the data section after the indexed chunks. That guards against a writer ignoring the recommendation, which would otherwise lose messages with no error. No writer is known to produce such a file; the Python and Go writers both fix the choice for a whole file at construction.

About two hundred of the lines in `tests/test_mcap.py` are mechanical: the record builders that hand-assemble a file move into an `McapBytes` mixin, which renames every call site but leaves the assembled bytes the same. The rest exercises the iterator against the fixtures and against hand-assembled files.

Related to #1286.

### DIFF
--- a/cpp/solvcon/mcap/mcap_reader.cpp
+++ b/cpp/solvcon/mcap/mcap_reader.cpp
@@ -10,6 +10,9 @@
 #include <cstring>
 #include <stdexcept>
 
+#include <lz4frame.h>
+#include <zstd.h>
+
 namespace solvcon
 {
 
@@ -31,6 +34,8 @@ enum Opcode : uint8_t
     OP_FOOTER = 0x02,
     OP_SCHEMA = 0x03,
     OP_CHANNEL = 0x04,
+    OP_MESSAGE = 0x05,
+    OP_CHUNK = 0x06,
     OP_CHUNK_INDEX = 0x08,
     OP_STATISTICS = 0x0b,
 }; /* end enum Opcode */
@@ -67,6 +72,9 @@ public:
     }
 
     void skip(uint64_t length) { bytes(length); }
+
+    /// The bytes after the cursor.
+    std::string_view rest() { return bytes(m_data.size() - m_pos); }
 
     bool done() const { return m_pos >= m_data.size(); }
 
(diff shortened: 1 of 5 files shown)


cpp/solvcon/mcap/mcap_reader.cpp
```diff
@@ -10,6 +10,9 @@
 #include <cstring>
 #include <stdexcept>
 
+#include <lz4frame.h>
+#include <zstd.h>
```
tigercosmos, line 14: Import lz4 and zstd.

cpp/solvcon/mcap/mcap_reader.cpp
```diff
@@ -117,6 +125,78 @@ static bool is_kept_record(uint8_t opcode)
...
+
+static std::string decompress(std::string_view compression, std::string_view input, uint64_t output_size)
+{
+    if (compression.empty())
+    {
+        return std::string(input);
+    }
+
+    std::string output(output_size, '\0');
+    if ("lz4" == compression)
```
tigercosmos, line 167: Handle the lz4 compression.

cpp/solvcon/mcap/mcap_reader.cpp
```diff
@@ -117,6 +125,78 @@ static bool is_kept_record(uint8_t opcode)
...
+        size_t dst_size = output.size();
+        size_t src_size = input.size();
+        size_t const ret = LZ4F_decompress(dctx, output.data(), &dst_size, input.data(), &src_size, nullptr);
+        LZ4F_freeDecompressionContext(dctx);
+        if (0 != ret || dst_size != output_size || src_size != input.size())
+        {
+            throw std::runtime_error("cannot decompress the lz4 MCAP chunk");
+        }
+    }
+    else if ("zstd" == compression)
```
tigercosmos, line 184: Handle zstd compression.

cpp/solvcon/mcap/mcap_reader.cpp
```diff
@@ -384,6 +477,146 @@ std::map<std::string, std::string> Reader::topics() const
...
+    {
+        uint64_t const start = m_offset;
+        m_offset += RECORD_HEADER_SIZE + length;
+
+        if (OP_MESSAGE == opcode)
+        {
+            m_buffer = m_reader->read_bytes(start, RECORD_HEADER_SIZE + length);
+            return true;
+        }
+        if (OP_CHUNK == opcode && m_reader->m_chunk_indices.empty())
```
tigercosmos, line 610: Chunk is optional in mcap, which contains messages.

tests/test_mcap.py
```diff
@@ -239,14 +309,104 @@ def test_duplicate_records_must_agree(self):
...
+                         [(10, b"in"), (20, b"out")])
 
     def test_summary_ending_inside_a_record_header(self):
         with self.assertRaisesRegex(RuntimeError, "ends inside a record"):
-            sc.mcap.Reader(self.build(_schema(1, "S"), trailing=b"\x03\x00"))
+            sc.mcap.Reader(self.build(self.schema(1, "S"),
+                                      trailing=b"\x03\x00"))
+
+
+class McapMessageTC(McapReaderTB):
```
tigercosmos, line 359: Test cases for reading messages in MCAP files.

PR #1402, opened 2026-08-24
Compile ROS 2 IDL schemas into a flat CDR decode plan

tigercosmos:
`DecodePlan` compiles the `ros2idl` schema text of an MCAP recording, plus a list of dotted field paths, into a flat CDR walk program. This is the second step of #1286: the plan is the contract that the C++ executor of the next PR runs over each message payload, so no C++ lands here. The module docstring states the instruction set.

A container is walked through a loop instruction that carries the length of its body, and a body may hold another body, so a sequence or an array of structs or strings does not block the fields behind it. Selection stays numeric and bool scalars, and an enum reads as its 32-bit ordinal.

Two CDR rules are easy to get wrong, so the plan settles them rather than leaving them to the executor. `skip_sequence` aligns to the element width on a nonzero count, because CDR pads between a sequence count and its elements. An annotation that moves bytes, such as `@optional`, is rejected rather than stripped; metadata annotations, including a misspelled one, still pass.

One scanner strips comments, annotations, and string literals in a single pass. `rosidl` writes each `.msg` comment as a `@verbatim (text="...")` annotation, and the stock `builtin_interfaces/msg/Time` carries a URL in one, so removing `//` comments first would end that annotation at the URL and discard the rest of the schema section.

`tests/test_mcap_decode_plan.py` covers each instruction form and each rejection path.

Related to #1286

### DIFF
--- a/solvcon/mcap/_decode_plan.py
+++ b/solvcon/mcap/_decode_plan.py
@@ -1,0 +1,412 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Compile ROS 2 IDL schemas into flat CDR decode instructions.
+
+MCAP ``ros2idl`` schemas contain the root IDL and dependency sections.
+``DecodePlan`` resolves their structs and compiles dotted field paths into
+the instruction sequence an executor runs over one message payload.  Every
+offset is relative to the start of the CDR body, which begins after the
+four-byte encapsulation header that also states the byte order.
+
+Each instruction is a tuple whose first item names it:
+
+- ``("align", n)`` advances to the next multiple of ``n`` bytes.
+- ``("skip", n)`` advances ``n`` bytes.
+- ``("skip_string", )`` reads a 4-byte length and advances past the text.
+- ``("skip_sequence", name)`` reads a 4-byte count and advances past that
+  many primitives of type ``name``.  CDR pads between the count and the
+  elements, so a nonzero count aligns to the width of ``name`` first.  An
+  empty sequence carries no padding.
+- ``("skip_sequence_body", n)`` reads a 4-byte count and runs the ``n``
+  instructions that follow it that many times.
+- ``("skip_array_body", count, n)`` runs the ``n`` instructions that follow
+  it ``count`` times.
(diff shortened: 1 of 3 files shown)


solvcon/mcap/_decode_plan.py
```diff
@@ -0,0 +1,412 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Compile ROS 2 IDL schemas into flat CDR decode instructions.
+
+MCAP ``ros2idl`` schemas contain the root IDL and dependency sections.
```
tigercosmos, line 6: Do decode the payload, we need to resolve the structures from ROS2 IDE.

solvcon/mcap/_decode_plan.py
```diff
@@ -0,0 +1,412 @@
...
+  output column ``column``.
+
+A body may hold another body, so an executor walks nested containers by
+recursion.  Selection is numeric and bool scalars only, and an enum reads as
+its 32-bit ordinal; a container, a string, or a nested struct in the output
+position is rejected, as is an indexed path.
+"""
+
+import dataclasses
+import re
```
tigercosmos, line 35: The structure decoder only runs once, so using re should be fine.

tests/test_mcap_decode_plan.py
```diff
@@ -0,0 +1,277 @@
...
+
+import types
+import unittest
+
+import solvcon as sc
+
+
+class DecodePlanTC(unittest.TestCase):
+
+    SCHEMA_TEXT = """\
```
tigercosmos, line 12: Example of the IDL.

tests/test_mcap_decode_plan.py
```diff
@@ -0,0 +1,277 @@
...
+        )
+
+    def test_nested_struct_fields(self):
+        plan = sc.mcap.DecodePlan(
+            self.schema(), fields=["dt", "current.x", "current.active"])
+        self.assertEqual(plan.fields,
+                         ("dt", "current.x", "current.active"))
+        self.assertEqual(plan.types, ("float64", "float64", "bool"))
+        self.assertEqual(
+            plan.instructions,
```
tigercosmos, line 52: Example of the result.

Issue #1287, opened 2026-08-07
MCAP signal processing step 3: Grow solvcon/track/DataFrame into the signal-log container: from_columns, asof, window, read_from_mcap


tigercosmos:
Third of three issues adding signal-processing support to solvcon (first: time-series kernels; second: columnar MCAP reader). This issue composes the pieces in pure Python: `solvcon/track/DataFrame` becomes the container that holds extracted signal columns, aligns series recorded on different sample grids, and hands out half-open window views for analysis.

`track/` already has a `DataFrame` with `SimpleArrayFloat64` columns and a `SimpleArrayUint64` epoch-ns index, CSV ingest, and sort-by-index. The additions keep its conventions: integer-nanosecond timestamps end to end, zero-copy views over shared buffers, and lookup semantics that are zero-order hold and strictly causal (the value at t is the last sample at or before t; a query before the first sample yields no value).

### PRs (2, each under 500 changed lines)

**PR 1: `DataFrame.from_columns` + `asof` + `window`.** `from_columns(index=..., **columns)` constructs a frame from already-built arrays (the shape `extract()` produces). `asof(times, values)` ZOH-aligns a foreign series onto the frame's grid via searchsorted, with index < 0 meaning before-first-sample and therefore no value. `window(start, end)` is two searchsorteds plus zero-copy views, half-open `[start, end)`. Tests in `tests/test_track.py` alongside the existing DataFrame tests; the boundary-sample-belongs-to-the-next-window rule is pinned explicitly.

Done in #1403

**PR 2: `DataFrame.read_from_mcap`.** Thin wiring of `Reader` + `DecodePlan` + `extract` into a frame per topic. Skipped or xfailed cleanly when solvcon is built with `BUILD_MCAP=OFF`. After this PR the full envisioned pipeline runs end to end: read a recording, extract columns, run `deriv`/`movavg`/`held` over the whole log, window by time boundaries, and evaluate with `buffer/` reductions (`abs().max()`, `std(ddof=0)`, `any()`) plus `true_intervals` for the intervals where a condition held. That pipeline becomes an integration test on a small fixture log, and, if worthwhile, a documented example.

### Example (pseudo-code)

The full pipeline this issue makes runnable end to end:

```python
import numpy as np

from solvcon import timeseries as ts
from solvcon.track import DataFrame

# PR 2: one call from recording to columns-in-a-frame.
df = DataFrame.read_from_mcap("recording.mcap",
                              topic="/vehicle/status",
[...]

PR #1403, opened 2026-08-24
Add from_columns, asof, and window to the track DataFrame

tigercosmos:
Grow `solvcon/track/DataFrame` into the container that holds extracted signal columns, the first of two pull requests for issue #1287. `from_columns` builds a frame from arrays that are already in memory, the shape a signal extractor produces. `asof` aligns a series recorded on another sample grid onto the frame index with a strictly causal zero-order hold, and returns the aligned values together with a mask that is false where there was no sample to hold. `window` cuts the half-open interval `[start, end)` as views over the same buffers. A column may now carry any SimpleArray element type rather than float64 alone, because a signal log records flags and counters beside its measurements.

`window` and `asof` search through `SimpleArray::searchsorted` rather than numpy. For `window` numpy is not merely slower but wrong: it has no integer type holding both uint64 and int64, so it compares a uint64 index against a plain Python int through float64, which cannot tell apart the epoch nanoseconds this index exists to carry.

The binding work the container needs comes with it, and is the part to review first. Casting a SimpleArray out of a plex took the plex by value. Copying a plex allocates a fresh SimpleArray and clones the whole buffer, and the caster then pointed into that copy and let it die at the end of the load, so a binding handed a plex read freed memory after paying a deep copy of its argument. Binding the plex by reference removes both. It also changes aliasing: a loaded SimpleArray now points at the caller's data rather than a private copy, so a binding that writes to its argument writes through to the caller.

A plex also builds from a numpy array through the maker that already serves the typed constructor, so it honors the strides and the offset of its source instead of reading the buffer as one flat block. That maker moves out of `WrapSimpleArray` into `array_common.hpp` as a free function template, so neither binding unit has to reach into the other. The constructor consequently rejects what the flat-block path accepted: a dtype mismatch, a stride that is not a multiple of the item size, and an underaligned data pointer.

`solvcon/timeseries.py` gains one Python wrapper per kernel, so the kernels take a numpy array or a plex as well as a typed array. The module's six names become real functions rather than objects forwarded from the extension.

Related to #1287, #1285, and #1286.

Left for follow-up: `asof` and `window` each rescan the index for order on every call, which dominates a windowed sweep over a large log; the caster could accept a numpy array directly and retire the Python wrappers in `solvcon/timeseries.py`; and the dtype table in `dataframe.py` duplicates a dispatch C++ already carries.

### DIFF
--- a/cpp/solvcon/buffer/pymod/array_common.hpp
+++ b/cpp/solvcon/buffer/pymod/array_common.hpp
@@ -480,6 +480,124 @@ pybind11::object ArrayPropertyHelper<T>::shift_slice_bound(
     return pybind11::reinterpret_steal<pybind11::object>(index) + pybind11::int_(offset);
 }
 
+/**
+ * Build an array viewing the memory of a numpy array.
+ *
+ * The result shares the buffer with @p arr_in and keeps it alive, so it
+ * carries the per-axis strides in elements and the offset a negative stride
+ * puts the first element at.
+ *
+ * @param[in] arr_in Source array, whose dtype must be @p T.
+ * @return An array over the same memory.
+ */
+template <typename T>
+SimpleArray<T> make_array_from_numpy(pybind11::array & arr_in)
+{
+    namespace py = pybind11;
+
+    using value_type = typename SimpleArray<T>::value_type;
+    using array_order_type = typename SimpleArray<T>::ArrayOrder;
+
+    if (!dtype_is_type<T>(arr_in))
+    {
+        throw std::runtime_error("dtype mismatch");
+    }
+
+    solvcon::detail::shape_type shape;
+    solvcon::detail::shape_type stride;
+    constexpr auto itemsize = static_cast<ssize_t>(sizeof(value_type));
+    ssize_t byte_span_begin = 0;
+    ssize_t byte_span_end = 0;
+    bool has_element = true;
+    for (ssize_t i = 0; i < arr_in.ndim(); ++i)
+    {
+        shape.push_back(arr_in.shape(i));
+        ssize_t const byte_stride = arr_in.strides(i);
+        if (byte_stride % itemsize != 0)
+        {
+            throw std::runtime_error(
+                std::format("NumPy byte stride {} in dimension {} is not divisible by item size {}",
+                            byte_stride,
+                            i,
+                            itemsize));
+        }
+        stride.push_back(byte_stride / itemsize);
+        if (shape[i] == 0)
+        {
+            has_element = false;
+            continue;
+        }
+        ssize_t const axis_byte_offset = (shape[i] - 1) * byte_stride;
+        if (axis_byte_offset < 0)
+        {
+            byte_span_begin += axis_byte_offset;
+        }
+        else
+        {
+            byte_span_end += axis_byte_offset;
+        }
+    }
+    if (!has_element)
+    {
+        byte_span_begin = 0;
+        byte_span_end = 0;
+    }
+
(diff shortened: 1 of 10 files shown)


cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -480,6 +480,124 @@ pybind11::object ArrayPropertyHelper<T>::shift_slice_bound(
...
+ *
+ * The result shares the buffer with @p arr_in and keeps it alive, so it
+ * carries the per-axis strides in elements and the offset a negative stride
+ * puts the first element at.
+ *
+ * @param[in] arr_in Source array, whose dtype must be @p T.
+ * @return An array over the same memory.
+ */
+template <typename T>
+SimpleArray<T> make_array_from_numpy(pybind11::array & arr_in)
```
tigercosmos, line 494: Move to the `array_common.hpp`

cpp/solvcon/timeseries/pymod/wrap_timeseries.cpp
```diff
@@ -24,13 +26,14 @@ void wrap_timeseries(pybind11::module & mod)
             small_vector<SimpleArray<uint64_t> const *> arrays;
             for (py::handle const arg : args)
             {
-                if (!py::isinstance<SimpleArray<uint64_t>>(arg))
+                py::detail::make_caster<SimpleArray<uint64_t>> caster;
+                if (!caster.load(arg, true)) // implicit conversion
```
tigercosmos, line 30: Allow Numpy to SimpleArray.

tests/test_timeseries.py
```diff
@@ -33,6 +33,73 @@ def bl(values):
                                        'complex64', 'complex128')
 
 
+class NumpyInputTC(unittest.TestCase):
+
+    def test_every_kernel_accepts_ndarrays(self):
```
tigercosmos, line 38: Test that kernels accept numpy.

solvcon/track/dataframe.py
```diff
@@ -87,25 +165,82 @@ def read_from_text_file(
...
     @property
     def index(self):
-        return self._index_data.ndarray
+        return np.asarray(self._index_data)
+
+    @property
+    def empty(self):
+        return self.shape[0] == 0
+
+    def asof(self, times, values):
```
tigercosmos, line 202: Meat.

solvcon/track/dataframe.py
```diff
@@ -22,6 +39,67 @@ def _init_members(self):
...
+        """Build a frame from parts that are already normalized."""
+        ret = cls()
+        ret._index_data = index
+        ret._index_name = index_name
+        ret._columns = list(names)
+        ret._data = list(data)
+        return ret
+
+    @classmethod
+    def from_columns(cls, index, **columns):
```
tigercosmos, line 92: Meat.

tests/test_timeseries_dataframe.py
```diff
@@ -182,4 +196,190 @@ def test_read_from_text_file_missing_raises_filenotfound(self):
...
+        message = "data frame has no index"
+
+        with self.assertRaisesRegex(ValueError, message):
+            frame['speed'] = self._f64([])
+        with self.assertRaisesRegex(ValueError, message):
+            frame.asof(self._u64([]), self._f64([]))
+        with self.assertRaisesRegex(ValueError, message):
+            frame.window(0, 1)
+
+    def test_asof_is_causal_and_uses_last_duplicate(self):
```
tigercosmos, line 256: Demo of `from_columns` with `asof`.

Issue #1397, opened 2026-08-22
Add device-aware SimpleArray storage and explicit Metal execution


ThreeMonth03:
## Motivation

Without device-aware storage, every Metal operation must copy a CPU-backed `SimpleArray` into an `MTLBuffer` and copy the result back. An operation chain repeats both copies and synchronizes after every step.

```text
CPU array -> copy -> Metal -> copy -> CPU array -> copy -> Metal -> copy -> CPU array
```

This issue lets owned `ConcreteBuffer` storage use a shared `MTLBuffer`. Metal output then becomes the next Metal input without crossing a CPU boundary:

```text
Metal array -> matmul -> Metal array -> matmul -> Metal array -> one final CPU copy
```

The result is fewer intermediate copies and synchronization points, asynchronous operation chains, and one `SimpleArray` API for CPU and Metal storage. CPU storage remains the default on every platform.

```python
a = SimpleArrayFloat16(..., device="metal")
b = SimpleArrayFloat16(..., device="metal")
c = SimpleArrayFloat16(..., device="metal")

result = a.matmul_metal(b).matmul_metal(c)
result.wait()                       # synchronize, keep Metal storage
cpu = result.to(device="cpu")       # wait and deep-copy
```

```text
device="metal"       -> shared Metal storage
wait()               -> synchronize only
to(device="cpu")     -> new CPU-owned snapshot

Metal storage        -> Float16, Float32, Int8, Int16
Float16 @ Float16    -> Float16
Float32 @ Float32    -> Float32
Int8 @ Int8          -> MPS Float32 -> GPU cast -> Int8
Int16 @ Int16        -> MPS Float32 -> GPU cast -> Int16
other dtypes         -> clear error
```

Integer cast-back stays on Metal, so it does not force a wait or a CPU copy. It narrows the MPS Float32 result to the input dtype. Large integer reductions are not promised to be bit-exact with CPU integer accumulation after Float32 loses integer precision.

```python
a = SimpleArrayInt8([[100, 100]], device="metal")
b = SimpleArrayInt8([[2], [2]], device="metal")

(a.matmul_metal(b).to(device="cpu")).ndarray
# array([[-112]], dtype=int8)
```

## Tasks

Keep each task and pull request at or below 500 changed lines, including tests.

Task 1 and #1399 can land independently. Task 2 depends on both. Task 3 uses fake work and does not depend on the Task 6 executor.

[...]

PR #1408, opened 2026-08-25
Add shared Metal buffer storage

ThreeMonth03:
Metal construction allocates CPU-visible shared `MTLBuffer` storage on a unified-memory device. The Metal layer returns storage and its owner; `ConcreteBuffer` remains the only buffer factory.

Alignment, empty buffers, device limits, and unavailable build or runtime configurations are handled explicitly. Synchronization, `SimpleArray` device APIs, and MPS execution remain in later tasks.

Related to #1397.

### DIFF
--- a/cpp/solvcon/buffer/BufferBase.hpp
+++ b/cpp/solvcon/buffer/BufferBase.hpp
@@ -49,7 +49,7 @@ public:
     explicit operator bool() const { return static_cast<bool>(m_begin); }
     size_type size() const
     {
-        return static_cast<size_type>(this->m_end - this->m_begin);
+        return this->m_begin == this->m_end ? 0 : static_cast<size_type>(this->m_end - this->m_begin);
     }
     size_t nbytes() const { return size() * sizeof(int8_t); }
 
--- a/cpp/solvcon/buffer/BufferStorage.hpp
+++ b/cpp/solvcon/buffer/BufferStorage.hpp
@@ -1,0 +1,99 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Ownership contracts for ConcreteBuffer storage.
+ *
+ * @ingroup group_core
+ */
+
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+#include <memory>
+
+#ifdef _WIN32
+#include <malloc.h>
+#endif
+
+namespace solvcon
+{
+
+namespace detail
+{
+
+/**
+ * Base class for buffer-storage deallocation.
+ *
+ * ConcreteBuffer's data deleter calls this object to release the backing
+ * storage. A derived remover may instead release its owner when the remover
+ * itself is destroyed.
+ */
+struct ConcreteBufferRemover
+{
+
+    ConcreteBufferRemover() = default;
+    ConcreteBufferRemover(ConcreteBufferRemover const &) = default;
+    ConcreteBufferRemover(ConcreteBufferRemover &&) = default;
+    ConcreteBufferRemover & operator=(ConcreteBufferRemover const &) = default;
+    ConcreteBufferRemover & operator=(ConcreteBufferRemover &&) = default;
+    virtual ~ConcreteBufferRemover() = default;
+
+    /**
+     * Release heap storage with its original alignment.
+     * @param p Storage to release.
+     * @param alignment Original alignment; 0 means no explicit alignment.
+     */
+    static void deallocate_memory(int8_t * p, size_t alignment);
+
+    /**
+     * Release storage or its backing resource.
+     * @param p CPU-visible storage pointer.
+     * @param alignment Storage alignment; 0 means no explicit alignment.
+     */
+    // NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays,readability-non-const-parameter)
+    virtual void operator()(int8_t * p, size_t alignment) const { deallocate_memory(p, alignment); }
+
+}; /* end struct ConcreteBufferRemover */
+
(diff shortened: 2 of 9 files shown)


cpp/solvcon/buffer/BufferStorage.hpp
```diff
@@ -0,0 +1,101 @@
...
+using buffer_storage_remover_type = ConcreteBufferRemover;
+
+/**
+ * Transfer object for device-backed storage.
+ *
+ * The remover moves into ConcreteBuffer, where it keeps the backing resource
+ * alive and handles release. A successful allocation has a non-null remover;
+ * its data pointer is null only for zero-byte storage.
+ */
+struct BufferStorage
```
ThreeMonth03, line 87: Added a neutral storage carrier so Metal can return memory and ownership without constructing `ConcreteBuffer`. This avoids a Metal-to-`ConcreteBuffer` dependency and keeps `ConcreteBuffer` as the only factory.

cpp/solvcon/device/metal/metal.mm
```diff
@@ -5,23 +5,68 @@
...
+{
+
+public:
+
+    explicit MetalBufferRemover(id<MTLBuffer> buffer)
+        : m_buffer(buffer)
+    {
+    }
+
+    // ARC keeps the allocation alive until this remover is destroyed.
```
ThreeMonth03: ARC retains the `MTLBuffer` through this remover and releases it when the remover is destroyed. Moving the remover into `ConcreteBuffer` keeps storage alive after `MetalManager::shutdown()`.

cpp/solvcon/device/metal/metal.mm
```diff
@@ -33,10 +78,46 @@
...
 
+detail::BufferStorage MetalManager::Impl::allocate_buffer(size_t nbytes, size_t alignment) const
+{
+    if (nbytes == 0)
+    {
+        return {.m_data = nullptr, .m_remover = std::make_unique<MetalBufferRemover>(nil)};
+    }
+
+    size_t const padding = alignment == 0 ? 0 : alignment - 1;
+    size_t const capacity = nbytes + padding;
```
ThreeMonth03, line 108: Allocating `nbytes + alignment - 1` gives `std::align` enough room to satisfy the requested alignment. The visible buffer size remains `nbytes`.

cpp/solvcon/buffer/BufferBase.hpp
```diff
@@ -49,7 +49,7 @@ class BufferBase
     explicit operator bool() const { return static_cast<bool>(m_begin); }
     size_type size() const
     {
-        return static_cast<size_type>(this->m_end - this->m_begin);
+        return this->m_begin == this->m_end ? 0 : static_cast<size_type>(this->m_end - this->m_begin);
```
ThreeMonth03, line 52: Empty buffers use equal null endpoints, so `size()` checks equality before subtraction. This avoids undefined null-pointer arithmetic while preserving a zero size.

ThreeMonth03 (commented):
@yungyuc Please review the latest revision.

- [x] Keep `ConcreteBuffer` as the only buffer factory.
- [x] Transfer shared Metal storage ownership through the existing buffer deleter.
- [x] Preserve alignment and empty-buffer invariants.

cpp/solvcon/buffer/BufferStorage.hpp
yungyuc: Is it only for "device" memory that is accessible to metal?  If so, should it be named as `BufferStorage` or something else?

cpp/solvcon/buffer/BufferBase.hpp
yungyuc: Makes sense.

cpp/solvcon/buffer/ConcreteBuffer.cpp
```diff
@@ -18,7 +22,14 @@ std::shared_ptr<ConcreteBuffer> ConcreteBuffer::construct(size_t nbytes, size_t
     case BufferDevice::Cpu:
         return construct(nbytes, alignment);
     case BufferDevice::Metal:
+#ifdef SOLVCON_METAL
+    {
+        detail::BufferStorage storage = device::MetalManager::instance().allocate_buffer(nbytes, alignment);
```
yungyuc, line 27: Is it correct that the memory managed by `storage` is accessible from both CPU and GPU?

yungyuc (changes requested):
Code is OK.  Intention needs to be discussed.

- [x] Clarify the name for the new class `BufferStorage`.

cpp/solvcon/buffer/BufferStorage.hpp
ThreeMonth03: Renamed it to `DeviceBufferStorage`. It is used only by the device allocation path; ordinary CPU buffers keep their current allocation. The carrier is backend-neutral: a device backend returns a CPU-visible address plus its owner, so Metal is one implementation rather than part of the type.

cpp/solvcon/buffer/ConcreteBuffer.cpp
ThreeMonth03: Yes. For Metal, `m_data` points to `MTLBuffer.contents`, and the remover retains that same `MTLBuffer`; CPU and GPU therefore access one shared allocation. This task manages lifetime only. Command ordering and CPU/GPU synchronization are added in Task 3, so shared accessibility does not mean concurrent access is automatically safe.

ThreeMonth03 (commented):
> - [x] Clarify the name for the new class `BufferStorage` by renaming it to `DeviceBufferStorage`.

cpp/solvcon/buffer/ConcreteBuffer.cpp
yungyuc: Unsafe concurrency is OK.  It's everywhere in HPC.  We just need to be careful and wise enough to use proper preventive measures.

yungyuc (approved):
LGTM

PR #1409, opened 2026-08-25
Complete Windows setup prerequisites

tigercosmos:
The Windows dependency instructions now install user-local CMake, Ninja, and WinLibs/gfortran packages, and the NumPy/SciPy path fails before creating an environment when no supported Fortran compiler is available.

`build.ps1` now considers every CMake executable on `PATH`, so Visual Studio's older bundled copy cannot hide a compatible CMake 4.x installation.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -204,7 +204,7 @@ function Resolve-Cmake {
     # A CMake >= 4.0.1 (solvcon's minimum): the one on PATH if new enough (vcvars
     # may put VS 2022's 3.31.6 there), else the newest bundled with a VS.
     $candidates = @()
-    $onpath = Get-Command cmake.exe -ErrorAction SilentlyContinue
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -204,7 +204,7 @@ function Resolve-Cmake {
     # A CMake >= 4.0.1 (solvcon's minimum): the one on PATH if new enough (vcvars
     # may put VS 2022's 3.31.6 there), else the newest bundled with a VS.
     $candidates = @()
-    $onpath = Get-Command cmake.exe -ErrorAction SilentlyContinue
+    $onpath = @(Get-Command cmake.exe -All -ErrorAction SilentlyContinue)
```
tigercosmos, line 207: List all CMake versions to find the compatible one.

PR #1410, opened 2026-08-25
Share the bounds-checked byte cursor of the MCAP reader

tigercosmos:
`detail::ByteCursor` in `mcap_cursor.hpp` carries the bounds-checked read and skip that `FieldReader` used over one record. The cursor over a CDR message body that the decode plan executor needs derives from it, so the two share one bounds check instead of repeating it.

`("skip_sequence", ...)` now carries the element width instead of the type name. The executor aligns to the width and skips past the elements, so the width is all it reads from the instruction.

This is the first of four stacked PRs that split the executor step of issue #1286 (the former single PR here). The next three are pushed as `worktree-issue-1286-pr3b-decode-plan` (the plan check, `core.McapDecodePlan`), `worktree-issue-1286-pr3c-executor` (the walk and `McapReader.extract`), and `worktree-issue-1286-pr3d-python-extract` (`Reader.extract` and `ColumnSet`); each opens once the one before it merges. #1381, #1386, and #1401 built the container reader, and #1402 built the plan compiler.

Related to #1286.

### DIFF
--- a/cpp/solvcon/mcap/mcap_cursor.hpp
+++ b/cpp/solvcon/mcap/mcap_cursor.hpp
@@ -1,0 +1,98 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Bounds-checked cursor over the bytes of an MCAP record or message.
+ *
+ * @ingroup group_inout
+ */
+
+#include <cstdint>
+#include <cstring>
+#include <stdexcept>
+#include <string_view>
+
+namespace solvcon
+{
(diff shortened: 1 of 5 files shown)


cpp/solvcon/mcap/mcap_cursor.hpp
```diff
@@ -0,0 +1,98 @@
...
+namespace detail
+{
+
+/**
+ * @internal
+ * Cursor over a run of bytes.  MCAP writes its integers little-endian, and
+ * so does every platform solvcon builds for, so a value is a plain copy out
+ * of the buffer.
+ */
+class ByteCursor
```
tigercosmos, line 35: Add a byte cursor for reading MCAP file.

PR #1412, opened 2026-08-26
Share the Windows CI setup and ship the MKL runtime in the portable artifact

tigercosmos:
Moves the MSVC, CMake, Qt, Python, MKL or vcpkg, and pip steps that `devbuild_windows.yml`, `nightly_build_windows.yml`, and `nightly_sanitizer.yml` each copied into one composite action, `.github/actions/setup_windows`.

The nightly builds `_solvcon` against Intel MKL, so the extension imports `mkl_rt.3.dll`, but the portable artifact shipped no MKL DLL. Running the 2026-08-25 artifact's `python.exe` with `PATH` limited to the artifact directory and `System32` gives `ImportError: DLL load failed while importing _solvcon`. The portable step now copies the `mkl_rt` runtime closure into the tree and imports `_solvcon` under that restricted `PATH` before archiving. It also verifies the embedded CPython zip against a pinned SHA-256 and drops the `get-pip.py` download.

The workflows have not run with this change yet.

### DIFF
--- a/.github/actions/setup_windows/action.yml
+++ b/.github/actions/setup_windows/action.yml
@@ -1,0 +1,159 @@
+name: Set up solvcon CI dependencies on Windows
+description: Set up the necessary dependencies for running CI tests on Github action Windows runners
+
+inputs:
+  workflow:
+    description: The name of the workflow that will use this action
+    required: true
+  cmake-build-type:
+    description: Build type of the calling job, used in the compiler cache key
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+
+    - name: check inputs
+      if: ${{ inputs.workflow == 'build' && inputs.cmake-build-type == '' }}
+      shell: pwsh
+      run: |
+        throw ('setup_windows: workflow=build needs cmake-build-type, ' +
+          'otherwise Debug and Release share one compiler-cache key')
+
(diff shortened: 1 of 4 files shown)


.github/actions/setup_windows/action.yml
```diff
@@ -0,0 +1,159 @@
+name: Set up solvcon CI dependencies on Windows
```
tigercosmos, line 1: `setup_windows` follows the current convention.

.github/workflows/nightly_build_windows.yml
```diff
@@ -227,6 +151,97 @@ jobs:
...
           Copy-Item -Path ".\solvcon" -Destination $destination -Recurse
+          Copy-Item -Path ".\_solvcon*.pyd" -Destination $destination
+
+          # mkl_rt dispatches at run time to a threading layer, the CPU kernels
+          # for the host ISA, and the vector-math kernels; that closure plus the
+          # OpenMP runtime is all solvcon reaches.  The wheel also ships cluster
+          # and offload libraries the portable tree never loads, so drop them.
+          # mkl_tbb_thread stays: with MKL_THREADING_LAYER=TBB already in the
+          # user's environment, a missing mkl_tbb_thread is fatal.
+          $mkl_runtime_keep = @(
```
tigercosmos, line 162: Make sure shipping MKL DLL

PR #1413, opened 2026-08-26
Verify scdv downloads with SHA-256

tigercosmos:
Replaces the MD5 check in `build-scdv.sh` and `build-scdv-windows.ps1` with SHA-256. The old helpers continued after a mismatch, and the CPython, libclang, OpenBLAS, and `7zr.exe` downloads carried no digest. The new helpers refuse a download without a digest, download to a temporary file, and abort on a mismatch. An overridden version must supply its digest through `PYTHON_SHA256`, `LIBCLANG_SHA256`, `QT_SHA256`, or `PYSIDE_SHA256`.

CPython moves to 3.14.7 and OpenSSL to 3.5.8, since 1.1.1 is end of life. OpenSSL 3 needs `--libdir=lib` and `install_sw install_ssldirs` to keep the prefix layout.

### DIFF
--- a/contrib/dependency/windows/build-scdv-windows.ps1
+++ b/contrib/dependency/windows/build-scdv-windows.ps1
@@ -63,6 +63,10 @@
 #     QT_SUB_VER: Qt patch version.
 #     PYSIDE_VERSION: Qt for Python (pyside-setup) source release version.
 #     LIBCLANG_VERSION: Qt prebuilt libclang version for shiboken.
+#     PYTHON_SHA256: required when overriding PYTHON_VERSION.
+#     LIBCLANG_SHA256: required when overriding LIBCLANG_VERSION.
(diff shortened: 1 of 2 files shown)


contrib/dependency/windows/build-scdv-windows.ps1
```diff
@@ -184,6 +188,27 @@ $PysideVersion = Get-EnvOrDefault 'PYSIDE_VERSION' "$QtMajorVer.$QtSubVer"
 # release is published as a vs2022_64 7z.
 $LibClangVersion = Get-EnvOrDefault 'LIBCLANG_VERSION' '21.1.2'
 
+# SHA-256 of every version-bearing download.  The keys are literal on
+# purpose: interpolating the version variables would make an overridden
+# version match its own key and receive the default version's digest,
+# which would make <NAME>_SHA256 unreachable.
+$PinnedHashes = @{
```
tigercosmos, line 195: Meat.

PR #1419, opened 2026-08-27
Check the MCAP decode plan at construction

tigercosmos:
`solvcon::mcap::DecodePlan` is the executable form of the walk program that `solvcon.mcap.DecodePlan` compiles. Construction checks the plan once, so the executor that runs it over the messages of a topic can walk without checking. It rejects a read inside a container, a column read twice or never, a read of a type no CDR primitive decodes into, an alignment that is not a power of two, a sequence of a width no primitive has, and a container body that runs past the range holding it.

`core.McapDecodePlan(instructions, column_count)` builds one from the instruction tuples the compiler emits. The tests state plans by hand, because the compiler emits only plans the check accepts, and confirm that a compiled plan passes.

This is the second of the four stacked PRs that split the executor step of issue #1286. #1410 shared the byte cursor. `worktree-issue-1286-pr3c-executor` (the walk and `McapReader.extract`) and `worktree-issue-1286-pr3d-python-extract` (`Reader.extract` and `ColumnSet`) follow, each opening once the one before it merges.

Related to #1286.

### DIFF
--- a/cpp/solvcon/mcap/mcap_decoder.cpp
+++ b/cpp/solvcon/mcap/mcap_decoder.cpp
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/mcap/mcap_decoder.hpp>
+
+#include <algorithm>
+#include <stdexcept>
+
+namespace solvcon
+{
+
+namespace mcap
+{
+
+/// Number of instructions a container body runs per element; they follow the step.
+static size_t body_length(PlanStep const & step)
+{
+    return PlanOp::SkipArrayBody == step.op ? step.extra : step.operand;
+}
+
+static bool is_power_of_two(uint32_t value)
+{
+    return 0 != value && 0 == (value & (value - 1));
+}
+
+/**
+ * Whether a CDR primitive decodes into the type.  IDL declares no
+ * half-precision and no complex type, so a column is never of Float16,
+ * Complex64, or Complex128.
+ */
+static bool is_cdr_primitive(DataType type)
+{
+    switch (type)
+    {
+    case DataType::Bool:
+    case DataType::Int8:
+    case DataType::Int16:
+    case DataType::Int32:
+    case DataType::Int64:
+    case DataType::Uint8:
+    case DataType::Uint16:
+    case DataType::Uint32:
+    case DataType::Uint64:
+    case DataType::Float32:
+    case DataType::Float64:
+        return true;
+    default:
+        return false;
+    }
+}
+
+/**
+ * Check one range of instructions and recurse into each container body with
+ * the range of that body.
+ *
+ * @param steps Every instruction of the plan.
+ * @param begin Index of the first instruction of the range.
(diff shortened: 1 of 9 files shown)


cpp/solvcon/mcap/pymod/wrap_McapDecodePlan.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <unordered_map>
+
+namespace solvcon
+{
+
+namespace python
+{
+
+/**
+ * Translate one instruction tuple that solvcon.mcap.DecodePlan compiled.  The
```
tigercosmos, line 18: Decode the complex schema (IDL) on Python side (one-time job), convert the decode plan from Python to C++ for later executor usage.

cpp/solvcon/mcap/pymod/wrap_McapDecodePlan.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * Python side owns the schema grammar, so the instruction names it emits are
+ * the contract between the two halves.
+ */
+static mcap::PlanStep plan_step(pybind11::sequence const & item)
+{
+    static std::unordered_map<std::string, mcap::PlanOp> const ops = {
```
tigercosmos, line 24: 7 operators of the plan.

cpp/solvcon/mcap/mcap_decoder.cpp
```diff
@@ -0,0 +1,140 @@
...
+            if (!is_power_of_two(step.operand) || 8 < step.operand)
+            {
+                throw std::invalid_argument("the MCAP decode plan skips a sequence of no CDR primitive width");
+            }
+            break;
+        case PlanOp::SkipSequenceBody:
+        case PlanOp::SkipArrayBody:
+        {
+            size_t const body_end = it + body_length(step);
+            if (body_end > end)
```
tigercosmos, line 94: `end` needs to be within the container's body end.

cpp/solvcon/mcap/mcap_decoder.cpp
```diff
@@ -0,0 +1,140 @@
...
+        case PlanOp::Read:
+        {
+            // A read inside a container appends one element per container
+            // element, not one per message.
+            if (nested)
+            {
+                throw std::invalid_argument("the MCAP decode plan reads a field inside a container");
+            }
+            DataType const type(static_cast<DataType::enum_type>(step.operand));
+            if (!is_cdr_primitive(type))
```
tigercosmos: Not accept float16 or complex number.

cpp/solvcon/mcap/mcap_decoder.cpp
```diff
@@ -0,0 +1,140 @@
...
+            {
+                throw std::invalid_argument("the MCAP decode plan names a type CDR has no primitive for");
+            }
+            if (step.extra >= types.size() || DataType::Undefined != types[step.extra])
+            {
+                throw std::invalid_argument("the MCAP decode plan does not read every column exactly once");
+            }
+            types[step.extra] = type;
```
tigercosmos, line 120: Record the column's type, also mark the column as read.

PR #1428, opened 2026-08-29
Add `Float16` Python and NumPy integration

ThreeMonth03:
`SimpleArrayFloat16` is available in C++, but Python cannot yet construct it or exchange its storage with NumPy.

This change exposes `SimpleArrayFloat16` and `SimpleCollectorFloat16`, adds [PEP 3118](https://peps.python.org/pep-3118/) `e` buffers and `SimpleArrayPlex` dispatch, and supports:

```python
source = np.arange(4, dtype=np.float16)
array = SimpleArrayFloat16(array=source)  # shared storage
array[...] = np.array([1, 2, 3, 4], dtype=np.float64)
array[0] = 1.25                           # Python scalar
view = array.ndarray                      # zero-copy NumPy view
```

Float16 copies preserve raw bits, cross-real conversions convert directly, and real/Complex crossings are rejected.

Array calculations remain in Task 4. 
Related to #1393.


cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -19,6 +19,53 @@ namespace python
 namespace detail
 {
 
+template <typename To, typename From>
```
ThreeMonth03: [P3] Keep this conversion policy local

This partially specialized caster hierarchy is used by only one assignment and one pre-copy validity check. Could we keep a small validity predicate before recursion, then handle only the `Float16` source case with a local `if constexpr` through `float`? The other valid paths can continue using `static_cast`, which preserves empty-array rejection without adding a new extension hierarchy.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -75,19 +122,8 @@ struct TypeBroadcastImpl
...
-                auto * ptr_out = arr_out.logical_data() + offset_out;
-                // FIXME: NOLINTNEXTLINE(bugprone-signed-char-misuse,cert-str34-c)
-                *ptr_out = static_cast<out_type>(*ptr_in);
-            }
-            else
-            {
-                throw std::runtime_error("Cannot convert between complex and non-complex types");
-            }
+            auto * ptr_out = arr_out.logical_data() + offset_out;
+            *ptr_out = detail::NumpyElementCaster<out_type, D>::cast(*ptr_in);
```
ThreeMonth03: [P3] Make the conversion direction explicit

`D` is the source NumPy element type here, while `T`/`out_type` is the destination, which is opposite to the comments on `TypeBroadcastImpl<T, D>`. Could we rename these parameters to `To` and `From` (or `Output` and `Input`) and remove the misleading comments?

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -103,13 +139,22 @@ struct TypeBroadcastImpl
...
+        using out_type = typename SimpleArray<T>::value_type;
 
-        ssize_t const ndim = arr_out.ndim();
-        shape_type const left_shape = detail::shape_from_slices(slices);
-        shape_type const sidx_init(ndim, 0);
-        copy_idx(arr_out, slices, arr_new, left_shape, sidx_init, ndim - 1);
+        if constexpr (detail::NumpyElementCastable<out_type, D>)
+        {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+            auto * arr_new = reinterpret_cast<pybind11::array_t<D> const *>(&arr_in);
```
ThreeMonth03: [P3] Rename this typed view

`arr_new` is neither newly allocated nor a new array; it is a typed view of `arr_in`. Could we call it `typed_input` or `typed_arr_in` so its ownership and role are clear?

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -103,13 +139,22 @@ struct TypeBroadcastImpl
...
+
+            ssize_t const ndim = arr_out.ndim();
+            shape_type const left_shape = detail::shape_from_slices(slices);
+            shape_type const sidx_init(ndim, 0);
+            copy_idx(arr_out, slices, arr_new, left_shape, sidx_init, ndim - 1);
+        }
+        else
+        {
+            throw std::runtime_error("Cannot convert between complex and non-complex types");
```
ThreeMonth03: [P3] Use an accurate conversion error

This branch also rejects `complex64 -> complex128`, where both sides are complex, so the current message is factually incorrect. Please report the source and destination dtypes, or use neutral wording such as `Cannot convert between these element types`.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -193,12 +193,17 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
                                    { return pybind11::cast(SimpleArrayPlex(arr)); })
             .wrap_modifiers()
-            .wrap_calculators()
-            .wrap_matrix()
-            .wrap_sort()
-            .wrap_search()
             // ATTENTION: always keep the same interface between WrapSimpleArrayPlex and WrapSimpleArray
             ;
+
+        if constexpr (!std::is_same_v<T, Float16>)
```
ThreeMonth03: [P3] Update the interface invariant

The preceding comment says these wrappers must always keep the same interface, but this condition intentionally omits calculator, matrix, sort, and search methods from `SimpleArrayFloat16`. Please remove the stale invariant or reword it to document the temporary Task 4 exception.

tests/test_float16.py
```diff
@@ -0,0 +1,128 @@
...
+        np.testing.assert_array_equal(target.ndarray.astype('float64'),
+                                      wider.ndarray)
+
+        payload = np.array([0x7c01, 0x8000], dtype='uint16').view('float16')
+        copied = sc.SimpleArrayFloat16((2,))
+        copied[...] = payload
+        np.testing.assert_array_equal(payload.view('uint16'),
+                                      copied.ndarray.view('uint16'))
+
+        midpoint = 1.0 + 2.0**-11 + 2.0**-25
```
ThreeMonth03: [P3] Name this value `above_midpoint`

`1 + 2**-11 + 2**-25` is deliberately above the binary16 midpoint, not at it. Please rename both occurrences and add a short note that the `2**-25` nudge survives Float64 but is lost when first rounded to Float32; that is the reason for the two different expected bit patterns below.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -258,14 +258,31 @@ class SimpleArrayMixinModifiers
...
+    auto athis = static_cast<A *>(this);
+    if (athis->size() == 0)
     {
-        auto athis = static_cast<A *>(this);
-        std::fill(athis->begin(), athis->end(), value);
         return *athis;
     }
-
-}; /* end class SimpleArrayMixinModifiers */
+    if (athis->is_c_contiguous() || athis->is_f_contiguous())
```
ThreeMonth03: `fill()` keeps the contiguous fast path without assuming every view is contiguous:

```text
C/F contiguous -> fill_n(logical_data, size)
strided view    -> IndexRange -> at(index)
```

The second path avoids writing padding or skipped elements.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -2623,7 +2634,7 @@ class SimpleArray
     {
         if (shape.empty())
         {
-            return 0;
+            return 1;
```
ThreeMonth03: Rank zero is a scalar, not an empty array:

```text
shape ()   -> size 1 -> one storage item
shape (0,) -> size 0 -> no storage items
```

This keeps `.at({})`, transpose, fill, copy, and `where()` consistent.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,73 @@
...
+inline bool type_caster<solvcon::Float16>::load(pybind11::handle src, bool convert)
+{
+    if (!src ||
+        (!convert && !PyFloat_Check(src.ptr())) ||
+        !solvcon::python::is_real_python_number(src))
+    {
+        return false;
+    }
+
+    double const number = PyFloat_AsDouble(src.ptr());
```
ThreeMonth03: Python scalar conversion narrows only at the final boundary:

```text
int / float / NumPy real -> double -> Float16
Complex                  -> rejected
```

Using `double` avoids an unintended FP32 rounding step.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,73 @@
...
+    }
+    value = solvcon::Float16(number);
+    return true;
+}
+
+template <>
+struct npy_format_descriptor<solvcon::Float16>
+{
+    static constexpr auto name = const_name("numpy.float16");
+    static pybind11::dtype dtype() { return pybind11::dtype("float16"); }
```
ThreeMonth03: One descriptor connects the three public views:

```text
C++ Float16 <-> NumPy float16 <-> PEP 3118 "e"
```

`ndarray` and `memoryview` therefore expose the same two-byte storage without copying.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -32,95 +72,104 @@ inline solvcon::detail::shape_type shape_from_slices(
...
+    }
+
+    static From input_at(pybind11::array const & arr_in, shape_type const & sidx)
+    {
+        static_assert(std::is_trivially_copyable_v<From>);
+        auto const * data = static_cast<char const *>(arr_in.data());
         for (pybind11::ssize_t axis = 0; axis < arr_in.ndim(); ++axis)
         {
-            offset += arr_in.strides(axis) / arr_in.itemsize() * sidx[axis];
+            data += arr_in.strides(axis) * sidx[axis];
```
ThreeMonth03: NumPy strides are byte strides:

```text
address = base + sum(stride[axis] * index[axis])
value   = memcpy<From>(address)
```

`memcpy` also makes negative, odd, and unaligned strides safe.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -32,95 +72,104 @@ inline solvcon::detail::shape_type shape_from_slices(
...
 
-            if constexpr (valid_conversion)
+    static void broadcast(SimpleArray<To> & arr_out, std::vector<shape_type> const & slices, pybind11::array const & arr_in)
+    {
+        if constexpr ((!is_complex_v<To> && !is_complex_v<From>) || std::is_same_v<To, From>)
+        {
+            shape_type const output_shape = detail::shape_from_slices(slices);
+            if constexpr (std::is_same_v<From, Float16> && std::is_integral_v<To> && !std::is_same_v<To, bool>)
+            {
+                detail::for_each_index(output_shape, [&](shape_type const & sidx)
```
ThreeMonth03: Integer conversion is preflighted for the whole source:

```text
validate every value -> write every value
any invalid value    -> throw; destination unchanged
```

This preserves assignment atomicity for NaN, infinity, and out-of-range values.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -32,95 +72,104 @@ inline solvcon::detail::shape_type shape_from_slices(
...
+                                       { detail::validate_float16_integer<To>(input_at(arr_in, sidx)); });
+            }
+            if (may_overlap(arr_out, arr_in))
```
ThreeMonth03: Assignment chooses its path from aliasing:

```text
disjoint -> read -> write
overlap  -> read -> staged copy -> write
```

Staging makes `a[1:] = a[:-1]` behave like NumPy.

cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -216,6 +228,15 @@ static SimpleArrayPlex make_arrayplex_from_numpy(pybind11::array & arr_in)
 #undef SC_DECL_MAKE_ARRAYPLEX
 }
 
+// NOLINTNEXTLINE(misc-use-anonymous-namespace)
+static void verify_calculator_supported(SimpleArrayPlex const & array)
+{
+    if (array.data_type() == DataType::Float16)
```
ThreeMonth03, line 245: Task 3 supports Float16 storage, but not array calculations:

```python
typed_f16 = SimpleArrayFloat16((2,))
plex_f16 = SimpleArray((2,), dtype="float16")
plex_f64 = SimpleArray((2,), dtype="float64")

typed_f16.min()  # AttributeError: the method is not bound
plex_f16.min()   # RuntimeError: rejected after the dtype check
plex_f64.min()   # works
```

`SimpleArrayFloat16` has a dedicated Python class, so unsupported methods can be omitted. Every Plex dtype shares the same `SimpleArray` class, so `min()` must remain available for supported dtypes and reject Float16 at runtime. `==` and `!=` are bound on the typed class only to prevent Python from silently falling back to object identity.

Task 4 will bind and verify the Float16 operations. The wrapper structure will remain different, but the calculation rejection shown above should disappear.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -192,13 +192,24 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-            .wrap_modifiers()
-            .wrap_calculators()
-            .wrap_matrix()
-            .wrap_sort()
-            .wrap_search()
-            // ATTENTION: always keep the same interface between WrapSimpleArrayPlex and WrapSimpleArray
-            ;
+            .wrap_modifiers();
+
+        if constexpr (std::is_same_v<T, Float16>)
```
ThreeMonth03, line 197: Task 3 limits the typed wrapper to storage and movement:

    Task 3: construct -> NumPy view -> slice/fill -> Plex
    Task 4: calculators -> matrix -> sort/search

SimpleArrayFloat16 binds modifiers but omits the calculation APIs. Equality and inequality are bound only to raise the explicit unsupported-calculation error, which prevents Python from silently falling back to object identity.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -109,6 +166,10 @@ struct TypeBroadcastImpl
         ssize_t const ndim = arr_out.ndim();
         shape_type const left_shape = detail::shape_from_slices(slices);
         shape_type const sidx_init(ndim, 0);
+        if constexpr (std::is_same_v<D, Float16> && std::is_integral_v<T> && !std::is_same_v<T, bool>)
```
ThreeMonth03: Float16-to-integer assignment validates the full source before the first write:

    all values valid -> copy
    invalid value    -> throw -> destination unchanged

This makes non-finite and out-of-range failures atomic.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,138 @@
...
+}
+
+template <HasScalarCodec T>
+bool try_load_scalar(pybind11::handle src, bool convert, T & value)
+{
+    if (!src)
+    {
+        return false;
+    }
+    if (ScalarCodec<T>::load_exact(src, value))
```
ThreeMonth03: Low-precision scalar conversion has one shared numeric path and one representation-specific exact path:

    exact NumPy scalar -> ScalarCodec<T>::load_exact() -> preserve payload
    other real scalar -> double -> T
    complex/non-number -> reject

Float16 supplies the first codec. Future low-precision scalar types can supply their own codec without duplicating the Python acceptance policy.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,138 @@
...
+inline bool type_caster<solvcon::Float16>::load(pybind11::handle src, bool convert)
+{
+    return solvcon::python::detail::try_load_scalar(src, convert, value);
+}
+
+template <>
+struct npy_format_descriptor<solvcon::Float16>
+{
+    static constexpr auto name = const_name("numpy.float16");
+    static pybind11::dtype dtype() { return solvcon::python::detail::ScalarCodec<solvcon::Float16>::dtype(); }
```
ThreeMonth03: ScalarCodec<Float16> owns the NumPy dtype, PEP 3118 format, and exact scalar representation. npy_format_descriptor delegates to it so buffer exposure and scalar ingestion stay aligned:

    C++ Float16 <-> NumPy float16 <-> PEP 3118 "e"

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,123 @@
...
+    return true;
+}
+
+inline bool try_load_float16_scalar(pybind11::handle src, bool convert, Float16 & value)
+{
+    if (!src)
+    {
+        return false;
+    }
+    if (try_load_exact_float16(src, value))
```
ThreeMonth03, line 84: Float16 scalar input has two explicit paths:

    exact np.float16 -> copy two raw bytes -> preserve representation
    other real       -> double -> Float16
    complex/invalid  -> reject

The helpers remain Float16-specific because no other scalar type currently promises bit-exact NumPy scalar loading.

cpp/solvcon/buffer/pymod/array_common.hpp
```diff
@@ -19,12 +19,123 @@
...
+inline bool type_caster<solvcon::Float16>::load(pybind11::handle src, bool convert)
+{
+    return solvcon::python::detail::try_load_float16_scalar(src, convert, value);
+}
+
+template <>
+struct npy_format_descriptor<solvcon::Float16>
+{
+    static constexpr auto name = const_name("numpy.float16");
+    static pybind11::dtype dtype() { return solvcon::python::detail::float16_dtype(); }
```
ThreeMonth03, line 135: The Float16 dtype and PEP 3118 format use the same helpers as exact scalar loading, so both paths describe one storage representation:

    C++ Float16 <-> NumPy float16 <-> PEP 3118 "e"

ThreeMonth03 (commented):
- [x] Expose `Float16` storage, movement, Python buffers, and zero-copy NumPy views.
- [x] Preserve exact bits for `np.float16` scalars while converting other real scalars numerically.
- [x] Keep calculations outside Task 3 for both typed and Plex wrappers.

@yungyuc Please take a look.

tests/test_float16.py
```diff
@@ -0,0 +1,133 @@
...
+        array = sc.SimpleArrayFloat16((3,))
+        for index, value in enumerate((1, 2.25, np.float16(-3.5))):
+            array[index] = value
+        assert_array(np.array([1, 2.25, -3.5], dtype='float16'), array.ndarray)
+
+        scalar = sc.SimpleArrayFloat16((1,))
+        scalar[0] = ABOVE_HALF_MIDPOINT
+        self.assertEqual(0x3c01, scalar.ndarray.view('uint16')[0])
+
+        signaling_nan = np.array([0x7c01], dtype='uint16').view('float16')[0]
```
yungyuc, line 41: I guess the two-step conversion of `np.array([0x7c01], dtype='uint16').view('float16')` is necessary.  Could you please share why `np.array([0x7c01], dtype='float16')` does not work?

tests/test_float16.py
```diff
@@ -0,0 +1,133 @@
...
+        scalar[0] = signaling_nan
+        self.assertEqual(0x7c01, scalar.ndarray.view('uint16')[0])
+
+        with self.assertRaises(RuntimeError):
+            scalar[0] = np.complex64(1 + 2j)
+
+        collector = sc.SimpleCollectorFloat16()
+        for value in (1, 2.25, np.float16(-3.5)):
```
yungyuc, line 49: Better add a comment stating that `int`, `float`, and `np.float16` are accepted to make it obvious.  It's good to have all 3 types tested.

Numpy integer and other numpy float types are not tested.  They can be added in a followup PR or this one.

yungyuc (approved):
Let me know if you want this to be merged now or do more changes before merging.

----

I expect complex32 to use another ~~PR~~ issue #1459.

tests/test_float16.py
ThreeMonth03: `dtype='float16'` converts the integer value `31745`; it does not store the raw bits `0x7c01`:

```text
31745 --numeric conversion--> 31744 (bits 0x77c0)
0x7c01 --bit reinterpretation--> signaling NaN
```

The test needs the exact signaling-NaN representation, so it creates `uint16` storage first and reinterprets it with `.view('float16')`.

tests/test_float16.py
ThreeMonth03: This loop covers the three scalar paths used in this task:

```text
int        -> numeric conversion
float      -> numeric conversion
np.float16 -> exact two-byte copy
```

NumPy integer and other NumPy floating scalars can be covered in a follow-up PR.

PR #1368, opened 2026-08-16
Add resolution, run-ending, and CFL controls to the reflection panel

yungyuc:
For issue #1271

Add stop, reset, and remesh buttons for the Euler solver.  Remesh takes `nx` and `ny`.

Report CFL number as it runs.  Make reopening domain view to remember where it were.  Add a size grip for the domain view sub-window to make it easier to resize.  Reduce blank space in the fold boxes.

PR #1429, opened 2026-08-29
Let any sub-window carry a resize grip

### LINKED ISSUE #1272: Show the Field Against the Analytic Answer (polish issue 5)
Use a slim colorbar legend under the controls to show the colormap with the min, the max, and the analytic zone ticks on the same pinned range the viewer uses.

- PR #1374

Allow to record the frames and output a movie.  Movie formats include mp4, webp, and gif.

- PR #1375
- PR #1377
- PR #1424
  - PR #1425
- PR #1429

Use a separate sub-window to plot the line profile with the analytic three-zone overlay.  Use native plot code, e.g., PR #1231.  If it is not ready, develop something similar.

- PR #1431
- PR #1432

Window-lane tests check the pinned range, the tick values, and that the lines track the session.

yungyuc:
Allow to add a resize grip to all sub-window rather than just the 3D viewer (which was introduced with PR #1368).  `RManager::addSubWindowGrip` is now the one entry point that adds the grip but stays private to the translation unit. `add3DWidget` calls it.

For issue #1272.

### DIFF
--- a/cpp/solvcon/pilot/wrap_pilot.cpp
+++ b/cpp/solvcon/pilot/wrap_pilot.cpp
@@ -1210,6 +1210,12 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRManager
                     return subwin;
                 },
                 py::arg("widget"))
+            .def(
+                "addSubWindowGrip",
+                &wrapped_type::addSubWindowGrip,
+                py::arg("subwin"),
+                "Put a resize grip over a sub-window's lower-right corner. A "
+                "sub-window that already carries one is left alone.")
             .def_property(
                 "windowTitle",
                 [](wrapped_type & self)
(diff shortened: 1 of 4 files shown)


cpp/solvcon/pilot/wrap_pilot.cpp
```diff
@@ -1210,6 +1210,12 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapRManager
                     return subwin;
                 },
                 py::arg("widget"))
+            .def(
+                "addSubWindowGrip",
```
yungyuc, line 1214: Allow to control adding the grip from Python.

tests/test_pilot_domain_widget.py
```diff
@@ -2008,6 +2008,59 @@ def test_open_3d_keeps_native_window(self):
         self.assertEqual(int(mw.winId()), before)
 
 
+@unittest.skipUnless(solvcon.HAS_PILOT, "Qt pilot is not built")
+class RManagerSubWindowGripTC(unittest.TestCase):
```
yungyuc, line 2012: Test for adding the grip.

PR #1446, opened 2026-08-30
Add allocation-wide buffer synchronization

ThreeMonth03:
## Motivation

Device work can outlive its submitting call. CPU access through any array alias must wait for that work and exclude new device submissions until CPU computation ends.

## Method

Both guards run on CPU threads; the published token tracks device completion.

```text
CPU access:  HostLease  -> compute -> release lease
Device work: Submission -> order dependencies -> enqueue -> publish token
```

Each Metal allocation owns one state; CPU-only buffers allocate none. Guards borrow the state; `shared_ptr` retains completion tokens.

```mermaid
flowchart TB
    subgraph Access["Buffer ownership and borrowed access"]
        Array["Array / slices"] --> Buffer["ConcreteBuffer"]
        Buffer -->|owns| State["BufferAccessState<br/>Phase / lease count / completion"]
        Host["HostLease"] -.->|borrows| State
        Submit["Submission"] -.->|borrows| State
    end
    subgraph Tokens["DeviceCompletionToken ownership (shared_ptr)"]
        direction LR
        Waiter["Host acquisition"] -->|temporary| Prior["Prior token<br/>Wait before CPU use"]
        StateOwner["State"] -->|latest| Latest["Published token<br/>Track completion"]
        SubmitOwner["Submission"] -->|dependencies| Dependencies["Prior tokens<br/>Order device work"]
    end
    Access ~~~ Tokens
```

Separate token boxes may refer to the same instance. Tokens do not own buffer allocations.

| Caller | Existing access | Action |
| --- | --- | --- |
| `HostLease` | Submission reservation | Wait for reservation release |
| `HostLease` | Unfinished device work | Wait on token |
| `Submission` | CPU lease / reservation / host export | Throw |

CPU leases may coexist; callers synchronize conflicting CPU reads/writes. Publishing or abandoning a reservation releases it; abandonment alone cannot cancel device work.

The state mutex protects three separate records:

| Record | Meaning |
| --- | --- |
| Phase | Submission reservation or permanent host export |
| CPU lease count | New submissions require zero leases |
| Latest completion token | Prior operation for CPU waits and device ordering |

`Unreserved` does not mean idle: CPU leases or device work may still be active.

Host acquisition waits for reservation release, then prior device completion. This example follows the publication path through a future device executor:

```mermaid
sequenceDiagram
    participant Host as CPU host caller
    participant Submit as CPU submitter
    participant Device as Device/backend
    Submit->>Submit: Reserve buffers (mutex)
    Host->>Host: Acquire HostLease
    Note over Host: Wait for publication<br/>CV releases state mutex
    Submit->>Submit: Order dependencies (unlocked)
    Submit->>Device: Enqueue work
    activate Device
    Submit->>Submit: Publish token (mutex)
    Submit-->>Host: Unlock and notify
    Note over Submit: Submission may end
    Host->>Host: Recheck phase + count lease<br/>Copy token (mutex)
    Host->>Device: Wait on token (unlocked)
    Device-->>Host: Complete (writes visible)
    deactivate Device
    Note over Host: CPU / BLAS work (unlocked)<br/>All workers finish
    Host->>Host: Release lease (mutex)
```

Publication replaces the states' token and releases the guard's reservations. Example with one shared prior operation:

```text
                         before publish          after publish
states' completion       prior token             new token
guard dependencies       [prior token]           [prior token]
guard reserved states    [source, destination]   []
```

Multi-buffer submissions lock states in pointer order. Backend waits run unlocked; the token atomic rejects reuse. After enqueue, finish or safely cancel work before releasing an unpublished reservation.

- Copies and CPU matmul hold leases; callers join workers before releasing them.
- Raw pointer export permanently excludes device submissions; unchecked `data()` does not trigger export.
- The executor must retain buffers and order dependencies; its implementation is outside this PR.

## Validation

283 Metal-enabled C++ tests pass. The 13 state tests pass 100 shuffled runs under ThreadSanitizer with simulated device completions. `make lint` and `make pilot_clang_tidy_diff` pass.

Related to #1397, task 3.

### DIFF
--- a/cpp/solvcon/buffer/BufferAccess.cpp
+++ b/cpp/solvcon/buffer/BufferAccess.cpp
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+#include <solvcon/buffer/BufferAccess.hpp>
+
+#include <algorithm>
+#include <stdexcept>
+
+namespace solvcon
+{
+
+namespace detail
+{
+
+void BufferAccessState::begin_host_access()
+{
+    std::shared_ptr<DeviceCompletionToken> completion;
+    {
+        std::unique_lock lock(m_mutex);
+        m_submission_changed.wait(lock, [this]()
+                                  { return m_phase != AccessPhase::SubmissionReserved; });
+        ++m_host_access_count;
+        completion = m_last_completion;
+    }
+    try
+    {
+        wait_for_completion(completion);
+    }
+    catch (...)
+    {
+        // Construction failed, so no HostLease destructor will decrement the lease count.
+        end_host_access();
+        throw;
+    }
+}
+
+void BufferAccessState::wait_for_completion(std::shared_ptr<DeviceCompletionToken> const & completion)
+{
+    if (!completion)
+    {
+        return;
+    }
+    completion->wait();
+    std::scoped_lock const lock(m_mutex);
+    // A wait-only caller must not clear work published after its snapshot.
+    if (m_last_completion == completion)
+    {
+        m_last_completion.reset();
+    }
+}
+
+BufferAccessState::Submission::Submission(std::span<state_type * const> states)
+    : m_reserved_states(states.begin(), states.end())
+{
+    std::ranges::sort(m_reserved_states);
+    m_reserved_states.erase(std::unique(m_reserved_states.begin(), m_reserved_states.end()), m_reserved_states.end());
+    if (m_reserved_states.empty() || std::ranges::find(m_reserved_states, nullptr) != m_reserved_states.end())
+    {
+        throw std::invalid_argument("BufferAccessState::Submission: access states must be non-empty and non-null");
+    }
+
+    auto locks = lock_states();
+    for (state_type const * state : m_reserved_states)
+    {
+        if (state->m_phase != AccessPhase::Unreserved || state->m_host_access_count != 0)
+        {
+            throw std::runtime_error("BufferAccessState::Submission: buffer is unavailable for device submission");
+        }
+        if (state->m_last_completion && std::ranges::find(m_dependencies, state->m_last_completion) == m_dependencies.end())
+        {
+            m_dependencies.push_back(state->m_last_completion);
+        }
+    }
+    // No state is reserved until validation and dependency allocation have both succeeded.
+    for (state_type * state : m_reserved_states)
+    {
+        state->m_phase = AccessPhase::SubmissionReserved;
+    }
+}
+
+std::vector<std::unique_lock<std::mutex>> BufferAccessState::Submission::lock_states() const
+{
+    std::vector<std::unique_lock<std::mutex>> locks;
+    locks.reserve(m_reserved_states.size());
+    for (state_type * state : m_reserved_states)
+    {
+        locks.emplace_back(state->m_mutex);
+    }
+    return locks;
+}
+
+void BufferAccessState::Submission::publish(std::shared_ptr<completion_type> const & completion)
+{
+    {
+        auto locks = lock_states();
+        if (!completion || m_reserved_states.empty())
+        {
+            throw std::invalid_argument("BufferAccessState::Submission::publish: active guard and non-null token required");
+        }
+        if (!completion->try_claim())
+        {
+            throw std::invalid_argument("BufferAccessState::Submission::publish: token already claimed");
+        }
+        // Dependencies retain old tokens, preventing backend deletion under these locks.
+        for (state_type * state : m_reserved_states)
+        {
+            state->m_last_completion = completion;
+            state->m_phase = AccessPhase::Unreserved;
+        }
+    }
+    for (state_type * state : m_reserved_states)
+    {
+        state->m_submission_changed.notify_all();
+    }
+    // A later submission may already hold new reservations; this guard must leave them intact.
+    m_reserved_states.clear();
+}
+
+BufferAccessState::Submission::~Submission()
+{
+    for (state_type * state : m_reserved_states)
+    {
+        {
+            std::scoped_lock const lock(state->m_mutex);
+            state->m_phase = AccessPhase::Unreserved;
+        }
+        state->m_submission_changed.notify_all();
+    }
+}
+
+void BufferAccessState::wait()
+{
+    std::shared_ptr<DeviceCompletionToken> completion;
+    {
+        std::unique_lock lock(m_mutex);
+        m_submission_changed.wait(lock, [this]()
+                                  { return m_phase != AccessPhase::SubmissionReserved; });
+        completion = m_last_completion;
+    }
+    wait_for_completion(completion);
+}
+
+bool BufferAccessState::ready() const
+{
+    std::shared_ptr<DeviceCompletionToken> completion;
+    AccessPhase phase;
+    {
+        std::scoped_lock const lock(m_mutex);
+        phase = m_phase;
+        completion = m_last_completion;
+    }
+    return phase != AccessPhase::SubmissionReserved && (!completion || completion->ready());
+}
+
+void BufferAccessState::export_host_access()
+{
+    if (host_exported())
+    {
+        return;
+    }
+    HostLease const access(this);
+    std::scoped_lock const lock(m_mutex);
+    m_phase = AccessPhase::HostExported;
+}
+
+} /* end namespace detail */
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
--- a/cpp/solvcon/buffer/BufferAccess.hpp
+++ b/cpp/solvcon/buffer/BufferAccess.hpp
@@ -1,0 +1,301 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * Array aliases share memory: CPU access must wait for prior device work and exclude new submissions.
+ * HostLease and Submission run on CPU threads; DeviceCompletionToken tracks device completion.
+ * @code
+ * array + slices --> ConcreteBuffer --owns--> BufferAccessState
+ *                                              ^           ^
+ * HostLease --------------------borrows--------+           |
+ * Submission -------------------borrows--------------------+
+ *
+ * DeviceCompletionToken ownership (shared_ptr):
+ *   state -------------> latest operation
+ *   host acquisition --> prior operation (temporary, while waiting)
+ *   submission --------> prior operations (dependencies)
+ * @endcode
+ * Token arrows may share an instance. Tokens do not own buffer allocations.
+ * @ingroup group_core
+ */
+
+#include <atomic>
+#include <condition_variable>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <span>
+#include <vector>
+
+namespace solvcon
+{
+
+namespace detail
+{
+
+/**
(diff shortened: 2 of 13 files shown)


ThreeMonth03 (commented):
@yungyuc This pull request is still in progress. Please hold off on reviewing it for now. I will let you know when it is ready.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,255 @@
...
+    /// Wait for completion and make its writes host-visible.
+    virtual void wait() const = 0;
+    /// Claim this token for one submission. @return ``false`` when already claimed.
+    bool try_claim() noexcept { return !m_claimed.exchange(true, std::memory_order_relaxed); }
+
+private:
+    std::atomic<bool> m_claimed{false};
+}; /* end class BufferWorkToken */
+
+/// Coordinate host access and device work for every alias of one allocation. @ingroup group_core
```
ThreeMonth03: Retain tokens while waiting outside the state mutex.

```text
state  -> newer token
waiter -> prior token (still alive)
```

`shared_ptr` keeps the prior token alive across replacement. Reference counting does not reacquire the state mutex; it still has a synchronization cost. Buffer lifetime remains the executor's responsibility.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,255 @@
...
+class BufferWorkToken // NOLINT(cppcoreguidelines-special-member-functions)
+{
+public:
+    virtual ~BufferWorkToken() = default;
+    /// Test whether the work has completed. @return ``true`` after completion.
+    virtual bool ready() const = 0;
+    /// Wait for completion and make its writes host-visible.
+    virtual void wait() const = 0;
+    /// Claim this token for one submission. @return ``false`` when already claimed.
+    bool try_claim() noexcept { return !m_claimed.exchange(true, std::memory_order_relaxed); }
```
ThreeMonth03, line 58: A completion token describes one GPU operation. Reusing an old token could make unfinished work appear complete.

```text
first GPU operation -> publish its completion token -> accepted
later GPU operation -> reuse that same token        -> rejected
later GPU operation -> publish a fresh token        -> accepted
```

`try_claim()` enforces this rule even when CPU threads race to publish the same token. Publishing a token does not mean its GPU operation has finished.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            throw std::runtime_error("BufferAccessState::Submission: buffer is unavailable for device submission");
+        }
+        if (state->m_last_work && std::ranges::find(m_dependencies, state->m_last_work) == m_dependencies.end())
+        {
+            m_dependencies.push_back(state->m_last_work);
+        }
+    }
+    std::ranges::for_each(m_states, [this](auto const * state)
+                          { state->m_pending_submission = m_reservation; });
```
ThreeMonth03: Buffers stay reserved while the executor prepares and queues the next GPU operation.

```text
collect distinct buffers
  -> record their previous GPU operations
  -> reserve the buffers
  -> executor orders and queues the new GPU operation
  -> publish its completion token
  -> release the reservations
```

During this interval, CPU access waits and another device submission using these buffers is rejected. This prevents CPU access between checking the previous work and recording the new work.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,255 @@
...
+    std::shared_ptr<BufferWorkToken> const work = m_last_work;
+    lock.unlock();
+    try
+    {
+        if (work)
+        {
+            work->wait();
+            std::scoped_lock const work_lock(m_mutex);
+            // A wait-only caller must not clear work published after its snapshot.
+            if (m_last_work == work)
```
ThreeMonth03: An old waiter only clears the token it observed:

```text
wait snapshots W1 -> W2 is published -> W1 completes
                                        W2 remains current
```

The identity check preserves newer device work.

cpp/solvcon/buffer/ConcreteBuffer.hpp
```diff
@@ -231,12 +234,30 @@ class ConcreteBuffer
...
+
+    /// Snapshot current completion without reserving host access. @return ``true`` when nothing observed is pending.
+    bool ready() const { return m_access_state == nullptr || m_access_state->ready(); }
+    /// Return the synchronization state. @return Borrowed state valid for this buffer's lifetime.
+    access_state_type const * access_state() const noexcept { return m_access_state.get(); }
+
+    /** Acquire recoverable host access for one internal operation.
+     * @return Lease that blocks device submission until destruction.
+     * @note This buffer must outlive the returned lease. */
+    host_access_type acquire_host_access() const { return host_access_type(m_access_state.get()); }
```
ThreeMonth03, line 248: Scoped CPU access and an exported raw pointer have different lifetimes.

```text
internal CPU operation:
  acquire HostLease (wait for GPU work) -> use memory -> destroy HostLease

raw pointer export:
  wait for GPU work -> permanently reject later device submissions
```

The raw pointer can outlive the call, so its export must call `export_host_access()`. Reading an unchecked raw accessor alone does not perform this transition. CPU-only storage needs neither synchronization path.

cpp/solvcon/buffer/SimpleArray.cpp
```diff
@@ -165,7 +165,9 @@ SimpleArrayCopier::SimpleArrayCopier(
     shape_type const & dst_stride,
     shape_type const & shape,
     size_t const itemsize)
-    : m_src(src_buffer.data<int8_t>() + src_body_offset)
+    : m_src_access(src_buffer.acquire_host_access())
```
ThreeMonth03, line 168: The copy must protect both the memory it reads and the memory it writes.

```text
source: wait for previous GPU work and hold a host lease
destination: wait for previous GPU work and hold a host lease
both ready -> capture pointers -> copy elements -> release both leases
```

Both leases are acquired before either pointer is captured. They reject new device submissions on these buffers until the copy scope ends.

gtests/test_nopython_buffer.cpp
```diff
@@ -98,6 +148,94 @@ TEST(ConcreteBuffer, iterator)
...
+    auto first_wait = std::async(std::launch::async, &buffer_access_state_type::wait, &first);
+    ASSERT_TRUE(wait_started_or_complete(*previous));
+    buffer_submission_type submission(std::span{accesses});
+    ASSERT_EQ(size_t{1}, submission.dependencies().size());
+    EXPECT_EQ(previous, submission.dependencies().front());
+    auto next = std::make_shared<BlockingBufferWork>(first);
+    submission.publish(next);
+    previous->complete();
+    first_wait.get();
+    EXPECT_FALSE(first.ready());
```
ThreeMonth03, line 169: Finishing an older GPU operation must not erase a newer operation recorded on the same buffer.

```text
CPU thread starts waiting for the previous GPU operation
  -> a newer GPU operation is published on the same buffer
  -> the previous GPU operation completes
  -> the CPU wait returns
  -> ready() must still report the newer operation as pending
```

The test uses fake device work to force this sequence. It also checks that a completed token cannot be reused and that a rejected publication can retry with a fresh token.

cpp/solvcon/buffer/BufferAccess.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+    std::unique_lock lock(m_mutex);
+    while (m_pending_submission && m_pending_submission->load(std::memory_order_acquire))
+    {
+        std::shared_ptr<std::atomic<bool>> const reservation = m_pending_submission;
+        lock.unlock();
+        reservation->wait(true, std::memory_order_acquire);
+        lock.lock();
+    }
+    m_host_access_count += static_cast<size_t>(reserve_host_access);
```
ThreeMonth03: CPU access must block new device submissions before it waits for the previous GPU operation.

```text
reserve host access -> wait for previous GPU work -> use memory on CPU

if the wait throws:
  remove the host reservation -> propagate the error
```

Otherwise, another CPU thread could submit device work before host access begins. The error path releases the reservation explicitly because `~HostLease()` does not run when construction throws.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,294 @@
...
+     */
+    ~Submission() { release_reservation(); }
+    /**
+     * Expose the prior work captured while reserving states; this function does not wait.
+     * @code
+     * dependency tokens -> executor establishes ordering -> new device work
+     * @endcode
+     * @return Borrowed span of unique tokens, valid only while this Submission exists.
+     */
+    std::span<std::shared_ptr<completion_type> const> dependencies() const { return m_dependencies; }
```
ThreeMonth03, line 220: All submissions acquire distinct state mutexes in the same order.

```text
[source, destination] -> sort + deduplicate -> lock in pointer order
[destination, source] -> sort + deduplicate -> same lock order
```

Backend waits and token destruction run unlocked. This prevents state-lock cycles. Executors must wait on `dependencies()`, not their own reserved states, which would self-wait.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,294 @@
...
+}; /* end class BufferAccessState::HostLease */
+
+/**
+ * Reserve allocation states for one device submission.
+ * @code
+ * construct -> dependencies() -> executor orders/enqueues work -> publish(new token)
+ * abandon before enqueue -> destructor releases the reservation
+ * @endcode
+ * The executor must honor dependencies and keep allocations alive until device work finishes.
+ * Destruction releases an unpublished reservation; it cannot cancel already enqueued device work.
```
ThreeMonth03: Cleanup depends on whether device work was enqueued and published.

```text
HostLease construction fails -> undo lease count
abandon before enqueue      -> release reservation
publish fails after enqueue -> finish / safely cancel, then release
publish succeeds            -> clear reserved states, retain prior tokens
```

A published guard releases only its dependencies on destruction. Its empty reservation list leaves later submissions alone. An unpublished guard cannot cancel queued device work.

ThreeMonth03 (commented):
@yungyuc Please review this pull request. It might look a little complex.

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,294 @@
...
+     *      -> wait for token -> lock, clear token only if still current, unlock
+     * wait throws -> undo provisional host count if any -> rethrow
+     * @endcode
+     * Counting the lease before waiting prevents another device submission from entering the host-access gap.
+     * @param reserve_host_access True for a lease; false for a wait-only caller that allows new submissions.
+     */
+    void synchronize(bool reserve_host_access);
+
+    mutable std::mutex m_mutex; ///< Guards token, gate pointer, and host count; unlocked for backend waits.
+    std::shared_ptr<DeviceCompletionToken> m_last_completion; ///< Latest published work; null means none recorded.
```
yungyuc: `std::shared_ptr` already uses a lock for its counter.  Does it make sense to use a shared pointer to hold the token?

cpp/solvcon/buffer/BufferAccess.hpp
```diff
@@ -0,0 +1,294 @@
...
+    std::shared_ptr<std::atomic<bool>> m_pending_submission; ///< Shared gate; true blocks host access and new submissions.
```
yungyuc: It looks very strange to use a `std::shared_ptr` (which has a counter with lock) to hold a `bool`.  Please explain clearly how this works.  I am not sure how to take an address of a `bool` and how the address of `bool` works with a shared pointer.

yungyuc (changes requested):
@ThreeMonth03 I feel the code is OK but will need elaboration from you.

- [ ] Elaborate how `std::shared_ptr` is used for tracking the state.  This probably need an interactive discussion to clarify.

Issue #1445, opened 2026-08-30
Fix `SimpleArray` assignment and fill for overlapping and strided views


ThreeMonth03:
## Problem

`SimpleArray` has three layout bugs in NumPy assignment and `fill()`. The following `Float64` examples currently produce wrong results without raising an exception.

### Overlapping assignment reads overwritten values

```python
import numpy as np
import solvcon as sc

array = sc.SimpleArrayFloat64(
    array=np.arange(1, 5, dtype="float64")
)
array[1:] = array.ndarray[:-1]

print(array.ndarray)
# Current:  [1. 1. 1. 1.]
# Expected: [1. 1. 2. 3.]
```

The right-hand side and destination share storage. Assignment must read all right-hand-side values before writing the destination.

### Assignment misreads NumPy byte strides

```python
raw = bytearray([
    0, 0, 0, 0, 0, 0, 240, 63,
    64, 0, 0, 0, 0, 0, 0, 0,
])
source = np.ndarray(
    (2,), dtype="float64", buffer=raw, strides=(1,)
)
destination = sc.SimpleArrayFloat64(shape=(2,), value=-1)
destination[...] = source

print(source)
# [1.     31.9375]

print(destination.ndarray)
# Current:  [1. 1.]
# Expected: [1.     31.9375]
```

NumPy strides count bytes, while `SimpleArray` strides count elements. Assignment currently truncates a one-byte stride to a zero-element stride. It must read valid NumPy sources by byte address without unaligned typed access, including non-itemsize-multiple strides and unaligned data pointers.

### `fill()` writes outside the logical view

```python
base = np.arange(12, dtype="float64").reshape(3, 4)
array = sc.SimpleArrayFloat64(array=base[:, ::2])
array.fill(-2)

print(base)
# Current:
# [[-2. -2. -2. -2.]
#  [-2. -2.  6.  7.]
#  [ 8.  9. 10. 11.]]
#
# Expected:
# [[-2.  1. -2.  3.]
#  [-2.  5. -2.  7.]
#  [-2.  9. -2. 11.]]
```

`fill()` must follow the view shape and strides and must not modify storage outside the logical view.

## Task

- [x] **Make `SimpleArray` assignment and `fill()` layout-safe.** Read valid byte-strided sources without unaligned typed access, stage the right-hand side when it overlaps destination writes, and make `fill()` visit only logical view addresses. Ordering among destination indices that share an address is outside this task. (#1448)

## Acceptance

[...]

PR #1448, opened 2026-08-30
Make `SimpleArray` assignment and `fill()` layout-safe

ThreeMonth03:
`SimpleArray` assignment could misread NumPy byte strides or reuse overwritten source values. `fill()` could modify storage outside a strided view. The corrected behavior is:

```python
import numpy as np
import solvcon as sc

values = np.arange(1, 5, dtype="float64")
array = sc.SimpleArrayFloat64(array=values)
array[1:] = array.ndarray[:-1]
# values: [1. 1. 2. 3.]

base = np.arange(6, dtype="float64")
view = sc.SimpleArrayFloat64(array=base[::2])
view.fill(-1)
# base: [-1. 1. -1. 3. -1. 5.]
```

Assignment now reads NumPy values by byte address and stages every source value before writing if storage may overlap. `fill()` follows logical shape and signed strides while retaining dense C- and F-contiguous fast paths.

Regression tests cover overlapping assignment, non-standard source strides, and strided `fill()` views. `make lint` and the full buffer suite pass.

Related to #1445.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -13,6 +13,7 @@
  */
 
 #include <solvcon/buffer/ConcreteBuffer.hpp>
+#include <solvcon/buffer/loop.hpp>
 #include <solvcon/buffer/matmul.hpp>
 #include <solvcon/buffer/signed_stride_layout.hpp>
 #include <solvcon/math/math.hpp>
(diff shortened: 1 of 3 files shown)


cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -32,95 +59,122 @@ inline solvcon::detail::shape_type shape_from_slices(
...
-        ssize_t offset = 0;
-        for (pybind11::ssize_t axis = 0; axis < arr_in.ndim(); ++axis)
-        {
-            offset += arr_in.strides(axis) / arr_in.itemsize() * sidx[axis];
-        }
-        return offset;
+        data += strides[axis] * sidx[axis];
     }
+    D value;
+    std::memcpy(&value, data, sizeof(value));
```
ThreeMonth03, line 83: NumPy strides are byte offsets, so address calculation deliberately stays in `char *` space:

```text
address(i, j) = data + i * strides[0] + j * strides[1]
```

`memcpy` then reads the value without forming a potentially unaligned `D *`.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -32,95 +59,122 @@ inline solvcon::detail::shape_type shape_from_slices(
...
+    constexpr bool valid_conversion =
+        (!is_complex_v<T> && !is_complex_v<D>) ||
+        (is_complex_v<T> && is_complex_v<D> && std::is_same_v<T, D>);
+
+    if constexpr (valid_conversion)
+    {
+        shape_type const output_shape = detail::shape_from_slices(slices);
+        auto const * data = static_cast<char const *>(arr_in.data());
+        auto const * strides = arr_in.strides();
+        if (may_overlap(arr_out, arr_in))
```
ThreeMonth03, line 143: The overlap check uses bounding byte envelopes, so a sparse view may take this path conservatively. Once here, staging forms a read/write barrier:

```text
RHS bytes -> convert every value -> staged
staged    -> write every output
```

No destination write can change a source byte that has not been read yet.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -258,14 +259,33 @@ class SimpleArrayMixinModifiers
...
+        std::fill_n(athis->logical_data(), size, value);
         return *athis;
     }
 
-}; /* end class SimpleArrayMixinModifiers */
+    LoopDomain const domain(athis->shape());
+    MappedOffsetCursor::mapping_type const mappings{OperandMapping(athis->stride())};
+    for (MappedOffsetCursor cursor(domain, mappings); cursor; cursor.advance())
+    {
+        athis->logical_data()[cursor.offset(size_t{0})] = value;
```
ThreeMonth03, line 285: A non-dense view is traversed by logical coordinates, not by its buffer interval:

```text
shape=(3, 2), stride=(4, 2)
element offsets -> 0, 2, 4, 6, 8, 10
```

`logical_data()` also keeps ghost entries in the fill range; dense C- and F-contiguous layouts stay on the fast path above.

yungyuc (approved):
LGTM.  The code is clear.

PR #1452, opened 2026-08-30
Prepare one `matmul` comparison

ThreeMonth03:
## Motivation

The Route Inspector needs reproducible operands and numerical differences from NumPy before it compares kernel runtime.

## Approach

Generate deterministic components in bounded FP64 chunks, then cast them into the target real or complex storage:

```text
bounded FP64 chunks
        |
        v
target real-component dtype
        |
        v
float or interleaved complex storage
        |
        v
SimpleArray.typed_class(dtype)
```

This gives every dtype the same component stream without a dtype-specific RNG table. Construct NumPy views with the requested shape, dtype, and element strides. The registered typed arrays share operand storage, so the collector keeps no duplicate dtype-to-class table and performs no input copy.

Let C++ report an unavailable forced kernel through a typed Python exception.

```text
execution result
|-- unavailable                       -> ineligible
|-- wrong shape, dtype, or non-finite -> invalid
`-- otherwise                         -> measured
```

Each non-empty `measured` result reports its maximum absolute difference from NumPy. It also reports that difference relative to NumPy's maximum magnitude. The collector applies no tolerance or numerical pass/fail rule.

Unexpected allocation and execution failures propagate. This PR adds no warmup, timing, storage, runtime limit, or memory limit.

Related to #1369.

### DIFF
--- a/cpp/solvcon/buffer/matmul_executor.hpp
+++ b/cpp/solvcon/buffer/matmul_executor.hpp
@@ -32,6 +32,17 @@
 namespace solvcon
 {
 
+/**
+ * Report that a forced matmul kernel is unavailable for this call.
+ *
+ * @ingroup group_core
+ */
+class MatmulKernelUnavailable : public std::invalid_argument
+{
+public:
+    using std::invalid_argument::invalid_argument;
+}; /* end class MatmulKernelUnavailable */
+
 namespace detail
 {
 
@@ -316,8 +327,7 @@ void MatmulExecutor<Array>::execute(MatmulKernel kernel)
         std::string_view const reason = use_matmul_blas_v<value_type>
                                             ? "is not eligible for these operands"
                                             : "requires a BLAS backend";
-        throw std::invalid_argument(
-            std::format("matmul(): kernel '{}' {}", matmul_kernel_name(kernel), reason));
+        throw MatmulKernelUnavailable(std::format("matmul(): kernel '{}' {}", matmul_kernel_name(kernel), reason));
     }
     run(selection.value());
 }
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
@@ -240,6 +240,34 @@ static SimpleArrayPlex make_arrayplex_from_numpy(pybind11::array & arr_in)
 }
 
 // NOLINTNEXTLINE(misc-use-anonymous-namespace)
+static pybind11::type get_typed_array_class(std::string const & dtype)
+{
+#define SC_DECL_GET_TYPED_ARRAY_CLASS(DataTypeValue, ArrayType) \
+    case DataTypeValue: return pybind11::type::of<ArrayType>();
+
+    switch (DataType(dtype))
+    {
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Bool, SimpleArrayBool)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Int8, SimpleArrayInt8)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Int16, SimpleArrayInt16)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Int32, SimpleArrayInt32)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Int64, SimpleArrayInt64)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Uint8, SimpleArrayUint8)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Uint16, SimpleArrayUint16)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Uint32, SimpleArrayUint32)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Uint64, SimpleArrayUint64)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Float16, SimpleArrayFloat16)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Float32, SimpleArrayFloat32)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Float64, SimpleArrayFloat64)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Complex64, SimpleArrayComplex64)
+        SC_DECL_GET_TYPED_ARRAY_CLASS(DataType::Complex128, SimpleArrayComplex128)
+    default: throw std::invalid_argument("Unsupported datatype");
+    }
+
+#undef SC_DECL_GET_TYPED_ARRAY_CLASS
+}
+
+// NOLINTNEXTLINE(misc-use-anonymous-namespace)
 static void verify_calculator_supported(SimpleArrayPlex const & array)
 {
     if (array.data_type() == DataType::Float16)
@@ -295,6 +323,7 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArrayPlex : public WrapBase<Wr
             ;
 
         (*this)
+            .def_static("typed_class", &get_typed_array_class, pybind11::arg("dtype"))
             .def("clone",
                  [](wrapped_type const & self)
                  { return wrapped_type(self); }) // cloning the object using the copy constructor. never add the clone method to the C++ class.
--- a/solvcon/benchmark/collector.py
+++ b/solvcon/benchmark/collector.py
@@ -1,0 +1,160 @@
+# Copyright (c) 2026, solvcon team <contact@solvcon.net>
+# BSD 3-Clause License, see COPYING
+
+"""Prepare exact operands and compare matmul kernel results."""
+
+import numpy as np
+
+import solvcon as sc
+
+
+_CHUNK_SIZE = 1 << 20
+
+
+def _storage_bounds(operand):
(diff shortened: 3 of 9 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3062,6 +3064,37 @@ SimpleArray<T> SimpleArray<T>::to_row_major() const
...
+ *
+ * @details
+ *      Unlike to_row_major(), this operation always allocates storage
+ *      containing exactly the elements in shape(). Every stored row,
+ *      including a ghost row, becomes an ordinary row in the result.
+ *
+ * @return Freshly allocated C-contiguous SimpleArray without ghost metadata.
+ */
+template <typename T>
+SimpleArray<T> SimpleArray<T>::to_compact_row_major() const
```
ThreeMonth03: This is the packing boundary, not a replacement for `to_row_major()`. The existing method may preserve extra backing storage and ghost metadata, while a packed operand must own exact zero-offset storage.

```text
source  shape=(2,2), strides=(-257,-128), nghost=1
        [[1,2],[3,4]]
packed  shape=(2,2), strides=(2,1), nghost=0
        buffer=[1,2,3,4]
```

cpp/solvcon/buffer/matmul_executor.hpp
```diff
@@ -662,11 +662,11 @@ void MatmulExecutor<Array>::pack(PackingState const & packing)
 {
     if (packing.lhs)
     {
-        m_packed_lhs.emplace(m_lhs.to_row_major());
+        m_packed_lhs.emplace(m_lhs.to_compact_row_major());
```
ThreeMonth03: `MatmulExecutor` now consumes one public packing contract for both operands. Layout policy remains in `SimpleArray`, so the executor needs neither `friend` access nor a ghost-specific branch.

```text
strided / offset / ghosted
            |
            v
exact C-order array -> rebuild MatmulPlan -> execute
```

solvcon/benchmark/collector.py
```diff
@@ -0,0 +1,141 @@
...
+    storage = np.empty(maximum - minimum + 1, dtype=dtype.name)
+    real_name = (
+        'float32' if dtype.name in ('float32', 'complex64')
+        else 'float64')
+    components = storage.view(real_name)
+    generator = np.random.default_rng(seed)
+    generator.random(components.shape, dtype=real_name, out=components)
+    components *= 2
+    components -= 1
+    return np.ndarray(
```
ThreeMonth03, line 47: The smallest reachable storage interval preserves the requested element strides without an intermediate contiguous copy. Negative, zero, and overlapping strides therefore reach NumPy and `SimpleArray` unchanged.

```text
shape=(2,3), strides=(-4,0), origin=storage[4]
row 0 -> storage[4,4,4]
row 1 -> storage[0,0,0]
```

solvcon/benchmark/collector.py
```diff
@@ -0,0 +1,141 @@
...
+    if not np.allclose(
+            result, reference, rtol=tolerance, atol=tolerance):
+        return {'status': 'incorrect', 'reason': 'values differ'}
+    return {'status': 'passed', 'reason': None}
+
+
+def _check(spec, execute):
+    """Classify every requested kernel and the NumPy reference."""
+
+    reference = np.atleast_1d(execute('numpy'))
```
ThreeMonth03, line 132: NumPy is evaluated once, then every requested kernel is classified independently. Only an unavailable forced kernel is caught; unexpected execution and allocation failures still propagate.

```text
NumPy reference
  |-- shape/dtype mismatch -> raise
  |-- non-finite           -> all invalid
  `-- valid                -> measured / invalid / ineligible
```

tests/test_matrix.py
```diff
@@ -639,6 +639,34 @@ def test_matmul_with_strided_vectors(self):
...
+
+    def test_matmul_packs_stored_ghosts(self):
+        """Packing preserves ghost rows after a strided layout mutation."""
+        dtype = np.dtype(self.dtype).name
+        lhs_data = self.make_strided_view(
+            np.arange(8, dtype=dtype).reshape((4, 2)) + 1, 1, -128)
+        rhs_data = np.arange(4, dtype=dtype) + 1
+        lhs = self.SimpleArray(array=lhs_data)
+        lhs.nghost = 4
+        lhs.transpose()
```
ThreeMonth03: The transpose changes the stored shape from `(4,2)` to `(2,4)` while `nghost` remains 4. Forced GEMV verifies that packing copies the stored layout instead of interpreting ghost metadata through a nested special case.

```text
(4,2), nghost=4 --transpose--> (2,4), nghost=4
                                     |
                                     `--pack--> compact (2,4)
```

cpp/solvcon/buffer/pymod/wrap_SimpleArrayPlex.cpp
```diff
@@ -239,6 +239,34 @@ static SimpleArrayPlex make_arrayplex_from_numpy(pybind11::array & arr_in)
 #undef SC_DECL_MAKE_ARRAYPLEX
 }
 
+// NOLINTNEXTLINE(misc-use-anonymous-namespace)
+static pybind11::type get_typed_array_class(std::string const & dtype)
```
ThreeMonth03, line 243: One C++ registry selects the typed Python wrapper:

```text
NumPy dtype -> SimpleArray.typed_class(dtype) -> SimpleArray<T>
```

The collector needs no second Python dtype table.

solvcon/benchmark/collector.py
```diff
@@ -0,0 +1,142 @@
...
+        dtype=dtype.name,
+        buffer=storage,
+        offset=-minimum * dtype.itemsize,
+        strides=tuple(stride * dtype.itemsize for stride in operand.strides),
+    )
+
+
+class _MatmulExecutor:
+    def __init__(self, spec, lhs, rhs):
+        array_type = sc.SimpleArray.typed_class(spec.dtype)
```
ThreeMonth03, line 58: Both paths consume the same allocation:

```text
NumPy ndarray -- shared storage --> SimpleArray<T>
      |                               |
  np.matmul()                   forced kernel
```

Shape and strides reach both implementations without a contiguous copy.

solvcon/benchmark/collector.py
```diff
@@ -0,0 +1,142 @@
...
+        return self._native_lhs.matmul(self._native_rhs, kernel=name)
+
+
+def _make_executor(spec):
+    lhs = _make_operand(spec.lhs, spec.dtype, 0)
+    rhs = _make_operand(spec.rhs, spec.dtype, 1)
+    return _MatmulExecutor(spec, lhs, rhs)
+
+
+def _difference_metrics(result, reference):
```
ThreeMonth03, line 76: The two metrics answer different questions:

```text
max_abs_diff = max(|result - NumPy|)
relative_diff = max_abs_diff / max(|NumPy|)
```

The first reports raw drift; the second normalizes it to the reference scale. `np.nditer` reduces both in `_CHUNK_SIZE` buffers, so comparison adds bounded scratch instead of output-sized subtraction and absolute-value temporaries.

tests/test_benchmark_collector.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class OperandConstructionTC(unittest.TestCase):
+    def test_quantizes_shared_component_stream(self):
+        operand = benchmark_spec.OperandSpec.from_dict({
+            'shape': [2, 3], 'strides': [3, 1],
+        })
+        expected = 2 * np.random.default_rng(7).random(
+            12, dtype='float64') - 1
+        with unittest.mock.patch.object(
```
ThreeMonth03, line 50: The expected stream is generated in one call, while `_make_operand()` is forced through multiple chunks. This catches RNG restarts at chunk boundaries and proves real and complex operands are quantizations of the same FP64 component prefix.

cpp/solvcon/buffer/matmul_executor.hpp
```diff
@@ -32,6 +32,17 @@
 namespace solvcon
 {
 
+/**
+ * Report that a forced matmul kernel is unavailable for this call.
+ *
+ * @ingroup group_core
+ */
+class MatmulKernelUnavailable : public std::invalid_argument
```
ThreeMonth03, line 40: Forced-kernel selection has one typed failure boundary:

```text
unknown kernel                         -> ValueError
known kernel unavailable for this call -> MatmulKernelUnavailable(ValueError)
```

The collector catches only the second path, so unexpected allocation and execution failures still propagate.

tests/test_benchmark_collector.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def test_rejects_invalid_numpy_reference(self):
+        cases = (
+            (np.ones((1, 4), dtype='float64'), 'shape does not match'),
+            (np.ones((2, 2), dtype='float32'), 'dtype does not match'),
+        )
+        for reference, reason in cases:
+            with self.subTest(reason=reason):
+                execute = FakeExecutor({'numpy': reference})
+                with self.assertRaisesRegex(RuntimeError, reason):
```
ThreeMonth03, line 160: Each case names the validation contract it violates. The regex checks the exact reason, while the call log proves validation stops after the single NumPy oracle call and never reaches a native kernel.

ThreeMonth03 (commented):
- [x] Generate reproducible operands in bounded chunks and share their storage between NumPy and typed `SimpleArray`.
- [x] Report bounded-memory absolute and relative numerical differences without imposing a tolerance.
- [x] Distinguish unavailable forced kernels from unexpected execution failures and cover the boundary with tests.

@yungyuc Please take a look.

yungyuc (approved):
LGTM.

It would save some minutes for reviewing the PR to make it clearer that the work is to prepare for collecting benchmark comparison.

PR #1455, opened 2026-08-31
Fix `TypeBroadcast` clang-tidy failures

ThreeMonth03:
The merge of #1448 exposed clang-tidy 22 failures in `TypeBroadcast`. The signed-character suppressions covered only the first lines of multiline expressions, while the diagnostics were emitted on the following lines. After those diagnostics were removed, the full template build also reported that the object filled by `memcpy` was not value-initialized.

Move each suppression onto its diagnostic line and value-initialize the object before `memcpy`. The numerical conversion remains signed because casting NumPy `int8` through `unsigned char` would change negative values.

The int, uint, and complex wrapper translation units pass clang-tidy 22 with warnings as errors. `make lint` also passes.

Follow-up to #1448 and the failing master run: https://github.com/solvcon/solvcon/actions/runs/33389762749

Related to #1445.

### DIFF
--- a/cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
+++ b/cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
@@ -79,7 +79,7 @@ D TypeBroadcastImpl<T, D>::input_at(char const * data, pybind11::ssize_t const *
     {
         data += strides[axis] * sidx[axis];
     }
-    D value;
+    D value{};
     std::memcpy(&value, data, sizeof(value));
     return value;
 }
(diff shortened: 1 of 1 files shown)


ThreeMonth03 (commented):
> - [x] Move both signed-character suppressions onto their diagnostic lines.
> - [x] Value-initialize the object filled by `memcpy`.
> - [x] Verify the int, uint, and complex wrappers with clang-tidy 22 and run `make lint`.

@yungyuc Please take a look.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -79,7 +79,7 @@ D TypeBroadcastImpl<T, D>::input_at(char const * data, pybind11::ssize_t const *
     {
         data += strides[axis] * sidx[axis];
     }
-    D value;
+    D value{};
```
ThreeMonth03, line 82: `D value{}` satisfies `cppcoreguidelines-pro-type-member-init` before `memcpy` overwrites the complete trivially copyable object. The value returned from the byte load is unchanged.

cpp/solvcon/buffer/pymod/TypeBroadcast.hpp
```diff
@@ -146,9 +146,9 @@ void TypeBroadcastImpl<T, D>::broadcast(
             size_t staged_index = 0;
             detail::for_each_index(output_shape, [&](shape_type const & sidx)
                                    {
-                                       // FIXME: NOLINTNEXTLINE(bugprone-signed-char-misuse,cert-str34-c)
                                        staged[staged_index++] = static_cast<T>(
-                                           input_at(data, strides, sidx)); });
+                                           input_at(data, strides, sidx)); // FIXME: NOLINT(bugprone-signed-char-misuse,cert-str34-c)
```
ThreeMonth03, line 150: NumPy `int8_t` is numeric data, so widening must preserve negative values; casting through `unsigned char` would change `-1` to `255`. Matching same-line suppressions now cover the diagnostic on `input_at(...)` in both the staged and direct paths.

yungyuc (approved):
LGTM

Issue #994, opened 2026-07-04
Improve visualizing of vectors


yungyuc:
Pilot needs to visualize vector field.  Issue #976 added basic support for visualizing mesh surface normal vector.  We will need to polish the visualization.

c1ydehhx:
After take a look about #976, I think we can start polish normal vector by support scale feature. In #976 promise `widget.showNormals(on, scale)` API, but currently only have `widget.showNormals(on)`. It will be a good chance to realize how normal vector works, and may have more idea about how to polish.

c1ydehhx:
> After take a look about [#976](https://github.com/solvcon/solvcon/issues/976), I think we can start polish normal vector by support scale feature. In [#976](https://github.com/solvcon/solvcon/issues/976) promise `widget.showNormals(on, scale)` API, but currently only have `widget.showNormals(on)`. It will be a good chance to realize how normal vector works, and may have more idea about how to polish.

#1456 already set. The scale feature of normal vector already set and waiting for the review. 
Now I'm thinking is it possible to have glyphs. Currently I have no idea what scenerio the normal vector will have glyphs.

PR #1456, opened 2026-08-31
Allow normal vector scaling

c1ydehhx:
In this PR, according to the #994, we found that the current version of API still not have scale feature on normal vector. Therefore, I implement these:

 - `showNormals` supports the scale parameter to scale the normal vector.
 - GUI supports with a slider to scale the normal vector. The minimal ratio can be x0.01 and maximal ratio can be x3.0.

image

image

Moreover, the normal vector will scale automatically according to the dense of faces. We measure the spatial of 2D/3D mesh from the bounding box, and divide the spatial by the number of faces to get the average spatial for each faces. After get the average spatial of faces, we can take (d)-dimension root to get the approxmately length of faces as candidiate length. For the length of normal vector, we take 30% length of candidiate length in default.

We added 3 tests to test the GUI and the mechanism of recommended scale:
 - `test_recommended_normal_scale_follows_mesh_density` in `test_tree_panel.py`: Test the lenght of normal vector will be scaled by the dense of faces. We prepare two different dense of mesh to compare the length of normal vector. The mesh with more dense faces will have smaller faces.
 - `test_normal_scale_slider_drives_viewer` in `test_tree_panel.py`: Test the slider in GUI will changes the length of normal vector.
 - `test_normals_scale_changes_arrow_length` in `test_pilot_domain_widget.py`: Test the scale values will change the length of normal vector. The test compared the length of normal vector in small and larger scale value.

Thanks codex for the GUI implementation.

### DIFF
--- a/solvcon/pilot/panel/_tree_panel.py
+++ b/solvcon/pilot/panel/_tree_panel.py
@@ -14,8 +14,8 @@ from PySide6.QtGui import QColor, QPainter
 from PySide6.QtWidgets import (QWidget, QVBoxLayout, QTreeWidget,
                                QTreeWidgetItem, QFrame, QDockWidget,
                                QStackedWidget, QHBoxLayout, QButtonGroup,
-                               QRadioButton, QPushButton,
-                               QSizePolicy, QAbstractButton)
+                               QRadioButton, QPushButton, QSizePolicy,
+                               QAbstractButton, QSlider)
 
 from ... import core
 from .._style import PaletteStyled
@@ -99,6 +99,13 @@ class MeshInfoTree(TreePanelBase):
     _ROLE_IBC = Qt.UserRole + 1
     _ROLE_STYLE = Qt.UserRole + 2
 
+    _NORMAL_SCALE_MIN = 10
+    _NORMAL_SCALE_MAX = 300
+    _NORMAL_SCALE_DEFAULT = 100
+    _NORMAL_SCALE_SLIDER_WIDTH = 150
+    _NORMAL_SCALE_BASE_FRACTION = 0.04
+    _NORMAL_SCALE_SPACING_FRACTION = 0.30
+
     # Map cell type numbers to human-readable names.
     CELL_TYPE_NAME = {
         core.StaticMesh.POINT: "point",
@@ -117,6 +124,11 @@ class MeshInfoTree(TreePanelBase):
         if self.style_status is not None:
             self.style_status.changed.connect(self.refresh_style_checks)
         self._style_items = {}
+        self._normal_item = None
+        self._normal_scale_item = None
+        self._normal_scale_slider = None
+        self._normal_scale_user_set = False
+        self._mesh = None
         self.boundary_toggled = None
         self.edges_toggled = None
         self.normals_toggled = None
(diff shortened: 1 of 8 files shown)


yungyuc (changes requested):
Please test using a more complex mesh, e.g., the unstructured mesh in a rectangle:

image

Currently it looks like:

image

It is good to have an adjustable scaling.  It is not convenient enough if one needs to adjust it every time turning on the normals.  The scale should be set to the best value automatically while remaining adjustable.

yungyuc (changes requested):
Leave inline review comments in the PR to highlight the key points of your change.

solvcon/pilot/panel/_tree_panel.py
```diff
@@ -183,11 +195,46 @@ def make_boundary_info(cls, mh):
...
+        if not nface or not crd.size:
+            return cls._NORMAL_SCALE_DEFAULT
+
+        ndim = int(mh.ndim)
+        extent = crd.max(axis=0)[:ndim] - crd.min(axis=0)[:ndim]
+        diag = float(np.linalg.norm(extent))
+        if diag <= np.finfo('float64').eps:
+            return cls._NORMAL_SCALE_DEFAULT
+
+        measure = float(np.prod(extent))
```
c1ydehhx, line 218: We calculate the diagonal (diag) and spatial (measure). The reason why to have diagonal is the spatial may too small that not able to calculate the candidate length.

solvcon/pilot/panel/_tree_panel.py
```diff
@@ -183,11 +195,46 @@ def make_boundary_info(cls, mh):
...
+        if measure > np.finfo('float64').eps:
+            spacing = measure ** (1.0 / ndim) / nface ** (1.0 / ndim)
+        else:
+            spacing = diag / nface ** (1.0 / ndim)
```
c1ydehhx, line 222: Convert mesh spatial to length by take (d)-dimension root. If the measure is too small, take diagonal for candidate length.

solvcon/pilot/panel/_tree_panel.py
```diff
@@ -183,11 +195,46 @@ def make_boundary_info(cls, mh):
...
+        scale = (cls._NORMAL_SCALE_SPACING_FRACTION * spacing
+                 / (cls._NORMAL_SCALE_BASE_FRACTION * diag))
```
c1ydehhx, line 224: We expected to have 30% of candidate length as the length of normal vector. `_NORMAL_SCALE_SPACING_FRACTION` represent `0.3`.

solvcon/pilot/panel/_tree_panel.py
```diff
@@ -237,12 +284,71 @@ def _add_overlay_toggles(self, root):
...
+    def _sync_normal_scale_text(self):
+        scale = self.normal_scale()
+        self._normal_scale_item.setText(0, f"scale: {scale:.2f}x")
+
+    def _on_normal_scale_changed(self, _value):
+        self._normal_scale_user_set = True
+        self._sync_normal_scale_text()
+        if (self._normal_item.checkState(0) == Qt.Checked
+                and self.normals_toggled is not None):
+            self.normals_toggled(True)
```
c1ydehhx: The code section is used for the GUI design. We added a slider to adjust the scale.

c1ydehhx (commented):
`recommended_normal_scale_value` function is used to calculate the best-fit scale for normal vector.

solvcon/pilot/panel/_tree_panel.py
yungyuc, line 222: Taking `spacing /= nface ** (1.0 / ndim)` outside `if/else` will make the mathematic expression clearer and less error-prone (less chance to update one but forget the other):

```suggestion
        if measure > np.finfo('float64').eps:
            spacing = measure ** (1.0 / ndim)
        else:
            spacing = diag
        spacing /= nface ** (1.0 / ndim)
```

But it's OK to have it inside the `if/else`.

yungyuc (approved):
LGTM.

30% looks like a reasonable value to start with.

PR #1460, opened 2026-09-01
Complete `Float16` SimpleArray operations

ThreeMonth03:
Enable the existing `SimpleArray` calculation, sorting, searching, matrix, and matrix multiplication interfaces for `Float16`.

Reuse the existing algorithms and result types:

```text
Float16 storage -> existing SimpleArray algorithm -> existing result type
```

Add the required `abs`, `sqrt`, and NaN handling. Tests cover calculations, NaN ordering, matrix helpers, and naive matrix multiplication.

Existing generic zero-dimensional and non-contiguous limitations remain unchanged.

Related to #1393.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -293,6 +293,22 @@ template <typename U>
 inline constexpr bool is_bool_v = std::is_same_v<bool, std::remove_const_t<U>>;
 
 template <typename U>
+inline constexpr bool is_array_real_v =
+    is_real_v<std::remove_cv_t<U>> || std::is_same_v<std::remove_cv_t<U>, Float16>;
+
+template <typename U>
+constexpr U min_identity() { return std::numeric_limits<U>::max(); }
+
+template <>
+constexpr Float16 min_identity<Float16>() { return Float16::from_bits(0x7c00U); }
+
+template <typename U>
+constexpr U max_identity() { return std::numeric_limits<U>::lowest(); }
+
+template <>
+constexpr Float16 max_identity<Float16>() { return Float16::from_bits(0xfc00U); }
+
+template <typename U>
 struct select_real_t
 {
     using type = U;
@@ -712,7 +728,7 @@ public:
 
     real_type std_op(small_vector<value_type> & sv, size_t ddof) const
     {
-        return static_cast<real_type>(std::sqrt(var_op(sv, ddof)));
+        return static_cast<real_type>(solvcon::sqrt(var_op(sv, ddof)));
     }
 
     auto std(const shape_type & axis, size_t ddof) const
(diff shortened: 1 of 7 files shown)


cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -192,24 +192,11 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
...
-        else
-        {
-            (*this)
-                .wrap_calculators()
-                .wrap_matrix()
-                .wrap_sort()
-                .wrap_search();
-        }
+            .wrap_modifiers()
+            .wrap_calculators()
```
ThreeMonth03: The typed wrapper no longer branches on `Float16`:

```text
SimpleArray<T>
  |-- modifiers
  |-- calculators
  |-- matrix
  `-- sort/search
```

For `T = Float16`, the same bindings call the shared algorithms; no parallel calculation implementation is added.

cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -292,6 +292,22 @@ A & SimpleArrayMixinModifiers<A, T>::fill(value_type const & value)
...
 
+template <typename U>
+inline constexpr bool is_array_real_v =
+    is_real_v<std::remove_cv_t<U>> || std::is_same_v<std::remove_cv_t<U>, Float16>;
+
+template <typename U>
+constexpr U min_identity() { return std::numeric_limits<U>::max(); }
+
+template <>
+constexpr Float16 min_identity<Float16>() { return Float16::from_bits(0x7c00U); }
```
ThreeMonth03, line 303: `Float16` is a class and has no `std::numeric_limits` specialization, so these array-local adapters isolate the two type-system gaps:

```text
classification: Float32/64 -> is_real_v
                Float16    -> explicit
min/max seed:   built-ins  -> numeric_limits
                Float16    -> +inf / -inf
```

The operation bodies below stay generic.

cpp/solvcon/math/math.hpp
```diff
@@ -74,6 +74,8 @@ inline auto real(T const & val)
     }
 }
 
+inline Float16 abs(Float16 val) { return Float16::from_bits(val.bits() & 0x7fffU); }
```
ThreeMonth03, line 77: `Float16` cannot use the standard math overload set directly. These scalar adapters keep its representation details at one boundary:

```text
abs   : clear sign bit
sqrt  : Float16 -> float -> sqrt -> Float16
isnan : exponent=all ones && fraction!=0
```

The array algorithms keep one math call path for every real dtype.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
ThreeMonth03, line 196: The typed wrapper no longer branches on `Float16`:

```text
SimpleArray<T>
  |-- modifiers
  |-- calculators
  |-- matrix
  `-- sort/search
```

For `T = Float16`, the same bindings call the shared algorithms; no parallel calculation implementation is added.

ThreeMonth03 (commented):
@yungyuc  Please review this pull request. Thanks.

cpp/solvcon/buffer/SimpleArray.hpp
yungyuc: This is an important detail.  In the future, we should study how to properly support meta-programming for the custom real number types in solvcon.

yungyuc (approved):
LGTM

PR #1474, opened 2026-09-03
Parse ros2idl with a tokenizer instead of line regexes

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
This PR replaces the line-based `ros2idl` parser in `solvcon.track.mcap` with a recursive-descent parser over a token stream. The registry now records typedefs, unions, and integer constants next to structs and enums. Every field carries a parsed type instead of its source text.

The old parser matched a field declaration with one regular expression. It read only a declaration that fits on one line, and it kept anything else as raw text. A real recording defeats that shape. The schema arrives as a bundle of IDL blocks, and those blocks carry typedefs, constants, unions, multi-dimensional arrays, and annotations that wrap across lines. Across 20 recordings holding 48 distinct `ros2idl` schemas, one raised `IndexError` and one lost its root struct. All 48 parse now.

A type is a tuple headed by `scalar`, `string`, `sequence`, `array`, `unsupported`, or `named`. That vocabulary is what the decoder compiles against in the next PR of this stack.

The decode model does not change here. The plan still flattens the root struct into scalar leaves and reads them with one `unpack_from`. A string, sequence, array, or union field still raises `McapError`, so this PR adds no new column to any topic.

This is the first of four stacked PRs that take the reader from the prototype parser to one that decodes real recordings. The next PR compiles the schema into a walk-and-skip program, which is the change that makes a message containing a string decodable.

Related to #1438.

### DIFF
--- a/solvcon/track/mcap/_decodeplan.py
+++ b/solvcon/track/mcap/_decodeplan.py
@@ -4,18 +4,16 @@
 """
 Schema layer: from ``ros2idl`` text to a ``DecodePlan``.
 
-IDL is the OMG Interface Definition Language.  ROS 2 embeds IDL text
-in a recording under the ``ros2idl`` schema encoding, and this module
-parses that text.
+IDL is the OMG Interface Definition Language.  ROS 2 embeds IDL text in a
+recording under the ``ros2idl`` schema encoding, and ``parse_schema``
+turns that text into one ``Registry`` of structs, enums, unions,
+typedefs, and integer constants.
 
-``parse_schema`` turns the schema text into a registry of structs and
-enums.  ``DecodePlan`` flattens the root struct into scalar leaves and
-compiles the selected ones into one ``struct`` format per byte order, so a
-message decodes with a single ``unpack_from``.
-
-The plan flattens structs whose fields are scalars, enums, or other
-structs.  A string, sequence, array, or union field raises ``McapError``
-when a plan reaches it.
+``DecodePlan`` flattens the root struct into scalar leaves and compiles
+the selected ones into one ``struct`` format per byte order, so a message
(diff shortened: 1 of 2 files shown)


solvcon/track/mcap/_decodeplan.py
```diff
@@ -62,83 +68,308 @@
...
+    """Drop annotations, comments, includes, and the bundle separators."""
+    text = _COMMENT.sub(lambda match: match.group(1) or "", text)
+    text = _ENUM_VALUE.sub(r"\2 = \1", text)
+    lines = [line for line in _ANNOTATION.sub("", text).splitlines()
+             if not line.lstrip().startswith(("#", "IDL:"))
+             and not _SEPARATOR.match(line.strip())]
+    return "\n".join(lines)
+
+
+class _IdlParser:
```
tigercosmos, line 119: Add IDL parser to support multi-line type defination.

tests/test_track_mcap.py
```diff
@@ -67,11 +67,242 @@
 };
 """
 
+SIGNALS_IDL = b"""
+module demo_msgs { module msg {
```
tigercosmos, line 71: Add new complex test cases.

solvcon/track/mcap/_decodeplan.py
```diff
@@ -62,83 +68,308 @@
...
+        if self.peek() == token:
+            self.pos += 1
+
+    def skip_statement(self):
+        while self.next() not in (";", None):
+            pass
+
+    def definitions(self, modules, end):
+        while True:
+            token = self.next()
```
tigercosmos, line 150: Parse IDL into tokens.

PR #1483, opened 2026-09-03
MCAP: Walk a CDR payload instead of unpacking it flat

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
The decode plan walks the CDR body over a schema tree, so a message with a string, a sequence, or an array decodes into columns.

Related to #1438.

### DIFF
--- a/solvcon/track/mcap/_decodeplan.py
+++ b/solvcon/track/mcap/_decodeplan.py
@@ -9,11 +9,11 @@ recording under the ``ros2idl`` schema encoding, and ``parse_schema``
 turns that text into one ``Registry`` of structs, enums, unions,
 typedefs, and integer constants.
 
-``DecodePlan`` flattens the root struct into scalar leaves and compiles
-the selected ones into one ``struct`` format per byte order, so a message
-decodes with a single ``unpack_from``.  The plan flattens structs whose
-fields are scalars, enums, or other structs.  A string, sequence, array,
-or union field raises ``McapError`` when a plan reaches it.
+``DecodePlan`` turns the root struct into a tree and walks it over the
+CDR body of each message.  A scalar or enum leaf reached through nested
+structs is a column.  The walk steps over a string, a sequence, or an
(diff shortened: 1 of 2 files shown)


solvcon/track/mcap/_decodeplan.py
```diff
@@ -373,56 +381,106 @@ def __init__(self, schema, fields=None):
...
+        fmt = _SCALARS[order][node[1]]
+        pos += -pos % fmt.size
+        if values is not None:
+            values.append(fmt.unpack_from(body, pos)[0])
+        return pos + fmt.size
+    if kind == "struct":
+        for _, child in node[1]:
+            pos = _walk(child, body, pos, order, values)
+        return pos
+    # TODO: a string, a sequence, or an array is only skipped; none of them
```
tigercosmos, line 462: TODO: we are able to parse string, squence and array, but we don't decode them.

solvcon/track/mcap/_decodeplan.py
```diff
@@ -373,56 +381,106 @@ def __init__(self, schema, fields=None):
...
+        if name in visiting:
+            raise McapError("struct {} contains itself".format(name))
+        return ("struct", [(field, _tree(registry, field_type,
+                                         visiting + (name,)))
+                           for field, field_type in registry.structs[name]])
+    if kind == "string":
+        return ("string",)
+    if kind == "sequence":
+        return ("sequence", _tree(registry, type_[1], visiting))
+    if kind == "array":
```
tigercosmos, line 427: Support the types of string, sequence and array.

tests/test_track_mcap.py
```diff
@@ -117,6 +118,16 @@
 
       sequence<Reason, kReasonCapacity> reasons;
 
+      sequence<string<256>, 32> notes;
```
tigercosmos, line 121: New types for testing.

Issue #1443, opened 2026-08-30
Prepare the build system for CUDA on Linux and Windows


yungyuc:
Create the stub for building CUDA code on Linux and Windows.

We will need:
- Install and maintain CUDA development and runtime environment (`nvcc`)
- CI
- `BUILD_CUDA=ON` and `CMAKE_CUDA_ARCHITECTURES`

We do not consider to enable CUDA on mac for now.

tigercosmos:
WIP in https://github.com/solvcon/solvcon/pull/1417, which includes the CUDA setup

yungyuc:
> WIP in #1417, which includes the CUDA setup

PR #1417 is to add CUDA FFT, so it depends on the CUDA build system to be developed with this issue.  The build system needs to be done before application code.

We are laying out the GPU code framework using the Metal work (discussion #1440).  CUDA code should be done after that.

tigercosmos:
> We are laying out the GPU code framework using the Metal work (discussion https://github.com/solvcon/solvcon/discussions/1440). CUDA code should be done after that.

I wonder if we should consider both Metal and Cuda at the same time for the GPU framework?

yungyuc:
> > We are laying out the GPU code framework using the Metal work (discussion [#1440](https://github.com/solvcon/solvcon/discussions/1440)). CUDA code should be done after that.
> 
> I wonder if we should consider both Metal and Cuda at the same time for the GPU framework?

No.  We should finish one and then work on the second.  Architecture design is not something that can be parallelized.

PR #1488, opened 2026-09-04
Prepare Linux CUDA builds

tigercosmos:
This change adds an opt-in CUDA compile probe and a dedicated Ubuntu 24.04 CI lane with pinned CUDA 13.0 packages. The compile-only probe stays separate from the C++ core because nvcc uses a different supported host compiler.

Local verification passed with CUDA 13.0, GCC 16 for the C++ core, and GCC 14 for nvcc: `make lint`; the CUDA-enabled Release build; 1,867 Python tests with 606 skips and 3 expected failures; all 264 C++ tests; and the CUDA plus sanitizer compile probe.

Related to #1443.

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -283,8 +283,66 @@ jobs:
         make run_pilot_pytest VERBOSE=1
         echo "::endgroup::"
 
+  build_cuda:
+
+    needs: [check_skip, standalone_buffer]
+
+    if: |
+      needs.check_skip.outputs.skip != 'true' && (
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'pull_request' ||
+      (github.event_name == 'push' &&
+       ((vars.SCGH_PUSH_RUN_BRANCH || vars.MMGH_PUSH_RUN_BRANCH) == '*' ||
+        contains(vars.SCGH_PUSH_RUN_BRANCH || vars.MMGH_PUSH_RUN_BRANCH, github.ref_name))))
+
+    name: build_cuda_ubuntu-24.04
+
+    runs-on: ubuntu-24.04
+
+    timeout-minutes: ${{ fromJSON(vars.SCGH_TIMEOUT_BUILD || vars.MMGH_TIMEOUT_BUILD || '45') }}
+
+    env:
+      MAKE_PARALLEL: -j4
+      CMAKE_BUILD_TYPE: Release
+      PIP_BREAK_SYSTEM_PACKAGES: 1
+
+    steps:
+
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 1
+
+    - name: event name
+      run: |
+        echo "github.event_name: ${{ github.event_name }}"
+
+    - name: setup dependencies for Linux
+      uses: ./.github/actions/setup_linux
+      with:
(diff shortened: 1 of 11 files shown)


.github/workflows/devbuild.yml
```diff
@@ -283,8 +283,77 @@ jobs:
         make run_pilot_pytest VERBOSE=1
         echo "::endgroup::"
 
+  build_cuda:
```
tigercosmos, line 286: Buiild CUDA in devbuild.

cpp/solvcon/device/cuda/cuda.cu
```diff
@@ -0,0 +1,41 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace solvcon
+{
+
+namespace device
+{
+
+__global__ void cuda_build_probe() {}
```
tigercosmos, line 35: Simple build probe.

CMakeLists.txt
```diff
@@ -120,6 +120,24 @@ if(BUILD_METAL)
     add_compile_options($<$<COMPILE_LANGUAGE:OBJCXX>:-fobjc-arc>)
 endif()
 
+option(BUILD_CUDA "build with CUDA" OFF)
+message(STATUS "BUILD_CUDA: ${BUILD_CUDA}")
+if(BUILD_CUDA)
+    if(APPLE)
+        message(FATAL_ERROR "BUILD_CUDA is not supported on macOS")
```
tigercosmos, line 127: Reject Apple. Handle Windows in the next PR.

CMakeLists.txt
```diff
@@ -120,6 +120,24 @@ if(BUILD_METAL)
...
+    endif()
+    if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
+        set(CMAKE_CUDA_ARCHITECTURES native)
+    endif()
+    # The probe does not include C++23 core headers. C++17 keeps the build
+    # compatible with CUDA 11 and newer.
+    set(CMAKE_CUDA_STANDARD 17)
```
tigercosmos, line 134: C++17 for CUDA 11.

tigercosmos (commented):

.github/workflows/devbuild.yml
```diff
@@ -283,8 +283,77 @@ jobs:
...
+    - name: event name
+      run: |
+        echo "github.event_name: ${{ github.event_name }}"
+
+    - name: setup dependencies for Linux
+      uses: ./.github/actions/setup_linux
+      with:
+        workflow: 'build_cuda'
+
+    - name: install CUDA toolkit
```
ExplorerRay: Why choose installing cuda toolkit in `devbuild` instead of in `setup_linux` action? 

I think we can install it in that prepare prerequisites action and use workflow `build_cuda` to determine that only this step will install it.

tigercosmos: True, moved.

yungyuc (changes requested):
- [x] scdv scripts need corresponding code for the hints to install cuda compiler and runtime.

contrib/dependency/build-scdv.sh
tigercosmos, line 1: Update `contrib/dependency/build-scdv.sh` for CUDA.

yungyuc (approved):
LGTM

PR #1489, opened 2026-09-04
MCAP: Walk a union case by its discriminator

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
This PR lets the decode plan in `solvcon.track.mcap` walk an IDL union. The plan turns a union into a tree node that holds the discriminator and a map from case value to case subtree. The walk reads the discriminator and steps into the case it selects. When no case matches and no default exists, the walk steps past the discriminator alone.

A schema with a union could not build a plan before this PR, so no field of that message, union or not, reached a column. A leaf inside a union is not a column, the same rule as a leaf inside a sequence or an array. A union that contains itself, or that switches on a non-integral type, raises `McapError`. The parser follows a typedef to the enum it names, so a case label of an enum typedef resolves to its ordinal.

Related to #1438.

### DIFF
--- a/solvcon/track/mcap/_decodeplan.py
+++ b/solvcon/track/mcap/_decodeplan.py
@@ -11,9 +11,9 @@ typedefs, and integer constants.
 
 ``DecodePlan`` turns the root struct into a tree and walks it over the
 CDR body of each message.  A scalar or enum leaf reached through nested
-structs is a column.  The walk steps over a string, a sequence, or an
-array, so a leaf inside one is not a column.  A union field raises
-``McapError`` when a plan reaches it.
+structs is a column.  The walk steps over a string, a sequence, an
+array, or a union, so a leaf inside one is not a column.  A union reads
+its discriminator to find the case the payload carries.
 """
 
 import re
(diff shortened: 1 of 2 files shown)


tests/test_track_mcap.py
```diff
@@ -242,7 +242,66 @@ def test_signals(self):
...
+                self.assertEqual(plan.decode(pack_signals(order, kind)),
+                                 (7, kind, True, 3.5), (order, kind))
+        with self.assertRaisesRegex(mcap.McapError, "ends before"):
+            plan.decode(pack_signals("<", 2)[:-12])
+
+    def test_union_cases(self):
+        """A union walks the case its discriminator selects, or none."""
+        schema = mcap.Schema(1, "x/msg/Y", "ros2idl", b"""
+            module x { module msg {
+              union Flag switch(boolean) {
```
tigercosmos, line 265: Test `union`.

solvcon/track/mcap/_decodeplan.py
```diff
@@ -404,31 +416,47 @@ def _tree(registry, type_, visiting):
...
     if kind == "sequence":
         return ("sequence", _tree(registry, type_[1], visiting))
     if kind == "array":
         return ("array", _tree(registry, type_[1], visiting), type_[2])
+    if kind == "union":
+        return _union_tree(registry, type_[1], visiting)
     raise McapError("unsupported type {!r}".format(type_[1]))
 
 
+def _union_tree(registry, name, visiting):
```
tigercosmos, line 448: Support union type in IDL.

PR #1493, opened 2026-09-05
MCAP: Decode a named string, sequence, or array as a column

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
This PR lets a `DecodePlan` in `solvcon.track.mcap` decode a string, a sequence, or an array into a column. A field list may name a string field, whose type is `str`. A sequence or array of scalars, enums, or strings has the element type followed by `[]`. Each decodes to a Python `str` or `list` per message. The reader stacks them into a NumPy `object` array, and `Extraction.to_frame` rejects such a column with `McapError`.

Before this PR the walk only stepped over these fields, so no such field could reach a column. The auto plan still selects scalar leaves only, so a recording full of `frame_id` strings extracts a frame without a field list. The walk decodes a string or container only when the plan selects it. A container of structs, unions, or nested containers is still not a column. A string without its NUL terminator, or with bytes that are not UTF-8, raises `McapError`.

Related to #1438.

### DIFF
--- a/solvcon/track/mcap/_decodeplan.py
+++ b/solvcon/track/mcap/_decodeplan.py
@@ -11,14 +11,19 @@ typedefs, and integer constants.
 
 ``DecodePlan`` turns the root struct into a tree and walks it over the
 CDR body of each message.  A scalar or enum leaf reached through nested
-structs is a column.  The walk steps over a string, a sequence, an
-array, or a union, so a leaf inside one is not a column.  A union reads
-its discriminator to find the case the payload carries.
+structs is a column.  A string is a column when the field list names
+it, and so is a sequence or array of scalars, enums, or strings.  Such
+a field decodes to a ``str`` or a ``list`` per message.  The walk steps
+over a union and over a container of any other element, so a leaf
+inside one is not a column.  A union reads its discriminator to find
+the case the payload carries.
 """
 
 import re
 import struct
 import collections
+
+import numpy as np
 
 import solvcon as sc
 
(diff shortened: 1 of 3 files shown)


solvcon/track/mcap/_decodeplan.py
```diff
@@ -502,16 +541,40 @@ def _walk(node, body, pos, order, values):
         pos += 4
     if count > len(body) - pos:
         raise McapError(_TRUNCATED)
+    selected = values is not None and node[-1]
     if kind == "string":
+        if selected:
+            values.append(_text(body, pos, count))
```
tigercosmos, line 547: Support string type parsing.

solvcon/track/mcap/_decodeplan.py
```diff
@@ -502,16 +541,40 @@ def _walk(node, body, pos, order, values):
...
         return pos + count
-    if node[1][0] == "scalar":
-        size = _SCALARS[order][node[1][1]].size
-        return pos + (-pos % size if count else 0) + size * count
+    child = node[1]
+    if child[0] == "scalar":
```
tigercosmos, line 550: Support sequence type parsing.

tests/test_track_mcap.py
```diff
@@ -320,35 +322,73 @@ def test_bundle(self):
...
         with self.assertRaisesRegex(mcap.McapError, "ends before"):
-            plan.decode(packer.payload()[:-3])
+            plan.decode(payload[:-3])
         with self.assertRaisesRegex(mcap.McapError, "ends before"):
             plan.decode(CdrPacker("<").scalar("I", 999).payload())
         with self.assertRaisesRegex(mcap.McapError, "encapsulation"):
-            plan.decode(b"\0\x02\0\0" + packer.payload()[4:])
+            plan.decode(b"\0\x02\0\0" + payload[4:])
+
+    def test_named_containers_decode(self):
```
tigercosmos, line 346: Test case for container types.

PR #1494, opened 2026-09-05
Add a fake MCAP recording as a test fixture

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
The change adds a small MCAP recording under tests/data/mcap, together with the script that writes it.

solvcon reads MCAP through solvcon/track/mcap, but the repository has no recording to read. The tests build a one-topic file at run time, which does not show the shape a real recording has. The new fixture does: four `cdr` channels over three `ros2idl` schemas, two channels sharing one schema, IDL blocks bundled under separator lines, a nested header struct, an enum field, a sequence field, and a string field. Every topic, package, field name, and value is invented, so the file carries no content from any real recording.

The generator stays in the tree next to its output, so a reader can see how the fixture is built and can write it again after a schema change. The chunks stay uncompressed, because the reader imports `zstandard` and `lz4` only on demand and neither is a solvcon dependency. The file is 9827 bytes.

Related to #1438


tests/data/mcap/make_fake_recording.py
tigercosmos, line 1: Add a script to generate MCAP files for the viewer testing.

tests/data/mcap/make_fake_recording.py
```diff
@@ -0,0 +1,295 @@
...
+EGO_STATE_IDL = bundle(EGO_STATE_BLOCK, HEADER_BLOCK, POINT2_BLOCK)
+DRIVE_MODE_IDL = bundle(DRIVE_MODE_BLOCK, HEADER_BLOCK)
+FAULT_REPORT_IDL = bundle(FAULT_REPORT_BLOCK, HEADER_BLOCK)
+
+EGO_TOPIC = "/sim/ego/state"
+FILTERED_TOPIC = "/sim/ego/state_filtered"
+MODE_TOPIC = "/sim/ego/drive_mode"
+FAULT_TOPIC = "/sim/diag/fault_report"
+
+COUNT = 6
```
tigercosmos, line 153: Small `COUNT` to ensure the output file is smaller enough (less than 10KB). Change to a big number while testing.

PR #1496, opened 2026-09-05
Read a recording that has no summary section

### LINKED ISSUE #1438: Create a pure Python prototype of MCAP reader

tigercosmos:
`Reader` opens recordings it refused before, and decodes channels it could not reach.

What `Reader` can do now:

- Read a recording that has no summary section. A recorder that dies leaves the file without one, and without a footer or a trailing magic.
- Read a recording that stores its messages outside chunks.
- Decode a channel that carries no schema, with a `DecodePlan` that the caller builds from IDL of its own.
- List a topic whose channel carries no schema. `topics()` gives an empty schema name and `schema()` gives `None`, where both raised `KeyError` before.
- Drop the record that the end of the file cuts short, and raise `McapError` for a header the file cuts short.

Without a summary section there is no index, so `Reader` walks the data section at open time. The message index records name every channel that carries a message, so the walk reads chunk bodies only until it accounts for them all. A 1.3 GB recording opens after 8 MB.

`extract_many` accepts a plan whose schema differs from the one the file carries. The guard on that plan compares the message type the plan names against the type the file gives the topic.

Related to #1438

### DIFF
--- a/solvcon/track/mcap/_reader.py
+++ b/solvcon/track/mcap/_reader.py
@@ -7,7 +7,9 @@ Container layer: the MCAP file format.
 ``Reader`` reads the magic, the header, the footer, and the summary section
 the footer points to.  The summary gives the schemas, the channels, the
 statistics, and the chunk indexes; the chunks themselves are read on demand
-by ``messages()`` and the ``extract*`` methods.
+by ``messages()`` and the ``extract*`` methods.  A file without a summary
+section gives no statistics, and ``Reader`` builds the index by walking the
+data section at open time instead.
 """
 
 import os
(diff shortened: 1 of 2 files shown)


solvcon/track/mcap/_reader.py
```diff
@@ -132,53 +137,163 @@ def _read_at(self, offset, size):
         self._file.seek(offset)
         return self._file.read(size)
 
-    def _read_summary(self):
-        self._file.seek(0, 2)
-        file_size = self._file.tell()
-        if self._read_at(0, 8) != MAGIC or \
-                self._read_at(file_size - 8, 8) != MAGIC:
+    def _read_index(self):
```
tigercosmos, line 140: Be able to handle the MCAP file without the optional summary section.

solvcon/track/mcap/_reader.py
```diff
@@ -132,53 +137,163 @@ def _read_at(self, offset, size):
...
+            self._schemas[schema_id] = Schema(schema_id, name, encoding,
+                                              bytes(data))
+        elif opcode == OP_CHANNEL:
+            (channel_id, schema_id), pos = unpack("<HH", body, 0)
+            topic, pos = unpack_string(body, pos)
+            message_encoding, pos = unpack_string(body, pos)
+            self._channels[channel_id] = Channel(
+                channel_id, schema_id, topic, message_encoding)
+
+    def _scan_data(self, offset):
```
tigercosmos, line 206: If there is no summary, we can through the data to do indexing.

tests/test_track_mcap.py
```diff
@@ -485,6 +488,14 @@ def test_string_const_does_not_shadow(self):
                          ("array", ("scalar", "float64"), 3))
 
 
+def cut_before_summary(data):
```
tigercosmos, line 491: Cut the summary for testing the MCAP file without a summary.

PR #1499, opened 2026-09-05
Fix inplace arithmetic bindings to return self by reference

### LINKED ISSUE #1475: Investigate python object creation when using non-copy return value policy in pybind11
During implementation of PR #1362, it is found that using [non-copy return value policy](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies) cannot keep the python object ID the same after returning reference to self. This issue should investigate and figure out the necessity of using `std::move` when returning self reference or the proper alternative.

KHLee529:
The way fluent API implemented in PR #1362 use `py::cast(std::move(object))` to make the wrapper python object have same python object ID which looks unintuitive. This PR rewrite the implementation by explicitly annotate the return type of wrapped function as reference type and remove the `SimpleArrayCaster::cast` implementation to use `cast` function from base class.

Related to #1475.

### DIFF
--- a/cpp/solvcon/buffer/pymod/SimpleArrayCaster.hpp
+++ b/cpp/solvcon/buffer/pymod/SimpleArrayCaster.hpp
@@ -56,12 +56,6 @@ namespace detail
             value = const_cast<solvcon::SimpleArray##DATATYPE *>(array_from_arrayplex); /* NOLINT(cppcoreguidelines-pro-type-const-cast) */                   \
             return true;                                                                                                                                      \
         }                                                                                                                                                     \
-                                                                                                                                                              \
-        /* Conversion from C++ to Python object FIXME: NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved) */                                  \
-        static pybind11::handle cast(solvcon::SimpleArray##DATATYPE && src, pybind11::return_value_policy policy, pybind11::handle parent)                    \
-        {                                                                                                                                                     \
-            return base::cast(src, policy, parent);                                                                                                           \
-        }                                                                                                                                                     \
     } /* end struct type_caster */
 
 SC_DECL_SIMPLE_ARRAY_CASTER(Bool);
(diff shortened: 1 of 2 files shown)


cpp/solvcon/buffer/pymod/SimpleArrayCaster.hpp
```diff
@@ -55,12 +55,6 @@ namespace detail
             const solvcon::SimpleArray##DATATYPE * array_from_arrayplex = reinterpret_cast<const solvcon::SimpleArray##DATATYPE *>(arrayplex.instance_ptr()); \
             value = const_cast<solvcon::SimpleArray##DATATYPE *>(array_from_arrayplex); /* NOLINT(cppcoreguidelines-pro-type-const-cast) */                   \
             return true;                                                                                                                                      \
-        }                                                                                                                                                     \
-                                                                                                                                                              \
-        /* Conversion from C++ to Python object FIXME: NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved) */                                  \
-        static pybind11::handle cast(solvcon::SimpleArray##DATATYPE && src, pybind11::return_value_policy policy, pybind11::handle parent)                    \
```
KHLee529, line 61: Remove unnecessary caster function to use the inherited function from base class.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -466,19 +464,19 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
              */
             .def(
                 "iadd",
-                [](wrapped_type & self, wrapped_type const & other)
-                { return py::cast(std::move(self.iadd(other))); })
+                [](wrapped_type & self, wrapped_type const & other) -> wrapped_type &
```
KHLee529, line 467: Remove `std::move` and annotate return type.

ThreeMonth03 (approved):
LGTM

tigercosmos (approved):
This sounds more like pybind11 limitation. Maybe pybind26 will be smarter.

PR #1503, opened 2026-09-05
Release old heap storage in `small_vector` move assignment

### LINKED ISSUE #1502: Fix the heap leak in `small_vector` move assignment
`small_vector` move assignment omits `delete[]` before overwriting the destination's heap pointer. When both containers use heap storage, the old destination allocation leaks.

```cpp
solvcon::small_vector<size_t> destination(4);
solvcon::small_vector<size_t> source(5);

destination = std::move(source);
// The old destination allocation is lost without delete[].
```

Both sizes exceed the default inline capacity of three. The bug also affects containers of integers; it does not depend on `shared_ptr` support.

- [x] Delete the destination's old heap storage before taking the source allocation, and add a regression for the leak. (#1503)

```text
heap-to-heap move assignment
  -> delete destination's old storage
  -> take source pointer, size, and capacity
  -> reset source to empty inline storage
```

ThreeMonth03:
`small_vector` move assignment forgot to delete the destination's old heap storage before taking the source pointer. This PR adds the missing `delete[]` when the destination owns heap storage.

```text
heap-to-heap move assignment
  -> delete destination's old storage
  -> take source pointer, size, and capacity
  -> reset source to empty inline storage
```

The regression uses `weak_ptr::expired()` to detect whether the old elements were released. It also checks that the destination takes the source allocation.

The C++ suite passes 265 tests with Qt and Metal disabled. `make lint` passes.

Related to #1502.

### DIFF
--- a/cpp/solvcon/buffer/small_vector.hpp
+++ b/cpp/solvcon/buffer/small_vector.hpp
@@ -175,6 +175,10 @@ public:
             }
             else
             {
+                if (m_head != m_data.data())
+                {
+                    delete[] m_head;
+                }
                 m_size = other.m_size;
                 m_capacity = other.m_capacity;
                 m_head = other.m_head;
--- a/gtests/test_nopython_buffer.cpp
+++ b/gtests/test_nopython_buffer.cpp
@@ -8,6 +8,7 @@
 #include <cmath>
 #include <cstdint>
 #include <limits>
+#include <memory>
 #include <random>
 #include <type_traits>
 #ifdef Py_PYTHON_H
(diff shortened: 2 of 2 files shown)


gtests/test_nopython_buffer.cpp
```diff
@@ -661,6 +662,23 @@ TEST(SimpleCollector, pop_back)
...
+TEST(small_vector, heap_move_assignment_releases_destination)
+{
+    solvcon::small_vector<std::shared_ptr<int>> destination(4, std::make_shared<int>(7));
+    std::weak_ptr<int> previous = destination[0];
+    solvcon::small_vector<std::shared_ptr<int>> source(5, std::make_shared<int>(42));
+    auto * const source_data = source.data();
+
+    destination = std::move(source);
+
+    EXPECT_TRUE(previous.expired());
```
ThreeMonth03, line 674: Both sizes exceed the inline capacity (`N = 3`), so this exercises heap-to-heap assignment.

```text
old destination storage -> delete[] -> last shared_ptr released -> previous.expired()
source storage          -> same pointer now owned by destination
```

The weak pointer makes the leaked element ownership observable without a leak sanitizer.

ThreeMonth03 (commented):
@tigercosmos Please review this pull request. Thank you.

tigercosmos (approved):
Code itself is good. Just not sure if people will ever use small vector for pointers as I mentioned in the issue. I think we can merge this anyway.

cpp/solvcon/buffer/small_vector.hpp
```diff
@@ -175,6 +175,10 @@ class small_vector
             }
             else
             {
+                if (m_head != m_data.data())
```
yungyuc, line 178: Good catch.

PR #1507, opened 2026-09-06
Fix transpose ownership and axis validation

### LINKED ISSUE #1505: Fix SimpleArray transpose ownership and axis validation
`SimpleArray.transpose()` accepts repeated axes and leaks memory when `inplace=False`.

### Repeated axes are accepted

```python
import numpy as np
import solvcon as sc

array = sc.SimpleArrayFloat64(
    array=np.arange(6, dtype="float64").reshape(2, 3)
)
result = array.transpose(axis=(1, 1), inplace=False)

print(result.shape)
# Current:  (3, 3)
# Expected: raise an exception because axis 1 appears twice.
```

The axis validation checks whether each output position has been assigned, so it does not detect repeated source axes. The supplied axes must form a permutation of the input axes.

### Out-of-place transpose leaks memory

```python
import gc
from pathlib import Path

import numpy as np
import solvcon as sc


def rss_kib():
    for line in Path("/proc/self/status").read_text().splitlines():
        if line.startswith("VmRSS:"):
            return int(line.split()[1])


array = sc.SimpleArrayFloat64(array=np.ones((1000, 1000)))

for _ in range(2):
    result = array.transpose(inplace=False)
    del result

gc.collect()
before = rss_kib()

for _ in range(6):
    result = array.transpose(inplace=False)
    del result

gc.collect()
print(rss_kib() - before)
# Observed on Linux: 46848 KiB retained after six calls.
```

The Python wrapper allocates a copy with `new`, returns its value, and never deletes the allocated object. Discarding the returned Python object therefore leaves the intermediate allocation alive. The RSS measurement above illustrates the leak; its exact value depends on the allocator.

### Acceptance

- Reject repeated axes before changing the receiver, for both `inplace` modes and both `copy` modes.
- Release intermediate allocations on successful calls and when axis validation raises an exception.
- Preserve the existing behavior of valid transposes.
- Add regression coverage for repeated axes and verify that repeated out-of-place calls no longer retain an allocation per call.

ThreeMonth03:
## Motivation

A repeated axis returns an invalid shape:

```python
import numpy as np
import solvcon as sc

a = sc.SimpleArrayFloat64(
    array=np.arange(6, dtype="float64").reshape(2, 3))
a.transpose(axis=(1, 1), inplace=False)
# Before: returns an array with shape (3, 3)
# After:  RuntimeError: SimpleArray::transpose: axis already set
```

The old out-of-place path leaks the allocated array:

```cpp
auto *ret = new wrapped_type(self);
ret->transpose(copy);
return *ret; // Copies the result; never deletes ret.
```

## Approach

For `axis=(1, 1)`, track source axis `1` instead of separate output positions:

```text
        old check                   new check
it=0    new_shape[0] == -1: accept   seen[1] == false: mark true
it=1    new_shape[1] == -1: accept   seen[1] == true: throw
```

Own the temporary copy with `unique_ptr`. It releases the copy on return and on exceptions:

```cpp
auto owner = inplace ? nullptr : std::make_unique<wrapped_type>(self);
wrapped_type &ret = inplace ? self : *owner;
ret.transpose(copy); // Default-axis path.
return ret;         // Copies the result; owner then frees the temporary.
```

No-Qt tests: 552 passed, 2 skipped, 2 xfailed. `make lint` passes. The allocator probe drops from ~48 MiB retained to 0-64 bytes after six calls.

Related to #1505.

### DIFF
--- a/cpp/solvcon/buffer/SimpleArray.hpp
+++ b/cpp/solvcon/buffer/SimpleArray.hpp
@@ -3020,7 +3020,8 @@ void SimpleArray<T>::transpose(shape_type const & axis, bool copy)
     {
         throw std::runtime_error("SimpleArray::transpose: axis size mismatch");
     }
-    shape_type new_shape(m_shape.size(), -1);
+    small_vector<bool> seen(m_shape.size(), false);
+    shape_type new_shape(m_shape.size());
     shape_type new_stride(m_stride.size());
     for (ssize_t it = 0; it < ndim(); ++it)
     {
@@ -3028,10 +3029,11 @@ void SimpleArray<T>::transpose(shape_type const & axis, bool copy)
         {
             throw std::runtime_error("SimpleArray::transpose: axis out of range");
         }
-        if (new_shape[it] != -1)
+        if (seen[axis[it]])
         {
             throw std::runtime_error("SimpleArray::transpose: axis already set");
         }
+        seen[axis[it]] = true;
         new_shape[it] = m_shape[axis[it]];
         new_stride[it] = m_stride[axis[it]];
     }
--- a/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
+++ b/cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
@@ -11,6 +11,7 @@
 
 #include <cctype>
 #include <cstdint>
+#include <memory>
 #include <string>
 
 namespace solvcon
@@ -145,16 +146,17 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "transpose",
                 [](wrapped_type & self, py::object const & axis, bool inplace, bool copy)
                 {
-                    wrapped_type * ret = inplace ? &self : new wrapped_type(self);
+                    auto owner = inplace ? nullptr : std::make_unique<wrapped_type>(self);
+                    wrapped_type & ret = inplace ? self : *owner;
                     if (axis.is_none())
                     {
-                        ret->transpose(copy);
+                        ret.transpose(copy);
                     }
                     else
                     {
-                        ret->transpose(make_shape(axis), copy);
+                        ret.transpose(make_shape(axis), copy);
                     }
-                    return *ret;
+                    return ret;
                 },
                 py::arg("axis") = py::none(),
                 py::arg("inplace") = true,
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -2,6 +2,7 @@
 # BSD 3-Clause License, see COPYING
 
 
+import itertools
 import operator
 import unittest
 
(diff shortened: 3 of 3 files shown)


cpp/solvcon/buffer/SimpleArray.hpp
```diff
@@ -3020,18 +3020,20 @@ void SimpleArray<T>::transpose(shape_type const & axis, bool copy)
...
+    shape_type new_shape(m_shape.size());
     shape_type new_stride(m_stride.size());
     for (ssize_t it = 0; it < ndim(); ++it)
     {
         if (axis[it] < 0 || axis[it] >= ndim())
         {
             throw std::runtime_error("SimpleArray::transpose: axis out of range");
         }
-        if (new_shape[it] != -1)
+        if (seen[axis[it]])
```
ThreeMonth03, line 3032: A concrete failure is transposing a `(2, 3)` array with `axis=(1, 1)`:

```python
a.transpose(axis=(1, 1), inplace=False)
# Before: shape (3, 3)
# After:  RuntimeError("SimpleArray::transpose: axis already set")
```

Both old output slots start unused; the new check sees source axis `1` twice:

```text
it=0: new_shape[0] == -1     seen[1] = false; set true
it=1: new_shape[1] == -1     seen[1] = true; throw
```

The exception occurs before changing the receiver's shape or stride.

cpp/solvcon/buffer/pymod/wrap_SimpleArray.hpp
```diff
@@ -145,16 +146,17 @@ class SOLVCON_PYTHON_WRAPPER_VISIBILITY WrapSimpleArray
                 "transpose",
                 [](wrapped_type & self, py::object const & axis, bool inplace, bool copy)
                 {
-                    wrapped_type * ret = inplace ? &self : new wrapped_type(self);
+                    auto owner = inplace ? nullptr : std::make_unique<wrapped_type>(self);
```
ThreeMonth03, line 149: For `inplace=False` with default axes:

```cpp
// Before:
auto *ret = new wrapped_type(self);
ret->transpose(copy);
return *ret; // Returns a copy; the allocated ret is never deleted.

// After:
auto owner = inplace ? nullptr : std::make_unique<wrapped_type>(self);
wrapped_type &ret = inplace ? self : *owner;
ret.transpose(copy);
return ret; // Returns a copy; owner releases the temporary.
```

If `transpose()` throws, `owner` also releases the temporary during stack unwinding.

tests/test_buffer.py
```diff
@@ -513,6 +513,46 @@ def check_equal(sarr, ndarr):
...
+                                           copy=copy)
+                        self.assertEqual(sarr.shape, ndarr.shape)
+                        self.assertEqual(sarr.stride, stride)
+                        np.testing.assert_array_equal(sarr.ndarray, ndarr)
+
+    def test_SimpleArray_transpose_modes(self):
+        ndarr = np.arange(24, dtype='float64').reshape(2, 3, 4)
+        for inplace in (False, True):
+            for copy in (False, True):
+                for axes in (None, (2, 0, 1)):
```
ThreeMonth03: Use a permutation that differs from its inverse:

```text
axes:    (2, 0, 1)
inverse: (1, 2, 0)
```

NumPy checks the values in all four `inplace`/`copy` modes. Mutating an out-of-place result must leave the receiver unchanged.

tests/test_buffer.py
```diff
@@ -513,6 +514,44 @@ def check_equal(sarr, ndarr):
...
+                with self.assertRaisesRegex(RuntimeError, message):
+                    sarr.transpose(axis=axes, inplace=inplace, copy=copy)
+
+                self.assertEqual(sarr.shape, ndarr.shape)
+                self.assertEqual(sarr.stride, original_stride)
+                np.testing.assert_array_equal(sarr.ndarray, ndarr)
+
+    def test_SimpleArray_transpose_modes(self):
+        ndarr = np.arange(24, dtype='float64').reshape(2, 3, 4)
+        for inplace, copy, axes in itertools.product(
```
ThreeMonth03, line 540: This permutation checks a concrete coordinate mapping and independent result storage:

```python
source = np.arange(24, dtype="float64").reshape(2, 3, 4)
a = solvcon.SimpleArrayFloat64(array=source)
b = a.transpose(axis=(2, 0, 1), inplace=False)

assert b.shape == (4, 2, 3)
assert b[2, 1, 0] == a[1, 0, 2] == 14

b[2, 1, 0] = -1
assert a[1, 0, 2] == 14
```

The test compares every value with NumPy across all four `inplace`/`copy` modes.

yungyuc (commented):
LGTM

PR #1511, opened 2026-09-06
Validate shapes in array property setters

### LINKED ISSUE #1510: Reject mismatched shapes in array property setters
Array property setters can accept a mismatched shape or misread a NumPy view with the correct shape.

The byte-count check accepts different ranks and axis lengths:

```python
import numpy as np
import solvcon as sc

mesh = sc.StaticMesh(ndim=2, nnode=4)
mesh.ndcrd = np.arange(8, dtype='float64').reshape(2, 4)
# Current: accepts (2, 4), although the mesh requires (4, 2).
# Expected: reject the assignment and preserve the original array.
```

Even with matching shapes, the conversion discards NumPy strides and interprets the source as C-order:

```python
mesh = sc.StaticMesh(ndim=2, nnode=4)
source = np.arange(8, dtype='float64').reshape(2, 4).T
mesh.ndcrd = source

print(source.tolist())
# [[0, 4], [1, 5], [2, 6], [3, 7]]
print(mesh.ndcrd.ndarray.tolist())
# Current: [[0, 1], [2, 3], [4, 5], [6, 7]]
# Expected: the same logical values as source.
```

Require matching rank and axis lengths, including for empty arrays. Use the existing NumPy view conversion to retain strides, the logical origin, and shared storage ownership. Reject layouts that the converter cannot safely represent before replacing the property. Failed assignments must preserve the original shape, values, and buffer.

Regression coverage should include C-order, F-order, stepped and reversed views, buffer sharing and lifetime, and mesh interior, boundary, and ghost construction using the assigned coordinates.

ThreeMonth03:
Property assignment checks byte counts but can still accept a wrong shape or misread a matching-shape NumPy view:

```text
internal shape   source shape   source layout   before                 after
(4, 2)           (2, 4)         C-order         accepts wrong shape    reject
(0, 2)           (0, 3)         empty           accepts wrong shape    reject
(4, 2)           (4, 2)         F-order         reads as C-order       retain layout
```

The matching-shape failure changes coordinate values:

```python
import numpy as np
import solvcon as sc

mesh = sc.StaticMesh(ndim=2, nnode=4)
source = np.arange(8, dtype='float64').reshape(2, 4).T
mesh.ndcrd = source

# source: [[0, 4], [1, 5], [2, 6], [3, 7]]
# before: [[0, 1], [2, 3], [4, 5], [6, 7]]
# after:  [[0, 4], [1, 5], [2, 6], [3, 7]]
```

Check rank and axis lengths in both property setter helpers. Move `make_array_from_numpy()` into the common wrapper header and use it for the replacement, sharing the same stride, storage-span, origin, and ownership handling as `SimpleArray(array=...)`. Finish conversion before swapping, so shape or layout errors leave the property unchanged.

```text
byte count -> shape -> NumPy view conversion -> swap
                 error at any check -> keep the original array
```

Retain the input NumPy array itself to keep its storage alive. Do not traverse the overridable `.base` property, which can form multi-level cycles in an ndarray subclass:

```text
A.base -> B
B.base -> A

before: A -> B -> A -> ...
after:  retain A without traversing .base
```

The tests check rejected-assignment preservation, valid empty arrays, and shared writes for C/F, stepped and reversed layouts. Cycles of lengths 1, 2 and 3 check shared storage and owner release. One reversed view checks owner lifetime; one 2D triangle checks mesh interior, boundary and ghost construction against contiguous coordinates.

No-Qt buffer, mesh, grid, and multidimensional solver tests: 471 passed, 8 skipped. `make lint` passes.

Related to #1510.


cpp/solvcon/python/common.hpp
```diff
@@ -215,6 +216,15 @@ std::enable_if_t<is_simple_array_v<S>, pybind11::array> to_ndarray(S && sarr)
     );
 }
 
+template <typename T>
+static void validate_array_shape(SimpleArray<T> const & array, pybind11::array const & ndarr)
+{
+    if (array.ndim() != ndarr.ndim() || !std::equal(array.shape().begin(), array.shape().end(), ndarr.shape()))
```
ThreeMonth03: Equal byte counts do not imply equal shapes:

```text
(4, 2) vs (8,)       -> same 64 bytes, different rank
(4, 2) vs (4, 2, 1) -> same leading axes, different rank
(4, 2) vs (2, 4)    -> same rank and bytes, different axis lengths
```

Check rank before `std::equal` so the comparison reads equally sized shape vectors. Both property setter helpers call this check before replacing the array.

tests/test_mesh.py
```diff
@@ -60,6 +61,43 @@ def _test(cls, ndim):
...
+                source = np.full(shape, -1, dtype='float64')
+
+                with self.assertRaisesRegex(
+                        ValueError,
+                        'input array shape differs from internal array shape'):
+                    mesh.ndcrd = source
+
+                self.assertEqual(expected.shape, mesh.ndcrd.shape)
+                np.testing.assert_array_equal(expected, mesh.ndcrd.ndarray)
+                self.assertEqual(original.ctypes.data,
```
ThreeMonth03: A rejected assignment must preserve the original storage as well as its shape and values:

```text
before: (4, 2), original values, original data pointer
reject: source shape (2, 4)
after:  (4, 2), original values, original data pointer
```

The pointer comparison also works for empty arrays, where `np.shares_memory` cannot establish sharing because there are no elements. The separate matching-shape test checks that successful assignments still share the source buffer.

cpp/solvcon/python/common.hpp
```diff
@@ -215,6 +216,133 @@ std::enable_if_t<is_simple_array_v<S>, pybind11::array> to_ndarray(S && sarr)
...
+    const auto data_offset = static_cast<size_t>(-byte_span_begin);
+    auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(owner);
+    const auto buffer = ConcreteBuffer::construct(storage_nbytes, storage_ptr, std::move(remover));
+    return SimpleArray<T>(shape, stride, buffer, data_offset, array_order);
+}
+
+template <typename T>
+static void validate_array_shape(SimpleArray<T> const & array, pybind11::array const & ndarr)
+{
+    if (array.ndim() != ndarr.ndim() || !std::equal(array.shape().begin(), array.shape().end(), ndarr.shape()))
```
ThreeMonth03: Equal byte counts do not imply equal shapes:

```text
(4, 2) vs (8,)       -> same 64 bytes, different rank
(4, 2) vs (4, 2, 1) -> same leading axes, different rank
(4, 2) vs (2, 4)    -> same rank and bytes, different axis lengths
```

Check rank before `std::equal` so the comparison reads equally sized shape vectors. Both property setter helpers call this check before replacing the array.

tests/test_mesh.py
```diff
@@ -60,6 +63,132 @@ def _test(cls, ndim):
...
```
ThreeMonth03: A rejected assignment must preserve the original storage as well as its shape and values:

```text
before: (4, 2), original values, original data pointer
reject: source shape (2, 4)
after:  (4, 2), original values, original data pointer
```

The pointer comparison also works for empty arrays, where `np.shares_memory` cannot establish sharing because there are no elements. The separate matching-shape test checks that successful assignments still share the source buffer.

cpp/solvcon/python/common.hpp
```diff
@@ -374,7 +502,8 @@ WrapBase
                             << this_array.nbytes() << " bytes of internal array";
                         throw std::length_error(msg.str());
                     }
-                    makeSimpleArray(ndarr).swap(this_array);
+                    validate_array_shape(this_array, ndarr);
+                    make_array_from_numpy<typename array_type::value_type>(ndarr).swap(this_array);
```
ThreeMonth03: Reuse the same converter as `SimpleArray(array=...)`. A reversed view needs both its strides and an offset into the retained storage:

```python
base = np.arange(8, dtype='float64').reshape(4, 2)
source = base[::-1]
# NumPy byte strides: (-16, 8)
# SimpleArray strides: (-2, 1), logical origin: 48 bytes into the buffer
```

The converter was moved unchanged to the common header. It retains the reachable storage span and its NumPy owner, and rejects unsupported layouts before this swap.

tests/test_mesh.py
```diff
@@ -60,6 +63,132 @@ def _test(cls, ndim):
...
+                np.testing.assert_array_equal(expected,
+                                              mesh.ndcrd.clone().ndarray)
+
+                for current in (mesh, reference):
+                    current.cltpn.ndarray[:] = cell_type
+                    current.clnds.ndarray[0, :len(expected) + 1] = (
+                        len(expected), *range(len(expected)))
+                    current.build_interior()
+                    current.build_boundary()
+                    current.build_ghost()
```
ThreeMonth03: The assigned view must work in the mesh algorithms, beyond returning the right `ndarray` values:

```text
C / F / stepped / reversed coordinates
    -> build_interior -> build_boundary -> build_ghost
    -> match contiguous-reference coordinates and geometry
```

These cases run for a 2D triangle and a 3D tetrahedron. Before building, they also check shared writes, release the Python source references, and verify that the mesh still owns readable and cloneable data.

cpp/solvcon/python/common.hpp
ThreeMonth03: Equal byte counts do not imply equal shapes:

```text
(4, 2) vs (8,)       -> same 64 bytes, different rank
(4, 2) vs (4, 2, 1) -> same leading axes, different rank
(4, 2) vs (2, 4)    -> same rank and bytes, different axis lengths
```

Check rank before `std::equal` so the comparison reads equally sized shape vectors. Both property setter helpers call this check before replacing the array.

tests/test_mesh.py
```diff
@@ -60,6 +62,81 @@ def _test(cls, ndim):
...
+                mesh = solvcon.StaticMesh(ndim=2, nnode=nnode)
+                expected = np.arange(nnode * 2, dtype='float64').reshape(-1, 2)
+                dst = expected.copy()
+                mesh.ndcrd = dst
+
+                with self.assertRaisesRegex(ValueError, error):
+                    mesh.ndcrd = src
+
+                np.testing.assert_array_equal(expected, mesh.ndcrd.ndarray)
+                self.assertEqual(dst.ctypes.data,
```
ThreeMonth03: A rejected assignment must preserve values and the original data pointer:

```text
wrong rank / wrong shape / empty wrong shape / read-only input
    -> ValueError
    -> keep the original array
```

The destination first receives a NumPy buffer, giving even the empty case a stable pointer to compare.

cpp/solvcon/python/common.hpp
ThreeMonth03: Reuse the same converter as `SimpleArray(array=...)`. A reversed view needs both its strides and an offset into the retained storage:

```python
base = np.arange(8, dtype='float64').reshape(4, 2)
source = base[::-1]
# NumPy byte strides: (-16, 8)
# SimpleArray strides: (-2, 1), logical origin: 48 bytes into the buffer
```

The converter was moved unchanged to the common header. It retains the reachable storage span and its NumPy owner, and rejects unsupported layouts before this swap.

tests/test_mesh.py
```diff
@@ -60,6 +62,81 @@ def _test(cls, ndim):
...
+        view[:] = coords
+        meshes = []
+        for src in (coords, view):
+            mesh = solvcon.StaticMesh(ndim=2, nnode=3, ncell=1)
+            mesh.ndcrd = src
+            mesh.cltpn.ndarray[:] = solvcon.StaticMesh.TRIANGLE
+            mesh.clnds.ndarray[0, :4] = (3, 0, 1, 2)
+            mesh.build_interior()
+            mesh.build_boundary()
+            mesh.build_ghost()
```
ThreeMonth03: Keep one integration case: identical triangle coordinates in contiguous and reversed storage must produce the same mesh geometry after interior, boundary and ghost construction. Other layouts and owner lifetime have separate focused tests.

cpp/solvcon/python/common.hpp
ThreeMonth03, line 340: Equal byte counts do not imply equal shapes:

```text
(4, 2) vs (8,)       -> same 64 bytes, different rank
(4, 2) vs (4, 2, 1) -> same leading axes, different rank
(4, 2) vs (2, 4)    -> same rank and bytes, different axis lengths
```

Check rank before `std::equal` so the comparison reads equally sized shape vectors. Both property setter helpers call this check before replacing the array.

tests/test_mesh.py
```diff
@@ -60,6 +62,83 @@ def _test(cls, ndim):
...
+                expected = np.arange(nnode * 2, dtype='float64').reshape(-1, 2)
+                dst = expected.copy()
+                mesh.ndcrd = dst
+
+                with self.assertRaises(ValueError) as caught:
+                    mesh.ndcrd = src
+                self.assertEqual(error, str(caught.exception))
+
+                np.testing.assert_array_equal(expected, mesh.ndcrd.ndarray)
+                self.assertEqual(dst.ctypes.data,
```
ThreeMonth03, line 85: A rejected assignment must preserve values and the original data pointer:

```text
wrong rank / wrong shape / empty wrong shape / read-only input
    -> ValueError
    -> keep the original array
```

The destination first receives a NumPy buffer, giving even the empty case a stable pointer to compare.

cpp/solvcon/python/common.hpp
ThreeMonth03, line 506: Reuse the same converter as `SimpleArray(array=...)`. A reversed view needs both its strides and an offset into the retained storage:

```python
base = np.arange(8, dtype='float64').reshape(4, 2)
source = base[::-1]
# NumPy byte strides: (-16, 8)
# SimpleArray strides: (-2, 1), logical origin: 48 bytes into the buffer
```

The converter was moved unchanged to the common header. It retains the reachable storage span and its NumPy owner, and rejects unsupported layouts before this swap.

tests/test_mesh.py
```diff
@@ -60,6 +62,83 @@ def _test(cls, ndim):
...
+        view[:] = coords
+        meshes = []
+        for src in (coords, view):
+            mesh = solvcon.StaticMesh(ndim=2, nnode=3, ncell=1)
+            mesh.ndcrd = src
+            mesh.cltpn.ndarray[:] = solvcon.StaticMesh.TRIANGLE
+            mesh.clnds.ndarray[0, :4] = (3, 0, 1, 2)
+            mesh.build_interior()
+            mesh.build_boundary()
+            mesh.build_ghost()
```
ThreeMonth03, line 136: Keep one integration case: identical triangle coordinates in contiguous and reversed storage must produce the same mesh geometry after interior, boundary and ghost construction. Other layouts and owner lifetime have separate focused tests.

ThreeMonth03 (commented):

cpp/solvcon/python/common.hpp
```diff
@@ -215,6 +216,133 @@ std::enable_if_t<is_simple_array_v<S>, pybind11::array> to_ndarray(S && sarr)
...
+            break;
+        }
+        auto next = b.cast<py::array>();
+        /*
+         * Prevent the infinite loop.
+         * For example, the following code will create a loop:
+         * nparr = np.arange(24, dtype='float64').reshape((2, 3, 4))
+         * nparr = nparr[::2, ::2, ::2]
+         */
+        if (next.ptr() == owner.ptr())
```
yungyuc: The cyclic referencing check is only one level: `a.base is a`.  We should also prevent multi-level cyclic referencing check like `a.base.base is a` or `a.base.base.base is a`.

If you prefer to handle multi-level cyclic prevention in a later PR, make it explicit in PR comment.

yungyuc (approved):
- [x] Decide whether or not handle multi-level cyclic referencing in this PR.

cpp/solvcon/python/common.hpp
```diff
@@ -215,6 +216,106 @@ std::enable_if_t<is_simple_array_v<S>, pybind11::array> to_ndarray(S && sarr)
...
+        throw std::runtime_error(
+            std::format("NumPy data pointer is not aligned for item alignment {}", alignof(value_type)));
+    }
+    char * storage_ptr = view_ptr + byte_span_begin;
+    const size_t storage_nbytes = has_element
+                                      ? static_cast<size_t>(byte_span_end - byte_span_begin + itemsize)
+                                      : 0;
+    const auto data_offset = static_cast<size_t>(-byte_span_begin);
+    // The input retains its storage without traversing an overridable base.
+    auto remover = std::make_unique<ConcreteBufferNdarrayRemover>(arr_in);
```
ThreeMonth03, line 305: The input array already keeps its backing storage alive. Retain it directly so an ndarray subclass cannot make owner lookup loop through an overridden `.base` property. This also removes the single-level cycle guard.

```text
A.base -> B -> A
retain A; do not traverse .base
```

tests/test_buffer.py
```diff
@@ -450,6 +452,35 @@ def test_SimpleArray_from_numpy_negative_stride(self):
         self.assertEqual(200, ndarr[1, 2])
         self.assertEqual(100, clone.ndarray[0, 0])
 
+    def test_SimpleArray_base_cycle(self):
+        class CyclicArray(np.ndarray):
+            @property
+            def base(self):
+                # Fail on a second visit instead of hanging on a regression.
+                if self.visited:
```
ThreeMonth03, line 460: Raise on a repeated visit so the old traversal fails instead of hanging the test. Cover cycles of lengths 1, 2 and 3, then check the data pointer, storage lifetime after removing Python references, and release after destroying the SimpleArray.

ThreeMonth03 (commented):
@yungyuc Please review this pr again.

cpp/solvcon/python/common.hpp
yungyuc: So you rely on numpy/Python to keep the buffer alive in the same way `ConcreteBufferNdarrayRemover` does instead of chasing the base buffer ownership.  I like this direct approach and it is consistent in the code already implemented in solvcon.

yungyuc (approved):
LGTM

PR #1513, opened 2026-09-06
Validate NumPy contiguity in `ConcreteBuffer`

### LINKED ISSUE #1512: Reject non-contiguous NumPy inputs in `ConcreteBuffer`
`ConcreteBuffer(array=...)` requires contiguous NumPy storage, but the constructor does not check the layout. It wraps `nbytes` consecutive bytes starting at the view's data pointer, even when those bytes are not the view's elements.

```python
import numpy as np
import solvcon as sc

base = np.arange(8, dtype='int8')
source = base[::2]  # [0, 2, 4, 6]
buffer = sc.ConcreteBuffer(array=source)

print(buffer.ndarray.tolist())  # [0, 1, 2, 3]
buffer[1] = 99
print(base.tolist())            # [0, 99, 2, 3, 4, 5, 6, 7]
# base[1] is outside the source view.
```

Reject non-contiguous inputs before wrapping their pointers. Accept either C- or F-contiguous storage, preserving the existing shared byte view, including scalar and empty arrays.

```text
input layout                    expected
C- or F-contiguous              share the existing bytes
stepped / reversed / zero-stride reject if non-contiguous
```

Add regressions for rejected layouts and for the physical byte order and memory sharing of accepted inputs.

ThreeMonth03:
A non-contiguous NumPy view is currently treated as a contiguous byte range, so the returned buffer exposes bytes outside the view:

```python
import numpy as np
import solvcon as sc

base = np.arange(8, dtype='int8')
source = base[::2]  # [0, 2, 4, 6]
buffer = sc.ConcreteBuffer(array=source)
# Before: exposes [0, 1, 2, 3].
# After:  ValueError: ConcreteBuffer: input array must be C- or F-contiguous
```

Check NumPy's C/F-contiguous flags before wrapping the data pointer. Either flag is sufficient because `ConcreteBuffer` exposes physical bytes and carries no shape or strides:

```text
C-contiguous     -> share bytes in their existing order
F-contiguous     -> share bytes in their existing order
neither          -> reject before constructing the buffer
```

The regressions reject stepped, reversed, column-sliced and zero-stride inputs. They also check byte order and sharing for accepted C/F layouts, including singleton axes, empty arrays and scalars. Update the constructor documentation to state the accepted layouts and exception.

No-Qt buffer, mesh, grid, and multidimensional solver tests: 467 passed, 8 skipped. `make lint` passes.

Related to #1512.

### DIFF
--- a/cpp/solvcon/buffer/pymod/wrap_ConcreteBuffer.cpp
+++ b/cpp/solvcon/buffer/pymod/wrap_ConcreteBuffer.cpp
@@ -38,6 +38,10 @@ WrapConcreteBuffer::WrapConcreteBuffer(pybind11::module & mod, char const * pyna
             py::init(
                 [](py::array & arr_in, size_t alignment)
                 {
+                    if ((arr_in.flags() & (py::array::c_style | py::array::f_style)) == 0)
+                    {
+                        throw std::invalid_argument("ConcreteBuffer: input array must be C- or F-contiguous");
+                    }
                     return wrapped_type::construct(
                         arr_in.nbytes(), arr_in.mutable_data(), std::make_unique<ConcreteBufferNdarrayRemover>(arr_in), alignment);
                 }),
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -79,6 +79,44 @@ class ConcreteBufferBasicTC(unittest.TestCase):
         buf.ndarray.fill(0)
         self.assertTrue((ndarr == 0).all())
 
+    def test_from_ndarray_rejects_noncontiguous(self):
+        storage = np.arange(64, dtype='int8')
+        cases = (
+            ('step', storage[:8:2]),
+            ('reverse', storage[:8][::-1]),
+            ('columns', storage.reshape(8, 8)[:, ::2]),
+            ('zero', np.ndarray((4,), dtype='int8', buffer=storage,
+                                strides=(0,))),
+        )
+        for layout, source in cases:
+            with self.subTest(layout=layout):
+                with self.assertRaisesRegex(
+                        ValueError,
+                        'input array must be C- or F-contiguous'):
+                    solvcon.ConcreteBuffer(array=source)
+
+    def test_from_ndarray_contiguous_layouts(self):
+        shapes = ((2, 3), (1, 3), (0, 3), ())
+        for shape, order, dtype in itertools.product(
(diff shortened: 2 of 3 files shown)


cpp/solvcon/buffer/pymod/wrap_ConcreteBuffer.cpp
```diff
@@ -38,6 +38,10 @@ WrapConcreteBuffer::WrapConcreteBuffer(pybind11::module & mod, char const * pyna
             py::init(
                 [](py::array & arr_in, size_t alignment)
                 {
+                    if ((arr_in.flags() & (py::array::c_style | py::array::f_style)) == 0)
```
ThreeMonth03, line 41: The wrapper currently treats the logical byte count as a contiguous storage span:

```text
source = base[::2]   selects offsets: 0, 2, 4, 6
old byte buffer     exposes offsets: 0, 1, 2, 3
```

`ConcreteBuffer` cannot retain the source strides. Check contiguity before wrapping its pointer, accepting either C or F storage because both provide a contiguous byte range.

tests/test_buffer.py
```diff
@@ -79,6 +79,44 @@ def test_ConcreteBuffer_from_ndarray(self):
...
+                    solvcon.ConcreteBuffer(array=source)
+
+    def test_from_ndarray_contiguous_layouts(self):
+        shapes = ((2, 3), (1, 3), (0, 3), ())
+        for shape, order, dtype in itertools.product(
+                shapes, ('C', 'F'), ('int8', 'float64')):
+            with self.subTest(shape=shape, order=order, dtype=dtype):
+                source = np.empty(shape, dtype=dtype, order=order)
+                source.flat = np.arange(source.size, dtype=dtype)
+                expected = np.frombuffer(source.tobytes(order='A'),
```
ThreeMonth03, line 105: For an F-contiguous source, the byte view must preserve physical order:

```text
logical array:  [[0, 1, 2], [3, 4, 5]]
physical bytes: [0, 3, 1, 4, 2, 5]  (int8, F-order)
```

The byte oracle uses `order='A'`, and the pointer and shared-write checks catch an implementation that silently copies into C-order. Scalar, empty and singleton-axis cases also keep NumPy's contiguity rules.

ThreeMonth03 (commented):
LGTM

PR #1517, opened 2026-09-07
Fix segfault reopening a Canvas/Samples window

### LINKED ISSUE #1471: Segmentation fault when reopening a Canvas sample after closing the previous one
Reopening a Canvas/Samples window after closing the previously opened one crashes the pilot GUI with a segmentation fault. The whole pilot process dies, not just the canvas window. 

This happens for any entry under Canvas/Samples and only after the canvas window has been closed first. Picking a different sample while a canvas window is still open does not crash; it overlays the new sample onto the existing canvas instead of crashing.

### Steps to reproduce
1. Build and launch the pilot GUI.
2. Open the "Canvas" menu, then "Samples", and pick any sample.
3. Close the canvas window that was opened.
4. Open "Canvas/Samples" again and pick any sample from the submenu.

### Expected behavior
A new canvas window opens showing the selected sample, same as the first time.

### Actual behavior
The program segfaults on step 4 and the entire pilot window closes.

SivanYeh:
Fixes a segfault when closing a Canvas sample-pick window.

Canvas caches its 2D widget across sample picks, but never cleared that cache on window close. Since Qt frees the underlying widget on close, the cache was left pointing at freed memory — the next access segfaulted instead of raising a normal Python exception.

The fix adds a close filter that clears the cached reference before Qt deletes the widget, so the next pick rebuilds a fresh one. The filter itself moved from obsrefl/_viewer.py into the shared pilot/base/_gui_common.py; see inline comments for why.

Related to #1471.

### DIFF
--- a/solvcon/pilot/apps/obsrefl/_viewer.py
+++ b/solvcon/pilot/apps/obsrefl/_viewer.py
@@ -18,6 +18,7 @@ close through the :attr:`closed` callback.
 
 from PySide6.QtCore import Qt, QObject, QEvent
 
+from ...base import _gui_common
 from ._colorbar import ColorBar
 from ._plots import AnalysisLinePlots
 
@@ -25,23 +26,6 @@ __all__ = [  # noqa: F822
     'DomainViewer',
     'LinePlotViewer',
 ]
-
-
-class _SubWindowCloseFilter(QObject):
-    """Report a watched sub-window's close synchronously.
-
-    A ``QMdiSubWindow`` has no close signal, so this filter is the only way
-    to stop the march before Qt frees the viewer.
-    """
-
-    def __init__(self, on_close, parent):
-        super().__init__(parent)
-        self._on_close = on_close
-
-    def eventFilter(self, obj, event):
-        if event.type() == QEvent.Type.Close:
-            self._on_close()
-        return False
 
 
 class _ResizeRelay(QObject):
@@ -109,7 +93,7 @@ class DomainViewer(object):
         self._subwin = self._mdi.activeSubWindow()
         if self._subwin is not None:
             self._subwin.setAttribute(Qt.WA_DeleteOnClose, True)
-            self._close_filter = _SubWindowCloseFilter(
+            self._close_filter = _gui_common.SubWindowCloseFilter(
                 self._on_subwin_closed, self._subwin)
             self._subwin.installEventFilter(self._close_filter)
             host = self._host()
@@ -273,7 +257,7 @@ class LinePlotViewer(object):
         self._subwin.setWindowTitle("reflection analysis")
         self._subwin.resize(*self.SIZE)
         self._mgr.addSubWindowGrip(self._subwin)
-        self._close_filter = _SubWindowCloseFilter(
+        self._close_filter = _gui_common.SubWindowCloseFilter(
             self._on_subwin_closed, self._subwin)
         self._subwin.installEventFilter(self._close_filter)
 
--- a/solvcon/pilot/base/_gui_common.py
+++ b/solvcon/pilot/base/_gui_common.py
@@ -70,6 +70,25 @@ class ToggleActionBridge(QtCore.QObject):
         self._store_changed.emit(tg.get(self._key, self._action.isChecked()))
 
 
+class SubWindowCloseFilter(QtCore.QObject):
+    """Report a watched sub-window's close synchronously.
+
+    A ``QMdiSubWindow`` has no close signal, so this filter is the only way
+    for an owner to drop its cached widget reference before Qt frees the
+    widget the sub-window wraps (``Qt.WA_DeleteOnClose`` runs the delete on
+    the next event-loop turn, not inside ``close()`` itself).
+    """
+
+    def __init__(self, on_close, parent):
+        super().__init__(parent)
+        self._on_close = on_close
+
+    def eventFilter(self, obj, event):
+        if event.type() == QtCore.QEvent.Type.Close:
+            self._on_close()
+        return False
+
+
 _SHORTCUT_CONTEXTS = {
     "application": QtCore.Qt.ApplicationShortcut,
     "window": QtCore.Qt.WindowShortcut,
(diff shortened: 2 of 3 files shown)


solvcon/pilot/apps/obsrefl/_viewer.py
```diff
@@ -27,23 +28,6 @@
 ]
 
 
-class _SubWindowCloseFilter(QObject):
```
SivanYeh, line 30: Close filter moved to pilot/base/_gui_common.py so Canvas can share
it without an app-specific import — see _gui_common.py.

solvcon/pilot/base/_gui_common.py
```diff
@@ -70,6 +70,25 @@ def _on_store_changed(self):
         self._store_changed.emit(tg.get(self._key, self._action.isChecked()))
 
 
+class SubWindowCloseFilter(QtCore.QObject):
```
SivanYeh, line 73: Moved here since Canvas needs this too and isn't obsrefl-specific.
Same pattern as DomainViewer.

solvcon/pilot/canvas/_canvas_gui.py
```diff
@@ -346,11 +348,40 @@ def mesh_iccad_2013(self):
...
+    def _clear_widget_2d(self):
+        self._widget_2d = None
+        self._widget_2d_close_filter = None
+
     def _update_widget(self):
         # The canvas world is planar geometry, so it renders in the 2D canvas
         # (the 3D domain viewer is for meshes and fields).
         if self._widget is None:
             self._widget = self._mgr.add2DWidget()
+            self._widget_close_filter = self._track_widget(
```
SivanYeh, line 393: Fires before Qt deletes the widget, so we clear the cache while
it's still valid.

solvcon/pilot/canvas/_canvas_gui.py
```diff
@@ -346,11 +348,40 @@ def mesh_iccad_2013(self):
         self._draw_layer(self._world, layer)
         self._update_widget()
 
+    def _track_widget(self, clear):
```
SivanYeh, line 355: Stale after close: QMdiArea's WA_DeleteOnClose frees the widget,
but this pointer was never cleared — next access hits freed memory.

solvcon/pilot/canvas/_canvas_gui.py
```diff
@@ -346,11 +348,40 @@ def mesh_iccad_2013(self):
...
+        the widget it wraps.
+
+        Closing a canvas sub-window deletes the underlying C++ widget
+        (``QMdiArea.addSubWindow`` sets ``WA_DeleteOnClose`` by default), so
+        a reference cached across sample picks goes stale the moment the
+        window closes. ``clear`` runs before that happens, so the next
+        sample or open call rebuilds a fresh, live widget instead of
+        touching a freed one.
+        """
+        subwin = self._mgr.mdiArea.activeSubWindow()
```
tigercosmos: `self._mgr.mdiArea` object will have a lifetime issue.
You can do `self._mdiArea = self._mgr.mdiArea` to keep it.

tigercosmos (changes requested):

solvcon/pilot/canvas/_canvas_gui.py
yungyuc: The lifetime issue is that the return (`subwin`) of `mdiArea.activeSubWindow()` does not keep the life of `mdiArea`.  See https://github.com/solvcon/solvcon/pull/1163/changes#diff-d847065d580e55936ae8f0fa032e163f55f825b8d0af4fc353972e6d9964a47eR262-R265:

```python
        # Held for the panel's lifetime: a throwaway mdiArea wrapper is
        # garbage-collected right after use, invalidating any sub-window handle
        # taken through it.
        self._mdi = None
```

A more complete solution is to fix it in shiboken, but it should be a separate work.

solvcon/pilot/canvas/_canvas_gui.py
```diff
@@ -346,11 +348,40 @@ def mesh_iccad_2013(self):
...
+        if subwin is None:
+            return None
+        subwin.setAttribute(QtCore.Qt.WA_DeleteOnClose, True)
+        close_filter = _gui_common.SubWindowCloseFilter(clear, subwin)
+        subwin.installEventFilter(close_filter)
```
yungyuc, line 377: Clarify why you do not need to prevent installation of the filter twice here in PR review.

yungyuc (changes requested):
- [x] Fix `mdiArea` lifetime issue.
- [x] Clarify if there is double installation issue.

solvcon/pilot/canvas/_canvas_gui.py
SivanYeh: Thanks for @tigercosmos's suggestion and @yungyuc's explanation point at the same fix! Fixed in eb7d41b: I now cache self._mdi = self._mgr.mdiArea once (mirroring DomainViewer.open's pattern in _viewer.py) and reuse self._mdi for the activeSubWindow() lookup instead of re-accessing the mdiArea property each time. See https://github.com/solvcon/solvcon/pull/1517/changes/eb7d41b79202e0696385a26c562904baf6186fb5#diff-3caad8a28480c7d6251789fa8ba67c491d2706109de40f951a449c6024077df4R370

solvcon/pilot/canvas/_canvas_gui.py
SivanYeh: `_track_widget` is only ever invoked immediately following the instantiation of a brand-new subwindow—specifically inside `_update_widget` and `_open_2d`, both of which are gated by `if self._widget is None:`.
Because of this guard:
> * **While open:** `self._widget` remains populated, preventing any redundant calls to `_track_widget` on an active instance.
> * **On close:** The filter's callback resets `self._widget` to `None` right before Qt destroys the C++ widget via `WA_DeleteOnClose`.
> * **On reopen:** Any subsequent request triggers the creation of an entirely new `QMdiSubWindow`, which receives its own initial filter.
> 
> 
Consequently, every subwindow instance undergoes `installEventFilter` exactly once throughout its lifecycle; duplicate installation on the same object is structurally impossible. This invariant has also been documented directly in `_track_widget` in commit [eb7d41b.](https://github.com/solvcon/solvcon/pull/1517/changes/eb7d41b79202e0696385a26c562904baf6186fb5#diff-3caad8a28480c7d6251789fa8ba67c491d2706109de40f951a449c6024077df4R365)

yungyuc (approved):
LGTM

Issue #1479, opened 2026-09-03
Post-merge CI failure detection and assignment


yungyuc:
CI sometimes fails after merging, e.g.,

- 09/02, merge b08336ac, `modernize-use-auto` at `SimpleArray.hpp:747` and `:761`. Introduced by #1460 (ThreeMonth03). Not fixed: `master` is red as this is written, with no fix pull request open.
- 08/31, merge 62e86088, `bugprone-signed-char-misuse` at `TypeBroadcast.hpp:151` and `:162`. Introduced by #1448 (ThreeMonth03), fixed by #1455 (ThreeMonth03) after 1 h 28 min.
- 08/30, merge c2a853fb, `google-objc-global-variable-declaration` at `base.hpp:62`, reached through `metal.mm` on macOS only. Introduced by [#1408](https://github.com/solvcon/solvcon/pull/1408) (ThreeMonth03), fixed by #1444 (yungyuc) after 1 h 32 min.
- 08/29, merge 47f7fc1d, `cppcoreguidelines-pro-type-member-init` at `World.hpp:520`. Introduced by #1348 (tychuang1211), fixed by #1430 (yungyuc) after 1 h 34 min.

When there is a red `master`, `send_email_on_fail.yml` mails `solvcon@googlegroups.com` to tell the list that something broke, but does not tell anyone to fix it.

We need something to automatically open an issue to name the pull request that fails a merge and name the PR author in it.  We may consider to implement it using a post-merge workflow to detect the offending pull request open an issue to reference the PR and mention its author.  For one failure there should only one issue opened.

rockleona:
Hi @yungyuc, I like to take this task

PR #1518, opened 2026-09-07
Add open_issue_on_failure workflow for post-merge failures

rockleona:
Add `open_issue_on_failure` workflow for post-merge failures

`send_email_on_fail.yml` mails `solvcon@googlegroups.com` when a master push breaks CI, but the mail never says which merged pull request caused it or who to ask to fix it. Add `.github/workflows/report_ci_failure.yml`, a reusable workflow that resolves the pull request for the failing commit via `gh api repos/{repo}/commits/{sha}/pulls`, then `gh issue create`s an issue naming the pull request and labels it `build`. The issue is assigned to a fixed pair (`chestercheng`, `rockleona`) rather than the pull request author, so post-merge failures always reach whoever triages them; the author is still named in the issue text. A repeat failure on a pull request that already has an open report gets a `gh issue comment` on that issue instead of a duplicate. Both the issue and the comment are formatted with GitHub markdown.

Call the new workflow from `devbuild.yml`, `lint.yml`, and `devbuild_windows.yml` as a job alongside the existing `send_email_on_failure` job, reusing its `needs`/`if` gate verbatim so it fires under the same push-to-master-failure conditions.

Related to #1479

### DIFF
--- a/.github/workflows/devbuild.yml
+++ b/.github/workflows/devbuild.yml
@@ -363,3 +363,25 @@ jobs:
     secrets:
       EMAIL_USERNAME: ${{ secrets.EMAIL_USERNAME }}
       EMAIL_PASSWORD: ${{ secrets.EMAIL_PASSWORD }}
+
+  open_issue_on_failure:
+    needs: [check_skip, standalone_buffer, build, build_cuda]
+    # Report a master push failure only, on the same gate as
+    # send_email_on_failure above.
+    if: |
+      always() &&
+      (contains(needs.*.result, 'failure') ||
+       contains(needs.*.result, 'cancelled')) &&
+      github.event_name == 'push' &&
+      github.ref == 'refs/heads/master' &&
+      github.event.repository.fork == false
+    permissions:
+      issues: write
+      pull-requests: read
+    uses: ./.github/workflows/open_issue_on_failure.yml
+    with:
+      job_results: |
+        - check_skip: ${{ needs.check_skip.result }}
+        - standalone_buffer: ${{ needs.standalone_buffer.result }}
+        - build: ${{ needs.build.result }}
+        - build_cuda: ${{ needs.build_cuda.result }}
--- a/.github/workflows/open_issue_on_failure.yml
+++ b/.github/workflows/open_issue_on_failure.yml
@@ -43,7 +43,7 @@ jobs:
           existing_issue=$(gh issue list \
             --repo "$GITHUB_REPOSITORY" \
             --state open \
-            --limit 100 \
+            --search "$title" \
             --json number,title \
             --jq "map(select(.title == \"$title\")) | .[0].number // empty")
 
--- a/.github/workflows/open_issue_on_failure.yml
+++ b/.github/workflows/open_issue_on_failure.yml
@@ -1,0 +1,91 @@
+name: open_issue_on_failure
+
+on:
+  workflow_call:
+    inputs:
+      job_results:
+        description: 'String of job results'
+        required: true
+        type: string
+
+jobs:
+  open_issue_on_failure:
+    name: open_issue_for_pr_that_broke_master
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: read
+    steps:
+      - name: Find and notify the pull request that broke master
+        env:
+          GH_TOKEN: ${{ github.token }}
+          JOB_RESULTS: ${{ inputs.job_results }}
+        run: |
+          echo "::group::Resolve the pull request for ${{ github.sha }}"
+          pr_number=$(gh api "repos/$GITHUB_REPOSITORY/commits/${{ github.sha }}/pulls" --jq '.[0].number // empty')
+          pr_author=$(gh api "repos/$GITHUB_REPOSITORY/commits/${{ github.sha }}/pulls" --jq '.[0].user.login // empty')
+          echo "pr_number=${pr_number:-<none>} pr_author=${pr_author:-<none>}"
+          echo "::endgroup::"
+
+          if [ -z "$pr_number" ]; then
+            echo 'No pull request is associated with this commit; nothing to report.'
+            exit 0
+          fi
+
+          # The title carries only the PR number so that any post-merge
+          # workflow failing on this PR lands on the same tracking issue.
+          title="CI failure after merging #${pr_number}"
+          run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          # The fixed pair triages post-merge CI regardless of who broke it;
+          # the pull request author is still named in the issue text below.
+          assignees='chestercheng,rockleona'
+
+          existing_issue=$(gh issue list \
+            --repo "$GITHUB_REPOSITORY" \
+            --state open \
+            --limit 100 \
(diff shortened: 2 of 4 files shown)


.github/workflows/devbuild.yml
```diff
@@ -363,3 +363,25 @@ jobs:
...
+    needs: [check_skip, standalone_buffer, build, build_cuda]
+    # Report a master push failure only, on the same gate as
+    # send_email_on_failure above.
+    if: |
+      always() &&
+      (contains(needs.*.result, 'failure') ||
+       contains(needs.*.result, 'cancelled')) &&
+      github.event_name == 'push' &&
+      github.ref == 'refs/heads/master' &&
+      github.event.repository.fork == false
```
rockleona, line 377: Only works for master branch, the workflow needs to be triggered in each triggers.

.github/workflows/open_issue_on_failure.yml
```diff
@@ -0,0 +1,89 @@
...
+          if [ -z "$pr_number" ]; then
+            echo 'No pull request is associated with this commit; nothing to report.'
+            exit 0
+          fi
+
+          title="CI failure in ${{ github.workflow }} after merging #${pr_number}"
+          run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          # The fixed pair triages post-merge CI regardless of who broke it;
+          # the pull request author is still named in the issue text below.
+          assignees='chestercheng,rockleona'
```
rockleona, line 41: Issue assignees are me and @chestercheng by our discussion, please let me know if this is not a proper behavior

.github/workflows/report_ci_failure.yml
```diff
@@ -0,0 +1,89 @@
...
+          pr_author=$(gh api "repos/$GITHUB_REPOSITORY/commits/${{ github.sha }}/pulls" --jq '.[0].user.login // empty')
+          echo "pr_number=${pr_number:-<none>} pr_author=${pr_author:-<none>}"
+          echo "::endgroup::"
+
+          if [ -z "$pr_number" ]; then
+            echo 'No pull request is associated with this commit; nothing to report.'
+            exit 0
+          fi
+
+          title="CI failure in ${{ github.workflow }} after merging #${pr_number}"
```
rockleona: I did not come out a better title, still struggling if the name do not bring out precise info.

.github/workflows/open_issue_on_failure.yml
```diff
@@ -0,0 +1,89 @@
...
+
+          **Job status**
+
+          ${JOB_RESULTS}
+
+          **Run:** ${run_url}
+          EOF
+          )"
+            exit 0
+          fi
```
rockleona, line 65: If the issue had been created, will leave a comment under the existed issue.

yungyuc (approved):
LGTM.

@rockleona Do you have an example issue created in your fork or somewhere?  It helps understand how it will look.

I can merge it as is so that we can directly experiment in the upstream repo.  Let me know what you want me to do.

@ExplorerRay @chestercheng Please help take a look.

.github/workflows/report_ci_failure.yml
chestercheng: I think it would be better to remove the workflow name from the title.

Since the current title includes the workflow name, the same merge could create separate issues if it fails in more than one workflow.

I would prefer a single issue per merged PR.

chestercheng (changes requested):

.github/workflows/devbuild.yml
```diff
@@ -363,3 +363,25 @@ jobs:
     secrets:
       EMAIL_USERNAME: ${{ secrets.EMAIL_USERNAME }}
       EMAIL_PASSWORD: ${{ secrets.EMAIL_PASSWORD }}
+
+  report_ci_failure:
```
ExplorerRay: I feel this name is a little bit too general. Sending email is also a kind of report ci failure. It might be okay to combine these two into one unified workflow called `report_ci_failure` in different steps.

If we don't want to combine them, I think some names like `open_issue_on_failure` looks more clear for me.

.github/workflows/open_issue_on_failure.yml
```diff
@@ -0,0 +1,91 @@
...
+          title="CI failure after merging #${pr_number}"
+          run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          # The fixed pair triages post-merge CI regardless of who broke it;
+          # the pull request author is still named in the issue text below.
+          assignees='chestercheng,rockleona'
+
+          existing_issue=$(gh issue list \
+            --repo "$GITHUB_REPOSITORY" \
+            --state open \
+            --limit 100 \
```
ExplorerRay: I am thinking whether we need this limit when we already have jq arg for filter title.

.github/workflows/devbuild.yml
rockleona: I will change to `open_issue_on_failure`, for me I don't want to combine it with sending mail in this stage. We may consider integrate next PR.

.github/workflows/open_issue_on_failure.yml
rockleona: I think this is a good practice when the number of open issues are over 100. IMO, we won't take care of CI failed issue for a long time, so I think the limit is reasonable.

ExplorerRay: I originally thought `--jq` would filter across all open issues, but I found there's a default limit of 30. Using `--search` to narrow by title and `--jq` to confirm the exact match seems like a good practice, since then we don't have to worry about older issues falling outside the window.

So the purpose of this arg is to widen the window because the default 30 is too narrow?

ExplorerRay: Another thing comes to mind: will this `existing_issue` comment be triggered often? Since every PR is only merged to master once, it seems like only a manual re-run of the workflow would reach that comment code. 

Does manual re-run workflow happens often? If not, maybe we don't need to handle the case.

rockleona: > I originally thought `--jq` would filter across all open issues, but I found there's a default limit of 30. Using `--search` to narrow by title and `--jq` to confirm the exact match seems like a good practice, since then we don't have to worry about older issues falling outside the window.
> 
> So the purpose of this arg is to widen the window because the default 30 is too narrow?

`--jq` is just a filter to help us to search on JSON. The default limit of 30 was defined by `gh issue list`, and I was intentionally to make it larger to 100 for visibility.

However `--search` could help us improving in this case, I will change to use this param.

> Another thing comes to mind: will this `existing_issue` comment be triggered often? Since every PR is only merged to master once, it seems like only a manual re-run of the workflow would reach that comment code.
> 
> Does manual re-run workflow happens often? If not, maybe we don't need to handle the case.

Merge is just a one-time action, this is correct, however, master branch will have scheduled daily jobs to trigger workflows again. You may take the frequence as the same like email sending.

ExplorerRay: Ah, I see. I didn't notice that the scheduled job would resolve to the same PR. Thanks for the clarification.

chestercheng (approved):
@rockleona thanks for the update. It looks to me now.

PR #1519, opened 2026-09-09
Let the 2D canvas sub-window carry a resize grip

tigercosmos:
`add2DWidget` now calls `RManager::addSubWindowGrip`, so the sub-window that hosts a 2D canvas carries a resize grip in its lower-right corner.

The sub-window frame is a few pixels wide, which makes a resize drag hard to aim. `addSubWindowGrip` is the one entry point for the grip, and `add3DWidget` and the obsrefl line-plot window already call it. `add2DWidget` was the only sub-window factory that did not.

`RManagerSubWindowGripTC` already pins what the grip does once it is there, so the new test in `tests/test_pilot_2d.py` only checks that the canvas asks for one.

### DIFF
--- a/cpp/solvcon/pilot/app/RManager.cpp
+++ b/cpp/solvcon/pilot/app/RManager.cpp
@@ -267,6 +267,7 @@ R2DWidget * RManager::add2DWidget()
         auto * subwin = this->addSubWindow(viewer);
         subwin->resize(400, 300);
         viewer->resize(400, 300);
+        this->addSubWindowGrip(subwin);
     }
     return viewer;
 }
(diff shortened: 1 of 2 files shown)


cpp/solvcon/pilot/app/RManager.cpp
```diff
@@ -267,6 +267,7 @@ R2DWidget * RManager::add2DWidget()
         auto * subwin = this->addSubWindow(viewer);
         subwin->resize(400, 300);
         viewer->resize(400, 300);
+        this->addSubWindowGrip(subwin);
```
tigercosmos, line 270: Meat

Issue #1521, opened 2026-09-09
Read single-row and single-column CSV files in DataFrame


ThreeMonth03:
`DataFrame.read_from_text_file()` raises `IndexError` for a nonempty CSV with only one data row or one column.

```python
import io
from solvcon.track.dataframe import DataFrame

for text, has_timestamp in (
    ('time,value\n1,2\n', True),
    ('value\n1\n2\n', False),
):
    frame = DataFrame()
    try:
        frame.read_from_text_file(
            io.StringIO(text), timestamp_in_file=has_timestamp)
        print(frame['value'].tolist())
    except IndexError as error:
        print(type(error).__name__, str(error))
# Current: IndexError for both inputs.
# Expected: [2.0], then [1.0, 2.0].
```

`np.genfromtxt` squeezes these inputs to one dimension, but the reader subsequently indexes a two-dimensional array. Preserve both dimensions when parsing. Add regressions for one row, one column, one data cell, and ordinary multi-row input, checking column names, indices, and values. Empty-file behavior is outside this task.

Reproduced at [52d4f6bb](https://github.com/solvcon/solvcon/blob/52d4f6bb22ac7f40786bbbaad2ae1fbe76bc5588/solvcon/track/dataframe.py#L144-L165), using Python 3.12.7 and NumPy 2.5.1.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

ThreeMonth03:
There are no current plans to proceed with this work, so I'm closing this issue for now.

PR #1524, opened 2026-09-09
Fix singleton CSV dimensions in DataFrame reader

nightcityblade:
Pass ``ndmin=2`` to ``numpy.genfromtxt`` so CSV input retains separate row and column axes even when either dimension has length one. Add regression coverage for one-row, one-column, and one-cell inputs, including column names, indices, and values.

Related to #1521.


ThreeMonth03 (changes requested):
Coding style is bad. Please elaborate this pr.

PR #1528, opened 2026-09-10
Skip the Winograd boundary case on the macOS runner

tigercosmos:
The nightly_profiling macOS job dies at its 30-minute timeout, which reports as `cancelled` and mails a failure. It did so on two consecutive nights. Almost all of that time goes to `profiling/profile_matrix_ops.py`, and almost all of that to the square case at the Winograd threshold. That case needs a 16384-square operand because `MatmulTuning::Winograd::minimum_side` is 16384, so the float64 pair peaks near 7 GiB and the runner swaps. The float32 pair peaks near 4 GiB and fits. The ubuntu runner has the memory for both and finishes the same script in about 11 minutes.

This change adds `--skip-winograd-boundary`, which names one dtype and repeats to skip both, and sets it to `float64` on the macOS matrix entry. The macOS job keeps the float32 measurement and the ubuntu entry keeps both.

`make pyprof` ran every profiling script bare, so the options `profile_matrix_ops.py` already parses could not reach it from CI. The target now forwards options per script through `PYPROF_OPTS_<script>`, the way `pytest` forwards `PYTEST_OPTS`. A flag for one script does not reach the other six.

I ran `make pyprof PROFFILES=profiling/profile_matrix_ops.py` on macOS four ways. The default takes 1m34s and reports both dtypes. `--skip-winograd-boundary float64` takes 28s and reports float32 alone. Repeating the flag for both dtypes takes 4s. An unknown dtype fails at argument parsing. `make lint` passes.

### DIFF
--- a/.github/workflows/nightly_profiling.yml
+++ b/.github/workflows/nightly_profiling.yml
@@ -42,8 +42,15 @@ jobs:
         include:
           - os: ubuntu-24.04
             cmake_build_type: Release
+            pyprof_matrix_opts: ''
           - os: macos-26
             cmake_build_type: RelWithDebInfo
+            # The float64 case at the Winograd threshold peaks near 7 GiB.
+            # The macOS runner swaps on it, which ran the job past its
+            # 30-minute timeout on two consecutive nightly builds. The
+            # float32 case peaks near 4 GiB and still runs here. The ubuntu
+            # runner has the memory and keeps measuring both dtypes.
+            pyprof_matrix_opts: '--skip-winograd-boundary float64'
       fail-fast: false
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly_profiling.yml
```diff
@@ -42,8 +42,15 @@ jobs:
...
             cmake_build_type: Release
+            pyprof_matrix_opts: ''
           - os: macos-26
             cmake_build_type: RelWithDebInfo
+            # The float64 case at the Winograd threshold peaks near 7 GiB.
+            # The macOS runner swaps on it, which ran the job past its
+            # 30-minute timeout on two consecutive nightly builds. The
+            # float32 case peaks near 4 GiB and still runs here. The ubuntu
+            # runner has the memory and keeps measuring both dtypes.
+            pyprof_matrix_opts: '--skip-winograd-boundary float64'
```
tigercosmos, line 53: Skip `float64` to avoid macOS memory swap issue.

yungyuc: @ThreeMonth03 Can you create an issue to change the case to use smaller memory, while making the larger memory test opt-in?  CI should be fixed ASAP so we use a new issue to track the improvement for the test itself.

yungyuc (approved):
LGTM

@ExplorerRay @chestercheng FYI

Issue #1245, opened 2026-07-31
Pilot bug: `Window` > `Cascade` collapses the sub-windows to bare title bars


yungyuc:
The `Cascade` action added in PR #1229 offsets the open MDI sub-windows correctly, but it also shrinks every one of them to the style's minimum frame size. What is left is a stack of stubs that show the window buttons and nothing else: no title text, no viewer content.

The [review comment](https://github.com/solvcon/solvcon/pull/1229#issuecomment-5133376650) on PR #1229 reports exactly this, with a screenshot of the collapsed result next to a hand-arranged cascade that keeps each title readable.

## Steps to reproduce

1. Start the pilot and open a few viewers (`2D canvas` and `Domain viewer` in any mix).
2. Choose `Window` > `Cascade`.

Observed: the sub-windows line up diagonally at the top-left corner, each about one title bar tall and barely wider than its window buttons. The titles are elided away and the viewers are no longer usable without resizing every frame by hand.

<img width="3148" height="2122" alt="Image" src="https://github.com/user-attachments/assets/dce465b6-b5f7-493c-8979-9461f77c1ac0" />

Expected: the sub-windows keep a usable size, offset by one title bar each, so every title stays readable and the frontmost viewer still fills a good part of the area.

## Analysis

[`WindowManager._cascade`](https://github.com/solvcon/solvcon/blob/a84df895966d4b8c7efd88fb57cbef305fe54c78/solvcon/pilot/panel/_window_manager.py#L94-L95) delegates straight to Qt:

```python
def _cascade(self):
    self._mgr.mdiArea.cascadeSubWindows()
```

We should also review a reasonable way to size the sub-windows when rearranging them.

Yuli-ytd:
I'm interested in working on this issue and would like to take it.

PR #1529, opened 2026-09-10
Give cascaded windows a consistent size

Yuli-ytd:
## Summary

This PR preserves the positions computed by Qt during Cascade and resizes each visible, non-minimized sub-window to `400` by `300`. The fixed size matches the initial dimensions used by the 2D and 3D viewer factories.

Minimized sub-windows retain both their state and size. The active sub-window remains unchanged, and the viewer content remains usable.

Manual verification used 2D and 3D viewers.
截圖 2026-09-10 14 45 43

## Observation

On macOS 26 with Qt 6.11.1, sub-window titles are absent immediately after Cascade. Clicking the active sub-window displays only its title. Clicking an inactive sub-window, or switching away from and back to pilot, displays every title. The geometry and viewer content remain unchanged.

| Immediately after Cascade | After clicking the active sub-window | After clicking an inactive sub-window |
| --- | --- | --- |
| 截圖 2026-09-09 23 00 52 | 截圖 2026-09-09 23 01 11 | 截圖 2026-09-09 23 01 28 |

Calling `repaint()` after `resize()` did not make the missing titles appear. The cause remains unknown, and the behavior has not been verified on other platforms. Verification on Linux, Windows, or other Qt versions would help determine whether the behavior is specific to macOS, Qt 6.11.1, or their interaction.

This PR does not include the tested `repaint()` call because it did not change the observed behavior.

Does the macOS title-bar behavior block this change, or should it be tracked separately as a Qt/platform follow-up?

Related to #1245.

### DIFF
--- a/solvcon/pilot/panel/_window_manager.py
+++ b/solvcon/pilot/panel/_window_manager.py
@@ -91,7 +91,7 @@ class WindowManager(_gui_common.PilotFeature):
     def _tile(self):
         self._mgr.mdiArea.tileSubWindows()
 
-    def _cascade(self):
+    def _cascade(self, width=400, height=300):
         mdi = self._mgr.mdiArea
         subwins = [s for s in mdi.subWindowList() if s.isVisible()]
         if not subwins:
@@ -101,7 +101,7 @@ class WindowManager(_gui_common.PilotFeature):
 
         for subwin in subwins:
             if not subwin.isMinimized():
-                subwin.resize(400, 300)
+                subwin.resize(width, height)
 
     def _set_tabbed(self, on):
         """Switch the MDI area between sub-window and tabbed view.
--- a/solvcon/pilot/panel/_window_manager.py
+++ b/solvcon/pilot/panel/_window_manager.py
@@ -92,7 +92,16 @@ class WindowManager(_gui_common.PilotFeature):
         self._mgr.mdiArea.tileSubWindows()
 
     def _cascade(self):
-        self._mgr.mdiArea.cascadeSubWindows()
+        mdi = self._mgr.mdiArea
+        subwins = [s for s in mdi.subWindowList() if s.isVisible()]
+        if not subwins:
+            return
+
+        mdi.cascadeSubWindows()
+
+        for subwin in subwins:
+            if not subwin.isMinimized():
+                subwin.resize(400, 300)
 
     def _set_tabbed(self, on):
         """Switch the MDI area between sub-window and tabbed view.
--- a/tests/gui/test_window_menu.py
+++ b/tests/gui/test_window_menu.py
@@ -205,12 +205,66 @@ class WindowLayoutTC(unittest.TestCase):
         one, two = [s.geometry() for s in self.area.subWindowList()]
         self.assertFalse(one.intersects(two))
 
-    def test_cascade_offsets_the_windows(self):
+    def test_cascade_preserves_qt_positions(self):
+        self._two_stacked_windows()
+        subwins = self.area.subWindowList()
+
+        self.area.cascadeSubWindows()
+        QtWidgets.QApplication.processEvents()
+        qt_positions = [subwin.pos() for subwin in subwins]
+        self.assertNotEqual(qt_positions[0], qt_positions[1])
+
+        for subwin in subwins:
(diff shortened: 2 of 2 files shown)


solvcon/pilot/panel/_window_manager.py
```diff
@@ -92,7 +92,16 @@ def _tile(self):
         self._mgr.mdiArea.tileSubWindows()
 
     def _cascade(self):
-        self._mgr.mdiArea.cascadeSubWindows()
+        mdi = self._mgr.mdiArea
+        subwins = [s for s in mdi.subWindowList() if s.isVisible()]
+        if not subwins:
+            return
+
+        mdi.cascadeSubWindows()
```
Yuli-ytd, line 100: Qt remains responsible for calculating the cascade offsets. The resize loop changes only the width and height, so the platform's cascade positions remain unchanged.

solvcon/pilot/panel/_window_manager.py
```diff
@@ -92,7 +92,16 @@ def _tile(self):
...
+
+        for subwin in subwins:
+            if not subwin.isMinimized():
+                subwin.resize(400, 300)
```
Yuli-ytd: `400` by `300` matches the initial size used by the 2D and 3D sub-window factories. The minimized check avoids stretching the minimized window while leaving its minimized state intact.

tests/gui/test_window_menu.py
```diff
@@ -205,12 +205,66 @@ def test_tile_separates_the_windows(self):
...
+        self._two_stacked_windows()
+        subwins = self.area.subWindowList()
+
+        self.area.cascadeSubWindows()
+        QtWidgets.QApplication.processEvents()
+        qt_positions = [subwin.pos() for subwin in subwins]
+        self.assertNotEqual(qt_positions[0], qt_positions[1])
+
+        for subwin in subwins:
+            subwin.setGeometry(0, 0, 200, 150)
```
Yuli-ytd, line 218: The test captures Qt's native cascade positions first, then resets the windows before triggering the menu action. The reset ensures that the test fails if the action stops invoking Qt's cascade or changes its positions.

solvcon/pilot/panel/_window_manager.py
```diff
@@ -92,7 +92,16 @@ def _tile(self):
         self._mgr.mdiArea.tileSubWindows()
 
     def _cascade(self):
```
yungyuc: Allow overriding sub window width and height from function arguments:

```suggestion
    def _cascade(self, width=400, height=300):
```

It's OK to default to (400, 300).

solvcon/pilot/panel/_window_manager.py
yungyuc: It will work better to allow overriding from outside the function:

```suggestion
                subwin.resize(width, height)
```

yungyuc (changes requested):
- [x] Allow setting width and height when cascading.

solvcon/pilot/panel/_window_manager.py
Yuli-ytd: I updated _cascade() to accept optional width and height arguments, and added a GUI test covering custom values.

yungyuc (approved):
LGTM.

@Yuli-ytd Let me know if you want it to get merged.

Yuli-ytd:
@yungyuc I would like this PR to be merged.


> > > @Yuli-ytd Can you set up a Linux environment to check?
> > 
> > 
> > It's OK if not.
> 
> Sure. I’ll set up a Linux VM and test the same workflow.

I set up an Ubuntu 24.04 ARM64 VM on Apple Silicon to check the title-bar behavior on Linux. The pilot build encountered a GCC/NEON vector-type error in code unrelated to this PR, so I could not perform the Linux GUI check.

yungyuc:
> @yungyuc I would like this PR to be merged.

Merged.

> I set up an Ubuntu 24.04 ARM64 VM on Apple Silicon to check the title-bar behavior on Linux. The pilot build encountered a GCC/NEON vector-type error in code unrelated to this PR, so I could not perform the Linux GUI check.

In the original issue #1245, please leave a comment indicating the Linux and Windows behaviors have not been checked in this PR #1529.

PR #1531, opened 2026-09-10
Give the plot limits their named 2d type

j8xixo12:
The xy plot moved its data and view limits around as std::optional<std::array<double, 4>>, so every consumer read the bounds by position and a swapped index would draw a wrong frame without a diagnostic.  The bounds now live in PlotLimits2d, a mutable value with named xmin, xmax, ymin, and ymax, the union as merge, and defaulted equality; the series, the model, and the Python binding all speak the type, and set_view_limits accepts it as an overload beside the four scalars.

The dimension suffix is the point of the name: the API cannot change once delivered, so a future 3d plot arrives as a sibling type and new overloads instead of a retrofit of this one, following the 2d and 3d naming of ViewTransform2d and the geometry types.

Related to #1331, #1049

### DIFF
--- a/cpp/solvcon/pilot/plot/PlotLimits2d.hpp
+++ b/cpp/solvcon/pilot/plot/PlotLimits2d.hpp
@@ -1,0 +1,57 @@
+#pragma once
+
+/*
+ * Copyright (c) 2026, solvcon team <contact@solvcon.net>
+ * BSD 3-Clause License, see COPYING
+ */
+
+/**
+ * @file
+ * The named rectangular limits shared by the native xy plot data and view.
+ *
+ * @ingroup group_domain
+ */
+
+#include <algorithm>
+
+namespace solvcon
+{
+
+/**
+ * One xy rectangle, with named bounds rather than positional array entries.
+ * The dimension suffix keeps this delivered type honestly 2d: a future 3d
+ * plot adds a sibling type instead of changing this one.
+ */
+struct PlotLimits2d
+{
+    double xmin = 0.0;
+    double xmax = 0.0;
+    double ymin = 0.0;
+    double ymax = 0.0;
+
+    constexpr PlotLimits2d() = default;
+
+    constexpr PlotLimits2d(double xmin, double xmax, double ymin, double ymax)
+        : xmin(xmin)
+        , xmax(xmax)
+        , ymin(ymin)
+        , ymax(ymax)
+    {
+    }
+
+    void merge(PlotLimits2d const & other);
+
+    constexpr bool operator==(PlotLimits2d const &) const = default;
+}; /* end struct PlotLimits2d */
+
+inline void PlotLimits2d::merge(PlotLimits2d const & other)
+{
+    xmin = std::min(xmin, other.xmin);
+    xmax = std::max(xmax, other.xmax);
+    ymin = std::min(ymin, other.ymin);
+    ymax = std::max(ymax, other.ymax);
+}
+
+} /* end namespace solvcon */
+
+// vim: set ff=unix fenc=utf8 et sw=4 ts=4 sts=4:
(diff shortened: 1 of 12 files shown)


cpp/solvcon/pilot/plot/PlotLimits2d.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace solvcon
+{
+
+/**
+ * One xy rectangle, with named bounds rather than positional array entries.
+ * The dimension suffix keeps this delivered type honestly 2d: a future 3d
+ * plot adds a sibling type instead of changing this one.
+ */
+struct PlotLimits2d
```
tigercosmos, line 25: Just thinking maybe there is a better name for `PlotLimits2d`.

tigercosmos (approved):
LGTM

cpp/solvcon/pilot/plot/PlotLimits2d.hpp
```diff
@@ -0,0 +1,57 @@
...
+{
+    double xmin = 0.0;
+    double xmax = 0.0;
+    double ymin = 0.0;
+    double ymax = 0.0;
```
j8xixo12, line 30: Collect these variables into a structure so that users don't need to access the data limits through array indices. https://github.com/solvcon/solvcon/blob/bbdc16ab1fee2da6208efbcb4f64d77349441d1f/cpp/solvcon/pilot/plot/RPlotModel.hpp#L65

cpp/solvcon/pilot/plot/PlotLimits2d.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+    constexpr PlotLimits2d(double xmin, double xmax, double ymin, double ymax)
+        : xmin(xmin)
+        , xmax(xmax)
+        , ymin(ymin)
+        , ymax(ymax)
+    {
+    }
+
+    void merge(PlotLimits2d const & other);
```
j8xixo12, line 42: If two sets of plot limits need to be merged, this function can be used.

cpp/solvcon/pilot/plot/RPlotModel.cpp
```diff
@@ -140,10 +136,10 @@ ViewTransform2dFp64 RPlotModel::view(double width, double height) const
...
     }
 
-    double const zoom = std::min(width / (m_view_limits[1] - m_view_limits[0]),
-                                 height / (m_view_limits[3] - m_view_limits[2]));
-    double const center_x = 0.5 * (m_view_limits[0] + m_view_limits[1]);
-    double const center_y = 0.5 * (m_view_limits[2] + m_view_limits[3]);
+    double const zoom = std::min(width / (m_view_limits.xmax - m_view_limits.xmin),
+                                 height / (m_view_limits.ymax - m_view_limits.ymin));
+    double const center_x = 0.5 * (m_view_limits.xmin + m_view_limits.xmax);
+    double const center_y = 0.5 * (m_view_limits.ymin + m_view_limits.ymax);
```
j8xixo12, line 142: By using a structured data type, users no longer need to access plot limits through magic-number indices, it also improves code readability.

tests/test_pilot_plot.py
```diff
@@ -57,6 +62,33 @@ def test_color_is_an_immutable_value(self):
...
+        ser = _series([0.0, 1.0], [2.0, 3.0])
+        limits = ser.data_limits()
+        limits.xmin = -1.0
+        self.assertEqual(_limits(0.0, 1.0, 2.0, 3.0), ser.data_limits())
+
+        model = pilot.RPlotModel()
+        limits = model.view_limits()
+        limits.xmin = -1.0
+        self.assertEqual(_limits(0.0, 1.0, 0.0, 1.0), model.view_limits())
+
```
j8xixo12, line 91: Ensure that the `PlotLimits2d` object returned from c++ side is a independent  copy. Modifying the returned object should not affect the plot limits in original series or plot. The internal limits should only be updated explicitly through `set_view_limits`.

yungyuc (approved):
LGTM

PR #1532, opened 2026-09-10
Keep CI logic in contrib/ci scripts instead of workflow YAML

tigercosmos:
Adds a "Continuous Integration" section to CLAUDE.md and Rule 14 to contrib/prompt/general-rule.md. A workflow step's run block stays a few lines of glue. When the logic needs loops, conditionals, more than about ten lines, or a local run, the agent puts that logic in a script under contrib/ci/ and the workflow calls the script, as cache_cleanup.yml calls contrib/ci/prune-caches.sh.

A script in YAML cannot be linted, tested, or run outside CI, so long inline blocks are a maintenance liability. The rule also asks an agent that edits a step with a long inline script to move that script into contrib/ci/ as part of the change.

Codex reviewed the diff and reported no issues. make checkascii and make checktws pass.

[skip-ci]

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -298,6 +298,18 @@ any dependency requires user review and consent.
 **macOS**: System Integrity Protection (SIP) may interfere with
 `DYLD_LIBRARY_PATH`. The Makefile sets `PYTHONPATH` as a workaround.
 
+## Continuous Integration
+
+Workflows live in `.github/workflows/`; helper scripts live in `contrib/ci/`.
+Keep a workflow step's `run:` block short: a single command or a few lines of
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -298,6 +298,18 @@ any dependency requires user review and consent.
 **macOS**: System Integrity Protection (SIP) may interfere with
 `DYLD_LIBRARY_PATH`. The Makefile sets `PYTHONPATH` as a workaround.
 
+## Continuous Integration
```
tigercosmos, line 301: Add this part to teach AI how to contribute CI code.

Issue #1536, opened 2026-09-12
nightly_devbuild fails with startup_failure after #1528 (issues: write exceeds caller permissions)


rockleona:
Every scheduled nightly_devbuild run has failed immediately since 2026-09-10, with conclusion "startup_failure" and zero jobs executed (the Actions UI shows "This run likely failed because of a workflow file issue").
Example run: https://github.com/solvcon/solvcon/actions/runs/34641745733

The last successful nightly run was https://github.com/solvcon/solvcon/actions/runs/34398291058 (head 52d4f6bb, 2026-09-09). The first failing one was https://github.com/solvcon/solvcon/actions/runs/34523655082 (head bbdc16ab, 2026-09-10). The only workflow change between them is PR #1528 (commit d0681e4a, "Add open_issue_on_failure workflow for post-merge failures"), which added an open_issue_on_failure job to both devbuild.yml and lint.yml.

That job declares job-level permissions of issues: write and pull-requests: read, and calls ./.github/workflows/open_issue_on_failure.yml via workflow_call. devbuild.yml and lint.yml are themselves called via workflow_call from nightly_devbuild.yml, whose top-level permissions block grants only issues: read. A workflow_call chain cannot escalate permissions beyond what an ancestor caller grants, so GitHub rejects the whole run at plan time before any job starts, producing startup_failure with zero jobs.

The open_issue_on_failure job's own if condition restricts it to github.event_name == 'push', so it would never actually run under the nightly schedule or workflow_dispatch triggers. That guard does not help here: GitHub validates the permission ceiling for the whole call graph up front, regardless of whether a job's if would later skip it.

This is why the regular push-triggered devbuild.yml/lint.yml runs on master were unaffected and #1528 merged cleanly: those runs are not wrapped in another workflow_call, so there is no permission ceiling to violate. Only the nightly_devbuild.yml wrapper (added earlier) exposes the mismatch.

Net effect: nightly_devbuild has not produced a build or lint result since 2026-09-10.

Possible directions (left to the implementor):
- Grant nightly_devbuild.yml issues: write at the top level, accepting that the token could write issues even though only the push path uses it.
- Keep open_issue_on_failure out of the workflow_call chain that nightly_devbuild.yml uses, for example by not needing it when called with workflow_call, or moving the push-only reporting job somewhere nightly_devbuild.yml does not reuse.

PR #1537, opened 2026-09-12
Raise `nightly_devbuild.yml`'s `issues` permission to `write`

rockleona:
This grants `issues: write` in `nightly_devbuild.yml`'s top-level `permissions` block instead of `issues: read`.

The `open_issue_on_failure` job that `devbuild.yml` and `lint.yml` each nest, added by #1528, declares `issues: write`. `nightly_devbuild.yml` calls both workflows through `workflow_call` but only granted `issues: read`, so GitHub rejected the whole call graph before any job could start. Every scheduled and `workflow_dispatch` run has failed with `startup_failure` since 2026-09-10.

The job's own `if` restricts it to `github.event_name == 'push'`, so it never actually runs under this workflow's triggers, but GitHub checks the permission ceiling for the full call graph up front regardless of that condition. Raising the grant satisfies the check without changing what the job does.

Related to #1536.

### DIFF
--- a/.github/workflows/nightly_devbuild.yml
+++ b/.github/workflows/nightly_devbuild.yml
@@ -13,10 +13,6 @@ concurrency:
 
 # devbuild and lint both call check_skip_ci, which asks for these three, and
 # each also nests an open_issue_on_failure call that asks for issues: write.
-# A called workflow cannot hold more than its caller grants, and GitHub
-# checks that ceiling for the whole call graph before any job starts, so the
-# write scope is required here even though open_issue_on_failure's own `if`
-# keeps it from ever running under a schedule or workflow_dispatch trigger.
 permissions:
   contents: read
   pull-requests: read
--- a/.github/workflows/nightly_devbuild.yml
+++ b/.github/workflows/nightly_devbuild.yml
@@ -11,12 +11,16 @@ concurrency:
   group: nightly_devbuild
   cancel-in-progress: false
 
-# devbuild and lint both call check_skip_ci, which asks for these three, and a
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly_devbuild.yml
```diff
@@ -11,12 +11,16 @@ concurrency:
...
+# each also nests an open_issue_on_failure call that asks for issues: write.
+# A called workflow cannot hold more than its caller grants, and GitHub
+# checks that ceiling for the whole call graph before any job starts, so the
+# write scope is required here even though open_issue_on_failure's own `if`
+# keeps it from ever running under a schedule or workflow_dispatch trigger.
 permissions:
   contents: read
   pull-requests: read
-  issues: read
+  issues: write
```
rockleona, line 19: A single line change due to `open_issue_on_failure` workflow needs.

.github/workflows/nightly_devbuild.yml
```diff
@@ -11,12 +11,16 @@ concurrency:
   group: nightly_devbuild
   cancel-in-progress: false
 
-# devbuild and lint both call check_skip_ci, which asks for these three, and a
-# called workflow cannot hold more than its caller grants.
+# devbuild and lint both call check_skip_ci, which asks for these three, and
```
yungyuc, line 14: Make code comments concise and accurate.  Use PR comments / inline annotations to explain context.

yungyuc (changes requested):
- [x] Make code comments concise and accurate.

yungyuc (approved):
LGTM

PR #1548, opened 2026-09-16
Block AI attribution in commits and PRs

tigercosmos:
Add a hook that blocks a git or gh write when its message or body credits an AI agent. Examples are a `Co-authored-by` trailer naming an agent and a "Generated with" line.

The commit-code and create-pr skills require human-authored commit messages and PR bodies, but nothing enforced the rule. This change also states the rule in CLAUDE.md.

One script serves Claude Code, Codex, and Cursor. Codex runs the hook only after the user trusts it with `/hooks`.

The 56 tests in `tests/test_check_ai_attribution.py` pass. In a live run, the hook blocked a commit with a co-author trailer in all three agents.

PR #1556, opened 2026-09-18
Skip the attribution hook test outside a checkout

tigercosmos:
`tests/test_check_ai_attribution.py` skips itself when the repository tree is not next to `tests/`.

The `nouse_install` nightly job copies `tests/` out of the checkout and runs the copy against the installed package. The hook under `.claude/hooks/` is then out of reach, and all 56 tests fail on a missing file. The skip condition keys on the repository tree rather than on the hook, so a moved or renamed hook still fails the tests instead of skipping them.

In a source checkout the 56 tests pass. In a copy of `tests/` run from outside a checkout they skip. With the hook removed from a checkout they fail, so the skip does not mask a missing hook.

Related to #1548.


tests/test_check_ai_attribution.py
```diff
@@ -15,6 +15,12 @@
 HOOK = (pathlib.Path(__file__).resolve().parent.parent
         / ".claude" / "hooks" / "check-ai-attribution.py")
 
+# The nouse_install job copies tests/ out of the checkout and runs the copy
+# against the installed package, so the repository tree is out of reach. Key
+# the skip on that tree, not on HOOK, so a moved hook fails rather than skips.
+pytestmark = pytest.mark.skipif(not HOOK.parent.parent.is_dir(),
```
tigercosmos, line 21: Instead of modifying nouse_install, let's skip this test because there is no coverage difference for the AI hook test.

